PR #13, opened 2019-08-05
go/go.{mod,sum}: Fix mmap-go dependency version

mjesuele:
I changed line 17 of `go.mod`; the other changes were automatic.

Before this, I was getting this error when trying to use any of the go tools on this project on a fairly fresh installation of Linux Mint 19.2:

```
github.com/edsrzf/mmap-go@v1.0.0-20181222142022-904c4ced31cd: invalid pseudo-version: does not match version-control timestamp (2019-01-08T06:59:03Z)
```

Now it seems to be working.

### DIFF
--- a/go/Godeps/LICENSES
+++ b/go/Godeps/LICENSES
@@ -794,38 +794,6 @@ SOFTWARE.
 ================================================================================
 
 ================================================================================
-= github.com/edsrzf/mmap-go licensed under: =
-
-Copyright (c) 2011, Evan Shaw <edsrzf@gmail.com>
-All rights reserved.
-
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
-    * Redistributions of source code must retain the above copyright
-      notice, this list of conditions and the following disclaimer.
-    * Redistributions in binary form must reproduce the above copyright
-      notice, this list of conditions and the following disclaimer in the
-      documentation and/or other materials provided with the distribution.
-    * Neither the name of the copyright holder nor the
-      names of its contributors may be used to endorse or promote products
-      derived from this software without specific prior written permission.
-
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
-ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
-WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
-DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
-(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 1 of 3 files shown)


Hydrocharged:
This PR is reverting us from the latest "[version](https://github.com/edsrzf/mmap-go/tree/904c4ced31cdffe19e971afa0b3d319ff06d9c72)". There have been [two commits](https://github.com/edsrzf/mmap-go/commits/904c4ced31cdffe19e971afa0b3d319ff06d9c72) to [mmap-go](https://github.com/edsrzf/mmap-go) since the owner updated the `v1.0.0` tag, so it looks like this other stuff in [go.mod](https://github.com/liquidata-inc/dolt/blob/60c282ec8aea29ef9f127e8f0d0742ace21de118/go/go.mod#L17) is referring to the timestamp and commit hash.

Looking at the actual changes [I see that there is a fix](https://github.com/edsrzf/mmap-go/compare/188cc3b666ba704534fa4f96e9e61f21f1e1ba7c..904c4ced31cdffe19e971afa0b3d319ff06d9c72) implemented specifically around `FlushFileBuffers`, which is the error that we are now encountering (thanks to @mjesuele for bringing this to my attention).

Curiously I looked at the [liquidata fork](https://github.com/liquidata-inc/mmap-go/) and I see [a different fix](https://github.com/liquidata-inc/mmap-go/pull/1) for `FlushFileBuffers`.

So it looks like we can't get around the need for a fix for this specific issue, and [this PR](https://github.com/liquidata-inc/dolt/pull/13) would just revert us to a known-bad state. It looks like we're no longer using our fork, so here is a suggestion. We can purge [our changes](https://github.com/liquidata-inc/mmap-go/commit/df51bed0d3c488cd47f0540571b5212763632bc4) and rebase [our fork](https://github.com/liquidata-inc/mmap-go/) to the [latest](https://github.com/edsrzf/mmap-go/tree/904c4ced31cdffe19e971afa0b3d319ff06d9c72). From there we can make a release of our fork's latest, say `v1.0.1`, and change our [go.mod](https://github.com/liquidata-inc/dolt/blob/60c282ec8aea29ef9f127e8f0d0742ace21de118/go/go.mod#L17) to use that, which should bypass the pseudo-version error we are currently getting.

Pinging @bheni for input on this as well.

mjesuele:
Thanks for looking into this @Hydrocharged. Sounds reasonable to me. Looking at this again, it seems clear that some dodgy rebasing or something similar went on here: 
image
```
$ go get https://github.com/edsrzf/mmap-go@904c4ced31cdffe19e971afa0b3d319ff06d9c72
go: github.com/edsrzf/mmap-go@v1.0.0-20181222142022-904c4ced31cd: invalid pseudo-version: does not match version-control timestamp (2019-01-08T06:59:03Z)
```

mjesuele:
Okay, I've pushed our fork of mmap-go one commit past the upstream and tagged it `v1.0.2`: https://github.com/liquidata-inc/mmap-go/releases/tag/v1.0.2

It seems to have solved the original issue on my Linux system and no longer fails the Windows tests.

I doubt we'll need it, but I backed up our previous patch just in case: https://github.com/liquidata-inc/mmap-go/tree/mj/windows-fix-BAK

mjesuele:
Also, I suppose we should look into reporting this to `edsrzf/mmap-go` and/or the Go team...

mjesuele:
Yeah, good question. If we're not going to file an issue upstream (either because we don't care enough to spend the time or we judge that they don't care enough to address it), that's probably a factor in favor of us directly referencing our fork.

I'm going to merge this for now, but happy to continue discussion.

Hydrocharged:
There has been a [PR](https://github.com/edsrzf/mmap-go/pull/19) since Jan 8 to release a new version, so we would only be adding a +1 to it. I'm more in favor of us changing it to our fork completely now. I'll make a separate PR.

PR #20, opened 2019-08-08
go/utils: Improve checkcommitters; add prepr.sh.

reltuk:


go/utils/checkcommitters/fix_commiter.sh
```diff
@@ -0,0 +1,48 @@
...
+
+force=
+if [[ $# -gt 0 ]]; then
+    if [ "$1" = "-f" ]; then
+        force=`shift`
+    fi
+fi
+
+if [[ $# -ne 4 && $# -ne 2 ]]; then
+    echo "Usage: fix_commiter.sh [-f] TARGET_BRANCH WRONG_EMAIL [RIGHT_NAME RIGHT_EMAIL]" 1>&2
```
mjesuele: I assume we want this file to be named `fix_committer.sh`

go/utils/checkcommitters/main.go
```diff
@@ -55,23 +55,65 @@ var AllowedCommitters = map[string]*struct{}{
 // - This would be a place to enforce DCO or CLA requirements.
```
zachmu: Should we just turn on DCO bot? I did so for our fork of go-mysql-server, it's a one-click deal. It would mean anyone with outstanding commits would need to rebase.

reltuk: That's fine too, but if we enforce DCO, using this in prepr.sh is still nice?

PR #26, opened 2019-08-13
Add brew Formula for initial support

FanaHOVA:
First time creating a tap, but this should allow `brew tap liquidata-inc/dolt https://github.com/liquidata-inc/dolt.git`to allow MacOS users to install `dolt`. Should actually be able to tap from this PR too, testing in a sec.


FanaHOVA:
Closing until I have it working on my fork.

PR #31, opened 2019-08-15
support importing and exporting data to and from stdin and std out

bheni:
In the current releases it was possible to chain dolt with other programs via stdout/stdin like so:

    dolt table export table_name --file-type csv /dev/stdout -f|python row_cleaner.py|dolt table import cleaned_data -u --file-type csv /dev/stdin

Which only works in environments where stdin / stdout are mapped to files on the filesystem.  This change will use the stdin / stdout streams for import / export when a file is not provided.


bheni:
Need to go back through and comment some exported items, and update the help text for import / export.

go/libraries/doltcore/table/untyped/xlsx/writer.go
```diff
@@ -1,117 +0,0 @@
-// Copyright 2019 Liquidata, Inc.
```
bheni: This class never worked correctly.  It's been stripped out, and we'll need to implement.

go/libraries/doltcore/mvdata/stdio_data_loc.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return nil, false, errors.New(string(dl.Format) + "is an unsupported format to read from stdin")
+}
+
+// NewCreatingWriter will create a TableWriteCloser for a DataLocation that will create a new table, or overwrite
+// an existing table.
+func (dl StdIODataLocation) NewCreatingWriter(ctx context.Context, mvOpts *MoveOptions, root *doltdb.RootValue, fs filesys.WritableFS, sortedInput bool, outSch schema.Schema, statsCB noms.StatsCB) (table.TableWriteCloser, error) {
+	switch dl.Format {
+	case CsvFile:
+		return csv.NewCSVWriter(iohelp.NopWrCloser(cli.CliOut), outSch, csv.NewCSVInfo())
```
zachmu: It's odd to have a reference to the cli output in doltcore like this. Do we have to do it this way because we have previous replaced STDOUT? It would be better to use os.Stdout directly if you can think of a way.

go/cmd/dolt/cli/stdio.go
```diff
@@ -37,6 +38,7 @@ func InitIO() (restoreIO func()) {
 	if err == nil {
 		os.Stdout = f
 		os.Stderr = f
+		mvdata.SetIOStreams(os.Stdout, os.Stdin)
```
zachmu: Order wrong

go/libraries/doltcore/mvdata/data_loc.go
```diff
@@ -65,38 +83,38 @@ func (df DataFormat) ReadableStr() string {
...
+	NewCreatingWriter(ctx context.Context, mvOpts *MoveOptions, root *doltdb.RootValue, fs filesys.WritableFS, sortedInput bool, outSch schema.Schema, statsCB noms.StatsCB) (table.TableWriteCloser, error)
 
-	if fileFmtStr == "" {
+	// NewUpdatingWriter will create a TableWriteCloser for a DataLocation that will update and append rows based on
+	// their primary key.
+	NewUpdatingWriter(ctx context.Context, mvOpts *MoveOptions, root *doltdb.RootValue, fs filesys.WritableFS, srcIsSorted bool, outSch schema.Schema, statsCB noms.StatsCB) (table.TableWriteCloser, error)
+}
+
+// NewDataLocation creates a DataLocation object from a path and a format string.  If the path is the name of a table
+// then a TableDataLocation will be returned.  If the path is empty a StdioDataLocationo is returned.  Otherwise a
```
zachmu: typo StdioDataLocation

zachmu: Also looks like you renamed this to StreamDataLocation since this comment

go/libraries/doltcore/mvdata/stream_data_loc.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return nil, false, errors.New(string(dl.Format) + "is an unsupported format to read from stdin")
+}
+
+// NewCreatingWriter will create a TableWriteCloser for a DataLocation that will create a new table, or overwrite
+// an existing table.
+func (dl StreamDataLocation) NewCreatingWriter(ctx context.Context, mvOpts *MoveOptions, root *doltdb.RootValue, fs filesys.WritableFS, sortedInput bool, outSch schema.Schema, statsCB noms.StatsCB) (table.TableWriteCloser, error) {
+	switch dl.Format {
+	case CsvFile:
+		return csv.NewCSVWriter(dl.Writer, outSch, csv.NewCSVInfo())
```
zachmu: Don't you need to wrap this in a noopCloser so it doesn't attempt to close stdout?

(same might apply to the reader method, but not sure)

bheni: fixed

go/libraries/doltcore/mvdata/data_loc.go
bheni: fixed

PR #40, opened 2019-08-17
Fixed JSON imports and disallowed schemas on import updates

### LINKED ISSUE #36: Unable to update tables using JSON files
Currently importing a JSON file using `dolt table import` will ask for a schema even though the table exists. We should just read the existing table's schema if none is provided.

Hydrocharged:
Fixes #36


bats/update-tables.bats
```diff
@@ -0,0 +1,71 @@
...
+
+@test "update table using schema with csv" {
+    run dolt table create -s `batshelper 1pk5col-ints.schema` test
+    [ "$status" -eq 0 ]
+    run dolt table import -u -s `batshelper 1pk5col-ints.schema` test `batshelper 1pk5col-ints.csv`
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "schema is not supported for update operations" ]] || false
+}
+
+@test "update table using csv with newlines" {
```
Hydrocharged: @bheni I added a test here for importing CSVs with new lines too.

bheni: I don't think you want to conflate this with what you are addressing in this PR.

Hydrocharged: Okay, I can make this a separate one

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -291,7 +293,7 @@ func executeMove(dEnv *env.DoltEnv, force bool, mvOpts *mvdata.MoveOptions) int
 			return 1
 		}
 
-		if srcFileLoc.Format == mvdata.JsonFile && mvOpts.SchFile == "" {
+		if srcFileLoc.Format == mvdata.JsonFile && mvOpts.Operation != mvdata.UpdateOp && mvOpts.SchFile == "" {
```
bheni: I think it's easier to read == positiveCase rather than != negativeCase.

```suggestion
		if srcFileLoc.Format == mvdata.JsonFile && mvOpts.Operation == mvdata.OverwriteOp && mvOpts.SchFile == "" {
```

Hydrocharged: I didn't know you could suggest code changes! GitHub has so many features that I don't know about...but I removed the CSV test and this seems like a good change to make so I've committed it as well. I'll merge as soon as Jenkins passes!

PR #44, opened 2019-08-20
Changed default remote host to use the env constant

Hydrocharged:
Before we were using `dolthub.com` as the default, which is incorrect. I've changed it to the appropriate environment constant so that it also properly updates when we change from our beta domain.

### DIFF
--- a/go/cmd/dolt/commands/remote_test.go
+++ b/go/cmd/dolt/commands/remote_test.go
@@ -45,14 +45,14 @@ func TestGetAbsRemoteUrl(t *testing.T) {
 		{
 			"",
 			config.NewMapConfig(map[string]string{}),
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/commands/remote_test.go
```diff
@@ -45,14 +45,14 @@ func TestGetAbsRemoteUrl(t *testing.T) {
 		{
 			"",
 			config.NewMapConfig(map[string]string{}),
-			"https://dolthub.com",
```
reltuk: Should this just reference `env. DefaultRemotesApiHost`?

Hydrocharged: That's probably a better change for when this is changed from not being the beta site.

PR #45, opened 2019-08-20
Add support to get the last modified time from the filesys

bheni:


go/libraries/utils/filesys/inmemfs.go
```diff
@@ -108,8 +120,10 @@ func NewInMemFS(dirs []string, files map[string][]byte, cwd string) *InMemFS {
 				panic("Initializing InMemFS with invalid data.")
 			}
 
-			newFile := &memFile{path, val, targetDir}
+			now := time.Now()
```
zachmu: If we're going to the trouble of having an in-memory file system, we probably want to have it use a fake clock for this kind of thing. Maybe in a second pass? That would make it easier to write repeatable tests for mtime.

go/libraries/utils/filesys/inmemfs.go
```diff
@@ -460,17 +476,20 @@ func (fs *InMemFS) MoveFile(srcPath, destPath string) error {
 			return err
 		}
 
-		destObj := &memFile{destPath, obj.(*memFile).data, destParentDir}
+		now := time.Now()
+		destObj := &memFile{destPath, obj.(*memFile).data, destParentDir, time.Now()}
```
zachmu: Should use now instead of a new call to time.Now()

bheni: fixed

go/libraries/utils/filesys/inmemfs.go
bheni: Put in something quick and dirty.

PR #46, opened 2019-08-20
Add client_event.proto and compiled .go file

bheni:


proto/dolt/services/eventsapi/v1alpha1/client_event.proto
```diff
@@ -0,0 +1,86 @@
...
+    string value = 2;
+}
+
+message ClientEventMetric {
+    oneof metric_oneof {
+        google.protobuf.Duration duration = 1;
+        int32 count = 2;
+    }
+
+    MetricID metric_id = 100;
```
zachmu: Why the jump in tag number?

bheni: I was expecting what's inside the oneof to grow and I prefer to keep those consecutive... but really doesn't matter.

PR #52, opened 2019-08-23
Add metrics to command events push, pull, fetch, clone

coffeegoddd:
Added "standard" attributes to all command events.
Added a counter and timer metric to push, pull, fetch, clone. [Needs to be fixed, Need to get exact bytes transferred and not the approximation]

Will comment on each code section. Looking for feedback on general approach and metrics I should be generating.


go/cmd/dolt/commands/clone.go
```diff
@@ -91,6 +93,11 @@ func Clone(ctx context.Context, commandStr string, args []string, dEnv *env.Dolt
 				if verr == nil {
 					verr = cloneRemote(ctx, srcDB, remoteName, branch, dEnv)
 
+					if verr == nil {
+						evt := events.GetEventFromContext(ctx)
+						evt.SetAttribute(eventsapi.AttributeID_CLONE_REMOTE_URL, remoteUrl)
+					}
+
```
coffeegoddd: For `clone` command, my thinking is we probably want to see the url of the remote.

go/cmd/dolt/commands/clone.go
```diff
@@ -231,10 +241,11 @@ func cloneAllBranchRefs(branches []ref.DoltRef, srcDB *doltdb.DoltDB, ctx contex
 
 		progChan := make(chan datas.PullProgress)
 		doneChan := make(chan struct{})
-		go progFunc(progChan, doneChan)
+		go progFunc(progChan, doneChan, evt)
```
coffeegoddd: So I modified `progFunc` , in order to get the bytes transferred and the duration into the metrics and register the metrics into the `evt`, but I'm not sure this is the right move...? I think I need to dig a bit deeper to find the exact byte count and not reference the approximation for metrics?

go/cmd/dolt/commands/push.go
```diff
@@ -264,18 +268,40 @@ func pushToRemoteBranch(ctx context.Context, srcRef, destRef, remoteRef ref.Dolt
...
+
+				timer.Stop()
+				counter.Add(int32(bytesWritten))
+
+				evt.AddMetric(timer)
+				evt.AddMetric(counter)
 			}
+
 			latest = progress
+			bytesWritten = latest.ApproxWrittenBytes
```
coffeegoddd: This is my main entry point for adding the `BYTES_UP/DOWNLOADED` metric and the `UP/DOWNLOAD_MS_ELAPSED` metric. As you can see I'm relying on `latest.ApproxWrittenBytes` to give me the bytes transferred, but need to find where I can get the exact number.

go/cmd/dolt/dolt.go
```diff
@@ -131,7 +131,7 @@ func runMain() int {
 	emitter := createMetricsEmitter(dEnv)
 	defer func() {
 		ces := events.GlobalCollector.Close()
-		// events.WriterEmitter{cli.CliOut}.LogEvents(Version, ces)
+		events.WriterEmitter{cli.CliOut}.LogEvents(Version, ces)
 		_ = emitter.LogEvents(Version, ces)
 	}()
```
coffeegoddd: Will remove before merge.

go/libraries/events/events.go
```diff
@@ -57,6 +59,26 @@ func (evt *Event) SetAttribute(attID eventsapi.AttributeID, attVal string) {
...
+
+// GetClientEventType returns a pointer to the Client Event. This method is thread safe
+func (evt *Event) GetClientEventType() eventsapi.ClientEventType {
+	evt.m.Lock()
+	defer evt.m.Unlock()
+
+	return evt.ce.Type
+}
+
 func (evt *Event) close() *eventsapi.ClientEvent {
```
coffeegoddd: I added two helper methods I use to implement the above changes.

go/libraries/events/events.go
```diff
@@ -76,3 +98,26 @@ func (evt *Event) close() *eventsapi.ClientEvent {
...
+
+				evt.SetAttribute(eventsapi.AttributeID_LOCAL_REMOTE_URLS, strings.Join(urls[:], " "))
+			}
+		}
+
+		if len(evt.GetAttribute(eventsapi.AttributeID_BRANCH_NAME)) < 1 {
+			evt.SetAttribute(eventsapi.AttributeID_BRANCH_NAME, dEnv.RepoState.Head.Ref.GetPath())
+		}
+	}
+}
```
coffeegoddd: I'm not sure this is the right file for this code? Anyway, I figure with every command, we want to the current branch and the remote url added to the `evt.attributes` map. 

However, I wasn't sure what my strategy should be for storing a remote in the `evt.attributes` map. There may be many remotes, there may or may not be an 'origin' remote, and I'm not sure if we want the get the information for all local remotes. That said, I need to define the key for this in the `.proto` and store the value as a string. So for now, I'm just making a larger string and storing it under `LOCAL_REMOTE_URLS`. Not sure this is the ideal solution either...

coffeegoddd:
Outdated

PR #62, opened 2019-08-29
Progress on switching inserts to new engine

Hydrocharged:
I don't think I'll need to necessarily add anything more to this PR as I think the rest of my changes will be in go-mysql-server.


Hydrocharged:
Changed the `Inserter` interface. On my other PR I wrote a few more details: https://github.com/liquidata-inc/go-mysql-server/pull/8#issuecomment-526412916

go/libraries/doltcore/sqle/tables.go
```diff
@@ -96,6 +99,56 @@ func (t *DoltTable) PartitionRows(ctx *sql.Context, _ sql.Partition) (sql.RowIte
...
+	dRow, err := SqlRowToDoltRow(t.table.Format(), sqlRow, t.sch)
+	if err != nil {
+		return false, err
+	}
+
+	taggedVals := row.TaggedValues{}
+	for _, pkTag := range t.sch.GetPKCols().Tags {
+		dCol, ok := dRow.GetColVal(pkTag)
+		if !ok {
+			return false, errors.New("error: primary key missing from values")
```
zachmu: No need to prefix these error messages with "error:"

go/libraries/doltcore/sqle/types.go
```diff
@@ -80,45 +81,135 @@ func nomsValToSqlVal(val types.Value) interface{} {
...
-		return types.Uint(e)
-	case uint64:
-		return types.Uint(e)
-	case float32:
-		return types.Float(e)
-	case float64:
-		return types.Float(e)
-	case string:
+	switch kind {
+	case types.BoolKind:
```
zachmu: So go-mysql-server has logic for doing this kind of conversion which you should copy or use. It would be weird for us to evaluate boolean values differently on insert than in other contexts. Check it out: 

https://github.com/src-d/go-mysql-server/blob/master/sql/type.go#L698

The same logic applies to all these things: you should use the go-mysql-server logic to get as close to the dolt type as possible, then apply any dolt-specific logic (like unsigned, uuid, etc)

go/cmd/dolt/commands/sql_test.go
```diff
@@ -272,42 +272,77 @@ func TestInsert(t *testing.T) {
 			query: "insert into dne (id) values (00000000-0000-0000-0000-000000000005)", expectedRes: 1,
 		},
 		{
-			name: "insert one row",
-			query: `insert into people (id, name, age, is_married) values 
+			name:  "insert one row",
```
zachmu: These command line tests are good to have, but they really shouldn't be where the main body of functional tests live -- these are really more like smoke tests to make sure that everything is wired together correctly. The exhaustive correctness tests should live in the package being exercised. Basically, copy the pattern I did with sqlselect_test.go, and apply it to sqlinsert_test.go.

go/libraries/doltcore/sqle/types.go
Hydrocharged: Ohhh I wasn't aware of this, I can make the changes.

go/libraries/doltcore/sql/sqltestutil/insertqueries.go
```diff
@@ -0,0 +1,126 @@
...
+	{
+		Name:           "insert partial columns multiple values",
+		InsertQuery:    "insert into people (id, first, last) values (0, 'Bart', 'Simpson'), (1, 'Homer', 'Simpson')",
+		SelectQuery:    "select id, first, last from people where id < 2 order by id",
+		ExpectedRows:   Rs(NewResultSetRow(types.Int(0), types.String("Bart"), types.String("Simpson")),
+			NewResultSetRow(types.Int(1), types.String("Homer"), types.String("Simpson"))),
+		ExpectedSchema: NewResultSetSchema("id", types.IntKind, "first", types.StringKind, "last", types.StringKind),
+	},
+	{
+		Name:        "insert partial columns existing pk",
```
zachmu: Should also have a test that uses AdditionalSetupFn to set up a database that already has the row being duplicated before the insert statement.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -96,6 +99,55 @@ func (t *DoltTable) PartitionRows(ctx *sql.Context, _ sql.Partition) (sql.RowIte
...
+		return  err
+	}
+
+	taggedVals := row.TaggedValues{}
+	for _, pkTag := range t.sch.GetPKCols().Tags {
+		dCol, ok := dRow.GetColVal(pkTag)
+		if !ok {
+			return errors.New("primary key missing from values")
+		}
+		dColVal, err := dCol.Value(ctx)
```
zachmu: No need to assemble separate tagged values here. You can use drow.NomsMapKey to get the tuples you need for GetRowByPKVals

zachmu: Actually it's GetRowByPKTuple, and you will need to cast the result of NomsMapKey to call it. But it's better than doing additional IO to assemble the noms key.

go/libraries/doltcore/sqle/types.go
```diff
@@ -80,45 +82,142 @@ func nomsValToSqlVal(val types.Value) interface{} {
...
+		case int8:
+			return types.Int(e), nil
+		case int16:
+			return types.Int(e), nil
+		case int32:
+			return types.Int(e), nil
+		case int64:
+			return types.Int(e), nil
+		case uint:
+			if e > 9223372036854775807 {
```
zachmu: Probably should be a constant for this

zachmu:
Brian, could you take a look at the interface change to PutTable and see what you think?

go/libraries/doltcore/sqle/tables.go
Hydrocharged: `GetRowByPKTuple` doesn't seem to exist anywhere in the codebase? I just did a full directory scan for that function and couldn't find anything.

Hydrocharged: Nvm, the function doesn't exist but I looked up how `NomsMapKey` was used and figured it out from there.

PR #67, opened 2019-09-03
Add file emitter that writes event data file

coffeegoddd:
Added file emitter that saves event data to files, and a flush that parses the files and sends them to the grpc server.


go/cmd/dolt/commands/send_metrics.go
```diff
@@ -0,0 +1,31 @@
...
+
+	dolt := dbfactory.DoltDir
+
+	eventGrpcFlush := events.NewEventGrpcFlush(fs, root, dolt, dEnv)
+	if err := eventGrpcFlush.FlushEvents(); err != nil {
+		// log error?
+		return 1
+	}
+	return 0
+}
```
coffeegoddd: `SendMetrics` is the `commandFunc` for the dolt command "dolt send-metrics". Looking for feedback on how to handle the error logging here. This also gets run in a separate process on all other dolt commands using `os.exec`

go/cmd/dolt/dolt.go
```diff
@@ -144,32 +153,52 @@ func runMain() int {
...
+	// 		log.Println(color.YellowString("The config value of '%s' is '%s' which is not a valid true/false value", env.MetricsInsecure, *insecureStr))
+	// 	}
 
-		return events.NewGrpcEmitter(conn)
-	}
+	// 	hostAndPort := fmt.Sprintf("%s:%d", *host, port)
+	// 	conn, _ := dEnv.GrpcConnWithCreds(hostAndPort, insecure, nil)
+
+	// 	return events.NewGrpcEmitter(conn)
+	// }
```
coffeegoddd: Commented this code out just during development while working on the `FileEmitter`. 
It seems like this function is used to handle the creation of the default emitter we want, based on the various emitters we've created? 
Do we want the `FileEmitter` to now be the default emitter we instantiate?

go/cmd/dolt/dolt.go
```diff
@@ -144,32 +153,52 @@ func runMain() int {
...
+				if err := cmd.Start(); err != nil {
+					// log error
+					return err
+				}
+			}
+			return err
+		}
+	}
+	return nil
+}
```
coffeegoddd: Running the hidden command "send-metrics" on a new process for every other dolt command. This block does not run if the user types it in herself, since otherwise it would loop infinitely.

go/go.sum
```diff
@@ -195,6 +195,7 @@ github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/lib/pq v1.0.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
 github.com/lib/pq v1.1.1 h1:sJZmqHoEaY7f+NPP8pgLB/WxulyR3fewgCM2qaSlBb4=
 github.com/lib/pq v1.1.1/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
+github.com/liquidata-inc/dolt v0.9.8 h1:uup9UA59oSvYsqYd9A3zF0CFXUnF1qI8odE+Cxtn+pg=
 github.com/liquidata-inc/go-mysql-server v0.4.1-0.20190710171053-b2883167103a h1:Z8uOEqaBW1a0vpu+S9MlWwiyIDrjtn9sZ/CCcs32e9s=
 github.com/liquidata-inc/go-mysql-server v0.4.1-0.20190710171053-b2883167103a/go.mod h1:bXirMMsoAKR1OI9GTQJPGSyuwJlOseI9AFNAdpwRY2k=
 github.com/liquidata-inc/ishell v0.0.0-20190514193646-693241f1f2a0 h1:phMgajKClMUiIr+hF2LGt8KRuUa2Vd2GI1sNgHgSXoU=
```
coffeegoddd: I'm not to sure how the `go.sum` works yet, which makes me pretty nervous. I don't touch this file, but it seems to change at times automatically... Planning to look into this.

go/libraries/events/directory_proc.go
```diff
@@ -0,0 +1,208 @@
...
+	for !empty {
+		select {
+		case completed := <-compCh:
+			delete(beingProcessed, completed)
+
+		default:
+			empty = true
+		}
+	}
+}
```
coffeegoddd: Same code from the events server. Just changed the log extension variable to be `evtDataExt`.
Will add the test file as well.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// insecure, err := strconv.ParseBool(*insecureStr)
+
+	// if err != nil {
+	// 	log.Println(color.YellowString("The config value of '%s' is '%s' which is not a valid true/false value", env.MetricsInsecure, *insecureStr))
+	// }
+
+	// hostAndPort := fmt.Sprintf("%s:%d", *host, port)
+	// conn, _ := dEnv.GrpcConnWithCreds(hostAndPort, insecure, nil)
+
```
coffeegoddd: This is the code used to connect to the actual grpc service. Will comment back in before merge.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,147 @@
...
+// NewEventGrpcFlush creates a new EventGrpcFlush
+func NewEventGrpcFlush(fs filesys.Filesys, userHomeDir string, doltDir string, dEnv *env.DoltEnv) *EventGrpcFlush {
+
+	fbp := NewFileBackedProc(fs, userHomeDir, doltDir)
+
+	if exists := fbp.EventsDirExists(); !exists {
+		// should panic?
+		panic(ErrEventsDataDir)
+	}
+
```
coffeegoddd: Not sure if we should panic here? This panic would occur in the separate process.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,147 @@
...
+// FlushEvents sends event logs to the events server
+func (egf *EventGrpcFlush) FlushEvents() error {
+	dp := NewDirectoryProcessor(egf.fbp.ed.fs, egf.fbp.GetEventsDirPath(), 4, time.Second, time.Minute, egf.sendToServer)
+	time.Sleep(2 * time.Second)
+	if err := dp.Close(); err != nil {
+		return err
+	}
+	return nil
+}
```
coffeegoddd: Think I need to implement a default params struct similar to how we do in the events server? That way we can use the default params, but later we can easily supply new ones...

go/cmd/dolt/dolt.go
```diff
@@ -128,11 +129,19 @@ func runMain() int {
 
 	dEnv := env.Load(context.TODO(), env.GetCurrentUserHomeDir, filesys.LocalFS, doltdb.LocalDirDoltDB)
 
+	if err := processEventsDir(args, dEnv); err != nil {
+		// log err
+		log.Fatal(err)
```
bheni: We shouldn't ever let metrics failures fail dolt commands.

go/go.sum
bheni: go.sum works via witchcraft

go/libraries/events/directory_proc.go
bheni: I think this might be a bit of overkill and could result in a ton of send-metrics processes running when someone is working in offline mode.  I think instead of watching the directory and looping as long and there are still files, you should just loop over all files until you have a failure to send, and then exit on first failure.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// if err != nil {
+	// 	log.Println(color.YellowString("The config value of '%s' is '%s' which is not a valid true/false value", env.MetricsInsecure, *insecureStr))
+	// }
+
+	// hostAndPort := fmt.Sprintf("%s:%d", *host, port)
+	// conn, _ := dEnv.GrpcConnWithCreds(hostAndPort, insecure, nil)
+
+	// local grpc for dev
+	hostAndPort := fmt.Sprintf("%s:%d", "localhost", 50051)
```
bheni: you can connect locally without having to hard code this.  You can run these commands:

    dolt config --global --add metrics.host localhost
    dolt config --global --add metrics.port 50051
    dolt config --global --add metrics.insecure true

in order to point your dolt cli at your local metrics server.

go/libraries/events/event_flush.go
bheni: It's probably fine.

go/cmd/dolt/dolt.go
```diff
@@ -128,11 +129,17 @@ func runMain() int {
...
+
 	defer func() {
 		ces := events.GlobalCollector.Close()
 		// events.WriterEmitter{cli.CliOut}.LogEvents(Version, ces)
 		_ = emitter.LogEvents(Version, ces)
+
+		if err := processEventsDir(args, dEnv); err != nil {
+			log.Print(err)
+		}
 	}()
```
coffeegoddd: Update: This is no longer a `log.Fatal`

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,145 @@
...
+// FlushEvents sends event logs to the events server
+func (egf *EventGrpcFlush) FlushEvents() error {
+	fs := egf.fbp.GetFileSys()
+
+	if err := fs.Iter(egf.fbp.GetEventsDirPath(), false, egf.flush); err != nil {
+		return err
+	}
+
+	return nil
+}
```
coffeegoddd: Much simpler implementation now using `fs.Iter()` and the `flush()` method.

go/cmd/dolt/dolt.go
```diff
@@ -144,32 +151,56 @@ func runMain() int {
...
+
+func processEventsDir(args []string, dEnv *env.DoltEnv) error {
+	if len(args) > 0 {
+		if args[0] != commands.SendMetricsCommand {
+			disabled, err := events.AreMetricsDisabled(dEnv)
+			if !disabled && err == nil {
+				cmd := exec.Command("dolt", commands.SendMetricsCommand)
+
+				if err := cmd.Start(); err != nil {
+					log.Print(err)
```
bheni: I think we should fail silently until a later date when we support a verbose logging mode.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,145 @@
...
+func (egf *EventGrpcFlush) flush(path string, size int64, isDir bool) (stop bool) {
+	fs := egf.fbp.GetFileSys()
+
+	data, err := fs.ReadFile(path)
+	if err != nil {
+		log.Print(err)
+		return false
+	}
+
+	isFileValid, err := checkFilenameAndExt(data, path)
```
bheni: if isFileValid is false, and err is nil that means the file is corrupted.  Retries on this file won't fix it, so we should go ahead and delete it.

go/libraries/events/file_backed_proc.go
```diff
@@ -0,0 +1,186 @@
...
+		Platform:  plat,
+		Events:    evts,
+	}
+
+	data, err := proto.Marshal(req)
+	if err != nil {
+		return err
+	}
+
+	if _, err := f.Write(data); err != nil {
```
bheni: it's possible that Write will write less than the entire buffer.  iohelp.WriteAll will take care of flushing the entire contents of data.

go/libraries/events/file_backed_proc.go
```diff
@@ -0,0 +1,216 @@
...
+
+	data, err := proto.Marshal(req)
+	if err != nil {
+		return err
+	}
+
+	if err := iohelp.WriteAll(f, data); err != nil {
+		return err
+	}
+
```
coffeegoddd: updated to ioutil.WriteAll()

go/cmd/dolt/dolt.go
```diff
@@ -143,33 +149,22 @@ func runMain() int {
...
+			cmd := exec.Command("dolt", commands.SendMetricsCommand)
 
-		if err != nil {
-			log.Println(color.YellowString("The config value of '%s' is '%s' which is not a valid port.", env.MetricsPort, *portStr))
-			return emitter
-		}
-
-		insecure, err := strconv.ParseBool(*insecureStr)
+			if err := cmd.Start(); err != nil {
+				log.Print(err)
```
bheni: disable logging until we have verbose mode.

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,128 @@
...
+	data, err := fs.ReadFile(path)
+	if err != nil {
+		log.Print(err)
+		return false
+	}
+
+	isFileValid, err := egf.fbp.CheckingFunc(data, path)
+
+	if isFileValid && err == nil {
+		ctx, cnclFn := context.WithDeadline(context.Background(), time.Now().Add(time.Second+500*time.Millisecond))
```
bheni: Since we are running in a new process we can up this to time.Minute

go/libraries/events/event_flush_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"path/filepath"
+	"testing"
+
+	eventsapi "github.com/liquidata-inc/dolt/go/gen/proto/dolt/services/eventsapi_v1alpha1"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+)
+
+var globalEvents []*eventsapi.ClientEvent
```
bheni: using a global will be problematic if 2 of these tests run concurrently.  I'd make this a member of the TestClient

go/libraries/events/event_flush_test.go
```diff
@@ -0,0 +1,85 @@
...
+	eventsapi "github.com/liquidata-inc/dolt/go/gen/proto/dolt/services/eventsapi_v1alpha1"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+)
+
+type TestClient struct {
+	CES []*eventsapi.ClientEvent
+}
+
```
coffeegoddd: no longer using the global events slice.

go/cmd/dolt/commands/send_metrics.go
```diff
@@ -0,0 +1,48 @@
...
+	disabled, err := events.AreMetricsDisabled(dEnv)
+	if !disabled && err == nil {
+		fs := filesys.LocalFS
+
+		root, err := env.GetCurrentUserHomeDir()
+		if err != nil {
+			// log.Print(err)
+			return 1
+		}
+
```
coffeegoddd: logging commented out in both `dolt.go` and `send_metrics.go`

go/libraries/events/event_flush.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Print(err)
+		return false
+	}
+
+	isFileValid, err := egf.fbp.CheckingFunc(data, path)
+
+	if isFileValid && err == nil {
+		ctx, cnclFn := context.WithDeadline(context.Background(), time.Now().Add(time.Minute))
+		defer cnclFn()
+
```
coffeegoddd: changed to time.Minute

go/libraries/events/file_backed_proc.go
```diff
@@ -0,0 +1,235 @@
...
+	if !exists {
+		if err := fs.WriteFile(lp, []byte("lockfile for dolt \n")); err != nil {
+			panic(err)
+		}
+	}
+
+	return &FileBackedProc{ed: eventsDataDir, namingFunc: nf, CheckingFunc: cf, LockPath: lp}
+}
+
```
coffeegoddd: This is the new implementation of the locking system. Instead of locking the entire events directory, I create a `dolt.lock` file if it does not exist, then lock/unlock this file.

PR #71, opened 2019-09-05
Db/events dir lock

coffeegoddd:
initial implementation of making event flush concurrency safe


go/cmd/dolt/commands/send_metrics.go
```diff
@@ -18,21 +18,22 @@ const SendMetricsCommand = "send-metrics"
...
-		eventGrpcFlush := events.NewEventGrpcFlush(fs, root, dolt, dEnv)
+		egf := events.NewEventGrpcFlush(dEnv.FS, root, dolt, dEnv)
 
-		if err := eventGrpcFlush.FlushEvents(); err != nil {
-			// log.Print(err)
-			return 1
+		err = egf.FlushEvents(ctx)
+		if err != nil {
+			return 2
 		}
```
coffeegoddd: Now handling the filesys' in a type switch inside `FlushEvents`

go/libraries/events/file_backed_proc.go
```diff
@@ -175,42 +182,42 @@ func (fbp *FileBackedProc) WriteEvents(version string, evts []*eventsapi.ClientE
 		plat = eventsapi.Platform_WINDOWS
 	}
 
-	if err := fbp.ed.MakeEventsDir(); err != nil {
-		return err
-	}
+	if dirExists := fbp.EventsDirExists(); dirExists {
+		eventsPath := fbp.ed.getPath()
+		tempFilename := filepath.Join(eventsPath, localPath)
```
coffeegoddd: I was having an issue by not creating the events directory until `WriteEvents()` was called, so I changed that.
I now create the directory when I instantiate a `FileBackedProc`, and just ensure the directory exists before I write the files.

go/libraries/events/event_flush.go
```diff
@@ -29,15 +29,21 @@ import (
...
 
-// ErrEventsDataDir occurs when events are trying to be  flushed, but the events data directory
-// does not yet exist
-var ErrEventsDataDir = errors.New("unable to flush, events data directory does not exist")
+// GrpcEventFlusher parses dolt event logs sends the events to the events server
+type GrpcEventFlusher struct {
+	em       *GrpcEmitter
+	fbp      *FileBackedProc
+	LockPath string
+}
```
coffeegoddd: Now using name `GrpcEventFlusher`

go/libraries/utils/filesys/lock.go
```diff
@@ -0,0 +1,91 @@
...
+	// 	return nil
+	// }
+
+	if new != 0 {
+		return errLockUnlock
+	}
+
+	return nil
+}
+
```
coffeegoddd: Looked at the implementation of `Mutex` lock and unlock methods to see how to use this. I think I'm correct or close. A test would confirm this possibly incorrect suspicion of this being close.

go/libraries/events/event_flush_test.go
```diff
@@ -73,13 +95,62 @@ func TestEF(t *testing.T) {
...
 
-			err = flush.FlushEvents()
+			err = flushTester.Flusher.FlushEvents(ctx)
 
 			assert.Equal(t, err, nil)
-			assert.Equal(t, len(client.CES), len(ces))
+			assert.Equal(t, len(flushTester.Client.CES), len(ces))
 		})
 	}
 }
```
coffeegoddd: This is the new test for the local filesys. I will figure out a way to not duplicate test logic here as these tests are largely the same...

bats/send_metrics.bats
```diff
@@ -0,0 +1,47 @@
...
+@test "create events data then flush data to stdout non-concurrently" {
+    run dolt ls
+    [ "$status" -eq 0 ]
+    run sleep 1
+    run dolt send-metrics --output
+    [ "$status" -eq 0 ]
+}
+
+# Create events data then flush data to stdout concurrently
+@test "create events data then flush data to stdout concurrently" {
```
bheni: I feel like this test could be flaky.  I believe it is supposed to do the following:

1. "run dolt ls" creates an event and kicks off a new process to send the event
2. The new process created by the ls command grabs the lock and sends the event
3. "run dolt send-metrics" tries to acquire the lock and fails.
4. "run dolt send-metrics" tries to acquire the lock and fails.

This could fail if 2 & 3 get run in a different order, or if 2 completes before 3 starts.

I think I would do the following 

Prep work:
Create multiple event files in the helper directory and check those in to source control

Then in the test:
1. copy the events to the place where dolt expects them.
2. "run dolt send-metrics --output >file1.txt &"
3. pid1=$!
4. "run dolt send-metrics --output >file2.txt &"
5. pid2=$!
6. wait $pid1
7. exit_code1=$?
8. wait $pid2
9. exit_code2=$?
10. check results.  something like:

if $exit_code1 = 0; then
  if $exit_code2 = 0; then
    # test file1 and file2 for some text like "0 events processed" and make sure the process writes that out
  elif exit_code2 != 2; then
    # fail test
  fi
elif $exit_code2 != 0; then
  # fail test
else
  if exit_code1 != 2; then
    # fail
  fi
fi

go/cmd/dolt/commands/send_metrics.go
```diff
@@ -2,47 +2,74 @@ package commands
...
+	if err != nil {
+		// log.Print(err)
+		return 1
+	}
+
+	if disabled {
+		return 0
+	}
+
 	if !disabled && err == nil {
```
bheni: dont need err == nil condition here anymore.  handled already above

go/cmd/dolt/commands/send_metrics.go
```diff
@@ -2,47 +2,74 @@ package commands
...
+		if len(args) > 0 {
+			if args[0] == outputFlag {
+				iof := events.NewIOFlusher(dEnv.FS, root, dolt, dEnv)
+
+				err := iof.Flush(ctx)
+				if err != nil {
+					if err == events.ErrFileLocked {
+						return 2
+					}
+
```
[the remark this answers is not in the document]
bheni: rather than having 2 code paths that do similar things, its better to do something like

```
var flusher Flusher
if outputFlag {
    flusher = events.NewIOFlusher(dEnv.FS, root, dolt, dEnv)
}
else {
    flusher = events.NewGrpcEventFlusher(dEnv.FS, root, dolt, dEnv)
}

err = flusher.Flush(ctx)
... etc
```

go/libraries/events/event_flush.go
```diff
@@ -64,63 +75,167 @@ func getGRPCEmitter(dEnv *env.DoltEnv) *GrpcEmitter {
...
 	if err != nil {
-		log.Print(err)
-		return false
+		return err
 	}
 
 	isFileValid, err := egf.fbp.CheckingFunc(data, path)
 
 	if isFileValid && err == nil {
-		ctx, cnclFn := context.WithDeadline(context.Background(), time.Now().Add(time.Minute))
```
bheni: I believe the context already has a deadline set in the caller.  Probably can skip and just use the ctx object

go/libraries/utils/filesys/lock.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"sync/atomic"
+
+	"github.com/juju/fslock"
+	"github.com/pkg/errors"
+)
+
+const originalStateValue int32 = 0
+const newStateValue int32 = 1
```
bheni: lets rename to 

unlockedStateValue = 0
lockedStateValue = 1

go/libraries/utils/filesys/lock.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// NewInMemFileLock creates a new InMemFileLock
+func NewInMemFileLock(fs Filesys) *InMemFileLock {
+	return &InMemFileLock{originalStateValue}
+}
+
+// TryLock attempts to lock the lock or fails if it is already locked
+func (memLock *InMemFileLock) TryLock() (bool, error) {
+	if atomic.CompareAndSwapInt32(&memLock.state, originalStateValue, newStateValue) {
```
bheni: nice!

go/cmd/dolt/commands/send_metrics.go
```diff
@@ -2,47 +2,74 @@ package commands
...
+	"time"
 
+	"github.com/fatih/color"
+	"github.com/liquidata-inc/dolt/go/cmd/dolt/cli"
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/dbfactory"
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/env"
 	"github.com/liquidata-inc/dolt/go/libraries/events"
-	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/argparser"
 )
```
coffeegoddd: now using ArgParser

bats/send_metrics.bats
```diff
@@ -0,0 +1,64 @@
...
+      if [ $exit_code2 -eq 0 ]; then
+        # we expect for only one output file to contain 4 lines, corresponding to the 4 event files successfully processed
+        # check that the line counts of the output files match what is expected
+        if [[ "$event_count1" = *"4 file1.txt" ]] && [[ "$event_count2" = *"0 file2.txt" ]]; then
+            return 0
+        fi
+        if [[ "$event_count1" = *"0 file1.txt" ]] && [[ "$event_count2" = *"4 file2.txt" ]]; then
+            return 0
+        fi
+        return 1
```
coffeegoddd: Now working!!

bats/send_metrics.bats
```diff
@@ -0,0 +1,89 @@
...
+    if [[ "$event_count1" = *"4 file1.txt" ]] && [[ "$event_count2" = *"0 file2.txt" ]]; then
+      return 0
+    fi
+    if [[ "$event_count1" = *"0 file1.txt" ]] && [[ "$event_count2" = *"4 file2.txt" ]]; then
+      return 0
+    fi
+
+    return 1
+}
+
```
coffeegoddd: New and improved test.. It's the cleaned up version of the commented out test below. Basically I removed the additional if else logic since we are suppressing the status codes from the wait command and just checking the file line length yet.

go/cmd/dolt/dolt.go
```diff
@@ -133,8 +133,21 @@ func runMain() int {
...
+			return
+		}
+
+		// write events
 		_ = emitter.LogEvents(Version, ces)
 
+		// flush events
 		if err := processEventsDir(args, dEnv); err != nil {
 			// log.Print(err)
 		}
```
coffeegoddd: No longer writing or flushing events if metrics.disabled is true.

go/cmd/dolt/dolt.go
```diff
@@ -151,17 +164,25 @@ func runMain() int {
 // processEventsDir runs the dolt send-metrics command in a new process
 func processEventsDir(args []string, dEnv *env.DoltEnv) error {
 	if len(args) > 0 {
-		if args[0] != commands.SendMetricsCommand {
-			cmd := exec.Command("dolt", commands.SendMetricsCommand)
+		ignoreCommands := map[string]struct{}{
+			commands.SendMetricsCommand: struct{}{},
+			"init":                      struct{}{},
+			"config":                    struct{}{},
+		}
```
coffeegoddd: map of commands we do not want to kick off a flush call

PR #81, opened 2019-09-13
Progress on switching deletes to new engine

Hydrocharged:
Currently works for deletes but not thoroughly testing.


Hydrocharged:
Not including the special casing logic for plain deletes, I'll make a separate PR for that.

bats/1pk5col-ints.bats
```diff
@@ -295,24 +295,24 @@ if rows[2] != "9,8,7,6,5,4".split(","):
     dolt sql -q "insert into test (pk,c1,c2,c3,c4,c5) values (0,1,2,3,4,5),(1,11,12,13,14,15),(2,21,22,23,24,25)"
     run dolt sql -q "delete from test where pk=2"
     [ "$status" -eq 0 ]
-    [ "$output" = "Rows deleted: 1" ]
+    [[ "$output" =~ "| 1       |" ]] || false
```
zachmu: I think this is fine, but what's the full output from the engine? Is the column named "rows deleted" or similar? Looks like you need to change this on your go-mysql-server PR.

PR #82, opened 2019-09-16
Naked deletes now just delete everything instead of iterating

Hydrocharged:
I mean this works but it's ugly and I'm not sure of a better way to do it really

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -387,7 +387,17 @@ func processQuery(ctx context.Context, query string, dEnv *env.DoltEnv, root *do
 			err = prettyPrintResults(ctx, root.VRW().Format(), sqlSch, rowIter)
 		}
 		return nil, err
-	case *sqlparser.Insert, *sqlparser.Delete:
+	case *sqlparser.Insert:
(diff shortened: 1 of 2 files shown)


Hydrocharged:
Also, darn it Zach but an empty map was causing a panic lol

go/cmd/dolt/commands/sql.go
```diff
@@ -684,6 +694,38 @@ func sqlDelete(ctx context.Context, dEnv *env.DoltEnv, root *doltdb.RootValue, u
...
+					newTable, err := table.UpdateRows(ctx, emptyMap)
+					if err != nil {
+						return nil, false
+					}
+					newRoot, err := doltdb.PutTable(ctx, root, root.VRW(), tName, newTable)
+					if err != nil {
+						return nil, false
+					}
+					// This is so the results are displayed as expected
+					if _, sqlSch, rowIter, err := sqlNewEngine(fmt.Sprintf("select COUNT(*) as updated from `%v`", tName), root); err == nil {
```
zachmu: This is going to be expensive -- it loads the entire table of data just to count it. You should get the number of rows by calling .size() or similar on the table map

PR #87, opened 2019-09-23
{go, bats}: Replace table works with file with schema in different order

tbantle22:


timsehn:
@bheni can we merge this? It's blocking me testing my import script fully.

go/libraries/doltcore/schema/schema.go
```diff
@@ -95,10 +95,10 @@ func VerifyInSchema(inSch, outSch Schema) (bool, error) {
 	}
 
 	match := true
-	err := outSchCols.Iter(func(tag uint64, outCol Column) (stop bool, err error) {
-		inCol, ok := inSchCols.GetByTag(tag)
+	err := inSchCols.Iter(func(tag uint64, inCol Column) (stop bool, err error) {
```
bheni: Maybe just update the comment to tell how the comparison is being done now, but LGTM.

tbantle22: Updated: [#diff-4275cde1329d049d2dce6d6a85c88acaR88](https://github.com/liquidata-inc/dolt/pull/87/files#diff-4275cde1329d049d2dce6d6a85c88acaR88)

PR #88, opened 2019-09-24
Db/metrics attributes

coffeegoddd:
Similar to previous PR db/event-metrics, but this time, no byte measurements on `clone` as the implementation is different. Some things in the events package have been refactored to prevent circular dependencies. Adding `StandardAttributes` will help me generate the info for my new metrics.


go/libraries/doltcore/env/environment.go
```diff
@@ -579,3 +581,31 @@ func (dEnv *DoltEnv) GetDefaultRemote() (Remote, errhand.VerboseError) {
...
+			if len(remotes) > 0 {
+				urls := make([]string, len(remotes))
+
+				for _, remote := range remotes {
+					urls = append(urls, remote.Url)
+				}
+
+				joined := strings.Join(urls[:], " ")
+
+				evt.SetAttribute(eventsapi.AttributeID_LOCAL_REMOTE_URLS, strings.Trim(joined, " "))
```
bheni: I'm not sure you want to capture all the remote urls for every call.  If a push is called, I want to be able to figure out which repo was pushed to.  if multiple remotes are here then I won't be able to distinguish between them.

go/libraries/doltcore/env/environment.go
```diff
@@ -579,3 +581,28 @@ func (dEnv *DoltEnv) GetDefaultRemote() (Remote, errhand.VerboseError) {
...
+			}
+		}
+
+		if len(evt.GetAttribute(eventsapi.AttributeID_BRANCH_NAME)) < 1 {
+			if dEnv.RepoState != nil {
+				evt.SetAttribute(eventsapi.AttributeID_BRANCH_NAME, dEnv.RepoState.Head.Ref.GetPath())
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
bheni: I'm not sure branch name is useful.  How were you planning on using it?

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -112,8 +116,13 @@ func NewDoltChunkStore(ctx context.Context, nbf *types.NomsBinFormat, org, repoN
 		},
 	})
 	if err != nil {
+		evt.SetAttribute(eventsapi.AttributeID_REMOTEAPI_ERROR, err.Error())
```
bheni: I don't know that you'll want to capture the error in an attribute. Might be better to just have a counter to count errors per rpc.

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -101,6 +103,8 @@ func NewDoltChunkStore(ctx context.Context, nbf *types.NomsBinFormat, org, repoN
 		csClient = RetryingChunkStoreServiceClient{csClient}
 	}
 
+	evt := events.NewEvent(eventsapi.ClientEventType_REMOTEAPI_GET_REPO_METADATA)
```
bheni: Rather than putting:

    events.GlobalCollector.CloseEventAndAdd(evt)

in multiple places, it would be better to:

    defer events.GlobalCollector.CloseEventAndAdd(evt)

bheni: The same thing for each of the calls below.

go/libraries/doltcore/remotestorage/chunk_store.go
bheni: Same thing for each error case below.

go/libraries/events/emitter.go
```diff
@@ -19,20 +19,21 @@ import (
...
-	"github.com/liquidata-inc/dolt/go/libraries/doltcore/env"
 	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
 	"github.com/liquidata-inc/dolt/go/libraries/utils/iohelp"
 )
 
+const (
+	doltDir = ".dolt"
+)
+
 // Emitter is an interface used for processing a batch of events
```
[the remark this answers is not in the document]
bheni: Maybe pass it in to the NewFileEmitter call rather than duplicating the constant.

go/store/chunks/chunk.go
```diff
@@ -49,6 +49,10 @@ func (c Chunk) IsEmpty() bool {
 	return len(c.data) == 0
 }
 
+func (c Chunk) Size() int {
```
bheni: Lets kill this change along with the datas/pull.go changes as they are no longer applicable in the latest and will just cause conflicts.

go/store/datas/pull.go
```diff
@@ -40,21 +40,24 @@ import (
```
bheni: Lets kill all the changes in this file as they are no longer applicable in the latest and will just cause conflicts.

go/cmd/dolt/commands/push.go
```diff
@@ -264,18 +269,40 @@ func pushToRemoteBranch(ctx context.Context, srcRef, destRef, remoteRef ref.Dolt
 	return nil
 }
 
-func progFunc(progChan chan datas.PullProgress, stopChan chan struct{}) {
```
bheni: these changes will no longer be applicable with the v2 push/pull.  Should revert so we don't have to deal with conflicts.

go/libraries/doltcore/env/environment.go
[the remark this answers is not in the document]
coffeegoddd: Tim mentioned he wanted to know what branch people were using most in a given repository if I recall correctly. I plan to track how often a branch of a given name is being used in each active repository over a given time range. Maybe there is a more valuable attribute related to this that is worth adding?

go/libraries/doltcore/env/environment.go
coffeegoddd: So I considered the AttributeID `LOCAL_REMOTE_URLS` for these "inactive" remotes, and then making an `ACTIVE_REMOTE_URL` AttributeID that is specifically used on CLONE, PUSH, PULL, FETCH, (and maybe REMOTE_ADD ?)... what do you think of something like that?

Also do we care very much about these "inactive" `LOCAL_REMOTE_URLS` that aren't actually being used outside of the commands listed above?

go/cmd/dolt/commands/fetch.go
```diff
@@ -158,6 +160,9 @@ func fetchRefSpecs(ctx context.Context, dEnv *env.DoltEnv, rem env.Remote, refSp
 }
 
 func fetchRemoteBranch(ctx context.Context, rem env.Remote, srcDB, destDB *doltdb.DoltDB, srcRef, destRef ref.DoltRef) errhand.VerboseError {
+	evt := events.GetEventFromContext(ctx)
+	evt.SetAttribute(eventsapi.AttributeID_ACTIVE_REMOTE_URL, rem.Url)
+
 	cs, _ := doltdb.NewCommitSpec("HEAD", srcRef.String())
```
coffeegoddd: collects the remote during FETCH

go/cmd/dolt/commands/push.go
```diff
@@ -230,6 +232,9 @@ func deleteRemoteBranch(ctx context.Context, toDelete, remoteRef ref.DoltRef, lo
 }
 
 func pushToRemoteBranch(ctx context.Context, srcRef, destRef, remoteRef ref.DoltRef, localDB, remoteDB *doltdb.DoltDB, remote env.Remote) errhand.VerboseError {
+	evt := events.GetEventFromContext(ctx)
+	evt.SetAttribute(eventsapi.AttributeID_ACTIVE_REMOTE_URL, remote.Url)
+
 	cs, _ := doltdb.NewCommitSpec("HEAD", srcRef.GetPath())
```
coffeegoddd: collects the remote during PUSH

go/cmd/dolt/dolt.go
```diff
@@ -131,13 +133,21 @@ func runMain() int {
...
 	dEnv := env.Load(context.TODO(), env.GetCurrentUserHomeDir, filesys.LocalFS, doltdb.LocalDirDoltDB)
 
-	emitter := events.NewFileEmitter()
+	root, err := env.GetCurrentUserHomeDir()
+
+	if err != nil {
+		return 1
+	}
+
+	emitter := events.NewFileEmitter(root, dbfactory.DoltDir)
```
coffeegoddd: now passing `DoltDir` to file emitter

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -101,6 +103,11 @@ func NewDoltChunkStore(ctx context.Context, nbf *types.NomsBinFormat, org, repoN
...
+	defer events.GlobalCollector.CloseEventAndAdd(evt)
+
+	counter := events.NewCounter(eventsapi.MetricID_REMOTEAPI_ERROR_COUNT)
+
```
coffeegoddd: Using counter metric now

PR #90, opened 2019-09-24
Implemented REPLACE

Hydrocharged:
Mostly tests since this just uses the `Delete` and `Insert` functions that we already have. The previous delete would ignore a delete on a non-existent row, so I just changed it to throw the correct error if the row does not exist so that `REPLACE` works properly now (else it will always say a `REPLACE` did both a delete & insert).

### DIFF
--- a/go/libraries/doltcore/sql/sqltestutil/replacequeries.go
+++ b/go/libraries/doltcore/sql/sqltestutil/replacequeries.go
@@ -53,8 +53,8 @@ var BasicReplaceTests = []ReplaceTest{
 		ExpectedSchema: CompressSchema(PeopleTestSchema),
 	},
 	{
-		Name:           "replace set",
-		ReplaceQuery:   "replace into people set id = 2, first = 'Bart', last = 'Simpson'," +
+		Name: "replace set",
+		ReplaceQuery: "replace into people set id = 2, first = 'Bart', last = 'Simpson'," +
 			"is_married = false, age = 10, rating = 9, uuid = '00000000-0000-0000-0000-000000000002', num_episodes = 222",
 		SelectQuery:    "select * from people where id = 2",
 		ExpectedRows:   CompressRows(PeopleTestSchema, Bart),
@@ -213,7 +213,7 @@ var BasicReplaceTests = []ReplaceTest{
 					(0, "Homer", "Simpson", true, 45, 100),
 					(8, "Milhouse", "Van Houten", false, 8, 3.5),
 					(7, "Maggie", null, false, 1, 5.1)`,
-		ExpectedErr:  "Constraint failed for column 'last': Not null",
+		ExpectedErr: "Constraint failed for column 'last': Not null",
 	},
 	{
 		Name: "type mismatch int -> string",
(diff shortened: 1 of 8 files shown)


go/libraries/doltcore/sql/sqltestutil/replacequeries.go
```diff
@@ -0,0 +1,321 @@
...
+		SkipOnSqlEngine: true,
+	},
+	{
+		Name: "replace partial columns multiple rows null pk",
+		ReplaceQuery: `replace into people (id, first, last, is_married, age, rating) values
+					(0, "Homer", "Simpson", true, 45, 100),
+					(8, "Milhouse", "Van Houten", false, 8, 3.5),
+					(7, "Maggie", null, false, 1, 5.1)`,
+		ExpectedErr:     "Constraint failed for column 'last': Not null",
+		SkipOnSqlEngine: true,
```
zachmu: Should work with your recent change, right?

go/libraries/doltcore/sql/sqltestutil/replacequeries.go
```diff
@@ -0,0 +1,321 @@
...
+var BasicReplaceTests = []ReplaceTest{
+	{
+		Name:           "replace no columns",
+		ReplaceQuery:   "replace into people values (2, 'Bart', 'Simpson', false, 10, 9, '00000000-0000-0000-0000-000000000002', 222)",
+		SelectQuery:    "select * from people where id = 2",
+		ExpectedRows:   CompressRows(PeopleTestSchema, Bart),
+		ExpectedSchema: CompressSchema(PeopleTestSchema),
+	},
+	{
+		Name:           "replace no columns",
```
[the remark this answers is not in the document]
Hydrocharged: lol whoops

go/libraries/doltcore/sql/sqltestutil/replacequeries.go
```diff
@@ -0,0 +1,321 @@
...
+			"is_married", types.BoolKind, "age", types.IntKind, "rating", types.FloatKind),
+	},
+	{
+		Name:         "replace partial columns multiple rows null pk",
+		ReplaceQuery: "replace into people (id, first, last) values (0, 'Bart', 'Simpson'), (1, 'Homer', null)",
+		ExpectedErr:  "column <last> received nil but is non-nullable",
+	},
+	{
+		Name:           "replace partial columns multiple rows duplicate",
+		ReplaceQuery:   "replace into people (id, first, last) values (2, 'Bart', 'Simpson'), (2, 'Bart', 'Simpson')",
```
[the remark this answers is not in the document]
Hydrocharged: It does, just tested it.

go/libraries/doltcore/sql/sqltestutil/replacequeries.go
Hydrocharged: It indeed does, I meant to remove these skips but I guess I forgot to.

Hydrocharged: And I forgot to remove them from the Insert test file too lol

PR #96, opened 2019-09-30
update status messages for push/pull

bheni:


go/cmd/dolt/commands/push.go
```diff
@@ -308,10 +308,10 @@ func pullerProgFunc(pullerEventCh chan datas.PullerEvent) {
 			cli.Println("")
 
 		case datas.StartUploadTableFile:
-			pos = cli.DeleteAndPrint(pos, fmt.Sprintf("Uploading table file %d of %d. %s bytes", evt.TFEventDetails.TableFilesUploaded+1, evt.TFEventDetails.TableFileCount, humanize.Bytes(uint64(evt.TFEventDetails.CurrentFileSize))))
```
bheni: humanize.Bytes will take a number like 10000000 and turn it into "10.0MBs", so putting "bytes" after it is redundant.

coffeegoddd: looks good

PR #98, opened 2019-09-30
Bh/output updates

bheni:


go/libraries/doltcore/mvdata/data_mover.go
```diff
@@ -169,17 +170,19 @@ func NewDataMover(ctx context.Context, root *doltdb.RootValue, fs filesys.Filesy
 	return imp, nil
 }
 
-func (imp *DataMover) Move(ctx context.Context) error {
+func (imp *DataMover) Move(ctx context.Context) (int64, error) {
```
zachmu: You should document the new return param

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -230,8 +230,8 @@ func (dcs *DoltChunkStore) GetManyCompressed(ctx context.Context, hashes hash.Ha
 }
 
 const (
-	getLocsBatchSize      = 4096
-	getLocsMaxConcurrency = 32
+	getLocsBatchSize      = 128 * 1024
```
zachmu: This file unrelated?

bheni: yeah.  Pullin a sneaky.

go/libraries/doltcore/mvdata/data_mover.go
bheni: done

PR #99, opened 2019-10-01
Implemented UPDATE

Hydrocharged:
I think we should delete the old SQL methods that are in the `sql.go` file. I know at first you mentioned keeping them there for reference, but they're not being used at all at this point, and they're still in git history if we want to look at them again in the future for some reason. It's clutter at this point.

I'm skipping that one test at the end because of a WHERE decision in `go-mysql-server`. The code looks intentional, in that converting strings to ints will return 0 if the string is not parsable. I'll file it as a non-conforming bug on their end, but for now I'm skipping the test.

### DIFF
--- a/go/libraries/doltcore/sqle/tables.go
+++ b/go/libraries/doltcore/sqle/tables.go
@@ -182,6 +182,65 @@ func (t *DoltTable) Delete(ctx *sql.Context, sqlRow sql.Row) error {
 	return nil
 }
 
+func (t *DoltTable) Update(ctx *sql.Context, oldRow sql.Row, newRow sql.Row) error {
+	dOldRow, err := SqlRowToDoltRow(t.table.Format(), oldRow, t.sch)
+	if err != nil {
+		return err
+	}
+	dNewRow, err := SqlRowToDoltRow(t.table.Format(), newRow, t.sch)
+	if err != nil {
+		return err
+	}
+	deleteFirst := false
+
+	// If the PK is changed then we have to delete the old row first
+	// This is assuming that the new PK is not taken
+	dOldKey := dOldRow.NomsMapKey(t.sch)
+	dOldKeyVal, err := dOldKey.Value(ctx)
+	if err != nil {
+		return err
+	}
+	dNewKey := dNewRow.NomsMapKey(t.sch)
+	dNewKeyVal, err := dNewKey.Value(ctx)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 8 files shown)


go/libraries/doltcore/sqle/tables.go
```diff
@@ -182,6 +182,65 @@ func (t *DoltTable) Delete(ctx *sql.Context, sqlRow sql.Row) error {
...
+	if err != nil {
+		return err
+	}
+	if !dOldKeyVal.Equals(dNewKeyVal) {
+		_, rowExists, err := t.table.GetRow(ctx, dNewKeyVal.(types.Tuple), t.sch)
+		if err != nil {
+			return errhand.BuildDError("failed to read table").AddCause(err).Build()
+		}
+		if rowExists {
+			return errors.New("cannot update primary key column")
```
zachmu: I would make this more specific, something like "Primary key collision" and give detail about which key is causing the issue.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -182,6 +182,65 @@ func (t *DoltTable) Delete(ctx *sql.Context, sqlRow sql.Row) error {
...
+func (t *DoltTable) Update(ctx *sql.Context, oldRow sql.Row, newRow sql.Row) error {
+	dOldRow, err := SqlRowToDoltRow(t.table.Format(), oldRow, t.sch)
+	if err != nil {
+		return err
+	}
+	dNewRow, err := SqlRowToDoltRow(t.table.Format(), newRow, t.sch)
+	if err != nil {
+		return err
+	}
+	deleteFirst := false
```
zachmu: Declare this right above where it's first used and call it deleteOldRow

go/libraries/doltcore/sqle/tables.go
```diff
@@ -182,6 +182,65 @@ func (t *DoltTable) Delete(ctx *sql.Context, sqlRow sql.Row) error {
...
+		if err != nil {
+			return errhand.BuildDError("failed to read table").AddCause(err).Build()
+		}
+		if rowExists {
+			return errors.New("cannot update primary key column")
+		}
+		deleteFirst = true
+	}
+
+	typesMap, err := t.table.GetRowData(ctx)
```
zachmu: This section could use some newlines to break it up a bit

Hydrocharged: Does it really though? It's in line with the rest of the functions 😤

PR #101, opened 2019-10-01
go/{store,libraries/doltcore/remotestorage}: Make the code peddling in nbs table file formats a little more explicit about it.

reltuk:


go/store/nbs/mem_table.go
```diff
@@ -153,8 +153,19 @@ func (mt *memTable) getMany(ctx context.Context, reqs []getRecord, foundChunks c
 	return remaining
 }
 
-func (mt *memTable) getManyCompressed(ctx context.Context, reqs []getRecord, foundCmpChunks chan<- chunks.Chunkable, wg *sync.WaitGroup, ae *atomicerr.AtomicError, stats *Stats) bool {
-	panic("not implemented")
+func (mt *memTable) getManyCompressed(ctx context.Context, reqs []getRecord, foundCmpChunks chan<- CompressedChunk, wg *sync.WaitGroup, ae *atomicerr.AtomicError, stats *Stats) bool {
```
reltuk: @bheni We implemented this with `ChunkToCompressedChunk`.

go/store/nbs/table_reader.go
```diff
@@ -75,14 +74,31 @@ func (cmp CompressedChunk) ToChunk() (chunks.Chunk, error) {
 	return chunks.NewChunk(data), nil
 }
 
+func ChunkToCompressedChunk(chunk chunks.Chunk) CompressedChunk {
+	compressed := snappy.Encode(nil, chunk.Data())
+	length := len(compressed)
+	compressed = append(compressed, []byte{0, 0, 0, 0}...)
+	binary.BigEndian.PutUint32(compressed[length:], crc(compressed[:length]))
+	return CompressedChunk{H: chunk.Hash(), FullCompressedChunk: compressed, CompressedData: compressed[:length]}
```
reltuk: @bheni Much open to suggestions on improving this and/or reducing any foreseeable problems in duplicating this from `table_writer`.

go/store/nbs/table_reader.go
```diff
@@ -75,14 +74,31 @@ func (cmp CompressedChunk) ToChunk() (chunks.Chunk, error) {
...
+
 // Hash returns the hash of the data
 func (cmp CompressedChunk) Hash() hash.Hash {
 	return cmp.H
 }
 
 // IsEmpty returns true if the chunk contains no data.
 func (cmp CompressedChunk) IsEmpty() bool {
-	return len(cmp.CompressedData) == 0
+	return len(cmp.CompressedData) == 1 && cmp.CompressedData[0] == 0
```
reltuk: @bheni Fixed this logic here...

go/store/nbs/table_reader.go
```diff
@@ -75,14 +74,31 @@ func (cmp CompressedChunk) ToChunk() (chunks.Chunk, error) {
...
+}
+
+var EmptyCompressedChunk CompressedChunk
+
+func init() {
+	EmptyCompressedChunk = ChunkToCompressedChunk(chunks.EmptyChunk)
+	if !EmptyCompressedChunk.IsEmpty() {
+		panic("EmptyCompressedChunk should be empty...")
+	}
```
reltuk: @bheni Should this be here? Or in a unit test?

go/libraries/doltcore/doltdb/table.go
```diff
@@ -261,9 +261,7 @@ func (t *Table) GetSchema(ctx context.Context) (schema.Schema, error) {
 	}
 
 	schemaRef := schemaRefVal.(types.Ref)
-	sch, _ := refToSchema(ctx, t.vrw, schemaRef)
-
-	return sch, nil
+	return refToSchema(ctx, t.vrw, schemaRef)
```
reltuk: Fixes a bug where `GetSchema` returned `err == nil` even when there was an error.

go/libraries/doltcore/remotestorage/noop_chunk_cache.go
```diff
@@ -29,21 +29,21 @@ var noopChunkCache = &noopChunkCacheImpl{}
...
 }
 
-func (*noopChunkCacheImpl) Put(chnks []chunks.Chunkable) {
+func (*noopChunkCacheImpl) Put(chnks []nbs.CompressedChunk) {
 }
 
-func (*noopChunkCacheImpl) Get(hashes hash.HashSet) map[hash.Hash]chunks.Chunkable {
-	return make(map[hash.Hash]chunks.Chunkable)
+func (*noopChunkCacheImpl) Get(hashes hash.HashSet) map[hash.Hash]nbs.CompressedChunk {
+	return make(map[hash.Hash]nbs.CompressedChunk)
```
reltuk: This was broken when it returned `map[hash.Hash]chunks.Chunkable` because `chunks.Chunkable` zero value is `nil`, and caller is expecting non-`nil` zero value where `v.IsEmpty() == true`.

go/libraries/doltcore/doltdb/table.go
bheni: yay

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -440,7 +440,7 @@ func (dcs *DoltChunkStore) HasMany(ctx context.Context, hashes hash.HashSet) (ha
 				absent[currHash] = struct{}{}
 				j++
 			} else {
-				c := chunks.NewChunkWithHash(currHash, []byte{})
+				c := nbs.ChunkToCompressedChunk(chunks.NewChunkWithHash(currHash, []byte{}))
```
bheni: should use chunks.EmptyCompressedChunk

go/libraries/doltcore/remotestorage/noop_chunk_cache.go
bheni: Maybe make this a member so you don't have to make a new one for every get call?

go/store/datas/puller.go
```diff
@@ -303,14 +302,23 @@ func limitToNewChunks(absent hash.HashSet, downloaded hash.HashSet) {
...
 func (p *Puller) getCmp(ctx context.Context, twDetails *TreeWalkEventDetails, leaves, batch hash.HashSet, completedTables chan FilledWriters) (hash.HashSet, hash.HashSet, error) {
-	found := make(chan chunks.Chunkable, 4096)
+	found := make(chan nbs.CompressedChunk, 4096)
 	processed := make(chan CmpChnkAndRefs, 4096)
 
 	ae := atomicerr.New()
 	go func() {
 		defer close(found)
-		err := p.srcDB.chunkStore().GetManyCompressed(ctx, batch, found)
+		st, ok := p.srcDB.chunkStore().(nbsGetManyCompressedable)
```
bheni: maybe create a new interface that is both a ChunkStore and an nbsGetManyCompressedable:

    type CompressedableChunkStore interface {
        chunks.ChunkStore
        nbsGetManyCompressedable
    }

and then cache the chunkstores as members in the NewPuller call.  You can error there if the interface isn't implemented and then you don't have to wait til the code gets here to fail in this go routine.

go/store/nbs/mem_table.go
bheni: should be fine.  Not optimal, but we aren't using memtable for push and pull anymore, so memTable.getManyCompressed may not even be used in the current codebase.

go/store/nbs/table_reader.go
bheni: I guess a unit test.  Better for Jenkins to fail it rather than for it to panic at startup.

go/store/types/indexed_sequences.go
```diff
@@ -200,7 +200,7 @@ func LoadLeafNodes(ctx context.Context, cols []Collection, startIdx, endIdx uint
 		var ok bool
 		childCols[i], ok = v.(Collection)
 		if !ok {
-			return nil, 0, errors.New("corrupted database; nil where child collection .(Collection) should be")
+			return nil, 0, fmt.Errorf("corrupted database; nil where child collection .(Collection) should be; i: %v, h: %v", i, hs[i])
```
bheni: maybe add 

  ; indexed_sequence;

to here to be consistent from the meta_sequence error.

go/store/nbs/table_reader.go
bheni: So the TableWriter uses a snappyEncoder interface.  Sometimes this is a "realSnappyEncoder", but not necessarily.  Not sure how these play together.  I guess I'd leave the code like this, and just have CompressedChunks always use real compression.

go/store/nbs/table_reader.go
reltuk: Move it!

go/libraries/doltcore/remotestorage/noop_chunk_cache.go
reltuk: Decided to leave it.

go/libraries/doltcore/remotestorage/chunk_store.go
reltuk: But I need the hash, right? Left it like this...

go/store/datas/puller.go
reltuk: This was better. Thanks for the suggestion!

PR #121, opened 2019-10-07
Added bats test for semicolon bug in piped input

timsehn:


Hydrocharged:
https://github.com/liquidata-inc/dolt/pull/122

PR #122, opened 2019-10-07
Fixed SQL batch insert breaking on semicolons in strings

Hydrocharged:
Fixes SQL bug as found in new Bats test introduced here: https://github.com/liquidata-inc/dolt/pull/124


timsehn:
LGTM

PR #123, opened 2019-10-07
Fix cell count function for diff summary cells modified

tbantle22:


go/store/types/tuple.go
```diff
@@ -448,37 +448,48 @@ func (t Tuple) Less(nbf *NomsBinFormat, other LesserValuable) (bool, error) {
...
+	otherMap, err := other.fieldsToMap()
+
+	if err != nil {
+		return 0, err
+	}
+
+	for i, v := range tMap {
+		if otherMap[i] == nil {
+			changed++
+		} else if otherMap[i] != v {
```
bheni: ==, != may work for some types, but will not work for all noms types.  I would use 

    v.Equals(otherMap[i]) 

which will also work if otherMap[i] returns nil, so should be able to handle both the if and else if case here.

tbantle22: Fixed [91f0fe2](https://github.com/liquidata-inc/dolt/pull/123/files#diff-34bc7aa1bf7e0235e247e99bc62049caR459-R461)

PR #130, opened 2019-10-08
fix active remote url to only save the url scheme

coffeegoddd:


go/cmd/dolt/commands/clone.go
```diff
@@ -97,7 +97,13 @@ func Clone(ctx context.Context, commandStr string, args []string, dEnv *env.Dolt
...
 						evt := events.GetEventFromContext(ctx)
-						evt.SetAttribute(eventsapi.AttributeID_ACTIVE_REMOTE_URL, remoteUrl)
+						u, err := earl.Parse(remoteUrl)
+
+						if err == nil {
+							if u.Scheme != "" {
+								evt.SetAttribute(eventsapi.AttributeID_REMOTE_URL_SCHEME, u.Scheme)
+							}
+						}
 					}
```
coffeegoddd: now using url scheme

PR #131, opened 2019-10-08
bats: Factor setup and teardown logic out of tests, greatly reducing duplication

mjesuele:
This has been bothering me for a while.

### DIFF
--- a/bats/helper/common.bash
+++ b/bats/helper/common.bash
@@ -13,19 +13,18 @@ set_dolt_user() {
 	dolt config --global --add user.email "$2" > /dev/null 2>&1
 }
 
-setup_common() {
+setup_no_dolt_init() {
 	export PATH=$PATH:~/go/bin
-	export NOMS_VERSION_NEXT=1
 	cd $BATS_TMPDIR
 	# Append the directory name with the pid of the calling process so
 	# multiple tests can be run in parallel on the same machine
 	mkdir "dolt-repo-$$"
 	cd "dolt-repo-$$"
+}
 
-	SKIP_INIT=$1
-	if [ "$SKIP_INIT" != true ] ; then
-		dolt init
-	fi
+setup_common() {
+	setup_no_dolt_init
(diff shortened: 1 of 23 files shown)


bats/helper/common.bash
```diff
@@ -5,10 +5,33 @@ if [ -z "$BATS_TMPDIR" ]; then
...
+	export PATH=$PATH:~/go/bin
+	export NOMS_VERSION_NEXT=1
+	cd $BATS_TMPDIR
+	# Append the directory name with the pid of the calling process so
+	# multiple tests can be run in parallel on the same machine
+	mkdir "dolt-repo-$$"
+	cd "dolt-repo-$$"
+
+	SKIP_INIT=$1
+	if [ "$SKIP_INIT" != true ] ; then
```
mjesuele: Please do let me know if you know a better way to write this. I tried a few, actually...

timsehn: What if you made two functions? One called setup_common() and one called setup_no_dolt() and make the setup_no_dolt() call setup_common(). Passing the variable is a little janky. The above seems to be more readable.

bats/helper/common.bash
```diff
@@ -5,10 +5,33 @@ if [ -z "$BATS_TMPDIR" ]; then
...
 batshelper() { echo `nativebatsdir helper/$1`; }
+
+set_dolt_user() {
+	dolt config --global --add user.name "$1" > /dev/null 2>&1
+	dolt config --global --add user.email "$2" > /dev/null 2>&1
+}
+
+setup_common() {
+	export PATH=$PATH:~/go/bin
+	export NOMS_VERSION_NEXT=1
```
timsehn: I think this NOMS_VERSION_NEXT is not be needed anymore.

bats/helper/common.bash
mjesuele: I like that, thanks.

PR #133, opened 2019-10-09
[RFC] Basic design doc for dolt blame

mjesuele:
Definitely interested in people's thoughts. I'll be plugging away with the approach I describe herein in the meantime.

### DIFF
--- a/go/cmd/dolt/commands/blame_design_doc.md
+++ b/go/cmd/dolt/commands/blame_design_doc.md
@@ -1,0 +1,63 @@
+# Dolt Blame
+
+## Goal
+
+Row-based blame annotation for Dolt tables (column-based coming soon)
+
+Assume we have a `startup_employees` table with some data with schema `pk:int`,
+`name:string`, and `salary:int`.
+
+In the example below, the `commit`, `author`, and `time` columns are
+added by `blame`.
+
+```
+# find out who last modified the software engineers
+dolt blame HEAD startup_employees --where "salary >= 100000"
+
++----+---------+--------+----------------------------------+----------------+---------------------------+
+| pk | name    | salary | commit                           | author         | time                      |
++----+---------+--------+----------------------------------+----------------+---------------------------+
+| 1  | Alex    | 100000 | 91osuaj28kpbl8fc2jahq90nv0chpps5 | Barbara Dwyer  | 2019-04-07 14:37:48 -0700 |
+| 2  | Barbara | 125000 | 91osuaj28kpbl8fc2jahq90nv0chpps5 | Barbara Dwyer  | 2019-04-07 14:37:48 -0700 |
+| 5  | Robert  | 125000 | bluojp2qu05o7h2j055kmfskkqm0a28i | Robert Loblaw  | 2019-04-08 11:27:15 -0700 |
+| 8  | Diana   | 120000 | vnolpfad1h277en46qq5084a0dfpkgju | Thomas Foolery | 2019-04-08 12:10:22 -0700 |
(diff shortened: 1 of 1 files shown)


timsehn:
I’m generally good with the interface. Implantation I have no thoughts.
precomputing this data seems more complicated so I would only do it if this
walk was prohibitively slow or worse, memory intensive.

mjesuele:
That's my thinking as well. Definitely feeling inclined to implement and profile the on-the-fly option before complicating things further.

oscarbatori:
How do we feel about actually saving this information in a table? Then the command line interface just queries the SQL table, and we can surface programmatic access.

I started out less dogmatic on this, but @reltuk has convinced me to adopt a maximalist position on surfacing everything in SQL.

mjesuele:
Sounds like alternative 2. I definitely see the benefit. The major downside is that it's no longer an orthogonal process and a lot more care is required. Which can be fine.

I think it still makes sense to implement it the other way to start; if we decide to make blame information a table or tables, the logic will be reusable at a minimum for adding blame to repos that don't have it yet, so no effort is wasted and the path to a basic feature release is shorter.

Does that make sense?

oscarbatori:
> I think it still makes sense to implement it the other way to start; if we decide to make blame information a table or tables, the logic will be reusable at a minimum for adding blame to repos that don't have it yet, so no effort is wasted and the path to a basic feature release is shorter.

This sold me, we can revisit but given that we think this will be a nice "candy" feature let's proceed with the low tech solution you described.

reltuk:
I think the implementation walking the commit graph to calculate the values at query time is the right way to do things.

I do like Oscar's lean into SQL though. I think a basic `dolt blame` interface is probably fine if it looks a lot like a dolt table list, but implementing this as a system table that a query can join against or as system columns on every table that can be addressed with standard syntax (`my_table.__dolt_blame_commit_id__`?) would be really cool and would avoid further special casing in CLI commands, etc.

oscarbatori:
If we want to surface this in the UI, i.e. click on a cell and see lineage, then would it be feasible to compute it in real time? I would imagine that storing it as a system table might be our only choice if we want that feature to be snappy.

reltuk:
I think the question of whether you materialize a derivative dataset to make common queries faster is a somewhat orthogonal concern --- the fundamental design should be "blame information is inherently derivable from the commit graph / merkle tree" and a totally credible pass at an implementation is "go ahead and derive it when we need it".

Storage, materialization and distribution policy for "materialized views" or derivative datasets in general is something we definitely need to solve eventually. But I don't think we need to start from a place where we special case this use case; we a have a credible analogous existence proof in an implementation approach in things like git and mercurial, and we can always iterate on the performance of that solution when it turns out to be unusably slow.

mjesuele:
Thanks for the very helpful comments @reltuk; I'm feeling good about proceeding. Closing this PR, but not opposed to additional discussion. (Also -- is there somewhere we'd like this design doc to persist?)

mjesuele:
Actually, where exactly do you guys think we land on the issue of interface? I think it's probably good to try to retain the git semantics even if they're just sugar over a lower-level SQL query, but could be convinced otherwise. Either way am definitely down to expose them as tables usable for joins as proposed.

oscarbatori:
How about design docs in Notion?

I agree on the interface, I like your presentation of it, and agree we
should keep the git semantics even if it’s on top of SQL.

mjesuele:
Added to Notion: https://www.notion.so/liquidata/Dolt-blame-ec7dd64997144b61847e326017b8f7fa

PR #138, opened 2019-10-14
[Temporary] Noms Byte Array and Dolt Types

Hydrocharged:
I apologize for the size of this PR for a _first look_. I made a new branch and removed a lot of my changes to get the size down (was originally at 39 files changed, and I'm not nearly done yet). The main purpose of this PR is so you guys can see how I'm approaching this.

The idea will be that adding new types will just be making a new file in `go/libraries/doltcore/dolttypes` and making the appropriate changes in `dolttype.go`, and it should all work. At least that's the plan. I haven't added anything for SQL stuff to work, so the interface will most definitely change.

Of note, I am working with a new "kind", which I'm calling `DoltKind`. I've got a bit of special-case logic to handle `types.UnderlyingArray` (and associates) differently, so that `NomsKind` is used for most things, with `DoltKind` being used for others. Also there's the `Unknown` type, so that most calls don't outright fail for certain operations (such as `String()`).

### DIFF
--- a/go/libraries/doltcore/dolttypes/dolttype.go
+++ b/go/libraries/doltcore/dolttypes/dolttype.go
@@ -1,0 +1,124 @@
+// Copyright 2019 Liquidata, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dolttypes
+
+import (
+	"fmt"
+)
+
+// DoltKind is Dolt's version of NomsKind for Dolt-defined types built on UnderlyingArrayKind.
+type DoltKind byte
+const DoltKindLength = 1
+
+const (
+	UnknownKind DoltKind = iota
+	TimestampKind
+)
+
+type DoltType interface {
+	// Compare returns whether the DoltType is equal to another DoltType. If different types, compares their kinds.
+	Compare(DoltType) int
+	// Decode takes in a byte array (prepended with the kind) and returns a new DoltType representing the given data.
+	Decode([]byte) (DoltType, error)
+	// Encode returns a byte array fully representing the type (prepended with the kind) suitable for decoding or storage.
+	Encode() ([]byte, error)
+	// Equals returns whether this DoltType represents the exact same data as another DoltType.
+	Equals(DoltType) bool
+	// Kind returns the DoltKind of this DoltType.
+	Kind() DoltKind
+	// MarshalBool returns a bool representing the DoltType.
+	MarshalBool() (bool, error)
+	// MarshalDoltType returns a DoltType (defined by the DoltKind) representing the parent DoltType.
+	MarshalDoltType(DoltKind) (DoltType, error)
+	// MarshalFloat returns a float64 representing the DoltType.
+	MarshalFloat() (float64, error)
+	// MarshalInt returns a int64 representing the DoltType.
+	MarshalInt() (int64, error)
+	// MarshalString returns a string representing the DoltType.
+	MarshalString() (string, error)
+	// MarshalUint returns a uint64 representing the DoltType.
+	MarshalUint() (uint64, error)
+	// UnmarshalBool returns this DoltType representing the given bool.
+	UnmarshalBool(bool) (DoltType, error)
+	// UnmarshalDoltType returns this DoltType representing the given DoltType, which may be of a different DoltKind.
+	UnmarshalDoltType(DoltType) (DoltType, error)
+	// UnmarshalFloat returns this DoltType representing the given float64.
+	UnmarshalFloat(float64) (DoltType, error)
+	// UnmarshalInt returns this DoltType representing the given int64.
+	UnmarshalInt(int64) (DoltType, error)
+	// UnmarshalString returns this DoltType representing the given string.
+	UnmarshalString(string) (DoltType, error)
+	// UnmarshalUint returns this DoltType representing the given uint64.
+	UnmarshalUint(uint64) (DoltType, error)
+	fmt.Stringer
+}
+
+var kindMap = map[DoltKind]DoltType {
+	UnknownKind:   Unknown{},
+	TimestampKind: Timestamp{},
+}
+var KindName = map[DoltKind]string {
+	UnknownKind:   "unknown",
+	TimestampKind: "Timestamp",
+}
+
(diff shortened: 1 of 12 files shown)


go/store/types/noms_kind.go
```diff
@@ -52,52 +52,55 @@ const (
 	UintKind
 	NullKind
 	TupleKind
+	UnderlyingArrayKind
```
zachmu: Probably want to call this a ByteArray, but let's see what Aaron thinks.

go/libraries/doltcore/type_conversion.go
```diff
@@ -59,66 +60,82 @@ func GetUnderlyingError(err error) error {
...
-		types.NullKind:   convToNullFunc},
+		types.StringKind:          convBoolToString,
+		types.UUIDKind:            nil,
+		types.UintKind:            convBoolToUint,
+		types.IntKind:             convBoolToInt,
+		types.FloatKind:           convBoolToFloat,
+		types.BoolKind:            identityConvFunc,
+		types.UnderlyingArrayKind: convBoolToUnderlyingArray,
+		types.NullKind:            convToNullFunc},
+	types.UnderlyingArrayKind: {
```
zachmu: I don't think these conversions make sense. The array type is only the storage details for some concrete type. You don't "convert" between a binary format and a high-level type, you "marshall" and "unmarshall". To put it another way, it makes sense to convert between a string and a timestamp (which is built on top of a byte array), but not between a string the underlying byte array.

go/libraries/doltcore/str_to_noms.go
```diff
@@ -16,17 +16,17 @@ package doltcore
...
 
-	"github.com/google/uuid"
-
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
-// StringToValue takes a string and a NomsKind and tries to convert the string to a noms Value.
-func StringToValue(s string, kind types.NomsKind) (types.Value, error) {
+// StringToValue takes a string, NomsKind, and DoltKind and tries to convert the string to a noms Value.
+func StringToValue(s string, kind types.NomsKind, doltKind dolttypes.DoltKind) (types.Value, error) {
```
zachmu: To keep these type signatures sane, you might want to introduce every type built on top of ByteArray as a new types.NomsKind constant.

go/libraries/doltcore/schema/alterschema/addcolumn.go
```diff
@@ -37,18 +38,18 @@ const (
 // table, since we must write a value for each row. If the column is not nullable, a default value must be provided.
 //
 // Returns an error if the column added conflicts with the existing schema in tag or name.
-func AddColumnToTable(ctx context.Context, db *doltdb.DoltDB, tbl *doltdb.Table, tag uint64, newColName string, colKind types.NomsKind, nullable Nullable, defaultVal types.Value) (*doltdb.Table, error) {
+func AddColumnToTable(ctx context.Context, db *doltdb.DoltDB, tbl *doltdb.Table, tag uint64, newColName string, colNomsKind types.NomsKind, colDoltKind dolttypes.DoltKind, nullable Nullable, defaultVal types.Value) (*doltdb.Table, error) {
```
zachmu: All these double column kind params are an issue. There should be exactly one kind parameter for these methods, not two.

go/libraries/doltcore/dolttypes/dolttype.go
```diff
@@ -0,0 +1,124 @@
...
+// DoltKind is Dolt's version of NomsKind for Dolt-defined types built on UnderlyingArrayKind.
+type DoltKind byte
+const DoltKindLength = 1
+
+const (
+	UnknownKind DoltKind = iota
+	TimestampKind
+)
+
+type DoltType interface {
```
zachmu: I think this is probably the wrong approach. We don't want to mimic the Noms storage layer in the dolt layer, just build on top of it.

Rather than have a separate NomsKind and DoltKind everywhere, you probably want to just stick with the NomsKind constants, and mark some of them only as used by Dolt for record keeping, not actually reified as those types. Example:

        BoolKind NomsKind = iota
	FloatKind
	StringKind
        ...
        ByteArrayKind // New noms type you're implementing, fully reified in noms layer
        DoltTimestampKind // Not ever directly stored in noms, just used to identify the type stored in the ByteArrayKind in a dolt Schema
        DoltPointKind // Not ever directly stored in noms, just used to identify the type stored in the ByteArrayKind in a dolt Schema

Aaron should take a look as well.

go/store/types/noms_kind.go
Hydrocharged: That's totally fine, I wasn't too sure about the name anyway lol

go/libraries/doltcore/type_conversion.go
Hydrocharged: It is marshalling between types. Further down in the same file you'll see the implementation of each of these.

go/libraries/doltcore/str_to_noms.go
Hydrocharged: From what I can see, a `NomsKind` more or less maps with a `types.Value`. In a few places in the codebase, I'm seeing things like `uuid, ok := someParam.(types.UUID)`, which then works off of the assumption that you're dealing with a `UUID` without ever calling `Kind()`.

go/libraries/doltcore/dolttypes/dolttype.go
Hydrocharged: This would require one to add `ByteArrayKind`, `DoltTimestampKind`, and `DoltPointKind` in every place that's doing a check for the type, which isn't clear. Even for my (name changing soon) `UnderlyingArrayKind`, I'm not 100% sure I've added the type to all of the places that handles the kind at this moment, so to force that workflow seems prone to introducing bugs.

Hydrocharged:
The primary feedback point seems to be with the introduction of the `DoltKind` type. While I'm not the biggest fan of it as well, adding more `NomsKind`s seems like a far higher implementation cost once more than one kind is implemented.

go/libraries/doltcore/schema/column.go
```diff
@@ -15,22 +15,39 @@
...
 	"math"
 	"strings"
 
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
-// KindToLwrStr maps a noms kind to the kinds lowercased name
-var KindToLwrStr = make(map[types.NomsKind]string)
+// FullKind is the combination of both a DoltKind and NomsKind, represents the absolute kind
+type FullKind struct {
```
bheni: I don't really like this approach. A DoltKind should only ever be backed by one NomsKind.  Why would you need to know anything other than the DoltKind if it can be mapped?

go/libraries/doltcore/str_to_noms.go
bheni: you don't want to keep adding noms kinds for each.  Will make adding new kinds suck as much as before.  You'll have to go and fix a bunch of switch statements each time, etc.

Hydrocharged:
The PR has served its purpose and a decision has been offline. Closing this PR since I don't think we need it anymore.

PR #143, opened 2019-10-16
go/libraries/utils/argparser/parser.go: Clean up Parse and fix handling of -- passed as commit spec

mjesuele:
I noticed this while trying to get `dolt blame` to support `--` as a valid value for commit spec. Previously, the parser would stop parsing when it encountered a `--`. This led to problems when that argument was not the last argument in the list, as in

```
dolt blame -- some_table
```

This fixes that. In the process I think I simplified the logic by taking advantage of early returns and continues to avoid else blocks and the indentation that comes with them.

### DIFF
--- a/go/libraries/utils/argparser/parser.go
+++ b/go/libraries/utils/argparser/parser.go
@@ -144,58 +144,56 @@ func (ap *ArgParser) Parse(args []string) (*ArgParseResults, error) {
 	for ; i < len(args); i++ {
 		arg := args[i]
 
-		if len(arg) == 0 || arg[0] != '-' { // empty strings should get passed through like other naked words
+		if len(arg) == 0 || arg[0] != '-' || arg == "--" { // empty strings should get passed through like other naked words
 			list = append(list, arg)
-		} else {
-			if arg == "--" {
-				break
+			continue
+		}
+
+		optName, value := splitOption(arg)
+
+		if optName == "help" || optName == "h" {
+			return nil, ErrHelp
+		}
+
+		supOpt, ok := ap.NameOrAbbrevToOpt[optName]
+
+		if !ok {
+			return nil, UnknownArgumentParam{optName}
+		}
+
+		if _, exists := results[optName]; exists {
+			//already provided
+			return nil, errors.New("error: flag `" + supOpt.Name + "' should not have a value")
+		}
+
+		if supOpt.OptType == OptionalFlag {
+			if value != nil {
+				return nil, errors.New("error: multiple values provided for `" + supOpt.Name + "'")
 			}
 
-			optName, value := splitOption(arg)
+			results[supOpt.Name] = ""
+			continue
+		}
 
-			if optName == "help" || optName == "h" {
-				return nil, ErrHelp
+		if value == nil {
+			i++
+			if i >= len(args) {
+				return nil, errors.New("error: no value for option `" + arg + "'")
 			}
 
-			supOpt, ok := ap.NameOrAbbrevToOpt[optName]
+			valueStr := args[i]
+			value = &valueStr
+		}
 
-			if !ok {
-				return nil, UnknownArgumentParam{optName}
-			} else {
-				if _, exists := results[optName]; exists {
-					//already provided
-					return nil, errors.New("error: flag `" + supOpt.Name + "' should not have a value")
-				}
+		if supOpt.Validator != nil {
+			err := supOpt.Validator(*value)
 
-				if supOpt.OptType == OptionalFlag {
-					if value != nil {
-						return nil, errors.New("error: multiple values provided for `" + supOpt.Name + "'")
-					}
-
-					results[supOpt.Name] = ""
-				} else {
-					if value == nil {
-						i++
-						if i >= len(args) {
-							return nil, errors.New("error: no value for option `" + arg + "'")
-						}
-
-						valueStr := args[i]
-						value = &valueStr
-					}
-
-					if supOpt.Validator != nil {
-						err := supOpt.Validator(*value)
-
-						if err != nil {
-							return nil, err
-						}
-					}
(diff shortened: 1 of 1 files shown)


mjesuele:
Okay, that spec makes sense. Yeah, if that's the intended behavior, I was seeing a bug -- the double dash and any subsequent parameters weren't being passed to the commands at all (I verified this with logging and in the debugger).

With my change, those subsequent args still get run through the normal arg parser, so things that look like flags will be treated as such, and that's not what we want.

Instead, they should just be passed verbatim from the `--` on, am I understanding correctly? Can have that change quickly if so.

reltuk:
Yeah, my understanding is that every argument appearing after `--` should be passed verbatim on to the command. The `--` itself should not get passed on. Right?

mjesuele:
I think it should be, or perhaps be replaced with an empty string. Otherwise, commands which depend on the later positional arguments being at a particular index will fail.

bheni:
I'm not sure I understand the issue.  Can you give an example.  I would assume that:

    command -option -option2 table file

would be the same as 

    command -option -option2 table -- file

and the same as

    command -option -option2 -- table file

in all cases table is positional arg 0, and file is positional arg 1.

bheni:
In the above example if -- gets passed as one of the positional arguments, then you can no longer guarantee that table is positional arg 0 and file is positional arg 1.  In each of the 3 cases above they are at different positional locations:

    ["table", "file"]
    ["table", "--", "file"]
    ["--","table","file"]

mjesuele:
The `--` _replaces_ a positional arg. In the case of, say, `git checkout`:

```
git checkout -- some/file
```
is equivalent (at least roughly and maybe precisely) to 
```
git checkout HEAD some/file
```

bheni:
Isn't that the same as

    git checkout some/file

reltuk:
The `--` stops argument parsing and gets removed from the argument list, or equivalently, gets replaced by the remaining arguments.

mjesuele:
Okay, that's fine. The parser still needs to be fixed, because it doesn't do that.

mjesuele:
I can make the change in a few minutes.

mjesuele:
Actually I think there's still an issue, though it's arguably because Git is abusing `--`:

```
git checkout file1 file2 file3
```
will check out file2 and file3 from a branch named `file1` if one exists, but you can prevent this by passing `--`:
```
git checkout -- file1 file2 file3
```

You can see they are not equivalent:

```
$ git checkout master foo.go bar.go
error: pathspec 'foo.go' did not match any file(s) known to git
error: pathspec 'bar.go' did not match any file(s) known to git
```

```
$ git checkout -- master foo.go bar.go
error: pathspec 'master' did not match any file(s) known to git
error: pathspec 'foo.go' did not match any file(s) known to git
error: pathspec 'bar.go' did not match any file(s) known to git
```

reltuk:
```
        /*
         * case 1: git checkout <ref> -- [<paths>]
         *
         *   <ref> must be a valid tree, everything after the '--' must be
         *   a path.
         *
         * case 2: git checkout -- [<paths>]
         *
         *   everything after the '--' must be paths.
         *
         * case 3: git checkout <something> [--]
         *
         *   (a) If <something> is a commit, that is to
         *       switch to the branch or detach HEAD at it.  As a special case,
         *       if <something> is A...B (missing A or B means HEAD but you can
         *       omit at most one side), and if there is a unique merge base
         *       between A and B, A...B names that merge base.
         *
         *   (b) If <something> is _not_ a commit, either "--" is present
         *       or <something> is not a path, no -t or -b was given, and
         *       and there is a tracking branch whose name is <something>
         *       in one and only one remote (or if the branch exists on the
         *       remote named in checkout.defaultRemote), then this is a
         *       short-hand to fork local <something> from that
         *       remote-tracking branch.
         *
         *   (c) Otherwise, if "--" is present, treat it like case (1).
         *
         *   (d) Otherwise :
         *       - if it's a reference, treat it like case (1)
         *       - else if it's a path, treat it like case (2)
         *       - else: fail.
         *
         * case 4: git checkout <something> <paths>
         *
         *   The first argument must not be ambiguous.
         *   - If it's *only* a reference, treat it like case (1).
         *   - If it's only a path, treat it like case (2).
         *   - else: fail.
         *
         */
```

```c
                if (!arg[2]) { /* "--" */
                        if (!(ctx->flags & PARSE_OPT_KEEP_DASHDASH)) {
                                ctx->argc--;
                                ctx->argv++;
                        }
                        break;
                }
```

```
        if (opts->accept_pathspec && opts->accept_ref)
                parseopt_flags = PARSE_OPT_KEEP_DASHDASH;

        argc = parse_options(argc, argv, prefix, options,
                             usagestr, parseopt_flags);
```


git is an absolute beast. Is it important to keep this? Can't the user just pass `HEAD`?

mjesuele:
That is the most incredible thing I've read all week.

No, I do not feel strongly about preserving this behavior. I've sat here for a few minutes trying to think of why we shouldn't axe it. Does `--` _ever_ mean anything other than `HEAD` when used that way? If not, it imposes a pretty high cost for allowing the user to save two keystrokes...

mjesuele:
But then again, it will violate people's expectations in kind of a gnarly way if they're used to what Git does...

mjesuele:
So where do we land on this? Should @timsehn weigh in?

timsehn:
I don't know enough about this to care. My default position is do what get does even if it's weird and hard to implement.

mjesuele:
That's kind of where I'm at too.

So that said, @reltuk and @bheni, how do we want to handle this? What I've done effectively punts the `--` to each individual command to decide what to do with it, but we could also do something like what Git does and actually (ugh, gross) modify the argument list based on whether one of a certain class of commands was called.

PR #156, opened 2019-10-17
Bh/schema import

bheni:


go/cmd/dolt/commands/schcmds/import.go
```diff
@@ -16,26 +16,313 @@ package schcmds
...
+	var existingSch schema.Schema = schema.EmptySchema
+	if tblExists {
+		existingSch, err = tbl.GetSchema(ctx)
+
+		if err != nil {
+			return errhand.BuildDError("error: failed to read schema from '%s'", tblName).AddCause(err).Build()
+		}
+	}
+
+	val, pksOK := apr.GetValue(pksParam)
```
zachmu: Same for this and other uses of apr below, don't even pass apr into this function

go/cmd/dolt/commands/schcmds/import.go
```diff
@@ -16,26 +16,313 @@ package schcmds
...
+	if pksOK {
+		pks = strings.Split(val, ",")
+		goodPKS := make([]string, 0, len(pks))
+
+		for _, pk := range pks {
+			pk = strings.TrimSpace(pk)
+
+			if pk != "" {
+				goodPKS = append(goodPKS, pk)
+			}
```
zachmu: Probably should throw an error on malformed pk strings, no?

go/cmd/dolt/commands/schcmds/import.go
```diff
@@ -16,26 +16,313 @@ package schcmds
...
+		defer f.Close()
+
+		rd, err = csv.NewCSVReader(nbf, f, csv.NewCSVInfo().SetDelim(args.delim))
+
+		if err != nil {
+			return nil, errhand.BuildDError("error: failed to create a CSVReader.").AddCause(err).Build()
+		}
+
+		defer rd.Close(ctx)
+	}
```
zachmu: Probably should throw an error here on unsupported file types.

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -0,0 +1,419 @@
...
+// MapMapper is a StrMapper implementation that is backed by a map[string]string
+type MapMapper map[string]string
+
+// Map maps a string to another string.  If a string is not in the mapping ok will be false, otherwise it is true.
+func (m MapMapper) Map(str string) (string, bool) {
+	v, ok := m[str]
+	return v, ok
+}
+
+// InferenceArgs are arguments that can be passed to the schema inferrer to modify it's inference behavior.
```
zachmu: its

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -0,0 +1,419 @@
...
+
+	tag := uint64(0)
+	for i, name := range inf.colNames {
+		if mappedName, ok := inf.impArgs.ColMapper.Map(name); ok {
+			name = mappedName
+		}
+
+		_, partOfPK := inf.pkColToIdx[name]
+		typeToCount := inf.colType[i]
+		hasNegatives := inf.negatives[i]
```
zachmu: I think this is probably the wrong call. Given that we can't really change the type of a column and preserve history, we should be more permissive with unsigned ints, and make them signed by default (maybe controllable with a flag?)

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+
+			strVal := string(val.(types.String))
+			kind, hasNegs := leastPermissiveKind(strVal, inf.impArgs.FloatThreshold)
+
+			if hasNegs {
+				inf.negatives[i] = true
+			}
+
+			count, _ := inf.colType[i][kind]
```
zachmu: Does inf.colType[i][kind] += 1 not work?

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -0,0 +1,419 @@
...
+		}
+
+		switch kind {
+		case types.StringKind:
+			if nullable {
+				return types.StringKind, true
+			}
+
+		case types.UUIDKind:
+			//cli.PrintErrln(color.YellowString("warning: column %s has a mix of uuids and non uuid strings.", name))
```
zachmu: Remove these commented cli prints

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -0,0 +1,419 @@
...
+				}
+			} else {
+				kind = types.StringKind
+			}
+
+		case types.UintKind:
+			if t == types.IntKind {
+				if hasNegatives {
+					//cli.PrintErrln(color.YellowString("warning: %s has values larger than a 64 bit signed integer can hold, and negative numbers.  This will be interpreted as a string.", name))
+					kind = types.StringKind
```
zachmu: Seems like having negatives should only downgrade to an int, not a string

go/libraries/doltcore/env/actions/infer_schema_test.go
```diff
@@ -0,0 +1,578 @@
...
+
+func TestLeastPermissiveKind(t *testing.T) {
+	tests := []struct {
+		name           string
+		valStr         string
+		floatThreshold float64
+		expKind        types.NomsKind
+		expHasNegs     bool
+	}{
+		{"empty string", "", 0.0, types.NullKind, false},
```
zachmu: I think these are all gonna fail go vet

go/libraries/doltcore/env/actions/infer_schema.go
mjesuele: Syntactically it's fine but believe it or not go-lint would complain about it:
image

...I think inf.colType[i][kind]++ is fine, though kinda gross.

go/cmd/dolt/commands/schcmds/import.go
bheni: added more errors

go/cmd/dolt/commands/schcmds/import.go
bheni: fixed

go/libraries/doltcore/env/actions/infer_schema.go
bheni: we talked about this.  it is signed be default.

go/libraries/doltcore/env/actions/infer_schema.go
bheni: you actually have to have a 0 value in the map to use these don't you?  can you increment an int in a map if there is no value for the key?

bheni: just tested.  Answer is no.  I will update.

go/libraries/doltcore/env/actions/infer_schema.go
bheni: if it is a uint to begin with it's because it doesn't fit in a signed int64, so can't downgrade.

PR #157, opened 2019-10-18
Basic dolt blame

mjesuele:
This still needs some more BATS tests and maybe some UI touches like what @Hydrocharged suggested, but overall I think it's ready for some eyes.

The biggest thing I don't like is the logic surrounding pretty-printing of primary keys (you'll see why) but please do tell me if you notice other things that are janky.

In general, feedback and suggestions are very welcome.


mjesuele:
Checklist:

- [ ] More BATS tests, including merges and some assertions on primary keys
- [ ] Nice big doc comment for the top-level `Blame` function
- [ ] ???

bats/common.bash.bats
```diff
@@ -0,0 +1,28 @@
...
+  stash_current_dolt_user
+  [ "$STASHED_DOLT_USER_NAME" = "Bats Tests" ]
+  [ "$STASHED_DOLT_USER_EMAIL" = "bats@email.fake" ]
+
+  set_dolt_user "Bats Tests 1" "bats-1@email.fake"
+  NAME=`current_dolt_user_name`
+  EMAIL=`current_dolt_user_email`
+  [ "$NAME" = "Bats Tests 1" ]
+  [ "$EMAIL" = "bats-1@email.fake" ]
+
```
mjesuele: I imagine there's a way to interpolate the command result but I quickly tried a few other things and couldn't get any to work.

go/cmd/dolt/commands/blame_test.go
```diff
@@ -0,0 +1,45 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import "testing"
+
+func TestTruncateString(t *testing.T) {
```
mjesuele: This test (and the function it's testing) should maybe live somewhere else; not sure where yet, there are still corners of the project structure I'm not familiar with.

go/cmd/dolt/commands/log.go
```diff
@@ -103,8 +103,8 @@ func logWithLoggerFunc(ctx context.Context, commandStr string, args []string, dE
 	return logCommits(ctx, dEnv, cs, loggerFunc, numLines)
 }
 
-func getCommitSpec(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) (*doltdb.CommitSpec, error) {
-	if apr.NArg() == 0 {
+func parseCommitSpec(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) (*doltdb.CommitSpec, error) {
```
mjesuele: Perhaps this should be factored out somewhere too now that it's also being used in blame.

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	blameGraph, err := blameGraphFromRows(ctx, rows)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, c := range commits {
```
reltuk: I think this would be better with the loops inverted, and possibly converted to a pipeline?

Since we're going row by row and getting the diff that way, we can (psuedo-y):

```go
for _, row := range rows {
    rowPK := row
    for _, c := range commits {
        parent, err := dEnv.DoltDB.ResolveParent(ctx, c, 0)
        changed, err := rowChanged(ctx, parent, c, tableName, rowPK)
        if changed {
            emit(rowPK, c)
            break
        }
}
```

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+		return err
+	}
+
+	cli.Println(blameGraph.String(ctx, pkColNames))
+	return nil
+}
+
+func blameGraphFromCommit(ctx context.Context, dEnv *env.DoltEnv, commit *doltdb.Commit, tableName string) (*blameGraph, error) {
+	// get the commits sorted from newest to oldest ending with `commit`
+	commits, err := actions.TimeSortedCommits(ctx, dEnv.DoltDB, commit, -1)
```
reltuk: Do we want time sorted or topological here?

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+		// get the first parent
+		parent, err := dEnv.DoltDB.ResolveParent(ctx, c, 0)
+		if err != nil {
+			return nil, err
+		}
+
+		// for each unblamed node, see if it changed between `c` and `parent`.
+		// if so, mark it blamed with `c` as the blame origin
+		for _, rowPK := range blameGraph.UnblamedNodeKeys() {
+			changed, err := rowChanged(ctx, parent, c, tableName, rowPK)
```
reltuk: Should we be comparing the value of the row in `HEAD` against the value of the row in this commit, and taking the last commit where the value is the same? Is the most recent time sorted commit where the value of this node changed against its parent the same thing? Maybe they only ever differ for intermediate schema changes where the columns that are in `c` and `parent` aren't in `HEAD` anymore...?

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	return str[0:maxLength-1] + "…"
+}
+
+var dataColNames = []string{"Commit Msg", "Author", "Time", "Commit"}
+
+// String returns the string representation of this blame graph
+func (bg *blameGraph) String(ctx context.Context, pkColNames []string) string {
+	// here we have two []string and need one []interface{} (aka table.Row)
```
reltuk: Seems right to me. A slight variation that comes to mind:

```go
header := []interface{}{}
for _, cellText := range append(pkColNames, dataColNames...) {
	header = append(header, cellText)
}
```

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		return fmt.Errorf("error getting commit hash: %v", err)
+	}
+
+	meta, err := c.GetCommitMeta()
+	if err != nil {
+		return fmt.Errorf("error getting metadata for commit %s: %v", commitHash.String(), err)
+	}
+
+	pkHash, err := rowPK.Hash(types.Format_7_18)
```
reltuk: Can we thread the right nbf from the database here?

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	if newRow != nil && oldRow == nil {
+		return true, nil
+	}
+
+	oldSchema, err := oldTable.GetSchema(ctx)
+	if err != nil {
+		return false, fmt.Errorf("error getting schema from %s in old commit %s: %v", tableName, oldHash, err)
+	}
+
+	return !row.AreEqual(*oldRow, *newRow, oldSchema), nil
```
reltuk: Does `row.AreEqual` return false if a column was added in `newTable`'s schema, `newRow` was populated with a non-nil value for that column and then we called `row.AreEqual(*oldRow, *newRow, oldSchema)`?

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+func blameGraphFromRows(ctx context.Context, rows types.Map) (*blameGraph, error) {
+	graph := make(blameGraph)
+	err := rows.IterAll(ctx, func(key, _ types.Value) error {
+		hash, err := key.Hash(types.Format_7_18)
```
reltuk: Thread the nbf here.

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	if !ok {
+		return nil, nil
+	}
+
+	return &row, err
+}
+
+func schemaFromCommit(ctx context.Context, c *doltdb.Commit, tableName string) (schema.Schema, error) {
+	t, err := maybeTableFromCommit(ctx, c, tableName)
```
reltuk: Would be great to package these (`schema` and `t`) up with the `commits` slice we're ranging up in `blameGraphFromCommit` since we're doing this same work for every row * every commit.

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	blameGraph, err := blameGraphFromRows(ctx, rows)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, c := range commits {
+		// get the first parent
+		parent, err := dEnv.DoltDB.ResolveParent(ctx, c, 0)
```
reltuk: Would be great to package this up with the values in `commits` themselves since we're doing this same work every row * every commit.

go/cmd/dolt/commands/blame.go
mjesuele: Hah, now that I'm looking at it again, is there even a need to get the full list of commits ahead of time? Could probably just keep getting the first parent and using that as the child for the next iteration, right?

go/cmd/dolt/commands/blame.go
mjesuele: Dude I had no idea what I was doing here, happy to talk about it.

go/cmd/dolt/commands/blame.go
reltuk: Maybe, but I think we want to walk non-first-parents too, even if we only compare against first parent in the case of a merge commit?

go/cmd/dolt/commands/blame.go
mjesuele: Same comment as previous.

go/cmd/dolt/commands/blame.go
mjesuele: Rad, thanks.

go/cmd/dolt/commands/blame.go
mjesuele: Yeah I know there's a decent bit of extra work being done in this function. If things are looking good overall I'm happy to start zeroing in on optimizations and the like.

go/cmd/dolt/commands/blame.go
reltuk: You can take it from the table or something:

`go/libraries/doltcore/doltdb/table.go`:

```go
func (t *Table) Format() *types.NomsBinFormat {
        return t.vrw.Format()
}
```

It's just a sentinel value that will effect how values serialize themselves at the noms layer.

mjesuele: Perfect, thank you

go/cmd/dolt/commands/blame.go
mjesuele: Makes sense.

go/cmd/dolt/commands/blame.go
mjesuele: I think it should be equivalent. If the blame of a row is the most recent commit where the row changed against its parent, and schema changes count as changes to every row (I am not confident that this is actually what happens but I think it's what we want, right?), then intermediate schema changes like this are precluded.

Am I missing something?

go/cmd/dolt/commands/blame.go
mjesuele: Yeah that's a good question. The existing tests don't consider this case but I'm trying to add it now.

go/cmd/dolt/commands/blame.go
mjesuele: What do you think of the temporal topological sort described on here? https://pvigier.github.io/2019/05/06/commit-graph-drawing-algorithms.html

go/cmd/dolt/commands/blame.go
mjesuele: Totally on board to make it a pipeline. The way pretty-printing works, we still need all the output before printing anything to the screen, but a) we could do something different for that, and b) other users of blame may not have the same constraints.

I spent some time thinking about the loop inversion and couldn't come up with any good objections; I'll give the change a try.

go/cmd/dolt/commands/blame.go
reltuk: Yeah, I think a topological order with concurrent commits ordered by timestamp is absolutely the thing we want.

This code does a topological sort with ties broken by "height of the tree", and some extra stuff for excluding certain commits from the walk:

https://github.com/liquidata-inc/dolt/blob/master/go/libraries/doltcore/env/actions/commitwalk/commitwalk.go#L133

But it might be the basis for a simpler thing that breaks ties with commit timestamp. (I think basically rip out `SetInvisible` and change `height` and `height` comparisons to a timestamp).

go/cmd/dolt/commands/blame.go
reltuk: I think pretty printing at the end of a pipeline just samples the first `n` rows and picks column sizes from that. But I could be wrong...definitely Zach and Brian are the people that know about how this currently works...

go/cmd/dolt/commands/blame.go
reltuk: Thinking about this further, the reason it's currently height instead of timestamp is that that's the way it gets a topological sort...it needs to be dfs as well, which it isn't in the existing code because it's trying to bail early in the two dot log case.

go/cmd/dolt/commands/blame.go
mjesuele: Boom: https://github.com/liquidata-inc/dolt/pull/157/commits/0665eb6fa9aff9442d12f20c33532893ee7e16f7

go/cmd/dolt/commands/blame.go
mjesuele: Okay, a bunch of inputs now get precomputed and packaged: https://github.com/liquidata-inc/dolt/compare/0665eb6fa9aff9442d12f20c33532893ee7e16f7...2721ea4

reltuk: Cool!

go/cmd/dolt/commands/blame.go
mjesuele: Okay, I think this is right: https://github.com/liquidata-inc/dolt/pull/157/commits/10cec98af42ffadaebdea442a7681c2e80638e84

go/libraries/doltcore/env/actions/commitwalk/commitwalk.go
```diff
@@ -101,33 +117,32 @@ func (q *q) load(ctx context.Context, h hash.Hash) (*doltdb.Commit, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	h, err := l.Height()
+	if err != nil {
+		return nil, err
+	}
+	c := &c{commit: l, height: h}
+	q.loaded[id] = c
+	return c, nil
```
mjesuele: I just did this to appease go-lint

mjesuele:
Cleaned up the commit history after implementing feedback. I think this is ready for one more look before merge.

go/cmd/dolt/commands/blame.go
```diff
@@ -0,0 +1,499 @@
...
+	// if the row is in the parent table but not the child one...something's wrong. bail!
+	if parentRow != nil && childRow == nil {
+		return false, fmt.Errorf("expected to find row with PK %v in table %s in child commit %s, but didn't", rowPK, input.TableName, input.Hash)
+	}
+	// if the row is in the child table but not the parent one, it must be new; return true
+	if childRow != nil && parentRow == nil {
+		return true, nil
+	}
+
+	return !row.AreEqual(*parentRow, *childRow, input.ParentSchema), nil
```
reltuk: Did we get an answer to the question about a new column in `childRow` with a non-nil value returning `false` here?

mjesuele: No, I wasn't sure how to update the test. I'm sure I could do so with more effort, but I was thinking...if schema changes count as a change to every row (for now) we can actually avoid this case by checking equality of parent and child schemas.

I added [some code](https://github.com/liquidata-inc/dolt/pull/157/commits/26212307bdf6a25cd2bfca1d4d6bed477f9d0ffc) for this that seems to be behaving well. It probably deserves a BATS test, though.

Does that make sense?

mjesuele:
Sweet! Thanks for the reviews.

PR #167, opened 2019-10-23
dolt/{go,bats}: Transform empty strings as null from csv

katiemcculloch:


katiemcculloch:
@bheni Just pushed another commit, this is ready for review again.

go/libraries/doltcore/type_conversion.go
```diff
@@ -145,6 +151,10 @@ func convStringToFloat(val types.Value) (types.Value, error) {
 		return nil, nil
 	}
 
+	if val.Equals(types.String("\"\"")) {
```
bheni: If the value in the csv is """""" then the val should be "\"\"" I don't think we want this to be Null.  It seems like that should be an error.  The same goes for the ones below.

go/libraries/doltcore/type_conversion_test.go
```diff
@@ -42,6 +42,20 @@ func TestConv(t *testing.T) {
...
 		{types.String("true"), types.Bool(true), convStringToBool, false},
 		{types.String("anything"), types.NullValue, convToNullFunc, false},
+		{types.String(""), types.String(""), identityConvFunc, false},
+		{types.String(""), types.NullValue, convStringToUUID, false},
+		{types.String(""), types.NullValue, convStringToUint, false},
+		{types.String(""), types.NullValue, convStringToInt, false},
+		{types.String(""), types.NullValue, convStringToFloat, false},
+		{types.String(""), types.NullValue, convStringToBool, false},
+		{types.String(""), types.NullValue, convToNullFunc, false},
+		{types.String("\"\""), types.String(""), identityConvFunc, false},
```
bheni: The expected value should be types.String("\"\"") same as what was input.

go/libraries/doltcore/type_conversion_test.go
```diff
@@ -42,6 +42,20 @@ func TestConv(t *testing.T) {
...
+		{types.String("\"\""), types.NullValue, convStringToUUID, false},
```
bheni: I think all of these should error

katiemcculloch:
Re-did this work in #199

PR #172, opened 2019-10-25
Types package simplification

Hydrocharged:


Hydrocharged:
I'll fix the merge conflicts once the following PR has been merged: https://github.com/liquidata-inc/dolt/pull/167

go/store/types/make_type.go
```diff
@@ -128,17 +106,11 @@ func makePrimitiveType(k NomsKind) *Type {
...
-var UUIDType = makePrimitiveType(UUIDKind)
-var IntType = makePrimitiveType(IntKind)
-var UintType = makePrimitiveType(UintKind)
-var NullType = makePrimitiveType(NullKind)
-var InlineBlobType = makePrimitiveType(InlineBlobKind)
+// PrimitiveTypeMap auto populates with Value types that return true from isPrimitive().
+// Only include a type here manually if it has no associated Value type.
+var PrimitiveTypeMap = map[NomsKind]*Type{
+	ValueKind: makePrimitiveType(ValueKind),
+}
```
bheni: The more I read this code, the more I question this change.  Maybe I'm not understanding the intention of this change, but I'm not sure it adds much value and it adds an unnecessary level of indirection.  In almost every case this is used like: 

    PrimitiveTypeMap[types.SOMEKIND] 

where SOMEKIND is known at compilation time.  What is the goal of this change?

Hydrocharged: So I wanted to make the `Value` interface as complete as possible, in that to add a new type, as long as you satisfy `Value`, you've no other changes to make (couldn't get around defining the kind in `noms_kind.go` without breaking _everything_, but I definitely tried). At least for the types that we need to add for SQL compliance (which I'm assuming will be the majority of the new types we'll be adding). With that, I added the `isPrimitive()` function so that the primitive determination is moved from the declaration being in `make_type.go` to whatever the result of `isPrimitive()` is.

From the 680 usages of `PrimitiveTypeMap` in the codebase, excluding the definitions and `typeOf` functions, there are only 27 calls to it that aren't in a `_test.go` file, so at least the prevalence of the changes just comes from the tests doing a lot more type-related stuff than the actual codebase.

The only primitive kind that doesn't have an associated `Value` is `ValueKind`, which is why it's special-cased here, but that shouldn't be true for most/any new types we'll be adding in the near future.

So right now, to add a new primitive type, it's adding the file and implementing the `Value` interface, adding the relevant `NomsKind`, and adding mappings in `KindToType` and `KindToString` (which the latter 3 are all linearly listed in `noms_kind.go`). Without the `PrimitiveTypeMap` change, it would be the above, plus a new case in `noms_kind.go -> IsPrimitiveKind()`, a new case in `make_type.go -> MakePrimitiveType()`, and of course the type instantiation in `make_type.go`. It's not a ton more work, but I prefer the simplification.

go/store/types/value.go
```diff
@@ -73,13 +86,31 @@ type Value interface {
...
+
+	// readFrom reads the encoded version of the value from a binaryNomsReader
+	readFrom(*NomsBinFormat, *binaryNomsReader) (Value, error)
+
+	// skip takes in a binaryNomsReader and skips the encoded version of the value
+	skip(*NomsBinFormat, *binaryNomsReader)
+}
+
+// PrimitiveValue is the interface all primitive values implement.
+type DoltValue interface {
```
bheni: Is this still used?

Hydrocharged: It is not, whoops!

PR #179, opened 2019-10-29
clone reliability

bheni:


bheni:
successfully cloned the Liquidata/primes repo on hotel wifi. 16Gs in a little over 2.5 hours.

    ~/datasets>time dolt clone Liquidata/primes
    cloning https://doltremoteapi.beta.dolthub.com/Liquidata/primes
    4,036,451 of 4,036,451 chunks complete. 0 chunks being downloaded currently.
    
    real	152m4.456s
    user	3m29.330s
    sys	5m45.656s

go/store/datas/pull.go
```diff
@@ -91,6 +92,18 @@ type TableFileEvent struct {
 	TableFiles []nbs.TableFile
 }
 
+func mapTableFiles(tblFiles []nbs.TableFile) ([]string, map[string]nbs.TableFile) {
```
zachmu: Comment describing the return type would be nice (table files keyed by ID)

go/store/datas/pull.go
```diff
@@ -102,56 +115,108 @@ func clone(ctx context.Context, srcTS, sinkTS nbs.TableFileStore, eventCh chan<-
...
 		eventCh <- TableFileEvent{Listed, tblFiles}
 	}
 
-	// should parallelize at some point
-	for _, tblFile := range tblFiles {
-		if eventCh != nil {
-			eventCh <- TableFileEvent{DownloadStart, []nbs.TableFile{tblFile}}
-		}
+	i := 0
+	op := func() error {
```
zachmu: Would call this download or similar

go/store/datas/pull.go
```diff
@@ -102,56 +115,108 @@ func clone(ctx context.Context, srcTS, sinkTS nbs.TableFileStore, eventCh chan<-
...
 			}
 
-			return nil
-		}()
+			i++
+		}
 
 		if err != nil {
+			var sourcesErr error
+			_, tblFiles, sourcesErr = srcTS.Sources(ctx)
```
zachmu: Not obvious why this would return a different set of table files than it did outside this function. Could use an explanatory comment

bheni: It shouldn't return a different set.  But it will refresh the signed urls.  Not the best solution but doesn't require a server deployment.

go/store/datas/pull.go
```diff
@@ -102,56 +115,108 @@ func clone(ctx context.Context, srcTS, sinkTS nbs.TableFileStore, eventCh chan<-
...
+	op := func() error {
+		var err error
+		for i < len(desiredFiles) {
+			fileID := desiredFiles[i]
+			tblFile, ok := fileIDToTF[fileID]
 
-		err := func() (err error) {
-			var rd io.ReadCloser
-			rd, err = tblFile.Open()
+			if !ok {
```
zachmu: If the set of files in fileIDToTF can't change from reassignment below, then I don't understand the point of this check or the conjoin comment. It seems like this is unreachable.

bheni: fileIDToTF can change.  For instance if there was a conjoin that ran between the first call to srcTS.Sources and the second then the fileIDToTF map would have different fileIDs.

go/store/datas/pull.go
bheni: added

go/store/datas/pull.go
bheni: updated

go/store/datas/pull.go
bheni: comments added.

PR #182, opened 2019-10-31
Moved SQL types to an interface

Hydrocharged:
Have a look! Just make an empty struct type that implements `SqlTypeInit` and add the struct to `sqlTypeInitializers` and you've got a type that works in SQL now!


go/libraries/doltcore/sqle/types/uuid.go
```diff
@@ -0,0 +1,64 @@
...
+func (uuidType) Value() types.Value {
+	return types.KindToType[types.UUIDKind]
+}
+
+func (uuidType) PrimaryType() sql.Type {
+	return sql.Text
+}
+
+func (uuidType) RepresentingTypes() []sql.Type {
+	return []sql.Type{}
```
zachmu: At some point, we need to need to handle SQL-> noms conversion for UUID strings. What you have now assumes a 1:1 mapping between all the types, but that won't be good enough for UUIDs. Would add a note here.

PR #184, opened 2019-11-01
RFC on joiner

bheni:


go/libraries/doltcore/schema/schema_impl.go
```diff
@@ -44,7 +44,7 @@ func SchemaFromCols(allCols *ColCollection) Schema {
 	}
 
 	if len(pkCols) == 0 {
-		panic("no primary key columns specified")
```
bheni: should this be legal?

bheni: Maybe we should enforce pks on schemas in a different place.  Such as when you attempt to create a table with the schema.

bheni: I think we must enforce it in other places already, otherwise we would be seeing a bunch of panics.

zachmu: We have UnkeyedSchemaFromColsfor this, which doesn't enforce the primary key check. We use that for e.g. SQL result sets and other unkeyed schemas.

PR #189, opened 2019-11-02
diff source refactor

bheni:


go/libraries/doltcore/diff/diff_splitter.go
```diff
@@ -0,0 +1,136 @@
...
+	joiner  *rowconv.Joiner
+	oldConv *rowconv.RowConverter
+	newConv *rowconv.RowConverter
+}
+
+func NewDiffSplitter(joiner *rowconv.Joiner, oldConv, newConv *rowconv.RowConverter) *DiffSplitter {
+	return &DiffSplitter{joiner, oldConv, newConv}
+}
+
+func a(rows map[string]row.Row, name string, rc *rowconv.RowConverter) (row.Row, error) {
```
zachmu: This could use a better name

go/libraries/doltcore/rowconv/joiner.go
```diff
@@ -0,0 +1,165 @@
...
+
+type Joiner struct {
+	srcSchemas map[string]schema.Schema
+	tagMaps    map[string]map[uint64]uint64
+	revTagMap  map[uint64]stringUint64Tuple
+	tags       map[string][]uint64
+	joined     schema.Schema
+}
+
+func NewJoiner(namedSchemas []NamedSchema, namers map[string]ColNamingFunc) (*Joiner, error) {
```
zachmu: A comment about the high-level purpose of the joiner interface and what it does would be nice

bheni: updated

PR #190, opened 2019-11-02
Missed Kind to String

Hydrocharged:
In my last PR, it looks like I missed that we were using the old `DoltToSQLType` hardcoded map from the original SQL implementation. I didn't change it everywhere (it's used heavily in the old SQL code that isn't even being called anymore), but it's changed where it matters. Added a new interface function and changed the printing code to be a bit more consistent (we were mixing uppercase with lowercase).

I'm also returning different values, such as `BIGINT` for `sql.Int64`, as `int` parses in MySQL to a 32-bit integer, which isn't correct. Essentially made it so that if you took the `CREATE` statement exactly as-is and exported your data to a bunch of inserts and ran it in MySQL then it wouldn't error out, as it previously would have.


Hydrocharged:
I forgot to fix those old tests, I'll do that tomorrow.

PR #191, opened 2019-11-02
fix branch name panic with period

coffeegoddd:
Looked into supporting periods in branch names, but it looks like `noms` relies on periods specifically pretty heavily. Seems to be excluded from the regex below by design, since they build some types on the expectation that a branch name or `ref` contain  a period.

My understanding is that a user's branch name is used to look up a particular dataset within the `noms` layer and this variable (go/store/datas/dataset.go): 
```
// DatasetRe is a regexp that matches a legal Dataset name anywhere within the
// target string.
var DatasetRe = regexp.MustCompile(`[a-zA-Z0-9\-_/]+`)
```
acts as the regex source of "truth" for branch names/ dataset look ups, and I believe more. 
Noms also expects to be able to append a `.` to this string in order to parse the string later and correctly create it's `Path` types...

I went down a rabbit hole trying to change all of the `noms` `Path` delimiters to be a different character, but the changes go pretty deep and start breaking a lot of things. Happy to continue down that course in order to support periods in branch names, but it might take me a bit of time change everything. I'm also not sure what character should replace the period... asterisk? Anyway, this PR seemed like low hanging fruit fix to resolve the panic at least.

### DIFF
--- a/bats/empty-repo.bats
+++ b/bats/empty-repo.bats
@@ -203,9 +203,8 @@ teardown() {
     [ "$output" = "" ]
 }
 
-@test "unsupported branch characters" {
-    run dolt branch "are.dots.supported"
-    skip "branch names with . panic right now"
-    [ "$status" -eq 0 ]
-    [ "$output" = "" ]
+@test "branch names do not support ." {
+    run dolt branch "dots.are.not.supported"
+    [ "$status" -eq 1 ]
+    [ "$output" = "fatal: 'dots.are.not.supported' is an invalid branch name." ]
 }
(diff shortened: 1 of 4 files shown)


reltuk:
Long term not supporting periods seems problematic...common in version numbers, especially when/if we support tags?

bats/empty-repo.bats
```diff
@@ -203,9 +203,8 @@ teardown() {
...
-    run dolt branch "are.dots.supported"
-    skip "branch names with . panic right now"
-    [ "$status" -eq 0 ]
-    [ "$output" = "" ]
-}
\ No newline at end of file
+@test "branch names do not support ." {
+    run dolt branch "dots.are.not.supported"
+    [ "$status" -eq 1 ]
+    [ "$output" = "fatal: 'dots.are.not.supported' is an invalid branch name." ]
```
coffeegoddd: added `output` check

timsehn:
> Long term not supporting periods seems problematic...common in version numbers, especially when/if we support tags?

Agreed. This is just to fix the panic. We need to be able to use "." in branch names and tags when supported.

PR #195, opened 2019-11-06
Added Timestamp to Dolt and Datetime to SQL

Hydrocharged:
Have a look!

I ran into an import cycle issue that I just could not figure out how to avoid, except by putting the tests into their own test folder (`sqle/types/tests`), so that's why they're in there. In particular, the cycle was that `sqle` imports `sqle/types`, and the tests rely on (and thus must import) `sqle`, causing the cycle.

I'm thinking of adding tests for the other SQL types later so that we have a few more built-in tests using the server portion, rather than everything using the `-q` pathway. That will be a different/future PR though.

### DIFF
--- a/go/libraries/doltcore/sqle/types/tests/datetime_test.go
+++ b/go/libraries/doltcore/sqle/types/tests/datetime_test.go
@@ -1,0 +1,87 @@
+// Copyright 2019 Liquidata, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/src-d/go-mysql-server/sql"
+
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+func TestDatetimeQueries(t *testing.T) {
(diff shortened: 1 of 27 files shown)


go/libraries/doltcore/sqle/types/tests/datetime_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/src-d/go-mysql-server/sql"
+
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+func TestDatetimeQueries(t *testing.T) {
+	tests := []struct {
+		inputSQLVal interface{}
+		inputValue  types.Timestamp
+	}{
```
zachmu: Should have some test cases for non-UTC timezones

go/libraries/doltcore/sqle/types/tests/datetime_test.go
```diff
@@ -0,0 +1,76 @@
...
+		"9999-12-31 23:59:59",
+		"1970-01-01 00:00:00",
+		"1610-06-08 09:10:37",
+		"1945-05-08 19:33:41",
+		"3005-02-14 10:02:25",
+		"2019-11-01 03:23:48",
+	}
+
+	for _, test := range tests {
+		roundTrip(t, test, sql.Datetime)
```
zachmu: What's the latency of starting and stopping a server for every test case? Worried about these being too slow.

go/store/types/timestamp.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/liquidata-inc/dolt/go/store/hash"
+)
+
+const (
+	timestampNumBytes = 15
+	timestampFormat   = "2006-01-02 15:04:05 -0700"
+)
+
+type Timestamp time.Time
```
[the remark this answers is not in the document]
Hydrocharged: Nope, `types.Value` was expanded to handle more functionality, so as long as you implement that interface you're good.

go/libraries/doltcore/sqle/types/tests/datetime_test.go
Hydrocharged: Funny enough it looks like the queries themselves are what is taking the time, not the starting/stopping of the server.

PR #198, opened 2019-11-07
checkout a remote only branch

bheni:


go/cmd/dolt/commands/checkout.go
```diff
@@ -17,6 +17,8 @@ package commands
 import (
 	"context"
 
+	ref2 "github.com/liquidata-inc/dolt/go/libraries/doltcore/ref"
```
zachmu: What is wrong with you

bheni: fixed.

bheni: I mean I fixed the import.  What is wrong with me can't be fixed.

PR #212, opened 2019-11-13
Zachmu/sql indexes2

zachmu:


go/libraries/doltcore/sqle/indexes.go
```diff
@@ -136,7 +136,30 @@ func keyColToValue(v interface{}, column schema.Column) types.Value {
 	case types.FloatKind:
 		return types.Float(v.(float64))
```
Hydrocharged: Should this also have a switch case that checks for `float32` as well?

PR #213, opened 2019-11-13
Zachmu/sql performance

zachmu:


go/libraries/doltcore/sqle/tables.go
```diff
@@ -183,10 +219,43 @@ func (t *DoltTable) Update(ctx *sql.Context, oldRow sql.Row, newRow sql.Row) err
 			}
 			return fmt.Errorf("primary key collision: (%v)", strings.Join(newRowAsStrings, ", "))
 		}
-		keyToDelete = dOldKey
+		u.ed.Remove(dOldKey)
```
Hydrocharged: We need to set the new key right? Here we're just removing the old one without doing anything else.

zachmu: Correct. I ended up rewriting update because of this issue, take a look if you're interested.

PR #214, opened 2019-11-15
Bh/super schema

bheni:


go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -0,0 +1,267 @@
...
+	Name string
+
+	// Kind is the NomsKind of the column
+	Kind types.NomsKind
+}
+
+// SuperSchemaGen is a utility class used to generate the superset of several schemas.
+type SuperSchemaGen struct {
+	tagKindToDestTag map[TagKindPair]uint64
+	usedTags         map[uint64]struct{}
```
zachmu: For existence-check maps like this, I find it's way easier to use bool as the value type. Then you can use the simpler key lookup syntax (instead of _, ok)

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -0,0 +1,267 @@
...
+
+		for _, nameKindPair := range additionalCols {
+			var err error
+			colColl, err = colColl.Append(schema.NewColumn(nameKindPair.Name, nextReserved, nameKindPair.Kind, false))
+
+			if err != nil {
+				return nil, err
+			}
+
+			nextReserved++
```
zachmu: Technically you should have the more complicated tag logic above in here too, unless you can guarantee that the reserved tags will always be assigned sequentially

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	return schema.UnkeyedSchemaFromCols(colColl), nil
+}
+
+// AddHistoryOfTableAtCommit will traverse a commit graph adding all versions of a tables schema to the schemas being
+// supersetted.
+func (ss *SuperSchemaGen) AddHistoryOfTableAtCommit(ctx context.Context, tblName string, ddb *doltdb.DoltDB, cm *doltdb.Commit) error {
+	addedSchemas := make(map[hash.Hash]struct{})
```
zachmu: Same comment about bool value types

go/libraries/doltcore/rowconv/super_schema_test.go
```diff
@@ -0,0 +1,259 @@
...
+	const tblName = "test_table"
+	initialSch := mustSchema(idColTag0TypeUUID, firstColTag1TypeStr, lastColTag2TypeStr)
+	addAddrAt3Sch := mustSchema(idColTag0TypeUUID, firstColTag1TypeStr, lastColTag2TypeStr, addrColTag3TypeStr)
+	addAgeAt3Sch := mustSchema(idColTag0TypeUUID, firstColTag1TypeStr, lastColTag2TypeStr, ageColTag3TypeInt)
+	readdAgeAt4Sch := mustSchema(idColTag0TypeUUID, firstColTag1TypeStr, lastColTag2TypeStr, addrColTag3TypeStr, ageColTag4TypeUint)
+
+	ctx := context.Background()
+	dEnv := envtestutils.CreateInitializedTestEnv(t, ctx)
+
+	history := envtestutils.HistoryNode{
```
zachmu: Probably want to have a test case that involves a merge, right?

go/libraries/doltcore/rowconv/super_schema.go
bheni: fixed

go/libraries/doltcore/rowconv/super_schema_test.go
bheni: not supported yet.  Maybe one day.

PR #215, opened 2019-11-15
diff table

bheni:


go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+	ad      *diff.AsyncDiffer
+	diffSrc *diff.RowDiffSource
+	sch     schema.Schema
+	to      string
+	from    string
+	fromTag uint64
+	toTag   uint64
+}
+
+func NewDiffRowItr(joiner *rowconv.Joiner, rowDataFrom, rowDataTo types.Map, convFrom, convTo *rowconv.RowConverter, from, to string, fromTag, toTag uint64) *DiffRowItr {
```
zachmu: Probably should be unexported

zachmu: fromHash, toHash

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+	ad := diff.NewAsyncDiffer(1024)
+	ad.Start(context.TODO(), rowDataTo, rowDataFrom)
```
zachmu: Might as well plumb through the context from PartitionRows

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+	sch, err := tbl.GetSchema(ctx)
+
+	if err != nil {
+		return types.EmptyMap, nil, err
+	}
+
+	return data, sch, nil
+}
+
+func (dt *DiffTable) PartitionRows(ctx *sql.Context, part sql.Partition) (sql.RowIter, error) {
```
zachmu: You're not gracefully handling the case where there isn't a WHERE clause that specifies from_commit and to_commit

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+	return handled
+}
+
+func (dt *DiffTable) WithFilters(filters []sql.Expression) sql.Table {
+	ctx := context.TODO()
+
+	for _, f := range filters {
+		var fieldName string
+		var value string
+		expression.Inspect(f, func(e sql.Expression) bool {
```
zachmu: This is going to be overly broad and match filters we don't support, like >, <, etc. You should instead look for *expression.Equal exclusively, then run what you have here on the Equal expression.

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+	}()
+
+	return nil
+}
+
+func (dt *DiffTable) HandledFilters(filters []sql.Expression) []sql.Expression {
+	handled := make([]sql.Expression, 0, len(filters))
+	for _, f := range filters {
+		expression.Inspect(f, func(e sql.Expression) bool {
+			if e, ok := e.(*expression.GetField); ok {
```
zachmu: Same comment as below. You probably should restrict this to *expression.Equal

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -0,0 +1,324 @@
...
+			case *expression.GetField:
+				fieldName = val.Name()
+			case *expression.Literal:
+				value = val.String()
+			}
+
+			return true
+		})
+
+		value = strings.Trim(value, " \t\n\r\"")
```
zachmu: Rather than panicking down below, you should either a) defer this verification logic to when you can return an error (PartitionRows), or b) save such as error to the dt struct and return it in PartitionRows

go/libraries/doltcore/sqle/diff_table.go
bheni: this isn't necessarily a hash.  could be a branch name.

go/libraries/doltcore/sqle/diff_table.go
bheni: done

go/libraries/doltcore/sqle/diff_table.go
bheni: default values for dt.fromRoot/dt.toRoot/dt.fromCommitVal/dt.toCommitVal are set in NewDiffTable

go/libraries/doltcore/sqle/diff_table.go
bheni: fixed

go/libraries/doltcore/sqle/diff_table.go
bheni: fixed

go/libraries/doltcore/sqle/diff_table.go
bheni: verified with a test

PR #220, opened 2019-11-21
update go-mysql-server dependency

bheni:
Signed-off-by: Brian Hendriks <brian@liquidata.co>


go/libraries/doltcore/sqle/database.go
```diff
@@ -48,47 +51,54 @@ func (db *Database) Name() string {
...
 
-		sch, err := table.GetSchema(ctx)
-		// TODO: fix panics
-		if err != nil {
-			panic(err)
-		}
+	if err != nil {
+		return nil, false, err
+	} else if !ok {
+		panic("Name had already been verified... This is a bug")
```
zachmu: Probably want to include the string being matched in the panic message

bheni: updated

PR #221, opened 2019-11-21
improved super schema names

bheni:


go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -111,55 +145,93 @@ func (ss *SuperSchemaGen) AddSchema(sch schema.Schema) error {
...
+	namedCols := ssg.nameCols()
+
 	colColl, _ := schema.NewColCollection()
-	for tagKind, destTag := range ss.tagKindToDestTag {
-		colName := fmt.Sprintf("%d_%s", tagKind.Tag, tagKind.Kind.String())
+	for tagKind, colName := range namedCols {
+		destTag, ok := ssg.tagKindToDestTag[tagKind]
+
+		if !ok {
+			panic("mismatch between namedCols and tagKindToDestTag")
```
zachmu: More appropriate to return an error here?

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -111,55 +145,93 @@ func (ss *SuperSchemaGen) AddSchema(sch schema.Schema) error {
...
+		}
+
 		col := schema.NewColumn(colName, destTag, tagKind.Kind, false)
 
 		var err error
 		colColl, err = colColl.Append(col)
 
 		if err != nil {
-			return nil, err
+			return SuperSchema{}, err
```
zachmu: Probably should be a pointer type now that you're returning a struct rather than an interface

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -16,29 +16,73 @@ package rowconv
...
+		panic("Bug: super schema not generated.")
+	}
+
+	return ss.sch
+}
+
+// RowConvForSuperSchema creates a RowConverter for transforming rows with the the given schema to this super schema.
 // This is done by mapping the column tag and type to the super schema column representing that tag and type.
-func RowConvForSuperSchema(sch, super schema.Schema) (*RowConverter, error) {
+func (ss SuperSchema) RowConvForSuperSchema(sch schema.Schema) (*RowConverter, error) {
```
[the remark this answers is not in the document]
bheni: updated

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -111,55 +145,93 @@ func (ss *SuperSchemaGen) AddSchema(sch schema.Schema) error {
 	return nil
 }
 
+func (ssg *SuperSchemaGen) nameCols() map[TagKindPair]string {
+	colNames := map[string][]TagKindPair{}
```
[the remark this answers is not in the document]
bheni: no.  changed to make.

go/libraries/doltcore/rowconv/super_schema.go
```diff
@@ -111,55 +145,93 @@ func (ss *SuperSchemaGen) AddSchema(sch schema.Schema) error {
...
+	for tagKind, names := range ssg.names {
+		name := fmt.Sprintf("%d_%s", tagKind.Tag, tagKind.Kind.String())
+		if names.Size() == 1 {
+			name = names.AsSlice()[0]
```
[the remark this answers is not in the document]
bheni: No Zach.

go/libraries/doltcore/rowconv/super_schema.go
bheni: Wanted it to be clearly immutable.

go/libraries/doltcore/rowconv/super_schema.go
bheni: if we hit this, there is a bug.  Likely it occurred when adding schemas.  Want to panic, because it should never get here.

PR #222, opened 2019-11-22
dolt_docs feature branch

katiemcculloch:


katiemcculloch:
After talking with Zach this morning, I'm going to move the dolt_doc table into the existing rootValue struct, in the tablesMap. This is instead of having a separate docsMap, and will allow us to take advantage of a lot of the existing code.

go/cmd/dolt/commands/status.go
```diff
@@ -43,19 +43,24 @@ func Status(ctx context.Context, commandStr string, args []string, dEnv *env.Dol
...
-
-	workingInConflict, _, _, err := actions.GetTablesInConflict(ctx, dEnv)
+	workingTblsInConflict, _, _, err := actions.GetTablesInConflict(ctx, dEnv)
 
 	if err != nil {
 		panic(err) // fix
 	}
 
-	printStatus(dEnv, stagedDiffs, notStagedDiffs, workingInConflict)
+	notStagedDcDiffs, err := actions.GetDocDiffs(ctx, dEnv)
```
bheni: feels like shortening doc to dc isn't worth the loss in clarity of the code

go/cmd/dolt/commands/status.go
```diff
@@ -43,19 +43,24 @@ func Status(ctx context.Context, commandStr string, args []string, dEnv *env.Dol
...
+
+	if err != nil {
+		panic(err)
```
bheni: Now is probably a good time to remove these panics.

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -475,3 +477,53 @@ func (root *RootValue) RemoveTables(ctx context.Context, tables ...string) (*Roo
...
+		rowMap, err = docTable.GetRowData(ctx)
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		licenseVal, _, err = rowMap.MaybeGet(ctx, types.String(licensePk))
+		readmeVal, _, err = rowMap.MaybeGet(ctx, types.String(readmePk))
+	}
+
+	if licenseVal == nil && newerLicenseBytes != nil {
```
bheni: could convert this if / else if block into a function so that if you add a 3rd doc type you don't have to copy so much code.

go/libraries/doltcore/env/actions/diffs.go
```diff
@@ -104,3 +120,66 @@ func GetTableDiffs(ctx context.Context, dEnv *env.DoltEnv) (*TableDiffs, *TableD
 
 	return stagedDiffs, notStagedDiffs, nil
 }
+
+func NewDocDiffs(ctx context.Context, newerLicenseBytes, newReadmeBytes []byte, older *doltdb.RootValue) (*DocDiffs, error) {
```
bheni: might want to modify function signature so that you can add new doc types in the future without having to change the function signature for each new type of document.

go/libraries/doltcore/env/environment.go
```diff
@@ -155,6 +162,50 @@ func (dEnv *DoltEnv) HasLocalConfig() bool {
 	return ok
 }
 
+// GetLicenseFile returns the path to the LICENSE.md file if it exists
+func (dEnv *DoltEnv) GetLicenseFile() string {
```
bheni: might want to change to something like GetDoc which takes some parameter which tells it which doc.  Then if you add docs you don't have to duplicate all these functions again.

bats/1pk5col-ints.bats
```diff
@@ -453,6 +453,8 @@ if rows[2] != "9,8,7,6,5,4".split(","):
 }
 
 @test "generate a merge conflict and try to roll back using dolt merge --abort" {
+    rm "LICENSE.md" 
+    rm "README.md"
```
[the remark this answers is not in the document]
katiemcculloch: Added comments everywhere I'm rm'ing the files, for example: https://github.com/liquidata-inc/dolt/pull/222/files#diff-d179fd867dc7119d19f23f21e8bf2d64R456

As a side note, these comments are temporary and will be updated/removed after i do dolt add.

go/cmd/dolt/commands/status.go
katiemcculloch: Good call, just referring to them by docs/Docs now: https://github.com/liquidata-inc/dolt/pull/222/commits/160188eb46eee8bc1331f7b69453f8d3d2033286

go/cmd/dolt/commands/status.go
katiemcculloch: Got rid of panics: https://github.com/liquidata-inc/dolt/pull/222/commits/747a0a21ae8b387d9271c04d3046757b6e3c94ef

go/libraries/doltcore/doltdb/root_val.go
katiemcculloch: Moved this code to a helper function called `appendDocDiffs`: https://github.com/liquidata-inc/dolt/pull/222/files#diff-6f09c9d7a33cd9d8093aa6d760e03081R524

go/libraries/doltcore/env/actions/diffs.go
katiemcculloch: Changed `NewDocDiffs` to take in a slice of `DocDetails`: https://github.com/liquidata-inc/dolt/pull/222/files#diff-0696625847c6b30eef8d68a0543236a1R124

...so now `rootVal.DocDiff` also takes in a slice of DocDetails, and can process any number of docs: https://github.com/liquidata-inc/dolt/pull/222/files#diff-6f09c9d7a33cd9d8093aa6d760e03081R490

go/libraries/doltcore/env/environment.go
katiemcculloch: Done: https://github.com/liquidata-inc/dolt/pull/222/files#diff-bd6777a87b5edb64be7736c8af71bbc1R165

go/libraries/doltcore/env/actions/diffs.go
```diff
@@ -104,3 +120,66 @@ func GetTableDiffs(ctx context.Context, dEnv *env.DoltEnv) (*TableDiffs, *TableD
...
+
+	if err != nil {
+		return nil, err
+	}
+	readmeText, err := dEnv.GetLocalFileText(env.ReadmeFile)
+	if err != nil {
+		return nil, err
+	}
+
+	docDetails := make([]*doltdb.DocDetails, 2)
```
bheni: if you get rid of the make function you can define this as:

    docDetails := []*doltdb.DocDetails {
        &doltdb.DocDetails{licenseText, doltdb.LicensePk, nil},
        &doltdb.DocDetails{readmeText, doltdb.ReadmePk, nil}
    }

PR #229, opened 2019-12-03
fix schema diff primary key changes

coffeegoddd:
Output looks like this for changing a pk:
```
--- a/test @ 4uvb6bb3p7dqudnuidh9oh4ccsehik7n
+++ b/test @ 2tl4quv92ot0jg4v3ai204rld00trbo4
  CREATE TABLE test (
    `pk` BIGINT NOT NULL COMMENT 'tag:0'
-   `c1` BIGINT COMMENT 'tag:1'
    `c2` BIGINT COMMENT 'tag:2'
    `c3` BIGINT COMMENT 'tag:3'
    `c4` BIGINT COMMENT 'tag:4'
    `c5` BIGINT COMMENT 'tag:5'
<    PRIMARY KEY (`pk`, `c1`)
>    PRIMARY KEY (`pk`)
  );
```

Also add the pk contstraint so it shows when it is not changing:
```
--- a/test @ idfqe6c5s2i9ohihkk4r4tj70tf3l8c7
+++ b/test @ 2tl4quv92ot0jg4v3ai204rld00trbo4
  CREATE TABLE test (
    `pk` BIGINT NOT NULL COMMENT 'tag:0'
    `c1` BIGINT COMMENT 'tag:1'
    `c2` BIGINT COMMENT 'tag:2'
<          `c3` BIGINT COMMENT 'tag:3'
>   `newColName3` BIGINT COMMENT 'tag:3'
    `c4` BIGINT COMMENT 'tag:4'
    `c5` BIGINT COMMENT 'tag:5'
     PRIMARY KEY (`pk`, `c1`)
  );
```


bats/schema-changes.bats
```diff
@@ -111,8 +112,9 @@ teardown() {
     dolt table create -f -s=`batshelper 1pk5col-ints-add-pk.schema` test
     run dolt diff --schema
     [ "$status" -eq 0 ]
-    skip "Schema diff output does not handle adding primary keys. Additionally, rows that should not be in the diff show up."
     [[ "$output" =~ "PRIMARY KEY" ]] || false
```
zachmu: Can remove this test now that you're looking for more specific output below

bats/schema-changes.bats
```diff
@@ -122,7 +124,41 @@ teardown() {
...
-}
\ No newline at end of file
+}
+
+@test "schema diff should show primary keys in output" {
+    dolt table create -s=`batshelper 1pk5col-ints.schema` test
+    dolt add test
+    dolt commit -m "committed table so we can see diffs"
+    dolt table create -f -s=`batshelper 1pk5col-ints-change-col-name.schema` test
+    echo dolt diff --schema
```
zachmu: rm this line

go/cmd/dolt/commands/diff.go
```diff
@@ -492,6 +525,10 @@ func sqlSchemaDiff(tableName string, tags []uint64, diffs map[uint64]diff.Schema
 	}
 }
 
+func formatColName(name string) string {
```
zachmu: sql.QuoteIdentifier already does this

go/cmd/dolt/commands/diff.go
```diff
@@ -421,15 +422,28 @@ func diffSchemas(tableName string, sch1 schema.Schema, sch2 schema.Schema, dArgs
 func tabularSchemaDiff(tableName string, tags []uint64, diffs map[uint64]diff.SchemaDifference) errhand.VerboseError {
```
zachmu: What you have here is fine, but it substantially duplicates the logic in sql.SchemaAsCreateStmt (before your change as well). I think this is fine to check in as is, but a better strategy would probably be to use a textual diff library, and just output the diff in sql.SchemaAsCreateStmt between the old schema and the new one.

coffeegoddd: not a lot of diff libraries for go that I found... this one (https://github.com/sergi/go-diff) is ok, but doesn't support line by line diffing afaik. Think it's probably better to leave this as is for now, then revisit updating this code to be less duplicative when I have a bit more time to focus on it.

PR #233, opened 2019-12-04
Zachmu/sql batch

zachmu:
Killed off original sql batch inserter and implemented equivalent functionality for new engine.


go/libraries/doltcore/sqle/database.go
```diff
@@ -94,7 +120,8 @@ func (db *Database) GetTableInsensitive(ctx context.Context, tblName string) (sq
 		return nil, false, err
 	}
 
-	return &DoltTable{name: tblName, table: tbl, sch: sch, db: db}, true, nil
+	db.tables[exactName] = &DoltTable{name: exactName, table: tbl, sch: sch, db: db}
```
Hydrocharged: Probably a micro-optimization, but since maps have bitten me in the past, why not make a new variable, assign to it the new `DoltTable`, set the map value, and return the variable? Instead of retrieving from the map again?

go/libraries/doltcore/sqle/sqlbatch_test.go
```diff
@@ -136,34 +137,42 @@ func TestSqlBatchInserts(t *testing.T) {
 	assertRowSetsEqual(t, expectedAppearances, allAppearanceRows)
 }
 
+func drainIter(iter sql.RowIter) error {
+	var returnedErr error
+	for {
+		_, err := iter.Next()
+		if err == io.EOF {
+			return returnedErr
+		} else if returnedErr != nil {
```
Hydrocharged: Should this be `else if returnedErr == nil`? It starts as `nil` at first, so we'd never assign to it.

go/libraries/doltcore/sqle/sqletestutil/executesql.go
```diff
@@ -136,3 +138,15 @@ func ExecuteSelect(root *doltdb.RootValue, query string) ([]sql.Row, error) {
...
 	return rows, nil
 }
+
+func drainIter(iter sql.RowIter) error {
+	var returnedErr error
+	for {
+		_, err := iter.Next()
+		if err == io.EOF {
+			return returnedErr
+		} else if returnedErr != nil {
```
Hydrocharged: Same as above comment.

go/libraries/doltcore/sqle/sqlbatch_test.go
```diff
@@ -188,29 +197,28 @@ func TestSqlBatchInsertIgnoreReplace(t *testing.T) {
...
+
+	_, rowIter, err := engine.Query(sql.NewEmptyContext(), `insert into people (id, first, last, is_married, age, rating, uuid, num_episodes) values
+					(0, "Maggie", "Simpson", false, 1, 5.1, '00000000-0000-0000-0000-000000000007', 677)`)
+	// This won't generate an error until we commit the batch (duplicate key)
+	assert.NoError(t, err)
+	assert.NoError(t, drainIter(rowIter))
+
+	_, _, err = engine.Query(sql.NewEmptyContext(), `insert into people values
+					(2, "Milhouse", "VanHouten", false, 1, 5.1, true, 677)`)
+	assert.Error(t, err)
```
Hydrocharged: The above does not error but this does before the flush? Probably something I'm overlooking...

go/libraries/doltcore/sqle/database.go
zachmu: Good suggestion, done.

go/libraries/doltcore/sqle/sqlbatch_test.go
zachmu: Good catch.

go/libraries/doltcore/sqle/sqlbatch_test.go
zachmu: Added a clarifying comment

PR #235, opened 2019-12-06
Fix typo in README for table import

namdnguyen:
I ran into this typo while using Dolt yesterday. The command keywords were in the incorrect order in the README.


bheni:
Thank you!!!

namdnguyen:
You're welcome! I'm loving how familiar it immediately feels because of the similarly with git.

PR #239, opened 2019-12-09
bats/create-views.bats: Pick up go-mysql-server support for views.

reltuk:


bats/create-views.bats
```diff
@@ -0,0 +1,101 @@
...
+    run dolt sql <<< '
+create view broken as select id from my_users'
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "table not found: my_users" ]] || false
+}
+
+@test "can create view referencing table" {
+    run dolt sql <<< '
+create table my_users (id int primary key);
+create view will_be_broken as select id from my_users'
```
zachmu: shouldn't be called will_be_broken

PR #245, opened 2019-12-12
change super_schema to use a CommitItr

bheni:


go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"errors"
+	"github.com/liquidata-inc/dolt/go/store/hash"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+// CommitItr is an interface for iterating over a set of unique commits
+type CommitItr interface {
+	// Next returns the hash of the next commit, and a pointer to that commit.  Implementations of Next must handle
+	// making sure the list of commits returned are unique.  When complete Next will return hash.Hash{}, nil, nil
```
zachmu: Returning io.EOF is more idiomatic for iterators in go, right?

go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+	ddb         *DoltDB
+	rootCommits []*Commit
+	currentRoot int
+
+	added       map[hash.Hash]bool
+	unprocessed []hash.Hash
+	curr        *Commit
+}
+
+// CommitItrForAllBranches returns a CommitItr which will iterate over all commits in all branches in a DoltDB
```
zachmu: Needs more detail about the order that results are returned in. Looks like this is a depth-first ancestor traversal for each top-level ref?

go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		rootCommits = append(rootCommits, cm)
+	}
+
+	cmItr := CommitItrForRoots(ddb, rootCommits...)
+	return cmItr, nil
+}
+
+// CommitItrForRoots will return a
```
zachmu: Suspense is killing me

go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+func CommitItrForRoots(ddb *DoltDB, rootCommits ...*Commit) CommitItr {
+	return &commitItr{
+		ddb:         ddb,
+		rootCommits: rootCommits,
+		added:       make(map[hash.Hash]bool, 4096),
+		unprocessed: make([]hash.Hash, 0, 4096),
+	}
+}
+
+// Next returns the hash of the next commit, and a pointer to that commit.  It handles making sure the list of commits
```
zachmu: Probably can omit the comment here, or just refer to the interface method.

go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+// returned are unique.  When complete Next will return hash.Hash{}, nil, nil
+func (cmItr *commitItr) Next(ctx context.Context) (hash.Hash, *Commit, error) {
+	for cmItr.curr == nil {
+		if cmItr.currentRoot >= len(cmItr.rootCommits) {
+			return hash.Hash{}, nil, nil
```
zachmu: Same comment about io.EOF

go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -0,0 +1,135 @@
...
+
+	numUnprocessed := len(cmItr.unprocessed)
+
+	if numUnprocessed == 0 {
+		cmItr.curr = nil
+		cmItr.currentRoot++
+		return cmItr.Next(ctx)
+	}
+
+	next := cmItr.unprocessed[numUnprocessed-1]
```
zachmu: Any reason to pop off the end of the list like this, instead of FIFO?

bheni: 1. performance.
2. maintain depth first ordering.

go/libraries/doltcore/doltdb/commit_itr.go
bheni: done

go/libraries/doltcore/doltdb/commit_itr.go
bheni: I don't think I want to guarantee an traversal order.  Users should just know they are going to get all unique commits.

go/libraries/doltcore/doltdb/commit_itr.go
bheni: updated

go/libraries/doltcore/doltdb/commit_itr.go
bheni: I think the generated go docs prefer this be here.

go/libraries/doltcore/doltdb/commit_itr.go
bheni: updated

PR #249, opened 2019-12-13
go-mysql-server types

Hydrocharged:


go/libraries/doltcore/sqle/types/types.go
```diff
@@ -96,11 +94,44 @@ func init() {
...
+		sqltypes.Float32: dtypes.FloatKind,
+		sqltypes.Float64: dtypes.FloatKind,
+		sqltypes.Geometry: dtypes.InlineBlobKind,
+		sqltypes.Int16: dtypes.IntKind,
+		sqltypes.Int24: dtypes.IntKind,
+		sqltypes.Int32: dtypes.IntKind,
+		sqltypes.Int64: dtypes.IntKind,
+		sqltypes.Int8: dtypes.IntKind,
+		sqltypes.Null: dtypes.NullKind,
+		sqltypes.Set: dtypes.StringKind,
```
zachmu: Probably should be unsupported

go/libraries/doltcore/sqle/types/types.go
```diff
@@ -96,11 +94,44 @@ func init() {
...
+		sqltypes.Text: dtypes.StringKind,
+		sqltypes.Time: dtypes.InlineBlobKind,
+		sqltypes.Timestamp: dtypes.TimestampKind,
+		sqltypes.TypeJSON: dtypes.InlineBlobKind,
```
zachmu: Should be unsupported

PR #254, opened 2019-12-16
Fix diff where clause

bheni:
Fixes diff where so that you no longer need to use to_col or from_col.

    --where col=val

is evaluated as:

    where to_col=val || from_col=val


timsehn:
<3

timsehn:
BATS test?

timsehn:
Maybe add a from_COLUMN_NAME to_COLUMN_NAME test?

timsehn:
I think this one is good to merge.

PR #256, opened 2019-12-16
Andy/sqlweb

andrew-wm-arthur:
I needed to reuse logic from SQL processing in Dolt, but as written it needed a CLI environment. Did some refactoring to de-couple env.DoltEnv and env.RepoState:
 - Removed env.DoltEnv as a dependency for sqlEngine in commands/sql
 - Removed FileSystem as a dependency in env.RepoState

👀 that symmetric stat line tho


zachmu:
Brian should take a look at this. My first instinct is that this kind of breaks the usefulness of the doltenv abstraction to some extent, maybe Brian has some thoughts. The SQL stuff looks fine.

go/cmd/dolt/commands/sql.go
```diff
@@ -441,7 +441,7 @@ func processQuery(ctx context.Context, query string, se *sqlEngine) error {
 	case *sqlparser.Select, *sqlparser.OtherRead, *sqlparser.Insert, *sqlparser.Update, *sqlparser.Show:
 		sqlSch, rowIter, err := se.query(ctx, query)
 		if err == nil {
-			err = prettyPrintResults(ctx, se.db.Root().VRW().Format(), sqlSch, rowIter)
+			err = prettyPrintResults(ctx, se.sdb.Root().VRW().Format(), sqlSch, rowIter)
```
bheni: rather than getting the VRW from the root se.ddb.ValueReadWriter().Format() seems better, but better still would be to add a Format method to DoltDB and call that directly here.

PR #258, opened 2019-12-16
First pass at bats tests for schema import. 

timsehn:
Modified a helper file that now causes a panic on a couple table import tests as well as a new schema import test. We should fix ASAP.


timsehn:
Added @katiemcculloch because a I broke a couple of here tests. The bug our user found that I am testing for may be caused by her changes.

bheni:
This has been merged into the PR with the fix: https://github.com/liquidata-inc/dolt/pull/259

PR #259, opened 2019-12-16
fix csv parsing

bheni:
Currently there is a panic if a csv line contains nothing but whitespace.


timsehn:
Presuming the bats tests pass this LGTM.

PR #262, opened 2019-12-17
Include required length value for "varchar" column type in README tutorial

jaredtkatz:


jaredtkatz:
The longest state name is 14 characters, hence the choice of `14` for the varchar length value.

reltuk:
Thanks for the contribution!

zachmu:
Thanks for the contribution, Jared, good catch!

We don't currently store or enforce the length of varchar columns, but @Hydrocharged is hard at work to change that. For the time being, any VARCHAR columns will be considered interchangeable with LONGTEXT by dolt.

PR #266, opened 2019-12-18
fix and unit test

andrew-wm-arthur:


andrew-wm-arthur:
I made sure that it failed before the fix

go/cmd/dolt/commands/sql_test.go
```diff
@@ -40,6 +42,23 @@ import (
...
 
+// Smoke test: Console opens and exits
+func TestSqlConsole(t *testing.T) {
+	t.Run("SQL console opens and exits", func(t *testing.T) {
+		dEnv := createEnvWithSeedData(t)
+		args := []string{}
+		commandStr := "dolt sql"
+
+		go func() {
+			time.Sleep(1 * time.Second)
```
zachmu: This is going to fail for some stupid reason on Jenkins eventually when it takes longer than a second. You might try adding a callback parameter to the Sql command that fires once it's ready to begin receiving input. It's gross, but also fault tolerant. Worth seeing how it looks anyway.

andrew-wm-arthur: For some reason the test doesn't actually need to read `exit` from STDIN in order the close the console. The console just closes without waiting for input. I removed the coroutine altogether

PR #268, opened 2019-12-19
added args check to dolt sql command

andrew-wm-arthur:
image

### DIFF
--- a/bats/sql-shell.bats
+++ b/bats/sql-shell.bats
@@ -24,3 +24,8 @@ teardown() {
     [ $status -eq 1 ]
     [[ ! "$output" =~ "666" ]] || false
 }
+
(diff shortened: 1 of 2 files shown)


bats/sql-shell.bats
```diff
@@ -23,4 +23,9 @@ teardown() {
     run dolt sql <<< "select * from test; This is bad sql; insert into test (pk) values (666); select * from test;"
     [ $status -eq 1 ]
     [[ ! "$output" =~ "666" ]] || false
+}
+
+@test "inline query with missing -q flag should error" {
+    run dolt sql "SELECT * FROM test;"
+    [ $status -eq 1 ]
```
timsehn: Maybe add a quick regex for some substring of the error message?

PR #269, opened 2019-12-19
go/store/util/verbose: Add ability to override Log function.

reltuk:


reltuk:
@bheni Ok, just adopted `zap.Logger` natively. PTAL.

go/store/util/verbose/verbose.go
```diff
@@ -36,35 +39,30 @@ var (
...
+			return zapcore.DebugLevel.Enabled(l)
 		} else {
-			log.Println(format)
+			return zapcore.WarnLevel.Enabled(l)
 		}
+	})
+	encoder := zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig())
+	core := zapcore.NewCore(encoder, zapcore.Lock(os.Stderr), enabler)
+	Logger = func(ctx context.Context) *zap.Logger {
+		return zap.New(core)
```
bheni: New logger for each call for the default dolt client?

bheni: Should we set the default log level to error for dolt?

reltuk: Warn seems fine...there is currently no logging though.

PR #279, opened 2019-12-31
{go,bats}: Remove DocTableName from dolt table, schema, ls, add, reset, diff

katiemcculloch:
This PR removes DocTableName from the outstanding commands so we don't expose the dolt docs table.


katiemcculloch:
Please hold off review on this for a few hours! I'm have another commit to push that will include dolt schema commands and a few other dolt commands.

PR #280, opened 2020-01-02
{bats, go/libraries/doltcore/sqle/database.go}: Remove DoltNamespace from `dolt sql` command

katiemcculloch:


go/libraries/doltcore/sqle/database.go
```diff
@@ -142,7 +142,17 @@ func (db *Database) GetTableInsensitive(ctx context.Context, tblName string) (sq
...
 
 func (db *Database) GetTableNames(ctx context.Context) ([]string, error) {
-	return db.root.GetTableNames(ctx)
+	tblNames, err := db.root.GetTableNames(ctx)
+	if err != nil {
+		return nil, err
+	}
+	for i, tbl := range tblNames {
+		if tbl == doltdb.DocTableName {
+			tblNames = append(tblNames[:i], tblNames[i+1:]...)
```
zachmu: Instead of setting and breaking here, just return the result of append.

Also does this work when dolt_docs is the last element of the list?

You might want to factor out a method called filterDoltInternalTables or similar, that takes the table names and returns a copy with the _dolt* tables filtered out.

go/libraries/doltcore/sqle/database.go
```diff
@@ -183,7 +193,7 @@ func (db *Database) DropTable(ctx *sql.Context, tableName string) error {
 // CreateTable creates a table with the name and schema given.
 func (db *Database) CreateTable(ctx *sql.Context, tableName string, schema sql.Schema) error {
 
-	if !doltdb.IsValidTableName(tableName) {
+	if !doltdb.IsValidTableName(tableName) || tableName == doltdb.DocTableName {
```
zachmu: I would just check strings.HasPrefix(tableName, "dolt_"). We want to reserve that entire namespace. Also make "dolt_" a constant somewhere in this file.

go/libraries/doltcore/sqle/database.go
katiemcculloch: Resolved: https://github.com/liquidata-inc/dolt/pull/280/commits/fc2e09aad5f6b2267c341262635a0aa48b76b2cf#diff-ff4ec6598b401e530eb84ad72ea62a9fR154

go/libraries/doltcore/sqle/database.go
katiemcculloch: Resolved: https://github.com/liquidata-inc/dolt/pull/280/commits/fc2e09aad5f6b2267c341262635a0aa48b76b2cf#diff-ff4ec6598b401e530eb84ad72ea62a9fR165

PR #281, opened 2020-01-03
fix select on system table that doesn't exist

bheni:
fix select on a system table that has a valid prefix but whose suffix does not match a valid table.

What makes this a little bit tough is that you can query diffs or the history of a table that no longer exists. So need to process the entire history and then see if at any time there was a schema'd table with the given name.


andrew-wm-arthur:
LGTM

PR #284, opened 2020-01-03
{go,bats}: Implement dolt diff <dic> by parsing docs from args, with …

katiemcculloch:
…bats test


katiemcculloch:
This PR is just for diffing one doc. Merging this into `km/docs-dolt-diff` to have all doc diff code reviewed in one PR.

PR #286, opened 2020-01-07
Dolt Types

Hydrocharged:


go/libraries/doltcore/schema/typeinfo/float32.go
```diff
@@ -0,0 +1,119 @@
...
+	return floatConvertValueToNomsValue(ti, v)
+}
+
+// Equals implements TypeInfo interface.
+func (ti *float32Impl) Equals(other TypeInfo) bool {
+	if other == nil {
+		return false
+	}
+	if ti2, ok := other.(*float32Impl); ok {
+		return *ti == *ti2
```
zachmu: This is always true (empty structs are always == to each other. Seems like you just want to do a type assertion here.

zachmu: Same comment applies to all of these types

go/libraries/doltcore/schema/typeinfo/null.go
```diff
@@ -0,0 +1,81 @@
...
+
+package typeinfo
+
+import (
+	"github.com/src-d/go-mysql-server/sql"
+
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+type nullImpl struct{}
```
zachmu: I don't think null is a type. I know there is a null type in noms, but I think it's kind of nonsensical in most applications.

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,243 @@
...
+	default:
+		return nil, fmt.Errorf(`no type info can be created from SQL base type "%v"`, sqlType.String())
+	}
+}
+
+// TypeInfoFromIdentifierParams constructs a TypeInfo from the given identifier and parameters.
+func TypeInfoFromIdentifierParams(id Identifier, params map[string]string) (TypeInfo, error) {
+	switch id {
+	case UnknownType:
+		return &unknownImpl{}, nil
```
zachmu: Probably should be an error (same as default case)

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,243 @@
...
+		return &float64Impl{}, nil
+	case types.StringKind:
+		return &unknownImpl{}, nil
+	case types.UUIDKind:
+		return &unknownImpl{}, nil
+	case types.IntKind:
+		return &int64Impl{}, nil
+	case types.UintKind:
+		return &uint64Impl{}, nil
+	case types.NullKind:
```
zachmu: Not a supported type in dolt

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,243 @@
...
+	// GetTypeParams returns a map[string]string containing the type parameters.  This is used for
+	// serialization and deserialization of type information.
+	GetTypeParams() map[string]string
+
+	// IsValid takes in a value (not a types.Value) and returns whether the value is valid for this
+	// type.
+	IsValid(v interface{}) bool
+
+	// ToSqlType returns the TypeInfo as a sql.Type.
+	ToSqlType() (sql.Type, error)
```
zachmu: What's the error case? Unsupported?

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -0,0 +1,243 @@
...
+	Uint32Type:  struct{}{},
+	Uint64Type:  struct{}{},
+	Float32Type: struct{}{},
+	Float64Type: struct{}{},
+}
+
+// TypeInfo is an interface used for encoding type information.
+type TypeInfo interface {
+	// AppliesToKind takes in a NomsKind and returns an error regarding whether this type applies to the value.
+	AppliesToKind(kind types.NomsKind) error
```
zachmu: This should return a bool like the IsValid method.

PR #288, opened 2020-01-07
add diff_type column to be able to select where diff_type is added, r…

bheni:
…emoved, or modified


go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -237,7 +252,21 @@ func (itr *diffRowItr) Next() (sql.Row, error) {
...
 
-	return doltRowToSqlRow(r, itr.sch)
+	sqlRow, err := doltRowToSqlRow(r, itr.sch)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if hasTo && hasFrom {
+		sqlRow = append(sqlRow, "modified")
```
zachmu: You should pull out constants for these strings, maybe put them in the same package as the differ / joiner stuff.

bheni: Moved to constants, but left in file.  I'm not sure I'd say this is how all diffs should label these fields in all cases.  It's just how I want them labeled for the diff_tables.

PR #290, opened 2020-01-08
go/go.mod: Run go get -u all. Migrate to dbr/v2.

reltuk:


zachmu:
We only use DBR for testing, should be fine

PR #293, opened 2020-01-09
go/cmd/dolt/commands/credcmds: Add documentation and a little bit of chrome to dolt creds commands.

reltuk:


go/libraries/doltcore/env/environment.go
```diff
@@ -428,16 +428,25 @@ func (dEnv *DoltEnv) getRPCCreds() (credentials.PerRPCCredentials, error) {
...
-		return dCreds, nil
+func (dEnv *DoltEnv) getRPCCreds() (credentials.PerRPCCredentials, error) {
+	dCreds, valid, err := dEnv.UserRPCCreds()
+
+	if err != nil {
+		return nil, ErrInvalidCredsFile
 	}
 
-	return nil, nil
+	if !valid {
```
bheni: If it's not valid, shouldn't it be an ErrInvalidCredsFile?

reltuk: Good call. Fixed!

reltuk: No, this was right. The `!valid` case is really the empty / missing case.

PR #294, opened 2020-01-09
Added indexes to dolt sqllogictest harness and updated dependency on …

zachmu:
…go-mysql-server.


go/libraries/doltcore/sqle/logictest/dolt/doltharness.go
```diff
@@ -198,6 +201,10 @@ func toSqlString(val interface{}) string {
 	case float32, float64:
 		// exactly 3 decimal points for floats
 		return fmt.Sprintf("%.3f", v)
+	case decimal.Decimal:
+		// exactly 3 decimal points for floats
+		f, _ := v.Float64()
+		return fmt.Sprintf("%.3f", f)
```
Hydrocharged: Rather than casting to a `Float64`, why not use the string functions included in `decimal`?

zachmu: Good idea

PR #298, opened 2020-01-11
go/cmd/dolt/commands/sql: Add view persistence into dolt database.

reltuk:


go/libraries/doltcore/sqle/database.go
```diff
@@ -246,3 +249,115 @@ func (db *Database) Flush(ctx context.Context) error {
 	}
 	return nil
 }
+
+// Implements sql.ViewCreator. Persists the view in the dolt database, so
+// it can exist in a sql session later. Reviews sql.ErrExistingView a view
```
zachmu: s/reviews/returns

go/libraries/doltcore/sqle/database.go
```diff
@@ -246,3 +249,115 @@ func (db *Database) Flush(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	iter, err := newRowIterator(tbl, ctx)
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+
+	var parseerrors []error
```
zachmu: broDoYouEvenCamelCase

go/libraries/doltcore/sqle/schema_table.go
```diff
@@ -0,0 +1,83 @@
...
+	keyVl := doltLookup.NomsMapKey(tbl.sch)
+	key, err := keyVl.Value(ctx)
+	if err != nil {
+		return false, err
+	}
+	rows, err := tbl.table.GetRowData(ctx)
+	if err != nil {
+		return false, err
+	}
+	f, err := rows.Has(ctx, key)
```
zachmu: Probably can just return rows.Has?

PR #305, opened 2020-01-14
bats/creds.bats: Some initial bats tests for dolt creds new, ls and rm.

reltuk:


bats/creds.bats
```diff
@@ -0,0 +1,51 @@
...
+}
+
+@test "ls empty creds" {
+    run dolt creds ls
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 0 ]
+}
+
+@test "ls new cred" {
+    run dolt creds new
```
timsehn: If you just check $status -eq 0 you don't need run. Just remove.

reltuk: Thanks. Fixed!

timsehn:
LGTM

PR #307, opened 2020-01-14
go/utils/publishrelease: First pass at an install.sh

reltuk:


reltuk:
Gonna land this for now. Open to iteration here.

PR #316, opened 2020-01-21
Buffered Sequence Iterator

andrew-wm-arthur:
- Created a new interface `sequenceIterator` for the use case when `sequenceCurosor` is simply accessing elements in its sequence (ie `MapIterator`, `SetIterator`, and `ListIterator`)
- Created a new buffered implementation of `sequenceIterator` designed by @reltuk to batch chunk fetching from the `ValueStore`. In use cases such as DoltHub where chunk fetching IO is slow, this will dramatically accelerate performance.


go/store/types/sequence_cursor.go
```diff
@@ -57,7 +64,7 @@ func (cur *sequenceCursor) getItem(idx int) (sequenceItem, error) {
 	return cur.seq.getItem(idx)
 }
 
-// sync loads the sequence that the cursor index points to.
+// syncAdvance loads the sequence that the cursor index points to.
```
reltuk: Fix this...

reltuk:
Generally looks good. One nit about a comment, then I think it's OK to land.

PR #333, opened 2020-01-23
fix regex

andrew-wm-arthur:


bats/no-repo.bats
```diff
@@ -47,8 +47,8 @@ teardown() {
 }
 
 @test "testing dolt version output" {
-    run dolt version
-    [ "$status" -eq 0 ]
+    dolt version > version.txt
+    run grep -E "dolt version \d+.\d+.\d+" version.txt
     [[ "$output" =~ "dolt version " ]] || false
```
timsehn: I think this output is not needed and remove run

timsehn: ie. if grep fails, the bats test will fail

timsehn: and we should probably delete version.txt.

PR #337, opened 2020-01-27
bats,{go/cmd/dolt/commands}: Hide doc schema for schema show w/no table args

katiemcculloch:


bats/docs.bats
```diff
@@ -618,6 +618,15 @@ SQL
     run dolt schema show dolt_docs
     [ "$status" -eq 1 ]
     [[ "$output" =~ "'dolt_docs' is not a valid table name" ]] || false
+    run dolt schema show
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "usage: dolt schema show" ]] || false
```
timsehn: Don't think this regex is necessary. When there are dolt_docs and no other table dolt schema show should error.

bats/docs.bats
```diff
@@ -618,6 +618,15 @@ SQL
     run dolt schema show dolt_docs
     [ "$status" -eq 1 ]
     [[ "$output" =~ "'dolt_docs' is not a valid table name" ]] || false
+    run dolt schema show
+    [ "$status" -eq 1 ]
```
zachmu: This doesn't make sense to me. Why should this command fail?

bats/docs.bats
timsehn: No need to test the help message.

bats/docs.bats
timsehn: Basically if only dolt_docs in the schema, error as if there is no schema...right Katie?

katiemcculloch: @zachmu `dolt schema show` should fail and return the print the command usage if there are no tables on the working root (this was also the logic before I started tweaking it for docs).

We don't want to expose `dolt_docs`, and if it is the only table on the working root, we want the behavior to be as if there are no tables on the working root.

bats/docs.bats
katiemcculloch: thanks, removed the message assertion

bats/docs.bats
zachmu: I think this is actually a bug. No tables should not be an error. An empty schema is a valid repo state. It should behave the same as dolt ls when there are no tables: print "No tables in working set" and exit 0.

Everything else you have here looks fine. You can address the no tables bug in a separate PR if you want.

katiemcculloch: Just pushed a commit that includes these updates: https://github.com/liquidata-inc/dolt/pull/337/files#diff-8dfc32e83b91e8c830d4d8bd789de412R132

PR #342, opened 2020-01-28
Zachmu/fix clone empty

### LINKED ISSUE #217: cloning an empty repository results in panic
After creating an empty repository on `dolthub`, attempting to clone it results in this panic:
`panic: nil commit spec`.

Here is my config:
```
metrics.host = localhost
remotes.default_host = <dev url>
user.creds = <creds>
user.email = <my email>
user.name = <my username>
```

Here is the command I ran:
`dolt clone https://doltremoteapi.dolthub.com/coffeegoddd/crashes`

zachmu:
Cloning an empty repo now works as expected. Fixes #217

### DIFF
--- a/go/cmd/dolt/commands/clone.go
+++ b/go/cmd/dolt/commands/clone.go
@@ -301,10 +301,31 @@ func cloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
 		for _, brnch := range branches {
 			branch = brnch.GetPath()
 
-			if branch == "master" {
+			if branch == doltdb.MasterBranch {
 				break
 			}
 		}
+	}
+
+	// If we couldn't find a branch but the repo cloned successfully, it's empty. Initialize it instead of pulling from
+	// the remote.
+	if branch == "" {
+		name := dEnv.Config.GetStringOrDefault(env.UserNameKey, "")
+		email := dEnv.Config.GetStringOrDefault(env.UserEmailKey, "")
+
+		if *name == "" {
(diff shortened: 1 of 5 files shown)


go/cmd/dolt/commands/clone.go
```diff
@@ -301,58 +301,80 @@ func cloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
...
+
+		err = dEnv.InitDBWithTime(ctx, types.Format_Default, *name, *email, doltdb.CommitNowFunc())
+		if err != nil {
+			return errhand.BuildDError("error: could not initialize repository").AddCause(err).Build()
+		}
+
+		branch = doltdb.MasterBranch
+		performPull = false
+	}
+
```
bheni: Maybe put this all in a function initIfEmpty or something like that with your comment at the top.

go/cmd/dolt/dolt.go
```diff
@@ -136,7 +136,7 @@ func runMain() int {
 	restoreIO := cli.InitIO()
 	defer restoreIO()
 
-	warnIfMaxFilesTooLow()
```
bheni: Why are we commenting this out?

zachmu: For some reason this fail in GoLand for me so I comment it out locally. Reverted.

go/cmd/dolt/commands/clone.go
zachmu: Done.

PR #347, opened 2020-01-29
go/cmd/dolt: commands/credcmds/new: Use a newly created credential if there is no existing selected credential.

reltuk:


go/cmd/dolt/commands/credcmds/new.go
```diff
@@ -74,7 +76,32 @@ func (cmd NewCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
+	} else {
+		return 0
+	}
+}
+
+func updateConfigToUseNewCredIfNoExistingCred(dEnv *env.DoltEnv, dCreds creds.DoltCreds) error {
+	gcfg, hasGCfg := dEnv.Config.GetConfig(env.GlobalConfig)
+
+	if !hasGCfg {
+		panic("global config not found.  Should create it here if this is a thing.")
```
zachmu: if this is a thing?

bats/creds.bats
```diff
@@ -19,17 +19,22 @@ teardown() {
...
     [ "${#lines[@]}" -eq 1 ]
-    [[ "${lines[0]}" =~ (^\ \ ) ]] || false
+    [[ "${lines[0]}" =~ (^\*\ ) ]] || false
+    cred=`echo "${lines[0]}" | awk '{print $2}'`
+    dolt creds new
+    run dolt creds ls
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 2 ]
+    # Initially chosen credentials is still the chosen one.
+    [[ "`echo "$output"|grep "$cred"`" =~ (^\*\ ) ]] || false
```
zachmu: Is this sufficient to exercise the new logic? Shouldn't we check the global config for the new key?

go/cmd/dolt/commands/credcmds/new.go
reltuk: No idea...that was copied from `dolt creds new`

bats/creds.bats
reltuk: The assert on line 23 is the thing that would have failed previously...there would have been no `*` at the start of the line for the newly created credential, because it wasn't selected. This assert is making certain that `dolt creds new` after one cred is selected doesn't change the selected one...it's sufficient for that AFAICT...it looks for the line in `dolt ls` that has that cred and makes sure it still starts with `*`. I don't think we should go looking at implementation details of the config storage, instead of testing through the dolt interfaces.

PR #350, opened 2020-01-30
Bh/schema import update

bheni:


go/libraries/doltcore/schema/constraint.go
```diff
@@ -78,6 +78,18 @@ func (nnc NotNullConstraint) String() string {
 	return "Not null"
 }
 
+// ConstraintOfTypeIndex returns the index in the supplied slice of the first constraint of matching type.  If none are
+// found then -1 is returned
+func ConstraintOfTypeIndex(constraints []ColConstraint, constraintType string) int {
```
zachmu: IndexOfConstraint?

Issue #275, opened 2019-12-28
Weird error message when using dolt import


oscarbatori:
```
[oscarbatori:scratch] >> dolt table import people --pk=first -c test_data.csv
Data already exists.  Use -f to overwrite.
```
Should say "Table already exists".

timsehn:
I think the export table path was changed by the PR above not the import path.

timsehn:
@andrew-wm-arthur is gonna fix and make a bats test.

timsehn:
So it turns out that error code was used on both the import and export path. So, when we merged the PR one of our integration tests failed. @andrew-wm-arthur is going to make two different paths, one for export and one for import.

PR #363, opened 2020-02-01
Fix error message when table doesn't exist

benraskin92:
Fixes https://github.com/liquidata-inc/dolt/issues/275.


timsehn:
Thanks for the PR. We'll look at this today :-)

PR #386, opened 2020-02-11
delete cell values on drop column

andrew-wm-arthur:


andrew-wm-arthur:
The primary objective here is to prevent cells that correspond to a tag/column that is not referenced in any schema. The scenario is: add column to working set, add values to column, drop column from working set. 

This implementation is more heavy-handed in that it always deletes values when we drop a column. If we want to be more careful will storage, we can only perform this pruning when we're deleting a column from the working set. 
@zachmu @bheni @reltuk

go/libraries/doltcore/schema/alterschema/dropcolumn.go
```diff
@@ -71,15 +76,60 @@ func DropColumn(ctx context.Context, tbl *doltdb.Table, colName string) (*doltdb
...
+	mi, err := rowData.BufferedIterator(ctx)
+
+	if err != nil {
+		return types.EmptyMap, err
+	}
+
+	for {
+		k, v, err := mi.Next(ctx)
+
+		if k == nil || v == nil {
```
zachmu: Doesn't this iterator return err == io.EOF when it has no more elements? More common in go to match this error than check a nil value returned by the iterator.

go/libraries/doltcore/schema/alterschema/dropcolumn_test.go
```diff
@@ -27,6 +27,46 @@ import (
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
+var TypedRowsSansAge []row.Row
```
zachmu: s/sans/without/g

Get out of here with this francophile nonsense USA! USA!

andrew-wm-arthur: https://lh5.googleusercontent.com/proxy/UAFpmthJ6_r8i644McoXNJ7_kq1hY-q4IdTiHoSp3P2WjvBtP6ZsmA2dhBy7QQO_Jl4YZMteGF1Bf_EpdmBZlGddsCpMAGyeiYJG4eEIxw

go/libraries/doltcore/schema/alterschema/dropcolumn.go
andrew-wm-arthur: it doesn't return io.EOF the types package just gives `nil`

PR #390, opened 2020-02-12
bats/sql.bats: Document some unsupported SQL features with some failing bats tests.

reltuk:


bats/sql.bats
```diff
@@ -345,3 +345,42 @@ teardown() {
...
 }
+
+@test "sql unix_timestamp function" {
+    skip "unix_timestamp is not supported"
+    run dolt sql -q "SELECT UNIX_TIMESTAMP(NOW()) FROM dual;"
+    [ $status -eq 0 ]
+    [ "${#lines[@]}" -eq 5 ]
+}
+
+@test "sql select union all" {
```
zachmu: Probably should cover union distinct while you're in here

PR #395, opened 2020-02-13
go/cmd/dolt/commands/sql.go: Always close row iters. Improves robustness when SQL returns an error.

reltuk:


reltuk:
@zachmu Ready for review!

timsehn:
Merging so we get this in the release.

PR #403, opened 2020-02-17
Tim/dateformat bats

timsehn:


bats/sql.bats
```diff
@@ -385,6 +394,18 @@ teardown() {
...
 
+@test "sql date_format function" {
+    skip "date_format() not supported" 
+    dolt sql -q "select date_format(date_created, '%Y-%m-%d') from has_datetimes"
+}
+
+@test "sql substr() and cast() functions" {
+    run dolt sql -q "select substr(cast(date_created as char), 1, 4) from has_datetimes"
+    [ $status -eq 0 ]
+    [[ "$output" =~ "2020" ]] || false
```
Hydrocharged:
```suggestion
    [[ "$output" =~ " 2020 " ]] || false
```
I like having the spaces around the value, just in case it does something wonky like returning `38202018`, but it's not required.

PR #404, opened 2020-02-18
iterate a map backward

bheni:
How much do you hate this?


andrew-wm-arthur:
...It works. And we have a roadmap to clean it up in the future

PR #415, opened 2020-02-20
publishrelease/install.sh: install -d /usr/local/bin if it does not exist.

reltuk:


timsehn:
Maybe also warn that it's not in their path?

reltuk:
Warning for path settings requires a little more investigation. Gonna fix the bug where we do the wrong thing and said we did the right thing first...happy to revisit path setting afterwards.

timsehn:
SGTM.

PR #416, opened 2020-02-23
bh/set-algebra

bheni:
Package setalgebra provides the ability to perform algebraic set operations on mathematical sets built directly on noms
types. Unlike standard sets in computer science, which define a finitely sized collection of unordered unique values,
sets in mathematics are defined as a well-defined collection of distint objects. This can include infinitely sized
groupings such as the set of all real numbers greater than 0.

See https://en.wikipedia.org/wiki/Set_(mathematics)

There are 3 types of sets defined in this package: FiniteSet, Interval, and CompositeSet.

FiniteSet is your typical computer science set representing a finite number of unique objects stored in a map. An
example would be the set of strings {"red","blue","green"}, or the set of numbers {5, 73, 127}.

Interval is a set which can be written as an inequality such as {n | n > 0} (set of all numbers n such that n > 0) or a
chained comparison {n | 0.0 <= n <= 1.0 } (set of all floating point values between 0.0 and 1.0)

CompositeSet is a set which is made up of a FiniteSet and one or more non overlapping intervals such as
{n | n < 0 or n > 100} (set of all numbers n below 0 or greater than 100) this set contains 2 non overlapping intervals
and an empty finite set. Alternatively {n | n < 0 or {5,10,15}} (set of all numbers n below 0 or n equal to 5, 10 or 15)
which would be represented by one Interval and a FiniteSet containing 5,10, and 15.

There are 2 special sets also defined in this package: EmptySet, UniversalSet.

The EmptySet is a set that has no values in it.  It has the property that when unioned with any set X, X will be the
result, and if intersected with any set X, EmptySet will be returned.

The UniversalSet is the set containing all values.  It has the property that when unioned with any set X, UniversalSet is
returned and when intersected with any set X, X will be returned.


go/libraries/doltcore/sqle/setalgebra/interval.go
```diff
@@ -0,0 +1,293 @@
...
+// points of another interval. The resulting intervalComparison is an array of 4 ints where a
+// value of -1 means that the point in interval 1 is less than the point in interval 2. A value
+// of 0 indicates equality, and a value of 1 indicates the value in interval 1 is greater than
+// the value in interval 2.
+func compareIntervals(in1, in2 Interval) (intervalComparison, error) {
+	comp := intervalComparison{}
+	endpoints := [2][2]*IntervalEndpoint{{in1.Start, in1.End}, {in2.Start, in2.End}}
+
+	var err error
+	for i := 0; i < 2; i++ {
```
zachmu: You only have N=4 here. This might be easier to read and verify if you just wrote 4 consecutive chunks of comparison logic, rather than structure it as a loop.

go/libraries/doltcore/sqle/setalgebra/interval.go
```diff
@@ -0,0 +1,293 @@
...
+	return comp, nil
+}
+
+// simplifyInterval will return:
+//  * a UniversalSet for an equivalent interval defined as: negative infinity < X < positive infinity
+//  * a FiniteSet with a single value N for an equivalent interval defined as: N <= X <= N
+//  * EmptySet for an interval defined as: N < X < N
+//  * EmptySet for an interval where end < start
+//  * an unchanged interval will be returned for all other conditions
+func simplifyInterval(in Interval) (Set, error) {
```
zachmu: I would consider removing this, not sure it's pulling its weight. Seems like a premature optimization.

go/libraries/doltcore/sqle/setalgebra/union.go
```diff
@@ -0,0 +1,342 @@
...
+
+	copyOf := *ep
+
+	return &copyOf
+}
+
+// finiteSetIntervalUnion will check all the points of the FiniteSet and see which ones are not in the given interval.
+// if all points are in the interval then the resulting set is the interval itself, otherwise a CompositeSet containing
+// the missing points as a new FiniteSet and the interval is returned.
+func finiteSetIntervalUnion(fs FiniteSet, in Interval) (Set, error) {
```
zachmu: Seems needlessly complex when you could just immediately return a CompositeSet that has the exact same semantics.

go/libraries/doltcore/sqle/setalgebra/union.go
```diff
@@ -0,0 +1,342 @@
...
+		return CompositeSet{newSet, []Interval{resInterval}}, nil
+	} else {
+		return resInterval, nil
+	}
+}
+
+// finiteSetCompositeSetUnion checks all the points in a FiniteSet against the CompositeSet to find all points not
+// represented in the CompositeSet (So not in any of it's intervals and not in it's existing FiniteSet), and adds those
+// points to the compositeSet
+func finiteSetCompositeSetUnion(fs FiniteSet, cs CompositeSet) (Set, error) {
```
zachmu: Shouldn't this just be Union(cs.Set, Union(fs, cs.Interval))?

go/libraries/doltcore/sqle/setalgebra/union.go
```diff
@@ -0,0 +1,342 @@
...
+			resIntervToReduce = Interval{in1.nbf, in2.Start, in1.End}
+		}
+	}
+
+	return simplifyInterval(resIntervToReduce)
+}
+
+// intervalCompositeSetUnion will check the CompositeSet's FiniteSet for points that the new Interval contains
+// and exclude those from the resulting composite and then union the Interval with its existing intervals.
+func intervalCompositeSetUnion(in Interval, cs CompositeSet) (Set, error) {
```
zachmu: Can this be Union(cs.fs, UnionWithMultiples(in,  cs.in...)

go/libraries/doltcore/sqle/setalgebra/union.go
```diff
@@ -0,0 +1,342 @@
...
+		return simplifyInterval(intervals[0])
+	} else {
+		return CompositeSet{FiniteSet{hashToVal}, intervals}, nil
+	}
+}
+
+// unionWithMultipleIntervals takes an interval and a slice of intervals and returns a slice of intervals containing
+// the minimum number of intervals required to represent the union.  The src []Interval argument must be in sorted
+// order and only contain non-overlapping intervals.
+func unionWithMultipleIntervals(in Interval, src []Interval) ([]Interval, error) {
```
zachmu: If this returned a Set (which would be a CompositeSet in practice) then you could use it as a primitive to compose other union operations

go/libraries/doltcore/sqle/setalgebra/finite_set.go
```diff
@@ -0,0 +1,81 @@
...
+		return finiteSetIntervalUnion(fs, otherTyped)
+	case CompositeSet:
+		return finiteSetCompositeSetUnion(fs, otherTyped)
+	case EmptySet:
+		return fs, nil
+	case UniversalSet:
+		return otherTyped, nil
+	}
+
+	panic("unknown set type")
```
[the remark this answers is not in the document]
bheni: done

go/libraries/doltcore/sqle/setalgebra/finite_set.go
```diff
@@ -0,0 +1,81 @@
...
+		return finiteSetIntervalIntersection(fs, otherTyped)
+	case CompositeSet:
+		return finiteSetCompositeSetIntersection(fs, otherTyped)
+	case EmptySet:
+		return otherTyped, nil
+	case UniversalSet:
+		return fs, nil
+	}
+
+	panic("unknown set type")
```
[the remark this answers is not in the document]
bheni: done

go/libraries/doltcore/sqle/setalgebra/intersection.go
```diff
@@ -0,0 +1,240 @@
...
+package setalgebra
+
+import (
+	"github.com/liquidata-inc/dolt/go/store/hash"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+// finiteSetIntersection returns the set of points that are in both fs1 and fs2
+func finiteSetIntersection(fs1, fs2 FiniteSet) (Set, error) {
+	hashToVal := make(map[hash.Hash]types.Value, len(fs1.HashToVal)+len(fs2.HashToVal))
```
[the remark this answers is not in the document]
bheni: removed the size.

go/libraries/doltcore/sqle/setalgebra/intersection.go
```diff
@@ -0,0 +1,240 @@
...
+	if len(hashToVal) == 0 {
+		return EmptySet{}, nil
+	} else {
+		return FiniteSet{HashToVal: hashToVal}, nil
+	}
+}
+
+// finiteSetInterval will return the set of points that are in the interval, or an EmptySet instance
+func finiteSetIntervalIntersection(fs FiniteSet, in Interval) (Set, error) {
+	hashToVal := make(map[hash.Hash]types.Value, len(fs.HashToVal))
```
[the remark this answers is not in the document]
bheni: removed

go/libraries/doltcore/sqle/setalgebra/intersection.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if len(hashToVal) == 0 {
+		return EmptySet{}, nil
+	} else {
+		return FiniteSet{HashToVal: hashToVal}, nil
+	}
+}
+
+// finiteSetInterval will return the set of points that are in the composite set, or an EmptySet instance
+func finiteSetCompositeSetIntersection(fs FiniteSet, composite CompositeSet) (Set, error) {
```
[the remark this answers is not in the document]
bheni: I actually tried this.  Because you still need to loop over all the comp.Intervals unioning all the results.  with the error handling on each of the unions and intersections the code actually comes out looking worse imo.

go/libraries/doltcore/sqle/setalgebra/interval.go
```diff
@@ -0,0 +1,293 @@
...
+	case UniversalSet:
+		return in, nil
+	}
+
+	panic("unknown set type")
+
+}
+
+// ValueInInterval returns true if the value falls within the bounds of the interval
+func ValueInInterval(in Interval, val types.Value) (bool, error) {
```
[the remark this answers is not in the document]
bheni: I had it in a separate file, then moved it here.  Will make it a member function as you suggest.

I think in the future this should be an interface method on the Set that all Sets implement.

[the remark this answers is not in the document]
bheni: done

go/libraries/doltcore/sqle/setalgebra/interval.go
```diff
@@ -0,0 +1,293 @@
...
+					lt = i < j
+				}
+			} else if ep1 == nil {
+				// if an intervalEndpoint is nil in the first point it will be less than the
+				// second point if it is a nil start point. In all other cases it is greater,
+				// and it can never be equal to a non nil intervalEndpoint
+				lt = i == 0
+			} else if ep2 == nil {
+				// if an intervalEndpoint is nil in the second point then the first point will be less
+				// if it is a nil end point. In all other cases it is greater greater, and it can never
```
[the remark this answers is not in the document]
bheni: fixed

go/libraries/doltcore/sqle/setalgebra/interval.go
bheni: updated

go/libraries/doltcore/sqle/setalgebra/union.go
bheni: that could result in values in the finiteset that are already covered by the interval.  In the code that actually uses this that would result in a row being in the result set 2x.

go/libraries/doltcore/sqle/setalgebra/union.go
bheni: same as finiteSetCompositeSetUnion

go/libraries/doltcore/sqle/setalgebra/union.go
bheni: Same as finiteSetCompositeSetUnion

PR #423, opened 2020-02-24
/go/libraries/doltcore/sqle/logictest/main/main.go: Add withdurations option

coffeegoddd:


go/libraries/doltcore/sqle/logictest/main/main.go
```diff
@@ -91,6 +91,7 @@ func NewDoltRecordResult(e *logictest.ResultLogEntry) *DoltResultRecord {
 		TestFile:     e.TestFile,
 		LineNum:      e.LineNum,
 		Query:        e.Query,
+		Duration: 	  e.Duration.String(),
```
zachmu: Would make this a numeric type and specify units. We don't have a duration type in dolt.

zachmu: Would make this a numeric type and specify units. We don't have a duration type in dolt.

PR #429, opened 2020-02-26
andy/Test compatibility between Dolt versions

andrew-wm-arthur:


bats/compatibility/test_files/bats/compatibility.bats
```diff
@@ -0,0 +1,158 @@
...
+    [ "$status" -eq 0 ]
+    [[ "${lines[0]}" =~ "abc @ working" ]] || false
+    [[ "${lines[1]}" =~ "CREATE TABLE \`abc\` (" ]] || false
+    [[ "${lines[2]}" =~ "\`pk\` BIGINT NOT NULL COMMENT 'tag:0'," ]] || false
+    [[ "${lines[3]}" =~ "\`a\` LONGTEXT COMMENT 'tag:100'," ]] || false
+    [[ "${lines[4]}" =~ "\`b\` DOUBLE COMMENT 'tag:101'," ]] || false
+    [[ "${lines[5]}" =~ "\`x\` BIGINT COMMENT 'tag:103'," ]] || false
+    [[ "${lines[6]}" =~ "\`y\` BIGINT COMMENT 'tag:104'," ]] || false
+    [[ "${lines[7]}" =~ "PRIMARY KEY (\`pk\`)" ]] || false
+    [[ "${lines[8]}" =~ ");" ]] || false
```
bheni: Checking exact line numbers is really fragile.  Might want to find a better way to do this in the future.

bats/compatibility/runner.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+if [[ $(git diff --stat) != '' ]]; then
```
bheni: Might move this block down below the functions to the main execution body of the script.

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -25,14 +25,21 @@ import (
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
+// Correct Marshalling & Unmarshalling is essential to compatibility across Dolt versions
+// any changes to the fields of Schema or other persisted objects must be append only, no
+// fields can ever be removed without break compatibility.
```
bheni: break -> breaking

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -25,14 +25,21 @@ import (
...
+//
+// the marshalling annotations of new fields must have the "omitempty" option to allow newer
+// versions of Dolt to read objects serialized by older Dolt versions where the field did not
+// yet exists. However, all fields must always be written.
```
bheni: You say:

    fields must have the "omitempty" option

But then you removed the omitempty option from the only field that had it.

PR #441, opened 2020-03-01
Added skipped bats test for update a datetime field

timsehn:


timsehn:
Merging because Zach is going to look at Andy's fix.

PR #451, opened 2020-03-05
/bats/remotes-file-system.bats: Add skipped bats test for fetch display branch bug

coffeegoddd:


bats/remotes-file-system.bats
```diff
@@ -117,3 +117,77 @@ SQL
...
+    [[ "${lines[1]}" != "tester" ]] || false
+
+    # fetch branches from remote
+    dolt fetch
+
+    # should display both branches fetched from remote
+    run dolt branch
+    [ "$status" -eq 0 ]
+    [[ "${lines[0]}" =~ "master" ]] || false
+    skip "Fetched branch not displayed in branch list" [[ "${lines[1]}" =~ "tester" ]] || false
```
andrew-wm-arthur: `skip "..."` exit the test the first time it's called, there's no need to call it again

bats/remotes-file-system.bats
```diff
@@ -117,3 +117,77 @@ SQL
...
+
+    # should not be able to create branch with same
+    # name as remote branch
+    skip "Allows creation of branch with same name as remote branch" run dolt checkout -b tester
+    skip "Allows creation of branch with same name as remote branch" [ "$status" -eq 1 ]
+    skip "Allows creation of branch with same name as remote branch" [[ "${lines[0]}" =~ "fatal: A branch named 'tester' already exists." ]] || false
+
+    # should not be able to checkout branch that is not displayed
+    skip "Allows checkout of branch not displayed in branch list" run dolt checkout tester
```
andrew-wm-arthur: once the bug is fixed, the checks on 191 & 192 will fail because `tester` will be fetched on 175

andrew-wm-arthur:
good catch!

PR #458, opened 2020-03-10
go/libraries/doltcore: Expose diff summary to use in dolthub api

tbantle22:


go/libraries/doltcore/diff/diff_summary.go
```diff
@@ -55,6 +60,50 @@ func Summary(ctx context.Context, ch chan DiffSummaryProgress, v1, v2 types.Map)
 	return nil
 }
 
+// DiffSummary returns a summary of diff changes between two values
+func DiffSummary(ctx context.Context, v1, v2 types.Map) (DifferenceSummary, error) {
```
tbantle22: Will performance be a problem for loading diffs for tables with a lot of rows on DoltHub?

tbantle22:
Decided to go in a different direction

PR #463, opened 2020-03-13
Andy/uniq tags

andrew-wm-arthur:
@bheni the biggest design decision I'm concerned about is highlighted [here](https://github.com/liquidata-inc/dolt/pull/463/files#r393827082)

Comments on exported functions are on the way.

I also enlisted @zachmu 's help to look over all of the `sqle`tests that needed to be modified. Those start [here](https://github.com/liquidata-inc/dolt/pull/463/files#r393837017) and go through the end of the file list.

I'm ironing out the rest of the backwards compatibility issues before I merge.


bats/create-tables.bats
```diff
@@ -94,13 +94,13 @@ CREATE TABLE test1 (
...
-  c3 BIGINT COMMENT 'tag:4',
-  c4 BIGINT COMMENT 'tag:5',
-  c5 BIGINT COMMENT 'tag:6',
+  pk1 BIGINT NOT NULL COMMENT 'tag:10',
+  pk2 BIGINT NOT NULL COMMENT 'tag:11',
+  c1 BIGINT COMMENT 'tag:12',
+  c2 BIGINT COMMENT 'tag:13',
+  c3 BIGINT COMMENT 'tag:14',
+  c4 BIGINT COMMENT 'tag:15',
+  c5 BIGINT COMMENT 'tag:16',
```
andrew-wm-arthur: Bats tests needed to be updated to either
1) Eliminate a tag conflict created by a tag annotation
2) Add a tag annotation to prevent random tag generation

bats/tags.bats
```diff
@@ -78,19 +78,7 @@ SQL
     dolt sql -q "alter table test drop column c1"
     dolt add test
     dolt commit	-m "Committed test table with c1 dropped"
-    # Adding the tag back with the same name and type should be allowed
```
andrew-wm-arthur: we cannot drop and re-add tables that use the same tags

go/cmd/dolt/dolt.go
```diff
@@ -144,7 +145,11 @@ func runMain() int {
 
 	warnIfMaxFilesTooLow()
 
-	dEnv := env.Load(context.TODO(), env.GetCurrentUserHomeDir, filesys.LocalFS, doltdb.LocalDirDoltDB, Version)
+	dEnv := env.Load(context.TODO(), env.GetCurrentUserHomeDir, filesys.LocalFS, doltdb.LocalDirDoltDB, doltdb.DoltVersion)
+
+	if dEnv.DBLoadError == nil {
+		dEnv.DBLoadError = rebase.MaybeMigrateUniqueTags(context.Background(), dEnv)
+	}
```
andrew-wm-arthur: @bheni This where we potentially migrate repos if the repo was created with an older Dolt version.
If this function doensn't find the Dolt version written to the root at each of the branches, it will performa a migration.
Going forward we will write the most recent dolt version to the root value.

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -44,18 +44,18 @@ func TestCreateTable(t *testing.T) {
 	}{
```
andrew-wm-arthur: @zachmu

bats/conflict-detection.bats
```diff
@@ -241,12 +241,12 @@ SQL
     dolt checkout add-table
     dolt sql <<SQL
 CREATE TABLE test2 (
-  pk BIGINT NOT NULL COMMENT 'tag:0',
-  c1 BIGINT COMMENT 'tag:1',
-  c2 BIGINT COMMENT 'tag:2',
-  c3 BIGINT COMMENT 'tag:3',
-  c4 BIGINT COMMENT 'tag:4',
-  c5 BIGINT COMMENT 'tag:5',
+  pk BIGINT NOT NULL COMMENT 'tag:10',
```
andrew-wm-arthur: Bats tests needed to be updated to either
1) Eliminate a tag conflict created by a tag annotation
2) Add a tag annotation to prevent random tag generation

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -186,6 +253,46 @@ func (root *RootValue) GetSuperSchema(ctx context.Context, tName string) (*schem
...
+			return true, err
+		}
+
+		ss, err := encoding.UnmarshalSuperSchemaNomsValue(ctx, root.vrw.Format(), ssVal)
+
+		if err != nil {
+			return true, err
+		}
+
+		sss = append(sss, ss) // go get -f parseltongue
```
zachmu: Nice

go/libraries/doltcore/schema/tag.go
```diff
@@ -0,0 +1,87 @@
...
+		} else if maxTagVal*128 < maxTagVal {
+			maxTagVal = ReservedTagMin - 1
+		} else {
+			maxTagVal = maxTagVal * 128
+		}
+	}
+
+	var randTag uint64
+	for {
+		randTag = uint64(randGen.Int63n(int64(maxTagVal)))
```
zachmu: It would be nice if tag generation is deterministic, so that two people who apply the same set of changes to a repo, in the same order, always end up with the same repo. It can be pseudorandom if you want, but it should use the current root hash for a seed if so. I would prefer sequential personally. If you go that route, then you should cache the highest assigned value in the superschema.

go/libraries/doltcore/schema/tag.go
```diff
@@ -0,0 +1,87 @@
...
+	//
+	SystemTableReservedMin uint64 = 1 << 51
+
+	// InvalidTag is used as an invalid tag
+	InvalidTag uint64 = math.MaxUint64
+)
+
+const (
+	// Tags for dolt_docs table
+	DocNameTag = iota + SystemTableReservedMin
```
zachmu: This is one place where this style of const declaration is a bad idea. These constants are persisted to repos. That means if I reorder this list, it silently changes the values and introduces a breaking change to dolt.

Each table should have its own const block with its own predefined keyspace. We have lots of overhead here. 1000 tags per system table gives us plenty of room to grow.

Also add a comment warning against reordering the consts

bats/conflict-detection.bats
zachmu: Unless it's a test specifically about tag generation, bats tests we shouldn't specify tag numbers when creating tables. Existing examples were mostly copy-pasted (since the output include tag number comments).

go/libraries/doltcore/schema/tag.go
bheni: Yeah, I agree.  @reltuk had some thoughts on this.  Think he was talking about using the hash of the super schema and the name and type to determine a tag.

go/libraries/doltcore/rebase/rebase_tag.go
```diff
@@ -0,0 +1,605 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	_, err = dEnv.PutDocsToStaged(ctx, dd)
+	return err
+}
+
+// TagRebaseForReg rebases the provided DoltRef, swapping all tags in the tagMapping.
```
bheni: typo: TagRebaseForReg -> TagRebaseForRef

go/libraries/doltcore/sqle/tables.go
```diff
@@ -224,16 +224,14 @@ func (t *AlterableDoltTable) AddColumn(ctx *sql.Context, column *sql.Column, ord
...
 
-	sch, err := table.GetSchema(ctx)
-	if err != nil {
-		return err
-	}
-
 	tag := extractTag(column)
 	if tag == schema.InvalidTag {
-		tag = schema.AutoGenerateTag(sch)
+		// TODO: are we sure we want to silently autogen if we fail to parse?
```
zachmu: Yes, because a comment that doesn't contain a tag is still legitimate SQL. We just treat it as if there were not comment on the column.

(We don't actually store the comment in dolt, but someday maybe)

Issue #467, opened 2020-03-14
Running piped SQL file that drops and adds a table produces different results on subsequent runs


timsehn:
Repo: corona-virus
Branch: case_details_virological_dot_org

This SQL: https://drive.google.com/open?id=1nZJyvSfZspMnIixZtmW3uRM0xYT7dloW

When run twice on the same working set produces no diff and the this diff summary:
```
Timothys-MacBook-Pro:corona-virus timsehn$ dolt diff --summary
diff --dolt a/case_details b/case_details
--- a/case_details @ ugcofv63pv8mqvq40h0esr4p7qm010qo
+++ b/case_details @ p11dee7ecl08h9g8ml3jrpnnqj48o5mm
48,601 Rows Unmodified (98.42%)                                         
0 Rows Added (0.00%)
779 Rows Deleted (1.58%)
0 Rows Modified (0.00%)
0 Cells Modified (0.00%)
(49,380 Entries vs 48,601 Entries)
```
It seems to deterministically delete all rows for the Hubei sheet source key with ID above 20463.

zachmu:
Reproduced this with latest build. Not sure what's happening yet, but have some ideas.

PR #471, opened 2020-03-17
Zachmu/batch bug

zachmu:
This fixes https://github.com/liquidata-inc/dolt/issues/467.

The SQL script in question deleted and recreated a subset of the table, and should have resulted in no diff. Before this change, it resulted in some subset of rows being deleted. The issue was that DoltDatabase in batch mode was using mapEditor.Remove and mapEditor.Add for the same keys, which doesn't work in all (most) cases. The solution is to flush the cache before *and* after any non-insert statement when running in batch mode.

This change also makes the output for batch mode more sensible.

Brian to review, Andy and Aaron FYI.

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -170,7 +170,7 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 			defer rowIter.Close()
 			err = se.prettyPrintResults(ctx, se.ddb.ValueReadWriter().Format(), sqlSch, rowIter)
 			if err != nil {
-				cli.Println(color.RedString(err.Error()))
+				return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 			}
 		}
 
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -81,6 +81,7 @@ const (
 	formatFlag  = "result-format"
 	saveFlag    = "save"
 	messageFlag = "message"
+	batchFlag   = "batch"
 	welcomeMsg  = `# Welcome to the DoltSQL shell.
 # Statements must be terminated with ';'.
 # "exit" or "quit" (or Ctrl-D) to exit.`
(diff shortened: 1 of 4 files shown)


go/cmd/dolt/commands/sql.go
```diff
@@ -580,64 +604,126 @@ func processBatchQuery(ctx context.Context, query string, se *sqlEngine) error {
...
+		if rowIter != nil {
+			defer rowIter.Close()
+			err = mergeResultIntoStats(sqlStatement, rowIter, batchEditStats)
+			if err != nil {
+				return fmt.Errorf("Error inserting rows: %v", err.Error())
+			}
 		}
 
-		if batchEditStats.numRowsInserted%maxBatchSize == 0 {
+		if batchEditStats.numUpdates()%maxBatchSize == 0 {
```
bheni: this seems odd to me.  This is only going to flush if it's an exact multiple of maxBatchSize?  shouldn't this be >= instead of % ?

go/cmd/dolt/commands/sql.go
```diff
@@ -580,64 +604,126 @@ func processBatchQuery(ctx context.Context, query string, se *sqlEngine) error {
 		return fmt.Errorf("Error parsing SQL: %v.", err.Error())
 	}
 
-	switch sqlStatement.(type) {
-	case *sqlparser.Insert:
+	if canProcessAsBatchInsert(sqlStatement) {
```
bheni: Seems preferable to put all this code into a separate function that will have it's defered blocks called on exit rather than leaving them dangling
.

go/cmd/dolt/commands/sql.go
```diff
@@ -110,6 +111,7 @@ func (cmd SqlCmd) createArgParser() *argparser.ArgParser {
 	ap.SupportsString(formatFlag, "r", "result output format", "How to format result output. Valid values are tabular, csv. Defaults to tabular. ")
 	ap.SupportsString(saveFlag, "s", "saved query name", "Used with --query, save the query to the query catalog with the name provided. Saved queries can be examined in the dolt_query_catalog system table.")
 	ap.SupportsString(messageFlag, "m", "saved query description", "Used with --query and --save, saves the query with the descriptive message given. See also --name")
+	ap.SupportsFlag(batchFlag, "b", "batch mode, to run more than one query with --query, separated by ';'. Piping input to sql with no arguments also uses batch mode")
```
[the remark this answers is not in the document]
zachmu: Done.

go/cmd/dolt/commands/sql.go
zachmu: Fixed.

go/cmd/dolt/commands/sql.go
zachmu: I extracted a couple functions here. But I did it for me!

PR #473, opened 2020-03-17
Better Syntax Error Messaging

bheni:
Need to update dependencies before merging this code.


go/cmd/dolt/commands/sql.go
```diff
@@ -215,7 +217,7 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 
 		err = runBatchMode(ctx, se, batchInput)
 		if err != nil {
-			_, _ = fmt.Fprintf(cli.CliErr, "Error processing batch: %s\n", err.Error())
+			cli.PrintErrln(color.RedString("Error processing batch"))
```
bheni: The specific error was already printed, so no need to print it again.

go/cmd/dolt/commands/sql.go
```diff
@@ -525,7 +568,7 @@ func processQuery(ctx context.Context, query string, se *sqlEngine) error {
 		// silently skip empty statements
 		return nil
 	} else if err != nil {
-		return fmt.Errorf("Error parsing SQL: %v.", err.Error())
```
zachmu: There are a couple other places where we parse SQL, need to cover those as well

go/cmd/dolt/commands/sql.go
```diff
@@ -202,6 +204,47 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 	return 0
 }
 
+func formatProcessQueryError(err error) errhand.VerboseError {
+	const (
+		maxStatementLen = 128
+		maxPosWhenTruncated = 64
+	)
+
+	verrBuilder := errhand.BuildDError("Error parsing SQL")
```
zachmu: This isn't always accurate -- the methods that return syntax errors also return other types of errors. You need to move this inside the type check

bats/1pk5col-ints.bats
```diff
@@ -129,11 +129,10 @@ if rows[2] != "9,8,7,6,5,4".split(","):
     run dolt sql -q "select * from test"
     [[ "$output" =~ "9" ]] || false
     run dolt sql -q "insert into test (c1,c3,c5) values (50,55,60)"
-    [ "$status" -eq 1 ]
-    [ "$output" = "column name 'pk' is non-nullable but attempted to set default value of null" ]
+    [[ "$output" =~ "column name 'pk' is non-nullable but attempted to set default value of null" ]]
```
zachmu: Seems like most of these changes are the sign of an issue. The error output shouldn't have changed except for syntax errors

go/cmd/dolt/commands/sql.go
```diff
@@ -240,6 +242,45 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
+		for i := 0; i < position; i++ {
+			marker[i] = ' '
+		}
+
+		marker[position] = '^'
+		verrBuilder.AddDetails(string(marker))
+
+		return verrBuilder.Build()
+	} else {
+		return errhand.BuildDError("Error Processing Query '%s'", query).AddCause(err).Build()
```
zachmu: Just capitalize the first word

zachmu: Actually, I think you just want to use errhand.VerboseErrorFromError here, rather than adding the underlying err as a cause. I don't think there's any good reason to be more verbose in the general case, let the client code decide what error message they want to print directly.

go/cmd/dolt/commands/sql.go
```diff
@@ -436,7 +478,8 @@ func runShell(ctx context.Context, se *sqlEngine, dEnv *env.DoltEnv) error {
 		}
 
 		if sqlSch, rowIter, err := processQuery(ctx, query, se); err != nil {
-			shell.Println(color.RedString(err.Error()))
+			verr := formatQueryError(query, err)
+			shell.Println(verr.Verbose())
```
zachmu: You forgot the red!

bheni: verr.Verbose() already prints the first line as red.

go/cmd/dolt/commands/sql.go
bheni: updated

bats/1pk5col-ints.bats
bheni: fixed

PR #476, opened 2020-03-19
Zachmu/as of timestamp

zachmu:
Support for AS OF queries with a timestamp


go/libraries/doltcore/sqle/sqlselect_test.go
```diff
@@ -129,7 +129,7 @@ type testCommitClock struct {
 
 func (tcc *testCommitClock) Now() time.Time {
 	now := time.Unix(0, tcc.unixNano)
-	tcc.unixNano += int64(time.Millisecond)
+	tcc.unixNano += int64(time.Hour)
```
reltuk: ?

zachmu: Most SQL times don't go to ms resolution. This makes it a bit easier to test

PR #487, opened 2020-03-20
fix issue with merge blowing away changes in working

bheni:
Previously we used to fail if there was anything checked out and you tried to merge.  Then we went to the git model where we would allow merging as long as the changes in the working set didn't touch the same tables as the changes in the commit being merged.  For fast forward merges this was stomping working table changes as we'd just set the root and not re-apply the changes to the tables that were modified in the working set.


go/cmd/dolt/commands/merge.go
```diff
@@ -123,13 +124,13 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
...
 				cli.Println("error: Merging is not possible because you have unmerged files.")
-				cli.Println("hint: Fix them up in the work tree, and then use {{.EmphasisLeft}}dolt add <table>{{.EmphasisRight}}")
+				cli.Println("hint: Fix them up in the work tree, and then use 'dolt add <table>'")
 				cli.Println("hint: as appropriate to mark resolution and make a commit.")
 				cli.Println("fatal: Exiting because of an unresolved conflict.")
 				return 1
 			} else if dEnv.IsMergeActive() {
 				cli.Println("error: Merging is not possible because you have not committed an active merge.")
-				cli.Println("hint: add affected tables using {{.EmphasisLeft}}dolt add <table>{{.EmphasisRight}} and commit using {{.EmphasisLeft}}dolt commit -m <msg>{{.EmphasisRight}}")
+				cli.Println("hint: add affected tables using 'dolt add <table>' and commit using {{.EmphasisLeft}}dolt commit -m <msg>{{.EmphasisRight}}")
```
bheni: looks like when docs got modified there was an over zealous search and replace.

timsehn:
This would be a good bats test. @andrew-wm-arthur can you add one as part of the commit?

bheni:
I'll add a bats test

PR #492, opened 2020-03-25
fix return value on failure to push

bheni:


go/cmd/dolt/commands/push.go
```diff
@@ -286,6 +286,7 @@ func pushToRemoteBranch(ctx context.Context, dEnv *env.DoltEnv, srcRef, destRef,
 				cli.Println("hint: Updates were rejected because the tip of your current branch is behind")
 				cli.Println("hint: its remote counterpart. Integrate the remote changes (e.g.")
 				cli.Println("hint: 'dolt pull ...') before pushing again.")
+				return errhand.BuildDError("").Build()
```
bheni: Return an empty error since the details of the error were already printed.

PR #495, opened 2020-03-26
Secondary Indexes Pt. 1

Hydrocharged:
What's Missing: https://github.com/liquidata-inc/dolt/pull/495#issuecomment-619745479
PR Notes: https://github.com/liquidata-inc/dolt/pull/495#issuecomment-623346656


go/libraries/doltcore/schema/index.go
```diff
@@ -0,0 +1,58 @@
...
+	tags    []uint64
+	allTags []uint64
+	pks     []uint64
+}
+
+func (ix *indexImpl) Name() string {
+	return ix.name
+}
+
+func (ix *indexImpl) FullName(tableName string) string {
```
zachmu: Could call this StorageTableName

go/libraries/doltcore/schema/index.go
```diff
@@ -0,0 +1,58 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+type Index interface {
+	// Name returns the name of the index.
+	Name() string
+	// FullName returns the full dolt name of the index.
```
zachmu: Name of the dolt storage table for index values

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+package schema
+
+import "fmt"
+
+//TODO: column renames need to change colToTag
+//TODO: tag changes need to do...a lot
+type IndexCollection interface {
+	// AddIndexCol adds an index with the given name, columns (in index order), and table primary keys (also in order).
+	// Ex. For table {k1, k2, k3} PK(k1, k2) IDX(k3, k2), columns should be (k3, k2) and pks (k1, k2)
+	AddIndexCol(indexName string, cols []string, pks []string) (Index, error)
```
zachmu: AddIndexByColNames

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	// AddIndexTag adds an index with the given name, column allTags (in index order), and table pk allTags (also in order).
+	AddIndexTag(indexName string, tags []uint64, pks []uint64) (Index, error)
```
zachmu: AddIndexByColTags

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	// AllIndexes returns a slice containing all of the indexes in this collection.
+	AllIndexes() []Index
+	// BestMatch takes a collection of columns and finds the index that best matches, if one exists.
+	BestMatch(cols ...string) Index
```
zachmu: What's this for? The engine should be in charge of matching expressions to indexes

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	// ContainsColumnCollection returns whether the collection contains an index that has this exact collection and ordering of columns.
+	ContainsColumnCollection(cols ...string) bool
+	// ContainsColumnTagCollection returns whether the collection contains an index that has this exact collection and ordering of columns.
+	ContainsColumnTagCollection(tags ...uint64) bool
+	// Count returns the number of indexes in this collection.
+	Count() int
+	// Get returns the index with the given name, or nil if it does not exist.
+	Get(indexName string) Index
+	// DropIndex removes an index from the table metadata.
+	DropIndex(indexName string) (Index, error)
```
zachmu: RemoveIndex? Drop is the DB action

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	// HasIndexes returns whether this collection has any indexes.
+	HasIndexes() bool
+	// RenameIndex renames an index in the table metadata.
+	RenameIndex(oldName, newName string) (Index, error)
+}
+
+type indexCollectionImpl struct {
+	indexes       map[string]*indexImpl
+	colTagToIndex map[uint64][]*indexImpl
+	colToTag      map[string]uint64
```
zachmu: Isn't this functionality better offloaded onto the linked schema object?

go/libraries/doltcore/schema/index_coll.go
zachmu: These Add* methods should take a schema obj instead of pk params

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func NewIndexCollection(cols *ColCollection) IndexCollection {
```
zachmu: If the add methods require a schema obj, then you don't need a col collection to create a new one of these

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, tag := range tags {
+		ixc.colTagToIndex[tag] = append(ixc.colTagToIndex[tag], index)
+	}
+	return index, nil
+}
+
+func (ixc *indexCollectionImpl) AllIndexes() []Index {
+	indexes := make([]Index, len(ixc.indexes))
+	i := 0
+	for _, index := range ixc.indexes {
```
zachmu: Can set i in the for loop expression

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,241 @@
...
+	delete(ixc.indexes, oldName)
+	index.name = newName
+	ixc.indexes[newName] = index
+	return index, nil
+}
+
+func combineAllTags(tags []uint64, pks []uint64) []uint64 {
+	allTags :=  make([]uint64, len(tags))
+	_ = copy(allTags, tags)
+	foundPKCols := make(map[uint64]bool)
```
zachmu: This is kind of convoluted. You could just put all tags in a boolean map with value true, then only add the pks that aren't in the map. The middle loop is unnecessary.

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -130,8 +130,15 @@ func (enc encodedTypeInfo) decodeTypeInfo() (typeinfo.TypeInfo, error) {
 	return typeinfo.FromTypeParams(id, enc.Params)
 }
 
+type encodedIndex struct {
+	Name string `noms:"name" json:"name"`
+	Tags []uint64 `noms:"tags" json:"tags"`
+	Pks []uint64 `noms:"pks" json:"pks"`
```
zachmu: Consider restructuring how this is stored. The index is inherently tied to the schema, so it's odd to list the PKs separately here. A couple different options:

* Just store the tags of the indexed columns, including any Pks in the index. On index load, add any missing PKs from the underlying schema as necessary.
* Store the entire index schema of all indexed columns and any additional Pk columns. Then store an indexLen param tells you how many columns are in the index. Then you know that any tags past that length are primary keys, not part of the indexed columns.

I think I like the first option better?

go/libraries/doltcore/schema/index_coll.go
Hydrocharged: This wouldn't be `i`, it would be the key. `ixc.indexes` is a map.

go/libraries/doltcore/schema/schema_impl.go
```diff
@@ -51,7 +53,10 @@ func SchemaFromCols(allCols *ColCollection) Schema {
 	nonPKColColl, _ := NewColCollection(nonPKCols...)
 
 	return &schemaImpl{
-		pkColColl, nonPKColColl, allCols,
+		pkCols: pkColColl,
+		nonPKCols: nonPKColColl,
+		allCols: allCols,
+		indexCollection: NewIndexCollection(allCols),
```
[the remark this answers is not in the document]
Hydrocharged: The created index collection is empty. The index collection holds a map from column name to tag, and the indexes only store the tags, so this just builds that map for the collection.

go/libraries/doltcore/sqle/indexes.go
```diff
@@ -197,6 +144,31 @@ func (di *doltIndex) Expressions() []string {
 	return strs
 }
 
+//TODO: this may not be needed depending on how index resolution is handled
+func (di *doltIndex) AllExpressions() [][]string {
```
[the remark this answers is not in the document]
Hydrocharged: Probably so, it was added when I was originally messing around with `SELECT` before seeing I needed partial key matching first. I left it in as I may need to reference the functionality (or change `Expressions()` as I had initially done), so this is just an artifact of this being a draft PR.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -104,6 +107,203 @@ func (t *DoltTable) PartitionRows(ctx *sql.Context, _ sql.Partition) (sql.RowIte
...
+	var pkNames []string
+	currentTag := uint64(0)
+	foundPKCols := make(map[string]struct{})
+	allTableCols := t.sch.GetAllCols()
+	for _, indexCol := range columns {
+		tableCol, ok := allTableCols.GetByName(indexCol.Name)
+		if !ok {
+			tableCol, ok = allTableCols.GetByNameCaseInsensitive(indexCol.Name)
+			if !ok {
+				return fmt.Errorf("column `%s` does not exist for the table", indexCol.Name)
```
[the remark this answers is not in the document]
Hydrocharged: So I did add an error in `go-mysql-server` that checks for it in there before it ever gets sent, however it's in `plan`. All of our other errors that are exposed for use in integrations are in `core`. As a whole, this error shouldn't really ever be hit (assuming `go-mysql-server` did its job properly), and I do like having the `go-mysql-server` error in `plan` as that's the place for that error to show up, and adding two errors for the same thing with just one being in the typical export location feels weird.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -104,6 +107,203 @@ func (t *DoltTable) PartitionRows(ctx *sql.Context, _ sql.Partition) (sql.RowIte
...
+	if err != nil {
+		return err
+	}
+
+	t.db.SetRoot(newRoot)
+	return nil
+}
+
+func (t *DoltTable) DropIndex(ctx *sql.Context, indexName string) error {
+	index, err := t.sch.Indexes().DropIndex(indexName)
```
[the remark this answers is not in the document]
Hydrocharged: So for this, `DropIndex` on the index collection does its own check, and returns an error there.

Hydrocharged:
What's still missing?

- Tests. All of the tests. I'm 99.9% sure that everything is working perfectly, as I'm still working on the tests. I've already caught some stuff, so instead of including some of the tests here, this is everything sans index-specific testing.
- Unique indexes don't yet enforce anything. It required working indexes first, so while I attempted it earlier, I decided to get the partial keys in first instead of making another hack, just to get replaced anyway.
- `SHOW INDEX` doesn't show anything. I don't think it's critically important, so I'll make another PR in the future to add it.
- `dolt diff` doesn't display index additions or modifications. Fairly critical functionality, so it's next on the menu, but that should also be a separate PR in my opinion. This one is already too large.
- Speed is pretty bad. With the missing partial key support, they're not performant.

So, my priority list on the missing things are as follows:

- Tests
- Diffs
- Partial Key Support
- Unique Constraint

Besides what's missing, go ahead and give this the full review treatment. It's fully passing all existing tests, so any additional non-feedback changes to the code will come from bugs discovered in testing, which will be a separate PR.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -104,6 +108,147 @@ func (t *DoltTable) PartitionRows(ctx *sql.Context, _ sql.Partition) (sql.RowIte
...
+	if err != nil {
+		return err
+	}
+	newTable, err := doltdb.NewTable(ctx, t.table.ValueReadWriter(), newSchemaVal, tableRowData, indexData)
+	if err != nil {
+		return err
+	}
+
+	// populate the index row data
+	indexRowData, err := types.NewMap(ctx, newTable.ValueReadWriter())
```
zachmu, line 155: Shouldn't this call table.RebuildIndex?

go/libraries/doltcore/sqle/index_lookup.go
```diff
@@ -0,0 +1,139 @@
...
+type doltIndexLookup struct {
+	idx     DoltIndex
+	keyIter IndexLookupKeyIterator
+}
+
+func (il *doltIndexLookup) Indexes() []string {
+	return []string{il.idx.ID()}
+}
+
+// No idea what this is used for, examples aren't useful. From stepping through the code I know that we get index values
```
zachmu, line 42: I know what this is for now :)

In the meantime you can delete the comment, as it's not helpful.

go/libraries/doltcore/schema/index.go
```diff
@@ -0,0 +1,136 @@
...
+	// The result is equivalent to a schema check on the associated dolt index table.
+	AllTags() []uint64
+	// Columns returns the names of the columns in the index.
+	Columns() []string
+	// Comment returns the comment that was provided upon index creation.
+	Comment() string
+	// GetColumn returns the column for the given tag and whether the column was found or not.
+	GetColumn(tag uint64) (Column, bool)
+	// InnerIndexes returns all inner indexes
+	InnerIndexes() []InnerIndex
```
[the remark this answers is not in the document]
zachmu: I think I figured this out, see my other comments on this type

go/libraries/doltcore/schema/inner_index.go
```diff
@@ -0,0 +1,63 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+import "strconv"
+
+type InnerIndex interface {
```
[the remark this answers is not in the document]
zachmu: So it looks like you're using this in DoltIndex to work with partial key matches. Personally I think you should move that complexity completely onto the sqle package to the extent that it needs to exist, and delete this.

go/libraries/doltcore/sqle/index_driver.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, nil
+	}
+
+	sch, err := tbl.GetSchema(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var sqlIndexes []sql.Index
+	for i := 0; i < sch.GetPKCols().Size(); i++ {
```
zachmu, line 75: This partial matching of primary keys is a separable concern. It will be good to have, but let's land your new code path for secondary indexes and then iterate. In the meantime just revert the existing working PK index lookup back to their old boring format and add a TODO.

go/libraries/doltcore/sqle/index_driver.go
```diff
@@ -0,0 +1,118 @@
...
+			db:        database,
+			sch:       sch,
+			tableName: table,
+			table:     tbl,
+			driver:    driver,
+			cols:      cols,
+			ctx:       ctx,
+		})
+	}
+	for _, index := range sch.Indexes().AllIndexes() {
```
zachmu, line 99: Matching partial indexes is nice to have, but I don't think it's necessary for the first pass here and it adds a lot of complexity. Start with getting complete exact matches working and iterate from there. Delete the InnerIndexes construct entirely -- when it comes time for partial key matching, we can compute all the partial keys here in a loop (or in a method private to this package)

go/libraries/doltcore/sqle/table_editor.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/src-d/go-mysql-server/sql"
```
zachmu, line 21: Lots of added complexity in here. Need robust tests.

go/cmd/dolt/commands/indexcmds/cat.go
```diff
@@ -0,0 +1,180 @@
+// Copyright 2020 Liquidata, Inc.
```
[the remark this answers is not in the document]
Hydrocharged: These aren't intended for users to use. The idea behind these are for usage in `bats`  tests. As the indexes aren't normal tables, I wanted some way to be able to interact with them for testing purposes, hence these hidden commands. No user should ever know about these, unless they're going through the source code, and that's what the warnings are for.

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -632,7 +632,12 @@ func (root *RootValue) CreateEmptyTable(ctx context.Context, tName string, sch s
 		return nil, err
 	}
 
-	tbl, err := NewTable(ctx, root.VRW(), schVal, m)
+	indexData, err := types.NewMap(ctx, root.VRW())
```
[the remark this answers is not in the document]
Hydrocharged: I did it this way since `types.Map` is a struct and can't be nil, but I can change it to a `*types.Map`

go/libraries/doltcore/merge/merge.go
```diff
@@ -170,7 +172,19 @@ func (merger *Merger) MergeTable(ctx context.Context, tblName string) (*doltdb.T
...
+		return nil, nil, err
+	}
+
+	mergedTable, err := doltdb.NewTable(ctx, merger.vrw, schUnionVal, mergedRowData, emptyIndexData)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	mergedTable, err = mergedTable.RebuildIndexData(ctx)
```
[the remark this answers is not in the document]
Hydrocharged: This is something we can discuss more, but the current merge path works with tables, of which these aren't anymore. To fix this would be to rewrite the merge logic, and I'm not sure how large of a change it would be, but that's something that I definitely do not plan on doing anytime soon.

go/libraries/doltcore/merge/resolve.go
```diff
@@ -42,54 +44,70 @@ func ResolveTable(ctx context.Context, vrw types.ValueReadWriter, tbl *doltdb.Ta
...
 
 	rowData, err := tbl.GetRowData(ctx)
-
 	if err != nil {
 		return nil, err
 	}
-
 	rowEditor := rowData.Edit()
 
+	indexes := tblSch.Indexes().AllIndexes()
```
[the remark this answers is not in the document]
Hydrocharged: When you do a merge, we rebuild the indexes based on the current merged data. However, if there are conflicts, then those must be resolved. This is where this code comes in, whenever you resolve a conflict, the indexes are updated. So they're only rebuilt on the initial merge.

go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -51,12 +51,37 @@ func ModifyColumn(
...
 		return nil, err
 	}
 
-	return updateTableWithModifiedColumn(ctx, tbl, newSchema, newCol)
+	updatedTable, err := updateTableWithModifiedColumn(ctx, tbl, newSchema, newCol)
+	if err != nil {
+		return nil, err
+	}
+
+	if newCol.IsPartOfPK != existingCol.IsPartOfPK {
```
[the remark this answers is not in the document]
Hydrocharged: Is it truly best to delete them and add a note? In the future, whenever they are added, if the implementor were to delete the comment in anticipation of writing the code and then forget to actually do so, then we'd be left with a bug. This is something we're planning to add at some point in the future, so I'd vote to leave the code to prevent a potential future bug.

go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -176,5 +206,17 @@ func replaceColumnInSchema(sch schema.Schema, oldColName string, newCol schema.C
 		return nil, err
 	}
 
-	return schema.SchemaFromCols(collection), nil
+	newSch := schema.SchemaFromCols(collection)
+
+	if oldCol.Tag != newCol.Tag {
```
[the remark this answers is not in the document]
Hydrocharged: Is this something that we'll never support?

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -171,7 +189,15 @@ func (sd schemaData) decodeSchema() (schema.Schema, error) {
 		return nil, err
 	}
 
-	return schema.SchemaFromCols(colColl), nil
+	sch := schema.SchemaFromCols(colColl)
```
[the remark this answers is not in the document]
Hydrocharged: Yes, the array will just be `nil`, which go will handle as the empty array.

go/libraries/doltcore/schema/index.go
```diff
@@ -0,0 +1,136 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+type Index interface {
+	// AllTags returns the tags of the columns in the entire index, including the primary keys.
+	// The result is equivalent to a schema check on the associated dolt index table.
```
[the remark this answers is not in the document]
Hydrocharged: The result of the call, the `[]uint64` return value. I can clarify this.

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,349 @@
...
+	AddIndex(indexes ...Index)
+	// AddIndexByColNames adds an index with the given name and columns (in index order).
+	AddIndexByColNames(indexName string, cols []string, isUnique bool, comment string) (Index, error)
+	// AddIndexByColTags adds an index with the given name and column tags (in index order).
+	AddIndexByColTags(indexName string, tags []uint64, isUnique bool, comment string) (Index, error)
+	// AllIndexes returns a slice containing all of the indexes in this collection.
+	AllIndexes() []Index
+	// ChangeTags changes the tags of any indexes that reference columns who have the original tag, and returns every index,
+	// including those that were not changed.
+	ChangeTags(changes ...IndexTagChange) ([]Index, error)
```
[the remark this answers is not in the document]
Hydrocharged: Indexes are stored by tag numbers only. So for
```sql
CREATE TABLE test2 (
    `pk1` INT NOT NULL COMMENT 'tag:6801'
    `pk2` INT NOT NULL COMMENT 'tag:4776'
    `v1` INT COMMENT 'tag:15071'
    `v2` INT COMMENT 'tag:13324'
     PRIMARY KEY (`pk1`, `pk2`),
     INDEX `idx_v` (`v2`,`v1`)
  );
```
The index `idx_v` is stored as `[13324, 15071]`. In the event that a tag is changed for a column, all indexes that refer to the old tag must be updated, and that's the purpose of this function. It does not the current `IndexCollection`, it instead returns an array of all of the indexes, with their tags changed to match the given `IndexTagChange`s.

This function is for those very particular circumstances.

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,349 @@
...
+	index := &indexImpl{
+		indexColl: ixc,
+		name:      indexName,
+		isUnique:  isUnique,
+		comment:   comment,
+	}
+	ixc.indexes[indexName] = index
+	tags := make([]uint64, len(cols))
+	for i, col := range cols {
+		tag := ixc.colColl.NameToCol[col].Tag
```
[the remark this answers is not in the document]
Hydrocharged: Ahh, I am checking at the parent call location, but not within the function. Thanks for the catch!

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func (ixc *indexCollectionImpl) AddIndexByColTags(indexName string, tags []uint64, isUnique bool, comment string) (Index, error) {
+	if ixc.Contains(indexName) {
+		return nil, fmt.Errorf("`%s` already exists as an index for this table", indexName)
+	}
+	if ixc.ContainsColumnTagCollection(tags...) {
+		return nil, fmt.Errorf("cannot create a duplicate index on this table")
+	}
+	index := &indexImpl{
```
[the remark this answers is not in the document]
Hydrocharged: A little different than above, this is only called when unmarshalling, so it's always going to have the tags, but I should still check here since that's not guaranteed to always be true.

go/libraries/doltcore/schema/inner_index.go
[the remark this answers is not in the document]
Hydrocharged: Right, this essentially represents a functioning subset of an index. For an index with columns `a, b, c`, it should support matches to `a`, `a, b`, and `a, b, c`. I wasn't sure what to call those three different subsets, so I just named them `InnerIndex`. I can see about moving it to `sqle`.

go/libraries/doltcore/sqle/tables.go
Hydrocharged: It should, I wrote this before writing `RebuildIndexData` and the related functions. I'll throw that in here instead, great catch!

go/libraries/doltcore/sqle/index_driver.go
Hydrocharged: Partial matching of primary keys and indexes are related due to their implementation. As you know, indexes are basically tables with every column being a primary key, so they necessitate the ability to partially match a primary key.

I can't really revert the code without deleting the current, working index usage. For the most part, adding true partial key support will just be adding/changing the `IndexLookupKeyIterator`, with everything else remaining fairly as-is.

go/libraries/doltcore/sqle/index_driver.go
Hydrocharged: Similar to what's written above, matching partial indexes and full indexes are the same problem. A partial index match is programmatically equivalent to a full index match since a full index is still a partial key on the map.

It does work, all of the code here works. The only issue is that it's doing a table scan in `IndexLookupKeyIterator.NextKey`, so it is as slow as having no index, since that's also a full table scan.

Without partial matching of any kind (which I'm just doing a table scan for now), indexes cannot exist with their current implementation. I can't really see a point in reverting this, but we can discuss this is further detail.

go/libraries/doltcore/doltdb/root_val.go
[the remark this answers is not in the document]
zachmu: Another way to approach it is to create a new method, NewTableWithIndexes, that takes the new value, and leave the existing method alone. Then the existing method can dispatch to the new one with types.EmptyMap as the other parameter.

go/libraries/doltcore/sqle/index_driver.go
zachmu: You're confusing the implementation details with the features. The implementation for secondary indexes involves partial primary key matching at the storage layer. As a feature, we can support matching *all* of the columns in an index, or *some subset* of the columns in an index. What I'm saying is that we should start by supporting just matching *all* the columns in an index, and add support for partial matches later. When we do that, it might make more sense to implement at least partially in the go-mysql-server layer, rather than declaring each partial index to the framework separately, as you're doing now. I'm not sure yet, let's decide later. Start small and iterate. You have enough complexity and testing burden here for 10 PRs already.

go/libraries/doltcore/sqle/index_driver.go
zachmu: See my other comment. Let's launch matches on every indexed column before moving on to deciding how to implement partial index matches.

go/cmd/dolt/commands/indexcmds/cat.go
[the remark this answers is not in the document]
zachmu: I think they're fine to leave for now, just keep them minimal and keep the bats tests reliance on them as low as possible. In the very near future we need to support all the SQL statements to manage indexes, and we should be writing bats tests that use SQL, not these commands.

Hydrocharged:
PR feedback, test fixes, and the bats tests now have their own commits under the **Commits** tab, so if you're interested in those, you can check them out individually there.

go/libraries/doltcore/sqle/table_editor.go
```diff
@@ -380,3 +281,98 @@ func (te *tableEditor) flush(ctx *sql.Context) error {
 	te.t.table = newTable
 	return te.t.db.SetRoot(ctx, newRoot)
 }
+
+func (te *tableEditor) updateIndexes(ctx *sql.Context, tbl *doltdb.Table, originalRowData types.Map, updated types.Map) (*doltdb.Table, error) {
+	if len(te.indexEds) == 0 {
+		return tbl, nil
+	}
+
+	for _, key := range te.affectedKeys {
```
zachmu, line 290: I notice that affectedKeys isn't written to for Delete. Won't that mean we don't update indexes when a row is deleted?

go/libraries/doltcore/sqle/table_editor.go
```diff
@@ -41,46 +42,39 @@ var ErrDuplicatePrimaryKeyFmt = "duplicate primary key given: (%v)"
...
 	ed           *types.MapEditor
-	idx          schema.Index
 	insertedKeys map[hash.Hash]types.Value
 	addedKeys    map[hash.Hash]types.Value
 	removedKeys  map[hash.Hash]types.Value
+	affectedKeys map[hash.Hash]types.Value
+	indexEds     []*indexEditor
+}
+
+type indexEditor struct {
```
zachmu, line 54: I like this abstraction a lot, much cleaner. Just make sure that it's thoroughly tested, this code has a lot of weird edge cases. It looks to me like you might have a bug on delete?

go/libraries/doltcore/sqle/table_editor.go
Hydrocharged: It's there, I just you just accidentally overlooked it lol
```go
func (te *tableEditor) Delete(ctx *sql.Context, sqlRow sql.Row) error {
	dRow, err := SqlRowToDoltRow(te.t.table.Format(), sqlRow, te.t.sch)
	if err != nil {
		return err
	}

	key, err := dRow.NomsMapKey(te.t.sch).Value(ctx)
	if err != nil {
		return errhand.BuildDError("failed to get row key").AddCause(err).Build()
	}
	hash, err := key.Hash(dRow.Format())
	if err != nil {
		return err
	}

	delete(te.addedKeys, hash)
	te.removedKeys[hash] = key
	te.affectedKeys[hash] = key

	if te.ed == nil {
		te.ed, err = te.newMapEditor(ctx)
		if err != nil {
			return err
		}
	}

	te.ed = te.ed.Remove(key)
	return nil
}
```

PR #497, opened 2020-03-26
Changes to support per connection sql state

bheni:


go/go.mod
```diff
@@ -87,7 +87,8 @@ require (
 
 replace github.com/liquidata-inc/dolt/go/gen/proto/dolt/services/eventsapi => ./gen/proto/dolt/services/eventsapi
 
-replace github.com/src-d/go-mysql-server => github.com/liquidata-inc/go-mysql-server v0.5.1-0.20200323172345-f4d8498ef438
+//replace github.com/src-d/go-mysql-server => github.com/liquidata-inc/go-mysql-server v0.5.1-0.20200323172345-f4d8498ef438
+replace github.com/src-d/go-mysql-server => ../../go-mysql-server
```
bheni: Once the go-mysql-server changes have been merged I'll fix this.

go/libraries/doltcore/sqle/database.go
```diff
@@ -44,62 +47,129 @@ const (
 	single  batchMode = false
 )
 
+type tableCache struct {
+	mu *sync.Mutex
+	tables map[string]map[*doltdb.RootValue]map[string]sql.Table
+}
```
bheni: I'm still not convinced that a per root cache is going to give the correct behavior in the long term, but maintaining it for now.

go/libraries/doltcore/sqle/database.go
```diff
@@ -44,62 +47,129 @@ const (
...
+var _ sql.Database = Database{}
+var _ sql.VersionedDatabase = Database{}
+var _ sql.TableDropper = Database{}
+var _ sql.TableCreator = Database{}
+var _ sql.TableRenamer = Database{}
 
 // NewDatabase returns a new dolt database to use in queries.
-func NewDatabase(name string, root *doltdb.RootValue, ddb *doltdb.DoltDB, rsr env.RepoStateReader) *Database {
-	return &Database{
+func NewDatabase(name string, defRoot *doltdb.RootValue, ddb *doltdb.DoltDB, rsr env.RepoStateReader) Database {
```
bheni: One of the larger changes is that I changed all the receivers for Database to be value types.  The Database object is immutable.

go/store/datas/database_common.go
```diff
@@ -282,7 +282,11 @@ func (db *database) CommitValue(ctx context.Context, ds Dataset, v types.Value)
 	return db.Commit(ctx, ds, v, CommitOptions{})
 }
 
-// doCommit manages concurrent access the single logical piece of mutable state: the current Root. doCommit is optimistic in that it is attempting to update head making the assumption that currentRootHash is the hash of the current head. The call to Commit below will return an 'ErrOptimisticLockFailed' error if that assumption fails (e.g. because of a race with another writer) and the entire algorithm must be tried again. This method will also fail and return an 'ErrMergeNeeded' error if the |commit| is not a descendent of the current dataset head
+// doCommit manages concurrent access the single logical piece of mutable state: the current Root. doCommit is
+// optimistic in that it is attempting to update head making the assumption that currentRootHash is the hash of the
+// current head. The call to Commit below will return an 'ErrOptimisticLockFailed' error if that assumption fails (e.g.
+// because of a race with another writer) and the entire algorithm must be tried again. This method will also fail and
+// return an 'ErrMergeNeeded' error if the |commit| is not a descendent of the current dataset head
```
bheni: Not sure why this one comment was on one line, but it annoyed me enough that I changed it.

go/cmd/dolt/commands/sql.go
```diff
@@ -927,36 +945,49 @@ const (
...
-	sdb          *dsqle.Database
+	sdb          dsqle.Database
 	ddb          *doltdb.DoltDB
 	engine       *sqle.Engine
 	resultFormat resultFormat
 }
 
 // sqlEngine packages up the context necessary to run sql queries against sqle.
-func newSqlEngine(ctx context.Context, dEnv *env.DoltEnv, db *dsqle.Database, format resultFormat) (*sqlEngine, error) {
+func newSqlEngine(ctx context.Context, dEnv *env.DoltEnv, root *doltdb.RootValue, db dsqle.Database, format resultFormat) (*sqlEngine, *sql.Context, error) {
```
zachmu: Similar to comments on your other PR, I don't like this coupling (returning a sql engine and the context required to use it). Would prefer a method on the sqlEngine object that creates new context objects on demand.

go/libraries/doltcore/sqle/database.go
zachmu: Why? Doesn't this result in a struct copy on every method call?

go/libraries/doltcore/sqle/rows.go
```diff
@@ -30,19 +31,19 @@ type doltTableRowIter struct {
 	sql.RowIter
 	table    *DoltTable
 	rowData  types.Map
-	ctx      *sql.Context
+	ctx      context.Context
```
zachmu: Everything in this package should use *sql.Context

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2019 Liquidata, Inc.
```
zachmu: Need to update your copyright headers

go/cmd/dolt/commands/sql.go
zachmu: This is my main comment on this PR. The rest of it looks fine.

go/libraries/doltcore/sqle/rows.go
bheni: fixed

go/libraries/doltcore/sqle/dolt_session.go
bheni: done

go/cmd/dolt/commands/sql.go
bheni: uncoupled.  creating the sql.Context externally and passing it in.

go/libraries/doltcore/sqle/database.go
bheni: It does. But I don't really see this as a performance issue.  This seems to be the way go prefers to do immutability unless the struct gets large.  Taking this conversation to gchat to get people's thoughts.

go/libraries/doltcore/sqle/database.go
```diff
@@ -44,62 +47,129 @@ const (
 	single  batchMode = false
 )
 
+type tableCache struct {
+	mu     *sync.Mutex
+	tables map[string]map[*doltdb.RootValue]map[string]sql.Table
```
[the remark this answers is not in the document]
bheni: fixed

PR #502, opened 2020-03-30
hashof function

bheni:


bheni:
rolling this into my writable dolt_branches PR

PR #506, opened 2020-03-31
Andy/force push & force fetch

andrew-wm-arthur:


go/cmd/dolt/commands/fetch.go
```diff
@@ -161,7 +168,7 @@ func mapRefspecsToRemotes(refSpecs []ref.RemoteRefSpec, dEnv *env.DoltEnv) (map[
 	return rsToRem, nil
 }
 
-func fetchRefSpecs(ctx context.Context, dEnv *env.DoltEnv, rem env.Remote, refSpecs []ref.RemoteRefSpec) errhand.VerboseError {
+func fetchRefSpecs(ctx context.Context, forceFetch bool, dEnv *env.DoltEnv, rem env.Remote, refSpecs []ref.RemoteRefSpec) errhand.VerboseError {
```
andrew-wm-arthur, line 171: unideal naked boolean. I could leave it, pass the argParser, or reorganize everything to create a separate forceFetch path. 
Thoughts?

reltuk: Can also technically do something like:

```go
type RefUpdateMode struct {
        bool
}
var ForceUpdate = RefUpdateMode{true}
var FastForwardOnly = RefUpdateMode{false}

func main() {
        a(ForceUpdate)
        a(FastForwardOnly)
}

func a(mode RefUpdateMode) {
        fmt.Println(mode.bool)
}
```

reltuk: I think naked boolean is fine here for now. This method isn't public, for one...

go/cmd/dolt/commands/push.go
```diff
@@ -204,6 +206,12 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
 					verr = bdr.Build()
 				} else if src == ref.EmptyBranchRef {
 					verr = deleteRemoteBranch(ctx, dest, remoteRef, dEnv.DoltDB, destDB, remote)
+				} else if apr.Contains(ForcePushFlag) {
+					verr = deleteRemoteBranch(ctx, dest, remoteRef, dEnv.DoltDB, destDB, remote)
```
reltuk, line 210: Is it worth adding a TODO or something? This should be first-class at the protocol level and the remote should handle the check...this suffers from dropping the remote ref before we get the new one up there...

go/cmd/dolt/commands/fetch.go
```diff
@@ -179,19 +186,32 @@ func fetchRefSpecs(ctx context.Context, dEnv *env.DoltEnv, rem env.Remote, refSp
...
 			if remoteTrackRef != nil {
-				verr := fetchRemoteBranch(ctx, dEnv, rem, srcDB, dEnv.DoltDB, branchRef, remoteTrackRef)
+				srcDBCommit, verr := fetchRemoteBranch(ctx, dEnv, rem, srcDB, dEnv.DoltDB, branchRef, remoteTrackRef)
 
 				if verr != nil {
 					return verr
 				}
+
+				if forceFetch {
+					err = dEnv.DoltDB.DeleteBranch(ctx, remoteTrackRef)
```
reltuk, line 196: How do we feel about this, versus just exposing a `SetHead` method? Seems like we're going to need/want that for an `update-ref` type functionality eventually anyway...and it would make this truly atomic.

```go
if forceFetch {
    err = dEnv.DoltDB.SetHead(ctx, remoteTrackRef, srcDBCommit)
} else {
    err = dEnv.DoltDB.FastForward(ctx, remoteTrackRef, srcDBCommit)
}
if err != nil {
    return errhand.BuildDError("error: fetch failed").AddCause(err).Build()
}
```

go/cmd/dolt/commands/fetch.go
```diff
@@ -179,19 +186,32 @@ func fetchRefSpecs(ctx context.Context, dEnv *env.DoltEnv, rem env.Remote, refSp
...
+					if err != nil {
+						return errhand.BuildDError("error: fetch failed").AddCause(err).Build()
+					}
+				}
+
+				err =  dEnv.DoltDB.FastForward(ctx, remoteTrackRef, srcDBCommit)
```
reltuk, line 202: I know you didn't cause this...issue...but reading through the code, I think before we call this method, we need to call `dEnv.DoltDB.CanFastForward(ctx, remoteTrackRef, srcDBCommit)`

The `FastForward` method itself doesn't actually implement the correctness check. But it does seem to go through some code that has a quick sanity check that will cause it to fail for a non-FF merge in some cases...

go/cmd/dolt/commands/push.go
andrew-wm-arthur: definitely

PR #514, opened 2020-04-03
CSV and JSON Import Behavior Changes

Hydrocharged:
Based on what was discussed before, this PR modifies the import behavior so that, when importing CSV files, if the destination schema has a bool replacement (`BIT(1)` or `TINYINT/BOOLEAN`) then it will parse `TRUE` and `FALSE` (and other variations) as their respective integers. This restores a behavior that was removed when the new type system was introduced, and all of the code paths were coalesced.

In addition, this PR also fixes a bug where importing a null value through JSON would ignore a column having `NOT NULL` set.


bats/helper/empty-strings-null-values-sch1.json
```diff
@@ -0,0 +1,30 @@
+{
```
zachmu, line 1: Are we still doing schema.json files? I think these will be ripped out soon. You should specify test table schemas with SQL create table commands

go/libraries/doltcore/rowconv/row_converter.go
```diff
@@ -40,7 +40,7 @@ func newIdentityConverter(mapping *FieldMapping) *RowConverter {
 }
 
 // NewRowConverter creates a a row converter from a given FieldMapping.
-func NewRowConverter(mapping *FieldMapping) (*RowConverter, error) {
+func NewRowConverter(mapping *FieldMapping, specialBoolHandling bool) (*RowConverter, error) {
```
zachmu, line 43: This should just always be true, and it should only go one way: when we are importing the string TRUE or FALSE (case insensitive) into an integer column, auto convert to 1 or 0.

Or is there some compelling reason not to always do this?

bats/helper/empty-strings-null-values-sch1.json
Hydrocharged: I updated the current test "_create a table with null values from csv import with json file_" in **create-tables.bats** so that the current JSON schema file had `-sch` appended to it. The test makes use of the `import -c` argument, hence the need for the JSON file. To minimize differences between the tests, I just re-used the JSON schema that was already there, and made a new one by making slight modifications.

go/libraries/doltcore/rowconv/row_converter.go
Hydrocharged: So `NewRowConverter` is used in 3 places from what I can see: imports, diffs, and conflicts. I haven't dove deep into exactly how the function is implemented in diffs and conflicts, so I can't say anything conclusive there.

I disagree that `"true"` and `"false"` strings should always autoconvert to 1 and 0 respectively. I prefer that all behavior is similar program-wide. I'm not too big of a fan of this change to begin with, as now SQL and imports deviate in behavior, and I'm **very** hesitant to apply it elsewhere.

zachmu: @bheni should weigh in here, since he is most familiar with this code's usage.

Agree with the idea we should be consistent. If this bothers you, one way to fix would be to pre-process the CSV / JSON files with a pipeline stage that specifically handles boolean values represented as strings. Then by the time it got here, you would already have numerical values.

As a matter of interface design, rather than passing a bool param to this method, just create a new method with a name that reflects its usage and the semantics you want it to have. Something like NewStringToBoolRowConverter or similar.

bats/helper/empty-strings-null-values-sch1.json
zachmu: I would just use a sql -q command to create the table, followed by import -u.

PR #518, opened 2020-04-03
Andy/autogen migrate

andrew-wm-arthur:
use the same tag generation method in the migration that we do in creating tables

The old migration just assigned sequential tags to all of the columns in the repo.

### DIFF
--- a/go/libraries/doltcore/rebase/rebase_tag.go
+++ b/go/libraries/doltcore/rebase/rebase_tag.go
@@ -580,27 +580,28 @@ func buildGlobalTagMapping(ctx context.Context, root *doltdb.RootValue, parentRo
 
 		var colNames []string
 		var colKinds []types.NomsKind
-		var existingTags []uint64
+		var oldTags []uint64
 		_ = cc.Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
 			colNames = append(colNames, col.Name)
 			colKinds = append(colKinds, col.Kind)
-			existingTags = append(existingTags, tag)
+			oldTags = append(oldTags, tag)
 			return false, nil
 		})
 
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/rebase/rebase_tag.go
```diff
@@ -73,6 +74,120 @@ func NeedsUniqueTagMigration(ctx context.Context, dEnv *env.DoltEnv) (bool, erro
 	return false, nil
 }
 
+// MigrateUniqueTags rebases the history of the repo to uniquify tags within branch histories.
```
andrew-wm-arthur, line 77: this is a copy/paste, it should be at top of this file

andrew-wm-arthur:
The diff for the PR isn't great. The [second commit](https://github.com/liquidata-inc/dolt/pull/518/commits/e6a9236275fafe8eec8c290cfd727ad35f64397f) contains the change

go/libraries/doltcore/rebase/rebase_tag.go
```diff
@@ -234,6 +351,12 @@ func replayCommitWithNewTag(ctx context.Context, root, parentRoot, rebasedParent
 			return nil, err
 		}
 
+		rebasedParentRows, err = dropValsForDeletedColumns(ctx, root.VRW().Format(), rebasedParentRows, rebasedSch, rebasedParentSch)
```
zachmu, line 354: Do we do this in the general case of dropping a column? We didn't use to

andrew-wm-arthur: We do, yes. https://github.com/liquidata-inc/dolt/pull/386

PR #523, opened 2020-04-04
fix logictest

bheni:


coffeegoddd:
Running `sqllogictests` against this branch seems to work fine... i think this is fine to merge, maybe just an eyeball from @zachmu as well

PR #525, opened 2020-04-05
Fixed bad regexes in diff --sql tests. 

timsehn:
Now these are failed tests that need to be fixed.

Andy, sorry I did not catch these bad regexes in review the first time. Unfortunately, these are working regexes now and the tests do not test what you think they do. We need to go back and fix the tests or fix the bad behavior in the diff --sql


bats/sql-diff.bats
```diff
@@ -154,7 +154,7 @@ SQL
     # confirm a difference exists
     run dolt diff --sql newbranch firstbranch
     [ "$status" -eq 0 ]
-    [[ "$output" != "" ]] || false
+    [[ ! "$output" =~ "" ]] || false
```
andrew-wm-arthur, line 157: this one is different than 37, 77, & 117

bats/sql-diff.bats
```diff
@@ -355,7 +355,7 @@ SQL
...
```
andrew-wm-arthur, line 358: this one also breaks the pattern

PR #530, opened 2020-04-07
store/datas/database.go: Update datas.CanUsePuller to check for write support in TableFileStore.

reltuk:
The TableFileStore optimizations accidentally broke push support for
non-dotlremoteapi/non-filepath chunk stores. Attempting to push an AWS remote
currently results in an error.

This is a quick minimal PR to check for write support in a TableFileStore and
not use the push/pull fast path if it isn't supported.


reltuk:
This is a backport of https://github.com/liquidata-inc/dolt/pull/531 in case we want to cut a patch release for 0.15.3 without the unique tags changes, etc., in order to fix this behavior.

reltuk:
Got a permission denied on clones from doltremoteapi in local testing. Will investigate tomorrow. Not safe to merge.

reltuk:
Clone is still broken under this change, but that can technically be worked around by initing a new repo, `dolt remote add ...`ing the remote and `dolt fetch`ing.

PR #539, opened 2020-04-08
Db/dolt harness test

coffeegoddd:
Pretty simple tests, but I think are effective. Tested against the commit that initially caused the breaks and these tests failed... should also have caught the "unable to find table errors" that occurred during the harness fixing iteration process. Feels pretty _seems_ _fine_ to me! LMK

### DIFF
--- a/go/libraries/doltcore/sqle/logictest/dolt/doltharness.go
+++ b/go/libraries/doltcore/sqle/logictest/dolt/doltharness.go
@@ -57,6 +57,18 @@ func (h *DoltHarness) Init() error {
 		panic("Current directory must be a valid dolt repository")
 	}
 
+	return innerInit(h, dEnv)
+}
+
+func (h *DoltHarness) ExecuteStatement(statement string) error {
+	return executeStatement(h, statement)
+}
+
+func (h *DoltHarness) ExecuteQuery(statement string) (schema string, results []string, err error) {
+	return executeQuery(h, statement)
+}
+
+func innerInit(h *DoltHarness, dEnv *env.DoltEnv) error {
 	root, verr := commands.GetWorkingWithVErr(dEnv)
 	if verr != nil {
 		return verr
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/logictest/dolt/doltharness_test.go
```diff
@@ -0,0 +1,136 @@
...
+type queryTest struct {
+	query      string
+	expErr     error
+	expSchema  string
+	expResults []string
+}
+
+func TestDoltHarness(t *testing.T) {
+	home := "/home"
+	wd := filepath.Join(home, "/harnesstester")
```
zachmu, line 56: Never put test files in home. Use ioutil.TempFile

go/libraries/doltcore/sqle/logictest/dolt/doltharness_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	t.Run("should execute simple sql statements against Dolt", func(t *testing.T) {
+		h := &DoltHarness{}
+		fs := filesys.NewInMemFS([]string{}, nil, home)
+		dEnv := createTestEnvWithFS(fs, wd)
+
+		err := innerInit(h, dEnv)
+		assert.Equal(t, nil, err)
+
+		for _, test := range statementTests {
```
zachmu, line 108: Since you're already doing this below, no need to do it here as well

go/libraries/doltcore/sqle/logictest/dolt/doltharness.go
```diff
@@ -57,6 +57,18 @@ func (h *DoltHarness) Init() error {
 		panic("Current directory must be a valid dolt repository")
 	}
 
+	return innerInit(h, dEnv)
+}
+
+func (h *DoltHarness) ExecuteStatement(statement string) error {
+	return executeStatement(h, statement)
```
zachmu, line 64: No reason to dispatch to these identical private methods, is there? Just test the public methods.

bheni: +1

go/libraries/doltcore/sqle/logictest/dolt/doltharness_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/doltdb"
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/env"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+func createTestEnvWithFS(fs filesys.Filesys, workingDir string) *env.DoltEnv {
+	testHomeDirFunc := func() (string, error) { return workingDir, nil }
+	const name = "test mcgibbins"
+	const email = "bigfakeytester@fake.horse"
```
bheni, line 33: fake.horse lives!!!

PR #544, opened 2020-04-09
Zachmu/bheni patch

zachmu:
Fixed bug in parsing URLs of relative file paths.

### DIFF
--- a/go/libraries/utils/earl/url.go
+++ b/go/libraries/utils/earl/url.go
@@ -60,7 +60,7 @@ func Parse(urlStr string) (*url.URL, error) {
 	}
 
 	// url.parse doesn't handle file paths that begin with . correctly
-	if strings.HasPrefix(u.Host, "."){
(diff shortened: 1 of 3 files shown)


zachmu:
This works, but is kind of fragile. A better solution would be to make sure that all file paths are absolute (file:// URLs are supposed to be absolute paths), which they aren't across the codebase right now.

go/libraries/utils/earl/url.go
```diff
@@ -59,11 +59,17 @@ func Parse(urlStr string) (*url.URL, error) {
 		return nil, err
 	}
 
+	// url.parse doesn't handle file paths that begin with . correctly
+	if strings.HasPrefix(u.Host, ".") {
+		u.Path = u.Host + u.Path
+		u.Host = ""
+	}
+
```
bheni, line 67: I'd limit this to scheme=="file"

PR #546, opened 2020-04-09
/bats/1pk5col-strings.bats: Add skipped test for exporting to csv then reimporting

coffeegoddd:


timsehn:
While you are at it, can you you can move all the export tests from 1pk5col.bats into export-tables.bats?

timsehn:
We should merge this :-)

PR #547, opened 2020-04-09
Km/doc checkout bug

katiemcculloch:
Taylor brought a bad docs bug to my attention. If you had a modified doc, and then `dolt checkout <branch>` or `dolt checkout -b <branch>`, your local docs would be overwritten by the working root (your changes vanish).

The intended behavior is to keep the working set exactly as it is on `dolt checkout <branch>`, given there are no conflicts. Since your "working set" for docs is technically on the filesystem, and not on the roots, it was getting wiped. Now i'm pulling the unstagedDocDiffs from the filesystem, and excluding those when it comes time to `saveDocsOnCheckout`. 

Added bats test coverage too

### DIFF
--- a/bats/docs.bats
+++ b/bats/docs.bats
@@ -511,7 +511,7 @@ SQL
      dolt add LICENSE.md README.md
      dolt commit -m "initial license and readme commit"
      echo updated-readme > README.md
-     skip dolt checkout -b test-branch
+     dolt checkout -b test-branch
      run dolt status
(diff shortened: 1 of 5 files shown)


timsehn:
Check out remotes.bats for the bats tests :-)

katiemcculloch:
While my doc tests in remotes.bats are _incredibly_ chill, Brian is right. Working docs get stomped when you pull. I'll fix that too and add coverage

PR #548, opened 2020-04-09
read only version of branches table

bheni:


go/libraries/doltcore/sqle/branches_table.go
```diff
@@ -0,0 +1,146 @@
...
+// BranchesTableName
+func (dt *BranchesTable) String() string {
+	return BranchesTableName
+}
+
+// Schema is a sql.Table interface function that gets the sql.Schema of the branches system table
+func (dt *BranchesTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{Name: "name", Type: sql.Text, Source: BranchesTableName, PrimaryKey: true},
+		{Name: "hash", Type: sql.Text, Source: BranchesTableName, PrimaryKey: true},
```
zachmu, line 60: Should this be a pkey? Aren't branch names unique?

bheni: good catch.

Issue #533, opened 2020-04-08
Feature:  dolt sql --format=json


tomdottom:
Whilst the the ability to run `dolt sql-server` is great, it is a little overkill for simple usecases.

The `dolt sql --format=csv` command works quite well, but loses type information.

Adding the json output format would provide a convenient way to retrieve typed data from cli queries.

    dolt sql --format=json

However, json would still be lossy (eg. no decimal, datetime, or raw bytes representations).

Working from Python any of the following output formats would also be useful:
- py-marshal (compatible with Pythons builtin `marshal` module`
- py-pickle (compatible with Pythons builtin `pickle` module`

Or there are certainly is a [large selection of serialisation formats](https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats)  which might be more general.

zachmu:
Hi Tom,

Have you found doltpy yet?

https://github.com/liquidata-inc/doltpy

That's our recommended python integration solution. Let us know what it's missing!

We'll add json format to our backlog. Expect it in a future release.

tomdottom:
I tried `doltpy` but found running `dolt sql-server` via a subprocess unreliable.

I was able to manually run `dolt sql-server` and then just use a the `mysql-connector-python` library to query the server.

However, non of this would match the convenience of a `--format=json` option.

I noticed `go/libraries/doltcore/table/typed/json` so I'm guessing this might be in the works already.

zachmu:
Yes, we have a JSON export format, it's just not hooked up to the SQL command yet. You can use `dolt table export` with a .json file destination if getting an entire table at once is good enough for what you are trying to do.

I think we can get --format=json into Monday's release.

PR #551, opened 2020-04-10
Zachmu/sql json

zachmu:
Added JSON result output to sql command. Also fixed handling of NULL values in CSV output.

This fixes https://github.com/liquidata-inc/dolt/issues/533


timsehn:
Let's get this in the next release if we can.

PR #553, opened 2020-04-11
Avoid segfaulting when dashes in 'dolt command --help' are missing

Asgavar:
So, recently I noticed a funny bug in dolt: running e.g. `dolt blame help`
in a non-repo directory results in a segfault [1], while `dolt blame whatever`,
as expected, complains that the current dir is not a proper repository.
After taking a look around the code it became apparent that there is a divergence
in what is considered a 'help request' by the parser and what subcommand
handlers (i.e. the `isHelp()` function in dolt/cli/command.go) mark as such.

This patch makes an arbitrary decision that `dolt whatever help` is equivalent
to `dolt whatever -help`, which is in line with current isHelp() definition.

[1] $> dolt blame help
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x55c491049594]

[...]

### DIFF
--- a/go/libraries/utils/argparser/parser.go
+++ b/go/libraries/utils/argparser/parser.go
@@ -142,15 +142,15 @@ func (ap *ArgParser) Parse(args []string) (*ArgParseResults, error) {
 	for ; i < len(args); i++ {
 		arg := args[i]
 
-		if len(arg) == 0 || arg[0] != '-' || arg == "--" { // empty strings should get passed through like other naked words
-			list = append(list, arg)
-			continue
-		}
-
 		optName, value := splitOption(arg)
 
 		if optName == "help" || optName == "h" {
 			return nil, ErrHelp
+		}
+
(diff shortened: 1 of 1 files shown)


timsehn:
First, thank you for the pull request! We'll review shortly. 

I'll add a bats test for this.

timsehn:
https://github.com/liquidata-inc/dolt/pull/554

timsehn:
We're working on an open source contributor agreement and once we have that we'll merge this in.

Asgavar:
Alright, CLA read & signed ;]

bheni:
@Asgavar, as you identified differences between isHelp and the Parse function of the ArgParser were causing the problem. I found that changing the Parser to be like the isHelp function caused issues for some commands if you have a branch named help or a table named help. As a result I opted to change the isHelp logic instead in https://github.com/liquidata-inc/dolt/pull/576

Thanks!

Asgavar:
Roger that! Glad it got resolved regardless.

Issue #562, opened 2020-04-13
dolt does not respect the $HOME env var


tomdottom:
Steps to reproduce:

1. Start with working dolt configuration:

```
$ dolt config --list

user.creds = ...
user.email = ...
user.name = ...
```

Updating `HOME` env var to an empty dir still lists the same config:

```
$ export HOME=$(mktemp -d)
$ dolt config --list

user.creds = ...
user.email = ...
user.name = ...
```

timsehn:
Thanks for the report. Adding Brian.

reltuk:
This is surprising. Appears to come from using `os/user` here:

https://github.com/liquidata-inc/dolt/blob/master/go/libraries/doltcore/env/paths.go#L42

Brian will know about how intentional this is.

Should probably be something like:
```go
func GetCurrentUserHomeDir() (string, error) {
        if doltRootPath, ok := os.LookupEnv(doltRootPathEnvVar); ok && doltRootPath != "" {
                return doltRootPath, nil
        }
        if homeFromEnv, ok := os.LookupEnv("HOME"); ok && homeFromEnv != "" {
                return homeFromEnv, nil
        }
        if usr, err := user.Current(); err != nil {
                return "", err
        } else {
                return usr.HomeDir, nil
        }
}
```

PR #564, opened 2020-04-13
go/libraries/doltcore/env: paths.go: Consult HOME environment variable for home location before consulting os/user.

reltuk:


reltuk:
Pontential on https://github.com/liquidata-inc/dolt/issues/562. Wanted to get it through CI.

PR #566, opened 2020-04-13
go/libraries/doltcore/creds: jwk.go: Create new creds storage files with 0600 permissions.

reltuk:


go/libraries/doltcore/creds/jwk.go
```diff
@@ -109,7 +109,7 @@ func JWKCredsRead(rd io.Reader) (DoltCreds, error) {
 
 func JWKCredsWriteToDir(fs filesys.Filesys, dir string, dc DoltCreds) (string, error) {
 	outFile := filepath.Join(dir, dc.KeyIDBase32Str()+JWKFileExtension)
-	wr, err := fs.OpenForWrite(outFile)
+	wr, err := fs.OpenForWrite(outFile, 0600)
```
bheni, line 112: Do we want to use 0600 on windows?  Does this cause problems?

reltuk: bats tests still pass...the client can still read the created files.

Issue #561, opened 2020-04-13
Feature: `dolt creds import ...`


tomdottom:
Currently, as far as I can tell, the only way to set up dolt credentials programatically for automated systems is by creating the appropriate files:

- `~/.dolt/config_global.json`
- `~/.dolt/creds/abcxyz.jwk`

I might be preferable to use the cli to import and setup creds. Say:

**Create from env var**
```
echo "${creds_file}" | dolt creds import --use --user-name="me" --user-email="me@dom.com"
```

**Create from file**
```
cat /path/to/creds_file.jwk | dolt creds import --use --user-name="me" --user-email="me@dom.com"
# or
dolt creds import --use --user-name="me" --user-email="me@dom.com" --creds-file=/path/to/creds_file
```

Alternative env vars could be used to provide temporary config:

**Using env var overrides**
```
DOLT_CONFIG_USER_NAME="Me"
DOLT_CONFIG_USER_EMAIL="me@dom.com"
DOLT_CONFIG_CREDS_PATH="...."
```

Or:
```
DOLT_CONFIG_USER_NAME="Me"
DOLT_CONFIG_USER_EMAIL="me@dom.com"
DOLT_CONFIG_CREDS_JWK="{...}"
```

Lastly, a `dolt creds export ...` command may also be useful.

timsehn:
Creds are @reltuk 's baby. Assigning to him.

reltuk:
I love import/export. Great ask. We'll get on it.

For now, the easiest way to use an external cred is maybe to drop the `*.jwk` file into `~/.dolt/creds` and then run `dolt creds use KEY_ID|PUBLIC_KEY`. KEY_ID is the same as the filename without the `.jwk` extension.

So something like...
```sh
$ cp 0gb511i5f5pi4u6o5ihkcaech9112d5o1rdufqbtqkkru.jwk ~/.dolt/creds
$ dolt creds use 0gb511i5f5pi4u6o5ihkcaech9112d5o1rdufqbtqkkru
```

I was just trying out this workflow, and noticed that

```sh
$ cp 0gb511i5f5pi4u6o5ihkcaech9112d5o1rdufqbtqkkru.jwk ~/.dolt/creds
$ dolt login 0gb511i5f5pi4u6o5ihkcaech9112d5o1rdufqbtqkkru
```

[...]

PR #567, opened 2020-04-13
go/cmd/dolt: commands/login.go: Change behavior so dolt login with existing creds checks server for association before opening a browser.

reltuk:


reltuk:
Realized this behavior could be better when investigating #561

PR #572, opened 2020-04-14
More Types V2

Hydrocharged:
Have fun @zachmu


Hydrocharged:
Something worth mentioning here, as I didn't discover this until just now: new Noms types are fully incompatible with old versions of Dolt. I don't think that's a big deal here (considering `GEOMETRY` may be the last type to add that requires a new Noms type), but it is something to keep in mind for the future.

go/store/types/decimal.go
```diff
@@ -0,0 +1,109 @@
...
+func (v Decimal) Kind() NomsKind {
+	return DecimalKind
+}
+
+func (v Decimal) valueReadWriter() ValueReadWriter {
+	return nil
+}
+
+func (v Decimal) writeTo(w nomsWriter, nbf *NomsBinFormat) error {
+	encodedDecimal, err := decimal.Decimal(v).GobEncode()
```
zachmu, line 76: This dependency means that if shopspring ever makes a backwards incompatible change to their binary encoding, things break. I don't think this is super likely, but after you merge this you should add a bats test that will fail if it ever happens.

PR #574, opened 2020-04-14
go/store/nbs: Support clone on S3 by using chunkSource.reader.

reltuk:
Care is needed to ensure that said reader actually generates EOFs appropriately
and that we do not forward ReadAtOffset requests to tableReaderAt
implementations that will run us off the end of the file. Some of those
implementations throw errors in those cases, even if a similar error is going
to be caught further the call stack.


bats/aws-remotes.bats
```diff
@@ -42,7 +42,6 @@ skip_if_no_aws_tests() {
 
 @test "can clone an existing aws remote" {
     skip_if_no_aws_tests
-    skip "clone does not currently work against aws remotes; use init, remote add, fetch..."
```
reltuk, line 45: Yay!

go/store/nbs/store.go
```diff
@@ -928,27 +901,57 @@ func (nbs *NomsBlockStore) Sources(ctx context.Context) (hash.Hash, []TableFile,
...
 		}
+		tf := tableFile{
+			info: info,
+			open: func() (io.ReadCloser, error) {
+				r, err := cs.reader(context.TODO())
+				if r != nil {
+					return ioutil.NopCloser(r), err
+				} else {
+					return nil, err
+				}
```
bheni, line 926: more idiomatic to do:

```suggestion
				if err != nil {
					return nil, err
				}
				
				return ioutil.NopCloser(r), err
```

Is there a case where r could be nil and err would also be nil that you are worried about?  If so seems like that should also return an error.

PR #576, opened 2020-04-14
Help Fix

bheni:
As identified by Asgavar in https://github.com/liquidata-inc/dolt/pull/553 there is a segfault caused by differences in logic between isHelp and the Parse function of the ArgParser.  I found that changing the Parser to be like the isHelp function caused issues for some commands if you have a branch named help or a table named help.  As a result I opted to change the isHelp logic instead.

Thank you @Asgavar

### DIFF
--- a/bats/no-repo.bats
+++ b/bats/no-repo.bats
@@ -226,6 +226,6 @@ NOT_VALID_REPO_ERROR="The current directory is not a valid dolt repository."
 @test "all versions of help work outside a repository" {
     dolt checkout --help
     dolt checkout -help
-    skip "No dashes in front of help segfaults right now"
-    dolt checkout help
+    run dolt checkout help
+    [ "$status" -ne 0 ]
 }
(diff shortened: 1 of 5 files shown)


bats/no-repo.bats
```diff
@@ -226,6 +226,6 @@ NOT_VALID_REPO_ERROR="The current directory is not a valid dolt repository."
 @test "all versions of help work outside a repository" {
     dolt checkout --help
     dolt checkout -help
-    skip "No dashes in front of help segfaults right now"
-    dolt checkout help
+    run dolt checkout help
+    [ "$status" -ne 0 ]
```
timsehn, line 230: You can just remove the run and the `[ "$status" -ne 0 ]`. Running a command naked just checks for $status -eq 0

bheni: It won't be equal to 0.  It'll exit non-zero which is what I'm checking

timsehn:
My bad. LGTM

PR #605, opened 2020-04-20
{bats,libraries/doltcore/env}: Fix init stomp bug

katiemcculloch:
Dont stomp/save any docs if one or more docs already exist in repo


katiemcculloch:
Per offline discussion with @oscarbatori and @timsehn, I removed all the logic for automatically creating docs on `dolt init`. Users now must create the file themselves.

oscarbatori:
Thanks Katie for the quick turnaround, much appreciated.

PR #614, opened 2020-04-23
partial key iteration

bheni:
This is simple partial key iteration.  Certainly not optimal.

### DIFF
--- a/go/libraries/doltcore/sqle/expreval/expression_evaluator.go
+++ b/go/libraries/doltcore/sqle/expreval/expression_evaluator.go
@@ -173,9 +173,7 @@ func GetComparisonType(be expression.BinaryExpression) ([]*expression.GetField, 
 	}
 
 	var compType ComparisonType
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/expreval/expression_evaluator.go
```diff
@@ -173,16 +173,16 @@ func GetComparisonType(be expression.BinaryExpression) ([]*expression.GetField,
...
+	if len(variables) == 2 {
 		compType = VariableVariableCompare
 	} else if len(variables) == 1 {
 		if len(consts) == 1 {
 			compType = VariableConstCompare
 		} else if len(consts) > 1 {
 			compType = VariableInLiteralList
 		}
+	} else if len(consts) == 2 {
+		compType = ConstConstCompare
```
bheni, line 185: This is a bug fix. if an "IN" clause has exactly two values then this was broken.

Issue #624, opened 2020-04-25
A `dolt clone file://...` of an invalid file remote does not error.


reltuk:
It just gives you a new directory with a new repo in a partially initialized state. This is confusing.

```sh
aaronson@Aarons-MacBook-Pro dolt_clone % dolt clone file:/// this-is-not-what-should-happen
cloning file:///
0 of 0 chunks complete. 0 chunks being downloaded currently.
aaronson@Aarons-MacBook-Pro dolt_clone % cd this-is-not-what-should-happen 
aaronson@Aarons-MacBook-Pro this-is-not-what-should-happen % dolt status
On branch master
Untracked files:
  (use "dolt add <table|doc>" to include in what will be committed)
	new doc:        LICENSE.md
	new doc:        README.md
aaronson@Aarons-MacBook-Pro this-is-not-what-should-happen % dolt log
commit tf7hi83mk075rt47mjvgcmr2mq3sbr33
Author: Aaron Son <aaron@liquidata.co>
Date:   Sat Apr 25 16:30:52 -0700 2020

	Initialize data repository

```

See later correspondence on #623 for example.

timsehn:
We have a skipped bats test for cloning a bad remote:

https://github.com/liquidata-inc/dolt/blob/005f2c315cd8be44db87e1665674d3e5690c038f/bats/remotes.bats#L220

I'll add one for file-based remotes because it seems even more confusing.

timsehn:
Skipped bats test in: https://github.com/liquidata-inc/dolt/pull/625

VinaiRachakonda:
This bats is now unskipped and passing. Closing

PR #625, opened 2020-04-26
Added skipped bats test for bad cloning behavior

timsehn:


timsehn:
Skipped bats test for: https://github.com/liquidata-inc/dolt/issues/624

PR #626, opened 2020-04-27
fixing column ordering after deleted columns

andrew-wm-arthur:
The primary fix is in the [first commit](https://github.com/liquidata-inc/dolt/pull/626/commits/a7d9667509ea9b4e2a232b7a5275ca123167b313). I changed the the `TaggedValues` api to make this mistake harder to make in the future. 

Before this changes, Noms tuples were being reordered by a column deletion which was causing erroneous diffs. The same column values were stored in different orders, leading to different content addresses between root values. 

_Should we add a pipeline stage to `diff` to filter out rows that are unchanged?_

### DIFF
--- a/go/libraries/doltcore/schema/alterschema/dropcolumn.go
+++ b/go/libraries/doltcore/schema/alterschema/dropcolumn.go
@@ -121,7 +121,7 @@ func dropColumnValuesForTag(ctx context.Context, nbf *types.NomsBinFormat, newSc
 		_, found := tv[dropTag]
 		if found {
 			delete(tv, dropTag)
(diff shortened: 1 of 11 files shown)


go/libraries/doltcore/schema/alterschema/dropcolumn.go
```diff
@@ -121,7 +121,7 @@ func dropColumnValuesForTag(ctx context.Context, nbf *types.NomsBinFormat, newSc
 		_, found := tv[dropTag]
 		if found {
 			delete(tv, dropTag)
-			re.Set(k, tv.NomsTupleForTags(nbf, newSch.GetNonPKCols().Tags, false))
```
andrew-wm-arthur, line 124: This is the actual bug fix.

PR #634, opened 2020-04-28
Added bats test for sql using alias as a function in a select

timsehn:


zachmu:
Doesn't work in mysql either:

mysql> select count(*) as a, count(*) as b, a/b as percent from dual;
ERROR 1054 (42S22): Unknown column 'a' in 'field list'

timsehn:
Seems like it should work...oh well...i'll close and leave unmerged\.

PR #641, opened 2020-04-30
Zachmu/test refactor

zachmu:
Refactored the SQL engine tests to make it possible to validate schemas of result sets are as expected. This involved moving a lot of code around to avoid import cycles. Also unskipped or deleted a bunch of tests that had been skipped.


zachmu:
This looks huge, but it's almost 100% moved files. I deleted the files that were full of test queries and inlined those queries into the relevant test files. (This was an artifact from when we had two SQL engines that we were testing, no longer useful). I mostly want you to look at the structure of the package moves and see if they seem sensible to you.

go/libraries/doltcore/sql/sqltestutil/data.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Liquidata, Inc.
+// Copyright 2020 Liquidata, Inc.
```
andrew-wm-arthur, line 1: it seems like this is the only file left in this package, can we move it over to the `sqle` package as well?

go/libraries/doltcore/sqle/rows.go
```diff
@@ -78,7 +78,7 @@ func (itr *doltTableRowIter) Close() error {
 }
 
 // Returns a SQL row representation for the dolt row given.
-func doltRowToSqlRow(doltRow row.Row, sch schema.Schema) (sql.Row, error) {
+func DoltRowToSqlRow(doltRow row.Row, sch schema.Schema) (sql.Row, error) {
```
andrew-wm-arthur, line 81: this doesn't need to be public

go/libraries/doltcore/sqle/sqldelete_test.go
```diff
@@ -18,22 +18,160 @@ import (
...
 	. "github.com/liquidata-inc/dolt/go/libraries/doltcore/sql/sqltestutil"
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
 // Set to the name of a single test to run just that test, useful for debugging
 const singleDeleteQueryTest = "" //"Natural join with join clause"
 
-// Set to false to run tests known to be broken
-const skipBrokenDelete = true
+// Structure for a test of a insert query
```
andrew-wm-arthur, line 36: comment seems wrong

go/cmd/dolt/dolt.go
```diff
@@ -146,7 +146,7 @@ func runMain() int {
 	restoreIO := cli.InitIO()
 	defer restoreIO()
 
-	warnIfMaxFilesTooLow()
```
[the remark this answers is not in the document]
zachmu: Reverted thanks

go/libraries/doltcore/sql/sqltestutil/data.go
zachmu: Yeah, I was prevented by an import cycle. I will take another pass before I'm done.

go/libraries/doltcore/sqle/sqldelete_test.go
zachmu: fixed

go/libraries/doltcore/sqle/rows.go
zachmu: fixed

Issue #630, opened 2020-04-27
`dolt diff --summary` panic


timsehn:
```
Timothys-MacBook-Pro:us-supreme-court-cases timsehn$ dolt diff --summary cases
diff --dolt a/cases b/cases
--- a/cases @ fadg8pam2i2s6d7hjgen4f27afpti7jr
+++ b/cases @ qfo8554mkh95dh9cvkrllriai88qv3r0
prev size: 7425, new size: 7425, adds: 0, deletes: 0, modifications: 0panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x15c9294]

goroutine 1836 [running]:
github.com/liquidata-inc/dolt/go/store/types.Null.Equals(...)
	/Users/timsehn/liquidata/git/dolt/go/store/types/null_value.go:46
github.com/liquidata-inc/dolt/go/store/types.Tuple.CountDifferencesBetweenTupleFields(0x26e7c60, 0xc00004a080, 0x33f3560, 0xc0003b46f1, 0x179, 0x1b59, 0x0, 0x0, 0x0, 0x26e7c60, ...)
	/Users/timsehn/liquidata/git/dolt/go/store/types/tuple.go:463 +0x1e5
github.com/liquidata-inc/dolt/go/libraries/doltcore/diff.reportChanges(0xc0006e4960, 0xc000012600, 0x0, 0x0)
	/Users/timsehn/liquidata/git/dolt/go/libraries/doltcore/diff/diff_summary.go:67 +0x27b
github.com/liquidata-inc/dolt/go/libraries/doltcore/diff.Summary(0x26e7820, 0xc00023c240, 0xc000012600, 0x270fda0, 0xc00067c2d0, 0x270fda0, 0xc00016d720, 0x0, 0x0)
	/Users/timsehn/liquidata/git/dolt/go/libraries/doltcore/diff/diff_summary.go:47 +0x29b
github.com/liquidata-inc/dolt/go/cmd/dolt/commands.diffSummary.func1(0xc000012600, 0x26e7820, 0xc00023c240, 0x270fda0, 0xc00067c2d0, 0x270fda0, 0xc00016d720, 0xc00041d800)
	/Users/timsehn/liquidata/git/dolt/go/cmd/dolt/commands/diff.go:904 +0x9b
created by github.com/liquidata-inc/dolt/go/cmd/dolt/commands.diffSummary
[...]

PR #645, opened 2020-05-01
fix panic

andrew-wm-arthur:
fix of [630](https://github.com/liquidata-inc/dolt/issues/630)

### DIFF
--- a/bats/diff.bats
+++ b/bats/diff.bats
@@ -61,7 +61,7 @@ SQL
     [[ "$output" =~ "(4 Entries vs 3 Entries)" ]] || false
 }
 
-@test "diff summary comparing row with a deleted column and an added column" {
+@test "diff summary comparing row with a deleted cell and an added cell" {
     dolt sql <<SQL
 CREATE TABLE test (
   pk BIGINT NOT NULL COMMENT 'tag:0',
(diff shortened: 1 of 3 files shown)


andrew-wm-arthur:
BATS incoming

timsehn:
No bats? How would we have caught this?

andrew-wm-arthur:
The issue I haven't resolved is how NULL values were written to the non-primary key tuples. 
I haven't been able to reproduce this with any bats test.

timsehn:
NULL values can be in non-primary keys. I don't understand.

--Tim

andrew-wm-arthur:
We don't story NULL values, tho. They don't get written to the noms tuple. This panic was from a noms NULL value being compared to a non-existant column in the other side of the diff.

timsehn:
Weird. Here's what I did:

1. Run that import script to create the table with an all text schema.
2. Inspect the schema looking for better types
3. For columns I wanted different type I dropped the column, added with a
new type, same name different tag
4. Run the import again to add the values back to the column

andrew-wm-arthur:
OK I'm going to merge this. I'll open another issue to investigate the NULLS being written

andrew-wm-arthur:
https://github.com/liquidata-inc/dolt/issues/659

PR #649, opened 2020-05-01
Disallowed tag changes

Hydrocharged:


bats/sql.bats
```diff
@@ -462,6 +467,28 @@ SQL
     [[ "$output" =~ "unsupported feature: column types cannot be changed" ]] || false
 }
 
+@test "sql alter table modify column with no actual change" {
+    # this specifically tests a previous bug where we would get a name collision and fail
+    dolt sql -q "alter table one_pk modify column c5 bigint"
+}
```
zachmu, line 473: Should validate the schema is the same after this statement

bats/sql.bats
```diff
@@ -462,6 +467,28 @@ SQL
...
+
+@test "sql alter table change column with no actual change" {
+    # this specifically tests a previous bug where we would get a name collision and fail
+    dolt sql -q "alter table one_pk change column c5 c5 bigint"
```
zachmu, line 477: Should validate that the schema is unchanged

go/libraries/doltcore/sqle/tables.go
```diff
@@ -344,11 +344,11 @@ func (t *AlterableDoltTable) ModifyColumn(ctx *sql.Context, columnName string, c
 	}
 
 	tag := extractTag(column)
-	if tag == schema.InvalidTag {
-		tag = existingCol.Tag
+	if tag != existingCol.Tag && tag != schema.InvalidTag {
+		return errors.New("tag changes are disallowed")
```
zachmu, line 348: "cannot change the tag of an existing column"

PR #650, opened 2020-05-04
Secondary Indexes Pt. 1 Episode 2

Hydrocharged:
This PR is strictly the tests. All of the PR feedback and whatnot can be seen from the commit history found on the parent PR https://github.com/liquidata-inc/dolt/pull/495.

### DIFF
--- a/go/libraries/doltcore/sqle/integration_tests/stockmarket_test.go
+++ b/go/libraries/doltcore/sqle/integration_tests/stockmarket_test.go
@@ -71,6 +71,12 @@ CREATE TABLE join_result (
 );
 `
 
+const createIndexes = `
+CREATE INDEX idx_country ON daily_summary (Country);
+CREATE INDEX idx_ipoyear ON symbols (IPOYear);
+CREATE INDEX idx_country ON join_result (Country);
+`
+
 const insertRows = `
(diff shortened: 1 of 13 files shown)


go/libraries/doltcore/sqle/integration_tests/stockmarket_test.go
```diff
@@ -20161,6 +20167,40 @@ func TestInserts(t *testing.T) {
...
+	root, err = sqle.ExecuteSql(dEnv, root, createIndexes)
+	require.NoError(t, err)
+
+	root, err = sqle.ExecuteSql(dEnv, root, insertRows)
+	require.NoError(t, err)
+
+	table, _, err := root.GetTable(ctx, "daily_summary")
+	assert.NoError(t, err)
+	rowData, err := table.GetIndexRowData(ctx, "idx_country")
+	assert.NoError(t, err)
```
zachmu, line 20188: You should always require.NoError if the next line is going to cause a null pointer panic otherwise

Hydrocharged: That's a fantastic point, never even crossed my mind.

PR #665, opened 2020-05-07
Zachmu/case insensitive

zachmu:
Updated tests for new case-insensitivity improvements in go-mysql-server. Relies on https://github.com/liquidata-inc/go-mysql-server/pull/102, which is not on master yet.


timsehn:
We have a bunch of bats tests in `capital-letters.bats`. Check that out and add some smoke tests in there? Probably should be called `case-sensitivity.bats`.

zachmu:
Will do

PR #667, opened 2020-05-08
add warning if we expect error on os.Rename

bheni:
This is temporary.  Until a real fix is ready.


go/cmd/dolt/system_checks.go
```diff
@@ -18,3 +18,7 @@ package main
 
 func warnIfMaxFilesTooLow() {
 }
+
```
andrew-wm-arthur, line 21: what do these do?

andrew-wm-arthur: RIP windows

bheni:
Deleting this in favor of: https://github.com/liquidata-inc/dolt/pull/671

PR #668, opened 2020-05-08
Added partial key lookups for indexes and primary keys

Hydrocharged:


bats/index.bats
```diff
@@ -713,7 +713,57 @@ SQL
     [[ "${#lines[@]}" = "1" ]] || false
 }
 
-@test "index: SELECT = Full Match" {
+@test "index: SELECT = Full Match Primary Key" {
```
zachmu, line 716: Match primary key? There are tests here for partial match as well

Hydrocharged: I should probably rename this to just `SELECT = Primary Key`. I added the partial part last since it seemed like a good idea later lol

timsehn:
Daylon, push on this and let's get it merged. It's close. We can get it in for the Monday release and talk about supporting Secondary Indexes!

Hydrocharged:
Sorry about that Tim, I'd already finished this, but got caught up in a different PR and forgot to push and get it merged in!

PR #670, opened 2020-05-08
Update doc generation

oscarbatori:
Update Dolt doc generation process to generate a single file with CLI docs for new docs site.

The only interesting thing going on here is `dump_docs.go`, and I'd hardly call that interesting. I still don't love this implementation but it works for now.


go/cmd/dolt/cli/command.go
```diff
@@ -56,7 +55,7 @@ type Command interface {
 	// Exec executes the command
 	Exec(ctx context.Context, commandStr string, args []string, dEnv *env.DoltEnv) int
 	// CreateMarkdown creates a markdown file containing the helptext for the command at the given path
```
bheni, line 57: This comment is no longer accurate

go/cmd/dolt/cli/command.go
```diff
@@ -56,7 +55,7 @@ type Command interface {
...
-	CreateMarkdown(fs filesys.Filesys, path, commandStr string) error
```
bheni, line 59: I'd rename this function to GetCommandDocs or something

PR #671, opened 2020-05-08
Fix issue with renaming temp files across volumes

bheni:


bheni:
This replaces https://github.com/liquidata-inc/dolt/pull/667

go/store/util/tempfiles/temp_files.go
```diff
@@ -0,0 +1,66 @@
...
+// Clean makes a best effort attempt to delete all temp files created by calls to NewFile
+func (tfp *TempFileProviderAt) Clean() {
+	for _, filename := range tfp.filesCreated {
+		// best effort. ignore errors
+		_ = os.Remove(filename)
+	}
+}
+
+// MovableTemFile is an object that implements TempFileProvider that is used by the nbs to create temp files that
+// ultimately will be renamed.  It is important not to use this instance rather than using os.TempDir, or ioutil.TempFile
```
andrew-wm-arthur, line 63: "It is important to use this instance" right?

go/store/util/tempfiles/temp_files.go
```diff
@@ -0,0 +1,66 @@
...
+// directly as those may have errors executing a rename against if the volume the default temporary directory lives on
+// is different than the volume of the destination of the rename.
+var MovableTempFile TempFileProvider = NewTempFileProviderAt(os.TempDir())
```
andrew-wm-arthur, line 66: I think this would be better as `MovableTempFileProvider`

go/store/util/tempfiles/temp_files.go
bheni: fixed

go/store/util/tempfiles/temp_files.go
bheni: updated

PR #675, opened 2020-05-11
Infer typeinfo from data files

andrew-wm-arthur:
This extends the old schema inference logic to use numeric type widths and time types


andrew-wm-arthur:
@bheni One open question I had: should we use `UNSIGNED INTS` if a numeric column doesn't have negatives?

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -16,47 +16,71 @@ package actions
...
+	"github.com/liquidata-inc/dolt/go/libraries/utils/funcitr"
 	"github.com/liquidata-inc/dolt/go/libraries/utils/set"
 	"github.com/liquidata-inc/dolt/go/store/types"
 )
 
 // StrMapper is a simple interface for mapping a string to another string
 type StrMapper interface {
 	// Map maps a string to another string.  If a string is not in the mapping ok will be false, otherwise it is true.
-	Map(str string) (mappedStr string, ok bool)
+	MaybeMap(str string) string
```
bheni, line 39: Update the comment.  With the boolean return value removed, I don't know that I'd call this "MaybeMap" instead of just "Map"...  even if it just gets mapped to itself.

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -92,384 +118,435 @@ func InferSchemaFromTableReader(ctx context.Context, rd table.TableReadCloser, p
...
 }
 
-func (inf *inferrer) inferSchema() (schema.Schema, error) {
-	nonPkCols, _ := schema.NewColCollection()
-	pkCols, _ := schema.NewColCollection()
-
-	if inf.impArgs.Update {
-		nonPkCols = inf.impArgs.ExistingSch.GetNonPKCols()
-		pkCols = inf.impArgs.ExistingSch.GetPKCols()
+func (inf *inferrer) inferSchema(ctx context.Context, root *doltdb.RootValue) (schema.Schema, error) {
```
bheni, line 151: This function is a bit of a beast.  Might be worth breaking up.

bheni:
@andrew-wm-arthur I would pretty much always opt for int over uint. It's rare that uint need that extra bit, and more likely that you'll have columns that need to support negative values.  The thing you get is a bit of error checking in that if you want it to be only uints, then you guarantee somebody can't stick a negative value in there... But at that point I think you are more of a power user that wants fine grained control and we have the ability to manage that yourself already.

PR #679, opened 2020-05-13
Index performance improvements

Hydrocharged:


go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -126,12 +138,15 @@ func (di *doltIndex) Expressions() []string {
...
+	tpl, err := di.keysToTuple(keys, false)
+	if err != nil {
+		return nil, err
 	}
-	return di.keysToIter(di.keysToExpr(keys, func(left sql.Expression, right sql.Expression) sql.Expression {
-		return expression.NewEquals(left, right)
-	}))
+	nbf := di.indexRowData.Format()
+	readRange := &noms.ReadRange{Start: tpl, Inclusive: true, Reverse: false, Check: func(tuple types.Tuple) (bool, error) {
+		return tuple.StartsWith(nbf, tpl)
```
bheni, line 147: love it

go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -151,31 +166,33 @@ func (di *doltIndex) Table() string {
...
+	for i, col := range di.cols {
+		val, err := col.TypeInfo.ConvertValueToNomsValue(keys[i])
+		if err != nil {
+			return types.EmptyTuple(nbf), err
+		}
+		vals = append(vals, types.Uint(col.Tag), val)
+	}
+	// In the case of possible partial keys, we may need to match at the beginning or end for matched values, so we
+	// append a tag that is beyond the allowed maximum. This will be ignored if it's a full key and not a partial key.
+	if appendMaxValue {
```
zachmu, line 188: I don't think this is right, Brian should take a look.

If you're trying to get a start value for iterating over the map, then you need a start tuple which has the same number of keys as the schema does. You're only including the number of indexed columns, plus maybe one optional value for the first column of the primary key. What you have now only works if the primary keys are all >= the zero value for the underlying noms type. If you had a table with negative primary keys, I believe this code would start iterating in the middle of the map (where keys >= 0). I'm not positive, Brian should say more.

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,637 @@
...
+CREATE TABLE twopk (
+  pk1 BIGINT,
+  pk2 BIGINT,
+  v1 BIGINT,
+  v2 BIGINT,
+  PRIMARY KEY (pk1,pk2)
+);
+CREATE INDEX idx_v1 ON onepk(v1);
+CREATE INDEX idx_v2v1 ON twopk(v2, v1);
+INSERT INTO onepk VALUES (1, 1, 1), (2, 1, 2), (3, 3, 3), (4, 4, 3);
```
zachmu, line 546: This is a great start for these tests, but it should be expanded a bit:

1) Negative numbers
2) Other types (strings and dates would be good for a start).

1) should be part of this PR, but you could do a followup for 2).

go/store/types/tuple.go
```diff
@@ -497,6 +497,32 @@ func (t Tuple) fieldsToMap() (map[Value]Value, error) {
 	return valMap, nil
 }
 
+func (t Tuple) StartsWith(nbf *NomsBinFormat, otherTuple Tuple) (bool, error) {
```
bheni, line 500: Does this work:

```
func (t Tuple) StartsWith(otherTuple Tuple) bool {
	return bytes.HasPrefix(t.buff, otherTuple.buff)
}
```

bheni: Should add tests to tuple_test.go for StartsWith

go/libraries/doltcore/sqle/dolt_index.go
bheni: So appendMaxValue looks like it's only true for:

DescendGreater - Should be good because it skips the current key and points to position before next possible key.
DescendLessOrEquald - Should be good because it is going past the last key, but is always before the next key.  Will work for partial and full keys.
AscendRange - Used for the end of the range which should be good assuming range is inclusive.  In this case it is the same as above.

bheni: Double check your test cases to make sure Zach's concerns are covered.

bheni:
Talked to Zach about this.  I guess the engine doesn't actually care about rows being returned in any particular order.

go/store/types/tuple.go
Hydrocharged: After some more thought, `HasPrefix` should actually work just fine. So I chose to not use it originally as I thought that you could potentially have a byte or string array as a Value in the Tuple, that could be a byte-wise match up to some point. For example:
```go
tupleA := NewTuple(nbf, types.Uint(1), types.InlineBlob{2, 3}, types.Uint(2), ...)
tupleB := NewTuple(nbf, types.Uint(1), types.InlineBlob{2, 3, ...})
```
So I thought that `tupleB` could have some end up having some value that would be equivalent to having `types.Uint(2)` encoded after as far as the raw bytes go, but after thinking again it wouldn't hold up, since all variable-length types also encode their length, so they would fail the prefix check since the length would be different. My worry isn't a real issue in practice.

Hydrocharged: `bytes.HasPrefix` should also be a good deal faster too so it's a win for everyone.

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,637 @@
...
+	require.NoError(t, err)
+
+	var readRows []sql.Row
+	var nextRow sql.Row
+	for nextRow, err = indexIter.Next(); err == nil; nextRow, err = indexIter.Next() {
+		readRows = append(readRows, nextRow)
+	}
+	require.Equal(t, io.EOF, err)
+
+	assert.ElementsMatch(t, convertSqlRowToInt64(expectedRows), readRows)
```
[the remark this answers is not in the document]
Hydrocharged: Order is deterministic at this point. I went this way because of the earlier conversations regarding the `ORDER BY` stuff, but I guess it's more useful to assert the expected order here too since that's further up the chain.

Hydrocharged:
> 
> 
> Talked to Zach about this. I guess the engine doesn't actually care about rows being returned in any particular order.

Yeah, from what I could dig through, these just seem like some names that they came up with. It doesn't appear like they care about the order at all above the call sight, I was going to rename all of this stuff anyway at some point in the future, as their `BETWEEN` support is fairly weird and requires mergeable indexes and other unnecessary stuff.

Hydrocharged:
@zachmu @bheni Alright guys, added a few hundred more tests and addressed the feedback! Check it out, it's all in the 2nd commit.

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,1170 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/doltdb"
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/dtestutils"
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/schema"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+type doltIndexTestType struct {
```
zachmu, line 34: s/TestType/TestCase/

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,1170 @@
...
+	indexName    string
+	keys         []interface{}
+	expectedRows []sql.Row
+}
+
+type doltIndexBetweenTestType struct {
```
zachmu, line 40: same comment

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,1170 @@
...
+			[]sql.Row{{2, 1, 4, 4}},
+		},
+	}
+
+	for _, varietyTest := range varietyTests {
+		tests = append(tests, doltIndexTestType{
+			varietyTest.indexName,
+			varietyTest.belowfirstValue,
+			[]sql.Row{
+				{int32(-3), uint64(1), forceParseTime(t, "2020-05-14 12:00:00"), "-3.30000", "a", -3.3, "a", "-00:03:03", "a", int16(1980)},
```
zachmu, line 336: Suggest extracting vars for all these variety table rows (typesTableRow1, typesTableRow2, etc) for all these tests. Would be easier to read.

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -0,0 +1,1170 @@
...
+}
+
+type doltIndexBetweenTestType struct {
+	indexName          string
+	greaterThanOrEqual []interface{}
+	lessThanOrEqual    []interface{}
+	expectedRows       []sql.Row
+}
+
+var varietyTests = []struct {
```
zachmu, line 47: Would call these typesTests

go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -31,73 +32,84 @@ type DoltIndex interface {
...
-	if len(di.cols) != len(keys) {
-		return nil, errors.New("keys must specify all columns for index")
+	tpl, err := di.keysToTuple(keys, true)
+	if err != nil {
+		return nil, err
 	}
-	return di.keysToIter(di.keysToExpr(keys, func(left sql.Expression, right sql.Expression) sql.Expression {
-		return expression.NewGreaterThan(left, right)
-	}))
+	readRange := &noms.ReadRange{Start: tpl, Inclusive: true, Reverse: false, Check: func(tuple types.Tuple) (bool, error) {
```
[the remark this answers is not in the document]
Hydrocharged: This is already tested for by the many tests in `dolt_index_test.go` and even the bats tests will catch it.

Brian said it best on another comment here in the PR:
> So appendMaxValue looks like it's only true for:
> 
> DescendGreater - Should be good because it skips the current key and points to position before next possible key. Will work for partial and full keys.
> DescendLessOrEquald - Should be good because it is going past the last key and points to position before next possible key. Will work for partial and full keys.
> AscendRange - Used for the end of the range which should be good assuming range is inclusive. In this case it is the same as above.

PR #684, opened 2020-05-18
proto: remoatesapi: GetRepoMetadataResponse: Add repository_size field for communicating approximate repository size.

reltuk:


reltuk:
Open to suggestions on naming and comments :).

PR #687, opened 2020-05-19
go/{store/nbs,libraries/doltcore/remotestorage}: Expose repository size as Size() on TableFileSource.

reltuk:


go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -461,6 +463,22 @@ func (dcs *DoltChunkStore) Rebase(ctx context.Context) error {
 		return NewRpcError(err, "Rebase", dcs.host, req)
 	}
 
+	mdReq := &remotesapi.GetRepoMetadataRequest{
```
reltuk, line 466: Maybe factor into a `func (dcs *DoltChunkStore) refreshRepoMetadata(ctx context.Context) error`?

go/store/datas/pull_test.go
```diff
@@ -421,6 +421,10 @@ func (ttfs *TestTableFileStore) Sources(ctx context.Context) (hash.Hash, []nbs.T
 	return ttfs.root, tblFiles, nil
 }
 
+func (ttfs *TestTableFileStore) Size(ctx context.Context) (uint64, error) {
```
reltuk, line 424: This fine or need to do better?

go/store/nbs/store.go
```diff
@@ -932,6 +932,44 @@ func (nbs *NomsBlockStore) Sources(ctx context.Context) (hash.Hash, []TableFile,
 	return contents.GetRoot(), tableFiles, nil
 }
 
+func (nbs *NomsBlockStore) Size(ctx context.Context) (uint64, error) {
```
reltuk, line 935: Thoughts on testing?

PR #693, opened 2020-05-20
optimize diff and hist tables

bheni:


go/libraries/doltcore/doltdb/commit_itr.go
```diff
@@ -163,114 +162,39 @@ func hashToCommit(ctx context.Context, vrw types.ValueReadWriter, h hash.Hash) (
...
-
-	return h, cm, nil
-}
-
-type TimeRange struct {
-	Min time.Time
-	Max time.Time
-}
+// CommitFilter is a function that returns true if a commit should be filtered out, and false if it should be kept
+type CommitFilter func(context.Context, hash.Hash, *Commit) (filterOut bool, err error)
```
zachmu, line 166: The logical case is inverted on this. Predicate functions should return true for a match, not true for exclusion. "(accept bool, err error)"

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -70,25 +74,21 @@ func NewDiffTable(ctx *sql.Context, dbName, tblName string) (*DiffTable, error)
...
 
-	cmItr := doltdb.CommitItrForRoots(ddb, rootCmt)
-	ss, partitions, err := calcSuperSchemaAndPartitions(ctx, cmItr, workingRoot, tblName)
+	ss, err := calcSuperSchema(ctx, workingRoot, tblName)
 
 	if err != nil {
 		return nil, err
 	}
 
+	/**/
```
zachmu, line 90: remove

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -341,178 +388,193 @@ func (dp diffPartition) Key() []byte {
 	return []byte(dp.toName + dp.fromName)
 }
 
-var _ sql.PartitionIter = &diffPartitions{}
+type partitionFilterFunc func(*sql.Context, diffPartition) (filterOut bool, err error)
```
zachmu, line 391: Same comment about predicate funcs

go/libraries/doltcore/sqle/diff_table.go
```diff
@@ -341,178 +388,193 @@ func (dp diffPartition) Key() []byte {
...
 
 		root, err := cm.GetRootValue()
 
 		if err != nil {
-			return nil, nil, err
+			return nil, err
 		}
 
-		tbl, ok, err := root.GetTable(ctx, exactName)
+		tbl, _, _, err := root.GetTableInsensitive(dp.ctx, dp.tblName)
```
zachmu, line 556: What happens if the table doesn't exist in this root? Do you have a test for this case?

Also, this doesn't track tables across renames, which would be nice to do. Might add a TODO.

bheni: If a table doesn't exist it is equal to nil.  This is hit in every test case you write because the table is never there for the entire history of the db.

PR #695, opened 2020-05-20
Better arg parsing

andrew-wm-arthur:


bats/arg-parsing.bats
```diff
@@ -12,29 +12,29 @@ teardown() {
...
-    [ $status -eq 0 ]
+    [ $status -eq 0 ] || echo $output
     [[ "$output" =~ "this-should-work" ]] || false
     dolt checkout master
     dolt branch -d "this-should-work"
 
     dolt checkout --b "this-should-work"
     run dolt branch
-    [ $status -eq 0 ]
+    [ $status -eq 0 ] || echo $output
```
timsehn, line 29: Probably should get rid ofthose echos

PR #697, opened 2020-05-21
support HEAD and ancestor spec syntax

bheni:


go/libraries/doltcore/doltdb/commit.go
```diff
@@ -264,3 +264,13 @@ func (c *Commit) CanFastReverseTo(ctx context.Context, new *Commit) (bool, error
 
 	return false, nil
 }
+
+func (c *Commit) WalkAncestorSpec(ctx context.Context, as *AncestorSpec) (*Commit, error) {
```
zachmu, line 268: I think these methods use the wrong verb. Typically, I reserve "walk" for cases of iteration. I would expect a method named "walk*" to accept a func that it calls on every node while traversing a databaset.

Suggest renaming this one and its lowercase equivalent to RetrieveAncestor / GetAncestor / ResolveAncestor or similar.

PR #700, opened 2020-05-21
swish

andrew-wm-arthur:


bats/sql-batch.bats
```diff
@@ -24,15 +24,18 @@ teardown() {
...
-    skip "Batch does not properly insert a newline after output"
-    NEWLINE_REGEX = "Rows deleted: 0\n"
-    [[ "$output" =~ $NEWLINE_REGEX  ]] || false
     dolt sql -batch -q "insert into test values (1,0,0,0,0,0)"
     run dolt sql -b -q "select * from test" 
     [ "$status" -eq 0 ]
     [[ "$output" =~ " 0 " ]] || false
     [[ "$output" =~ " 1 " ]] || false
-    [[ "$output" =~ " Rows inserted " ]] || false
+
```
andrew-wm-arthur, line 32: this test fails as expected if the `\n` is removed from this [line](https://github.com/liquidata-inc/dolt/blob/cc1cf590aac8c9fc5e79a3f060f8ede07f0ced93/go/cmd/dolt/commands/sql.go#L973)

PR #703, opened 2020-05-22
Zachmu/indexes

zachmu:
Killed off index driver and let tables declare indexes natively.


go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -141,26 +145,31 @@ func (di *doltIndex) Get(keys ...interface{}) (sql.IndexLookup, error) {
...
 func (di *doltIndex) Schema() schema.Schema {
 	return di.tableSch
 }
 
+// Table implements sql.Index
 func (di *doltIndex) Table() string {
 	return di.tableName
 }
 
+// TableData returns the map of table data for this index (the map of the index table, not the target table)
```
Hydrocharged, line 172: The comment is backwards, `di.tableData` is the target table's data. `di.indexRowData` is the map of the index table.

go/libraries/doltcore/sqle/indexed_dolt_table.go
```diff
@@ -24,12 +24,16 @@ type IndexedDoltTable struct {
 	indexLookup *doltIndexLookup
 }
 
-func (idt *IndexedDoltTable) WithIndexLookup(lookup sql.IndexLookup) sql.Table {
-	return idt.table.WithIndexLookup(lookup)
+var _ sql.IndexedTable = (*IndexedDoltTable)(nil)
+
+func (idt *IndexedDoltTable) GetIndexes(ctx *sql.Context) ([]sql.Index, error) {
+	// TODO: move index driver implementation here
+	panic("implement me")
```
Hydrocharged, line 31: Wouldn't this be the same?
`return idt.table.GetIndexes(ctx)`?

PR #708, opened 2020-05-27
Extra BATS for `table import` & `schema import`

andrew-wm-arthur:


bats/schema-import.bats
```diff
@@ -3,14 +3,39 @@ load $BATS_TEST_DIRNAME/helper/common.bash
...
+DELIM
+
+    cat <<DELIM > abc-xyz.csv
+pk,a,b,c,x,y,z
+0, red,  1.1, true,  green,  3.14, -1
+1, blue, 2.2, false, yellow, 2.71, -2
+DELIM
 }
 
 teardown() {
```
timsehn, line 33: Should you blow away the files in teardown?

andrew-wm-arthur: the whole directory for each test gets deleted when it completes

PR #710, opened 2020-05-28
New TableEditor, thread-safety, and table import

Hydrocharged:
So I created a new `TableEditor` under `doltdb`, similar to the `IndexEditor`. I renamed the old editor to `sqlTableEditor`, and it's now using `TableEditor` underneath. It's just to comply with the SQL interfaces.

Perhaps one of the most striking changes are the synchronization changes. The motivation was from a comment from @zachmu , which implied that some of the write paths are parallelized for performance, and the old editor was not thread-safe by any means. In order to support these new write paths, it was necessary to make `TableEditor` thread-safe. Because of this, `IndexEditor` also received the same changes. I verified that there was a lack of data races using `go test -race`, and also ran the concurrency tests with thousands more iterations and edits (they're lower now so that we're not spending an hour in Jenkins). Just to verify that my tests also worked, I made temporary intentional changes that would allow for data races, and the tests caught them, and `go test -race` also caught them.

In addition to the above changes, `dolt table import` and `dolt table cp` both use the `TableEditor` now. Due to the concerns mentioned in the previous paragraph, I decided to benchmark a few metrics to get an idea of the performance impact of these two major changes.

To measure just the impact of the threading changes, I imported 1,000,000 rows through SQL and timed the results:
```
Before: 67.074s
After:  67.363s
```
The result is close enough that it's within the margin of error, so it's safe to say that new `TableEditor` is just as performant as the old implementation.

The `table import` is a different story though. This time, I imported a 10,000,000 row `.psv`, with the results being:
```
Before: 53.521s
After:  107.816s
```
It takes roughly twice as long now, which is slower, but not tragically so. Primarily, the difference in speed is that the old code just wrote directly to the edit accumulator, where as the `TableEditor` does a ton more bookkeeping, being a generalized editor.


Hydrocharged:
Even if the pipeline code does not currently use multiple threads for writing to the `TableEditor`, we'll still have to make it thread-safe at some point in the future. I'd say these should have been thread-safe from the beginning, and not having them so does limit their potential usage.

I kept the lock implementations simple for the sake of clarity as well. We can go back over them later whenever do a performance overhaul of the editors. For `TableEditor`, it's all just `Lock()/defer Unlock()`, so it's not hard to trace.

For `IndexEditor`, I'm not using the `defer Unlock()` pattern so much, as the `Lock()/Unlock()` just surrounds portions of code that are 1-3 lines long and don't have reasons to panic, so it's also straightforward in its implementation.

Hydrocharged:
Apparently Jenkins is having issues with `doltpy`, so ignore the failure, as it passed everything.

zachmu:
I haven't looked at this yet, but wanted to address the performance regression.

A 50% performance penalty in CSV import is maybe barely permissible, but should be avoided if at all possible, even if it makes the code more complicated. This is the front door for most customers, especially those with large pre-existing data sets. It needs to be as fast as possible. I think 10M rows in 2 min is probably acceptable, but obv faster is better. It's worth doing some work profiling this to see where the time is going for CSV, and which parts could be avoided. If there are easy wins to be had here, let's take them.

I agree substantially about the thread safety concern (better to assume it will be used in a non-thread-safe way). Probably worth benchmarking with and without the locks to see if it costs us performance, though.

Hydrocharged:
I'll profile and see what I can come up with. I tried some simple multithreaded changes in the import code and only got around 2 seconds of speed increase for a lot more complexity, so I reverted that and shipped what is here.

> Probably worth benchmarking with and without the locks to see if it costs us performance, though.

Darn it @zachmu it's right there in my initial post lol, I'll just copy it here too:

> To measure just the impact of the threading changes, I imported 1,000,000 rows through SQL and timed the results:
> ```
> Before: 67.074s
> After:  67.363s
> ```
> The result is close enough that it's within the margin of error, so it's safe to say that new > `TableEditor` is just as performant as the old implementation.

zachmu:
Brian, this is about a 2x performance regression for import. I want you to see if you have any ideas on how to speed it up. Daylon is in the middle of profiling it, so he might have some low hanging fruit too.

Hydrocharged:
So from profiling...well I didn't find anything that can be considered "low hanging fruit". It looks like a significant chunk of time spent is on garbage collection, and I can't find anything quick and simple that can reduce that. Besides that though, the load is fairly spread out.

Performance of this path is something that I can really dig into and try to optimize after foreign keys are in. Since everything will be using this new path, it would lead to speed increases across the entire product.

PR #714, opened 2020-05-28
simple conflicts table

bheni:


go/libraries/doltcore/sqle/conflicts_table.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Schema is a sql.Table interface function that gets the sql.Schema of the log system table.
+func (dt *ConflictsTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{Name: "table", Type: sql.Text, Source: doltdb.ConflictsTableName, PrimaryKey: true},
+		{Name: "num_conflicts", Type: sql.Uint64, Source: doltdb.ConflictsTableName, PrimaryKey: false},
+	}
+}
+
+type conflictsPartition struct {
```
bheni, line 64: The original design for this was to show all the conflicts through the dolt_conflicts table... but creating the schema for that table sucked.  I may end up going back to that at some point so I didn't kill the partitions, but they may get killed in the future.

The new design will have dolt_conflicts_TABLENAME style tables in addition to this one.

zachmu: This seems fine. I don't know how else you would make a global conflicts table.

bheni:
@zachmu There is a bats test for now.  But I will revisit testing once I have the table specific conflicts table done.

bheni:
added an additional bats test.  Will revisit putting in a go test once I have SQL Merge function in.

PR #715, opened 2020-05-28
Special quoting logic for NULLS & ""

andrew-wm-arthur:


go/cmd/dolt/commands/sql.go
```diff
@@ -1130,10 +1130,6 @@ func (se *sqlEngine) prettyPrintResults(ctx context.Context, sqlSch sql.Schema,
 
 	// Parts of the pipeline depend on the output format, such as how we print null values and whether we pad strings.
 	switch se.resultFormat {
-	case formatCsv:
```
zachmu, line 1133: Was this just unnecessary? Already not printing nil values?

go/libraries/doltcore/table/untyped/csv/writer.go
```diff
@@ -30,14 +34,15 @@ import (
 
 // WriteBufSize is the size of the buffer used when writing a csv file.  It is set at the package level and all
```
zachmu, line 35: Update comment with new name

go/libraries/doltcore/table/untyped/csv/writer.go
```diff
@@ -97,42 +105,150 @@ func (csvw *CSVWriter) GetSchema() schema.Schema {
...
+		csvw.wr = nil
 		return errCl
 	} else {
 		return errors.New("Already closed.")
 	}
 }
+
+// write is directly copied from csv.Writer.Write() with the addition of the `isNull []bool` parameter
+// this method has been adapted for Dolt's special quoting logic, ie `10,,""` -> (10,NULL,"")
+func (csvw *CSVWriter) write(record []string, isNull []bool) error {
```
zachmu, line 154: Small nit: just make this take []*string and pass nils for the nil strings. No need for the second array.

go/cmd/dolt/commands/sql.go
andrew-wm-arthur: `NULL`/`nil` printing is handled by the sink now. We need to eliminate this transform to be able to disambiguate between `NULL` values and empty strings

PR #721, opened 2020-06-01
Last write paths to TableEditor

Hydrocharged:
`Resolve` and `Merge` were the last two write paths that didn't use the new `TableEditor` yet, so here they are!

### DIFF
--- a/go/libraries/doltcore/merge/merge.go
+++ b/go/libraries/doltcore/merge/merge.go
@@ -28,7 +28,6 @@ import (
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/env"
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/row"
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/schema"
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/merge/merge.go
```diff
@@ -425,40 +429,75 @@ func mergeTableData(ctx context.Context, sch schema.Schema, rows, mergeRows, anc
...
-func applyChange(me *types.MapEditor, stats *MergeStats, change types.ValueChanged) {
+func applyChange(ctx context.Context, tableEditor *doltdb.TableEditor, rowData types.Map, sch schema.Schema, stats *MergeStats, change types.ValueChanged) error {
 	switch change.ChangeType {
 	case types.DiffChangeAdded:
+		newRow, err := row.FromNoms(sch, change.Key.(types.Tuple), change.NewValue.(types.Tuple))
+		if err != nil {
+			return err
+		}
+		val, ok, err := rowData.MaybeGet(ctx, change.Key)
+		if err == nil && ok {
```
zachmu, line 462: You should handle err != nil before going into insert / update logic

PR #724, opened 2020-06-02
Bh/conflicts table3

bheni:


bats/sql-conflicts.bats
```diff
@@ -0,0 +1,190 @@
...
+  dolt sql -q "DELETE from dolt_conflicts_one_pk WHERE our_pk1 = 0"
+  dolt sql -q "DELETE from dolt_conflicts_two_pk WHERE our_pk1 = 0 and our_pk2 = 0"
+
+  EXPECTED=$( echo -e "table,num_conflicts")
+  run dolt sql -r csv -q "SELECT * FROM dolt_conflicts"
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "$EXPECTED" ]] || false
+}
+
+@test "delete modify conflict" {
```
zachmu, line 138: How does this test differ from the one above? Not obvious to me

go/libraries/doltcore/merge/conflict_source.go
```diff
@@ -16,215 +16,167 @@ package merge
...
+
+// NextConflict can be called successively to retrieve the conflicts in a table.  Once all conflicts have been returned
+// io.EOF will be returned in the error field.  This can be used in a pipeline, or to iterate through all the conflicts
+// in a table.
+func (cr *ConflictReader) NextConflict(ctx context.Context) (row.Row, pipeline.ImmutableProperties, error) {
+	key, value, err := cr.confItr.Next(ctx)
 
 	if err != nil {
-		return nil, err
+		return nil, pipeline.ImmutableProperties{}, err
```
zachmu, line 113: NoProps?

zachmu: Here and elsewhere

go/libraries/doltcore/sqle/conflicts_tables.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/liquidata-inc/dolt/go/store/types"
+
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/doltdb"
+)
+
+const ()
+
+var _ sql.Table = ConflictsTable{}
+
+type ConflictsTable struct {
```
zachmu, line 30: You should write some unit tests verifying that you can't insert or update rows in this table. There are a bunch of these for other system tables already. E.g. TestExecuteUpdateSystemTables in sqlupdate_test.go

bats/sql-conflicts.bats
bheni: this one does deletes of the rows in the feature branch.  The other one updates in the different branch.

go/libraries/doltcore/merge/conflict_source.go
bheni: done

bheni:
For a table named "test" primary key column pk and columns c1 and c2, resolving and taking theirs looks like:

```
REPLACE INTO test (pk, c1, c2)
SELECT their_pk, their_c1, their_c2 FROM dolt_conflicts_test WHERE their_pk IS NOT NULL;

DELETE FROM test 
WHERE pk in (
	SELECT base_pk
	FROM dolt_conflicts_test
	WHERE their_pk IS NULL
);

DELETE FROM dolt_conflicts_test;
```

PR #725, opened 2020-06-03
Andy/diff schema changes

andrew-wm-arthur:
- Refactored `dolt diff` to match tables between roots using schema instead of table name, allowing us to track table renames
- Fixed command line parsing to allow for `dolt diff head^ head tbl1 tbl2 tbl3`
- Switched argument order from `dolt diff to_commit from_commit` to `dolt diff from_commit to_commit` to match Git. (I also changed most instances of `(newArg,oldArg)` to `(fromArg,toArg)` to try to avoid this confusion in the future

### DIFF
--- a/go/cmd/dolt/commands/diff.go
+++ b/go/cmd/dolt/commands/diff.go
@@ -309,7 +309,7 @@ func maybeResolve(ctx context.Context, dEnv *env.DoltEnv, spec string) (*doltdb.
 func diffUserTables(ctx context.Context, fromRoot, toRoot *doltdb.RootValue, dArgs *diffArgs) (verr errhand.VerboseError) {
 	var err error
 
-	tableDeltas, err := diff.GetUserTableDeltas(ctx, fromRoot, toRoot)
+	tableDeltas, err := diff.GetTableDeltas(ctx, fromRoot, toRoot)
 	if err != nil {
 		return errhand.BuildDError("error: unable to diff tables").AddCause(err).Build()
 	}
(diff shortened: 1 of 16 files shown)


go/cmd/dolt/commands/ls.go
```diff
@@ -115,6 +115,35 @@ func (cmd LsCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
 	return HandleVErrAndExitCode(verr, usage)
 }
 
+func getRootForCommitSpecStr(ctx context.Context, csStr string, dEnv *env.DoltEnv) (string, *doltdb.RootValue, errhand.VerboseError) {
```
andrew-wm-arthur, line 118: this is a straight copy/paste, it was only used in `dolt ls`

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -558,6 +557,43 @@ func (root *RootValue) HasConflicts(ctx context.Context) (bool, error) {
 	return len(cnfTbls) > 0, nil
 }
 
+func (root *RootValue) IterTables(ctx context.Context, cb func(name string, table *Table) (stop bool, err error)) error {
```
bheni, line 560: profileration of "stop" instead of "continue" return variable.  Should figure out if we actually want to fixk this.

go/cmd/dolt/commands/diff.go
```diff
@@ -103,6 +103,8 @@ In order to filter which diffs are displayed {{.EmphasisLeft}}--where key=value{
 type diffArgs struct {
 	diffParts  diffPart
 	diffOutput diffOutput
+	tableSet   *set.StrSet
+	docSet     *set.StrSet
```
bheni, line 107: Going through this file, all the special casing for docs is pretty painful.  Wish we had a better solution here (Not that it has anything to do with your PR)

PR #731, opened 2020-06-08
[DRAFT] Foreign Keys Part 1

Hydrocharged:
https://github.com/liquidata-inc/dolt/pull/731#issuecomment-643224267


Hydrocharged:
So, what's missing in foreign keys?

- All of the stuff pertaining to the write paths
- The commit stuff as discussed in chat earlier
- Disallow column drops if they're in a foreign key

bats/foreign-keys.bats
```diff
@@ -0,0 +1,1236 @@
...
+    ON DELETE CASCADE,
+  FOREIGN KEY (parent_v4)
+    REFERENCES parent(v4)
+    ON DELETE CASCADE
+    ON UPDATE CASCADE
+);
+SQL
+    run dolt index ls parent
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "dolt_fk_1(v1) HIDDEN" ]] || false
```
zachmu, line 234: I wouldn't match the fk name in these tests. If the behavior is undefined, don't test for it

bats/foreign-keys.bats
```diff
@@ -0,0 +1,1236 @@
+#!/usr/bin/env bats
```
zachmu, line 1: Some newlines in between chunks of setup and test assertions would make these tests easier to read, especially when you have multiple dolt sql <<SQL heredoc chunks

go/libraries/doltcore/sqle/tables.go
```diff
@@ -498,57 +497,292 @@ func (t *AlterableDoltTable) CreateIndex(ctx *sql.Context, indexName string, usi
...
 	}
 
+	err = t.db.SetRoot(ctx, newRoot)
+	if err != nil {
+		return err
+	}
+	return t.updateFromRoot(ctx, newRoot)
+}
+
+func (t *AlterableDoltTable) DropForeignKey(ctx *sql.Context, fkName string) error {
```
zachmu, line 619: Need golang tests of these new methods ASAP (can be engine level, but should examine the root object for the presence of the correct constraints and indexes)

bats/foreign-keys.bats
zachmu: Could just match the columns, e.g. "(v1) HIDDEN"

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -17,16 +17,17 @@ package schema
 import (
 	"fmt"
 	"sort"
+	"strings"
 )
 
 type IndexCollection interface {
```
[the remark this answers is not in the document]
zachmu: Speaking more generally, about the implementation as a whole, I'm a little troubled by how tightly "hidden" is coupled to foreign keys. I think some of this coupling is necessary, and should be made explicit rather than abstracted behind a generic "hidden" flag. E.g. consider what happens when you create then drop a foreign key constraint under two scenarios: 1) there is a pre-existing index on the referenced columns, or 2) there isn't one. In the case of 1), we don't remove the index when we drop the constraint. In the case of 2) we do. So these indexes do need to have some knowledge of what they're being used for. Hidden is just too general.

Several ways this could be resolved, start thinking about it. Looking forward to seeing a solution in a future PR.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -498,57 +497,292 @@ func (t *AlterableDoltTable) CreateIndex(ctx *sql.Context, indexName string, usi
...
+		realColNames = append(realColNames, tableCol.Name)
+	}
+
+	// create the index metadata, will error if index names are taken or an index with the same columns in the same order exists
+	var index schema.Index
+	var err error
+	if !hidden {
+		index, err = tblSch.Indexes().AddIndexByColNames(indexName, realColNames, constraint == sql.IndexConstraint_Unique, comment, false)
+	} else {
+		indexName := "dolt_fk_1"
```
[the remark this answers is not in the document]
Hydrocharged: I should move this to the Foreign Key section, that would make more sense. If no name is provided for a fk, then it autogenerates one.

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -0,0 +1,322 @@
...
+	TableColumns           []string
+	ReferencedTableName    string
+	ReferencedTableIndex   string
+	ReferencedTableColumns []string
+	OnUpdate               ForeignKeyReferenceOption
+	OnDelete               ForeignKeyReferenceOption
+}
+
+type ForeignKey struct {
+	Name                   string                    `noms:"name" json:"name"`
```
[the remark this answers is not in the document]
Hydrocharged: Perhaps not, but it's on all of the other serialized types. I wouldn't want to break anything by leaving it out, so in the future we could go through and remove all of the json references if need be.

PR #732, opened 2020-06-08
Andy/query diff

andrew-wm-arthur:
This is a first pass at "query diffs", a command that diffs the results of arbitrary queries. The initial motivation for this was the ability to diff views and saved queries, but it has more applications such as extending the functionality of `dolt diff --where`.


go/libraries/doltcore/diff/querydiff/sort_node.go
```diff
@@ -0,0 +1,252 @@
...
+			return 0, err
+		}
+		if cmp != 0 {
+			return rowCmp(cmp), nil
+		}
+	}
+	return 0, nil
+}
+
+type sqlNodeWrapper struct {
```
zachmu, line 184: This is probably not a good idea. Fine to wrap a node / row iterator, but actively harmful to pretend to be one. You can't meet the interface requirements in a sane way (panics, non-standard error behavior from Next), so this would die if you actually fed it to an engine to execute instead of iterating over yourself manually. Better to make it explicit that this isn't a Node.

andrew-wm-arthur:
@zachmu I struggled with how to structure this in general. I certainly agree that creating fake sql.Nodes and bypassing the RowIter pipeline with errSkip is an abuse of go-mysql-server's design. I think this is a gross but workable solution given that we don't want to pollute go-mysql-server with query_diff logic (yet).
The main issue is that you can't simply wrap two top level iterators and diff them. The most immediate reason is because of projections. Consider the example `select c0 from test order by pk`. Also, ordering/matching query results rows by whatever sort field is really only works if you order by the primary key, the next iteration of query_diff will need to match rows below the sort node at the table iter level.
For these reasons, the diffing logic has to happen at a lower level in the query execution tree and  whatever parent nodes exist above the sort node need to be able to consume the with node with `sql.Node.WithChildren(...sql.Node)`.

PR #743, opened 2020-06-17
Upgraded go-mysql-server and vitess

zachmu:


go/go.mod
```diff
@@ -89,6 +93,6 @@ require (
 
 replace github.com/liquidata-inc/dolt/go/gen/proto/dolt/services/eventsapi => ./gen/proto/dolt/services/eventsapi
 
-replace vitess.io/vitess => github.com/liquidata-inc/vitess v0.0.0-20200604002149-44dae2034412
+replace vitess.io/vitess => github.com/liquidata-inc/vitess v0.0.0-20200617010845-e40f55ea16e3
```
Hydrocharged, line 96: `go-mysql-server` and `dolt` should reference the same Vitess version.

zachmu: Now using versions of both from master branches. Take another look when you can.

PR #744, opened 2020-06-17
Foreign Keys Part 2

Hydrocharged:
Tests for this PR are in PR https://github.com/liquidata-inc/dolt/pull/748

### DIFF
--- a/go/libraries/doltcore/doltdb/sessioned_table_editor.go
+++ b/go/libraries/doltcore/doltdb/sessioned_table_editor.go
@@ -1,0 +1,343 @@
+// Copyright 2020 Liquidata, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package doltdb
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/row"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+// SessionedTableEditor represents a table editor obtained from a TableEditSession. This table editor may be shared
+// by multiple callers. It is thread safe.
+type SessionedTableEditor struct {
+	tableEditSession  *TableEditSession
+	tableEditor       *TableEditor
+	referencedTables  []*ForeignKey
+	referencingTables []*ForeignKey
+}
+
+// ContainsIndexedKey returns whether the given key is contained within the index. The key is assumed to be in the
+// format expected of the index, similar to searching on the index map itself.
+func (ste *SessionedTableEditor) ContainsIndexedKey(ctx context.Context, key types.Tuple, indexName string) (bool, error) {
+	return ste.tableEditor.ContainsIndexedKey(ctx, key, indexName)
+}
+
+// GetIndexedRows returns all matching rows for the given key on the index. The key is assumed to be in the format
+// expected of the index, similar to searching on the index map itself.
+func (ste *SessionedTableEditor) GetIndexedRows(ctx context.Context, key types.Tuple, indexName string) ([]row.Row, error) {
+	return ste.tableEditor.GetIndexedRows(ctx, key, indexName)
+}
+
+// GetRow returns the row matching the key given from the TableEditor. This is equivalent to calling Table and then
+// GetRow on the returned table, but a tad faster.
+func (ste *SessionedTableEditor) GetRow(ctx context.Context, key types.Tuple) (row.Row, bool, error) {
+	return ste.tableEditor.GetRow(ctx, key)
+}
+
+// GetRoot is a shortcut to calling SessionTableEditor.GetRoot.
+func (ste *SessionedTableEditor) GetRoot(ctx context.Context) (*RootValue, error) {
+	return ste.tableEditSession.GetRoot(ctx)
+}
+
+// InsertRow adds the given row to the table. If the row already exists, use UpdateRow.
+func (ste *SessionedTableEditor) InsertRow(ctx context.Context, dRow row.Row) error {
+	ste.tableEditSession.writeMutex.RLock()
+	defer ste.tableEditSession.writeMutex.RUnlock()
+
+	err := ste.validateForInsert(ctx, dRow)
+	if err != nil {
+		return err
+	}
+
+	return ste.tableEditor.InsertRow(ctx, dRow)
+}
+
+// DeleteKey removes the given key from the table.
+func (ste *SessionedTableEditor) DeleteKey(ctx context.Context, key types.Tuple) error {
+	ste.tableEditSession.writeMutex.RLock()
+	defer ste.tableEditSession.writeMutex.RUnlock()
+
+	if !ste.tableEditSession.Props.ForeignKeyChecksDisabled && len(ste.referencingTables) > 0 {
+		dRow, ok, err := ste.tableEditor.GetRow(ctx, key)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return nil
+		}
+		err = ste.handleReferencingRowsOnDelete(ctx, dRow)
+		if err != nil {
+			return err
+		}
+	}
+
+	return ste.tableEditor.DeleteKey(ctx, key)
+}
+
+// DeleteRow removes the given row from the table, along with any applicable rows from tables that have a foreign key
+// referencing this table.
+func (ste *SessionedTableEditor) DeleteRow(ctx context.Context, dRow row.Row) error {
+	ste.tableEditSession.writeMutex.RLock()
+	defer ste.tableEditSession.writeMutex.RUnlock()
+
+	return ste.deleteRow(ctx, dRow)
+}
+
+// UpdateRow takes the current row and new row, and updates it accordingly. Any applicable rows from tables that have a
+// foreign key referencing this table will also be updated.
+func (ste *SessionedTableEditor) UpdateRow(ctx context.Context, dOldRow row.Row, dNewRow row.Row) error {
+	ste.tableEditSession.writeMutex.RLock()
+	defer ste.tableEditSession.writeMutex.RUnlock()
+
+	return ste.updateRow(ctx, dOldRow, dNewRow, true)
+}
+
+// deleteRow is the same as DeleteRow, except that it does not acquire any locks.
+func (ste *SessionedTableEditor) deleteRow(ctx context.Context, dRow row.Row) error {
+	err := ste.handleReferencingRowsOnDelete(ctx, dRow)
+	if err != nil {
+		return err
+	}
+
+	return ste.tableEditor.DeleteRow(ctx, dRow)
+}
+
+// handleReferencingRowsOnDelete handles updating referencing foreign keys on delete operations
+func (ste *SessionedTableEditor) handleReferencingRowsOnDelete(ctx context.Context, dRow row.Row) error {
+	if ste.tableEditSession.Props.ForeignKeyChecksDisabled {
+		return nil
+	}
+
+	for _, foreignKey := range ste.referencingTables {
+		var err error
+		referencingSte, err := ste.tableEditSession.getTableEditorOrFail(foreignKey.TableName)
+		if err != nil {
+			return err
+		}
+		indexKey, err := ste.reduceRowAndConvert(ste.tableEditor.nbf, foreignKey.ReferencedTableColumns, foreignKey.TableColumns, dRow)
+		if err != nil {
+			return err
+		}
+		referencingRows, err := referencingSte.tableEditor.GetIndexedRows(ctx, indexKey, foreignKey.TableIndex)
+		if err != nil {
+			return err
+		}
+		if len(referencingRows) == 0 {
+			continue
+		}
+
+		switch foreignKey.OnDelete {
+		case ForeignKeyReferenceOption_Cascade:
+			for _, rowToDelete := range referencingRows {
+				err = referencingSte.DeleteRow(ctx, rowToDelete)
+				if err != nil {
+					return err
+				}
+			}
+		case ForeignKeyReferenceOption_SetNull:
+			for _, oldRow := range referencingRows {
+				newRow := oldRow
+				for _, colTag := range foreignKey.TableColumns {
+					newRow, err = newRow.SetColVal(colTag, types.NullValue, referencingSte.tableEditor.tSch)
+					if err != nil {
+						return err
+					}
+				}
+				err = referencingSte.updateRow(ctx, oldRow, newRow, false)
+				if err != nil {
+					return err
+				}
+			}
+		case ForeignKeyReferenceOption_DefaultAction, ForeignKeyReferenceOption_NoAction, ForeignKeyReferenceOption_Restrict:
+			indexKeyStr, _ := types.EncodedValue(ctx, indexKey)
+			return fmt.Errorf("foreign key constraint violation on `%s`.`%s`: cannot delete rows with value `%s`",
+				foreignKey.TableName, foreignKey.Name, indexKeyStr)
+		default:
+			return fmt.Errorf("unknown ON DELETE reference option on `%s`: `%s`", foreignKey.Name, foreignKey.OnDelete.String())
+		}
+	}
+
+	return nil
+}
+
+func (ste *SessionedTableEditor) handleReferencingRowsOnUpdate(ctx context.Context, dOldRow row.Row, dNewRow row.Row) error {
+	if ste.tableEditSession.Props.ForeignKeyChecksDisabled {
+		return nil
+	}
+
+	for _, foreignKey := range ste.referencingTables {
+		var err error
+		referencedSte, err := ste.tableEditSession.getTableEditorOrFail(foreignKey.TableName)
+		if err != nil {
+			return err
+		}
+		indexKey, err := ste.reduceRowAndConvert(ste.tableEditor.nbf, foreignKey.ReferencedTableColumns, foreignKey.TableColumns, dOldRow)
+		if err != nil {
+			return err
+		}
+		referencingRows, err := referencedSte.tableEditor.GetIndexedRows(ctx, indexKey, foreignKey.TableIndex)
+		if err != nil {
+			return err
+		}
+		if len(referencingRows) == 0 {
+			continue
+		}
+
+		valueChanged := false
+		for _, colTag := range foreignKey.ReferencedTableColumns {
+			oldVal, oldOk := dOldRow.GetColVal(colTag)
+			newVal, newOk := dNewRow.GetColVal(colTag)
+			if (oldOk != newOk) || (oldOk && newOk && !oldVal.Equals(newVal)) {
+				valueChanged = true
+				break
+			}
+		}
+		if !valueChanged {
+			continue
+		}
+
+		switch foreignKey.OnUpdate {
+		case ForeignKeyReferenceOption_Cascade:
+			// NULL handling is usually done higher, so if a new value is NULL then we need to error
+			for i := range foreignKey.ReferencedTableColumns {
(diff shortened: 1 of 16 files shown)


go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+// UpdateRow takes the current row and new row, and updates it accordingly. Any applicable rows from tables that have a
+// foreign key referencing this table will also be updated.
+func (lte *LocalTableEditor) UpdateRow(ctx context.Context, dOldRow row.Row, dNewRow row.Row) error {
+	globalTableEditor.writeMutex.RLock()
+	defer globalTableEditor.writeMutex.RUnlock()
+
+	return lte.updateRow(ctx, dOldRow, dNewRow, true)
+}
+
+func (lte *LocalTableEditor) deleteReferencedRows(ctx context.Context, dRow row.Row) error {
```
zachmu, line 113: handleReferencingRowsOnDelete

go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	tpl, err := types.NewTuple(nbf, keyVals...)
+	if err != nil {
+		return types.EmptyTuple(nbf), err
+	}
+	return tpl, nil
+}
+
+func (lte *LocalTableEditor) updateRow(ctx context.Context, dOldRow row.Row, dNewRow row.Row, checkReferences bool) error {
+	if !globalTableEditor.DisableForeignKeyChecks {
```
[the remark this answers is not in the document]
zachmu: To make this symmetric with the delete method, I would extract the inner logic as handleReferencingRowsOnUpdate

go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+			return fmt.Errorf("unknown ON DELETE reference option on `%s`: `%s`", foreignKey.Name, foreignKey.OnDelete.String())
+		}
+	}
+
+	return nil
+}
+
+// deleteRow is the same as DeleteRow, except that it does not acquire any locks.
+func (lte *LocalTableEditor) deleteRow(ctx context.Context, dRow row.Row) error {
+	if !globalTableEditor.DisableForeignKeyChecks {
```
zachmu, line 169: This check belongs in the method being called, not here. Always call the method, which sometimes does nothing and returns nil.

go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+				return fmt.Errorf("unknown ON UPDATE reference option on `%s`: `%s`", foreignKey.Name, foreignKey.OnUpdate.String())
+			}
+		}
+	}
+
+	return lte.tableEditor.UpdateRow(ctx, dOldRow, dNewRow)
+}
+
+// validToInsert returns whether the given row is able to be inserted into the target table.
+func (lte *LocalTableEditor) validToInsert(ctx context.Context, dRow row.Row) error {
```
zachmu, line 287: validateForInsert

go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+func (lte *LocalTableEditor) GetRow(ctx context.Context, key types.Tuple) (row.Row, bool, error) {
+	return lte.tableEditor.GetRow(ctx, key)
+}
+
+// InsertRow adds the given row to the table. If the row already exists, use UpdateRow.
+func (lte *LocalTableEditor) InsertRow(ctx context.Context, dRow row.Row) error {
+	globalTableEditor.writeMutex.RLock()
+	defer globalTableEditor.writeMutex.RUnlock()
+
+	if !globalTableEditor.DisableForeignKeyChecks {
```
zachmu, line 61: Move this check inside validToInsert. Always call validate, let it decide which checks to perform.

go/libraries/doltcore/doltdb/global_table_editor.go
```diff
@@ -0,0 +1,278 @@
...
+
+var globalTableEditor = &GlobalTableEditorImpl{}
+
+func init() {
+	globalTableEditor.tables = make(map[string]*LocalTableEditor)
+	globalTableEditor.writeMutex = &sync.RWMutex{}
+}
+
+// GlobalTableEditor returns the global table editor.
+func GlobalTableEditor() *GlobalTableEditorImpl {
```
zachmu, line 46: This is the wrong abstraction to use here. Rather than a global variable shared by SQL all sessions, this needs to be local to a particular session or transaction. Suggest TableEditSession for a start (it's not really a Transaction yet, so probably don't call it that). Instead of accessing this globally, instantiate one for each DoltSession to make available to SQL methods. Could do this lazily if you feel ambitious.

go/libraries/doltcore/doltdb/global_table_editor.go
```diff
@@ -0,0 +1,278 @@
...
+func (gte *GlobalTableEditorImpl) GetTableEditorOverride(ctx context.Context, tableName string, t *Table, tableSch schema.Schema) (*LocalTableEditor, error) {
+	gte.writeMutex.Lock()
+	defer gte.writeMutex.Unlock()
+
+	return gte.getTableEditorOverride(ctx, tableName, t, tableSch)
+}
+
+// SetRoot sets the root fetching function to the one given. If the root function is nil then it is ignored. This must
+// be set before any other call.
+func (gte *GlobalTableEditorImpl) SetRoot(rootFunc GetRootFunc) {
```
zachmu, line 72: Again, this is the wrong abstraction. Set the root when the session begins, and update it when the session's root gets updated (or whenever else this is necessary).

go/libraries/doltcore/doltdb/global_table_editor.go
```diff
@@ -0,0 +1,278 @@
...
+	defer gte.writeMutex.Unlock()
+
+	if rootFunc != nil {
+		gte.getRoot = rootFunc
+	}
+}
+
+// Root updates the given root with all of the changed tables. If root is nil, then this fetches a root from the
+// function provided through SetRoot.
+func (gte *GlobalTableEditorImpl) ApplyToRoot(ctx context.Context, root *RootValue) (*RootValue, error) {
```
zachmu, line 83: Ideally the root value param here can be killed off. This object needs to be aware of the root of the current session. Fine to return the new root, but it should also update itself to have the new value.

go/libraries/doltcore/doltdb/global_table_editor.go
```diff
@@ -0,0 +1,278 @@
...
+	defer gte.writeMutex.Unlock()
+
+	return gte.getTableEditorOverride(ctx, tableName, nil, nil)
+}
+
+// GetTableEditor returns a LocalTableEditor using the given table. Fetches a root from the function provided to SetRoot
+// for any potential foreign keys. Any existing LocalTableEditors that reference this table will be updated with the
+// given table, losing any unsaved edits. If a nil table and schema are given, then this functions the same as
+// GetTableEditor.
+func (gte *GlobalTableEditorImpl) GetTableEditorOverride(ctx context.Context, tableName string, t *Table, tableSch schema.Schema) (*LocalTableEditor, error) {
```
zachmu, line 63: The concept of an "override" doesn't really make sense here. Or maybe I'm missing something fundamental that you could explain to me. But this doesn't seem like an "override", it's just creating a table editor for the table named, and sometimes throwing away an existing one. From the single use of this method, it's hard for me to understand your intent.

The job of this object should be to manage all open table editors, and create / update / validate them as necessary for clients to use them. If it's necessary to flush / cancel / update an existing table editor for a particular table, then make that explicit with a method of that name. But make that behavior explicit -- or else roll it into GetTableEditor itself, and intelligently handle invalidation there.

go/libraries/doltcore/doltdb/local_table_editor.go
```diff
@@ -0,0 +1,309 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/row"
+	"github.com/liquidata-inc/dolt/go/store/types"
+)
+
+type LocalTableEditor struct {
```
zachmu, line 25: Suggest  SessionedTableEditor.

Should have a reference to its editor session object, rather than accessing a global var.

go/libraries/doltcore/mvdata/table_data_loc.go
```diff
@@ -203,14 +212,14 @@ type tableEditorWriteCloser struct {
...
-	tableEditor *doltdb.TableEditor
+	tableEditor *doltdb.SessionedTableEditor
 	tableSch    schema.Schema
 }
 
 var _ DataMoverCloser = (*tableEditorWriteCloser)(nil)
 
-func (te *tableEditorWriteCloser) GetTable(ctx context.Context) (*doltdb.Table, error) {
-	return te.tableEditor.Table()
+func (te *tableEditorWriteCloser) GetRoot(ctx context.Context) (*doltdb.RootValue, error) {
```
zachmu, line 221: Same comment as on table editor. This does write IO, should not be called Get*. Should call this CommitChanges or similar.

go/libraries/doltcore/sqle/database.go
```diff
@@ -129,6 +129,7 @@ type Database struct {
 	rsw       env.RepoStateWriter
 	batchMode commitBehavior
 	tc        *tableCache
+	tes       *doltdb.TableEditSession
```
zachmu, line 132: This should be in the DoltSession, not the database

go/libraries/doltcore/doltdb/sessioned_table_editor.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		return err
+	}
+
+	return ste.tableEditor.DeleteRow(ctx, dRow)
+}
+
+// handleReferencingRowsOnDelete handles updating referencing foreign keys on delete operations
+func (ste *SessionedTableEditor) handleReferencingRowsOnDelete(ctx context.Context, dRow row.Row) error {
+	if ste.tableEditSession.Props.ForeignKeyChecksDisabled {
```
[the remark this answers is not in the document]
Hydrocharged: At least with the goal of this, I should rename it to `ForeignKeysDisabled` or `IgnoreForeignKeys` or something similar.

go/libraries/doltcore/doltdb/table_edit_session.go
```diff
@@ -0,0 +1,239 @@
...
+	defer tes.writeMutex.Unlock()
+
+	rootMutex := &sync.Mutex{}
+	wg := &sync.WaitGroup{}
+	wg.Add(len(tes.tables))
+
+	newRoot := tes.root
+	var tableErr error
+	var rootErr error
+	for tableName, ste := range tes.tables {
```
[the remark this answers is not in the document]
Hydrocharged: So a table editor, for the most part, is only concerned with itself. Let's say you have two tables, A and B, and two separate functions grab an editor for one of them. That editor is used just like the old editor, in that you issue your Insert, Update, etc. to them. Now, let's say that one of the editors is done and wants to grab the root, that's when they call this. That should have no observable effect on the other editor. The only time an editor should be deleted is if the table is deleted, otherwise you have to synchronize your table editors.

The goal of the `TableEditSession` is to track all of the internal changes for you. You, as a caller who only wants to edit or update a single table, should not be concerned with any other user of this edit session.

go/libraries/doltcore/doltdb/table_edit_session.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	tes.root = newRoot
+	return newRoot, nil
+}
+
+// SetRoot uses the given root to set all open table editors to the state as represented in the root. If any
+// tables are removed in the root, but have open table editors, then the references to those are removed. If those
+// removed table's editors are used after this, then the behavior is undefined.
+func (tes *TableEditSession) SetRoot(ctx context.Context, root *RootValue) error {
```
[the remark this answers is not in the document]
Hydrocharged: This is explicitly and intentionally not an error. There is only one time that a table editor should ever be deleted, and that's when the underlying table is deleted.

So updating the root will update all open table editors, that is the intent of this function. `TableEditSession` only handles updating row data of a table, and thus any other changes to a table must be done externally. Index updates, schema changes, foreign key changes...whatever it is, those happen outside of the editor. So in order for editors to pick up on those changes, you'd either have to 1) close all editors, make the changes, and require each previously-open editor to fetch a new one, or 2) block all writes session-wide and edit each editor in-place. I've gone with the 2nd, which fits with the _I only worry about my table_ design goal.

The only possible issue in my current implementation is if a caller sets the root without first grabbing the updated root from an edit session, as that may cause you to lose changes. I thought about adding a function with a callback to make this a bit more explicit (and prevent this issue from occurring), so I can add that if you feel it's clearer in its intent.

go/libraries/doltcore/doltdb/table_edit_session.go
```diff
@@ -0,0 +1,239 @@
...
+	err = tes.loadForeignKeys(ctx, localTableEditor)
+	if err != nil {
+		return nil, err
+	}
+
+	return localTableEditor, nil
+}
+
+// getTableEditorOrFail returns the table editor for this table or fails
+func (tes *TableEditSession) getTableEditorOrFail(tableName string) (*SessionedTableEditor, error) {
```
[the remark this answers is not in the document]
Hydrocharged: So this function is directly due to feedback left last time. I had:
```go
referencedLte, ok := globalTableEditor.tables[foreignKey.ReferencedTableName]
```
To which your feedback was:

> Should call a method here, rather than accessing the field directly

This function is only called from a `SessionedTableEditor`, and it's only in the places where I was using the above code. No clients know of or even care that this function exists (it's lowercased to make it package-private).

I think there may be some confusion on the overall goal. So the design goal of `TableEditSession` and friends is to be a sort of drop-in replacement for just using a standard, non-foreign-key-aware `TableEditor`. Meaning a client, given a session, should just call `GetTableEditor`, use it like before, and call `GetRoot` when they're done. They should not worry about anything besides that.

Internally, `TableEditSession` uses one editor per table. A client does not need to know nor care about that. This gives advantages though, such as using internal table editors for value cascading from foreign keys and the such, and if an editor already exists due to the `TableEditSession` having already instantiated one for internal use, it can just return that already-existing one, and the logic stays the same.

go/libraries/doltcore/mvdata/table_data_loc.go
```diff
@@ -105,8 +105,12 @@ func (dl TableDataLocation) NewCreatingWriter(ctx context.Context, mvOpts DataMo
 	if err != nil {
 		return nil, err
 	}
+	updatedRoot, err := root.PutTable(ctx, dl.Name, tbl)
+	if err != nil {
+		return nil, err
+	}
 
-	tableEditor, err := doltdb.NewTableEditor(ctx, tbl, outSch)
+	tableEditor, err := doltdb.CreateTableEditSession(updatedRoot, doltdb.TableEditSessionProps{}).GetTableEditor(ctx, dl.Name, outSch)
```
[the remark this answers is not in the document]
Hydrocharged: I'm heavily against that, as it can accidentally lead to devs using this `CreateSingleTableEditor` over creating a session and getting an editor from there. I want it to be clear and explicit that you need to get editors from a session, and that all of your editors need to come from the same session. If anything, I may need to separate this into two steps with an intermediate variable to make that more clear, but I'm against anything that may suggest you can create an editor without a session first.

Later on down the line, if the code is being edited and another table editor is added, it's easy to see two ways to make an editor and think "I only need this one" and call the single, even though a session exists further up the stack.

go/libraries/doltcore/sqle/database.go
Hydrocharged: Any real reason why? I looked at `DoltSession`, and it handles things like multiple databases. AFAICT, an edit session should be local to a database, thus it seems to make more sense here?

go/libraries/doltcore/doltdb/sessioned_table_editor.go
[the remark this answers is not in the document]
zachmu: I think the name is fine as is actually. I didn't understand how MySQL treats this variable, but it does what you have implemented: disables not only foreign key constraint checks, but also cascaded deletes / updates.

```
CREATE TABLE `child` (
  `b` int DEFAULT NULL,
  KEY `b` (`b`),
  CONSTRAINT `child_ibfk_1` FOREIGN KEY (`b`) REFERENCES `parent` (`a`) ON DELETE CASCADE;

mysql> select * from child;
+------+
| b    |
+------+
|    1 |
|    2 |
|    3 |
+------+
3 rows in set (0.00 sec)

mysql> select @@foreign_key_checks;
+----------------------+
| @@foreign_key_checks |
+----------------------+
|                    0 |
+----------------------+
1 row in set (0.00 sec)

mysql> select * from parent;
+------+
| a    |
+------+
|    1 |
|    2 |
|    3 |
+------+
3 rows in set (0.00 sec)

mysql> delete from parent where a = 1;

mysql> select * from child;
+------+
| b    |
+------+
|    1 |
|    2 |
|    3 |
+------+
3 rows in set (0.00 sec)

mysql> select * from parent;
+------+
| a    |
+------+
|    2 |
|    3 |
+------+
```

So just document what this property means.

go/libraries/doltcore/doltdb/table_edit_session.go
```diff
@@ -0,0 +1,239 @@
...
+// that is used for currently open editors (if any), then those editors will reload the table from the root.
+func (tes *TableEditSession) GetTableEditor(ctx context.Context, tableName string, tableSch schema.Schema) (*SessionedTableEditor, error) {
+	tes.writeMutex.Lock()
+	defer tes.writeMutex.Unlock()
+
+	return tes.getTableEditor(ctx, tableName, tableSch)
+}
+
+// GetRoot returns an updated root with all of the changed tables.
+func (tes *TableEditSession) GetRoot(ctx context.Context) (*RootValue, error) {
```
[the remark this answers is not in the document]
zachmu: Or CommitChanges

go/libraries/doltcore/doltdb/table_edit_session.go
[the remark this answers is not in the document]
zachmu: I think this makes sense, although I do still worry we'll get ourselves in trouble with this. But agree that letting clients have references to the table editors makes it necessary to keep them around. Fixing that (requiring all edits to go through the session) would be a lot of work and would have some drawbacks.

go/libraries/doltcore/doltdb/table_edit_session.go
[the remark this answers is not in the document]
zachmu: So calling a method rather than addressing the field is good. I just think it's the wrong method. If you want to do an existence check for a table's editor, make a method that does that. TableEditorExists or similar.

I don't remember the significance of the ok param in the previous revision. You sure it matters that a table editor already exists? Not obvious to me why that would be.

go/libraries/doltcore/doltdb/table_edit_session.go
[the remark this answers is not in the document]
zachmu: I would just make the comment here a little more explicit about how this is expected to be used (for schema changes) and add some caveats. Like, if I update a table's rows outside this edit session, and update it in the root that I pass here, and there is a table editor open for that table, then the edits that I made outside the session get lost. What I'm getting at is that there are many different root values you could pass this thing that would lead to bad / undefined behavior, so you should attempt to document them and give guidance on what to do instead.

go/libraries/doltcore/mvdata/table_data_loc.go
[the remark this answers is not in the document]
zachmu: Fine by me

go/libraries/doltcore/sqle/database.go
zachmu: Few reasons:
1) Mixing metaphors: session data belongs in the session object, not the database object

2) The database object isn't supposed to have any state that would prevent it from being used by another session. Like, the DB doesn't have its own root value, it gets it from the session. The one kind-of exception is the table cache, but even that is keyed by root value (which is immutable).

3) The session is in charge of managing things like the current root value for each database for the session. The table editor session has a root value. We will quickly get to a place where the DB session needs to manage this for the editor session.

You are correct that the DoltSession keys everything by database. The EditSession will have to be keyed by database as well.

(If this is going to be a lot of work, you can do it in a follow-up PR, just don't leave it long)

Hydrocharged:
Manually merged into https://github.com/liquidata-inc/dolt/pull/748

PR #746, opened 2020-06-18
query_diff via lazy projections

andrew-wm-arthur:
Another iteration of `query-diff`. The methodology is: 
- Generate a query plan at each `RootValue` from & to
- Alter the query plan to lazily evaluate projections. This allows access to any column that is used in ordering the query results
- Determine the order of the query results by extracting all primary key columns and `SortFields`
- Iterate over the modified query plans, diffing their results by comparing row order. Then evaluate the projections to produce the final diff

This version does not yet incorporate Noms layer diffing. It also does not yet handle aggregate functions. 

depends on https://github.com/liquidata-inc/go-mysql-server/pull/127


go/libraries/doltcore/diff/querydiff/analyze.go
```diff
@@ -0,0 +1,160 @@
...
+
+func getFieldsForTable(tbl sql.Table) []sql.Expression {
+	fields := make([]sql.Expression, len(tbl.Schema()))
+	for i, col := range tbl.Schema() {
+		fields[i] = expression.NewGetField(i, col.Type, col.Name, col.PrimaryKey)
+	}
+	return fields
+}
+
+func shiftFieldIndexes(composite []sql.Expression, offset int) []sql.Expression {
```
zachmu, line 128: Check out fix_field_indexes.go to see if there is anything you can use there by making it publicly exported. I think it's a more general solution to what you're trying to do here with offsets.

PR #748, opened 2020-06-22
Foreign Keys Part 2 Episode 1

Hydrocharged:
Tests for the write path, in particular PR https://github.com/liquidata-inc/dolt/pull/744. Of note, any tests that rely on commits are not yet in. Those will come along with the commit (and misc) PR.

### DIFF
--- a/go/libraries/doltcore/sqle/table_editor_fk_test.go
+++ b/go/libraries/doltcore/sqle/table_editor_fk_test.go
@@ -1,0 +1,479 @@
+// Copyright 2020 Liquidata, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sqle
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"sort"
+	"strings"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/table_editor_fk_test.go
```diff
@@ -0,0 +1,479 @@
...
+}
+
+func TestTableEditorForeignKeyViolations(t *testing.T) {
+	testRoot, err := ExecuteSql(fk_dEnv, fk_initialRoot, `
+ALTER TABLE two ADD FOREIGN KEY (v1) REFERENCES one(v1) ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE three ADD FOREIGN KEY (v1, v2) REFERENCES two(v1, v2) ON DELETE CASCADE ON UPDATE CASCADE;
+`)
+	require.NoError(t, err)
+
+	tests := []struct {
```
zachmu, line 313: Formatting for these and other test cases in this file are a bit weird. I would do two things: 1) give a name to the anonymous struct to reduce one level of nesting, 2) standardize on for indentation of these strings. Consider adding named struct keys (GoLand will do this for you).

It's not a huge deal, but the formatting does make some of these a bit harder to read than they should be

go/libraries/doltcore/sqle/table_editor_fk_test.go
```diff
@@ -0,0 +1,479 @@
...
+				root, err = executeModify(context.Background(), fk_dEnv, root, sqlStatement)
+				require.NoError(t, err)
+			}
+			root, err = executeModify(context.Background(), fk_dEnv, root, test.trigger)
+			assert.Error(t, err)
+		})
+	}
+}
+
+func testTableEditorRows(t *testing.T, root *doltdb.RootValue, expected []sql.Row, tableName string) {
```
zachmu, line 392: Should be assertTableRows or similar

Hydrocharged:
Manually merged into https://github.com/liquidata-inc/dolt/pull/753

PR #749, opened 2020-06-22
go/go.mod: go get -u ./....

reltuk:


reltuk:
Should we be running dependabot on dolt?

PR #754, opened 2020-06-23
proto/{third_party,Makefile}: Adopt protobuf-go and grpc-go for protobuf message generation.

reltuk:


reltuk:
I will come back through after this lands and remove the unused submodules.

PR #758, opened 2020-06-24
Added skipped test for json DATETIME bug. 

timsehn:
Also, added || false on some regexes to create deterministic behavior in old versions of bash.


timsehn:
Can I merge this @andrew-wm-arthur?

PR #760, opened 2020-06-24
/.github/workflows/bump-brew-formula.yml: Attempt to fix syntax and dispell yaml syntax error

coffeegoddd:
My best guess at the correct syntax... not a huge fan of actions rn lol

### DIFF
--- a/.github/workflows/bump-brew-formula.yml
+++ b/.github/workflows/bump-brew-formula.yml
@@ -9,13 +9,14 @@ jobs:
     steps:
       - name: Get the version
         id: get_version
-        run: echo "::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}" # GITHUB_REF is expected to be set in the format refs/tags/0.3.1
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-brew-formula.yml
```diff
@@ -9,13 +9,13 @@ jobs:
     steps:
       - name: Get the version
         id: get_version
-        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+        run: echo "::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}" # GITHUB_REF is expected to be set in the format refs/tags/0.3.1
```
oscarbatori, line 12: Move comment to line above?

coffeegoddd:
@oscarbatori yea the testing is harder... we can probably test by adding:
```
on:
  push:
    branches: [ some_test_branch ]
```

And then this will run when we push `some_test_branch` ... but if it works we'd have to manually close that PR that opens against Homebrew

PR #764, opened 2020-06-24
/.github/workflows/ci-bats-tests.yaml: Add dolt bats tests ci github actions

coffeegoddd:
Seems to run the tests for both `linux` and `macos` ... need to fix the broken tests that have specific dependency issues


coffeegoddd:
now passing

PR #768, opened 2020-06-25
Added GROUP BY support, tests to dolt query_diff

andrew-wm-arthur:


go/libraries/doltcore/diff/querydiff/query_differ_test.go
```diff
@@ -262,3 +362,506 @@ func testQueryDiffer(t *testing.T, test queryDifferTest) {
...
+		HAVING table_type IN ('TABLE', 'VIEW')
+		ORDER BY table_type, table_schema, table_name`,
+	},
+	{query: "SELECT substring(mytable.s, 1, 5) AS s FROM mytable INNER JOIN othertable ON (substring(mytable.s, 1, 5) = SUBSTRING(othertable.s2, 1, 5)) GROUP BY 1"},
+	{query: `SELECT SUBSTRING_INDEX(mytable.s, "d", 1) AS s FROM mytable INNER JOIN othertable ON (SUBSTRING_INDEX(mytable.s, "d", 1) = SUBSTRING_INDEX(othertable.s2, "d", 1)) GROUP BY 1 HAVING s = 'secon'`},
+	{query: `SELECT SUBSTRING(s, -3, 3) AS s FROM mytable WHERE s LIKE '%d row' GROUP BY 1`},
+	{query: `SELECT SUM(i) FROM mytable`},
+}
+
+// runs a subset of SELECT queries from enginetest/queries.go as a sanity check
```
zachmu, line 849: Any reason not to run all of them? If you're just checking that they run and produce no diff, then that should be fine, right?

Another interesting test would be to create the table, commit it, then insert the rows and verify that the "to" diff contains exactly the rows expected by the normal query in the engine test.

go/libraries/doltcore/diff/querydiff/query_differ_test.go
```diff
@@ -262,3 +362,506 @@ func testQueryDiffer(t *testing.T, test queryDifferTest) {
 	assert.Nil(t, to)
 	assert.Equal(t, io.EOF, err)
 }
+
+// subset of test data from go-mysql-server/enginetest/testdata.go:CreateSubsetTestData()
+var engineTestSetup = []testCommand{
```
zachmu, line 367: Probably should just create these with a doltHarness. Not sure what value you get by duplicating them here.

zachmu: That method even lets you create a subset of the tables you are interested in (but creating all of them should work)

go/libraries/doltcore/diff/querydiff/query_differ_test.go
```diff
@@ -262,3 +362,506 @@ func testQueryDiffer(t *testing.T, test queryDifferTest) {
...
+		assert.Equal(t, 0, exitCode)
+	}
+	return dEnv
+}
+
+type engineTestQuery struct {
+	query string
+}
+
+var engineTestSelectQueries = []engineTestQuery{
```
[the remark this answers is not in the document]
zachmu: And if you're concerned that not every test query is supported (those involving typestable e.g.), just use doltHarness.SkipQueryTest to filter them.

go/libraries/doltcore/diff/querydiff/query_differ_test.go
andrew-wm-arthur: that's a great idea

PR #771, opened 2020-06-26
Foreign Keys Part 3 Episode 1

Hydrocharged:
Here are the merge changes along with some `bats` tests

### DIFF
--- a/bats/foreign-keys.bats
+++ b/bats/foreign-keys.bats
@@ -1894,3 +1894,285 @@ SQL
     dolt add super_parent
     dolt commit -m "passes now"
 }
+
+@test "foreign-keys: Merge valid onto parent" {
+    dolt sql <<SQL
+CREATE TABLE parent (
+  pk BIGINT PRIMARY KEY,
+  v1 BIGINT
+);
+CREATE TABLE child (
+  pk BIGINT PRIMARY KEY,
+  v1 BIGINT,
+  CONSTRAINT fk_name FOREIGN KEY (v1)
+    REFERENCES parent(v1)
+);
+INSERT INTO parent VALUES (1, 1), (2, 2), (3, 3);
+INSERT INTO child VALUES (2, 1), (3, 2), (4, 3);
+SQL
+
+    dolt add -A
+    dolt commit -m "initial commit"
+    dolt checkout -b other
+    dolt checkout master
+    dolt sql <<SQL
+INSERT INTO parent VALUES (4, 3);
+SQL
+    dolt add -A
+    dolt commit -m "added row"
+    dolt checkout other
+
+    dolt sql <<SQL
+ALTER TABLE child DROP FOREIGN KEY fk_name;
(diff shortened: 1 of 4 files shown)


bats/foreign-keys.bats
```diff
@@ -1894,3 +1894,282 @@ SQL
...
+UPDATE parent SET v1 = v1 - 1;
+SQL
+    dolt add -A
+    dolt commit -m "updated parent"
+    dolt checkout master
+    dolt merge other
+    
+    run dolt sql -q "SELECT * FROM parent" -r=csv
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "pk,v1" ]] || false
```
zachmu, line 1936: When you're matching a whole bunch of lines in order like this, it's more compact to echo with newlines. Like this:

`EXPECTED=$( echo -e "table,num_conflicts\none_pk,1\ntwo_pk,1")`

zachmu: Also add an order by clause so that the order doesn't vary

bats/foreign-keys.bats
```diff
@@ -1894,3 +1894,282 @@ SQL
...
+    dolt checkout master
+    dolt sql <<SQL
+INSERT INTO parent VALUES (4, 4);
+SQL
+    dolt add -A
+    dolt commit -m "added row"
+    dolt checkout other
+
+    dolt sql <<SQL
+ALTER TABLE child DROP FOREIGN KEY fk_name;
```
zachmu, line 1979: This is actually kind of weird... I would expect the schema change (dropped a foreign key) to get merged along with the data, so that there wouldn't actually be a merge validation problem. I would add a note about this behavior -- it seems wrong to me.

zachmu: In the meantime, you could use a flag or session var instead to disable foreign key checks, right?

bats/foreign-keys.bats
```diff
@@ -1894,3 +1894,282 @@ SQL
...
+UPDATE parent SET v1 = v1 - 1;
+SQL
+    dolt add -A
+    dolt commit -m "updated parent"
+    dolt checkout master
+    run dolt merge other
+    [ "$status" -eq "1" ]
+    [[ "$output" =~ "violation" ]] || false
```
zachmu, line 1987: Does this error message include the key value for the constraint that was violated? It should

bats/foreign-keys.bats
```diff
@@ -1894,3 +1894,282 @@ SQL
...
+    dolt checkout master
+    dolt sql <<SQL
+INSERT INTO parent VALUES (4, 4);
+SQL
+    dolt add -A
+    dolt commit -m "added row"
+    dolt checkout other
+
+    dolt sql <<SQL
+ALTER TABLE child DROP FOREIGN KEY fk_name;
```
zachmu, line 2018: Same comment about schema changes here

PR #772, opened 2020-06-26
Zachmu/show indexes

zachmu:
Implemented new interface methods needed by latest go-mysql-server


go/libraries/doltcore/sqle/enginetest/dolt_harness.go
```diff
@@ -120,7 +120,7 @@ func (d *doltHarness) NewTableAsOf(db sql.VersionedDatabase, name string, schema
 // branch with the given name.
 func (d *doltHarness) SnapshotTable(db sql.VersionedDatabase, name string, asOf interface{}) error {
 	ddb := db.(sqle.Database)
-	e := enginetest.NewEngineWithDbs(d.t, d.Parallelism(), []sql.Database{db}, nil)
+	e := enginetest.NewEngineWithDbs(d.t, d, []sql.Database{db}, nil)
```
andrew-wm-arthur, line 123: we don't need an index driver?

go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -159,6 +160,21 @@ func (di *doltIndex) ID() string {
...
 }
 
+// IsUnique implements sql.Index
+func (di *doltIndex) IsUnique() bool {
+	return di.unique
+}
+
+// Comment implements sql.Index
+func (di *doltIndex) Comment() string {
+	return "" // TODO: store comments
```
Hydrocharged, line 170: In `go/libraries/doltcore/sqle/tables::GetIndexes`, where we create the `doltIndex` objects, you have access to the index. Just like you're doing for `Unique` you can do for for `Comment` too.

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
zachmu: Nope, got rid of that about a month ago. Tables provide their own indexes now, no driver needed.

go/libraries/doltcore/sqle/dolt_index.go
zachmu: Oh snap, I didn't think we had this! Added.

PR #774, opened 2020-06-29
Foreign Keys Part 3 Episode 2

Hydrocharged:
The last bit of things to add for foreign keys!

### DIFF
--- a/go/libraries/doltcore/sqle/tables.go
+++ b/go/libraries/doltcore/sqle/tables.go
@@ -371,7 +371,13 @@ func (t *AlterableDoltTable) DropColumn(ctx *sql.Context, columnName string) err
 		return err
 	}
 
-	updatedTable, err = alterschema.DropColumn(ctx, updatedTable, columnName)
+	fkCollection, err := root.GetForeignKeyCollection(ctx)
+	if err != nil {
+		return err
+	}
+	declaresFk, referencesFk := fkCollection.KeysForTable(t.name)
+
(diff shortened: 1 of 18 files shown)


go/libraries/doltcore/sqle/tables.go
```diff
@@ -634,6 +661,55 @@ func (t *AlterableDoltTable) CreateForeignKey(ctx *sql.Context, fkName string, c
 		return err
 	}
 
+	// ensure that the table's data conforms to the foreign key
```
zachmu, line 664: This new block should be a method on the foreign key / collection.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -568,8 +592,10 @@ func (t *AlterableDoltTable) CreateForeignKey(ctx *sql.Context, fkName string, c
 			Length: 0,
 		}
 		if !tblCols[i].TypeInfo.Equals(tableCol.TypeInfo) {
-			//TODO: make this more descriptive about our lack of support of different types for fks, figure it out later
-			return fmt.Errorf("column type mismatch on `%s` and `%s`", columns[i], tableCol.Name)
+			return fmt.Errorf("Column type mismatch on `%s` and `%s`. In other databases, we feel that type "+
```
zachmu, line 595: I wouldn't explain our thinking so thoroughly. Just note that the types of columns must match exactly.

PR #776, opened 2020-06-29
moved query_diff to diff -q

andrew-wm-arthur:


go/cmd/dolt/commands/diff.go
```diff
@@ -135,11 +136,12 @@ func (cmd DiffCmd) CreateMarkdown(fs filesys.Filesys, path, commandStr string) e
 func (cmd DiffCmd) createArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParser()
 	ap.SupportsFlag(DataFlag, "d", "Show only the data changes, do not show the schema changes (Both shown by default).")
-	ap.SupportsFlag(SchemaFlag, "s", "Show only the schema changes, do not show the data changes (Both shown by default).")
+	ap.SupportsFlag(SchemaFlag, "", "Show only the schema changes, do not show the data changes (Both shown by default).")
 	ap.SupportsFlag(SummaryFlag, "", "Show summary of data changes")
-	ap.SupportsFlag(SQLFlag, "q", "Output diff as a SQL patch file of {{.EmphasisLeft}}INSERT{{.EmphasisRight}} / {{.EmphasisLeft}}UPDATE{{.EmphasisRight}} / {{.EmphasisLeft}}DELETE{{.EmphasisRight}} statements")
+	ap.SupportsFlag(SQLFlag, "s", "Output diff as a SQL patch file of {{.EmphasisLeft}}INSERT{{.EmphasisRight}} / {{.EmphasisLeft}}UPDATE{{.EmphasisRight}} / {{.EmphasisLeft}}DELETE{{.EmphasisRight}} statements")
```
zachmu, line 141: I think this is probably the wrong interface. We want to mirror the command line of `dolt sql`, where you have a single flag that lets you control output format. Something like `-r sql` (default is `-r tabular`). Then we can introduce new diff output formats over time without adding new additional flags.

go/cmd/dolt/commands/diff.go
```diff
@@ -155,7 +157,13 @@ func (cmd DiffCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 	}
 
-	verr := diffUserTables(ctx, fromRoot, toRoot, dArgs)
+	var verr errhand.VerboseError
```
zachmu, line 160: No reason to move this to a var declaration, just keep it in the block. These are two different variables.

go/cmd/dolt/commands/diff.go
```diff
@@ -173,16 +180,41 @@ func (cmd DiffCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
-	dArgs.diffOutput = TabularDiffOutput
-	if apr.Contains(SQLFlag) {
+	f, _ := apr.GetValue(formatFlag)
+	switch strings.ToLower(f) {
+	case "tablular":
+		dArgs.diffOutput = TabularDiffOutput
+	case "sql":
 		dArgs.diffOutput = SQLDiffOutput
+	case "":
+		dArgs.diffOutput = TabularDiffOutput
```
zachmu, line 217: Should die on a default case here

andrew-wm-arthur: good call

PR #781, opened 2020-07-01
Zachmu/show foreign keys

zachmu:
Implemented new foreign key interfaces for go-mysql-server, and upgrade to latest vitess and go-mysql-server


go/libraries/doltcore/sqle/tables.go
```diff
@@ -726,6 +729,87 @@ func (t *AlterableDoltTable) DropForeignKey(ctx *sql.Context, fkName string) err
...
+	root, err := t.db.GetRoot(ctx)
+	if err != nil {
+		return nil, err
+	}
+	fkc, err := root.GetForeignKeyCollection(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	fks, _ := fkc.KeysForTable(t.name)
```
Hydrocharged, line 743: This should be `fks, err := fkc.KeysForDisplay(ctx, t.name, root)`, as `toForeignKeyConstraint` is essentially duplicating that function

zachmu: Much better, thanks.

PR #784, opened 2020-07-07
go/cmd/dolt: Change commit spec handling so that abbreviated forms of remotes are supported.

reltuk:
This changes `doltdb.CommitSpec` to carry the original input in the `refs` case, instead of trying to add a prefix or anything else. Instead, `doltdb.Resolve()` takes the current `HEAD` ref, and fully resolves the `CommitSpec` itself.

There was some confusion in usage across the code base with `NewCommitSpec`. In particular, there was a lot of `NewCommitSpec("HEAD", "some-branch-name-that-is-not-CWB")` in order to get `NewCommitSpec("some-branch-name-that-is-not-CWB", "")`. This collapsed all such uses to the same syntax. `"HEAD"` is only used for resolving the CWB now.

In places where the CommitSpec was statically known to not be `HEAD`, I've not bothered to always thread the `CWBRef` to the `Resolve` call.

This change will break `ld` Dolt usage, but I will follow up there when this lands in `master`.

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -273,26 +273,26 @@ func (ddb *DoltDB) Resolve(ctx context.Context, cs *CommitSpec, cwb ref.DoltRef)
 
 	var commitSt types.Struct
 	var err error
-	switch cs.CSType {
-	case HashCommitSpec:
-		commitSt, err = getCommitStForHash(ctx, ddb.db, cs.BaseRef)
-	case RefCommitSpec:
+	switch cs.csType {
+	case hashCommitSpec:
+		commitSt, err = getCommitStForHash(ctx, ddb.db, cs.baseSpec)
+	case refCommitSpec:
 		// For a ref in a CommitSpec, we have the following behavior.
 		// If it starts with `refs/`, we look for an exact match before
 		// we try any suffix matches. After that, we try a match on the
 		// user supplied input, with the following three prefixes, in
 		// order: `refs/`, `refs/heads/`, `refs/remotes/`.
 		candidates := []string{
-			"refs/" + cs.BaseRef,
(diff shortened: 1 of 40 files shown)


go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -187,8 +187,8 @@ func (ddb *DoltDB) WriteEmptyRepoWithCommitTime(ctx context.Context, name, email
 	return err
 }
 
-func getCommitStForRef(ctx context.Context, db datas.Database, dref ref.DoltRef) (types.Struct, error) {
-	ds, err := db.GetDataset(ctx, dref.String())
+func getCommitStForDatasetName(ctx context.Context, db datas.Database, datasetName string) (types.Struct, error) {
```
bheni, line 190: Typically don't let the term "dataset" bleed into dolt code as it requires devs to understand what a dataset is in noms.

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -265,17 +265,49 @@ func getAncestor(ctx context.Context, vrw types.ValueReadWriter, commitSt types.
 }
 
 // Resolve takes a CommitSpec and returns a Commit, or an error if the commit cannot be found.
-func (ddb *DoltDB) Resolve(ctx context.Context, cs *CommitSpec) (*Commit, error) {
+// If the CommitSpec is a HEAD, Resolve also needs the DoltRef of the current working branch.
```
bheni, line 268: "a HEAD"?

go/libraries/doltcore/doltdb/doltdb.go
reltuk: Is `getCommitStForRef` better here, even if it doesn't take a `ref.DoltRef`?

PR #785, opened 2020-07-07
go/cmd/dolt/commands: merge.go: Make merge operate on commit specs, not branch refs directly.

reltuk:


andrew-wm-arthur:
My changes all happen after we resolve refs to RootValues. No conflict here.

PR #795, opened 2020-07-13
go/libraries/doltcore/doltdb: Use parentsList in a Commit if it is available.

reltuk:


reltuk:
The approach here is to fully parse the parents anytime we construct a `doltdb.Commit` and then use the in-memory copy instead of the struct field. Alternatives include:
* Lazily parse on first access. Probably have to deal with concurrency. Maybe gated on a `sync.Once` or something.
* Keep the current access patterns and never parse. Choose the path on every top-level access.
This seemed simplest and workable for our use case. Let me know if that seems wrong to you though.

PR #803, opened 2020-07-21
Andy/foreign key changes

andrew-wm-arthur:
- Indexes can no longer be HIDDEN
- Foreign Keys must have a pre-existing index in the parent table
- Foreign Keys use pre-existing indexes in the child table if it exists

Still TODO: Diff, Merge, and Stage changes


bats/foreign-keys.bats
```diff
@@ -597,759 +211,139 @@ SQL
...
-    ! [[ "$output" =~ "(parent_id) HIDDEN" ]] || false
+    [[ "$output" =~ "(v1)" ]] || false
     run dolt schema show child
     [ "$status" -eq "0" ]
     ! [[ "$output" =~ 'CONSTRAINT `fk_name`' ]] || false
     run dolt sql -q "ALTER TABLE child DROP FOREIGN KEY fk_name"
     [ "$status" -eq "1" ]
 }
 
-@test "foreign-keys: ALTER TABLE SET NULL on non-nullable column" {
```
Hydrocharged, line 1209: Why remove this test?

go/libraries/doltcore/sqle/table_editor_fk_test.go
```diff
@@ -384,7 +379,9 @@ func assertTableEditorRows(t *testing.T, root *doltdb.RootValue, expected []sql.
 			sqlRows = append(sqlRows, sqlRow)
 			return nil
 		})
-		assert.Equal(t, convertSqlRowToInt64(expected), sqlRows)
+		if !assert.Equal(t, convertSqlRowToInt64(expected), sqlRows) {
+			fmt.Println("")
```
Hydrocharged, line 383: I'm assuming you wrote something here and forgot to remove it afterward lol

andrew-wm-arthur: Yeah I needed somewhere to put a breakpoint, good catch

bats/foreign-keys.bats
andrew-wm-arthur: good call

PR #811, opened 2020-07-25
Andy/foreign key changes

andrew-wm-arthur:


go/libraries/doltcore/schema/index_coll.go
```diff
@@ -28,6 +28,8 @@ type IndexCollection interface {
 	AddIndexByColNames(indexName string, cols []string, props IndexProperties) (Index, error)
 	// AddIndexByColTags adds an index with the given name and column tags (in index order).
 	AddIndexByColTags(indexName string, tags []uint64, props IndexProperties) (Index, error)
+	// todo: this method is trash, clean up this interface
```
Hydrocharged, line 31: In the meantime, we should add a doc comment stating what makes it unsafe.

go/libraries/doltcore/sqle/table_editor_fk_test.go
```diff
@@ -58,6 +58,8 @@ CREATE TABLE three (
   v1 BIGINT,
   v2 BIGINT
 );
+ALTER TABLE one ADD INDEX secondary (v1);
```
Hydrocharged, line 61: Should be able to move this up to the `CREATE TABLE` statement now

PR #812, opened 2020-07-28
Andy/range reader empty map

andrew-wm-arthur:


go/libraries/doltcore/table/typed/noms/range_reader_test.go
```diff
@@ -28,8 +28,68 @@ import (
...
+	},
+	{
+		"test multiple ranges",
+		[]*ReadRange{
+			NewRangeEndingBefore(mustTuple(10), greaterThanCheck(2)),
+			NewRangeStartingAt(mustTuple(10), lessThanCheck(20)),
+		},
+		[]int64{8, 6, 4, 10, 12, 14, 16, 18},
+	},
+	{
```
andrew-wm-arthur, line 69: The four tests below are new, the tests above were copied.

go/store/types/map.go
```diff
@@ -399,26 +399,28 @@ func (m Map) IteratorBackFrom(ctx context.Context, key Value) (MapIterator, erro
...
-	}
+
+		if !isLess && !key.Equals(entry.key) {
+			_, err := cur.advance(ctx)
+
+			if err != nil {
+				return nil, err
+			}
+		}
+	} // else: we're at the beginning of the map
```
andrew-wm-arthur, line 423: IDK what's happening with this diff. I added lines 402 and 423

bheni: Change looks good.  Not sure what you are asking.

If the map has at least one element in it the 3 cases (key < map element key, key == map element key, key > map element key) will all return a valid MapIterator even if it has no valid elements.

In the case of an empty map it sees it's passed the last element, goes back one and falsely assumes there is a valid element there... but there isn't.

PR #813, opened 2020-07-30
Foreign Key, Index changes

andrew-wm-arthur:
- Added `diff` support for foreign keys and indexes
- Updated commit, stage, and reset logic to match tables based on column tags instead of table names


go/libraries/doltcore/sqle/tables.go
```diff
@@ -706,28 +706,63 @@ func (t *AlterableDoltTable) GetForeignKeys(ctx *sql.Context) ([]sql.ForeignKeyC
...
-		ReferencedTable:   key.ReferencedTableName,
-		ReferencedColumns: key.ReferencedTableColumns,
-		OnUpdate:          toReferenceOption(key.OnUpdate),
-		OnDelete:          toReferenceOption(key.OnDelete),
+func toForeignKeyConstraint(fk doltdb.ForeignKey, childSch, parentSch schema.Schema) (cst sql.ForeignKeyConstraint, err error) {
+	cst = sql.ForeignKeyConstraint{
+		Name:              fk.Name,
+		Columns:           make([]string, len(fk.ReferencedTableColumns)),
+		ReferencedTable:   fk.ReferencedTableName,
+		ReferencedColumns: make([]string, len(fk.TableColumns)),
```
Hydrocharged, line 741: Got the columns reversed. Doesn't actually matter since their lengths are always equivalent but still needs fixing lol

andrew-wm-arthur: _that_ is some attention to detail, nice catch

bats/diff.bats
```diff
@@ -131,6 +131,40 @@ teardown() {
...
+    c2 bigint,
+    INDEX c1 (c1)
+);
+ALTER TABLE test ADD CONSTRAINT fk1 FOREIGN KEY (c1) REFERENCES parent(c1);
+SQL
+    dolt add -A
+    dolt commit -m "added parent table, foreign key"
+    dolt sql <<SQL
+ALTER TABLE parent ADD INDEX c2 (c2);
+ALTER TABLE parent DROP INDEX c1;
```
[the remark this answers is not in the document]
andrew-wm-arthur: Good catch, I'll fix

PR #820, opened 2020-08-03
foreign keys can use primary keys as parent table indexes

andrew-wm-arthur:
This change makes a special case for foreign keys that reference primary key(s) in the parent table. If all of the parent table columns are primary keys, we create an index to use for the foreign key.

Ideally we'd make use of the existing clustered index as the parent table index. Because we will soon be adding keyless tables and allowing primary key changes, I'm punting, for now, on a more elegant implementation.

### DIFF
--- a/bats/foreign-keys.bats
+++ b/bats/foreign-keys.bats
@@ -47,6 +47,16 @@ SQL
     [[ `echo "$output" | tr -d "\n" | tr -s " "` =~ 'CONSTRAINT `fk_named` FOREIGN KEY (`v1`) REFERENCES `parent` (`v1`)' ]] || false
 }
 
+@test "foreign-keys: parent table index recquired" {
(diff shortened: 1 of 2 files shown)


bats/foreign-keys.bats
```diff
@@ -47,6 +47,16 @@ SQL
     [[ `echo "$output" | tr -d "\n" | tr -s " "` =~ 'CONSTRAINT `fk_named` FOREIGN KEY (`v1`) REFERENCES `parent` (`v1`)' ]] || false
 }
 
+@test "foreign-keys: parent table index recquired" {
```
Hydrocharged, line 50: _recquired_

zachmu: rekwired

andrew-wm-arthur: I would like the record to show that neither of my coworkers were helpful in fixing my spelling error

PR #821, opened 2020-08-04
Memory Leak Fix Part 2

Hydrocharged:


go/libraries/doltcore/doltdb/index_editor.go
```diff
@@ -68,6 +68,8 @@ func (indexEd *IndexEditor) Flush(ctx context.Context) error {
 	indexEd.flushMutex.Lock()
 	defer indexEd.flushMutex.Unlock()
 
+	defer indexEd.ed.Close() // current edit accumulator is captured by defer
```
zachmu, line 71: Might add a note about why this is important so somebody after you doesn't reintroduce the leak

go/store/types/edits/async_sorted_edits.go
```diff
@@ -173,6 +176,12 @@ func (ase *AsyncSortedEdits) FinishedEditing() (types.EditProvider, error) {
 	return ase.Iterator(), nil
 }
 
+func (ase *AsyncSortedEdits) Close() {
+	if !ase.closed {
```
zachmu, line 180: There's a race here. You should add a comment to this method describing what it does, and note that it isn't thread safe, and must be synchronized externally.

PR #828, opened 2020-08-06
go/store/nbs: Make tableIndex into an interface, and add an off-heap implementation using mmap.

reltuk:
A table file index maps chunk addresses to offsets within the file, and is a core part of the chunk store read path. An in-memory index entry consists of:

1) A chunk address prefix – the first 8 bytes of the chunk address.
2) A `uint32` ordinal – which number chunk this prefix entry is in the file.
3) A `uint64` offset – the first byte of the chunk in the file.
4) A `uint32` length – the length of the chunk in the file.
5) A chunk address suffix – the last 12 bytes of the chunk address.

On disk, `offset` isn't stored, but it's computed from the lengths when the index is loaded.

Overall, an index entry takes up 36 bytes. Target chunk size is 4096 bytes, so this is about 0.9% of the total repository size of RAM required to access table files in the repository. A 300GB repository needs 2.8GB of RAM just to locate a chunk.

This initial PR adds a separate index serving structure that moves some of the above off-heap. Initially, this usage does not use file backed `mmap`s. This initial implementation is slower on the compaction path and uses almost as much memory as the old implementation. It's a start to be able to play with tradeoffs in this space though.

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/codahale/blake2 v0.0.0-20150924215134-8d10d0420cbf
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/dustin/go-humanize v1.0.0
+	github.com/edsrzf/mmap-go v1.0.0
 	github.com/fatih/color v1.9.0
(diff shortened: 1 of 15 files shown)


reltuk:
Initial RFC. Still needs some unit tests on `mmapTableIndex` itself.

PR #838, opened 2020-08-12
Andy/table renames

andrew-wm-arthur:
This is a temporary fix to address a workflow like:
```
1. CREATE TABLE sick_data (pk int PRIMARY KEY);
2. DROP TABLE sick_data;
3. CREATE TABLE sick_data (pk int PRIMARY KEY);
```
We want to allow table names to be reused, so we're relaxing validations table creation. However, step 3 will overwrite the super schema data written in step one. This could potentially cause tag uniqueness issues later on.
A more complete solution is needed to store different super schemas for the same table name.


andrew-wm-arthur:
In addition to fixing table name reuse, this PR also addresses erroneous output from `dolt status` and `dolt reset` see the new BATS files for testing examples. This is the tail end of work to change Dolt's table matching logic. Instead of matching by name, we always want to compare column tags in the schemas.

go/cmd/dolt/commands/reset.go
```diff
@@ -254,15 +254,28 @@ func printNotStaged(ctx context.Context, dEnv *env.DoltEnv, staged *doltdb.RootV
...
-		for _, tblName := range notStagedTbls.Tables {
-			tdt := notStagedTbls.TableToType[tblName]
+	if removeModified+notStagedDocs.NumRemoved+notStagedDocs.NumModified > 0 {
+		cli.Println("Unstaged changes after reset:")
 
-			if tdt != diff.AddedTable && !doltdb.IsReadOnlySystemTable(tblName) {
-				lines = append(lines, fmt.Sprintf("%s\t%s", tblDiffTypeToShortLabel[tdt], tblName))
+		var lines []string
+		for _, td := range notStagedTbls {
+			if td.IsAdd() {
```
zachmu, line 269: This logic is repeated nearly verbatim in several places. Can you extract a method?

andrew-wm-arthur: I plan on cleaning up the printing logic, but I have a larger refactor involving DocDiffs planned. 
For now, it just does the right thing

PR #839, opened 2020-08-14
Bug fix for reverse sequence cursor. Not sufficient to just mark a cu…

zachmu:
…rsor reverse after the fact, need to initialize cursors at all levels of the tree as reverse as well. Still needs tests, but fixes panics in sqllogictest.

### DIFF
--- a/go/store/types/ordered_sequences.go
+++ b/go/store/types/ordered_sequences.go
@@ -89,7 +89,7 @@ func newCursorAt(ctx context.Context, seq orderedSequence, key orderedKey, forIn
 	return cur, nil
 }
 
-// newCursorBackFromValue returns a reverse sequence iterator starting from the last value <= the one given
+// newCursorBackFromValue returns a reverse sequence iterator starting from the greatest value <= the one given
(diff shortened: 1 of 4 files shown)


zachmu:
Still needs unit tests, but does this seem like a reasonable solution? Fixes panics in sqllogictest.

zachmu:
Aaron, PTAL

go/store/types/ordered_sequences.go
```diff
@@ -89,6 +89,72 @@ func newCursorAt(ctx context.Context, seq orderedSequence, key orderedKey, forIn
...
+		if cs == nil {
+			break
+		}
+		seq = cs.(orderedSequence)
+	}
+
+	// If we overshot the key, back off by one. We want the greatest element <= key
+	if cur.valid() {
+		currKey, err := cur.seq.(orderedSequence).getKey(cur.idx)
+		d.PanicIfError(err)
```
reltuk, line 139: Is this an appropriate panic, or should be return err?

zachmu: Good call

PR #848, opened 2020-08-24
MySQL Client Tests

timsehn:
This PR will contain the MySQL client tests in each language. I've chosen BATS as the framework to drive the testing because I'm familiar with it. We can swap that out if necessary.


timsehn:
@coffeegoddd should we merge this?

PR #849, opened 2020-08-27
/mysql-client-tests/{node/, MySQLDockerfile}: POC dockerizing test env

coffeegoddd:
so the docker container seems to work with the following changes... my JS/Node is kinda wack these days too...
so this can run like this:
```
$ cd mysql-client-tests
$ docker build -t mysql-tests -f MySQLDockerfile .
$ docker run -it --rm mysql-tests:latest /bin/bash
root@e1fbe3c4579f:/mysql-client-tests# bats mysql-client-tests.bats 
 ✓ python mysql.connector client
 ✓ python pymysql client
 ✓ mysql-connector-java client
 ✓ node mysql client

4 tests, 0 failures
root@e1fbe3c4579f:/mysql-client-tests# exit
```

### DIFF
--- a/mysql-client-tests/node/index.js
+++ b/mysql-client-tests/node/index.js
@@ -1,12 +1,12 @@
-let mysql = require('mysql');
+const mysql = require('mysql');
 
-var args = process.argv.slice(2);
+const args = process.argv.slice(2);
 
-var user = args[0];
-var port = args[1];
-var db   = args[2];
+const user = args[0];
+const port = args[1];
+const db   = args[2];
 
-var config = {
+const config = {
     host: '127.0.0.1',
     user: user,
     port: port,
@@ -15,7 +15,8 @@ var config = {
 
 class Database {
     constructor( config ) {
-        this.connection = mysql.createConnection( config );
+				this.connection = mysql.createConnection( config );
+				this.connection.connect();
     }
 
     query( sql, args ) {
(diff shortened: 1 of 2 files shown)


timsehn:
Explain to me why you changed the JS to the way you did...

mysql-client-tests/node/index.js
```diff
@@ -1,12 +1,12 @@
-let mysql = require('mysql');
+const mysql = require('mysql');
```
coffeegoddd, line 1: js best practice is to use `const` where applicable... didn't necessarily see a hoisting prob in here, but `var` is like taboo js now

mysql-client-tests/node/index.js
```diff
@@ -15,41 +15,40 @@ var config = {
 
 class Database {
     constructor( config ) {
-        this.connection = mysql.createConnection( config );
+				this.connection = mysql.createConnection( config );
+				this.connection.connect();
```
coffeegoddd, line 19: `connection.connect()` call comes from example here: https://www.npmjs.com/package/mysql

mysql-client-tests/node/index.js
```diff
@@ -15,41 +15,40 @@ var config = {
...
-                resolve();
-            } );
-        } );
+      this.connection.end(err => {
+				if (err) {
+					console.error(err)
+				} else {
+					console.log("db connection closed")
+				}
+			})
```
coffeegoddd, line 38: My changes originated from this app hanging and not exiting properly in the docker container... locally it seemed to exit without hanging, but not when dockerized... just refactored this to match what was in the npm example.

mysql-client-tests/node/index.js
```diff
@@ -87,27 +86,29 @@ async function main() {
...
+					console.log("Expected:", expected);
+					throw new Error("Query failed")
+				} else {
+					console.log("Query succeeded:", query)
+				}
+			}).catch(err => {
+				console.error(err)
+				process.exit(1);
+			});
+		}));
```
coffeegoddd, line 108: I think `Promise.all()` is the recommended way to do Promises in a loop, also i was seeing errors in docker about not having `.catch()` after the Promise.

timsehn:
Word. Thanks. You can merge this. Do you want to try .net or c? Or just let me do it in this new way?

timsehn:
Oh. also add a README.md in the directory about how to run the tests now.

coffeegoddd:
@timsehn I'll add a README.md... this still needs some work, it needs to build dolt from source and i need to figure out a credible strategy for integrating a dockerfile like this with github actions... im gonna do that first over adding .net and c.

timsehn:
Word. Get it working on PRs and then we'll talk.

PR #851, opened 2020-08-27
Andy/tag redesign

andrew-wm-arthur:


go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -333,6 +357,57 @@ func (ddb *DoltDB) ResolveRef(ctx context.Context, ref ref.DoltRef) (*Commit, er
...
+	}
+
+	dsHead, hasHead := ds.MaybeHead()
+
+	if !hasHead {
+		return nil, nil, ErrBranchNotFound
+	}
+
+	if dsHead.Name() != datas.TagName {
+		return nil, nil, fmt.Errorf("dataset head is neither commit nor tag")
```
reltuk, line 374: Nit: something like `"tagRef head is not a tag"`

go/libraries/doltcore/doltdb/tag_meta.go
```diff
@@ -0,0 +1,120 @@
...
+// TagMeta contains all the metadata that is associated with a tag within a data repo.
+type TagMeta struct {
+	Name        string
+	Email       string
+	Timestamp   uint64
+	Description string
+}
+
+// NewTagMeta returns CommitMeta that can be used to create a tag.
+func NewTagMeta(name, email, desc string) *TagMeta {
```
reltuk, line 48: Seems a little weird to me that this doesn't take the `time.Time` as well. But maybe this is fine...? In dolt's architecture maybe I would expect the following:

```
   On Backdating Tags
       If you have imported some changes from another VCS and would like to add tags for major releases of your work, it is useful
       to be able to specify the date to embed inside of the tag object; such data in the tag object affects, for example, the
       ordering of tags in the gitweb interface.

       To set the date used in future tag objects, set the environment variable GIT_COMMITTER_DATE (see the later discussion of
       possible values; the most common form is "YYYY-MM-DD HH:MM").

       For example:

           $ GIT_COMMITTER_DATE="2006-10-02 10:31" git tag -s v1.0.1
```

would be implemented through a `dEnv` lookup or something?

go/store/datas/database.go
```diff
@@ -95,6 +95,8 @@ type Database interface {
 	// of a conflict, Commit returns an 'ErrMergeNeeded' error.
 	CommitValue(ctx context.Context, ds Dataset, v types.Value) (Dataset, error)
 
+	Tag(ctx context.Context, ds Dataset, ref types.Ref, opts TagOptions) (Dataset, error)
```
reltuk, line 98: Add a docstring here.

go/store/datas/database_common.go
```diff
@@ -417,6 +417,78 @@ func (db *database) doCommit(ctx context.Context, datasetID string, commit types
...
+	)
+}
+
+// doTag manages concurrent access the single logical piece of mutable state: the current Root. It uses
+// the same optimistic writing algorithm as doCommit (see above).
+func (db *database) doTag(ctx context.Context, datasetID string, tag types.Struct) error {
+	if is, err := db.validateTag(ctx, tag); err != nil {
+		return err
+	} else if !is {
+		d.Panic("Can't tag using a non-Commit struct (dataset %s)", datasetID)
```
reltuk, line 442: Should `validateTag` be `panicIfNotValidTag` or something? It could provide a better error message at least...there are lots of `false` returns there.

`err` is still reasonable to return in the case of i/o errors / non-logic errors.

go/store/datas/database_common.go
```diff
@@ -417,6 +417,78 @@ func (db *database) doCommit(ctx context.Context, datasetID string, commit types
...
+		}
+
+		_, hasHead, err := currentDatasets.MaybeGet(ctx, types.String(datasetID))
+
+		if err != nil {
+			return err
+		}
+
+		if hasHead {
+			d.Panic("datasets for tags (refs/tags/*) cannot be altered after creation")
```
reltuk, line 473: Should this be an `err` instead? Seems like it might represent a race condition, and not some fatal programming error in dolt itself.

go/store/datas/database_common.go
andrew-wm-arthur: I tried to mirror the existing logic for `Commit()`, but yeah it's not great. will change

go/libraries/doltcore/doltdb/tag_meta.go
andrew-wm-arthur: I added user timestamp to the meta struct. There's currently no way to pass it from the CLI, but the model is there

PR #856, opened 2020-09-01
/mysql-client-tests/dotnet: Iterating on dotnet mysql client test

coffeegoddd:


mysql-client-tests/dotnet/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        catch (Exception ex)
+        {
+            Console.WriteLine(ex.ToString());
+        }
+
+        conn.Close();
+        return 0;
+    }
+
+    public static void SetupTest(MySqlConnection conn)
```
timsehn, line 47: Should we add the read queries here and examine the results?

timsehn: ie. https://github.com/liquidata-inc/dolt/blob/c28953970d9cbf91621f9bd9ee53eb31edfe7be2/mysql-client-tests/python/pymysql-test.py#L4

I'm less concerned about examining the results than I am running a couple more queries.

timsehn: I see. You're just missing the describe. This is probably fine. Up to you...

mysql-client-tests/mysql-client-tests.bats
```diff
@@ -48,3 +48,8 @@ teardown() {
     (cd $BATS_TEST_DIRNAME/c; make clean; make)
     $BATS_TEST_DIRNAME/c/mysql-connector-c-test $USER $PORT $REPO_NAME
 }
+
+@test "dotnet mysql connector" {
+    cd $BATS_TEST_DIRNAME/dotnet/
+    dotnet run -- $USER $PORT $REPO_NAME 3>&-
```
timsehn, line 54: I know this is mine but I would add a comment:

`# dotnet run uses output channel 3 which conflicts with bats so we pipe it to null`

timsehn (approved):
Love. Just a couple comments.

PR #857, opened 2020-09-01
Added new column defaults

Hydrocharged:


go/libraries/doltcore/schema/alterschema/addcolumn.go
```diff
@@ -83,23 +83,36 @@ func updateTableWithNewSchema(ctx context.Context, tbl *doltdb.Table, tag uint64
...
 		return doltdb.NewTable(ctx, vrw, newSchemaVal, rowData, &indexData)
 	}
 
 	me := rowData.Edit()
 
+	newSqlSchema, err := sqleSchema.FromDoltSchema(tblName, newSchema)
+	if err != nil {
+		return nil, err
+	}
+	columnIndex := []int{-1}
```
zachmu, line 96: Why is this a slice instead of just an int?

go/libraries/doltcore/schema/column.go
```diff
@@ -70,22 +71,25 @@ type Column struct {
 	// TypeInfo states the type of this column.
 	TypeInfo typeinfo.TypeInfo
 
+	// Default is the default value of this column.
```
zachmu, line 74: Add a note about this being the string representation of the sql.expression

go/libraries/doltcore/sqle/schema/schema.go
```diff
@@ -33,31 +35,53 @@ import (
...
-	cols := make([]*sql.Column, sch.GetAllCols().Size())
-
-	var i int
-	err := sch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
-		var innerErr error
-		cols[i], innerErr = doltColToSqlCol(tableName, col)
-		if innerErr != nil {
-			return true, innerErr
+// ApplyDefaults applies the default values to the given indices, returning the resulting row.
+func ApplyDefaults(ctx context.Context, doltSchema schema.Schema, sqlSchema sql.Schema, indicesOfColumns []int, oldRow row.Row) (row.Row, error) {
```
zachmu, line 39: s/oldRow/row/

go/libraries/doltcore/sqle/schema/schema.go
```diff
@@ -12,19 +12,21 @@
...
-package sqle
+package schema
 
 import (
 	"context"
 	"fmt"
 	"strconv"
 	"strings"
 
+	mysqlServer "github.com/liquidata-inc/go-mysql-server"
```
zachmu, line 23: import this as engine as elsewhere

go/libraries/doltcore/sqle/schema/schema.go
Hydrocharged: So I didn't wanna use `row` because we have a package named `row` and I didn't want that naming conflict

go/libraries/doltcore/schema/alterschema/addcolumn.go
Hydrocharged: Because `ApplyDefaults` wants a slice, so I just made it here. I could have this be `columnIndex := -1`, and then later do `ApplyDefault(..., []int{columnIndex}, ...)`

PR #860, opened 2020-09-03
Fixed column renaming breaking default values

Hydrocharged:


go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
```diff
@@ -55,6 +55,10 @@ func FmtColWithNameAndType(indent, nameWidth, typeWidth int, colName, typeStr st
 		}
```
zachmu, line 55: We should kill of this whole file, make sure it always goes through the engine. (This was written before the engine supported a bunch of things). Add a TODO / do it?

zachmu (approved):

PR #865, opened 2020-09-08
backtick minimally

andrew-wm-arthur:


timsehn:
Are reserved words handled somewhere else?

andrew-wm-arthur:
@timsehn no, good catch. I'll see if I can get those out of the parser

PR #867, opened 2020-09-09
Added Python SQLAlchemy test to mysql-client-tests

timsehn:


timsehn:
@oscarbatori here's a stub for you to test sqlalchemy. 

Seems fine to connect to and run basic queries. It's a lot more chatty on connection and issues a lot of COMMIT and ROLLBACK statements by default. I'm a little worried it will break us in weird ways given the above.

coffeegoddd (approved):
nice

PR #874, opened 2020-09-14
Added skip bats test for ref spec panic on diff

timsehn:


bats/diff.bats
```diff
@@ -411,3 +411,15 @@ SQL
...
+
+@test "diff with branch@commit spec does not panic" {
+    dolt add .
+    dolt commit -m table
+    dolt checkout -b test-branch
+    dolt sql -q "insert into test values (0, 0, 0, 0, 0, 0)"
+    dolt add test
+    dolt commit -m "added row"
+    FIRST_COMMIT=`dolt log | grep commit | cut -d " " -f 2 | tail -1`
+    skip "The blow panics."
```
zachmu, line 423: below, probably?

zachmu: put down the rolled up hundo tim

zachmu (approved):

bats/diff.bats
timsehn: what you mean?

PR #875, opened 2020-09-14
go: Changes to support some commit walks used in Dolthub diffs when the commits come from different repositories.

reltuk:


go/store/datas/commit.go
```diff
@@ -76,8 +76,9 @@ func NewCommit(ctx context.Context, value types.Value, parentsList types.List, m
 
 // FindCommonAncestor returns the most recent common ancestor of c1 and c2, if
 // one exists, setting ok to true. If there is no common ancestor, ok is set
-// to false.
-func FindCommonAncestor(ctx context.Context, c1, c2 types.Ref, vr types.ValueReader) (a types.Ref, ok bool, err error) {
+// to false. Refs of |c1| are dereferenced through |vr1|, while refs of |c2|
+// are dereference through |vr2|.
+func FindCommonAncestor(ctx context.Context, c1, c2 types.Ref, vr1, vr2 types.ValueReader) (a types.Ref, ok bool, err error) {
```
zachmu, line 81: This is fine, but it seems like the signature might be more self explanatory if the refs and their vrs were package together via some struct

zachmu (approved):
LG, but probably want a test or two for the new case you added (where the two DBs are different)

PR #877, opened 2020-09-14
validate ref strings when resolving ref specs

andrew-wm-arthur:
fix for https://github.com/liquidata-inc/dolt/pull/874


timsehn (approved):
LGTM

PR #878, opened 2020-09-14
Go MySQL client test

timsehn:


mysql-client-tests/go/go-sql-driver-mysql-test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "database/sql"
+import _ "github.com/go-sql-driver/mysql"
+
+func main() {
+    var user = os.Args[1]
+    var port = os.Args[2]
+    var db   = os.Args[3]
+
+    var dsn = user + "@tcp(127.0.0.1:" + port + ")/" + db
```
zachmu, line 14: fmt.Sprintf is more idiomatic

mysql-client-tests/go/go-sql-driver-mysql-test.go
```diff
@@ -0,0 +1,32 @@
...
+    database, err := sql.Open("mysql", dsn)
+
+    if err != nil {	
+        panic(err)
+    }
+
+    defer database.Close()
+
+    // Ping opens a connection
+    err = database.Ping()
```
zachmu, line 26: No SELECT *?

mysql-client-tests/go/go-sql-driver-mysql-test.go
```diff
@@ -0,0 +1,32 @@
...
+    if err != nil {
+        panic(err)
+    }
+
+    os.Exit(1)
```
zachmu, line 31: Probably want exit(0) in the happy case right?

zachmu (approved):
LG but seems incomplete. Should have some selects in there or something

PR #884, opened 2020-09-16
CREATE TABLE LIKE, commands to SQL, and tag removal

Hydrocharged:
PR is currently broken as I still have existing tests to fix, so although this will fail CI, you can look over the implementation itself. Brand new tests will come in a separate PR.


go/cmd/dolt/commands/tblcmds/mv.go
```diff
@@ -87,51 +85,18 @@ func (cmd MvCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
...
-	working, err = working.RenameTable(ctx, oldName, newName)
-	if err != nil {
-		verr = errhand.BuildDError("Unable to remove '%s'", oldName).Build()
-		return commands.HandleVErrAndExitCode(verr, usage)
-	}
-
-	return commands.HandleVErrAndExitCode(commands.UpdateWorkingWithVErr(dEnv, working), usage)
+	queryStr = fmt.Sprintf("%sRENAME TABLE `%s` TO `%s`;", queryStr, oldName, newName)
+
+	cli.CliOut = ioutil.Discard // display nothing on success
```
zachmu, line 96: This could (will) take a long time for big tables. Batch mode does print update stats as it goes, leave output alone

go/libraries/doltcore/sqle/root_database.go
```diff
@@ -0,0 +1,58 @@
...
+package sqle
+
+import (
+	"github.com/liquidata-inc/go-mysql-server/sql"
+
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/doltdb"
+)
+
+// RootDatabase in an implementation of sql.Database for root values. Does not expose any of the internal dolt tables.
+type RootDatabase struct {
```
zachmu, line 24: UserSpaceDatabase?

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -45,18 +45,18 @@ func TestCreateTable(t *testing.T) {
 	}{
 		{
 			name:          "Test create single column schema",
-			query:         "create table testTable (id int primary key comment 'tag:100')",
+			query:         "create table testTable (id int primary key)",
 			expectedTable: "testTable",
 			expectedSchema: dtestutils.CreateSchema(
-				schemaNewColumn(t, "id", 100, sql.Int32, true, schema.NotNullConstraint{})),
+				schemaNewColumn(t, "id", 4817, sql.Int32, true, schema.NotNullConstraint{})),
```
zachmu, line 51: These are kind of annoying. Think about making a comparison method for the test that ignores tags?

go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
```diff
@@ -23,22 +23,13 @@ import (
 	"github.com/liquidata-inc/dolt/go/libraries/doltcore/schema"
 )
 
-//TODO: This is a relic from before `SHOW CREATE TABLE` was implemented. We should remove this file altogether.
```
zachmu, line 26: Is this file still necessary? What's it for?

zachmu (approved):
Looks pretty reasonable, no major course corrections

go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
Hydrocharged: `dolt diff` still uses it to construct statements for diffs. As well, it's used when we auto-generate a schema and want to display it to the user. Can't use the other method since nothing has been written to root yet.

zachmu: The update / insert / deletes can't be handled by the engine, so fine to keep those. But you should figure out how to get all schema serialization to happen in the engine. If not as part of this PR, then as a followup. Having this code lying around when it diverges from the engine implementation is an active liability at this point.

zachmu (approved):
Changes look good, just one comment about making sure that all CREATE TABLE statements pass through the engine.

PR #888, opened 2020-09-16
optimizations

bheni:


go/libraries/doltcore/table/pipeline/procfunc_help.go
```diff
@@ -142,27 +142,41 @@ func InFuncForChannel(rowChan <-chan row.Row) InFunc {
...
+			case <-noMoreChan:
+				more = false
 			case r, ok := <-rowChan:
 				if ok {
 					ch <- RowWithProps{Row: r, Props: NoProps}
 				} else {
 					return
 				}
-			case <-time.After(100 * time.Millisecond):
-				// wake up and check stop condition
```
bheni, line 165: This creates a lot of garbage. select on the channel instead

go/libraries/doltcore/sqle/rows.go
```diff
@@ -45,20 +50,28 @@ func newRowIterator(tbl *DoltTable, ctx *sql.Context, partition *doltTablePartit
 	}
 
 	var mapIter types.MapIterator
-	var end types.Value = nil
+	var end types.LesserValuable
```
bheni, line 53: TupleValueSlice implements LesserValuable, but not Value

bheni:
@Hydrocharged would appreciate a look at FastNomsToSql in typeinfo.go

bheni:
Roughly a 40% performance improvement on a full table scan on the wikipedia-ngrams dataset:


```
Dataset: https://www.dolthub.com/repositories/Liquidata/wikipedia-ngrams
Command: GOMAXPROCS=1 time dolt --prof cpu  sql -r csv -q 'select bigram from bigram_counts' > out.csv

Before Changes:     1234.47 real      1186.96 user        19.17 sys
After Changes :      715.69 real       680.54 user        13.84 sys
```

Hydrocharged (changes requested):
So there are a few errors I've spotted in the `FastNomsToSql` function, but I'm not a fan of the function in general. It looks like the goal is to reduce the time it takes to get a dolt value to a sql (go) value, which is the purpose of `ConvertNomsValueToValue`. Beforehand I was going through all of the conversion checks, which is unnecessary as everything stored will be valid, so no need to recheck every value (which is what `FastNomsToSql` is doing). In addition, the `TypeInfo` interface is meant to be as complete of an interface as possible, so I'd rather not have multiple locations to edit for when new types are added that adapt existing noms types.

Rather than ask you to remove `FastNomsToSql` and implement my feedback, I've made a PR that you can merge into yours! 😄 I was going to run the query that you posted above (before and after) to get an idea of the performance impact of my PR, but the ngrams dataset is downloading so slowly that it won't complete in a sane amount of time. If you wouldn't mind testing it on your machine then that would be great 😃

https://github.com/dolthub/dolt/pull/911

bheni:
@Hydrocharged I've merged your changes.

Hydrocharged (approved):
So I didn't run tests locally on my last PR, and I wasn't paying attention that only the formatting and license tests ran on Github (it just says All Checks Passed, didn't see the 2 under it), so it broke the tests that were relying on `ConvertNomsValueToValue` to do all the additional junk that's not being used (which should be in `IsValid` anyway). I fixed all of those tests here, and ran it locally since Github still isn't running the full suite of tests for the PR.

https://github.com/dolthub/dolt/pull/915

PR #892, opened 2020-09-19
Andy/gc table files

andrew-wm-arthur:


reltuk:
Generally implementation in fsTablePersister looks reasonable, and integration with NBS seems reasonable.

Is there a good way to handle these optional interfaces? Should Prune be represented in `TableFileStoreOps` and returned from `SupportedOperations` here? Do we do something like `QueryInterface`? Or put query methods for each optional interface on the super interface? Do we just make a `ErrUnsupportedOperation` error value that methods on unsupported interfaces can return instead of panicing? @bheni and @zachmu for their opinions here.

bats/garbage_collection.bats
```diff
@@ -0,0 +1,23 @@
...
+    run dolt gc
+    [ "$status" -eq "0" ]
+
+    dolt sql <<SQL
+CREATE TABLE test (pk int PRIMARY KEY);
+INSERT INTO test VALUES (1),(2),(3),(4),(5);
+SQL
+
+    run dolt gc
+    [ "$status" -eq "0" ]
```
reltuk, line 22: Can we add smoke sanity tests that directory still works, `test` is still around, etc.?

go/cmd/dolt/commands/gc.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/liquidata-inc/dolt/go/cmd/dolt/cli"
+	"github.com/liquidata-inc/dolt/go/cmd/dolt/errhand"
+	"github.com/liquidata-inc/dolt/go/libraries/doltcore/env"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/argparser"
+	"github.com/liquidata-inc/dolt/go/libraries/utils/filesys"
+	"github.com/liquidata-inc/dolt/go/store/datas"
+)
+
+var gcDocs = cli.CommandDocumentationContent{
```
reltuk, line 28: This don't look right...

go/store/datas/garbage_collection.go
```diff
@@ -0,0 +1,31 @@
...
+
+package datas
+
+import (
+	"context"
+
+	"github.com/liquidata-inc/dolt/go/store/nbs"
+)
+
+func PruneTableFiles(ctx context.Context, db Database) error {
```
reltuk, line 23: Is this used?

go/store/nbs/file_table_persister.go
```diff
@@ -208,3 +211,52 @@ func (ftp *fsTablePersister) ConjoinAll(ctx context.Context, sources chunkSource
...
+		return err
+	}
+
+	err = ftp.fc.ShrinkCache()
+
+	if err != nil {
+		return err
+	}
+
+	for _, info := range fileInfos {
```
reltuk, line 230: If this encounters an `err`, should it accumulate them, and do as much work as possible?

go/store/nbs/store.go
```diff
@@ -1095,8 +1101,51 @@ func (nbs *NomsBlockStore) WriteTableFile(ctx context.Context, fileId string, nu
...
+	nbs.mm.LockForUpdate()
+	defer func() {
+		unlockErr := nbs.mm.UnlockForUpdate()
+
+		if err == nil {
+			err = unlockErr
+		}
+	}()
+
+	for {
```
reltuk, line 1118: The idea of this loop is that it's like a commit, but without a moving `root`?

reltuk (approved):
Looks good to me. Some comments and one question for my own education.

go/cmd/dolt/commands/gc.go
andrew-wm-arthur: yeah... good catch. The command is still hidden, so I'll leave it blank for now.

go/store/datas/garbage_collection.go
andrew-wm-arthur: yes, it's called from `go/cmd/commands/gc.go`

go/store/nbs/store.go
andrew-wm-arthur: Right, we want to persist any chunks added through `nbs.Put()`, specifically via `ValueStore.WriteValue()`.
I originally had `nbs.Commit(ctx, nbs.upstream.root, nbs.upstream.root)` but that deadlocks because we already have the lock at the top of this method.

go/store/nbs/file_table_persister.go
andrew-wm-arthur: good idea

PR #896, opened 2020-09-22
Moved everything over to SHOW CREATE TABLE and fixed diff panic

Hydrocharged:


go/libraries/doltcore/sqle/single_table_info_db.go
```diff
@@ -0,0 +1,138 @@
...
+	foreignKeys []doltdb.ForeignKey
+	parentSchs  map[string]schema.Schema
+}
+
+var _ sql.Database = (*SingleTableInfoDatabase)(nil)
+var _ sql.Table = (*SingleTableInfoDatabase)(nil)
+var _ sql.IndexedTable = (*SingleTableInfoDatabase)(nil)
+var _ sql.ForeignKeyTable = (*SingleTableInfoDatabase)(nil)
+
+func NewSingleTableDatabase(tableName string, sch schema.Schema, foreignKeys []doltdb.ForeignKey, parentSchs map[string]schema.Schema) *SingleTableInfoDatabase {
```
zachmu, line 41: Should have a comment about the intended use case of this DB implementation. Either here or on the struct

zachmu (approved):
LG!

Issue #654, opened 2020-05-04
Create table X like otherdatabase.Y panic


bheni:
```
$ dolt clone Liquidata/irs-soi
$ dolt clone Liquidata/us-congress
$ mkdir workspace
$ cd workspace
$ dolt init
$ cd ..
$ dolt sql --multi-db-dir ./
# Welcome to the DoltSQL shell.
# Statements must be terminated with ';'.
# "exit" or "quit" (or Ctrl-D) to exit.
> use workspace;
Database changed
workspace> create table allagi_with_rep_info like irs_soi.allagi;
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1db2c99]

goroutine 1 [running]:
github.com/liquidata-inc/go-mysql-server/sql/parse.tableSpecToSchema(0x0, 0x0, 0x0, 0x8, 0xc00108cb80, 0x144245c, 0x0)
	/Users/brian/dev/go-mysql-server/sql/parse/parse.go:767 +0x29
github.com/liquidata-inc/go-mysql-server/sql/parse.convertCreateTable(0xc000422230, 0xc00058c580, 0x2328f7b, 0x6, 0x0, 0xc00058a5c0)
	/Users/brian/dev/go-mysql-server/sql/parse/parse.go:624 +0x3f
github.com/liquidata-inc/go-mysql-server/sql/parse.convertDDL(0xc000422230, 0xc00058a540, 0x35, 0xc00058c580, 0x0, 0x0, 0x0, 0x0)
	/Users/brian/dev/go-mysql-server/sql/parse/parse.go:463 +0x293
github.com/liquidata-inc/go-mysql-server/sql/parse.convert(0xc000422230, 0x269d6c0, 0xc00058c420, 0xc00058a540, 0x35, 0x0, 0xc00058a540, 0x35, 0x0)
	/Users/brian/dev/go-mysql-server/sql/parse/parse.go:153 +0x130
github.com/liquidata-inc/go-mysql-server/sql/parse.Parse(0xc000422230, 0xc00058a440, 0x36, 0x0, 0x0, 0x0, 0x0)
	/Users/brian/dev/go-mysql-server/sql/parse/parse.go:125 +0x53b
[...]

PR #897, opened 2020-09-22
Fixed CREATE LIKE multi-db

Hydrocharged:
Fixes https://github.com/liquidata-inc/dolt/issues/654

### DIFF
--- a/bats/sql-create-tables.bats
+++ b/bats/sql-create-tables.bats
@@ -348,3 +348,25 @@ SQL
     [[ "$output" =~ "\`c1\` bigint DEFAULT 5 COMMENT 'hi'" ]] || false
     [[ "$output" =~ "PRIMARY KEY (\`pk\`)" ]] || false
 }
+
+@test "sql-create-tables: create like from other database" {
+    mkdir otherdb
+    cd otherdb
+    dolt init
+    dolt sql -q "CREATE TABLE othertable(pk bigint primary key, v1 bigint comment 'other')"
+    dolt add -A
+    dolt commit -m "some commit"
+    cd ..
+    mkdir workspace
+    cd workspace
+    dolt init
+    cd ..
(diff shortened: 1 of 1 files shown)


go/libraries/doltcore/sqle/sqlselect_test.go
```diff
@@ -709,7 +709,7 @@ var BasicSelectTests = []SelectTest{
 				"0o2rnf5pq2s1nq3hj3609e1lt0socuf1",
 				"billy bob",
 				"bigbillieb@fake.horse",
-				time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC),
+				time.Date(1970, 1, 1, 0, 0, 0, 0, &time.Location{}),
```
zachmu, line 712: Hmm, this might be the sign of a bug.

zachmu (approved):
LG, but I'm suspicious I broke something by deleting convert_dates. Let me know if I should dig in.

go/libraries/doltcore/sqle/sqlselect_test.go
Hydrocharged: So I I did a test by referencing the commit immediately before the `triggers 3` PR merge, and then the PR's commit, and this issue only popped up after the PR was merged. So it's definitely from that.

Hydrocharged:
It looks like some of the failing tests have two issues. 1) They're wrong. Naked variables only work with system variables, and they're being used here for user variables, 2) Changing them to user variables causes this line to be hit (https://github.com/liquidata-inc/go-mysql-server/blob/fff5edda928de40709eec9b7b4f0559464560dfe/sql/plan/set.go#L78) with the following error: `unsupported type for set: *expression.UserVar`

So I'll just leave this PR as-is until we get those fixed, since this PR is basically just updating the `go-mysql-server` version and adding a bats test of previously-broken behavior.

Hydrocharged:
A separate PR fixed many of the issues previously in this PR, so this is just a bats test now.

PR #901, opened 2020-09-28
Benchmarking tools first draft

oscarbatori:
See README.md for details.


oscarbatori:
Back to the drawing board. The problem here is to make it easy to benchmark Dolt between two commits, rather than recording it on DoltHub (though we should do that too).

PR #903, opened 2020-09-29
new pipeline framework

bheni:


bheni:
```
+------------+------+------+--------------+		
|            |  old |  new |  improvement |
+------------+------+------+--------------+
| Tabular    | 1773 | 1338 |     24.5%    |
| csv        |  671 |  480 |     28.4%    |
| json       | 1004 |  575 |     42.7%    |
+------------+------+------+--------------+
```

go/libraries/utils/pipeline/local_storage.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Put stores an item in local storage
+func (ls LocalStorage) Put(id int, val interface{}) {
+	ls[id] = val
+}
+
+// GetLocalStorage retrieves the LocalStorage from the context.  This only works if the context was generated
+// by the pipeline package when starting the pipeline
+func GetLocalStorage(ctx context.Context) LocalStorage {
```
zachmu, line 37: This whole thing is kind of gross. It's only used one place outside tests. Can we get rid of it?

zachmu (approved):
Generally looks quite good, just a few comments.

go/libraries/utils/pipeline/local_storage.go
bheni: Yeah.  it was mostly something I was playing with as a possible way to optimize in the future.  I don't have a case where it really makes a difference atm.

bheni: added a todo.  will revisit post TG.

PR #904, opened 2020-09-30
Added trigger functionality to Dolt

Hydrocharged:


andrew-wm-arthur:
What are the tradeoffs between adding triggers to the `dolt_schemas` table vs creating a separate system table for them?
I'm concerned about the compatibility between existing `dolt_schemas` tables and new versions with the updated schema. 
Also we need tests for diffing/merging triggers. Specifically, the cycle detection we do on `CreateTrigger()` should throw a conflict when we merge triggers across branches.

go/libraries/doltcore/sqle/schema_table.go
```diff
@@ -15,65 +15,174 @@
...
+
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/row"
+	"github.com/dolthub/dolt/go/store/types"
 )
 
 // The fixed schema for the `dolt_schemas` table.
 func SchemasTableSchema() sql.Schema {
```
andrew-wm-arthur, line 29: We need to generate this schema as a Dolt `schema.Schema` in order to hard-code its tags. Currently it's are generated via the deterministic tag method, which is not stable when modifying the schema. See the [query catalog table](https://github.com/dolthub/dolt/blob/4f20d5f78c80885bc0a26a5a51d85cee6b5c14d7/go/libraries/doltcore/sqle/query_catalog_table.go#L29) for an example.

go/libraries/doltcore/doltdb/system_table.go
```diff
@@ -204,15 +204,17 @@ const (
...
 	SchemasTablesTypeCol = "type"
-
-	// // The name of the database entity.
+	// The name of the database entity.
 	SchemasTablesNameCol = "name"
 	// The schema fragment associated with the database entity.
 	// For example, the SELECT statement for a CREATE VIEW.
 	SchemasTablesFragmentCol = "fragment"
+	// The name of the index that is on the table.
+	SchemasTablesIndexName = "fragment_name"
```
andrew-wm-arthur, line 217: We also need to update the next `const` block with new tags. However, the current tags in that block are currently incorrect/unused. See the comment in `sqle/schema_table.go` for details.

andrew-wm-arthur (commented):
Per our conversation, there are some compatibility concerns stemming from the existing `dolt_schemas` table schema implementation. 
Also, we need some merge/diff tests. Trigger ordering and ref-cycle detection could be tricky in the merge case.

go/libraries/doltcore/sqle/schema_table.go
```diff
@@ -15,65 +15,174 @@
...
+	tbl, found, err := db.GetTableInsensitiveWithRoot(ctx, root, doltdb.SchemasTableName)
 	if err != nil {
 		return nil, err
 	}
+	var rowsToAdd []sql.Row
 	if found {
-		return tbl.(*WritableDoltTable), nil
+		schemasTable := tbl.(*WritableDoltTable)
+		// Old schemas table does not contain the `id` column.
+		if !tbl.Schema().Contains(doltdb.SchemasTablesIdCol, doltdb.SchemasTableName) {
```
zachmu, line 52: Suggest extracting a method for this migration logic

zachmu (approved):
LG, just add the logic about using predefined tags in the system space for the schemas table like we talked offline, and this is good to go

Hydrocharged:
@andrew-wm-arthur Since this all resides in the `dolt_schemas` table anyway, I actually don't think we need to worry about catching ordering and cycle detection issues. It's completely valid to write directly into the `dolt_schemas` table, including writing a trigger that would cause those issues, and thus it can be fixed just the same.

On another note, while writing a merge test, I found an unrelated issue with how we handle conflicts. Rather than fixing that and possibly delaying getting triggers in, I'll just merge this and fix that in another PR.

PR #909, opened 2020-10-02
use tr to lowercase output instead of {output,,}

andrew-wm-arthur:
lowercasing via parameter expansion `${output,,}` is only supported on Bash 4+. I switched to using `tr` so I could run the tests locally.


Hydrocharged (approved):
image

PR #910, opened 2020-10-05
Adding "Garbage Collection Generation" to manifest file

andrew-wm-arthur:
This new manifest field will support `NomsBlockStore` garbage collection and protect against `NBS` corruption. Storing `gcGen` in the manifest will support deleting chunks from an `NBS` in a safe way. `NBS` instances that see a different `gcGen` than they saw when they last read the manifest will error and require clients to re-attempt their write.

`NBS` will now have three forms of write errors (not including IO errors or other kinds of unexpected errors):
- `nbs.errOptimisticLockFailedTables`: Another writer landed a manifest update since the last time we read the manifest. The root chunk is unchanged and the set of chunks referenced in the manifest is either the same or has strictly grown. Therefore the NBS can handle this by rebasing on the new set of tables in the manifest and re-attempting to add the same set of novel tables.
- `nbs.errOptimisticLockFailedRoot`: Another writer landed a manifest update that includes a new root chunk. The set of chunks referenced in the manifest is either the same or has strictly grown, but it is not know which chunk are reachable from the new root chunk. The NBS has to pass this value to the client and let them decide. If the client is a `datas.database` it will attempt to rebase, read the head of the dataset it is committing to, and execute its `mergePolicy` (Dolt passes a noop `mergePolicy`). 
- `chunks.ErrGCGenerationExpired`: This is similar to a moved root chunk, but with no guarantees about what chunks remain in the ChunkStore. Any information from `CS.Has(ctx, chunk)` is now stale. Writers must rewrite all data to the chunkstore.


bats/compatibility/runner.sh
```diff
@@ -64,8 +64,12 @@ function test_dolt_version() {
   echo testing "$ver" at "$bin"
   PATH="`pwd`"/"$bin":"$PATH" setup_repo "$ver"
 
+# Changes to the NBS manifest broke forward compatibility when multiple clients with different
+# versions interact with the same repo instance. However, in a more realistic setting of
+# clients interacting with a repo through a remote, compatibility is not broken.
+# todo: update compatibility tests to have different client versions interact through a remote.
```
reltuk, line 70: There are two types of "remotes" – one is `doltremoteapi`, where the comment text seems to be true, and one is where each client actually has write access to the underlying storage assets, and in particular the manifest. Both `file:///` and `aws:` remotes are the second type. This comment isn't true for those remotes...

go/cmd/dolt/dolt.go
```diff
@@ -154,6 +155,14 @@ func runMain() int {
 	warnIfMaxFilesTooLow()
 
 	ctx := context.Background()
+
+	err := nbs.MaybeMigrateFileManifest(ctx, dbfactory.DoltDataDir)
```
reltuk, line 159: As mentioned in chat, we can do this optimistically (i.e., anytime a new dolt client runs against a directory.) But we don't _have_ to do this until the first use of GC version is needed (first GC run I guess...). On the balance, if we expect GC to be commonly used, the tradeoff doesn't matter too much. If we expect GC to remain experimental for a while and we want to ease into the impacts of introducing it, we might want to take the other approach...

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -91,6 +91,7 @@ func NewDoltChunkStoreFromPath(ctx context.Context, nbf *types.NomsBinFormat, pa
 		return nil, ErrInvalidDoltSpecPath
 	}
 
+	// todo:
```
reltuk, line 94: Dangling...

go/store/nbs/store.go
```diff
@@ -51,7 +51,8 @@ var ErrFetchFailure = errors.New("fetch failed")
 
 const (
 	// StorageVersion is the version of the on-disk Noms Chunks Store data format.
-	StorageVersion = "4"
+	// todo: how to handle discrepancies between file manifest and dynamo manifest
```
reltuk, line 54: Dangling?

reltuk (commented):
Couple comments I landed in chat as well.

go/libraries/doltcore/remotestorage/chunk_store.go
andrew-wm-arthur: this was just to annotate the line below it without changing the blame. might have overthought that

go/store/nbs/file_manifest.go
```diff
@@ -318,21 +513,18 @@ func (fm fileManifest) Update(ctx context.Context, lastLock addr, newContents ma
...
+	// this is where we assert that gcGen is unchanged
+	err = vd(upstream, newContents)
 
 	if err != nil {
 		return manifestContents{}, err
 	}
 
-	return newContents, nil
-}
+	err = os.Rename(tempManifestPath, manifestPath)
```
andrew-wm-arthur, line 523: @zachmu

go/store/nbs/file_manifest.go
```diff
@@ -33,21 +34,119 @@ import (
...
+	contents.gcGen = contents.lock
+
+	_, err = updateWithParseWriterAndChecker(ctx, dir, fm5.writeManifest, fm4.parseManifest, check, contents.lock, contents, nil)
+
+	if err != nil {
+		return false, nil
+	}
+
+	// clear manifest cache
+	makeGlobalCaches()
```
reltuk, line 106: These seems likely to violate concurrency expectations around global caches? In particular, was this behind a `sync.Once` before?

andrew-wm-arthur: Yes, all other instances were behind a sync.Once. I'll remove this

go/store/nbs/file_manifest.go
```diff
@@ -158,70 +415,14 @@ func (fm fileManifest) ParseIfExists(ctx context.Context, stats *Stats, readHook
...
-		root:  hash.Parse(slices[3]),
-		specs: specs,
-	}, nil
-}
-
-func (fm fileManifest) Update(ctx context.Context, lastLock addr, newContents manifestContents, stats *Stats, writeHook func() error) (mc manifestContents, err error) {
-	t1 := time.Now()
-	defer func() { stats.WriteManifestLatency.SampleTimeSince(t1) }()
-
+func updateWithParseWriterAndChecker(_ context.Context, dir string, wr manifestWriter, pr manifestParser, vd manifestChecker, lastLock addr, newContents manifestContents, writeHook func() error) (mc manifestContents, err error) {
```
reltuk, line 418: `write`, `parse` and `validate` instead of `wr`, `pr`, `vd`?

reltuk (approved):
Seems totally reasonable. A few comments.

PR #911, opened 2020-10-05
Daylon's suggestions for bheni perf PR

Hydrocharged:


go/libraries/doltcore/schema/typeinfo/varstring.go
```diff
@@ -83,17 +83,9 @@ func CreateVarStringTypeFromParams(params map[string]string) (TypeInfo, error) {
 // ConvertNomsValueToValue implements TypeInfo interface.
 func (ti *varStringType) ConvertNomsValueToValue(v types.Value) (interface{}, error) {
 	if val, ok := v.(types.String); ok {
-		strVal, err := ti.sqlStringType.Convert(string(val))
```
bheni, line 86: Profiling these changes, but looks good overall.  This line was the worst performer, so as long as it's gone I imagine this should be fine.

bheni (approved):
perf looks good.

PR #913, opened 2020-10-05
squash merge

bheni:


timsehn:
BATS tests lgtm. Like how you updated one to make sure both commits show up in the log.

PR #920, opened 2020-10-09
go/go.mod: Adopt dolthub/fslock fork. Forked version uses Open(RDRW) for lock file on *nix, which works on NFS.

reltuk:


bheni (approved):
lgtm

PR #921, opened 2020-10-09
fix gcp remotes

bheni:


OliverEvans96:
@bheni , thanks for looking at this so quickly! At first glance, this branch basically works for me. I do have a few notes, though, related to GCP's slightly weird handling of `/`:

1. `gs://my-super-secret-dolt-gcs-test/` and `gs://my-super-secret-dolt-gcs-test` (with and without trailing slash) both work as remotes. I can upload and download successfully.  That's great. However, it creates an unnecessary `/` directory where the dolt files are stored. I would recommend avoiding this if possible - just dump the files straight into the bucket. 

I can't remember the details, but I recall running into a similar issue where an application created a `/` directory for me, and I couldn't use rclone in that bucket because didn't know how to handle the `/` directory.

2. `gs://my-super-secret-dolt-gcs-test/subdir1/` seems to work. But again, I now have a `subdir1` directory inside of my `/` directory. I would rather have `subdir1` in the bucket root.

3. `gs://my-super-secret-dolt-gcs-test/subdir` (no slash) also works, which is a bit surprising given that `/subdir` is not a directory. Rather, it's just a prefix to the dolt files which are dumped into the `/` directory. e.g. `/subdirmanifest`, `/subdirdn6cle7i9rv9q6d14ik231c7cgk9hsh3`. That's clearly not what I want, but I can apparently read/write dolt repos this way.

bheni:
Thanks Oliver for giving it a try. I am putting together tests, and will address this before the PR gets merged.

bheni:
@OliverEvans96 I've updated with some testing (still need more) and the path issues should also be addressed (You'll need to delete any repos you pushed to GCS previously though as the it won't look for the old paths).

```
~/datasets/bad-words>gsutil mb gs://dolt-test-bucket0
Creating gs://dolt-test-bucket0/...
~/datasets/bad-words>gsutil mb gs://dolt-test-bucket1
Creating gs://dolt-test-bucket1/...

~/datasets/bad-words>dolt remote add bucket0 gs://dolt-test-bucket0
~/datasets/bad-words>dolt remote add bucket1 gs://dolt-test-bucket1/
~/datasets/bad-words>dolt remote add slash_prefix gs://dolt-test-bucket1/prefix/with/slash/
~/datasets/bad-words>dolt remote add no_slash_prefix gs://dolt-test-bucket0/prefix/no/slash

~/datasets/bad-words>gsutil ls -R gs://dolt-test-bucket0
gs://dolt-test-bucket0/5rvg2rgbmhijnjk9t5i8rmnit0np0t96
gs://dolt-test-bucket0/m14sn7t4c0dqoderfpud8fgnkndhla3l
gs://dolt-test-bucket0/manifest
gs://dolt-test-bucket0/prefix/:

gs://dolt-test-bucket0/prefix/no/:

gs://dolt-test-bucket0/prefix/no/slash/:
gs://dolt-test-bucket0/prefix/no/slash/5rvg2rgbmhijnjk9t5i8rmnit0np0t96
gs://dolt-test-bucket0/prefix/no/slash/m14sn7t4c0dqoderfpud8fgnkndhla3l
gs://dolt-test-bucket0/prefix/no/slash/manifest

~/datasets/bad-words>gsutil ls -R gs://dolt-test-bucket1
gs://dolt-test-bucket1/5rvg2rgbmhijnjk9t5i8rmnit0np0t96
gs://dolt-test-bucket1/m14sn7t4c0dqoderfpud8fgnkndhla3l
gs://dolt-test-bucket1/manifest
gs://dolt-test-bucket1/prefix/:

gs://dolt-test-bucket1/prefix/with/:

gs://dolt-test-bucket1/prefix/with/slash/:
gs://dolt-test-bucket1/prefix/with/slash/5rvg2rgbmhijnjk9t5i8rmnit0np0t96
gs://dolt-test-bucket1/prefix/with/slash/m14sn7t4c0dqoderfpud8fgnkndhla3l
gs://dolt-test-bucket1/prefix/with/slash/manifest
```

go/libraries/doltcore/dbfactory/gs.go
```diff
@@ -17,9 +17,11 @@ package dbfactory
 import (
```
zachmu, line 17: This file should be gcs.go, right?

zachmu: And the struct / methods should be GCSFactory, right?

go/libraries/doltcore/dbfactory/gs.go
```diff
@@ -38,7 +40,8 @@ func (fact GSFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat, ur
 		return nil, err
 	}
 
-	gcsStore, err := nbs.NewGCSStore(ctx, nbf.VersionString(), urlObj.Host, urlObj.Path, gcs, defaultMemTableSize)
+	bs := blobstore.NewGCSBlobstore(gcs, urlObj.Host, urlObj.Path)
```
zachmu, line 43: Here and elsewhere, I find embedded acronyms to be more readable if you only capitalize the first letter. So NewGcsBlobstore

go/libraries/doltcore/dbfactory/gs.go
```diff
@@ -48,3 +51,28 @@ func (fact GSFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat, ur
 
 	return db, err
 }
+
+// LocalBSFactory is a DBFactory implementation for creating a local filesystem blobstore backed databases for testing
+type LocalBSFactory struct {
```
zachmu, line 56: Probably should go in its own file?

go/store/blobstore/gcs.go
```diff
@@ -55,11 +63,12 @@ func (bs *GCSBlobstore) Exists(ctx context.Context, key string) (bool, error) {
...
 // its version
 func (bs *GCSBlobstore) Get(ctx context.Context, key string, br BlobRange) (io.ReadCloser, string, error) {
-	oh := bs.bucket.Object(bs.prefix + key)
+	absKey := path.Join(bs.prefix, key)
+	oh := bs.bucket.Object(absKey)
 	attrs, err := oh.Attrs(ctx)
 
 	if err == storage.ErrObjectNotExist {
-		return nil, "", NotFound{key}
+		return nil, "", NotFound{"gs://" + path.Join(bs.bucketName, absKey)}
```
zachmu, line 71: gcs://?

zachmu (approved):
Just some comments on naming and organization.

Why do we use GS in some places and GCS in others? Seems like it should always be GCS?

Also see my comments about camel casing using embedded acronyms. This gets pretty bad in the db factory code

go/libraries/doltcore/dbfactory/gs.go
bheni: Google uses the scheme gs:// for their GCS urls.  These factories refer to the schemes.  It is a factory for creating databases backed by the GS scheme.

go/libraries/doltcore/dbfactory/gs.go
bheni: consistent case is the preferred go style: https://github.com/golang/go/wiki/CodeReviewComments

```Initialisms
Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized "words", use for example "xmlHTTPRequest" or "XMLHTTPRequest".

This rule also applies to "ID" when it is short for "identifier" (which is pretty much all cases when it's not the "id" as in "ego", "superego"), so write "appID" instead of "appId".

Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.```

go/store/blobstore/gcs.go
bheni: Didn't you work at google?  `gs://` is the scheme used to refer to files in GCS.

go/libraries/doltcore/dbfactory/gs.go
bheni: probably.

PR #925, opened 2020-10-12
benchmark tools

oscarbatori:


benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
+## Background
+This folder contains a toolkit for benchmarking Dolt SQL performance. [`sysbench`](https://github.com/akopytov/sysbench) is an industry standard tool for benchmarking database performance, particularly MySQL. This directory contains a set of tools for benchmarking Dolt using `sysbench` in a way that should make it convenient for Dolt contributors to measuret the performance benefits their changes deliver.
```
reltuk, line 2:
```suggestion
This folder contains a toolkit for benchmarking Dolt SQL performance. [`sysbench`](https://github.com/akopytov/sysbench) is an industry standard tool for benchmarking database performance, particularly MySQL. This directory contains a set of tools for benchmarking Dolt using `sysbench` in a way that should make it convenient for Dolt contributors to measure the performance benefits their changes deliver.
```

benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Architecture
+The basic goal is to run these tools is, given a commit, to:
```
reltuk, line 5:
```suggestion
The basic goal of these tools is, given a commit, to:
```

benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
...
+- run Dolt SQL server
+- connect to it using the [`sysbench` MySQL driver](https://github.com/akopytov/sysbench/tree/master/src/drivers/mysql)
+- execute the benchmarks specified
+
+In order to to do this in a repeatable way we build Dolt at the specified commit inside a Docker container. We then launch a separate container, install `sysbench`, mount the binary we produced in the build step, and execute the benchmarks specified inside that container for maximum repeatability.
```
reltuk, line 10:
```suggestion
In order to to do this in a repeatable way we build Dolt at the specified commit inside a Docker container. We then launch a separate container, install `sysbench`, mount the binary we produced in the build step, and execute the specified benchmarks from inside that container.
```

benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+This will tell `run_benchmarks.sh` to take the following actions:
+- build a copy of Dolt at the current commit (noting in the binary name if it is dirty)
+- clone Dolt, checkout the specified commit, and build a binary
+- run the specified `sysbench` tests with both binaries
+
+If the result of `$(git rev-parse HEAD)` is `c84f34075e55a8e3322f584986ecf808076b309c`, and the repo is dirty then we will get the following:
+```
+output/c84f34075e55a8e3322f584986ecf808076b309c.csv
```
reltuk, line 30: Should this have `-dirty` on it?

benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
...
+  --mysql-db=test_db \
+  --mysql-user=root \
+  --mysql-host=127.0.0.1 \
+run
+```
+
+Suppose we wanted to define custom behavior, or some custom output method, then we would define `sysbench_scripts/my_custom_test.lua`, and then run:
+```
+sysbench \
+  sysbench_scripts/lua/my_custom_test.lua \
```
reltuk, line 66:
```suggestion
  sysbench_scripts/my_custom_test.lua \
```

benchmark-tools/README.md
```diff
@@ -0,0 +1,138 @@
...
+  --table-size=1000000 \
+  --db-driver=mysql \
+  --mysql-db=test_db \
+  --mysql-user=root \
+  --mysql-host=127.0.0.1 \
+run
+```
+
+This passes the script `sysbench_scripts/lua/my_custom_test.lua` which is then executed and measured. Here is an example, `bulk_insert`, taken from the [`sysbench` core out-of-the-box SQL benchmarks](https://github.com/akopytov/sysbench/tree/master/src/lua):
```
reltuk, line 75:
```suggestion
This passes the script `sysbench_scripts/my_custom_test.lua` which is then executed and measured. Here is an example, `bulk_insert`, taken from the [`sysbench` core out-of-the-box SQL benchmarks](https://github.com/akopytov/sysbench/tree/master/src/lua):
```

benchmark-tools/README.md
oscarbatori: Yes, that's a good shout!

benchmark-tools/README.md
oscarbatori: I think we in fact want to separate `lua` from the scripts directory since we may want to define shell wrappers around the sysbench scripts once we retire the Python code for managing the output.

PR #928, opened 2020-10-12
Andy/gc chunks

andrew-wm-arthur:
Implements garbage collection by traversing a `Database` from its root chunk and coping all reachable chunks to a new set of NBS tables.

While "garbage collection generation" will protect the NBS from corruption by out-of-process writers, GC is not currently thread safe for concurrent use in-process. Getting to online GC will require work around protecting in-progress writes that are not yet reachable from the root chunk.

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -23,15 +23,15 @@ import (
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/dbfactory"
 	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
-	"github.com/dolthub/dolt/go/store/chunks"
-	"github.com/dolthub/dolt/go/store/spec"
-	"github.com/dolthub/dolt/go/store/types/edits"
-
 	"github.com/dolthub/dolt/go/libraries/utils/filesys"
 	"github.com/dolthub/dolt/go/libraries/utils/pantoerr"
+	"github.com/dolthub/dolt/go/libraries/utils/set"
+	"github.com/dolthub/dolt/go/store/chunks"
 	"github.com/dolthub/dolt/go/store/datas"
 	"github.com/dolthub/dolt/go/store/hash"
+	"github.com/dolthub/dolt/go/store/spec"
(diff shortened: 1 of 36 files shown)


go/store/nbs/gc_copier.go
```diff
@@ -0,0 +1,120 @@
...
+	return &gcCopier{
+		mtSize: tableSize,
+		tables: newTableSet(ftp),
+		ftp:    ftp,
+		tmpDir: tmpDir,
+		stats:  NewStats(),
+	}, nil
+}
+
+func (gcc *gcCopier) addChunk(ctx context.Context, h addr, data []byte) bool {
```
reltuk, line 79: For the future, I would love us flesh out `CmpChunkTableWriter` so we could intermittently flush it to a `Writer` and it didn't have to hold all chunk data in memory, and then to use that write path here instead of mem table / table set.

reltuk (approved):
Generally looks great. Other than the `errgroup` PR I opened against this one, just a few comments.

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -909,6 +909,113 @@ func (ddb *DoltDB) DeleteTag(ctx context.Context, tag ref.DoltRef) error {
...
+	for i, h := range uncommitedVals {
+		v, err := ddb.db.ReadValue(ctx, h)
+		if err != nil {
+			return err
+		}
+		if v == nil {
+			return fmt.Errorf("empty value for value hash %s", h.String())
+		}
+
+		ds, err := ddb.db.GetDataset(ctx, fmt.Sprintf("tmp/%d", time.Now().UnixNano()))
```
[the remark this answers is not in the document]
andrew-wm-arthur: we don't have a concurrency model for GC anyway, but yes, we need to think about how to safely preserve tmp roots

go/store/nbs/gc_copier.go
andrew-wm-arthur: Yeah that's a good call. The memory footprint of GC isn't acceptable right now, which needs to be fixed given a major application of gc is to facilitate `dolt push` when we're overloading `doltremoteapi`

PR #932, opened 2020-10-13
/go/cmd/dolt/commands: Help Text Fix

andrew-wm-arthur:


bats/no-repo.bats
```diff
@@ -53,6 +53,34 @@ teardown() {
...
 }
 
+@test "check all commands for valid help text" {
+    # pipe all commands to a file
+    # cut -s suppresses the line if it doesn't contain the delim
+    dolt | cut -f 1 -d " - " -s | sed "s/ //g" > all.txt
+
+    # filter out commands without "-h"
+    cat all.txt \
+        | sed "s/remote//g"    \
```
zachmu, line 63: Odd that these don't have help text. Probably should have a skipped test that doesn't omit these commands

zachmu (approved):
LG, just one comment

bats/no-repo.bats
andrew-wm-arthur: remote actually does have help text, I removed that from the filter. The rest are subcommand handlers, so if you give them `-h` or anything invalid they'll print their usage

PR #938, opened 2020-10-15
go/store/{nbs,chunks}: Convert some core methods to provide results in callbacks. Convert some functions to use errgroup.

reltuk:


go/store/nbs/table_set.go
```diff
@@ -132,70 +133,82 @@ func (ts tableSet) get(ctx context.Context, h addr, stats *Stats) ([]byte, error
...
+				return false, nil
 			}
 		}
-
-		return true
+		return true, nil
 	}
 
-	return f(ts.novel) && f(ts.upstream)
+	remaining, err := f(ts.novel)
```
reltuk, line 162: A potential downside of this errgroup change is that now we have synchronous and asynchronous errors reported differently on paths like this one. It makes things like this here a little awkward.

go/store/nbs/table_reader.go
```diff
@@ -764,141 +746,126 @@ func (tr tableReader) readAtOffsetsWithCB(
...
-
-			if batch == nil {
-				batch = make(offsetRecSlice, 1)
-				batch[0] = offsetRecords[i]
-				readStart = rec.offset
-				readEnd = readStart + uint64(length)
-				i++
-				continue
+) error {
+	batches := toReadBatches(offsetRecords, tr.blockSize)
```
reltuk, line 845: Calculating these before spawning goroutines did not seem like a big loss and was a decent simplification. Creating the little `readBatch` helper type seemed like a win.

go/store/types/value_store.go
```diff
@@ -253,34 +252,23 @@ func (lvs *ValueStore) ReadManyValues(ctx context.Context, hashes hash.HashSlice
...
-			ae.SetIfError(err)
-		}()
-
-		var err error
-		for c := range foundChunks {
-			if err != nil {
-				continue // continue to drain even if there is an error
+		mu := new(sync.Mutex)
+		var decodeErr error
+		err := lvs.cs.GetMany(ctx, remaining, func(c *chunks.Chunk) {
```
reltuk, line 257: A potential downside of this callback change is that callers have to be more aware of thread safety. Here I introduced a race on the writes to `decodeErr` and `foundValues` before realizing it and adding the mutex.

go/store/nbs/table_set.go
bheni: Could always do something like:

```
if err != nil {
    eg.Go(func() error {
        return err
    }

    return false
}```

if you don't like the mix... But that seems worse.

File: go/libraries/doltcore/remotestorage/chunk_store.go [hunk 4]
```diff
@@ -135,22 +137,19 @@ func (dcs *DoltChunkStore) getRepoId() *remotesapi.RepoId {
...
+	} else {
 		return chunks.EmptyChunk, nil
 	}
 }
 
-func (dcs *DoltChunkStore) GetMany(ctx context.Context, hashes hash.HashSet, foundChunks chan<- *chunks.Chunk) error {
+func (dcs *DoltChunkStore) GetMany(ctx context.Context, hashes hash.HashSet, found func(*chunks.Chunk)) error {
 	ae := atomicerr.New()
 	wg := &sync.WaitGroup{}
 	foundCmp := make(chan nbs.CompressedChunk, 1024)
```
bheni, line 155: Thoughts on pushing this to the caller and allowing them to setup and use (on not) a channel in their callback as needed.  In cases where a channel isn't needed from a perf perspective?

bheni: I guess by kicking off a go routine here and using the channel it means that we don't have to do that everywhere this is called.

reltuk: Yeah, not using the channel is definitely nicer:

https://github.com/dolthub/dolt/pull/938/commits/89780ffd88a2dbd1b95625bed036e38e7b198104

Still partially highlights the subtlety around threading though. These interfaces don't take the `*errgroup.Group`, so maybe calls to the callback should be serialized.

PR #939, opened 2020-10-15
Added new indexes overwriting auto-generated indexes

Hydrocharged:


bats/index.bats
```diff
@@ -2316,3 +2316,89 @@ SQL
...
+  FOREIGN KEY (parent_value)
+    REFERENCES parent(v1)
+);
+INSERT INTO parent VALUES (1, 1), (2, 2), (3, 3), (4, NULL), (5, 5), (6, 6), (7, 7);
+INSERT INTO child VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5', 5);
+INSERT INTO child_idx VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5', 5);
+INSERT INTO child_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', NULL), ('5', 5);
+INSERT INTO child_non_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5', 5);
+SQL
+    dolt sql -q "CREATE INDEX abc ON child (parent_value);"
```
zachmu, line 2358: As part of this test, after creating the new key, you should do some inserts / deletes that are blocked by the foreign key constraint to verify that it still works. Also some inserts to verify that the unique constraint is being honored.

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -136,11 +136,11 @@ func (enc encodedTypeInfo) decodeTypeInfo() (typeinfo.TypeInfo, error) {
...
-	Name    string   `noms:"name" json:"name"`
-	Tags    []uint64 `noms:"tags" json:"tags"`
-	Comment string   `noms:"comment" json:"comment"`
-	Unique  bool     `noms:"unique" json:"unique"`
-	Hidden  bool     `noms:"hidden,omitempty" json:"hidden,omitempty"`
+	Name             string   `noms:"name" json:"name"`
+	Tags             []uint64 `noms:"tags" json:"tags"`
+	Comment          string   `noms:"comment" json:"comment"`
+	Unique           bool     `noms:"unique" json:"unique"`
+	IsNotUserDefined bool     `noms:"hidden,omitempty" json:"hidden,omitempty"` // Was previously named Hidden, do not change noms name
```
zachmu, line 143: No negations in field names. How about "implicit" or "system defined"

go/libraries/doltcore/schema/index.go
```diff
@@ -146,17 +162,77 @@ func (ix *indexImpl) Schema() Schema {
 	}
 }
 
+func (ix *indexImpl) VerifyMap(ctx context.Context, iter types.MapIterator, nbf *types.NomsBinFormat) error {
```
zachmu, line 165: Method comment plz

go/libraries/doltcore/sqle/tables.go
```diff
@@ -934,14 +958,29 @@ func parseFkReferenceOption(refOp sql.ForeignKeyReferenceOption) (doltdb.Foreign
...
+	ctx *sql.Context,
+	table *doltdb.Table,
+	indexName string,
+	using sql.IndexUsing,
+	constraint sql.IndexConstraint,
+	columns []sql.IndexColumn,
+	isUserDefined bool,
+	comment string,
+) (
+	newTable *doltdb.Table,
```
zachmu, line 971: This return type is getting a little chonky... I would take the time to declare a return struct instead. (leave the error as a separate return result)

go/libraries/doltcore/sqle/tables.go
```diff
@@ -966,43 +1005,72 @@ func createIndexForTable(ctx *sql.Context, table *doltdb.Table, indexName string
...
 	}
 	if !doltdb.IsValidTableName(indexName) {
-		return nil, nil, nil, fmt.Errorf("invalid index name `%s` as they must match the regular expression %s", indexName, doltdb.TableNameRegexStr)
+		return nil, nil, nil, nil, fmt.Errorf("invalid index name `%s` as they must match the regular expression %s", indexName, doltdb.TableNameRegexStr)
+	}
+
+	// if an index was already created for the column set but was not generated by the user then we replace it
+	replacingIndex := false
+	existingIndex, ok := sch.Indexes().GetIndexByColumnNames(realColNames...)
+	if ok && !existingIndex.IsUserDefined() {
```
zachmu, line 1014: Make sure we have tests of attempting to create a duplicate index when it was user-defined

bats/index.bats
```diff
@@ -2316,3 +2316,89 @@ SQL
...
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "pk,parent_value" ]] || false
+    [[ "$output" =~ "5,5" ]] || false
+    [[ "${#lines[@]}" = "2" ]] || false
+    run dolt sql -q "SELECT * FROM child_non_unq WHERE parent_value = 5" -r=csv
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "pk,parent_value" ]] || false
+    [[ "$output" =~ "5,5" ]] || false
+    [[ "${#lines[@]}" = "2" ]] || false
+}
```
zachmu, line 2404: You also want tests that the opposite order works: when you create an index first, then add a foreign key constraint. Everything should work the same.

zachmu (approved):
LG, just some testing and structure comments

go/libraries/doltcore/sqle/tables.go
Hydrocharged: It's in both the go and bats tests

bats/index.bats
Hydrocharged: This is in this test.
```sql
CREATE TABLE child_idx (
  pk varchar(10) PRIMARY KEY,
  parent_value bigint,
  INDEX (parent_value),
  FOREIGN KEY (parent_value)
    REFERENCES parent(v1)
);
```
When they're both declared in a `CREATE TABLE`, we declare the indexes first so that the foreign keys that are also declared will reference them.

PR #944, opened 2020-10-19
go/store/types: value_store: Optimize GC to work in parallel and use less memory.

reltuk:


go/store/types/value_store.go
```diff
@@ -563,63 +563,68 @@ func (lvs *ValueStore) GC(ctx context.Context) error {
...
-	eg.Go(func() error {
-		// send root chunk
-		err := sendHash(root)
-		if err != nil {
-			return err
+	const batchSize = 16384
+	batches := func(hs []hash.Hash) [][]hash.Hash {
+		copied := make([]hash.Hash, len(hs))
+		copy(copied, hs)
+		res := make([][]hash.Hash, 0)
```
andrew-wm-arthur, line 589: why make `copied` here? 
also Nit: `var [][]hash.Hash res`

go/store/types/parallel_ref_walker.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if i < len(vals) {
+		if err := w.sendWork(parallelRefWalkerWork{
+			vals[i:len(vals)],
+			resCh,
+		}); err != nil {
+			return 0, nil, err
+		}
+		numSent++
+	}
```
andrew-wm-arthur, line 97:
```suggestion
	for i < len(vals) {
		j := i + step
		if j > len(vals) {
			j = len(vals)
		}
		if err := w.sendWork(parallelRefWalkerWork{
			vals[i:j],
			resCh,
		}); err != nil {
			return 0, nil, err
		}
		i = j
		numSent++
	}
```

andrew-wm-arthur (approved):
LGTM! a few comments

go/store/types/value_store.go
reltuk: I added a comment. It's because the caller goes on to set `toVisit = toVisit[0:0]` and then append things onto `toVisit`. So the caller mutates the parameter slice after the call to `batches`.

https://play.golang.org/p/C1Rhoje_KPo

PR #948, opened 2020-10-21
/{sysbench, run_benchmarks.sh}: some fixes on the path to connection

coffeegoddd:


oscarbatori:
Sorry for sending you down a rabbit hole. After embarking down one of my own, I can report the root cause. 

The observed behavior was that `sysbench` could connect to neither MySQL, nor Dolt. I observed similar behavior for SQL Alchemy. This implied to me that the issue was Docker networking configurations. After banging my head against a wall for many hours and flailing with various setups, this proved incorrect. 

The problems were in fact as follows:
- Dolt requires `--host=0.0.0.0` passed to `dolt sql-server` if you want to connect from another machine, it needs to be bound
- MySQL 8+ has some change in default behavior of storing passwords that requires a configuration change to allow certain (such as `mysql-client` and `sysbench`) to connect, for now I just downgraded to 5.7 which should be perfectly adequate

Thanks for your help. I incorporated some of your suggestions into the final PR, and appreciate you taking a look.

PR #949, opened 2020-10-21
Andy/auto increment

andrew-wm-arthur:


bats/auto_increment.bats
```diff
@@ -0,0 +1,164 @@
...
+    [[ "$output" =~ "1,1" ]] || falseΩ
+    [[ "$output" =~ "2.1,2" ]] || false
+    [[ "$output" =~ "3,3" ]] || false
+    [[ "$output" =~ "3.9,4" ]] || false
+    [[ "$output" =~ "5,5" ]] || false
+}
+
+@test "create auto_increment tables with all numeric types" {
+    # signed integer types and floating point
+    for TYPE in TINYINT SMALLINT MEDIUMINT INT BIGINT FLOAT DOUBLE
```
timsehn, line 116: Love the for loop

timsehn (commented):
BATS feedback.

go/libraries/doltcore/schema/encoding/schema_marshaling_test.go
```diff
@@ -194,8 +191,8 @@ func TestMirroredTypes(t *testing.T) {
 
 // testEncodedColumn is a mirror type of encodedColumn that helps ensure compatibility between Dolt versions
 //
-// Fields in this test struct should be added WITHOUT the "omitempty" annotation in order to guarantee that
-// all fields in encodeColumn are always being written when encodedColumn is serialized.
+// If a field in encodedColumn is not options, it should be added WITHOUT the "omitempty" annotation here
```
zachmu, line 194: optional?

bats/auto_increment.bats
```diff
@@ -0,0 +1,189 @@
...
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "1,1" ]] || false
+}
+
+@test "insert into auto_increment table with skipped keys" {
+    dolt sql -q "INSERT INTO test VALUES (1,1),(100,100);"
+
+    run dolt sql -q "INSERT INTO test (c0) VALUES (101);"
+    [ "$status" -eq 0 ]
+    run dolt sql -q "SELECT * FROM test WHERE c0 > 100;" -r csv
```
zachmu, line 50: Should also do an insert below the greatest existing key

bats/auto_increment.bats
```diff
@@ -0,0 +1,189 @@
...
+@test "insert into auto_increment table with NULL" {
+    dolt sql -q "INSERT INTO test VALUES (1,1);"
+
+    run dolt sql -q "INSERT INTO test (pk,c0) VALUES (NULL,2);"
+    [ "$status" -eq 0 ]
+    run dolt sql -q "SELECT * FROM test WHERE c0 > 1;" -r csv
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "2,2" ]] || false
+
+    run dolt sql -q "INSERT INTO test VALUES (NULL,3);"
```
zachmu, line 64: Add a skipped test for inserting 0 (should behave the same as NULL)

bats/auto_increment.bats
```diff
@@ -0,0 +1,189 @@
...
+@test "insert into table via batch mode" {
+    # asserts proper batch mode handling
+    run dolt sql <<SQL
+CREATE TABLE test2 (
+    pk int PRIMARY KEY,
+    c0 int
+);
+INSERT INTO test2 VALUES (1,1),(2,2),(3,3);
+INSERT INTO test2 SELECT (pk + 10), c0 FROM test2;
+INSERT INTO test2 SELECT (pk + 20), c0 FROM test2;
```
zachmu, line 99: This actually looks like a bug: the table should contain ((pk + 30), c0) rows, and it doesn't

zachmu: Or if this works, the test should assert it

zachmu: Looks like this should work based on a read of changes to the batching code

zachmu (approved):
Looks great

PR #951, opened 2020-10-23
Mergeable Indexes Pt. 1

Hydrocharged:
Very incomplete implementation of mergeable indexes. All of the surrounding logic to handle the new `LookupOp`s have been implemented, but only the `LookupOp`s `All` and `None` are implemented, with `Equals` partially implemented (enough to run some small scale unofficial tests confirming intended behavior).

For this draft PR, you can review everything outside of the new `lookupops` folder, and within that folder I'd recommend the files `all.go`, `equals.go`, and `lookupop.go`, as everything else is either unimplemented or follows the same pattern.


Hydrocharged:
Also this won't pass tests so ignore that.

go/libraries/doltcore/sqle/index_lookup.go
```diff
@@ -29,20 +31,116 @@ type IndexLookupKeyIterator interface {
...
+	}
+	return &doltIndexLookup{
+		idx:       il.idx,
+		lookupOps: lookupOps,
+	}
+}
+
+// Difference implements sql.MergeableIndexLookup
+func (il *doltIndexLookup) Difference(...sql.IndexLookup) sql.IndexLookup {
+	panic("MergeableIndexLookup Difference unused at the time of implementation")
```
zachmu, line 128: Might remove this from the interface for now

go/libraries/doltcore/sqle/lookupops/lookupop.go
```diff
@@ -0,0 +1,159 @@
...
+// key larger, such as [7,9]. They do overlap, from [7,min] to [7,8].
+//
+// Each comparison may have a small comment similar to the above, or an example of a comparison using the standard
+// comparison signs (==, <, >, etc.) suffixed by a tuple.
+
+// NOTE: The function names all contain underscores to make it visually easier to recognize each function considering
+// the length of their names.
+
+// Returns a new LookupOp representing the result set of all possible values.
+func NewLookupOp_All(nbf *types.NomsBinFormat) LookupOp {
```
zachmu, line 72: Would move these into the files where the structs are defined, and get rid of the _ nomenclature. The package name has the type, so just do NewAll, NewNone, etc.

go/libraries/doltcore/sqle/lookupops/lookupop.go
```diff
@@ -0,0 +1,159 @@
...
+// Returns a new LookupOp representing the result set (x <= key).
+func NewLookupOp_LessThanOrEqual(nbf *types.NomsBinFormat, key types.Tuple) LookupOp {
+	return &lessThanOrEqual{
+		nbf: nbf,
+		key: key,
+	}
+}
+
+// Returns a new LookupOp representing the result set (lt > x > gt).
+func NewLookupOp_GreaterThan_LessThan(nbf *types.NomsBinFormat, gt types.Tuple, lt types.Tuple) LookupOp {
```
zachmu, line 126: Would call these Between, and potentially combine a bunch them with the addition of inclusiveStart, inclusiveEnd flags

zachmu (commented):
Approach and interfaces seem sound. I have some comments on the naming and organization, but it seems like this direction should work fine.

I would work on getting a thin vertical slice of this working, since there is a lot of surface area here, esp after you write tests. I.e. start with just Equals, and combine them only with other Equals lookups. This will allow you to deliver in smaller pieces and still add functionality. (For this approach to work, you probably want LookupOp to implement IsMergeable)

go/libraries/doltcore/sqle/index_lookup.go
```diff
@@ -29,20 +31,116 @@ type IndexLookupKeyIterator interface {
...
+func (il *doltIndexLookup) IsMergeable(indexLookup sql.IndexLookup) bool {
+	otherIl, ok := indexLookup.(*doltIndexLookup)
+	if !ok {
+		return false
+	}
+	return il.idx == otherIl.idx
+}
+
+// Intersection implements sql.MergeableIndexLookup
+//TODO: change signature to accept errors
```
reltuk, line 55: Another approach is to return an allows-erroring-IndexLookup implementation...

Hydrocharged: The interface was updated altogether (breaking change) in `go-mysql-server` already to allow for errors 😄

go/libraries/doltcore/sqle/lookupops/lookupop.go
Hydrocharged: I’ll look into these! And completely forgot about `setalgebra` as well.

Hydrocharged:
Fully working implementation, but all tests are missing. Those will come in another PR. Many, many thanks to @reltuk for writing a range implementation that I was able to use as a base!

go/libraries/doltcore/sqle/lookup/range.go
```diff
@@ -0,0 +1,273 @@
...
+	// neverContinueRangeCheck will immediately end.
+	neverContinueRangeCheck noms.InRangeCheck = func(tuple types.Tuple) (bool, error) {
+		return false, nil
+	}
+)
+
+// OpenRange returns a range representing {l < x < u}.
+func OpenRange(lower, upper types.Tuple) Range {
+	return Range{
+		Above{key: lower},
```
zachmu, line 42: Request OpenLower() and ClosedLower(), OpenUpper(), etc. aliases for these constructors. Trying to sort these out as is makes my brain hurt.

go/libraries/doltcore/sqle/lookup/range.go
```diff
@@ -0,0 +1,273 @@
...
+	return Range{
+		BelowAll{},
+		AboveAll{},
+	}
+}
+
+// EmptyRange returns the empty range.
+func EmptyRange() Range {
+	return Range{
+		Above{},
```
zachmu, line 114: This will panic for Contains

go/libraries/doltcore/sqle/lookup/rangeslice.go
```diff
@@ -0,0 +1,67 @@
...
+
+// RangeSlice is a sortable slice of ranges.
+type RangeSlice []Range
+
+func (r RangeSlice) Len() int      { return len(r) }
+func (r RangeSlice) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
+func (r RangeSlice) Less(i, j int) bool {
+	lc, err := r[i].LowerBound.Compare(r[j].LowerBound)
+	if err != nil {
+		panic(err) //TODO: write own sort that accepts errors
```
zachmu, line 27: Please do before shipping. Just make RangeSlice a struct type and store the error in there

zachmu (approved):
This looks pretty good, nice implementation.

My main comment is that intersect is broken for the case where you have multiple ranges in either the receiver or caller. Needs lots of tests when you write those.

Get this checked in with CanMerge() returning false. Looking forward to the tests!

go/libraries/doltcore/sqle/index_lookup.go
```diff
@@ -29,20 +31,107 @@ type IndexLookupKeyIterator interface {
...
+	return il.idx == otherIl.idx
+}
+
+// Intersection implements sql.MergeableIndexLookup
+func (il *doltIndexLookup) Intersection(indexLookups ...sql.IndexLookup) (sql.IndexLookup, error) {
+	lookupRange := lookup.AllRange()
+	var err error
+	var ok bool
+	for _, ilRange := range il.ranges {
+		lookupRange, ok, err = lookupRange.TryIntersect(ilRange)
```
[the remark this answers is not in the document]
Hydrocharged: Ohhh fantastic catch!

go/libraries/doltcore/sqle/lookup/above.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return a.key.Equals(other.key)
+}
+
+// Format implements Cut.
+func (a Above) Format() *types.NomsBinFormat {
+	if a.key.Empty() {
+		return types.Format_Default
+	}
+	return a.Format()
```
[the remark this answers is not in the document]
Hydrocharged: Whoops meant to put `a.key.Format()`

go/libraries/doltcore/sqle/lookup/range.go
Hydrocharged: I think I can just remove `Contains` altogether

go/libraries/doltcore/sqle/lookup/range.go
Hydrocharged: I could probably just rename these altogether.

PR #957, opened 2020-10-26
go/store/{datas,util/tempfiles}: Fix some races in map writes. One effects clones, one effects only tests.

reltuk:


andrew-wm-arthur (approved):
LGTM!

PR #966, opened 2020-10-28
go/store/types/edits: Rework AsyncSortedEdits to use errgroup, and a transient goroutine for each work item.

reltuk:


go/store/types/edits/async_sorted_edits.go
```diff
@@ -15,165 +15,93 @@
...
-
-	if err := ase.ae.Get(); err != nil {
+	if err := ase.sortGroup.Wait(); err != nil {
 		return nil, err
 	}
 
-	ase.Sort()
-
-	if err := ase.ae.Get(); err != nil {
+	if err := ase.mergeCollections(); err != nil {
```
reltuk, line 100: I made these methods private, since they were dangerous to use publically.

go/store/types/edits/async_sorted_edits.go
```diff
@@ -15,165 +15,93 @@
...
-func (ase *AsyncSortedEdits) pollSortedSlices() {
-	for {
-		select {
-		case val := <-ase.resultChan:
-			coll := NewKVPCollection(ase.nbf, val)
-			ase.sortedColls = append(ase.sortedColls, coll)
-
-		default:
+		coll := NewKVPCollection(ase.nbf, ase.accumulating)
+		ase.sortedColls = append(ase.sortedColls, coll)
```
reltuk, line 67: The result slice can go directly into `sortedColls`, because the sort that takes place in the background is in place.

go/store/types/edits/async_sorted_edits.go
```diff
@@ -184,87 +112,42 @@ func (ase *AsyncSortedEdits) Close() {
...
-		close(sortChan)
-
-		for numGoRs > 0 {
-			select {
-			case val := <-resChan:
-				ase.sortedColls = append(ase.sortedColls, val)
-
-			case <-ase.doneChan:
-				numGoRs--
+		ase.sortedColls = make([]*KVPCollection, len(pairs))
```
reltuk, line 122: Once we have our pairs, we know we're going to have one result collection per pair, and they go can directly into `sortedColls`.

go/store/types/edits/async_sorted_edits.go
```diff
@@ -15,165 +15,93 @@
...
-			defer func() {
-				doneChan <- true
-			}()
-
-			err := sorter(nbf, sortChan, resChan)
-			ae.SetIfError(err)
-		}()
-	}
-
+	group, groupCtx := errgroup.WithContext(context.TODO())
```
reltuk, line 49: Should I come back through and thread context?

go/store/types/edits/async_sorted_edits.go
bheni: I think this needs a comment.  Scares me that changing the way that sort is happening in the background could break this

go/store/types/edits/async_sorted_edits.go
```diff
@@ -15,165 +15,93 @@
...
-		case val := <-ase.resultChan:
-			coll := NewKVPCollection(ase.nbf, val)
-			ase.sortedColls = append(ase.sortedColls, coll)
-
-		default:
+		coll := NewKVPCollection(ase.nbf, ase.accumulating)
+		ase.sortedColls = append(ase.sortedColls, coll)
+		toSort := ase.accumulating
+		ase.accumulating = make([]types.KVP, 0, ase.sliceSize)
+		if err := ase.sema.Acquire(ase.sortCtx, 1); err != nil {
```
bheni, line 70: I could see this being a perf hit for large imports. It'll block the main thread and slow it from getting the data in.

bheni (commented):
I don't love the sema on AddEdit.  Don't want to block the main thread waiting for a worker.  Would do a go routine per batch here

bheni (approved):
Like it

PR #971, opened 2020-10-29
Added GitHub workflow tests for race conditions

Hydrocharged:
Will fail until https://github.com/dolthub/dolt/pull/967 is merged into `master`, however the workflow only works when the PR is based against `master`. Therefore this PR does not target the aforementioned PR's branch.


reltuk:
One question: if we're running `go test ./... -race` on all platforms, do we also need to run `go test ./...` on all platforms?

coffeegoddd:
IMO if `go test ./... -race` isn't doing something totally different than `go test ./...` then i think the ci-gotest file can be modified to just use the `-race` flag. Or if this needs to run separately from `go test ./...`, we can just add a subsequent step to that file like:
```yaml
- name: Test For Races
      working-directory: ./go
      run: go test -race ./...
```

Hydrocharged:
From what I can tell, `-race` will catch everything that the lack of `-race` would catch, so I can just modify the current ones instead.

Hydrocharged:
@reltuk @coffeegoddd Look good now?

coffeegoddd (approved):
LGTM! made this PR against your branch https://github.com/dolthub/dolt/pull/976 , seems not great we fail go tests on a workflow change.

PR #975, opened 2020-10-30
Extract some import logic to be used in dolthubapi

tbantle22:
In reference to this comment https://github.com/dolthub/ld/pull/5262#discussion_r514465176

I had some duplicate logic in `dolthubapi` for the import table api. I extracted some logic so that I can use `InferSchema` and `MoveDataToRoot` to root to reduce some of the duplications


reltuk (approved):
LGTM!

PR #980, opened 2020-10-31
In prettyPrint, defer closing the iterator before doing anything else…

zachmu:
…. We were missing close() when an UPDATE or INSERT etc. had an error during cursor iteration, therefore leaving a server process running. Also save sql history file before executing the query, so it gets saved even if the user interrupts execution.

Signed-off-by: Zach Musgrave <zach@dolthub.com>


bheni (approved):
uit's beautiful

PR #982, opened 2020-11-01
go/libraries/utils/async: Have ActionExecutor use sync.WaitGroup.

reltuk:


reltuk:
@Hydrocharged Curious for your thoughts here. The wait group tests were pretty heavy under `-race` and I saw (transient?) failures under PRs. The semantics of the thing with concurrent `Add` and `Wait` are non-deterministic and I was thinking probably not more useful than the behavior exhibited by this small change.

Hydrocharged (approved):
The idea was to be able to call `WaitForEmpty` and then let it process through an entire workload knowing that it would wait for the entire load to finish before the wait ended, but if there was ever a moment when the counter dropped to zero during that load then `WaitForEmpty` would have returned anyway, somewhat defeating that purpose.

The proper implementation of that specific case would have been to use a close function, which I was wanting to avoid (no need to close an `ActionExecutor` so it never leaks). So my reasoning was flawed, and this essentially accomplishes the same from a theoretical standpoint. Plus if we were getting failures beforehand then it's doubly worth replacing.

Looks good to me!

PR #990, opened 2020-11-04
mysql auto increment semantics

andrew-wm-arthur:


go/libraries/doltcore/sqle/tables.go
```diff
@@ -65,10 +65,31 @@ func init() {
...
 type DoltTable struct {
 	name   string
-	table  *doltdb.Table
-	sch    schema.Schema
 	sqlSch sql.Schema
 	db     SqlDatabase
+
+	table   *doltdb.Table
+	sch     schema.Schema
+	autoCol schema.Column
```
zachmu, line 73: autoIncCol?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -228,6 +258,7 @@ var _ sql.UpdatableTable = (*WritableDoltTable)(nil)
 var _ sql.DeletableTable = (*WritableDoltTable)(nil)
 var _ sql.InsertableTable = (*WritableDoltTable)(nil)
 var _ sql.ReplaceableTable = (*WritableDoltTable)(nil)
+var _ sql.AutoIncrementTable = (*WritableDoltTable)(nil)
```
zachmu, line 261: Ideally you would only implement this conditionally for tables that have autoincrement, but you can't really do that.

The other alternative is that you use some standard error type provided by go-mysql-server to return an error in the case that a table doesn't have any auto_increment columns.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -297,6 +328,14 @@ func (t *WritableDoltTable) Updater(ctx *sql.Context) sql.RowUpdater {
 	return te
 }
 
+// GetAutoIncrementValue gets the last AUTO_INCREMENT value
+func (t *WritableDoltTable) GetAutoIncrementValue(ctx *sql.Context) (interface{}, error) {
```
zachmu, line 332: See my comment above. How is go-mysql-server expected to respond to a table that declares that it supports auto increment values via the interface method, but doesn't declare any auto_inc columns? Needs to be part of the contract. You could make it explicit by adding a `HasAutoIncrementColumn` method to that interface, or you could sort it out with good error handling. But what you have now is kind of confusing.

(also I don't think you should need to implement this method on WritableDoltTable, since it embeds DoltTable)

zachmu (approved):
Looks generally good, but the interface you chose to bridge the gap with go-mysql-server leaves room for confusion: dolt tables with no auto_increment columns will still implement the sql.AutoIncrementTable interface. You need to be really explicit about what that means. Several ways to resolve it, but the semantics needs to be obvious and documented.

PR #997, opened 2020-11-06
Fix typo in brew bump

oscarbatori:
From failed run on most recent release:
Screen Shot 2020-11-05 at 6 29 02 PM


oscarbatori:
Merging this as it only touches our GitHub Actions `brew bump` job which is currently broken.

PR #1001, opened 2020-11-06
NewCreatingWriter breaks dolthubapi with recent changes

tbantle22:
There might be a better fix for this, but `dolthubapi` uses `NewCreatingWriter` which breaks with Andy's [recent changes](https://github.com/dolthub/dolt/commit/b6be196af9b31fd1d37cf423e7777202cf8e7843#diff-f6b6f55f2e449395d3b8e0d758e8f3e1de084bdd6369d9e8e8fe1d69c4878920R100) (it's being used in dolthubapi [here](https://github.com/dolthub/ld/blob/master/go/services/dolthubapi/pkg/domain/repositorydata.go#L1922))

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/export.go
+++ b/go/cmd/dolt/commands/tblcmds/export.go
@@ -262,7 +262,7 @@ func NewExportDataMover(ctx context.Context, root *doltdb.RootValue, dEnv *env.D
 	inSch := rd.GetSchema()
 	outSch := inSch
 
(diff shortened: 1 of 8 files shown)


andrew-wm-arthur (changes requested):
Sorry for stepping on your toes here. 
Instead of using `te.dEnv == nil` as the flag for gc, there should be an explicit boolean option in `DataMoverOptions`.

PR #1008, opened 2020-11-10
Mergeable Indexes Pt. 2

Hydrocharged:
Tests for mergeable indexes


go/libraries/doltcore/sqle/mergeable_indexes_test.go
```diff
@@ -0,0 +1,938 @@
...
+			[]lookup.Range{
+				lookup.AllRange(),
+			},
+			[]int64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9},
+		},
+		{
+			"v1 > 11 AND v1 <= 15",
+			2,
+			[]lookup.Range{
+				lookup.CustomRange(idxv1.tuple(true, 11), idxv1.tuple(true, 15),
```
zachmu, line 449: This abstraction is bugging me. The CustomRange method should figure out how to deliver a ranged iterator that delivers on its promise. The fact that the caller has to construct tuples that sometimes includes a max ID to get correct behavior is a sign that this abstraction is leaking in a gross way. Think about how you can fix that.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
```diff
@@ -0,0 +1,938 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/lookup"
+)
+
+// This tests mergeable indexes by using the SQL engine and intercepting specific calls. This way, we can verify that
+// the engine is intersecting and combining the proper number of lookups, and we can also examine the ranges before
+// they're converted into a format that Noms understands to verify that they were handled correctly. Lastly, we ensure
+// that the final output is as expected.
+func TestMergeableIndexes(t *testing.T) {
```
zachmu, line 33: These tests are great, but you're missing an important class of them. I want to see things like

v1 > 1 AND v1 < 5 OR v1 > 6 AND v1 < 9

And variations.

Also things like:

(v1 > 1 AND v1 < 5 OR v1 > 6 AND v1 < 9) AND v1 > 3

Things where you're combining complex sets of predicates together with and AND or OR.

zachmu: Conspicuously missing: testing use of index for IN clauses (v1 in (1,2,3)). If that doesn't work in the engine yet, then skip the test, make it work, and then unskip it

zachmu (approved):
This looks generally quite good, very thorough job with the tests as usual.

My main comment about how this turned out is that your earlier mechanism of artificially constructing the range tuples with the MaxInt64 value tag really conflicts with your new range / cut abstractions in a gross way. Something needs to give there. Think about how to resolve this tension. Doesn't have to be part of this PR, but should be addressed soon.

go/libraries/doltcore/sqle/mergeable_indexes_setup_test.go
```diff
@@ -0,0 +1,320 @@
...
+	t            *testing.T
+	countLookups func(int)
+	finalRanges  func([]lookup.Range)
+}
+
+func (il *testMergeableIndexLookup) IsMergeable(indexLookup sql.IndexLookup) bool {
+	return il.doltIndexLookup.IsMergeable(indexLookup.(*testMergeableIndexLookup).doltIndexLookup)
+}
+func (il *testMergeableIndexLookup) Intersection(indexLookups ...sql.IndexLookup) (sql.IndexLookup, error) {
+	il.countLookups(len(indexLookups) + 1) // include 1 for the caller
```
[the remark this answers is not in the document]
Hydrocharged: `countLookups` essentially lets us know that we've unioned and intersected the expected number of lookups. For instance `v1 = 1` doesn't require either of the two, so the `lookupCount` ends up being 0, while `v1 = 1 OR v1 = 2` ends up being 2 since that's two separate lookups.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
```diff
@@ -0,0 +1,938 @@
...
+	engine, db, idxv1, idxv2v1 := setupMergeableIndexes(t)
+
+	tests := []struct {
+		whereStmt      string
+		lookupsCounted int
```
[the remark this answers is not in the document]
Hydrocharged: They're more to make sure that we're actually attempting to union and intersect all of the applicable indexes. It is fairly specific to the current implementation details, as the interface allows for merging multiple indexes at once but it currently only does two at a time. So perhaps it's better to just make sure that we're merging at all? It's easily changeable.

Initially I was going to make a `testRange` that had a unique prime number for that test, and I'd multiply all merged numbers together. So for `v1 >= 1 OR v1 = 1 `, I'd give `v1 >= 1` the number 2 and `v1 = 1` would get 3, and I could check that the final range has the number 6. The issue is that there is no guarantee that the engine will acquire the lookups in a deterministic order, so I'd have to add an additional field to ensure that the correct numbers go where I expect them to. I can do it, I just decided to do the quicker/easier option.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
Hydrocharged: I can just move the max ID portion to the Range functions. I didn't do it as I felt it may break on some tuples, but that's only true if the tuple given has a dangling tag with no value. I'd say that that should never happen, and such a tuple is an error/bug so it's fine if it causes more issues further down the stack.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
Hydrocharged: So right now I have (replacing X for =, >, <, etc.):

X1
X1 OR X2
X1 AND X2
X1 OR X2 OR X3
X1 OR X2 AND X3
X1 AND X2 OR X3
X1 AND X2 AND X3

So you're saying I should add:

X1 OR X2 OR X3 OR X4
X1 OR X2 OR X3 AND X4
X1 OR X2 AND X3 OR X4
X1 OR X2 AND X3 AND X4
X1 AND X2 OR X3 OR X4
X1 AND X2 OR X3 AND X4
X1 AND X2 AND X3 OR X4
X1 AND X2 AND X3 AND X4
X1 OR X2 OR X3 OR X4 OR X5
X1 OR X2 OR X3 OR X4 AND X5
X1 OR X2 OR X3 AND X4 OR X5
X1 OR X2 OR X3 AND X4 AND X5
X1 OR X2 AND X3 OR X4 OR X5
X1 OR X2 AND X3 OR X4 AND X5
X1 OR X2 AND X3 AND X4 OR X5
X1 OR X2 AND X3 AND X4 AND X5
X1 AND X2 OR X3 OR X4 OR X5
X1 AND X2 OR X3 OR X4 AND X5
X1 AND X2 OR X3 AND X4 OR X5
X1 AND X2 OR X3 AND X4 AND X5
X1 AND X2 AND X3 OR X4 OR X5
X1 AND X2 AND X3 OR X4 AND X5
X1 AND X2 AND X3 AND X4 OR X5
X1 AND X2 AND X3 AND X4 AND X5

Or I should just add a few of them to verify that they're working as expected (not needing such an exhaustive amount)?

Also, I didn't even think about `IN` clauses, I'll get those added.

zachmu: What I was expecting to see (and didn't) was a where clause that resulted in two closed ranges on the lookup. That's the main thing I want to see covered.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
zachmu: Your call, but I don't want the callers of these functions to have to add MAX_TAG artifacts into their tuples. Probably multiple ways to achieve that.

go/libraries/doltcore/sqle/mergeable_indexes_test.go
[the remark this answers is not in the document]
zachmu: I actually just don't care about the intermediate merge operations that produce the result, as long as the final result is correct. Adding that kind of extra info to the test is just a way to make up for not having enough test cases in the first place (which I'm pretty sure you do). You shouldn't need to verify the internal operations of these objects to verify the algorithm works correctly.

go/libraries/doltcore/sqle/mergeable_indexes_setup_test.go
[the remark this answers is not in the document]
zachmu: See my comment below. I don't think these things are pulling their weight, and they make the tests more complex than they need to be (and potentially more prone to breakage if we change the algorithm).

go/libraries/doltcore/sqle/mergeable_indexes_test.go
Hydrocharged: Ahh gotcha. I can add a few that results in those.

PR #1011, opened 2020-11-10
Bh/commit diff

bheni:


go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+		return nil, err
+	}
+
+	fromRoot, fromName, fromDate, err := dt.rootValForFilter(ctx, dt.fromCommitFilter)
+
+	if err != nil {
+		return nil, err
+	}
+
+	toTable, _, err := toRoot.GetTable(ctx, dt.name)
```
zachmu, line 253: Probably want to handle `ok` on these, this will panic if the lookup fails otherwise

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+		toDate:   toDate,
+		fromDate: fromDate,
+	}}), nil
+}
+
+func (dt *CommitDiffTable) rootValForFilter(ctx *sql.Context, eqFilter *expression.Equals) (*doltdb.RootValue, string, *types.Timestamp, error) {
+	var hashStr string
+	var ok bool
+	if _, isGetField := eqFilter.BinaryExpression.Left.(*expression.GetField); !isGetField {
+		val, err := eqFilter.BinaryExpression.Left.Eval(ctx, nil)
```
zachmu, line 279: This wont' work for all expressions (like a subselect, for example). Should do something reasonable in that case.

zachmu: The row can't be nil I mean

zachmu: You probably need to do an inspection at set time and only allow certain types of expressions

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+
+		if err != nil {
+			return nil, "", nil, err
+		}
+
+		if hashStr, ok = val.(string); !ok {
+			return nil, "", nil, fmt.Errorf("received '%v' when expecting commit hash string", val)
+		}
+	} else if _, isGetField = eqFilter.BinaryExpression.Right.(*expression.GetField); !isGetField {
```
zachmu, line 288: This logic is kind of convoluted. Probably you want something like:

```
left, right := eqFilter.Left, eqFilter.Right
if _, ok := left.(GetField) {
   left, right = right, left
}
```

Then you just run your logic all on left

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+
+		t := meta.Time()
+		commitTime = (*types.Timestamp)(&t)
+	}
+
+	return root, hashStr, commitTime, nil
+}
+
+// HandledFilters returns the list of filters that will be handled by the table itself
+func (dt *CommitDiffTable) HandledFilters(filters []sql.Expression) []sql.Expression {
```
zachmu, line 337: This really needs to be able to return an error, so that we can reject queries that have incompatible WHERE clauses. Might add a TODO.

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+				return true
+			})
+		}
+
+		if !isCommitFilter {
+			remaining = append(remaining, filter)
+		}
+	}
+
+	return remaining
```
zachmu, line 377: Isn't this the opposite logic you want? I thought this interface returned the set of expressions handled. Am I way off base?

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+	if dt.toCommitFilter == nil || dt.fromCommitFilter == nil {
+		return nil
+	}
+
+	return []sql.Expression{dt.toCommitFilter, dt.fromCommitFilter}
+}
+
+// WithFilters returns a new sql.Table instance with the filters applied
+func (dt *CommitDiffTable) WithFilters(filters []sql.Expression) sql.Table {
+	return dt
```
zachmu, line 391: Seems wrong

go/libraries/doltcore/sqle/commit_diff_table.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// HandledFilters returns the list of filters that will be handled by the table itself
+func (dt *CommitDiffTable) HandledFilters(filters []sql.Expression) []sql.Expression {
+	var remaining []sql.Expression
+	for _, filter := range filters {
+		isCommitFilter := false
+
+		if eqFilter, isEquality := filter.(*expression.Equals); isEquality {
+			sql.Inspect(filter, func(e sql.Expression) (cont bool) {
```
zachmu, line 343: This isn't quite right: you can't use sql.Inspect here because it walks the entire expression tree, and you really just want the top level. You need to inspect the left and right of the equals expression directly. One of them must be a GetField to be handled.

zachmu: You can use the same `left, right = right, left` swapping trick here as well

zachmu: The way you have it now will catch expressions like `concat(from_commit, "abc")`, which will yield incorrect results with your logic below

zachmu (commented):
LG but has a few correctness issues you should address

bats/system-tables.bats
```diff
@@ -33,10 +33,11 @@ teardown() {
     [ $status -eq 0 ]
     [[ "$output" =~ "dolt_history_test" ]] || false
     [[ "$output" =~ "dolt_diff_test" ]] || false
+    [[ "$output" =~ "dolt_commit_diff_test" ]] || false
     run dolt ls --all
     [ $status -eq 0 ]
     [[ "$output" =~ "dolt_history_test" ]] || false
-    [[ "$output" =~ "dolt_diff_test" ]] || false
+    [[ "$output" =~ "dolt_commit_diff_test" ]] || false
```
zachmu, line 40: Shouldn't these be additions rather than changes?

zachmu (approved):

go/libraries/doltcore/sqle/commit_diff_table.go
bheni: it doesn't panic.  It handles nils as a table not existing and all rows being added.  I updated my bats test to verify.

go/libraries/doltcore/sqle/commit_diff_table.go
bheni: nice

go/libraries/doltcore/sqle/commit_diff_table.go
bheni: updated tests and fixed

go/libraries/doltcore/sqle/commit_diff_table.go
bheni: updated tests and fixed

go/libraries/doltcore/sqle/commit_diff_table.go
bheni: So the interface here is kinda garbage.  HandledFilters passes a list of the filters and asks which ones you are going to handle in the table.  Then WithFilters is called with the same set of filters and wants a Table returned which handles those filters.  Anyway, In this class HandledFilters mutates the object, so it is just returned by WithFilters.  Not sure where this breaks down, but I haven't seen issue with this yet.

PR #1012, opened 2020-11-10
added reset_hard() sql function

andrew-wm-arthur:


timsehn:
Should we prefix all of these functions with `dolt_`?

andrew-wm-arthur:
That would help to indicate it's a Dolt-specific function, at the cost of being more verbose.
We don't currently have user-defined functions, but we'll have to think about name-space concerns once we do

zachmu:
Any function that does the same thing as a dolt command should just be 1:1, so this should be `dolt_reset(hard string)`

andrew-wm-arthur:
@zachmu the CLI and SQL interface already diverge in this case. We have no concept of staging in SQL so `dolt_reset()` wouldn't mean anything

go/libraries/doltcore/sqle/dfunctions/reset.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/dolthub/go-mysql-server/sql"
+)
+
+const ResetHardFuncName = "reset_hard"
+
+type ResetHardFunc struct{}
+
+// NewResetHardFunc creates a new ResetHardFunc expression.
+func NewResetHardFunc() sql.Expression {
```
zachmu, line 28: See my top level comment, but I think for these it's important to exactly mirror the semantics of the command line. So this should be `dolt_reset` and take N args: a boolean `hard` param followed by N table names.

zachmu: (hard == false would be a no-op)

zachmu:
We can have additional aliases for the same function, that's fine. So `dolt_reset_hard()` can be an alias for `dolt_reset(hard)`. But we do want to have the latter even if soft is not yet supported.

I left a note about making this take a list of tables to reset. I think it's fine to ship what you have now (`dolt reset --hard .`) but we'll want to eventually make this match the command line.

zachmu (approved):

PR #1017, opened 2020-11-11
/go/libraries/doltcore/sqle:  Created system table package

andrew-wm-arthur:


go/libraries/doltcore/sqle/database.go
```diff
@@ -203,46 +204,53 @@ func (db Database) GetTableInsensitive(ctx *sql.Context, tblName string) (sql.Ta
...
-			tblName = tblName[len(prefix):]
-			dt, err := newFunc(ctx, db, tblName)
-
-			if err != nil {
-				return nil, false, err
-			}
-
-			return dt, true, nil
-		}
+	switch {
```
zachmu, line 215: Maybe extract a method getSystemTable?

zachmu (approved):
Generally looks good, nice cleanup.

The one concern I have is that you've changed the caching semantics of these tables. They used to get a root value at run time from the context, and now they get one at construction. That's fine, but it means we absolutely cannot reuse these tables between queries (where we could before). Make a note of that on these types.

andrew-wm-arthur:
Yeah, that's a good note.

PR #1021, opened 2020-11-12
!= operator now uses indexes

Hydrocharged:


zachmu (approved):
Looks great, excellent tests

PR #1024, opened 2020-11-13
dolt filter-branch

andrew-wm-arthur:


go/cmd/dolt/commands/filter-branch.go
```diff
@@ -0,0 +1,241 @@
...
+		ok := se.checkThenDeleteAllRows(sqlCtx, s)
+		if !ok {
+			_, itr, err = se.query(sqlCtx, query)
+		}
+	default:
+		// todo: support insert, update, ddl
+		return nil, fmt.Errorf("SQL statement not supported for filter-branch: '%v'.", query)
+	}
+
+	err, ok := captureTblNotFoundErr(err, mt, rh)
```
reltuk, line 155: Ideas for interfaces for this in the future? We can allow the query to parameterize based on the commit we're operating on...we can try to distinguish between logical and internal|i/o errors and have a mode where we continue across any logical errors...we can allow the user to give a class of logical errors which should be converted to warnings...not so easy...

go/libraries/doltcore/rebase/rebase.go
```diff
@@ -18,22 +18,169 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	_, err = dEnv.PutDocsToStaged(ctx, dd)
+	return err
+}
+
+// MigrateUniqueTags rebases the history of the repo to uniquify tags within branch histories.
+func CurrentBranch(ctx context.Context, dEnv *env.DoltEnv, replay ReplayCommitFn, nerf NeedsRebaseFn) error {
```
reltuk, line 125: Can this and `AllBranches` share a bunch of logic and just have the heads and branch refs to update be parameters?

go/libraries/doltcore/rebase/rebase.go
```diff
@@ -18,22 +18,169 @@ import (
...
+	err = dEnv.PutDocsToWorking(ctx, dd)
+	if err != nil {
+		return err
+	}
+
+	_, err = dEnv.PutDocsToStaged(ctx, dd)
+	return err
+}
+
+// MigrateUniqueTags rebases the history of the repo to uniquify tags within branch histories.
```
reltuk, line 124: Stale...

go/libraries/doltcore/rebase/rebase.go
```diff
@@ -18,22 +18,169 @@ import (
...
+
+		heads[i], err = dEnv.DoltDB.Resolve(ctx, cs, nil)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	return
+}
+
+// MigrateUniqueTags rebases the history of the repo to uniquify tags within branch histories.
```
reltuk, line 62: Stale...

reltuk (commented):
Looks great! Some small comments :)

go/cmd/dolt/commands/filter-branch.go
andrew-wm-arthur: I think we should handle the class of logical errors stemming from schema changes and error on everything else. 
Specifically, we should rewrite queries to handle column add/drops and table/column renames so that we always execute something useful when appropriate. 
This version doesn't have an ancestor commit option to stop at, so I thought we should swallow missing table errors. I think that's probably the right thing to do in general.

go/cmd/dolt/commands/filter-branch.go
```diff
@@ -0,0 +1,241 @@
...
+	roots, err := se.getRoots(sqlCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	return roots[dbName], nil
+}
+
+// monoSqlEngine packages up the context necessary to run sql queries against single root.
+func monoSqlEngine(ctx context.Context, dEnv *env.DoltEnv, root *doltdb.RootValue) (*sql.Context, *sqlEngine, error) {
```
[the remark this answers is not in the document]
andrew-wm-arthur: It's definitely time to cleanup our usage of the sql package. There are currently some import-cycle blockers. (The first thing that comes to mind is `dfunctions` depends on `sqle`

PR #1029, opened 2020-11-16
Added dolt_version()

Hydrocharged:
As `version()` is used to emulate the target MySQL version, I've added `dolt_version()` so that one may specifically query the dolt version.


andrew-wm-arthur (approved):
LGTM! 
I got a BATS for it:
```
@test "dolt_version() func" {
    SQL=$(dolt sql -q 'select dolt_version() from dual;' -r csv | tail -n 1)
    CLI=$(dolt version | cut -f 3 -d ' ')
    [ "$SQL" == "$CLI" ]
}
```

PR #1033, opened 2020-11-17
Temporary parallelism implementation on indexes

Hydrocharged:
A better implementation would make use of partitions of some form. My current idea is to split up the ranges so that each partition handles a subsection of a range, however I'd likely rewrite a portion of the ranges to facilitate this, along with some additions to the Value types (such as a new interface). In the meantime, this is a rather quick and dirty implementation of parallelizing reads on an index so that queries don't come to a standstill when adding an index.

I initially had a faster and less ugly implementation, but it completely changed the order of the output (still correct since order is arbitrary without an `ORDER BY` clause). Instead of rewriting every test, I changed it so that the old order is preserved by writing into a buffer. It's still far, far faster than the old single-threaded implementation. The following issue now has indexes performing faster than the full table scan on my machine: https://github.com/dolthub/dolt/issues/1005. Still not as fast as it could be, but it's no longer many times slower. Of note, running with `GOMAXPROCS=1` takes roughly the same time as before.

I didn't add any tests since this is a fairly deep change, and is touched by every single test above it that uses indexes.

### DIFF
--- a/go/libraries/doltcore/sqle/index_lookup.go
+++ b/go/libraries/doltcore/sqle/index_lookup.go
@@ -134,13 +134,9 @@ func (il *doltIndexLookup) RowIter(ctx *sql.Context) (sql.RowIter, error) {
 	for i, lookupRange := range il.ranges {
 		readRanges[i] = lookupRange.ToReadRange()
 	}
-	return &indexLookupRowIterAdapter{
-		idx: il.idx,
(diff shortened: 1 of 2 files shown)


zachmu (approved):
This seems fine, although I'm looking forward to splitting ranges and executing them in parallel instead.

PR #1036, opened 2020-11-17
go/cmd/dolt/commands: added write queries and ancestor commit to filter-branch

andrew-wm-arthur:


bats/filter-branch.bats
```diff
@@ -83,16 +86,141 @@ teardown() {
...
+        [[ ! "$output" =~ "to_drop" ]] || false
+    done
+}
+
+@test "filter-branch error on conflict" {
+    setup_write_test
+
+    run dolt filter-branch "INSERT INTO test VALUES (1,2);"
+    [ "$status" -ne 0 ]
+    [[ ! "$output" =~ "panic" ]] || false
```
reltuk, line 216: Can we add another error case test for something like, query doesn't successfully execute against the schema that prevails at one of the rewritten commits?

go/libraries/doltcore/rebase/rebase.go
```diff
@@ -27,13 +27,70 @@ import (
...
+		if h.Equal(sh) {
+			return false, nil
+		}
+
+		n, err := cm.NumParents()
+		if err != nil {
+			return false, err
+		}
+		if n == 0 {
+			return false, fmt.Errorf("commit %s not found in history", sh)
```
reltuk, line 63: Not necessarily not found, but was at least not found on one path from HEAD to some parent-less commit. Not exactly sure how to best word this...

reltuk (approved):
LGTM! One nit comment.

PR #1038, opened 2020-11-17
go/libraries/doltcore/remotestorage: Simplify concurrentExec implementation with errgroup.

reltuk:


go/libraries/doltcore/remotestorage/concurrency.go
```diff
@@ -26,76 +27,43 @@ func concurrentExec(work []func() error, concurrency int) error {
...
-						return
+					if err := w(); err != nil {
+						return err
 					}
-
-				// stop message received while waiting for work
-				case <-stopChan:
-					return
+				case <-ctx.Done():
+					return ctx.Err()
```
bheni, line 60: I think in my blog you said we should return eg.Wait() here.  No?

bheni (approved):
lgtm.  One comment.

PR #1039, opened 2020-11-17
go/libraries/doltcore/remotestorage: Refactor urlAndRanges a bit.

reltuk:


go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -202,44 +203,123 @@ const (
...
+					gr := (*GetRange)(loc)
+					if v, ok := res[gr.ResourcePath()]; ok {
+						v.Append(gr)
 					} else {
-						resourceToUrlAndRanges[resourcePath] = urlAndRanges{urlStr, typedLoc.HttpGetRange.Ranges}
+						res[gr.ResourcePath()] = gr
 					}
 				}
+			case <-ctx.Done():
+				return ctx.Err()
```
bheni, line 319: I would return nil here.  This go routine did not error.  This won't affect anything, I just feel like its more correct.

go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -202,44 +203,123 @@ const (
...
+	return gr.ChunkEndOffset(gr.NumChunks()-1) - gr.ChunkStartOffset(0)
+}
+
+func (gr *GetRange) ChunkByteRange(i int) (uint64, uint64) {
+	start := gr.ChunkStartOffset(i) - gr.ChunkStartOffset(0)
+	end := gr.ChunkEndOffset(i) - gr.ChunkStartOffset(0)
+	return start, end
+}
+
+func (gr *GetRange) WorkFunc(ctx context.Context, fetcher HTTPFetcher, chunkChan chan nbs.CompressedChunk) func() error {
```
bheni, line 268: Think I'd rename this to GetWorkFunc, or probably better GetDownloadFunc.

bheni (approved):
Few comments.  looks good overall

go/libraries/doltcore/remotestorage/chunk_store.go
reltuk: In these cases, I think it's best to still return `ctx.Err()`, because this could be the first goroutine that is encountering the canceled context. The cancelation could have been of a parent context, as opposed to it being of the `errgroup`'s context after the `errgroup` had already observed a non-nil error from one of its dispatched functions. This function was unable to fulfill its contract, which included reading `dlLocChan` to completion, so it returns an error about what went wrong...

bheni: sounds good.

Issue #1023, opened 2020-11-12
Remove tag info from EXTRA in dolt SQL schema


zachmu:
This breaks compatibility with standard mysql's info schema tables, and conflicts with AUTO_INCREMENT columns.

See https://github.com/dolthub/dolt/pull/1022

Main barrier to just ripping it out is current implementation of `dolt schema tags`

Hydrocharged:
How does this break compatibility with standard MySQL's infoschema tables? From reading [the docs](https://dev.mysql.com/doc/refman/8.0/en/information-schema-columns-table.html) it seems like `EXTRA` is meant to hold additional information that doesn't belong in the other columns, which would include our tags. There can be multiple values in `EXTRA`.

zachmu:
Makes it so we can't write information_schema tests that run on both Dolt and other implementation of go-mysql-server. Also the possibility that a client will read the EXTRA column and not know what to do with it.

Hydrocharged:
We could have our tests check for `contains` rather than `equals`, but we can change it if we're worried about clients getting confused seeings an unknown `tags`.

VinaiRachakonda:
So if we want to remove tags from the extra column does it make sense to add it to a different column in the information_schema?

Hydrocharged:
From previous conversations that we've had, I think we were wanting to get rid of it from SQL world altogether, and just have `dolt schema tags` for when one wants tags.

As a general rule of thumb though, for the built in tables like those under `information_schema`, their schema should mimic what MySQL has. Adding additional columns may cause clients to break as they may expect a specific number of columns.

VinaiRachakonda:
Should I go ahead and close this?

VinaiRachakonda:
Closed!

PR #1041, opened 2020-11-18
Vinai/1023 remove tag info

VinaiRachakonda:
Fixes #1022 and #1023

### DIFF
--- a/go/cmd/dolt/commands/schcmds/tags.go
+++ b/go/cmd/dolt/commands/schcmds/tags.go
@@ -141,7 +141,7 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
 			return commands.HandleVErrAndExitCode(errhand.VerboseErrorFromError(verr), usage)
 		}
 
-		err = commands.PrettyPrintResults(ctx, resultFormat , headerSchema, sql.RowsToRowIter(rows...))
+		err = commands.PrettyPrintResults(ctx, resultFormat, headerSchema, sql.RowsToRowIter(rows...))
 	} else {
 		// Default to tabular.
 		err = commands.PrettyPrintResults(ctx, 0, headerSchema, sql.RowsToRowIter(rows...))
--- a/go/cmd/dolt/commands/schcmds/tags.go
+++ b/go/cmd/dolt/commands/schcmds/tags.go
@@ -16,8 +16,6 @@ package schcmds
 
 import (
 	"context"
-	"strings"
-
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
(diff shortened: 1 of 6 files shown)


timsehn (commented):
Is it worth writing a couple bats tests to cover information schema @zachmu ?

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +89,58 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		{Name: "tag", Type: sql.Uint64, Default: nil},
 	}
 
-	//TODO: implement REGEXP_SUBSTR in go-mysql-server and use it here instead of SUBSTR, as this will eventually break
-	queryStr := fmt.Sprintf("SELECT table_name AS 'table', column_name AS 'column', "+
-		"SUBSTR(extra, 5) AS tag FROM information_schema.columns WHERE table_name IN (%s)", strings.Join(tables, ","))
+	rows := make([]sql.Row, 0)
+
+	for _, tableName := range tables {
+		table, ok, err := root.GetTable(ctx, strings.ToLower(tableName))
```
zachmu, line 102: This won't work in all cases. root.GetTable is case sensitive. So it's not enough to just lower-case the input like this.

You either need to get all the tables from the root and then look for case-insensitive matches with the tables given on the command line, or you need to implement a case-insensitive version of root.GetTable. The former is probably a smaller change, but think about what you would prefer.

Should be a unit / bats test for this case insensitive matching.

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +89,58 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+
+		if err != nil {
+			return 1
+		}
+
+		if ok {
```
zachmu, line 108: Should print an error and exit(1) if a given table cannot be found. (this logic will need to change when you make the above case sensitivity fix)

zachmu: I know the current semantics are to ignore any non-existent tables, but I think that's a mistake. It should fail loudly.

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +89,58 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+			}
 
-	if formatStr, ok := apr.GetValue(commands.FormatFlag); ok {
-		return commands.SqlCmd{}.Exec(ctx, "", []string{
-			fmt.Sprintf(`--%s=%s`, commands.FormatFlag, formatStr),
-			fmt.Sprintf(`--%s`, commands.QueryFlag),
-			queryStr + ";",
-		}, dEnv)
+			_ = sch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
+				currRow := sql.NewRow(strings.ToLower(tableName), col.Name, tag)
```
zachmu, line 116: Suggest inlining this var

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +89,58 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+			return commands.HandleVErrAndExitCode(errhand.VerboseErrorFromError(verr), usage)
+		}
+
+		err = commands.PrettyPrintResults(ctx, resultFormat, headerSchema, sql.RowsToRowIter(rows...))
 	} else {
-		return commands.SqlCmd{}.Exec(ctx, "", []string{
-			fmt.Sprintf(`--%s`, commands.QueryFlag),
-			queryStr + ";",
-		}, dEnv)
+		// Default to tabular.
```
zachmu, line 137: Instead of if / else here, you should declare an outputFmt var with default of tabular, then override it if the format flag was passed.

Basically reduce the scope of the conditional logic to just the one var that changes, rather than having a call to PrettyPrint with different params on two different conditional branches.

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +89,58 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		err = commands.PrettyPrintResults(ctx, 0, headerSchema, sql.RowsToRowIter(rows...))
 	}
+
+	if err != nil {
+		return 1
```
zachmu, line 142: Need to print an error message as well. This is what commands.HandleVerrAndExitCode does.

go/cmd/dolt/commands/sql.go
```diff
@@ -487,7 +487,7 @@ func formatQueryError(message string, err error) errhand.VerboseError {
 	}
 }
 
-func getFormat(format string) (resultFormat, errhand.VerboseError) {
+func GetFormat(format string) (resultFormat, errhand.VerboseError) {
```
zachmu, line 490: Now that this is exported, call it GetResultFormat

go/libraries/doltcore/table/untyped/tabular/tablewriter.go
```diff
@@ -174,7 +174,6 @@ func (ttw *TextTableWriter) WriteRow(ctx context.Context, r row.Row) error {
 	}
 
 	allCols := ttw.sch.GetAllCols()
-
```
zachmu, line 177: Revert this file since you didn't make any non-whitespace changes?

zachmu (commented):
Nice work, just a few comments and need a test for the case sensitivity.

Note that a lot of our commands have this problem because they pre-date the SQL engine. But dolt table names should always match case-insensitive, as they do in SQL.

go/cmd/dolt/commands/schcmds/tags.go
VinaiRachakonda: Does the GetTableInsensitive in root_val solve this?

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +88,63 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+
+		_ = sch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
+			rows = append(rows, sql.NewRow(foundTableKey, col.Name, tag))
+
+			return false, err
+		})
+
+	}
+
+	outputFmt, verr := commands.GetResultFormat("tabular")
```
zachmu, line 126: No need to call GetResultFormat here, just assign the var with the tabular const value directly

go/cmd/dolt/commands/schcmds/tags.go
```diff
@@ -87,24 +88,63 @@ func (cmd TagsCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		outputFmt, verr = commands.GetResultFormat(formatSr)
+
+		if verr != nil {
+			return commands.HandleVErrAndExitCode(errhand.VerboseErrorFromError(verr), usage)
+		}
 	}
+
+	err = commands.PrettyPrintResults(ctx, outputFmt, headerSchema, sql.RowsToRowIter(rows...))
+
+	if err != nil {
```
zachmu, line 145: Since this is the end of the function, you can just skip the condition and always return HandleVerrAndExitCode (it handles a nil verr correctly)

zachmu (approved):
LG, just a couple comments

PR #1052, opened 2020-11-25
Vinai/dolt commit author no config

VinaiRachakonda:
Add a bats test that models the following behavior.

1. Unsets name and user. 
2. Makes a sql change
3. Add a commit with --author

### DIFF
--- a/bats/config.bats
+++ b/bats/config.bats
@@ -148,15 +148,19 @@ teardown() {
       PRIMARY KEY (pk)
     );"
 
-     dolt config --global --unset user.name
-     dolt config --global --unset user.email
+    dolt config --global --unset user.name
+    dolt config --global --unset user.email
 
-     run dolt config --list
-     [ "$status" -eq 0 ]
-     [ "${#lines[@]}" -eq 0 ]
+    run dolt config --list
+    [ "$status" -eq 0 ]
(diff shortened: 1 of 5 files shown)


bats/config.bats
```diff
@@ -131,3 +131,32 @@ teardown() {
...
+     dolt config --global --unset user.email
+
+     run dolt config --list
+     [ "$status" -eq 0 ]
+     [ "${#lines[@]}" -eq 0 ]
+
+      dolt add .
+      run dolt commit --author="John Doe <john@doe.com>" -m="works"
+      [ "$status" -eq 0 ]
+
```
andrew-wm-arthur, line 161: can you run `dolt log` and assert the author was written correctly?

VinaiRachakonda: Added!

andrew-wm-arthur: This asserts that message was written correctly, but we also need to check that the author info was written correctly

VinaiRachakonda: Whoops! Think I got it now?

andrew-wm-arthur: Ideally you'd check both, but this looks right

andrew-wm-arthur (approved):
LGTM

PR #1053, opened 2020-11-25
Support opening XLSXReader from binary for xlsx import in dolthubapi

tbantle22:


go/libraries/doltcore/table/untyped/xlsx/marshaling_test.go
```diff
@@ -63,10 +64,36 @@ func TestDecodeXLSXRows(t *testing.T) {
...
+
+	if stateCols != nil || employeeCols == nil {
+		t.Fatal("error")
+	}
+}
+
+func getBytesFromXlsx() []byte {
+	path := "test_files/employees.xlsx"
+
+	file, err := openFile(path)
```
reltuk, line 88: Does this work as just

```go
f, err := os.Open("test_files/employees.xlsx")
if err != nil {
    panic(err)
}
defer f.Close()
bs, err := ioutil.ReadAll(f)
if err != nil {
    panic(err)
}
return bs
```

Any reason to actually parse the xslx here?

reltuk (approved):
LGTM! One small comment.

go/libraries/doltcore/table/untyped/xlsx/marshaling_test.go
tbantle22: No, that works I fixed here: https://github.com/dolthub/dolt/pull/1053/commits/a19577ef29e2247f28c21f57084dd49fbefec4da

I think my googling led me astray 😁

PR #1056, opened 2020-11-29
Vinai/add dolt sql commit

VinaiRachakonda:
This pr seeks to add 'DOLT_COMMIT' functionality to the SQL Interface. It currently supports all of the current cli commit behavior. It does work with doc diffs!

You can test on the CLI like this: `SELECT DOLT_COMMIT('-m', 'This is a commit')`


go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+	ap.SupportsString(commitMessageArg, "m", "msg", "Use the given {{.LessThan}}msg{{.GreaterThan}} as the commit message.")
+	ap.SupportsFlag(allowEmptyFlag, "", "Allow recording a commit that has the exact same data as its sole parent. This is usually a mistake, so it is disabled by default. This option bypasses that safety.")
+	ap.SupportsString(dateParam, "", "date", "Specify the date used in the commit. If not specified the current system time is used.")
+	ap.SupportsFlag(forceFlag, "f", "Ignores any foreign key warnings and proceeds with the commit.")
+	ap.SupportsString(authorParam, "", "author", "Specify an explicit author using the standard A U Thor <author@example.com> format.")
+	return ap
+}
+
+// we are more permissive than what is documented.
+var supportedLayouts = []string{
```
andrew-wm-arthur, line 62: we shouldn't duplicate this logic here. If there's something we're doing on the CLI and in the SQL engine, let's move it into the `Actions` package

bats/sql-commit.bats
```diff
@@ -0,0 +1,68 @@
...
+    dolt add .
+    run dolt sql -q "SELECT DOLT_COMMIT('-m', 'Commit1', '--author', 'John Doe <john@doe.com>')"
+    [ "$status" -eq 0 ]
+
+    run dolt log
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Commit1" ]] || false
+    regex='John Doe <john@doe.com>'
+    [[ "$output" =~ "$regex" ]] || false
+}
```
andrew-wm-arthur, line 68: Let's add one more test asserting that our commit is reflected in `dolt_log` and `dolt_commits`

andrew-wm-arthur (approved):
General approach looks good, I like the tests. A bit of refactoring is needed around `env.DoltEnv` and the arg parsing

go/libraries/doltcore/env/actions/commit.go
```diff
@@ -130,13 +130,13 @@ func CommitStaged(ctx context.Context, dEnv *env.DoltEnv, props CommitStagedProp
 		}
 	}
 
-	h, err := dEnv.UpdateStagedRoot(ctx, srt)
+	h, err := writer.UpdateStagedRoot(ctx, srt)
```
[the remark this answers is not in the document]
VinaiRachakonda: I'm actually a little confused about how this would work. Saving the hash in the UpdateStagedHoot requires a fs passed in through the dEnv that is not part of the RepoState.

go/libraries/doltcore/env/repo_state.go
```diff
@@ -169,3 +174,44 @@ func (rs *RepoState) WorkingHash() hash.Hash {
...
+func (rs *RepoState) IsMergeActive() bool {
+	return rs.Merge != nil
+}
+
+func (rs *RepoState) GetMergeCommit() string {
+	return rs.Merge.Commit
+}
+
+// TODO: This functionality is currently unsupported by the SQL session.
+func (rs *RepoState) GetAllValidDocDetails() ([]doltdb.DocDetails, error) {
```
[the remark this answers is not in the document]
VinaiRachakonda: I'm actually not sure now that you mention it...

go/cmd/dolt/commands/commit.go
```diff
@@ -123,7 +123,7 @@ func (cmd CommitCmd) Exec(ctx context.Context, commandStr string, args []string,
 		}
 	}
 
-	err = actions.CommitStaged(ctx, dEnv, actions.CommitStagedProps{
+	err = actions.CommitStaged(ctx, dEnv.DoltDB, dEnv.RepoStateReader(), dEnv.RepoStateWriter(), actions.CommitStagedProps{
```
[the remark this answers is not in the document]
VinaiRachakonda: I think it might not be worth it since there will a large refactor whenever we think there is going to be multiple writers. I also like having all params listed as that clearly states that this function is either modifying or just reading.

go/gen/proto/dolt/services/eventsapi/v1alpha1/client_event.pb.go
```diff
@@ -23,13 +23,14 @@
 package eventsapi
 
 import (
+	reflect "reflect"
```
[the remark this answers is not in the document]
VinaiRachakonda: Any reason why not? This would just cause a problem on someone else's branch right?

go/libraries/doltcore/env/environment.go
```diff
@@ -367,6 +367,42 @@ func (dEnv *DoltEnv) UpdateWorkingRoot(ctx context.Context, newRoot *doltdb.Root
...
+
+func (r *repoStateReader) GetAllValidDocDetails() ([]doltdb.DocDetails, error) {
+	return r.dEnv.GetAllValidDocDetails()
+}
+
+func (dEnv *DoltEnv) RepoStateReader() RepoStateReader {
+	return &repoStateReader{dEnv}
+}
+
 type repoStateWriter struct {
```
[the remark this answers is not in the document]
VinaiRachakonda: I agree. But i think that could mean changing what is stored in the Repo State. Let's discuss offline.

go/libraries/doltcore/env/repo_state.go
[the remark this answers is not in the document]
VinaiRachakonda: Ok I think I get it now. Since the sqle is taking in a repo state reader derived from dEnv it able to deal with doc changes. That might have to change in the future.

go/cmd/dolt/commands/commit.go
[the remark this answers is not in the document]
andrew-wm-arthur: yeah that's reasonable

go/libraries/doltcore/env/repo_state.go
```diff
@@ -169,3 +174,44 @@ func (rs *RepoState) WorkingHash() hash.Hash {
...
+	return nil, nil
+}
+
+func HeadRoot(ctx context.Context, ddb *doltdb.DoltDB, reader RepoStateReader) (*doltdb.RootValue, error) {
```
zachmu, line 191: Method doc for all exported methods

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+
+const (
+	allowEmptyFlag   = "allow-empty"
+	dateParam        = "date"
+	commitMessageArg = "message"
+	forceFlag        = "force"
+	authorParam      = "author"
+)
+
+func createArgParser() *argparser.ArgParser {
```
zachmu, line 51: Same comment as Andy made about duplicating logic. Share the arg parser used by the cmd package.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	// Get the commit message.
+	msg, msgOk := apr.GetValue(commitMessageArg)
+	if !msgOk {
+		return nil, fmt.Errorf("Must provide commit message.")
+	}
+
+	// Specify the time if the date parameter is not.
+	t := time.Now()
```
zachmu, line 175: Get this from the context

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+	err = actions.CommitStaged(ctx, doltdb, rsr, rsw, actions.CommitStagedProps{
+		Message:          msg,
+		Date:             t,
+		AllowEmpty:       apr.Contains(allowEmptyFlag),
+		CheckForeignKeys: !apr.Contains(forceFlag),
+		Name:             name,
+		Email:            email,
+	})
+
+	return "Staged", err
```
zachmu, line 194: Expression should evaluate to the hash of the new commit. Should have a test of this.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+func (d DoltCommitFunc) IsNullable() bool {
+	return false
+}
+
+func (d DoltCommitFunc) WithChildren(children ...sql.Expression) (sql.Expression, error) {
+	return NewDoltCommitFunc(children...)
+}
+
+func (d DoltCommitFunc) Resolved() bool {
+	return true
```
zachmu, line 219: Need to check if all children are resolved

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, fmt.Errorf("Could not load the %s RepoStateWriter", dbName)
+	}
+
+	ap := createArgParser()
+
+	// Get the args for DOLT_COMMIT.
+	args := make([]string, 1)
+	for i := range d.children {
+		temp := d.children[i].String()
+		str := trimQuotes(temp)
```
zachmu, line 149: This is unnecessary, see above

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+	if !ok {
+		return nil, fmt.Errorf("Could not load the %s RepoStateWriter", dbName)
+	}
+
+	ap := createArgParser()
+
+	// Get the args for DOLT_COMMIT.
+	args := make([]string, 1)
+	for i := range d.children {
+		temp := d.children[i].String()
```
zachmu, line 148: You don't want .String() here, you want to call child.Eval(ctx, row). See other functions for details.

zachmu: After you eval the child, you'll want to convert the value to a String using the sql package, something like sql.Longtext.Convert(val)

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,224 @@
...
+	rsw, ok := dSess.GetDoltDBRepoStateWriter(dbName)
+
+	if !ok {
+		return nil, fmt.Errorf("Could not load the %s RepoStateWriter", dbName)
+	}
+
+	ap := createArgParser()
+
+	// Get the args for DOLT_COMMIT.
+	args := make([]string, 1)
```
zachmu, line 146: Why 1 here?

zachmu (commented):
This looks pretty good, but has some issues.

The main one is Andy's comment about not duplicating logic from the cmd / action packages. You should get the array of string commands, then dispatch them to some exported function in the cmd package.

go/libraries/doltcore/env/environment.go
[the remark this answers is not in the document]
VinaiRachakonda: Think I got a fix. Pretty big change to get rid of this abstraction though.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
VinaiRachakonda: Do we have a bats manner for comparing hashes? Or do you want a go test?

go/libraries/doltcore/env/actions/commit.go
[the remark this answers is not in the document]
VinaiRachakonda: Is this is a blocker?

go/libraries/doltcore/env/environment.go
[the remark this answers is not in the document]
VinaiRachakonda: Deciding to keep this abstraction for now and rethink later.

bats/sql-commit.bats
```diff
@@ -0,0 +1,92 @@
...
+SQL
+    dolt add .
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "DOLT_COMMIT with a message and author" {
+    run dolt sql -q "SELECT DOLT_COMMIT('-m', 'Commit1', '--author', 'John Doe <john@doe.com>')"
```
zachmu, line 26: Should check that this returns the expected hash

go/cmd/dolt/commands/commit.go
```diff
@@ -74,32 +63,22 @@ func (cmd CommitCmd) Description() string {
 
 // CreateMarkdown creates a markdown file containing the helptext for the command at the given path
 func (cmd CommitCmd) CreateMarkdown(fs filesys.Filesys, path, commandStr string) error {
-	ap := cmd.createArgParser()
+	ap := actions.CreateCommitArgParser()
```
bheni, line 66: Not a fan of this in the actions package.

go/libraries/doltcore/env/actions/commit.go
```diff
@@ -40,6 +44,73 @@ type CommitStagedProps struct {
 	Email            string
 }
 
+// we are more permissive than what is documented.
+var SupportedLayouts = []string{
```
bheni, line 48: actions is not a command line parsing package.  We shouldn't just throw things in here because its convenient and accessible in the other packages that you are working in.

go/cmd/dolt/commands/commit.go
VinaiRachakonda: There was some repeated logic with it in dolt_commit. I felt that it was best to not be repeated. Should that change so that commit.go and dolt_commit.go have their own argparser?

bheni: I agree that not repeating code is great, but you shouldn't stick it in a random package just because its accessible from both of the packages you are working in.  You can create new packages.

go/libraries/doltcore/env/actions/commit.go
```diff
@@ -61,14 +132,14 @@ func GetNameAndEmail(cfg config.ReadableConfig) (string, string, error) {
 	return name, email, nil
 }
 
-func CommitStaged(ctx context.Context, dEnv *env.DoltEnv, props CommitStagedProps) error {
+func CommitStaged(ctx context.Context, ddb *doltdb.DoltDB, rsr env.RepoStateReader, rsw env.RepoStateWriter, props CommitStagedProps) (string, error) {
```
bheni, line 135: I'd create a combined interface

```type RepoStateReadWriter interface {
    RepoStateReader
    RepoStateWriter
}```

and pass that in here rather than having two arguments

Comment (zachmu) [inline go/libraries/doltcore/env/actions/commit.go:135]: Method doc on what this method does and its return value plz

File: go/libraries/doltcore/sqle/dfunctions/dolt_commit.go [hunk 20]
```diff
@@ -0,0 +1,171 @@
...
+	rsw, ok := dSess.GetDoltDBRepoStateWriter(dbName)
+
+	if !ok {
+		return nil, fmt.Errorf("Could not load the %s RepoStateWriter", dbName)
+	}
+
+	ap := actions.CreateCommitArgParser()
+
+	// Get the args for DOLT_COMMIT.
+	args := make([]string, 0)
```
zachmu, line 77: If you're building an array with append() calls like this in go, then it's more idiomatic to declare the slice like:

`var args []string`

On the other hand, if you know its size in advance, like you do here, then it's usually better (and more performant) to initialize it like:

`args := make([]string, len(d.children))`

Then assign to it with the index:

`args[i] = ...`

Instead of using append.

Either is fine, but pick one.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,171 @@
...
+	for i := range d.children {
+		temp, err := d.children[i].Eval(ctx, row)
```
zachmu, line 79: no `temp` please, be descriptive. `childVal` or just `val` would be better than `temp`

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,171 @@
...
+			return nil, err
+		}
+
+		text, err := sql.Text.Convert(temp)
+
+		if err != nil {
+			return nil, err
+		}
+
+		str := trimQuotes(text.(string))
```
zachmu, line 91: Should be a no-op now, remove this func

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,171 @@
...
+		args = append(args, str)
+	}
+
+	apr := cli.ParseArgs(ap, args, nil)
```
zachmu, line 95: From here down significantly duplicates the logic in the cmd / action packages. You should extract a library method which does this and use it in both places, unless the business logic is going to differ between the command line and the SQL interface (I don't think it is).

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if len(param) > 0 && param[len(param)-1] == '"' {
+		param = param[:len(param)-1]
+	}
+
+	return param
+}
+
+func (d DoltCommitFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
```
zachmu, line 51: Method doc

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+const DoltCommitFuncName = "dolt_commit"
+
+type DoltCommitFunc struct {
+	children []sql.Expression
+}
+
+// NewDoltCommitFunc creates a new DoltCommitFunc expression.
+func NewDoltCommitFunc(args ...sql.Expression) (sql.Expression, error) {
```
zachmu, line 34: Document what the expression returns

go/libraries/doltcore/env/environment.go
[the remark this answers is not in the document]
bheni: Yeah I don't understand these repoStateReader / repoStateWriter structs.  The DoltEnv should implement these interfaces.

go/cmd/dolt/commands/commit.go
VinaiRachakonda: Do you think the dolt cli package is the appropriate location? Maybe arg_parsers_helpers.go?

go/libraries/doltcore/diff/diffs.go
```diff
@@ -136,33 +136,33 @@ func (nd *DocDiffs) Len() int {
...
-func GetDocDiffs(ctx context.Context, dEnv *env.DoltEnv) (*DocDiffs, *DocDiffs, error) {
-	docDetails, err := dEnv.GetAllValidDocDetails()
+func GetDocDiffs(ctx context.Context, ddb *doltdb.DoltDB, rsr env.RepoStateReader) (*DocDiffs, *DocDiffs, error) {
+	docDetails, err := rsr.GetAllValidDocDetails()
 	if err != nil {
 		return nil, nil, err
 	}
 
-	workingRoot, err := dEnv.WorkingRoot(ctx)
+	workingRoot, err := env.WorkingRoot(ctx, ddb, rsr)
```
zachmu, line 145: Not crazy about these interfaces -- WorkingRoot etc. should be methods on RepoStateReader

go/libraries/doltcore/diff/diffs.go
```diff
@@ -136,33 +136,33 @@ func (nd *DocDiffs) Len() int {
...
 	}
 
-	notStagedDocDiffs, err := NewDocDiffs(ctx, dEnv, workingRoot, nil, docDetails)
+	notStagedDocDiffs, err := NewDocDiffs(ctx, workingRoot, nil, docDetails)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	headRoot, err := dEnv.HeadRoot(ctx)
+	headRoot, err := env.HeadRoot(ctx, ddb, rsr)
```
zachmu, line 155: Same

go/libraries/doltcore/diff/diffs.go
```diff
@@ -136,33 +136,33 @@ func (nd *DocDiffs) Len() int {
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	stagedRoot, err := dEnv.StagedRoot(ctx)
+	stagedRoot, err := env.StagedRoot(ctx, ddb, rsr)
```
zachmu, line 160: Same

go/libraries/doltcore/env/actions/commit.go
VinaiRachakonda: I actually disagree with this. I think we should be explicit about which methods are only doing the reading and which are only doing the writing. I think a combined interface would obfuscate it.

zachmu (approved):
Revert the pb.go files please (they're build artifacts and you didn't change anything that touches them).

Most of this feedback can be pushed off to a followup PR in the interest of time and making incremental progress, but I definitely want to see better code sharing and cleaner abstractions on the stuff you have touched.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
VinaiRachakonda: I think with how error handling is done between the cmd and dfunctions package I felt that extracting would just obfuscate too much.

go/libraries/doltcore/diff/diffs.go
VinaiRachakonda: Agree. I will refactor this in a separate pr.

PR #1059, opened 2020-12-02
pass in-memory gc gen when we conjoin

andrew-wm-arthur:


reltuk (approved):
Let's get a test for this after this lands.

PR #1060, opened 2020-12-02
Added IS NULL and IS NOT NULL to indexing

Hydrocharged:
So, right now (in `master`) we iterate over any existing `NULL` values in our indexes. This is due to how `NULL`s are ordered relative to the other values. Initially, I made the mistake of thinking that `NULL` always appeared at the beginning of the map, and thus rewrote a bit of the range implementation to have a special `NULL` section right before `BelowAll`, so it went `BelowAll`, `BelowValues`, `Values...`. This way, a range representing `LessThan(6)` changed from `Range(BelowAll, 6)` to `Range(BelowValues, 6)`, and the range of NULLs were `Range(BelowAll, BelowValues]`.

This was all fine and dandy until I realized that the `NULL` position is dependent on the noms kind, so `NULL` will be at the beginning of the map for some types, and at the end for others. I scrapped all of my work and made it as you see here: some iterators will return more values than they should, but they'll be filtered out at the SQL engine, so the correct overall result is returned. Definitely not ideal, but it's "no worse" than the current behavior...it's just more visible now that `NULL`s are reachable through indexing.

I didn't want to spend any more time working on this so I can move on to something else.


go/libraries/doltcore/sqle/mergeable_indexes_test.go
```diff
@@ -1263,3 +1276,187 @@ func TestMergeableIndexes(t *testing.T) {
...
+		(8, 18, 28),
+		(9, 19, 29);`)
+
+	tests := []struct {
+		whereStmt   string
+		finalRanges []lookup.Range
+		pks         []int64
+	}{
+		{
+			"v1 IS NULL",
```
zachmu, line 1307: What about IS NULL AND IS NOT NULL?

zachmu: IS NULL OR IS NOT NULL?

zachmu (approved):

PR #1068, opened 2020-12-03
Add -a flag to cli and DOLT_COMMIT

VinaiRachakonda:
This pr adds a `-a` flag to `dolt commit` and `DOLT_COMMIT`. It stages all tables. 

It also cleans up some of the previous work done in #1056 by removing all method handlers in repo_state and moving them to the RepoStateReader and RepoStateWriter

It does not refactor the RSR/RSW interfaces in environment.go. This will be done in a subsequent pr.

### DIFF
--- a/bats/sql-commit.bats
+++ b/bats/sql-commit.bats
@@ -26,6 +26,11 @@ teardown() {
     [ $status -eq 0 ]
     DCOMMIT=$output
 
+    # Check that everything was added
+    run dolt diff
+    [ "$status" -eq 0 ]
+    [ "$output" = "" ]
+
     run dolt log
     [ $status -eq 0 ]
     [[ "$output" =~ "Commit1" ]] || false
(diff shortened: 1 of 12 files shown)


bats/sql-commit.bats
```diff
@@ -53,7 +57,7 @@ teardown() {
 }
 
 @test "DOLT_COMMIT with just a message reads session parameters" {
-    run dolt sql -q "SELECT DOLT_COMMIT('-m', 'Commit1')"
+    run dolt sql -q "SELECT DOLT_COMMIT('-a', '-m', 'Commit1')"
```
zachmu, line 60: Make sure you have some variants that don't specify -a as well

zachmu (approved):
LG, nice cleanup on the repo state reader / writer interface.

You need to also add bats tests for the `dolt commit` command with this flag, otherwise LGTM

PR #1074, opened 2020-12-04
Add dolt_commit error check when autocommit is off

VinaiRachakonda:
Fails loudly when autocommit is off for dolt_commit.

### DIFF
--- a/bats/sql-server.bats
+++ b/bats/sql-server.bats
@@ -65,6 +65,57 @@ teardown() {
     [[ "$output" =~ "one_pk" ]] || false
 }
 
+
+@test "test dolt sql interface works properly with autocommit" {
+    skiponwindows "Has dependencies that are missing on the Jenkins Windows installation."
+
+    cd repo1
+    start_sql_server repo1
+
+    # No tables at the start
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "No tables in working set" ]] || false
+
+    # create table with autocommit off and verify there are still no tables
+    server_query 0 "CREATE TABLE one_pk (
+        pk BIGINT NOT NULL COMMENT 'tag:0',
+        c1 BIGINT COMMENT 'tag:1',
+        c2 BIGINT COMMENT 'tag:2',
+        PRIMARY KEY (pk)
+    )" ""
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "No tables in working set" ]] || false
+
+    # check that dolt_commit throws an error when autocommit is off
+    run dolt sql -q "SELECT DOLT_COMMIT('-a', '-m', 'Commit1')"
+    [ "$status" -eq 1 ]
+
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "No tables in working set" ]] || false
+
+    # create table with autocommit on and verify table creation
+    server_query 1 "CREATE TABLE one_pk (
+        pk BIGINT NOT NULL COMMENT 'tag:0',
+        c1 BIGINT COMMENT 'tag:1',
+        c2 BIGINT COMMENT 'tag:2',
+        PRIMARY KEY (pk)
+    )" ""
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "one_pk" ]] || false
+
+    # check that dolt_commit works properly when autocommit is on
+    run dolt sql -q "SELECT DOLT_COMMIT('-a', '-m', 'Commit1')"
+    [ "$status" -eq 0 ]
+
+    run dolt log
+    [ $status -eq 0 ]
+    [[ "$output" =~ "Commit1" ]] || false
+}
+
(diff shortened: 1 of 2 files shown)


bats/sql-server.bats
```diff
@@ -65,6 +65,57 @@ teardown() {
...
+        c1 BIGINT COMMENT 'tag:1',
+        c2 BIGINT COMMENT 'tag:2',
+        PRIMARY KEY (pk)
+    )" ""
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "No tables in working set" ]] || false
+
+    # check that dolt_commit throws an error when autocommit is off
+    run dolt sql -q "SELECT DOLT_COMMIT('-a', '-m', 'Commit1')"
```
zachmu, line 92: This is wrong... any single command line SQL query like this should be auto committed by default.

Easiest way to fix this is probably to set the session variable when we start the command line SQL session.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -41,6 +41,12 @@ func (d DoltCommitFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
 	dbName := ctx.GetCurrentDatabase()
 	dSess := sqle.DSessFromSess(ctx.Session)
 
+	_, val := dSess.Session.Get(sql.AutoCommitSessionVar)
```
zachmu, line 44: I wouldn't check for this specifically. The relevant question isn't whether autocommit is on, but whether there are any changes in the working set (for -a) or staged (without -a). If there aren't, and we haven't set --allow-empty, then a commit will be a no-op and should error out.

zachmu (commented):
Not quite the right way to solve this.

It's OK for the error message to mention @@autocommit, but the business logic shouldn't depend on that detail.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
VinaiRachakonda: How could you check for changes in the staged root?

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -134,6 +149,34 @@ func (d DoltCommitFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
...
+func hasWorkingSetChanges(rsr env.RepoStateReader) bool {
+	if rsr.WorkingHash() != rsr.StagedHash() {
+		return true
+	}
+
+	return false
+}
+
+func hasStagedSetChanges(ctx *sql.Context, rsr env.RepoStateReader) (bool, error) {
+	root, err := rsr.HeadRoot(ctx)
```
zachmu, line 161: Make a TODO here. When you follow up with the refactor of RepoStateReader, then you should address the fact that you're returning a root object here, and shouldn't be. Just need a CommitSpec that you can get the hash from.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -134,6 +149,34 @@ func (d DoltCommitFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
...
+		return false, err
+	}
+
+	headHash, err := root.HashOf()
+
+	if err != nil {
+		return false, err
+	}
+
+	if rsr.StagedHash() != headHash {
```
zachmu, line 173: Same comment here: just `return rsr.StagedHash() != headHash, nil`

bats/sql-server.bats
```diff
@@ -65,6 +65,57 @@ teardown() {
...
+        c1 BIGINT COMMENT 'tag:1',
+        c2 BIGINT COMMENT 'tag:2',
+        PRIMARY KEY (pk)
+    )" ""
+    run dolt ls
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "one_pk" ]] || false
+
+    # check that dolt_commit works properly when autocommit is on
+    run dolt sql -q "SELECT DOLT_COMMIT('-a', '-m', 'Commit1')"
```
zachmu, line 111: Also should have a test of the negative case, where there are no staged changes / no changes to stage. With and without -a.

go/libraries/doltcore/sqle/dfunctions/dolt_commit.go
```diff
@@ -81,10 +82,24 @@ func (d DoltCommitFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
 
 	apr := cli.ParseArgs(ap, args, nil)
 
-	// Check if the -all param is provided. Stage all tables if so.
 	allFlag := apr.Contains(cli.AllFlag)
+	allowEmpty := apr.Contains(cli.AllowEmptyFlag)
```
zachmu, line 86: This is substantially duplicating logic from the command / action packages. After you finish fixing RepoStateReader, you should take a pass at eliminating some of this duplication.

zachmu (approved):

PR #1080, opened 2020-12-06
Fixed shell error loop on UNIQUE violation

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1079


zachmu (commented):
This... seems fine? A test of this behavior would be better

Hydrocharged:
How do we test the interactive shell? I wasn't too sure so I didn't include it.

Hydrocharged:
Added a go test. Doesn't test the exact issue, but stems from the same root cause.

zachmu (approved):
LGTM

PR #1082, opened 2020-12-07
Fixed internal index comparisons considering unnecessary parameters

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1081


bats/regression-tests.bats
```diff
@@ -0,0 +1,54 @@
...
+setup() {
+    setup_common
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "regression-tests: dolt issue #1081" {
+    dolt sql <<SQL
```
zachmu, line 13: Any way to pare this down? Pretty hard to see what parts of this large schema are actually relevant to the bug.

go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -255,7 +255,10 @@ func (di *doltIndex) keysToTuple(keys []interface{}) (types.Tuple, error) {
 	}
 	var vals []types.Value
 	for i, col := range di.cols {
-		val, err := col.TypeInfo.ConvertValueToNomsValue(keys[i])
+		// As an example, if our TypeInfo is Int8, we should not fail to create a tuple if given the value 9001.
+		// Instead, the index will return no values. Comparisons should happen at the widest type, only storage
```
zachmu, line 259: "return no values" isn't right... we could have a < comparison, in which case we return all values in the index.

bats/regression-tests.bats
zachmu: Also the naming scheme is pretty opaque. Would be nice if a) the set of columns were minimal and b) the table / column names were a bit more meaningful.

zachmu (approved):
Test is a bit o_0 otherwise LG

go/libraries/doltcore/sqle/dolt_index.go
Hydrocharged: Good point, I'll update the comment.

bats/regression-tests.bats
Hydrocharged: This is dramatically pared down from the original while retaining the issue. Naming was because it was observed from private  data, and I renamed everything so that we can have a public test that matches the private one. I agree that the names aren't at all useful though.

PR #1084, opened 2020-12-07
bats/: keyless spec

andrew-wm-arthur:
This is a set of skipped BATS tests that provide a spec for keyless tables. 
- `keyless.bats` contains specs that are "uncontested", ie they will work the same in both implementations
- `keyless-bag.bats` contains specs of the "bag semantics" implementation of keyless tables. This implementation is an unordered multi-set. Rows are read/written in sorted order. It has complete history invariance at the price of arbitrary row-ordering and cell-wise diffs.
- `keyless-map.bats` contains specs of the "generated hidden key" implementation of keyless tables. This implementation will created a hidden primary key for each row on insert. It will use content-addressability to support diffs and merges. This implementation leverages the hidden key to maintain arbitrary row ordering and cell-wise diffs. The cost is that two tables with the same row contents can have different storage representations depending on their mutation history.


bats/keyless-bag.bats
```diff
@@ -0,0 +1,361 @@
...
+    dolt sql -q "INSERT INTO keyless VALUES (7,7),(8,8),(9,9);"
+    dolt commit -am "inserted on master"
+
+    dolt checkout other
+    dolt sql -q "INSERT INTO keyless VALUES (9,9),(8,8),(7,7);"
+    dolt commit -am "inserted on other"
+
+    run dolt merge master
+    [ $status -eq 0 ]
+    run dolt sql -q "SELECT * FROM keyless WHERE c0 > 6;" -r csv
```
reltuk, line 226: Does this also assert `count(*) where c0 > 6` is `3`?

bats/keyless-bag.bats
```diff
@@ -0,0 +1,361 @@
...
+    [ $status -eq 0 ]
+    [[ "$lines[@]" = "7,7" ]] || false
+    [[ "$lines[@]" = "8,8" ]] || false
+    [[ "$lines[@]" = "9,9" ]] || false
+}
+
+# convergent row data has convergent storage representation
```
reltuk, line 233:
```suggestion
# convergent row data history with convergent data has convergent storage representation
```

bats/keyless-bag.bats
```diff
@@ -0,0 +1,361 @@
...
+    dolt sql -q "INSERT INTO keyless VALUES (7,7),(8,8),(9,9);"
+    dolt commit -am "inserted on master"
+
+    dolt checkout other
+    dolt sql -q "INSERT INTO keyless VALUES (7,7),(7,7),(8,8),(9,9);"
+    dolt commit -am "inserted on other"
+
+    run dolt diff master
+    [ $status -eq 0 ]
+    [[ "$lines[@]" = "|  +  | 7  | 7  |" ]] || false
```
reltuk, line 291: Should this assert nothing else here too...?

bats/keyless-map.bats
```diff
@@ -0,0 +1,390 @@
...
+    [[ "$lines[@]" = "|  >  | 7  | 27  |" ]] || false
+    [[ "$lines[@]" = "|  <  | 8  | 18  |" ]] || false
+    [[ "$lines[@]" = "|  >  | 8  | 28  |" ]] || false
+    [[ "$lines[@]" = "|  <  | 9  | 19  |" ]] || false
+    [[ "$lines[@]" = "|  >  | 9  | 29  |" ]] || false
+}
+
+# where in-place updates are divergent, both versions are kept on merge
+# same for hidden key and bag semantics
+@test "keyless merge with in-place updates (branches)" {
```
reltuk, line 178: I would expect this to merge conflict?

reltuk (commented):
In general these look great. I think it might be instructive to add more merge cases...

1) We delete a row on both branches and merge.
2) We delete 1 of a row which appears more than once on both branches and merge.
3) We update a different column of the same row on each branch and merge.
4) We update the same column of the same row on each branch and merge. (We have this one I think...but in `-map` its spec is wrong?)
5) Same as 3 and 4, but we do it for a single row in a table where the row appears more than once.

Looking at those cases...I think there's a problem with 5 for the `-map` implementation. If the update statement on the two branches hit the same hidden extrinsic ID, then it will merge one way, and if they hit two different hidden extrinsic IDs it will merge a different way. Unpredictable behavior seems like the worst case...

bats/keyless.bats
```diff
@@ -0,0 +1,344 @@
...
+
+    run dolt diff master
+    [ $status -eq 0 ]
+    [[ "$output" = "|  -  | 1  | 1  |" ]] || false
+    [[ "$output" = "|  -  | 1  | 1  |" ]] || false
+    [[ "$output" = "|  -  | 1  | 1  |" ]] || false
+    [[ "$output" = "|  -  | 1  | 1  |" ]] || false
+}
+
+@test "keyless merge duplicate deletes" {
```
andrew-wm-arthur, line 233: @reltuk is this the correct behavior to you?

PR #1094, opened 2020-12-08
go/libraries/doltcore/{mvdata, table}: moved pipeline & data-location packages inside mvdata

andrew-wm-arthur:
before & after: 
image


zachmu:
OK, I think I'm on board with this. Thanks for taking this on and doing the work.

Ultimately I think a better split of functionality would be mvdata (pipeline and friends) and filetypes (text  / binary representations, readers and writers of same). It would be nice to be able to de/serialize these text formats outside the context of import / export, so putting them in their own packages, and using from mvdata, makes the most sense to me. Could do that as a followup, or as part of this PR. Worth thinking about though.

PR #1097, opened 2020-12-09
Fixed table import allowing NULLs in the primary key

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1096


zachmu (approved):
LG!

PR #1102, opened 2020-12-10
go/libraries/doltcore/{doltdb,table}: remove row access methods from doltdb.Table

andrew-wm-arthur:
Removed:
- Table.GetRowByPKVals()
- Table.GetRow()
- Table.GetRows()

Had do to some refactoring along the way to fix dependency cycles. 
Reversed dependency `rowconv -> pipeline` to `pipeline -> rowconv`

### DIFF
--- a/go/libraries/doltcore/schema/alterschema/addcolumn.go
+++ b/go/libraries/doltcore/schema/alterschema/addcolumn.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/row"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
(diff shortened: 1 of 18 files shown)


go/libraries/doltcore/schema/alterschema/addcolumn.go
```diff
@@ -105,7 +104,7 @@ func updateTableWithNewSchema(ctx context.Context, tblName string, tbl *doltdb.T
 	}
 
 	err = rowData.Iter(ctx, func(k, v types.Value) (stop bool, err error) {
-		oldRow, _, err := tbl.GetRow(ctx, k.(types.Tuple), newSchema)
+		oldRow, err := row.FromNoms(newSchema, k.(types.Tuple), v.(types.Tuple))
```
andrew-wm-arthur, line 107: This is temporary for this change, we'll need a general purpose way to set defaults

zachmu (approved):
Looks fine, but you should add a note in your commits / merge comment about the rationale for this change.

PR #1108, opened 2020-12-11
Fixed dolt status output incorrectly displayed for staged files

VyasSathya:
The function `printStagedDiffs` always returned 0, even when there were diffs not staged. This return was also causing it to print in `printStatus ` "nothing to commit, working tree clean". This was not the case.  

I changed  `printStagedDiffs` to return the number of the staged tables plus the number of staged docs instead. This prevents it from entering the if statement with the print also.

I've attached a picture of the bug with debug statements: 

Bug

and here is it after the fix:

Fixed

### DIFF
--- a/go/cmd/dolt/commands/status.go
+++ b/go/cmd/dolt/commands/status.go
@@ -185,6 +185,7 @@ func printStagedDiffs(wr io.Writer, stagedTbls []diff.TableDelta, stagedDocs *di
 		}
 
 		iohelp.WriteLine(wr, color.GreenString(strings.Join(lines, "\n")))
+		return len(stagedTbls) + stagedDocs.Len()
 	}
 
 	return 0


timsehn:
I'm confused as to what exactly the user sees that is borken?

timsehn:
I don't think we should return the number of files changed. This is the exit code of the program. 0 means `OK` and any other exit code means `ERROR`. What does Git return?

VyasSathya:
Hi Tim, the zero doesn't represent the error code in this case. The return number is being used to decide whether or not to print the 'nothing to commit' message. I believe it should be similar to the notStaged function.

VyasSathya:
In the file:
_dolthub\dolt\go\cmd\dolt\commands\status.go_
```
	n := printStagedDiffs(cli.CliOut, stagedTbls, stagedDocs, true)
	n = printDiffsNotStaged(ctx, dEnv, cli.CliOut, notStagedTbls, notStagedDocs, true, n, workingTblsInConflict)

	if dEnv.RepoState.Merge == nil && n == 0 {
		cli.Println("nothing to commit, working tree clean")
	}
```
if _printStagedDiffs_ returns zero all the time, even if there are staged files it will output "nothing to commit, working tree clean" whereas in git it won't output that. 

@timsehn 
@bheni

bheni (approved):
lgtm

PR #1109, opened 2020-12-11
Attempt to add default decimal type to FromKind

tbantle22:
Not sure if this is the behavior we want or if this is correct at all, but I'm implementing schema updates on the front end and I'm getting this error if I select the decimal type (all other types work)
Screen Shot 2020-12-11 at 2 50 11 PM

### DIFF
--- a/go/libraries/doltcore/schema/typeinfo/decimal.go
+++ b/go/libraries/doltcore/schema/typeinfo/decimal.go
@@ -37,6 +37,8 @@ type decimalType struct {
 
 var _ TypeInfo = (*decimalType)(nil)
 
+var DecimalDefaultType = &decimalType{sql.MustCreateDecimalType(10, 5)}
+
 func CreateDecimalTypeFromParams(params map[string]string) (TypeInfo, error) {
 	var precision uint8
 	if precisionStr, ok := params[decimalTypeParam_Precision]; ok {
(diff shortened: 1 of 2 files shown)


reltuk (approved):
This looks totally reasonable to me.

go/libraries/doltcore/schema/typeinfo/decimal.go
```diff
@@ -37,6 +37,8 @@ type decimalType struct {
 
 var _ TypeInfo = (*decimalType)(nil)
 
+var DecimalDefaultType = &decimalType{sql.MustCreateDecimalType(10, 5)}
```
Hydrocharged, line 40: I'm not sure that we should declare this as the "default type", as I don't really think that there should be one for the decimal type. Any place that wants to construct a decimal type as a default value should declare it at the call site IMO.

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -289,6 +289,8 @@ func FromKind(kind types.NomsKind) TypeInfo {
 		return Uint64Type
 	case types.UUIDKind:
 		return UuidType
+	case types.DecimalKind:
+		return DecimalDefaultType
```
Hydrocharged, line 293: I'd return `&decimalType{sql.MustCreateDecimalType(65, 30)}` here as it's far wider than `10,5`

Hydrocharged (changes requested):

go/libraries/doltcore/schema/typeinfo/typeinfo.go
tbantle22: Fixed https://github.com/dolthub/dolt/pull/1109/commits/290a79cc09c9da580ecd206528c9d8aae4095302

PR #1110, opened 2020-12-11
covering indexes

bheni:


go/libraries/doltcore/sqle/index_lookup.go
```diff
@@ -130,13 +130,40 @@ func (il *doltIndexLookup) Union(indexLookups ...sql.IndexLookup) (sql.IndexLook
...
+
+func (il *doltIndexLookup) indexCoversCols(cols []string) bool {
+	if cols == nil {
+		return false
+	}
+
+	idxCols := il.idx.IndexSchema().GetAllCols()
+	covers := true
+	for _, colName := range cols {
+		if _, ok := idxCols.GetByName(colName); !ok {
```
zachmu, line 160: Remember the columns coming back from the engine aren't guaranteed to match case

zachmu (commented):
Basic idea seems sound

PR #1111, opened 2020-12-12
README quotes changed bugfix for windows terminal

VyasSathya:
On the README there are instructions on how to add values into a table. The values in one portion have single quotes on the outside and have double-quotes for any string. While that format works in a Unix terminal, it doesn't work in Powershell or command prompt. 

The results before and after the fix are here

BugInReadMe
FixedREADMECommand

A few notes:
1) This issue only exists if the -q query is used, if dolt SQL is called by itself and the SQL query is called separately there is no problem. 
2) I tried to fix the code to see if there was a problem parsing it, but as I went up I found that the issue starts at main. If the quotes are single outside double inside then it will remove the quotes when it comes into args.


bheni (approved):
lgtm

PR #1112, opened 2020-12-14
Add cached functionality to dolt diff

VyasSathya:
https://github.com/dolthub/dolt/issues/1049

dolt diff vs dolt diff --cached:
FixedDoltDiffCached


go/cmd/dolt/commands/diff.go
```diff
@@ -281,41 +285,52 @@ func parseDiffArgs(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
...
+	stagedRoot, err := dEnv.StagedRoot(ctx)
 	workingRoot, err := dEnv.WorkingRootWithDocs(ctx)
+	headRoot, err := dEnv.HeadRoot(ctx)
 	if err != nil {
 		return nil, nil, nil, err
 	}
 
+	// default unstaged roots
+	fromRoot := stagedRoot
+	toRoot := workingRoot
```
bheni, line 298: It's unnecessary to introduce the variables fromRoot and toRoot.  You can just use:

``` go
from = stagedRoot
to = workingRoot
```

this will mean that all the places below where you do `to = toRoot` or `from = fromRoot` won't be needed.

bheni (changes requested):

VyasSathya:
@bheni 
Sorry Brian, just realized I never tagged you on the change. Just wanted to update you. I added the cached checks inside the loops to prevent creating another variable!

timsehn:
@bheni can we get another review? I could have used `--cached` a couple times reviewing bounties...

PR #1113, opened 2020-12-14
Added TRUNCATE and removed old optimized DELETE path

Hydrocharged:


zachmu (approved):
LG

timsehn:
Can we get this in @Hydrocharged ?

Hydrocharged:
I need to address the feedback in the `go-mysql-server` PR first, which I’ll do after `INSERT ON DUPLICATE KEY UPDATE`. It’ll all be done over the weekend.

PR #1117, opened 2020-12-15
Remove root methods from rsr and rsw

VinaiRachakonda:
First step in cleaning up RepoStatReader/Writer interface.


reltuk (approved):
LGTM!

PR #1120, opened 2020-12-16
Vinai/docs read writer

VinaiRachakonda:
This pr creates a DocsReadWriter which factors out some of the additional docs methods that were stuck in repo_state.json. Subsequent refactoring across files to account for this change.


reltuk (approved):
LGTM!

PR #1123, opened 2020-12-17
INSERT ON DUPLICATE KEY UPDATE

Hydrocharged:
Requires changes in `go-mysql-server`, but want to get this to you so you can look at the strategy here. Basically we're moving pk duplicate checks from the background thread to every INSERT query (so we can know when we have duplicates). I'm taking advantage of the idea that the row data map + collection of added/removed keys is equivalent to a row data map that has been through the edit accumulator.

Believe it or not, this actually seems to perform 4-5% **faster** on my completely unofficial benchmarks. I guess the background thread was running into a small IO bottleneck. If this seems like a decent-ish approach, I'm planning on the whole sysbench routine to see how it fairs there too.


Hydrocharged:
Also, obviously no tests. This is strictly looking at the general strategy.

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -276,14 +334,27 @@ func (te *pkTableEditor) InsertRow(ctx context.Context, dRow row.Row) error {
...
 			return err
 		}
-		return fmt.Errorf(ErrDuplicatePrimaryKeyFmt, keyStr)
+		return sql.ErrUniqueKeyViolation.New(keyStr)
+	}
+	// Check if the key is in the existing data, and is not scheduled to be removed.
+	if _, ok := te.tea.removedKeys[keyHash]; !ok {
+		pkExists, err := te.rowData.Has(ctx, keyHash, key)
+		if err != nil {
+			return errhand.BuildDError("failed to read table").AddCause(err).Build()
```
zachmu, line 349: Only simple errors in this package.

go/libraries/utils/async/action_executor.go
```diff
@@ -70,6 +70,13 @@ type work struct {
 	wg  *sync.WaitGroup
 }
 
+// ClearError removes the cached error if one is present.
+func (aq *ActionExecutor) ClearError() {
```
zachmu, line 74: Unused

go/libraries/utils/async/action_executor.go
```diff
@@ -94,8 +101,10 @@ func (aq *ActionExecutor) Execute(val interface{}) {
 
 // WaitForEmpty waits until all the work that has been submitted before
 // the call to |WaitForEmpty| has completed. It returns any errors that
-// any actions may have encountered.
-func (aq *ActionExecutor) WaitForEmpty() error {
+// any actions may have encountered. If retainError is true, then the
+// returned error is not cleared, meaning repeat calls to WaitForEmpty
+// will result in the same error being returned.
+func (aq *ActionExecutor) WaitForEmpty(retainError bool) error {
```
zachmu, line 107: The added param and semantics strike me as kind of sloppy. AFAICT the only use for `true` could be avoided by making pkTableEditor.flush() return an error (which seems like a good idea anyway).

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -97,6 +99,18 @@ type tableEditAccumulator struct {
...
 
+// currentTableData represents the current persisted state of the table's row map. A row map with the set of added and
+// removed keys is equivalent to a row map that has had those keys applied through the edit accumulator. As a result,
+// queries can make use of the most up-to-date version of the data for the sake of comparisons (such as to find
+// duplicate keys on insert rather than flush) while the edit accumulator rebuilds the map in the background. All
+// functions assume that, at most, only one edit accumulator is flushing at any given time.
+type currentTableData struct {
+	mutex       *sync.RWMutex
+	rowData     types.Map
+	addedKeys   map[hash.Hash]types.Value
```
zachmu, line 110: Having these fields defined both here and in the tableEditAccumulator seems kind of fragile and easy to mess up. It's not always obvious to me when these need to be copied from `tea` and when they need to be replaced with empty maps.

Some ideas to make this clearer:

1) Add rowData to tableEditAccumulator, move all the currentTableData methods onto that struct
2) Make tableEditAccumulator responsible for creating / updating currentTableData objects

zachmu (commented):
I like the idea, but specific details seem a little sloppy. I have some thoughts on how to improve the situation, but not exhaustive.

go/libraries/doltcore/table/editor/pk_table_editor.go
Hydrocharged: This was actually already in this package. We actually use it quite a bit in this same file in other places. I can remove those though.

go/libraries/utils/async/action_executor.go
Hydrocharged: I added it as I wanted a clean way to remove errors when someone used the modified `WaitForEmpty`, but I guess I don't need it anymore.

go/libraries/doltcore/table/editor/pk_table_editor.go
Hydrocharged: Thinking it through, it does make more sense for `rowData` map to be in the `tea`. Solves some of my other synchronization problems as well.

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -152,6 +166,45 @@ func newTableEditAcc(nbf *types.NomsBinFormat) *tableEditAccumulator {
 	}
 }
 
+// CreateFromCurrent returns a new tableEditAccumulator that references the current tableEditAccumulator.
+func (tea *tableEditAccumulator) CreateFromCurrent() *tableEditAccumulator {
+	return &tableEditAccumulator{
+		prevTea:      tea,
+		prevTeaAsMap: types.EmptyMap,
```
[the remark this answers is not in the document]
zachmu: Reading a bit farther, it seems like it's just a bad name.

I would just call this rowData. The field doc already mentions when it's set to nil or not.

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -152,6 +166,45 @@ func newTableEditAcc(nbf *types.NomsBinFormat) *tableEditAccumulator {
...
+
+// Has returns whether the current tableEditAccumulator contains the given key. This assumes that the given hash is for the given
+// key.
+func (tea *tableEditAccumulator) Has(ctx context.Context, keyHash hash.Hash, key types.Value) (bool, error) {
+	if _, ok := tea.addedKeys[keyHash]; ok {
+		return true, nil
+	}
+	if _, ok := tea.removedKeys[keyHash]; !ok {
+		// When prevTeaAsMap is updated, prevTea is set to nil. Therefore, if prevTea is non-nil, we use it.
+		if tea.prevTea != nil {
```
zachmu, line 191: You need a lock here to avoid a race with prevTea and prevTeaAsMap being updated.

go/libraries/utils/async/action_executor.go
```diff
@@ -94,7 +94,9 @@ func (aq *ActionExecutor) Execute(val interface{}) {
 
 // WaitForEmpty waits until all the work that has been submitted before
 // the call to |WaitForEmpty| has completed. It returns any errors that
-// any actions may have encountered.
+// any actions may have encountered. If retainError is true, then the
```
zachmu, line 97: Comment now out of date

zachmu (commented):
Looks good, much better than the first draft

go/libraries/doltcore/table/editor/pk_table_editor.go
Hydrocharged: There are locks around all of the calls and modifications, they're just done from outside of the struct methods. I'll add a note.

Hydrocharged:
@zachmu See the attached issue. Not a blocker (just added the order clause), but at least worth a read.

go/libraries/doltcore/sqle/testutil.go
```diff
@@ -165,6 +165,7 @@ func drainIter(iter sql.RowIter) error {
 		if err == io.EOF {
 			break
 		} else if err != nil {
+			_ = iter.Close()
```
zachmu, line 168: Probably prefer to just defer this at the beginning of the function

zachmu: And actually, we aren't capturing the error from Close(), if any, which is a serious bug that has caused problems in the past. Definitely needs to be fixed.

zachmu (approved):
LGTM

go/libraries/doltcore/sqle/testutil.go
Hydrocharged: This is only used in tests, we can just panic here instead.

PR #1124, opened 2020-12-17
/go/libraries/doltcore/diff: Keyless Table Diff

andrew-wm-arthur:


go/cmd/dolt/commands/diff.go
```diff
@@ -1009,26 +1029,23 @@ func diffSummary(ctx context.Context, from types.Map, to types.Map, colLen int)
...
+	} else if keyless {
+		formatKeylessSummary(acc)
+	} else {
+		formatSummary(acc, colLen)
 	}
 
 	return nil
 }
 
 func formatSummary(acc diff.DiffSummaryProgress, colLen int) {
```
zachmu, line 1048: These should be called print*, not format*

go/libraries/doltcore/diff/async_differ.go
```diff
@@ -16,17 +16,41 @@ package diff
...
+	GetDiffs(numDiffs int, timeout time.Duration) ([]*diff.Difference, bool, error)
+
+	// Close closes the RowDiffer.
+	Close() error
+}
+
+func NewRowDiffer(ctx context.Context, fromSch, toSch schema.Schema, buf int) RowDiffer {
+	ad := NewAsyncDiffer(buf)
+
+	// assumes no PK changes
```
zachmu, line 45: Expand on this comment a bit. What happens if one schema is keyless and one not?

go/libraries/doltcore/diff/async_differ.go
```diff
@@ -91,13 +118,97 @@ func (ad *AsyncDiffer) GetDiffs(numDiffs int, timeout time.Duration) ([]*diff.Di
...
+		if delta > 0 {
+			df.ChangeType = types.DiffChangeAdded
+			df.OldValue = nil
+			return df, uint64(delta), nil
+		} else if delta < 0 {
+			df.ChangeType = types.DiffChangeRemoved
+			df.NewValue = nil
+			return df, uint64(-delta), nil
+		} else {
+			return df, 0, fmt.Errorf("diff with delta = 0 for key: %s", df.KeyValue.HumanReadableString())
```
zachmu, line 209: Should this be a panic? Should be impossible, right?

go/libraries/doltcore/diff/async_differ.go
```diff
@@ -91,13 +118,97 @@ func (ad *AsyncDiffer) GetDiffs(numDiffs int, timeout time.Duration) ([]*diff.Di
...
+
+		case <-kd.egCtx.Done():
+			return nil, false, kd.eg.Wait()
+
+		case d, more = <-kd.diffChan:
+			if !more {
+				return diffs[:idx], more, nil
+			}
+
+			kd.df, kd.copiesLeft, err = convertDiff(d)
```
zachmu, line 161: Bet you're glad we talked you into using compatible noms storage now huh

go/libraries/doltcore/diff/diff_summary.go
```diff
@@ -86,3 +120,40 @@ func reportChanges(ctx context.Context, change *diff.Difference, ch chan<- DiffS
...
+	delta := int64(newCard) - int64(oldCard)
+	if delta > 0 {
+		summary = DiffSummaryProgress{Adds: uint64(delta)}
+	} else if delta < 0 {
+		summary = DiffSummaryProgress{Removes: uint64(-delta)}
+	} else {
+		return fmt.Errorf("diff with delta = 0 for key: %s", change.KeyValue.HumanReadableString())
+	}
+
+	select {
```
zachmu, line 153: Is this idiomatic / kosher? Never seen sending and receiving on channels in the same select block before

(for that matter I don't think I've ever seen sending in a select, at all)

zachmu (approved):
Looks great, just a few comments

go/libraries/doltcore/diff/diff_summary.go
andrew-wm-arthur: This was copied from AsyncDiffer, don't @ me

go/libraries/doltcore/diff/async_differ.go
andrew-wm-arthur: Coulda worked either way, consumers of `GetDiffs()` use `row.FromNoms` and then work with `row.Row`s

go/libraries/doltcore/diff/async_differ.go
andrew-wm-arthur: Should be, yes...

PR #1127, opened 2020-12-17
Quick add of --no-fk on merge for CLI merges

Hydrocharged:
Quick workaround for https://github.com/dolthub/dolt/issues/1126. Merge with `--no-fk`, and you can verify that the constraints are still satisfied afterward with `dolt verify-constraints table1 table2...`


Hydrocharged:
Bad PR, mistakes were made

PR #1129, opened 2020-12-18
DOLT_ADD SQL Implementation

VinaiRachakonda:
Implement DOLT_ADD


go/cmd/dolt/commands/branch.go
```diff
@@ -141,7 +141,7 @@ func printBranches(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
 
 	verbose := apr.Contains(verboseFlag)
 	printRemote := apr.Contains(remoteParam)
-	printAll := apr.Contains(allParam)
```
VinaiRachakonda, line 144: Need to double check this change is correct.

bats/sql-add.bats
```diff
@@ -0,0 +1,74 @@
...
+
+    dolt sql <<SQL
+CREATE TABLE test (
+    pk int primary key
+);
+CREATE TABLE test2 (
+    pk int primary key
+);
+INSERT INTO test VALUES (0),(1),(2);
+SQL
```
zachmu, line 15: Why two different SQL blocks?

bats/sql-add.bats
```diff
@@ -0,0 +1,74 @@
...
+    dolt sql <<SQL
```
zachmu, line 16: This second block doesn't seem to be doing anything useful. Just create the table with the data you want

bats/sql-add.bats
```diff
@@ -0,0 +1,74 @@
...
+DELETE FROM test WHERE pk = 0;
+INSERT INTO test VALUES (3);
+SQL
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "DOLT_ADD all flag works" {
```
zachmu, line 26: Duplicate this test with "." as well

bats/sql-add.bats
```diff
@@ -0,0 +1,74 @@
...
+    run dolt diff
+    [ "$status" -eq 0 ]
+    [ "$output" = "" ]
+
+    run dolt log
+    [ $status -eq 0 ]
+    [[ "$output" =~ "Commit1" ]] || false
+    regex='Bats Tests <bats@email.fake>'
+    [[ "$output" =~ "$regex" ]] || false
+}
```
zachmu, line 74: You should have a test that makes the handling of README.md files explicit. I think we decided that it doesn't make sense to deal with the local file system from SQL? So have a test that writes a README.md file, does SELECT  dolt_add(".") and commits, and confirms that README.md still hasn't been committed.

go/libraries/doltcore/env/actions/staged.go
```diff
@@ -216,28 +221,19 @@ func getStagedAndWorking(ctx context.Context, dEnv *env.DoltEnv) (*doltdb.RootVa
...
-
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return roots[WorkingRoot], roots[HeadRoot], nil
-}
-
-func getRoots(ctx context.Context, dEnv *env.DoltEnv, rootTypes ...RootType) (map[RootType]*doltdb.RootValue, error) {
+// Get all wanted roots.
```
zachmu, line 224: In golang always begin method doc with the name of the func, and use simple present tense verbs. So, "getRoots gets..."

go/libraries/doltcore/env/environment.go
```diff
@@ -461,6 +461,10 @@ func (d *docsReadWriter) ResetWorkingDocsToStagedDocs(ctx context.Context) error
 	return d.dEnv.ResetWorkingDocsToStagedDocs(ctx)
 }
 
+func (d *docsReadWriter) GetOneDocDetail(docName string) (doc doltdb.DocDetails, err error) {
```
zachmu, line 464: GetDocDetail, and add method doc

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+const DoltAddFuncName = "dolt_add"
+
+type DoltAddFunc struct {
+	children []sql.Expression
+}
+
+func (d DoltAddFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
+	dbName := ctx.GetCurrentDatabase()
```
zachmu, line 35: This can be empty. The error message below isn't great in that case.

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	apr := cli.ParseArgs(ap, args, nil)
+	allFlag := apr.Contains(cli.AllFlag)
+
+	if apr.NArg() == 0 && !allFlag {
+		return nil, fmt.Errorf("Nothing specified, nothing added.\n Maybe you wanted to say 'dolt add .'?")
```
zachmu, line 54: No newlines in error messages

zachmu, line 54: No newlines in error messages

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
```diff
@@ -0,0 +1,106 @@
...
+
+	for i, child := range d.children {
+		childrenStrings[i] = child.String()
+	}
+
+	return fmt.Sprintf("DOLT_ADD(%s)", strings.Join(childrenStrings, ","))
+}
+
+func (d DoltAddFunc) Type() sql.Type {
+	return sql.Text
```
zachmu, line 88: Seems like this is better off returning an integer value -- the only string it ever returns is empty.

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return "", nil
+}
+
+func (d DoltAddFunc) Resolved() bool {
+	for _, child := range d.Children() {
```
zachmu, line 69: Next one of these you do, add an NaryFunction type to go-mysql-server that defines an array of children and handles all these boilerplate methods. Then you can just embed that type everywhere you need it.

zachmu (commented):
Sorry for the slow turnaround here, bug me harder next time :)

Looks pretty good, just a few comments I want to see taken care of.

bats/sql-add.bats
VinaiRachakonda: To test addition with different table name

bats/sql-add.bats
VinaiRachakonda: Right now dolt add works with docs. Should I remove that behavior?

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
VinaiRachakonda: Good idea, that'll also introduce me to go-mysql

bats/sql-add.bats
zachmu: I mean that you had two `dolt sql <<SQL` commands, when one would do. You've fixed that now.

bats/sql-add.bats
zachmu: I think it's fine as is, we can change later.

go/libraries/doltcore/env/actions/staged.go
```diff
@@ -216,28 +221,19 @@ func getStagedAndWorking(ctx context.Context, dEnv *env.DoltEnv) (*doltdb.RootVa
...
-
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return roots[WorkingRoot], roots[HeadRoot], nil
-}
-
-func getRoots(ctx context.Context, dEnv *env.DoltEnv, rootTypes ...RootType) (map[RootType]*doltdb.RootValue, error) {
+// getRoots gets all wanted roots.
```
zachmu, line 224: Comment needs expanding. What does "wanted" mean? How does it relate to the params?

go/libraries/doltcore/sqle/dfunctions/dolt_add.go
zachmu: Not actually fixed. Need specific handling for the empty case.

zachmu (approved):
LG, just a couple comments.

PR #1132, opened 2020-12-20
Added release automation for Dolt

oscarbatori:
The Release workflow is kicked off by the user hitting the `run workflow` GUI, and then entering version.

The basic workflow is as follows:
- bump-version
  - checks out the code, and creates a branch, for example `v0.23.0-release` if the release parameter is `0.23.0`
  - updates the version string
  - snaps a commit
  - creates a tag
  - pushes the tag and branch
- create-release
  - checks out the code at the newly created tag
  - creates a release
  - builds the binaries
  - uploads the binaries to the release
- homebrew-bump
  - creates a PR to bump the Homebrew formula

Possible enhancements
- validate version string passed by user
- use GitHub API to create PR of release branch back to master
- notifications
- MSI creation


.github/workflows/cd-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+          git config --global user.email dolt-release-bot@dolthub.com
+          ./go/utils/publishrelease/bump-version.sh ${{ github.event.inputs.version }}
+
+  create-release:
+    needs: bump-version
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
coffeegoddd, line 30: does this work? may have a race between pushed ref and getting it here

.github/workflows/cd-release.yaml
```diff
@@ -0,0 +1,103 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: go/out/install.sh
+          asset_name: install.sh
+          asset_content_type: text/plain
+
+  homebrew-bump:
+    needs: bump-version
```
coffeegoddd, line 90: this is fine to do with the `create-release` step finishing first?

coffeegoddd (approved):
Seems good, though i'd say the final test is having a team member use it. Also i think we should add the MSI step if we can.

.github/workflows/cd-release.yaml
oscarbatori: Good point, it does actually use that. I will update the dependency chain.

PR #1133, opened 2020-12-20
Andy/keyless import/export

andrew-wm-arthur:
Adds keyless table support for `dolt table import ...`. `table import -c` now creates keyless tables if the `-pk` option is not provided.


go/libraries/doltcore/row/keyless_row.go
```diff
@@ -68,6 +68,26 @@ func KeylessRowsFromTuples(key, val types.Tuple) (Row, uint64, error) {
 	return r, card, err
 }
 
+func keylessRowFromTaggedValued(nbf *types.NomsBinFormat, sch schema.Schema, tv TaggedValues) (Row, error) {
```
zachmu, line 71: values

zachmu (approved):
LG

PR #1134, opened 2020-12-21
Andy/ungate keyless tables

andrew-wm-arthur:


go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -66,11 +66,6 @@ func TestCreateTable(t *testing.T) {
 			expectedTable: "testTable",
 			expectedErr:   "syntax error",
 		},
-		{
```
zachmu, line 69: You should add some smoke tests for keyless tables in here

zachmu (approved):
LG

PR #1139, opened 2020-12-21
index bug fix

bheni:


bheni:
there are new go-mysql-server tests that test this.  They fail prior to this change, and pass now.

andrew-wm-arthur (approved):
LGTM!

PR #1140, opened 2020-12-21
Fix hanging on duplicate rows within large csvs

VinaiRachakonda:


go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -283,7 +285,7 @@ func (te *pkTableEditor) InsertRow(ctx context.Context, dRow row.Row) error {
 		if err != nil {
 			return err
 		}
-		return fmt.Errorf(ErrDuplicatePrimaryKeyFmt, keyStr)
+		return table.NewBadRow(dRow, fmt.Sprintf(ErrDuplicatePrimaryKeyFmt, keyStr))
```
zachmu, line 288: This doesn't work with Daylon's outstanding change to handle duplicate primary keys here. That change requires this method to return an error type defined in the go-mysql-server package, which means you can't wrap it like this.

Change the logic that examines this error, rather than changing the error type.

timsehn:
Can this be covered by a bats test?

VinaiRachakonda:
@timsehn I honestly don't know what. The hanging problem on `p.StopWithErr(err)` is only reproducible with large imports. @andrew-wm-arthur Seems to think it happens with the batching threshold set in pkTableEditor. This is pretty a quick and dirty fix that address that `p.StopWithErr(err)` did not work appropriately.

timsehn:
KK. I'm good then. Large import would really make the bats tests slow.

bats/import-update-tables.bats
```diff
@@ -20,6 +20,12 @@ SQL
 pk,c1,c2,c3,c4,c5
 0,1,2,3,4,5
 1,1,2,3,4,5
+DELIM
+
+    cat <<DELIM > 1pk5col-rpt-ints.csv
```
zachmu, line 25: Do this setup work only in the one test you need it

go/libraries/doltcore/mvdata/table_data_loc.go
```diff
@@ -239,9 +239,13 @@ func (te *tableEditorWriteCloser) WriteRow(ctx context.Context, r row.Row) error
 	_ = atomic.AddInt64(&te.gcOps, 1)
 
 	if te.insertOnly {
-		_ = atomic.AddInt64(&te.statOps, 1)
-		te.stats.Additions++
-		return te.tableEditor.InsertRow(ctx, r)
+		if err := te.tableEditor.InsertRow(ctx, r); err != nil {
```
zachmu, line 242: Don't like this if / else style for error handling. It's not actually any more compact and is difficult to scan. Just put if err != nil on its own line.

zachmu: Same comment elsewhere in this file.

zachmu (approved):
LG, just a few comments.

PR #1143, opened 2020-12-22
First cut of Discrod notifications

oscarbatori:
This implements the following policy:
- notify on cancellation or failure of any job
- notify on release, including success

Currently release notifications are broken by a shortcoming in GitHub Actions, namely that one workflow cannot kick of another when using `GITHUB_TOKEN`. We will devise a workaround.


coffeegoddd (approved):
LGTM!

PR #1145, opened 2020-12-22
optimizations

bheni:


go/libraries/doltcore/sqle/tables.go
```diff
@@ -432,6 +439,15 @@ func (t *DoltTable) GetForeignKeys(ctx *sql.Context) ([]sql.ForeignKeyConstraint
...
 
+/*func (t *DoltTable) WithProjection(colNames []string) sql.Table {
+	t.projectedCols = colNames
+	return t
+}
+
+func (t *DoltTable) Projection() []string {
+	return t.projectedCols
+}*/
+
```
bheni, line 450: Disabled until I have a chance to look into an issue with colNames being off in some cases.

bheni:
Modest improvements just with the iterator changes:

`~/datasets/wikipedia-ngrams>GOMAXPROCS=1 time dolt --prof cpu sql -r null -q 'SELECT * from bigram_counts'`

Before: `404.65 real       372.61 user        12.59 sys`
After: `379.31 real       315.42 user        19.45 sys`

[after.pdf](https://github.com/dolthub/dolt/files/5748808/after.pdf)
[before.pdf](https://github.com/dolthub/dolt/files/5748809/before.pdf)

go/libraries/doltcore/sqle/dolt_map_iter.go
```diff
@@ -68,79 +107,83 @@ func (conv *KVToSqlRowConverter) ConvertKVToSqlRow(k, v types.Value) (sql.Row, e
 	}
 
 	cols := make([]interface{}, conv.rowSize)
-	filled, err := conv.processTuple(cols, 0, keyTup)
-	if err != nil {
-		return nil, err
+	if conv.valsFromKey > 0 {
+		// keys are not in sorted order so cannot use max tag to early exit
```
andrew-wm-arthur, line 111: For PK tables, key tuples _are_ sorted by tags and val tuples have the same order as columns in the `schema.Schema`.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -42,7 +42,7 @@ func limitTestQueriesTo(queries ...string) {
 }
 
 func TestQueries(t *testing.T) {
-	// limitTestQueriesTo(...) // whitelist queries you want run.
+	limitTestQueriesTo("SELECT i FROM niltable WHERE i2 IS NOT NULL ORDER BY 1") // whitelist queries you want run.
```
andrew-wm-arthur, line 45: this helper function is hiding test failures in the engine tests

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
```diff
@@ -114,7 +114,7 @@ func (d *DoltHarness) SupportsForeignKeys() bool {
 }
 
 func (d *DoltHarness) SupportsKeylessTables() bool {
-	return true
+	return false
```
andrew-wm-arthur, line 117: This will also hide test failures

andrew-wm-arthur (changes requested):
I'm seeing a handful of test failures when I run the full engine test suite

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
bheni: shit. didn't mean to check this in.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -86,6 +85,20 @@ func TestQueryPlans(t *testing.T) {
...
+		"SELECT a.pk1,a.pk2,b.pk1,b.pk2 FROM two_pk a, two_pk b WHERE a.pk1=b.pk2 AND a.pk2=b.pk1 ORDER BY 1,2,3",
+		"SELECT opk.c5,pk1,pk2 FROM one_pk opk, two_pk tpk WHERE pk=pk1 ORDER BY 1,2,3",
+		"SELECT one_pk.c5,pk1,pk2 FROM one_pk,two_pk WHERE pk=pk1 ORDER BY 1,2,3",
+		"SELECT pk,i,f FROM one_pk RIGHT JOIN niltable ON pk=i and pk > 0 ORDER BY 2,3",
+		"SELECT pk,pk1,pk2 FROM one_pk JOIN two_pk ON pk1-pk>0 AND pk2<1",
+		"SELECT pk,pk1,pk2 FROM one_pk JOIN two_pk ORDER BY 1,2,3",
+		"SELECT pk,pk1,pk2 FROM one_pk LEFT JOIN two_pk ON pk=pk1 ORDER BY 1,2,3",
+		"SELECT pk,pk1,pk2 FROM one_pk,two_pk WHERE one_pk.c1=two_pk.c1 ORDER BY 1,2,3",
+		"SELECT pk,pk1,pk2,one_pk.c1 AS foo, two_pk.c1 AS bar FROM one_pk JOIN two_pk ON one_pk.c1=two_pk.c1 ORDER BY 1,2,3",
+		"SELECT pk,pk1,pk2,one_pk.c1 AS foo,two_pk.c1 AS bar FROM one_pk JOIN two_pk ON one_pk.c1=two_pk.c1 WHERE one_pk.c1=10",
```
bheni, line 101: @zachmu had to add some new ones here now that projected tables are working.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -482,6 +505,43 @@ func (p doltTablePartition) Key() []byte {
...
+}
+
+type partitionIter struct {
+	pos  uint64
+	end  uint64
+	iter types.MapIterator
+}
+
+func newPartitionIter(ctx context.Context, m types.Map, start, end uint64) (*partitionIter, error) {
+	iter, err := m.IteratorAt(ctx, start)
```
andrew-wm-arthur, line 522: From what I can tell, all of the perf wins are coming here.

andrew-wm-arthur (commented):
See my comment in `tables.go:521`
By using `IteratorAt()` instead of `BufferedIteratorAt()` we're creating almost no garbage. The buffered version uses `GetCompositeChildSequence()` which seems to be screwing up the escape analysis or causing garbage creation in some other way. If we can remedy this we should get _even more_ wins by doing larger disk reads while also saving on GC overhead.

go/libraries/doltcore/sqle/tables.go
bheni: It really is dependent on the test query you are running with.  Projected tables have big benefits for some queries. Some schemas perform much better with the new iterators.  There are multiple wins in this PR.

go/libraries/doltcore/sqle/dolt_map_iter.go
```diff
@@ -68,79 +107,83 @@ func (conv *KVToSqlRowConverter) ConvertKVToSqlRow(k, v types.Value) (sql.Row, e
...
-type DoltMapIter struct {
-	kvGet KVGetFunc
-	conv  *KVToSqlRowConverter
-}
-
-// NewDoltMapIterFromNomsMapItr returns an iterator which returns sql.Row instances read from a types.Map.  The cols
-// passed in are used to limit the values that are processed
-func NewDoltMapIterFromNomsMapItr(mapItr types.MapIterator, cols []schema.Column) *DoltMapIter {
-	getFunc := func(ctx context.Context) (types.Value, types.Value, error) {
+func GetGetFuncForMapIter(mapItr types.MapIterator) func(ctx context.Context) (types.Value, types.Value, error) {
```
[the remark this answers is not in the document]
bheni: not sure what you mean.  Do you just mean in general we should change the noms map iterators to return io.EOF instead of using a nil key to signify all items have been iterated over?

andrew-wm-arthur (approved):

PR #1149, opened 2020-12-23
/go/libraries/doltcore/sqle: move table cache to DoltSession, purge on root change

andrew-wm-arthur:
I'm ambivalent about where the table cache lives, but we need to have access to it when we change the root of a `sqle.Database`

This PR purges the table cache when we change roots. IE `SET @@dolt_head = hashof(...)`. The intention is to limit the scope of edit sessions, and table mutations in general, and contain them to the working root of a `sqle.Database`. 

Currently there is a tricky reference chain within the table cache: `sqle.tableCache` -> `sqle.WritableDoltTable` -> `sqle.sqlTableEditor` -> `table/editor.TableEditSesson`

When we change roots in the database we also change the root within the `TableEditSession`. They're both referencing the new working root. Without purging the cache, we will keep old tables from the previous working root that still have a reference to the `TableEditSession`. 

This hasn't caused any issues (yet), but I think it's prudent to limit the scope and lifetime of these interconnected pieces of state. The next step is to create a fresh `TableEditSession` each time we switch roots.

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_session.go
+++ b/go/libraries/doltcore/sqle/dolt_session.go
@@ -17,6 +17,7 @@ package sqle
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/dolthub/go-mysql-server/sql"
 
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -39,18 +40,36 @@ type DoltSession struct {
...
+	Get(tableName string, root *doltdb.RootValue) (sql.Table, bool)
+
+	// Put stores a copy of |tbl| corresponding to |root|.
+	Put(tableName string, root *doltdb.RootValue, tbl sql.Table)
+
+	// AllForRoot retrieves all tables from the caches corresponding to |root|.
+	AllForRoot(root *doltdb.RootValue) (map[string]sql.Table, bool)
+
+	// Purge removes all entries from the cache.
+	Purge()
```
zachmu, line 62: Probably prefer Clear

zachmu (approved):
LG

PR #1151, opened 2020-12-26
Implement dolt status table

VinaiRachakonda:
Implement the dolt_status table. Schema is of the form table_name, staged(bool), status,

### DIFF
--- a/bats/sql-status.bats
+++ b/bats/sql-status.bats
@@ -4,10 +4,16 @@ load $BATS_TEST_DIRNAME/helper/common.bash
 setup() {
     setup_common
 
-    dolt sql <<SQL
-CREATE TABLE test (
-    pk int primary key
-);
+dolt sql <<SQL
+    CREATE TABLE test (
+      pk BIGINT NOT NULL,
+      c1 BIGINT,
+      c2 BIGINT,
+      c3 BIGINT,
+      c4 BIGINT,
+      c5 BIGINT,
+      PRIMARY KEY (pk)
+    );
 SQL
 }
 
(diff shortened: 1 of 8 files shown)


bats/sql-status.bats
```diff
@@ -0,0 +1,108 @@
...
+      PRIMARY KEY (pk)
+    );
+SQL
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "status properly works with working and staged tables" {
```
zachmu, line 24: Need one more test case: a table that has both staged and unstaged changes.

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -0,0 +1,250 @@
...
+
+// BranchItr is a sql.RowItr implementation which iterates over each commit as if it's a row in the table.
+type StatusItr struct {
+	tables   []string
+	isStaged []bool
+	statuses []string
+	idx      int
+}
+
+func NewStatusItr(ctx *sql.Context, st *StatusTable) (*StatusItr, error) {
```
zachmu, line 74: shouldn't be exported

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -0,0 +1,250 @@
...
+		return &StatusItr{}, err
+	}
+
+	tLength := len(stagedTables) + len(unstagedTables) + len(stagedDocDiffs.Docs) + len(notStagedDocDiffs.Docs) + len(workingTblsInConflict) + len(workingDocsInConflict.Docs)
+
+	tables := make([]string, tLength)
+	isStaged := make([]bool, tLength)
+	statuses := make([]string, tLength)
+
+	itr := StatusItr{tables: tables, isStaged: isStaged, statuses: statuses, idx: 0}
```
zachmu, line 109: Better to make the pointer here instead of everywhere it's used below

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		idx += 1
+	}
+
+	return idx
+}
+
+func handleUnstagedTables(notStaged []diff.TableDelta, itr *StatusItr, idx int) int {
+	for _, td := range notStaged {
```
zachmu, line 152: Lots of repeated code with the function above

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -0,0 +1,250 @@
...
+		itr.statuses[idx] = docDiffTypeToLabel[dType]
+
+		idx += 1
+	}
+
+	return idx
+}
+
+func handleUnstagedDocDiffs(notStaged *diff.DocDiffs, itr *StatusItr, idx int) int {
+	for _, docName := range notStaged.Docs {
```
zachmu, line 194: Same comment, these two functions are very similar

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -0,0 +1,250 @@
...
+func (s StatusTable) PartitionRows(context *sql.Context, _ sql.Partition) (sql.RowIter, error) {
+	return NewStatusItr(context, &s)
+}
+
+// NewStatusTable creates a StatusTable
+func NewStatusTable(_ *sql.Context, ddb *doltdb.DoltDB, rsr env.RepoStateReader, drw env.DocsReadWriter) sql.Table {
+	return &StatusTable{ddb: ddb, rsr: rsr, drw: drw}
+}
+
+// BranchItr is a sql.RowItr implementation which iterates over each commit as if it's a row in the table.
```
zachmu, line 66: copy pasta error

zachmu, line 66: sql.RowIter

zachmu (commented):
Looks like a good approach, just needs some cleanup. Biggest comment is that a lot of the code in the status table file could be refactored to eliminate duplication.

VinaiRachakonda:
@zachmu One overarching question I had about this is how status should work docs. Should I replicate what's going with `dolt_docs` or is copying what status.go does fine?

go/libraries/doltcore/sqle/dtables/status_table.go
VinaiRachakonda: Good catch.

zachmu (approved):
LG, thanks for the changes!

PR #1159, opened 2020-12-29
Bh/hang fix

bheni:
Fixes https://github.com/dolthub/dolt/issues/1153

and disables GC on import errors.

### DIFF
--- a/go/libraries/doltcore/mvdata/data_mover.go
+++ b/go/libraries/doltcore/mvdata/data_mover.go
@@ -90,14 +90,25 @@ func (dmce *DataMoverCreationError) String() string {
 	return string(dmce.ErrType) + ": " + dmce.Cause.Error()
 }
 
+type GCTableWriteCloser interface {
+	table.TableWriteCloser
+	GC(ctx context.Context) error
+}
+
 // Move is the method that executes the pipeline which will move data from the pipeline's source DataLocation to it's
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/mvdata/data_mover.go
```diff
@@ -90,15 +90,26 @@ func (dmce *DataMoverCreationError) String() string {
...
+
 // Move is the method that executes the pipeline which will move data from the pipeline's source DataLocation to it's
 // dest DataLocation.  It returns the number of bad rows encountered during import, and an error.
 func (imp *DataMover) Move(ctx context.Context) (badRowCount int64, err error) {
 	defer imp.Rd.Close(ctx)
 	defer func() {
 		closeErr := imp.Wr.Close(ctx)
-		if closeErr != nil {
+		if err == nil {
 			err = closeErr
```
bheni, line 105: I don't like overwriting an existing error if an error closing occurs.

Hydrocharged: Yeah first error is probably the most important

Hydrocharged (approved):
Oh snap there's now video support for reviews! LGTM!

VinaiRachakonda (commented):
If the changes are with cleaning up some channels, should we add a bats test.

bheni:
@VinaiRachakonda Yeah, adding the case from the bug as a bats test would be good.

PR #1162, opened 2020-12-31
unrolled decode varint decode loop

bheni:
30% faster on the benchmark in this PR.

BenchmarkUnrolledDecodeUVarint/binary.UVarint-8         	1000000000	         0.0372 ns/op
BenchmarkUnrolledDecodeUVarint/unrolled-8               	1000000000	         0.0258 ns/op

### DIFF
--- a/go/store/types/codec_test.go
+++ b/go/store/types/codec_test.go
@@ -22,9 +22,12 @@
 package types
 
 import (
+	"encoding/binary"
+	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCodecWriteFloat(t *testing.T) {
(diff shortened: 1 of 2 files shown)


go/store/types/codec_test.go
```diff
@@ -72,3 +75,122 @@ func TestCodecReadFloat(t *testing.T) {
...
+	const NumDecodes = 100000
+	masks := []uint64{
+		0xFF,
+		0xFFFF,
+		0xFFFFFF,
+		0xFFFFFFFF,
+		0xFFFFFFFFFF,
+		0xFFFFFFFFFFFF,
+		0xFFFFFFFFFFFFFF,
+		0xFFFFFFFFFFFFFFFF}
```
bheni, line 89: mask provides better distribution of encoded number sizes

reltuk:
Is there any difference in actual perf of something like a table scan? Not suggesting we shouldn’t do it, but very curious if it’s measurable outside of a tight microbenchmark loop.

PR #1166, opened 2020-12-31
Vinai/dolt reset

VinaiRachakonda:
This pr adds DOLT_RESET functionality.

Previous DOLT_RESET that works with session variables is renamed to RESET()


go/libraries/doltcore/env/repo_state.go
```diff
@@ -44,8 +45,13 @@ type RepoStateWriter interface {
 type DocsReadWriter interface {
 	GetAllValidDocDetails() ([]doltdb.DocDetails, error)
 	PutDocsToWorking(ctx context.Context, docDetails []doltdb.DocDetails) error
+	PutDocsToStaged(ctx context.Context, docDetails []doltdb.DocDetails) (*doltdb.RootValue, error)
 	ResetWorkingDocsToStagedDocs(ctx context.Context) error
 	GetDocDetail(docName string) (doc doltdb.DocDetails, err error)
+	UpdateFSDocsToRootDocs(ctx context.Context, root *doltdb.RootValue, docs Docs) error
+	GetDocsWithNewerTextFromRoot(ctx context.Context, root *doltdb.RootValue, docs Docs) (Docs, error)
+	GetFS() filesys.Filesys
+	GetDocs() Docs
```
zachmu, line 54: This seems to be redundant with GetAllValidDocDetails, but actually does something completely different. Yikes.

go/libraries/doltcore/env/repo_state.go
```diff
@@ -44,8 +45,13 @@ type RepoStateWriter interface {
 type DocsReadWriter interface {
```
zachmu, line 45: This interface is kind of a mess. Lots of methods that do very similar things. Some of them read from disk, some of them read from previous cached information, etc.

This interface should be about one thing: moving docs file details between memory and disk. And it does way way more than that.

My first pass for what this should look like:

GetDocsOnDisk([]string docs) Docs // reads and returns docs on disk, optionally filtered by name
WriteDocsToDisk(Docs) // writes the docs given to disk

Neither of these methods involve a file system, because any implementor needs a filesystem to do its job in general. And none of them involve root values.

If you want doc details that are in a RootValue, ask the root value.
If you want to move doc details that are on disk into a RootValue, get them from the above method then tell the RootValue to update its docs table (possibly creating it in the process).
If you want update the value of the staged or working roots, there are methods in RSW for doing that.

You didn't create this mess, but it's getting really out of hand at this point. You should clean it up.

You can come back and do this on another pass.

go/libraries/doltcore/env/actions/docs.go
```diff
@@ -58,17 +62,17 @@ func SaveDocsFromRoot(ctx context.Context, root *doltdb.RootValue, dEnv *env.Dol
...
 		return err
 	}
 
 	docs := removeUntrackedDocs(localDocs, docDiffs)
 
-	err = dEnv.UpdateFSDocsToRootDocs(ctx, targetRoot, docs)
+	err = drw.UpdateFSDocsToRootDocs(ctx, targetRoot, docs)
 	if err != nil {
-		localDocs.Save(dEnv.FS)
+		localDocs.Save(drw.GetFS())
```
zachmu, line 75: Definitely the wrong abstraction here... the docs read writer should know how to write to disk as necessary, not pass its filesystem to another object

zachmu (commented):
So the reset function itself is fine, the refactoring of reset command code is mostly fine, but there are two big issues.

1) Local filesystem stuff. What happens when somebody runs this on dolthub? Or in a webservices context? Somewhere where we don't have docs or a repo state file that exist as files in the filesystem. It should give a sane error back to the client about the operation not being supported. I'm not aware of any tests of that behavior, but we should have them. For this to be checked in, I want tests that verify that something reasonable happens in a context without a repo state / filesystem. Probably easiest to achieve in golang unit tests, but maybe you could figure something clever out with the command line as well, like using a filesystem with no permissions or something.

2) DocsReadWriter is just a mess. You didn't create the mess, but you're making it worse here. Cleaning this up can be a second pass, but for the time being you can't do things like introduce a Filesys dependency in this interface (breaks the webservices). I would rather ignore docs entirely from the `dolt_reset` sql path than ship what you have here. That's what I think you should ship for a first pass, then come back and get a better interface in place.

This is a lot of feedback so hit me up with any questions you have.

go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
```diff
@@ -0,0 +1,124 @@
...
+
+	working, staged, head, err := env.GetRoots(ctx, dbData.Ddb, dbData.Rsr)
+
+	if err != nil {
+		return 1, err
+	}
+
+	var verr errhand.VerboseError
+
+	if apr.ContainsAll(cli.HardResetParam, cli.SoftResetParam) {
```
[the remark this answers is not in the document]
VinaiRachakonda: Wdym by this?

go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if verr != nil {
+		return 1, fmt.Errorf(verr.Error())
+	}
+
+	return 0, nil
+}
+
+func (d DoltResetFunc) Resolved() bool {
+	for _, child := range d.Children() {
```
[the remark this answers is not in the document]
VinaiRachakonda: Lol forgot to export the param in the Nnary type

bats/sql-reset.bats
```diff
@@ -0,0 +1,134 @@
...
+    [[ "$output" =~ "Changes to be committed:" ]] || false
+    [[ "$output" =~ ([[:space:]]*new doc:[[:space:]]*LICENSE.md) ]] || false
+
+    # Explicitly defining the file ignores it.
+    run dolt sql -q "SELECT DOLT_RESET('LICENSE.md')"
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ ("error: the table(s) LICENSE.md do not exist") ]] || false
+}
+
+@test "DOLT_RESET works on specific tables" {
```
zachmu, line 113: Should have at least one test for combining --hard or --soft with a table argument

go/libraries/doltcore/env/repo_state.go
```diff
@@ -249,3 +251,38 @@ func UpdateStagedRoot(ctx context.Context, ddb *doltdb.DoltDB, rsw RepoStateWrit
...
+	case doltdb.ErrNomsIO:
+		return errhand.BuildDError("fatal: failed to write value").Build()
+	case ErrStateUpdate:
+		return errhand.BuildDError("fatal: failed to update the staged root state").Build()
+	}
+
+	return nil
+}
+
+func GetRoots(ctx context.Context, ddb *doltdb.DoltDB, rsr RepoStateReader) (*doltdb.RootValue, *doltdb.RootValue, *doltdb.RootValue, error) {
```
zachmu, line 268: Label the return values

go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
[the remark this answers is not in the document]
zachmu: You are doing IO above (getting roots from the env). Don't do that until you know you will need them.

zachmu (approved):
LG, just a couple comments

PR #1192, opened 2021-01-07
Use map.IterRange for iterating over rows

bheni:
Before
```
~/datasets/wikipedia-ngrams>GOMAXPROCS=1 time dolt sql -r null -q 'SELECT * FROM bigram_counts'
      310.13 real       285.36 user         9.16 sys
```

After
```
~/datasets/wikipedia-ngrams>GOMAXPROCS=1 time dolt sql -r null -q 'SELECT * FROM bigram_counts'
      253.91 real       233.10 user         9.67 sys
```


andrew-wm-arthur (approved):
LGTM!

PR #1200, opened 2021-01-08
go/libraries/doltcore/remotestorage: Tracing and configurability improvements.

reltuk:
Adds the ability to tune I/O parallelism and install a custom compressed chunk cache, if a client wants to.

Adds some simple trace spans to keep track of how much data was fetched when and in how many separate requests, etc.

Still a lot that could be done, but this was a good start from some insights and some improvements.


bheni (approved):
lgtm

PR #1205, opened 2021-01-11
Map Reading Performance Improvements

bheni:
Decreases allocations in reading values from tuples and removes additional go routine calls for reading from maps.


reltuk (approved):
Looks great!

bheni:
@benchmark

PR #1206, opened 2021-01-11
Read Tuple Perf Improvements

bheni:
Old implementation read fields of a tuple to types.Values, converted those values to go values, and then returned those as an interface{}.
New implementation skips types.Values stage and goes directly to interface{}

### DIFF
--- a/go/libraries/doltcore/schema/typeinfo/inlineblob.go
+++ b/go/libraries/doltcore/schema/typeinfo/inlineblob.go
@@ -43,6 +43,7 @@ func (ti *inlineBlobType) ConvertNomsValueToValue(v types.Value) (interface{}, e
 	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), v.Kind())
 }
 
+// ReadFrom reads a go value from a noms types.CodecReader directly
 func (ti *inlineBlobType) ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error) {
 	k := reader.ReadKind()
 	switch k {
--- a/go/libraries/doltcore/schema/typeinfo/inlineblob.go
+++ b/go/libraries/doltcore/schema/typeinfo/inlineblob.go
@@ -43,6 +43,19 @@ func (ti *inlineBlobType) ConvertNomsValueToValue(v types.Value) (interface{}, e
 	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), v.Kind())
 }
 
+func (ti *inlineBlobType) ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error) {
+	k := reader.ReadKind()
+	switch k {
+	case types.InlineBlobKind:
+		bytes := reader.ReadInlineBlob()
+		return bytes, nil
+	case types.NullKind:
+		return nil, nil
+	}
+
+	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), k)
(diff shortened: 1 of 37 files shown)


bheni:
@benchmark

bheni:
This has pretty big performance improvement:

from:

```
BenchmarkMapItr-8   	 4410943	      2183 ns/op	     957 B/op	      22 allocs/op
```

to

```
BenchmarkMapItr-8   	 7985829	      1209 ns/op	     506 B/op	      14 allocs/op
```

go/libraries/doltcore/schema/typeinfo/varbinary.go
```diff
@@ -85,6 +85,20 @@ func (ti *varBinaryType) ConvertNomsValueToValue(v types.Value) (interface{}, er
 	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), v.Kind())
 }
 
+// ReadFrom reads a go value from a noms types.CodecReader directly
+func (ti *varBinaryType) ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error) {
+	k := reader.ReadKind()
```
Hydrocharged, line 90: Panic here for now as I've changed all of this logic, and I don't wanna accidentally forget to make the change on my PR.

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -73,6 +73,9 @@ type TypeInfo interface {
 	// the given type.
 	ConvertNomsValueToValue(v types.Value) (interface{}, error)
 
+	// ReadFrom reads a go value from a noms types.CodecReader directly
+	ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error)
```
Hydrocharged, line 77: Just a nitpick, but since all of the other interfaces have a name for the variable, I'd say this one should too for the format.

go/libraries/doltcore/schema/typeinfo/inlineblob.go
```diff
@@ -43,6 +43,20 @@ func (ti *inlineBlobType) ConvertNomsValueToValue(v types.Value) (interface{}, e
 	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), v.Kind())
 }
 
+// ReadFrom reads a go value from a noms types.CodecReader directly
+func (ti *inlineBlobType) ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error) {
+	k := reader.ReadKind()
```
Hydrocharged, line 48: Panic here for now as I've changed all of this logic, and I don't wanna accidentally forget to make the change on my PR.

Hydrocharged (approved):
LGTM!

go/libraries/doltcore/schema/typeinfo/inlineblob.go
Hydrocharged: Wait no this is wrong, ignore that don't panic here. I'll leave another suggestion for what it should be.

go/libraries/doltcore/schema/typeinfo/inlineblob.go
```diff
@@ -43,6 +43,20 @@ func (ti *inlineBlobType) ConvertNomsValueToValue(v types.Value) (interface{}, e
...
+	switch k {
+	case types.InlineBlobKind:
+		bytes := reader.ReadInlineBlob()
+		return bytes, nil
```
Hydrocharged, line 52:
```suggestion
		return string(bytes), nil
```

PR #1209, opened 2021-01-12
Added binary types to dolt

Hydrocharged:
This is meaty, so I'll add tests in another PR. A ton of files changed, but that's from changing two methods on the `TypeInfo` interface, so I had to make the appropriate changes all around.

### DIFF
--- a/go/libraries/doltcore/schema/typeinfo/varbinary.go
+++ b/go/libraries/doltcore/schema/typeinfo/varbinary.go
@@ -74,8 +74,20 @@ func (ti *varBinaryType) ConvertNomsValueToValue(v types.Value) (interface{}, er
 
 // ReadFrom reads a go value from a noms types.CodecReader directly
 func (ti *varBinaryType) ReadFrom(_ *types.NomsBinFormat, reader types.CodecReader) (interface{}, error) {
-	// todo - implement
-	panic("not implemented yet")
+	k := reader.PeekKind()
+	switch k {
+	case types.BlobKind:
+		val, err := reader.ReadBlob()
+		if err != nil {
+			return nil, err
+		}
+		return ti.fromBlob(val)
+	case types.NullKind:
+		_ = reader.ReadKind()
+		return nil, nil
+	}
+
+	return nil, fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), k)
 }
 
 // ConvertValueToNomsValue implements TypeInfo interface.
--- a/go/libraries/doltcore/schema/typeinfo/varbinary.go
+++ b/go/libraries/doltcore/schema/typeinfo/varbinary.go
@@ -15,8 +15,13 @@
 package typeinfo
 
 import (
+	"bytes"
+	"context"
+	"encoding/binary"
 	"fmt"
+	"io"
 	"strconv"
+	"unsafe"
 
 	"github.com/dolthub/go-mysql-server/sql"
 	"github.com/dolthub/vitess/go/sqltypes"
@@ -25,11 +30,7 @@ import (
 )
 
 const (
-	varBinaryTypeParam_Length        = "length"
-	varBinaryTypeParam_SQL           = "sql"
-	varBinaryTypeParam_SQL_Binary    = "bin"
-	varBinaryTypeParam_SQL_VarBinary = "varbin"
-	varBinaryTypeParam_SQL_Blob      = "blob"
+	varBinaryTypeParam_Length = "length"
 )
 
 // As a type, this is modeled more after MySQL's story for binary data. There, it's treated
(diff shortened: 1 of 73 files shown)


Hydrocharged:
Also, there are changes in https://github.com/dolthub/dolt/pull/1206 that will affect this one, but it won't be anything major, so I'm excluding all of that.

go/store/types/chunkedstring.go
```diff
@@ -0,0 +1,332 @@
...
+var _ valueReadWriter = ChunkedString{}
+
+//TODO: Consolidate all chunk size literals into global constants (this should be the smallest constant)
+const maxChunkedStringChunk = 1<<27 - 1024
+
+// NewChunkedString returns a new ChunkedString.
+func NewChunkedString(ctx context.Context, vrw ValueReadWriter, s string) (ChunkedString, error) {
+	strVal := String(s)
+	var strChunks []String
+	for true {
```
zachmu, line 37: for {

go/store/types/chunkedstring.go
```diff
@@ -0,0 +1,332 @@
...
+
+type ChunkedString struct {
+	valueImpl
+}
+
+var _ Value = ChunkedString{}
+var _ valueReadWriter = ChunkedString{}
+
+//TODO: Consolidate all chunk size literals into global constants (this should be the smallest constant)
+const maxChunkedStringChunk = 1<<27 - 1024
```
zachmu, line 31: Discussion of how this number is derived?

go/store/types/chunkedstring.go
```diff
@@ -0,0 +1,332 @@
...
+		targetVal, err := valRef.TargetValue(ctx, vrw)
+		if err != nil {
+			return ChunkedString{}, err
+		}
+		if targetVal == nil {
+			_, err = vrw.WriteValue(ctx, val)
+			if err != nil {
+				return ChunkedString{}, err
+			}
+		}
```
zachmu, line 70: What happens if the target value is not nil?

go/store/types/value_store.go
```diff
@@ -99,12 +99,19 @@ const (
...
-// and is backed by a chunks.TestStore.
+// and is backed by a chunks.TestStore. Used for testing noms.
 func newTestValueStore() *ValueStore {
 	ts := &chunks.TestStorage{}
 	return NewValueStore(ts.NewView())
 }
 
+// NewInternalValueStore creates a simple struct that satisfies ValueReadWriter
+// and is backed by a chunks.TestStore. Used for dolt operations outside of noms.
+func NewInternalValueStore() *ValueStore {
```
zachmu, line 110: Kind of a bad name. ReadOnlyValueStore? The fact that it's internal doesn't really tell me anything useful about its limitations.

go/libraries/doltcore/schema/typeinfo/varbinary.go
```diff
@@ -171,11 +142,11 @@ func (ti *varBinaryType) IsValid(v types.Value) bool {
 
 // NomsKind implements TypeInfo interface.
 func (ti *varBinaryType) NomsKind() types.NomsKind {
-	return types.StringKind
+	return types.ChunkedStringKind
```
zachmu, line 145: I thought we had decided that the noms kind for these types depends on the length of the sql type. For LongBlob this makes sense, for other types not as much.

zachmu (approved):
This makes sense as far as I can tell, but I'm out of my depth for the noms kind implementation. Without doing a lot more homework I can't say if it looks good or not, so I'm passing off to Aaron on that part.

It looks like the Varbinary implementation always uses a refed type, whereas I remember you saying only the longer types would be stored in refs. Not sure if this is a mistake or you changed your mind. Seems like shorter varbinaries should be stored inline.

go/store/types/chunkedstring.go
Hydrocharged: From searching around the codebase to see what the lowest number for any table or chunk size limit that I could find, and then subtracting `1024` from it to account for prepended metadata (such as noms kinds, etc.). So the `1024` is arbitrary and can honestly be like `4` bytes, but it doesn't matter too much.

However, all over the code base we've got some places referencing a constant, and other places are just `1<<28`, so it's hard to search for them all, so I added the `TODO` so we can just have a collection of constants to reference.

go/store/types/chunkedstring.go
Hydrocharged: Then the `Ref` hash points to a `Value` that already exists, so we don't need to write anything new.

go/store/types/value_store.go
Hydrocharged: It's not read only. Maybe `NewMemoryValueStore` would be better. It's just a value store that in memory and should be used if something needs to write something to get a hash (such as a `Ref` used by `ChunkedString`) but we don't want to actually write to disk.

reltuk: Should this just panic if we try to write a value to it? It's hard for me to reason about times when we want a Blob/ChunkedSting instance floating around that has refs in it to a transient `ValueStore` but we know for certain that thing is on the right side of a firewall such that it never gets written to a persisted `ValueStore` or something...

Hydrocharged: So this is really only being used for places where we are displaying data, such as diffs and whatnot. If, in the process of diffing one of the blob types pops up and needs a VRW, we have this one that it can interact with. Everywhere that's on a write path (SQL, merge, import, etc.) is making use of the contextual VRW.

go/libraries/doltcore/schema/typeinfo/varbinary.go
Hydrocharged: So `BINARY/VARBINARY` use `inlineBlobType`, where as all of the `BLOB` types use `varBinaryType`, or `InlineBlob` and `ChunkedString` respectively.

Although the smaller `BLOB` types could technically be inlined as well, this mimics MySQL's implementation (where all `BLOB` types are stored out of band), and really someone should just use a `VARBINARY` over any `TINYBLOB`.

PR #1215, opened 2021-01-13
Write to root on every loop of sql shell.

VinaiRachakonda:
This pr fixes a bug where the root gets written on every loop of the dolt sql shell.

### DIFF
--- a/bats/sql-shell.bats
+++ b/bats/sql-shell.bats
@@ -33,11 +33,12 @@ teardown() {
     echo "$output"
 
     # 2 tables are created. 1 from above and 1 in the expect file.
-    [[ "$output" =~ "+----------+" ]] || false
-    [[ "$output" =~ "| COUNT(*) |" ]] || false
-    [[ "$output" =~ "+----------+" ]] || false
-    [[ "$output" =~ "| 2        |" ]] || false
-    [[ "$output" =~ "+----------+" ]] || false
+    [[ "$output" =~ "+-------------+" ]] || false
+    [[ "$output" =~ "| Table       |" ]] || false
+    [[ "$output" =~ "+-------------+" ]] || false
+    [[ "$output" =~ "| test        |" ]] || false
+    [[ "$output" =~ "| test_expect |" ]] || false
+    [[ "$output" =~ "+-------------+" ]] || false
 }
 
 @test "bad sql in sql shell should error" {
--- a/bats/sql-shell.bats
+++ b/bats/sql-shell.bats
@@ -27,6 +27,19 @@ teardown() {
     [[ "$output" =~ "pk" ]] || false
 }
 
+@test "sql shell writes to disk after every iteration (autocommit)" {
+    skiponwindows "Need to install expect and make this script work on windows."
+    run $BATS_TEST_DIRNAME/sql-shell.expect
+    echo "$output"
+
+    # 2 tables are created. 1 from above and 1 in the expect file.
+    [[ "$output" =~ "+----------+" ]] || false
+    [[ "$output" =~ "| COUNT(*) |" ]] || false
+    [[ "$output" =~ "+----------+" ]] || false
+    [[ "$output" =~ "| 2        |" ]] || false
+    [[ "$output" =~ "+----------+" ]] || false
+}
+
 @test "bad sql in sql shell should error" {
     run dolt sql <<< "This is bad sql"
     [ $status -eq 1 ]
--- a/bats/sql-shell.expect
+++ b/bats/sql-shell.expect
@@ -2,11 +2,20 @@
 
 set timeout 1
 spawn dolt sql
+set id $spawn_id
 
-expect "doltsql> "
-send "CREATE TABLE test_expect (pk int primary key);\r"
+expect -i id "doltsql> "
+send -i id "CREATE TABLE test_expect (pk int primary key);\r"
 
-expect "doltsql> "
-send "select COUNT(*) from dolt_status;\r"
+expect -i id "doltsql> "
 
-expect eof
+# spawn the second process
+spawn dolt sql
+set id2 $spawn_id
+
+# Todo: Should this be a dolt ls instead ?
+expect -i id2 "doltsql> "
+send -i id2 "select COUNT(*) from dolt_status;\r"
+
+expect -i id eof
+expect -i id2 eof
--- a/bats/sql-shell.expect
+++ b/bats/sql-shell.expect
@@ -1,0 +1,12 @@
+#!/usr/bin/expect
+
+set timeout 1
+spawn dolt sql
+
+expect "doltsql> "
+send "CREATE TABLE test_expect (pk int primary key);\r"
+
+expect "doltsql> "
+send "select COUNT(*) from dolt_status;\r"
+
+expect eof
--- a/bats/sql-shell.expect
+++ b/bats/sql-shell.expect
@@ -15,7 +15,7 @@ set id2 $spawn_id
 
 # Todo: Should this be a dolt ls instead ?
 expect -i id2 "doltsql> "
-send -i id2 "select COUNT(*) from dolt_status;\r"
+send -i id2 "show tables;\r"
 
 expect -i id eof
 expect -i id2 eof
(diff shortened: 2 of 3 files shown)


go/cmd/dolt/commands/sql.go
```diff
@@ -679,6 +679,20 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 		currPrompt := fmt.Sprintf("%s> ", ctx.GetCurrentDatabase())
 		shell.SetPrompt(currPrompt)
 		shell.SetMultiPrompt(fmt.Sprintf(fmt.Sprintf("%%%ds", len(currPrompt)), "-> "))
+		roots, err := se.getRoots(ctx)
```
zachmu, line 682: If you're going to do this, then extract a method that does it and make all the methods that return a root (execShell, execBatch, etc.) do it there, rather than in the main method.

bats/sql-shell.bats
```diff
@@ -27,6 +27,19 @@ teardown() {
     [[ "$output" =~ "pk" ]] || false
 }
 
+@test "sql shell writes to disk after every iteration (autocommit)" {
```
zachmu, line 30: This isn't testing what you think it is. The expect script ends, which means the shell exits. You need to spawn a new shell while the old one is still running (or otherwise issue SQL queries).

coffeegoddd:
#benchmark

coffeegoddd:
#benchmark

coffeegoddd:
#benchmark

bats/sql-shell.expect
```diff
@@ -0,0 +1,21 @@
...
+expect -i id "doltsql> "
+
+# spawn the second process
+spawn dolt sql
+set id2 $spawn_id
+
+# Todo: Should this be a dolt ls instead ?
+expect -i id2 "doltsql> "
+send -i id2 "show tables;\r"
+
```
zachmu, line 19: You're not actually examining the output of show tables here

go/cmd/dolt/commands/sql.go
```diff
@@ -303,66 +304,48 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
-	if err != nil {
-		return nil, errhand.BuildDError("unable to start shell").AddCause(err).Build()
+		return errhand.VerboseErrorFromError(err)
 	}
 
-	newRoots, err := se.getRoots(sqlCtx)
+	err = runShell(sqlCtx, se, mrEnv, roots)
 	if err != nil {
-		return nil, errhand.BuildDError("failed to get roots").AddCause(err).Build()
+		return errhand.BuildDError("unable to start shell").AddCause(err).Build()
```
zachmu, line 329: Wrong error message here. Shell could start but then die. Just return the error from above.

go/cmd/dolt/commands/sql.go
```diff
@@ -303,66 +304,48 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
 	if err != nil {
-		return nil, errhand.BuildDError("Error processing batch").Build()
+		return errhand.BuildDError("Error processing batch").Build()
 	}
 
-	newRoots, err := se.getRoots(sqlCtx)
-	if err != nil {
-		return nil, errhand.BuildDError("failed to get roots").AddCause(err).Build()
-	}
+	verr := writeRoots(sqlCtx, se, mrEnv, roots)
```
zachmu, line 346: Just return writeRoots

go/cmd/dolt/commands/sql.go
```diff
@@ -375,32 +358,29 @@ func newBatchedDatabase(name string, dEnv *env.DoltEnv) dsqle.Database {
...
-			return nil, errhand.VerboseErrorFromError(err)
+			return errhand.VerboseErrorFromError(err)
 		}
 	}
 
-	newRoots, err := se.getRoots(sqlCtx)
-	if err != nil {
-		return nil, errhand.BuildDError("failed to get roots").AddCause(err).Build()
-	}
+	verr := writeRoots(sqlCtx, se, mrEnv, roots)
```
zachmu, line 381: return writeRoots

go/cmd/dolt/commands/sql.go
```diff
@@ -687,6 +673,31 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 	return nil
 }
 
+// writeRoots updates the working root values.
```
zachmu, line 676: Explain params in comment

go/cmd/dolt/commands/sql.go
```diff
@@ -676,6 +656,12 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 			}
 		}
 
+		verr := writeRoots(ctx, se, mrEnv, initialRoots)
+
+		if verr != nil {
+			shell.Println(verr.Verbose())
```
zachmu, line 662: This should be fatal, just return the error here.

go/cmd/dolt/commands/sql.go
```diff
@@ -676,6 +656,12 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 			}
 		}
 
+		verr := writeRoots(ctx, se, mrEnv, initialRoots)
```
zachmu, line 659: You only want to write new roots in the case that err was not nil above.

Should have tests for this.

zachmu: Probably means we need to restructure above logic, since err is only defined in that if block scope.

zachmu (commented):
Couple small changes, almost there

bats/sql-shell.expect
VinaiRachakonda: It should get examined in the test case

go/cmd/dolt/commands/sql.go
VinaiRachakonda: Shouldn't the error be nil for me to write to the roots?

VinaiRachakonda: Also not sure how to test this, just simulate an error in the shell?

VinaiRachakonda:
I removed the check for the SET command since that break a bunch of bats test, I think we'll leave the autocommit feature on for now.

bats/sql-shell.expect
zachmu: You can't examine it in the test case, that's too late. The whole point is to make sure that the two SQL sessions can read each other's writes before they exit. In the bats script, both processes have exited by the time you do any assertions. It's too late.

go/cmd/dolt/commands/sql.go
```diff
@@ -676,6 +653,15 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 			}
 		}
 
+		if err == nil {
+			verr := writeRoots(ctx, se, mrEnv, initialRoots)
+
+			if verr != nil {
```
zachmu, line 659: You want to return this error from runShell, not just print it

zachmu (commented):
Code looks good mostly, but the tests are still not testing what you say they are

go/cmd/dolt/commands/sql.go
```diff
@@ -676,6 +654,15 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv) error {
 			}
 		}
 
+		if err == nil {
+			returnedVerr = writeRoots(ctx, se, mrEnv, initialRoots)
+
+			if returnedVerr != nil {
+				shell.Println(returnedVerr.Verbose())
```
zachmu, line 661: Sure we aren't going to double-print this error? Check and see how it's handled at the call site.

zachmu (approved):
LG, thanks for the many changes along the way

go/cmd/dolt/commands/sql.go
VinaiRachakonda: Good catch, it would eventually get double printed.

PR #1218, opened 2021-01-14
remotestorage: Fix download aggregation to correctly aggregated based on prior chunk in batch, not first chunk.

reltuk:


zachmu (approved):
A+++++ would review again

PR #1220, opened 2021-01-14
Print pipeline fix

bheni:
Fixes issue https://github.com/dolthub/dolt/issues/1219 by not calling on an iterator after it's already returned EOF

### DIFF
--- a/go/cmd/dolt/commands/sql_print.go
+++ b/go/cmd/dolt/commands/sql_print.go
@@ -93,12 +93,18 @@ func sqlColToStr(col interface{}) string {
 
 // getReadStageFunc is a general purpose stage func used by multiple pipelines to read the rows into batches
 func getReadStageFunc(iter sql.RowIter, batchSize int) pipeline.StageFunc {
+	isDone := false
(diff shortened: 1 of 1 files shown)


go/cmd/dolt/commands/sql_print.go
```diff
@@ -93,12 +93,18 @@ func sqlColToStr(col interface{}) string {
 
 // getReadStageFunc is a general purpose stage func used by multiple pipelines to read the rows into batches
 func getReadStageFunc(iter sql.RowIter, batchSize int) pipeline.StageFunc {
```
bheni, line 95: this function reads a batch of items and returns them.  If during a batch read an io.EOF was returned by the underlying iterator, the batch that had been read so far would be returned, and then it would rely on a subsequent call to the underlying iterator to properly return eof.  Rather than call into the iterator again store a variable saying that the iterator has been exhausted.

An alternative would be to return io.EOF with the batch of data, but I find that error prone, even though I've seen this done in other iterators / readers.

zachmu (approved):

PR #1221, opened 2021-01-15
cli output fix

bheni:
fixes issues where printf formatting was being unintentionally applied for some cli output


timsehn:
Bats test?

zachmu (approved):
LGTM

PR #1222, opened 2021-01-15
benchmark perfect hashing table index

andrew-wm-arthur:
We can hash 1M 20 byte keys in about 2-4 seconds, producing a hash function that requires 800K to store

```
BenchmarkCHDIndex
BenchmarkCHDIndex/build_MPH_table
BenchmarkCHDIndex/build_MPH_table-12         	       1	1859812551 ns/op
20000000
800004
PASS
```

This was tested with a 90% load factor, meaning we'll have an 11% space overhead in the storage table. The load factor can probably be faster without a significant slowdown in hash construction


coffeegoddd:
#benchmark

coffeegoddd:
#benchmark

andrew-wm-arthur:
@coffeegoddd what's going on with that second benchmark?

PR #1226, opened 2021-01-19
ALTER TABLE MODIFY COLUMN Pt. 1

Hydrocharged:
Here you can see the core idea of the approach: each type handling every other type. Before, we were using `FormatValue` and `ParseValue` to convert between types, which worked decently for importing from CSVs (where you could pretend everything is coming from a `varchar`), but that approach won't work between other types. Even within types (such as `ENUM('a','b','c')` to `ENUM('a','c')`) you can have altered values, so we have to be able to handle all of these special (and largely undocumented) behaviors. This way is easily the most tedious, but we can guarantee correct behavior with it (plus once it's done it's done).

The other approach I thought about was to just special case all of them within the current `Convert` function, but it looked like it would balloon into a less structured version of this approach anyway.

As a result, I'm also going to remove `ParseValue`, as its sole reason for existence was for type conversion, so the additional function would just cause confusion (we can resort to converting random strings by pretending they're `varchar`s as mentioned before). Along with that change, I also need to handle primary key column changes and foreign keys. And of course the ridiculous amount of testing (I may make use of the test harnesses, which I'll then translate to go and bats tests), which will modify some/most of the placeholder conversion logic.

Of note, I looked into keeping the row data (and therefore tag) the same for type conversions that would not fundamentally change the underlying data (such as `int` to `bigint`). I determined it's outside of the scope of this set of PRs, but it's something I'd like to look more into in the future, as it would open up better diff behavior for some conversions.

### DIFF
--- a/go/libraries/doltcore/schema/alterschema/modifycolumn.go
+++ b/go/libraries/doltcore/schema/alterschema/modifycolumn.go
@@ -16,16 +16,15 @@ package alterschema
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
-
-	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/row"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema/encoding"
+	"github.com/dolthub/dolt/go/libraries/doltcore/schema/typeinfo"
+	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
 	"github.com/dolthub/dolt/go/store/types"
 )
 
@@ -38,7 +37,6 @@ func ModifyColumn(
 	newCol schema.Column,
 	order *ColumnOrder,
 ) (*doltdb.Table, error) {
-
 	sch, err := tbl.GetSchema(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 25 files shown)


go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -86,33 +73,30 @@ func validateModifyColumn(ctx context.Context, tbl *doltdb.Table, existingCol sc
...
 
-	if existingCol.Kind != modifiedCol.Kind || !existingCol.TypeInfo.Equals(modifiedCol.TypeInfo) {
-		return errors.New("unsupported feature: column types cannot be changed")
-	}
+	if existingCol.Name != modifiedCol.Name {
+		cols := sch.GetAllCols()
+		err = cols.Iter(func(currColTag uint64, currCol schema.Column) (stop bool, err error) {
+			if currColTag == modifiedCol.Tag {
+				return false, nil
+			} else if currCol.Name == modifiedCol.Name {
```
zachmu, line 81: Lower-case these for comparison

zachmu: We don't do this elsewhere but we really need to start, it's on the backlog

go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -122,21 +106,56 @@ func updateTableWithModifiedColumn(ctx context.Context, tbl *doltdb.Table, newSc
...
+				if err != nil {
+					return true, err
+				}
+				newRowKey, err := r.NomsMapKey(newSch).Value(ctx)
+				if err != nil {
+					return true, err
+				}
+				mapEditor.Set(newRowKey, r.NomsMapValue(newSch))
+			} else {
+				_, err := convFunc(ctx, vrw, val)
```
zachmu, line 148: Why run this function at all if you're not saving the result? Comment plz

zachmu (commented):
This approach seems sound, nice work.

I would focus on getting a good enough set of covering tests so we can check this in and iterate on it.

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -667,6 +662,194 @@ func TestModifyAndChangeColumn(t *testing.T) {
...
+	tests := []struct {
+		name            string
+		setupStmts      []string
+		alterStmt       string
+		tableName       string
+		expectedRows    [][]types.Value
+		expectedIdxRows [][]types.Value
+		expectedErr     bool
+	}{
+		{
```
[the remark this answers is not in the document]
zachmu: This doesn't have to be exhaustive, but we should have more coverage than we do here.

zachmu (commented):
The code looks quite good, nice work.

The tests are a bit lacking. Even if you plan to add more later, you should get better coverage than this on your first pass. I left some suggestions. Nothing drastic, just a bit better than we have here, and better coverage in the bats tests.

Hydrocharged:
Merged into https://github.com/dolthub/dolt/pull/1281

PR #1231, opened 2021-01-20
Performance Benchmarks Email on Release

coffeegoddd:
This PR adds an action that sends an email to a list of recipients. It also adds a manually triggered workflow that benchmarks a version of dolt against mysql. This workflow will later be integrated into the release-workflow once the results and queries are properly tuned.


coffeegoddd:
merging with plan to iterate on tuning performance results, will then revisit and update this

PR #1233, opened 2021-01-20
fix panic when key does not exist in table

bheni:


go/libraries/doltcore/row/fmt.go
```diff
@@ -34,7 +36,6 @@ func FieldSeparatedFmt(delim rune) RowFormatFunc {
 		}
 
 		allCols := sch.GetAllCols()
-		kvps := make([]string, 0, allCols.Size())
```
bheni, line 37: this didn't actually do anything

bats/conflict-detection.bats
```diff
@@ -922,6 +922,12 @@ SQL
     dolt branch -d -f merge-into-modified
     dolt checkout -b merge-into-modified modifier
     dolt merge deleter
+
+    # Test resolve nonexistant key
```
zachmu, line 926: nonexistent

zachmu (approved):
LG

PR #1242, opened 2021-01-22
projected column fixes

bheni:
For both indexed and unindexed tables when the engine would call WithProjection on our sql.Table implementation we would return a new sql.Table that embedded the old one and had a member which was the list of projected columns.  Then, when sql.Table.PartitionRows was called, it was called on the embedded object directly and the projected columns were not accessible.


zachmu (approved):
LGTM

bheni:
#benchmark

PR #1245, opened 2021-01-25
fixes panic in the atomic package

bheni:
The `atomic` package has the following [limitation on 32-bit platforms:](https://golang.org/pkg/sync/atomic/#pkg-note-BUG)

> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/.idea/
 .vscode
+.run
 venv
 
 go.sum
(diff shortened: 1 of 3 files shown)


reltuk (approved):
LGTM!

PR #1248, opened 2021-01-25
Hidden command for changing the type of a column in place. 

### LINKED ISSUE #983: dolt should not allow columns with the same case-insensitive name
This is an error on the SQL path, but is currently allowed on other import paths. It leads to undefined behavior in SQL queries.

zachmu:
This tool can't safely be run because it will break repos read by older versions of dolt. We need to turn on feature version writing before that. So for now it's only safe to use on repos that haven't been shared (and won't be shared) outside dolt team.

This PR also cleans up a bunch of other issues that came up when digging in this space:
* Introduce column compatibility checks, for columns that have the same tag and storage type but different type params.
* Stop checking tag collision errors on NewColCollection. A couple different use cases construct ColCollections with duplicate tags by design.
* Introduce checks to prevent columns with the same case-insensitive name. This fixes #983
* Introduce last-resort sanity check for schemas before serializing them to disk


Hydrocharged (changes requested):
Using this to change a column type will break things.

```
$ dolt init

$ dolt sql <<SQL
> CREATE TABLE test(
>   pk BIGINT PRIMARY KEY,
>   v1 BIGINT
> );
> INSERT INTO test VALUES (0,0);
> SQL
Rows inserted: 1 Rows updated: 0 Rows deleted: 0

$  dolt add -A

$ dolt commit -m "message"
commit pogem5tks8o38k0dbqti7ahin5tfc822
Author: Daylon Wilkins <redacted>
Date:   Mon Jan 25 13:53:05 -0800 2021

        message


$ dolt schema change-type test v1 int

$ dolt diff
diff --dolt a/test b/test
--- a/test @ oaq7vpcslbuuccsebshblv6m4ss9sptf
+++ b/test @ o393iri8g82n6hq5lemufkfrh7qke9md
  CREATE TABLE test (
    `pk` BIGINT NOT NULL
<   `v1` BIGINT
>   `v1`    INT
     PRIMARY KEY (pk)
  );

Failed to merge schemas
cause: two different columns with the same tag

$ dolt add -A
Unknown error
cause: tag collision for columns v1 and v1, different definitions (tag: 5758)
```

doInfinitely (approved):
LGTM

zachmu:
@reltuk and @bheni 

Daylon caught an important issue, which I have now fixed and added tests for. Can you think of other places I should have coverage that I don't?

Is "compatibility" a meaningful concept, and is it defined correctly?

An important note is that we can't run this on any existing repository without breaking existing clients. Anybody working with those repositories would need to upgrade their dolt client or face panics / errors (bit of each). We should still get this code out, since we want in-place changes for schema changes that don't require data rewrites, and the existing schema / column collection code will break when we do that. But we have to think carefully about if / when we run this command.

Hydrocharged:
> 
> 
> @reltuk and @bheni
> 
> Daylon caught an important issue, which I have now fixed and added tests for. Can you think of other places I should have coverage that I don't?
> 
> Is "compatibility" a meaningful concept, and is it defined correctly?
> 
> An important note is that we can't run this on any existing repository without breaking existing clients. Anybody working with those repositories would need to upgrade their dolt client or face panics / errors (bit of each). We should still get this code out, since we want in-place changes for schema changes that don't require data rewrites, and the existing schema / column collection code will break when we do that. But we have to think carefully about if / when we run this command.

One of the planned additions for schema column type changes using `ALTER TABLE` is this exact functionality that you've added here, so it will be integrated into the SQL path rather soon. It makes diffing and merging far more elegant when switching between _"compatible"_ types.

andrew-wm-arthur (approved):
lgtm

PR #1269, opened 2021-01-28
Implement StatisticsTable interface to provide table statistics to the analyzer.

bheni:


zachmu (approved):
LG

Also note the latest GMS has a breaking test change for self-referential foreign keys. You'll need to add a skip for it.

PR #1270, opened 2021-01-28
remove layered parallelism.

bheni:


bheni:
#benchmark

bheni:
#benchmark

bheni:
investigating alternate approaches that don't hurt index scan times.

PR #1271, opened 2021-01-28
Enable sqllogictest runner to parse results to csv

coffeegoddd:
This PR enables `csv` formatted parsed `sqllogictest` results.

### DIFF
--- a/go/libraries/doltcore/sqle/logictest/main/main.go
+++ b/go/libraries/doltcore/sqle/logictest/main/main.go
@@ -16,7 +16,9 @@ package main
 
 import (
 	"bytes"
+	"encoding/csv"
 	"encoding/json"
+	"fmt"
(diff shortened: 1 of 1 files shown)


go/libraries/doltcore/sqle/logictest/main/main.go
```diff
@@ -59,14 +61,21 @@ func parseTestResults(version, f string) {
 		records[i] = NewDoltRecordResult(e, version)
 	}
 
-	b, err := JSONMarshal(records)
-	if err != nil {
-		panic(err)
-	}
+	if os.Getenv("FORMAT") == "csv" {
```
zachmu, line 64: Use a -r arg, not an env var

zachmu (approved):
LG

PR #1272, opened 2021-01-28
Create, list, delete workspace refs [no-release-notes]

tbantle22:
Edit: Removed workspace command, so the below is no longer relevant

~~Some questions that have come up while implementing this:~~
~~1. Should users be able to create a workspace using dolt, or only using dolthub?~~
~~2. Should users be able to name their workspace?~~
~~3. Should workspaces share the the same namespace as branches?~~

To do:
- [ ] Test fetch workspace ref
- [ ] Test push workspace ref

Also I refactored `CreateBranch` and `CreateWorkspace` so the `OnDB` versions can be easily used by dolthubapi (since it doesn't use `DoltEnv`)


reltuk (approved):
LGTM!

PR #1273, opened 2021-01-28
[no-release-notes] /go/libraries/doltcore/table/typed/schema.go: revert removal of schema.go file

coffeegoddd:
Not sure this should live in dolt... maybe we just move to ld and leave this out?


coffeegoddd:
Gonna close this and add this to service where its used

PR #1280, opened 2021-02-01
o/libraries/doltcore/table/editor: Add ability to configure map editor flush interval with environment variables.

reltuk:


Hydrocharged (approved):
Looks good, although perhaps we should have `DOLT_EDIT_TABLE_BUFFER_ROWS` control both the table and index editors if `DOLT_EDIT_INDEX_BUFFER_ROWS` is not present? I'd assume the average use case is to set both to the same value.

PR #1281, opened 2021-02-01
ALTER TABLE MODIFY COLUMN

Hydrocharged:


zachmu (approved):
LG, although I didn't review the generated tests.

My one major comment is that you should put a README in the same directory as the generated test files, describing what they're supposed to test and how you generated them, so that someone else can debug or reproduce the work you did in the future.

Hydrocharged:
> LG, although I didn't review the generated tests.
> 
> My one major comment is that you should put a README in the same directory as the generated test files, describing what they're supposed to test and how you generated them, so that someone else can debug or reproduce the work you did in the future.

Good idea!

zachmu (commented):
Also it looks like you're always rebuilding indexes in modifycolumn.go. Should only be necessary if the tag changes, right?

go/libraries/doltcore/schema/alterschema/modifycolumn_test.go
```diff
@@ -61,12 +61,12 @@ func TestModifyColumn(t *testing.T) {
 		},
 		{
 			name:           "remove null constraint",
-			existingColumn: schema.NewColumn("id", dtestutils.IdTag, types.UUIDKind, true, schema.NotNullConstraint{}),
-			newColumn:      schema.NewColumn("newId", dtestutils.IdTag, types.UUIDKind, true),
+			existingColumn: schema.NewColumn("age", dtestutils.AgeTag, types.UintKind, false, schema.NotNullConstraint{}),
```
zachmu, line 64: I'd expect to see some tests here of tag reuse. Are there some I'm not seeing?

Hydrocharged:
> Also it looks like you're always rebuilding indexes in modifycolumn.go. Should only be necessary if the tag changes, right?

Not always. Even if the tag remains the same the values can still change. For example, `ENUM` to other `ENUM` types.

go/libraries/doltcore/schema/alterschema/modifycolumn_test.go
zachmu: Testing of the tag reuse in general seems a little bit light. It was a bit confusing toggling between parts 1 and 2 of this change, so I missed this on the first pass. But while you do a good job of making sure the values are converted properly, I would like to see more tests that confirm that tag reuse works as expected.

PR #1290, opened 2021-02-04
Added DELIMITER support to the dolt shell

Hydrocharged:


Hydrocharged:
This will fail all of the tests until I can make a PR against https://github.com/dolthub/ishell, then I'll update this one with the proper commit.

bats/sql-delimiter.expect
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/expect
```
zachmu, line 1: Probably want this to include at least one test of multiple statements like a `BEGIN .. END` block. The only place this is supported in the grammar right now is when creating a trigger.

go/cmd/dolt/commands/sql.go
```diff
@@ -638,27 +638,53 @@ func runShell(ctx *sql.Context, se *sqlEngine, mrEnv env.MultiRepoEnv, initialRo
...
-			verr := formatQueryError("", err)
-			shell.Println(verr.Verbose())
-		} else if rowIter != nil {
-			err = PrettyPrintResults(ctx, se.resultFormat, sqlSch, rowIter)
-			if err != nil {
-				shell.Println(color.RedString(err.Error()))
+		shouldProcessQuery := true
+
+		//TODO: Parse this in a way that can handle comments
+		if len(singleLine) > 9 && strings.HasPrefix(strings.ToLower(singleLine[:9]), "delimiter") {
```
zachmu, line 649: Probably feel better about a regex match than doing prefixes and string splits

go/go.mod
```diff
@@ -78,4 +78,6 @@ require (
 
 replace github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi => ./gen/proto/dolt/services/eventsapi
 
+replace github.com/dolthub/ishell => ../../ishell
```
zachmu, line 81: Don't check this in

zachmu (approved):
LG, just a couple comments

PR #1293, opened 2021-02-04
Insertion optimizations

bheni:
Go from sql rows to types.Tuples directly and use those tuples for insertion without using row.Row.


andrew-wm-arthur (approved):
lgtm

Hydrocharged (approved):
LGTM

PR #1303, opened 2021-02-08
go/go.mod: Bump go-mysql-server

reltuk:


zachmu (approved):
LG assuming CI passes

PR #1306, opened 2021-02-09
Check constraints on commit

Hydrocharged:


bats/foreign-keys.bats
```diff
@@ -962,7 +977,7 @@ SET FOREIGN_KEY_CHECKS=0;
 UPDATE parent SET v1 = v1 - 1;
 SQL
     dolt add -A
-    dolt commit -m "updated parent"
+    dolt commit --force -m "updated parent"
```
timsehn, line 980: Why do we need these --forces now?

Hydrocharged: Previously we didn't check row data on commit (only schema information), but now we do. These tests are verifying that the merge behavior is working correctly when working with invalid table state, so now to commit bad tables we have to force them.

timsehn: Understood

timsehn: I see. All these have `SET FOREIGN_KEY_CHECKS=0;` above.

timsehn (approved):

andrew-wm-arthur (approved):
LGTM.

This is a good safety measure for now, but we need come up with a better, more efficient way to validate FKs. Doing a full scan over two indexes for every Foreign Key is prohibitively expensive. We should think about how to leverage diffs to make this more performant.

Hydrocharged:
> 
> 
> LGTM.
> 
> This is a good safety measure for now, but we need come up with a better, more efficient way to validate FKs. Doing a full scan over two indexes for every Foreign Key is prohibitively expensive. We should think about how to leverage diffs to make this more performant.

I'll add a TODO and make the suggestion so we know to come back to it later.

PR #1308, opened 2021-02-09
Andy/merge dupe tables

andrew-wm-arthur:
Quick fix to address views bug. Repro is described in the bats test `@test "Add views on two branches, merge" `.

This PR only addresses the case where a table created on two branches has the exact same schema. Further work is needed to merge tables when the schema is "overlapping" but not equal.

### DIFF
--- a/go/libraries/doltcore/merge/merge.go
+++ b/go/libraries/doltcore/merge/merge.go
@@ -56,8 +56,13 @@ func (merger *Merger) MergeTable(ctx context.Context, tblName string, sess *edit
 	}
 
 	var h hash.Hash
+	var tblSchema schema.Schema
 	if ok {
 		h, err = tbl.HashOf()
+		if err != nil {
+			return nil, nil, err
+		}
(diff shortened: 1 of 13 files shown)


timsehn (commented):
BATS lgtm

go/libraries/doltcore/merge/merge.go
```diff
@@ -107,7 +122,13 @@ func (merger *Merger) MergeTable(ctx context.Context, tblName string, sess *edit
 
 		if !ancOk {
 			if mergeOk && ok {
-				return nil, nil, ErrSameTblAddedTwice
+				if schema.SchemasAreEqual(tblSchema, mergeTblSchema) {
+					// hackity hack
+					ancTblSchema, ancTbl = tblSchema, tbl
```
bheni, line 127: so this will automatically take the merge table rows in conflict?

bheni (approved):
lgtm.

PR #1316, opened 2021-02-10
[no-release-notes] /{.github,bats}: pin runner, add timeouts

coffeegoddd:
This PR pins the mac runner and attempts to handle `timeout` and `failed` `expect` cases.


timsehn:
https://github.com/dolthub/dolt/blob/ed18a61545d53e314367f9dae2a0da923c6bf0f5/bats/sql.bats#L716

That requires a `run` if we're going to inspect `$output`. Also, we should add a `[ "$status" -eq "0" ]` check as well.

PR #1321, opened 2021-02-10
disable query diff

bheni:
query diff was a little known feature that was not well documented and was the source of a large amount of maintenance work.  We can recover this from git in the future, or reimplement once it is a priority.


reltuk:
Is it worth leaving in a bunch of untested dead code? Likely to rot more and more over time?

bheni:
> Is it worth leaving in a bunch of untested dead code? Likely to rot more and more over time?

deleted it.

reltuk (approved):
LGTM!

PR #1326, opened 2021-02-12
Max/ge bats

max-hoffman:


timsehn:
We gonna merge or close this @max-hoffman ?

max-hoffman:
@timsehn we should merge but I need to fix this to account for extra features. Can I try for Monday? A bit stretched with Aktify/Kedro/Hunter at the moment. @VinaiRachakonda also might be in a good spot to add some of the queries Oscar sent his way

PR #1328, opened 2021-02-13
Andy/write feature version

andrew-wm-arthur:
FeatureVersion is now checked every time a RootVal is read, and is written with every call to `doltdb.WriteRootValue()`.

A CLI flag was added to `dolt.go` to enable FeatureVersion testing.


reltuk (approved):
LGTM!

PR #1329, opened 2021-02-13
CREATE DATABASE (DBDDL)

VinaiRachakonda:
This pr brings `CREATE DATABASE` functionality to dolt and adds associated test cases.

Test cases are failing due to un-updated go-mysql-server and vitess


bats/sql-multi-db.bats
```diff
@@ -83,3 +83,122 @@ seed_repos_with_tables_with_use_statements() {
...
+DROP SCHEMA mydb;
+SQL
+    [[ "$output" =~ "dolt_repo_$$" ]] || false
+    [[ "$output" =~ "information_schema" ]] || false
+    [[ "$output" =~ "mydb" ]] || false
+    [[ "$output" =~ "1" ]] || false
+
+    run dolt sql -q "SHOW DATABASES"
+    [[ ! "$output" =~ "mydb" ]] || false
+}
```
timsehn, line 204: I think we need some more bats test coverage here. Here are some ideas:
* non-happy path tests. drop a db that does not exist, create a db that already exists, etc.
* Add some querying across two databases, both created and the default
* I think SHOW DATABASES and USE need to be explicitly tested. Happy and non-happy.

timsehn (commented):
More bats test coverage. It's our superpower.

bats/sql-create-database.bats
```diff
@@ -0,0 +1,164 @@
...
+
+    [ "$status" -eq 0 ]
+}
+
+@test "sql create and drop new database" {
+    run dolt sql << SQL
+CREATE DATABASE mydb;
+DROP DATABASE mydb;
+USE mydb;
+SQL
```
timsehn, line 59: This is an interesting pattern. Usually, I run these all as separate `dolt sql` commands and then test the one I want output for using `run`. I'm worried that running these all as on SQL block makes them harder to debug. So imagine we broke `DROP` in this case. You wouldn't be sure which one of those three statements broke. Make sense?

bats/sql-create-database.bats
```diff
@@ -0,0 +1,164 @@
...
+CREATE DATABASE IF NOT EXISTS test;
+SHOW DATABASES;
+USE test;
+CREATE TABLE test (
+    pk int primary key
+);
+INSERT INTO test VALUES (222);
+SELECT COUNT(*) FROM test WHERE pk=222;
+DROP DATABASE test;
+SQL
```
timsehn, line 80: This one is even harrier.

bats/sql-create-database.bats
VinaiRachakonda: Yup I usually do that but these in memory dbs expire on every session. So this was the best way to test besides the server.

bats/sql-multi-db.bats
```diff
@@ -82,4 +82,4 @@ seed_repos_with_tables_with_use_statements() {
     echo \"\"\"$output\"\"\"
     [ "$status" -eq 0 ]
     [[ "$output" =~ "$EXPECTED" ]] || false
-}
+}
```
VinaiRachakonda, line 85: will fix this after review

zachmu (approved):
LG

PR #1337, opened 2021-02-16
Bump vitess & engine & add Stored Procedure interfaces

Hydrocharged:
The 1337 PR! Adds stored procedures to Dolt!

### DIFF
--- a/bats/sql.bats
+++ b/bats/sql.bats
@@ -39,7 +39,7 @@ teardown() {
     teardown_common
 }
 
-@test "sql select from multiple tables" {
+@test "sql: select from multiple tables" {
     run dolt sql -q "select pk,pk1,pk2 from one_pk,two_pk"
     [ "$status" -eq 0 ]
     [ "${#lines[@]}" -eq 20 ]
@@ -53,7 +53,7 @@ teardown() {
     [ "${#lines[@]}" -eq 8 ]
 }
 
-@test "sql AS OF queries" {
+@test "sql: AS OF queries" {
     dolt add .
     dolt commit -m "Initial master commit" --date "2020-03-01T12:00:00Z"
 
@@ -157,7 +157,7 @@ teardown() {
     [[ "$output" =~ "not found" ]] || false
 }
 
-@test "sql output formats" {
+@test "sql: output formats" {
     dolt sql <<SQL
     CREATE TABLE test (
     a int primary key,
@@ -190,7 +190,7 @@ SQL
     [ "$output" == '{"rows": [{"a":1,"b":1.5,"c":"1","d":"2020-01-01 00:00:00 +0000 UTC"},{"a":2,"b":2.5,"c":"2","d":"2020-02-02 00:00:00 +0000 UTC"},{"a":3,"c":"3","d":"2020-03-03 00:00:00 +0000 UTC"},{"a":4,"b":4.5,"d":"2020-04-04 00:00:00 +0000 UTC"},{"a":5,"b":5.5,"c":"5"}]}' ]
 }
 
-@test "sql output for escaped longtext exports properly" {
+@test "sql: output for escaped longtext exports properly" {
  dolt sql <<SQL
     CREATE TABLE test (
     a int primary key,
@@ -213,19 +213,19 @@ SQL
     [[ "$output" =~ '2,"""Hello"""' ]] || false
 }
 
-@test "sql ambiguous column name" {
+@test "sql: ambiguous column name" {
     run dolt sql -q "select pk,pk1,pk2 from one_pk,two_pk where c1=0"
     [ "$status" -eq 1 ]
     [ "$output" = "ambiguous column name \"c1\", it's present in all these tables: one_pk, two_pk" ]
 }
 
-@test "sql select with and and or clauses" {
+@test "sql: select with and and or clauses" {
     run dolt sql -q "select pk,pk1,pk2 from one_pk,two_pk where pk=0 and pk1=0 or pk2=1"
     [ "$status" -eq 0 ]
     [ "${#lines[@]}" -eq 13 ]
 }
 
-@test "sql select the same column twice using column aliases" {
+@test "sql: select the same column twice using column aliases" {
     run dolt sql -q "select pk,c1 as foo,c1 as bar from one_pk"
     [ "$status" -eq 0 ]
     [[ ! "$output" =~ "NULL" ]] || false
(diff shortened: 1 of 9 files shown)


go/libraries/doltcore/sqle/database.go
```diff
@@ -746,6 +748,82 @@ func (db Database) DropTrigger(ctx *sql.Context, name string) error {
...
+		return nil, err
+	}
+	return spds, nil
+}
+
+// SaveStoredProcedure implements sql.StoredProcedureDatabase.
+func (db Database) SaveStoredProcedure(ctx *sql.Context, spd sql.StoredProcedureDetails) error {
+	spdAsString := StoredProcedureDetailsToString(spd)
+	return db.addFragToSchemasTable(ctx,
+		"storedprocedure",
```
zachmu, line 814: const plz

go/libraries/doltcore/sqle/database.go
```diff
@@ -886,3 +964,38 @@ func RegisterSchemaFragments(ctx *sql.Context, db Database, root *doltdb.RootVal
 
 	return nil
 }
+
+// StoredProcedureDetailsToString returns a string encoding of sql.StoredProcedureDetails, which may be used to store
+// the details in string format.
+func StoredProcedureDetailsToString(spd sql.StoredProcedureDetails) string {
+	return fmt.Sprintf("%d|%d|%s", spd.CreatedAt.UTC().UnixNano(), spd.ModifiedAt.UnixNano(), spd.CreateStatement)
```
zachmu, line 971: I worry this is not good enough. It's totally possible somebody puts a | character into the body of their stored procedure. Not least because they are doing their own crappy storage hacks that they want to use a stored procedure to make easier.

If the modified / created distinction is important enough to track, then we need to either
1) Modify dolt_schemas to add these fields, or 
2) Store them somewhere else, like on the root value instead of in a user visible table.

Honestly I'm not sure which is better. I like storing them in a table because it makes diffing trivial, but maybe we have outgrown this abstraction. Let's bring in Aaron and talk after standup.

zachmu: Another option would be to use the timestamps from the commit tree for these values. That changes the semantics a bit for changes in the working set, obviously.

zachmu (commented):
Looks fine for what it is, but I am not on board for storing additional metadata (timestamps) like this. We need a better solution.

go/libraries/doltcore/sqle/database.go
zachmu: I thought about this a little more, and the way you have this implemented doesn't have the issue I brought up ("|" characters in the stored procedure body).

I still don't like it, though.

go/libraries/doltcore/doltdb/doltdb_test.go
```diff
@@ -168,11 +168,14 @@ func TestSystemTableTags(t *testing.T) {
 		assert.Equal(t, queryCatalogMin+4, schema.QueryCatalogDescriptionTag)
 	})
 	t.Run("dolt_schemas tags", func(t *testing.T) {
-		doltSchemasMin := sysTableMin + uint64(4007)
+		doltSchemasMin := sysTableMin + uint64(4012)
```
zachmu, line 171: This means that we do a hard fork of the table, instead of just adding columns to it. Any reason that's preferable? You wouldn't be able to diff schemas before and after the cutoff.

go/libraries/doltcore/doltdb/system_table.go
```diff
@@ -185,7 +185,13 @@ const (
 	// The schema fragment associated with the database entity.
 	// For example, the SELECT statement for a CREATE VIEW.
 	SchemasTablesFragmentCol = "fragment"
-	// The name of the index that is on the table.
+	// The time that the entity was created at.
+	SchemasTablesCreatedAtCol = "createdat"
```
zachmu, line 189: created_at

go/libraries/doltcore/doltdb/system_table.go
```diff
@@ -185,7 +185,13 @@ const (
...
+	// The time that the entity was last modified at.
+	SchemasTablesModifiedAtCol = "modifiedat"
```
zachmu, line 191: modified_at

zachmu: or last_modified

go/libraries/doltcore/sqle/schema_table.go
```diff
@@ -136,7 +148,51 @@ func GetOrCreateDoltSchemasTable(ctx *sql.Context, db Database) (retTbl *Writabl
...
+		dRow, err := row.FromNoms(schemasTable.sch, key.(types.Tuple), val.(types.Tuple))
+		if err != nil {
+			return err
+		}
+		sqlRow, err := sqlutil.DoltRowToSqlRow(dRow, schemasTable.sch)
+		if err != nil {
+			return err
+		}
+		// append the new createdat, modifiedat, and metadata to each row
+		sqlRow = append(sqlRow, time.Unix(0, 0).UTC(), time.Unix(0, 0).UTC(), "")
```
zachmu, line 177: time.now() probably better

zachmu (approved):
This looks good, but do we have a backdoor on merge? What happens if I merge a root with the old table schema into one with the new table schema? Badness? Undefined? Would be nice to have a test for it. Also vice versa.

It's fine if we force people to migrate before doing such merges, but we should catch the error.

go/libraries/doltcore/doltdb/doltdb_test.go
Hydrocharged: It’s what we did for v1 to v2, so I’m just mimicking that behavior.

Hydrocharged:
Changed the logic to using a new system table, so take another look!

bats/sql.bats
```diff
@@ -848,14 +847,49 @@ SQL
...
 }
 
-@test "dolt_version() func" {
+@test "sql: dolt_version() func" {
     SQL=$(dolt sql -q 'select dolt_version() from dual;' -r csv | tail -n 1)
     CLI=$(dolt version | cut -f 3 -d ' ')
     [ "$SQL" == "$CLI" ]
 }
+
+@test "sql: stored procedures sanity check" {
```
zachmu, line 862: Need more tests here. At least a test of drop with N>1 procedures stored. And hopefully a test of show as well.

zachmu (approved):
LG, just a couple comments.

PR #1359, opened 2021-02-21
err handling fix

bheni:


go/libraries/doltcore/env/actions/commit.go
```diff
@@ -169,10 +169,16 @@ func CommitStaged(ctx context.Context, dbData env.DbData, props CommitStagedProp
...
+		return "", err
+	}
+
+	err = rsw.ClearMerge()
+
+	if err != nil {
+		return "", err
 	}
 
 	h, err = c.HashOf()
```
reltuk, line 184: Here too, right? `h` could be `nil` if `c.HashOf()` returns `err != nil`?

bheni: nah, can't be nil... but it's more idiomatic.

reltuk (approved):
LGTM!

PR #1360, opened 2021-02-22
Diff based foreign key validation

bheni:
Previous validation implementation did full scans for every foreign key constraint.  New implementation only validates against rows that have changed from the previous commit.


andrew-wm-arthur:
The idea here seems sound, but could we diff the FK indexes directly instead of diffing the row storage and looking for changes in the FK columns?

bheni:
@andrew-wm-arthur yeah, that's better.

Hydrocharged:
Since Brian will change the logic a bit, I’ll wait to review.

andrew-wm-arthur (approved):
looks good, but we need more error-case tests

Hydrocharged (approved):
LGTM as well

PR #1367, opened 2021-02-24
fix infinite loop

bheni:
if `length = length * 2` overflowed before it was greater than the necessary size `ensureCapacity` would loop forever.  After 2 gigs we allocate a half gig at a time and if the allocation size exceeds a uint32 we panic, rather than fail later.

### DIFF
--- a/go/store/types/codec.go
+++ b/go/store/types/codec.go
@@ -359,7 +359,6 @@ type binaryNomsWriter struct {
 func newBinaryNomsWriter() binaryNomsWriter {
 	return binaryNomsWriter{make([]byte, initialBufferSize), 0}
 }
-
 func (b *binaryNomsWriter) data() []byte {
 	return b.buff[0:b.offset]
 }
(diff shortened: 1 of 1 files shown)


go/store/types/codec.go
```diff
@@ -368,19 +367,42 @@ func (b *binaryNomsWriter) reset() {
...
 	}
 
 	old := b.buff
 
-	for b.offset+n > length {
+	if minLength > math.MaxUint32 {
+		panic("overflow")
+	}
+
+	for minLength > length {
```
zachmu, line 388: These loops seem like a strange way to ensure that length is bigger than minLength up to MaxUint32. Can't we just round up to the nearest half gig that's over the requested capacity?

zachmu (approved):
LG, just one comment

go/store/types/codec.go
bheni: the bit twiddling approach to rounding to the next power of 2 isn't readable, and this isn't happening in a tight loop where this needs to be super fast so leaving the doubling loop.

If you a good way to round to the next half gig that works, feel free to suggest it.  But I don't think it's more readable, and perf isn't a concern.

PR #1376, opened 2021-02-28
fix overflow issue

bheni:
Merges were failing when the changes were sufficiently large.  This was really exacerbated by merges where the common ancestor is extremely old.  getCompositeChildSequence would try to allocate contiguous blocks and encode the entire composite child sequence into it.  If the size of the data exceeded the max value of a uint32 then the indexes being used to access the buffers would overflow and wrap around and cause all kinds of badness.

### DIFF
--- a/go/store/types/codec.go
+++ b/go/store/types/codec.go
@@ -356,9 +356,21 @@ type binaryNomsWriter struct {
 	offset uint32
 }
 
+func newBinaryNomsWriterWithSizeHint(sizeHint uint64) binaryNomsWriter {
+	size := uint32(initialBufferSize)
+	if sizeHint >= math.MaxUint32 {
+		size = math.MaxUint32
+	} else if sizeHint > uint64(size) {
+		size = uint32(sizeHint)
+	}
+
+	return binaryNomsWriter{make([]byte, size), 0}
+}
+
 func newBinaryNomsWriter() binaryNomsWriter {
 	return binaryNomsWriter{make([]byte, initialBufferSize), 0}
 }
+
 func (b *binaryNomsWriter) data() []byte {
 	return b.buff[0:b.offset]
 }
@@ -374,7 +386,7 @@ const (
 
 func (b *binaryNomsWriter) ensureCapacity(n uint32) {
 	length := uint64(len(b.buff))
-	minLength := uint64(b.offset + n)
+	minLength := uint64(b.offset) + uint64(n)
 	if length >= minLength {
 		return
 	}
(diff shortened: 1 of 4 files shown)


go/store/types/codec.go
```diff
@@ -374,7 +386,7 @@ const (
 
 func (b *binaryNomsWriter) ensureCapacity(n uint32) {
 	length := uint64(len(b.buff))
-	minLength := uint64(b.offset + n)
```
bheni, line 377: this was overflowing, and then being cast to uint64.  Need to cast before adding.

go/store/types/map_leaf_sequence.go
```diff
@@ -194,41 +200,51 @@ func (ml mapLeafSequence) entries() (mapEntrySlice, error) {
 }
 
 func (ml mapLeafSequence) getCompareFn(other sequence) compareFn {
```
bheni, line 202: the diff git shows of this function sucks.  The existing logic is unchanged but put into the else block.  The if block is new.

go/store/types/codec.go
```diff
@@ -356,9 +356,21 @@ type binaryNomsWriter struct {
 	offset uint32
 }
 
+func newBinaryNomsWriterWithSizeHint(sizeHint uint64) binaryNomsWriter {
+	size := uint32(initialBufferSize)
+	if sizeHint >= math.MaxUint32 {
+		size = math.MaxUint32
+	} else if sizeHint > uint64(size) {
+		size = uint32(sizeHint)
```
andrew-wm-arthur, line 364: This seems like an explicit overflow

go/store/types/map_leaf_sequence.go
```diff
@@ -350,3 +367,216 @@ func (ml mapLeafSequence) getValue(idx int) (Value, error) {
...
+func (mes mapEntrySequence) typeOf() (*Type, error) {
+	panic("implement me")
+}
+
+func (mes mapEntrySequence) WalkRefs(nbf *NomsBinFormat, cb RefCallback) error {
+	panic("implement me")
+}
+
+func (mes mapEntrySequence) writeTo(writer nomsWriter, format *NomsBinFormat) error {
+	panic("implement me")
```
andrew-wm-arthur, line 561: AKAIK we're not using `getCompositeChildSequence` on the write path anywhere. I believe we'll panic here if we try to.

andrew-wm-arthur (approved):
lgtm

go/store/types/map_leaf_sequence.go
bheni: yeah, Goland auto filled the implementation for me.  Changed the panic methods as I'm not planning on implementing.

PR #1386, opened 2021-03-05
Added file names to bats tests

Hydrocharged:
Also edited some of the titles to cut down on redundancy since the file names are now in the title. This makes it far easier to see what test belongs in what file if a test ever fails.


timsehn:
I like it.

timsehn (approved):

PR #1402, opened 2021-03-08
diff between a range of values

bheni:
allows for performance improvements.  Not used in dolt yet, but used by the bounty attribution code.

### DIFF
--- a/go/libraries/doltcore/diff/async_differ.go
+++ b/go/libraries/doltcore/diff/async_differ.go
@@ -17,7 +17,6 @@ package diff
 import (
 	"context"
 	"fmt"
-	"math"
 	"time"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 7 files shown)


go/store/diff/diff.go
```diff
@@ -105,6 +105,27 @@ type differ struct {
...
 //    }
 func Diff(ctx context.Context, v1, v2 types.Value, dChan chan<- Difference, leftRight bool, descFunc ShouldDescFunc) error {
+	f := func(ctx context.Context, d differ, v1, v2 types.Value) error {
+		return d.diff(ctx, nil, v1, v2)
+	}
+
+	return diff(ctx, f, v1, v2, dChan, leftRight, descFunc)
+}
+
+func DiffMapRange(ctx context.Context, m1, m2 types.Map, start types.Value, inRange func(types.Value) (bool, error), dChan chan<- Difference, leftRight bool, descFunc ShouldDescFunc) error {
```
andrew-wm-arthur, line 115: I think a `type ValueInRange func(types.Value) (boot, error)` would help readability

go/libraries/doltcore/diff/async_differ.go
```diff
@@ -59,18 +60,20 @@ type AsyncDiffer struct {
...
 	return &AsyncDiffer{
-		make(chan diff.Difference, bufferedDiffs),
-		bufferedDiffs,
-		nil,
-		context.Background(),
-		func() {},
+		diffChan:   make(chan diff.Difference, bufferedDiffs),
+		bufferSize: bufferedDiffs,
+		egCtx:      context.Background(),
+		egCancel:   func() {},
```
andrew-wm-arthur, line 75: do we need to initialize these?

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/diff/async_differ.go
bheni: I just kept this because it was there?  I assume it's being called later  without being nil checked.

go/store/diff/diff.go
bheni: good call.

Issue #1389, opened 2021-03-06
stray characters in `dolt diff` output


tybug:
`dolt diff` is erroneously inserting the string `(MISSING)` in some columns of some rows.

* clone https://www.dolthub.com/repositories/tybug/national-course-catalog-us/doc/stray-characters-repro
* `dolt diff HEAD HEAD^1 > diff.diff`
* look for lines like `Blended/Hybrid up to 49%!W(MISSING)eb` - the actual content of this line is `Blended/Hybrid up to 49% Web`, as you can see with eg the following query: `select * from sections where section='82591' and course_id='ACCT 225'`.

timsehn:
Huh. We had another bug report of that this and I thought @Brian Hendriks
<brian@dolthub.com> fixed it. Thanks for the repro. We'll get this fixed
for the next release.

PR #1406, opened 2021-03-09
diff printing fix

bheni:
Columns with % signs were being printed using printf style functions which were looking at %s and escape characters.

### DIFF
--- a/bats/sql-shell.bats
+++ b/bats/sql-shell.bats
@@ -63,6 +63,9 @@ CREATE TABLE test2 (
   PRIMARY KEY (str)
 );
 SQL
+  dolt add .
+  dolt commit -m "created table"
+
   TESTSTR='0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ`~!@#$%^&*()){}[]/=?+|,.<>;:_-_%d%s%f'
   dolt sql -q "INSERT INTO test2 (str) VALUES ('$TESTSTR')"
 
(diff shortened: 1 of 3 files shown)


zachmu (approved):
bats test?

timsehn:
The bats test for this last time was pretty clever. Having a hard time locating it.

bheni:
@tim found the test:

```go
@test "validate string formatting" {
      dolt sql <<SQL
CREATE TABLE test2 (
  str varchar(256) NOT NULL,
  PRIMARY KEY (str)
);
SQL
  TESTSTR='0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ`~!@#$%^&*()){}[]/=?+|,.<>;:_-_%d%s%f'
  dolt sql -q "INSERT INTO test2 (str) VALUES ('$TESTSTR')"

  run dolt sql -q "SELECT * FROM test2"
  [ $status -eq 0 ]
  [[ "$output" =~ "$TESTSTR" ]] || false

  run dolt sql -q "SELECT * FROM test2" -r csv
  [ $status -eq 0 ]
  [[ "$output" =~ "$TESTSTR" ]] || false

  run dolt sql -q "SELECT * FROM test2" -r json
  [ $status -eq 0 ]
  [[ "$output" =~ "$TESTSTR" ]] || false
}
```

timsehn:
`dolt diff` at the end and we good. I was grepping for MISSING in the bats
directory.

bats/sql-shell.bats
```diff
@@ -77,4 +80,12 @@ SQL
   run dolt sql -q "SELECT * FROM test2" -r json
   [ $status -eq 0 ]
   [[ "$output" =~ "$TESTSTR" ]] || false
+
+  dolt add .
```
timsehn, line 84: Why not just diff the working set?

timsehn (approved):

timsehn:
This fixes: #1389

PR #1412, opened 2021-03-09
Fixed dropped errors in libraries/doltcore/sqle/dfunctions

alrs:
This fixes dropped test & code `err` variables, as well as some `err` variables that were being passed into functions and then dropped immediately.

This PR goes beyond straightforward `err` drop fixes into spookier territory.

### DIFF
--- a/go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
+++ b/go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
@@ -62,7 +62,6 @@ func (d DoltResetFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) 
 
 	// Get all the needed roots.
 	working, staged, head, err := env.GetRoots(ctx, dbData.Ddb, dbData.Rsr)
-
 	if err != nil {
 		return 1, err
 	}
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
```diff
@@ -67,8 +67,9 @@ func (d DoltResetFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
 		return 1, err
 	}
 
+	var h string
 	if apr.Contains(cli.HardResetParam) {
-		h, err := actions.ResetHardTables(ctx, dbData, apr, working, staged, head)
+		h, err = actions.ResetHardTables(ctx, dbData, apr, working, staged, head)
```
zachmu, line 72: Restore this line plz. h should be scoped to within the if block

alrs: OK. I scoped it outside the block so that I could get the `err` that was being swallowed out of the block. I'll just add an error check inside.

go/libraries/doltcore/sqle/dfunctions/dolt_reset.go
```diff
@@ -88,8 +88,13 @@ func (d DoltResetFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
...
+				return 1, err
+			}
 		} else {
-			err = setHeadAndWorkingSessionRoot(ctx, h)
+			if err := setHeadAndWorkingSessionRoot(ctx, h); err != nil {
+				return 1, err
+			}
 		}
 	} else {
 		_, err = actions.ResetSoftTables(ctx, dbData, apr, staged, head)
```
zachmu, line 100: Move the error handling block below up into here for symmetry

go/libraries/doltcore/sqle/dfunctions/hashof.go
```diff
@@ -82,6 +82,9 @@ func (t *HashOf) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 		}
 
 		cm, err = ddb.ResolveRef(ctx, branchRef)
+		if err != nil {
```
zachmu, line 85: Same comment as above. If we're going to return an err from inside conditional branch blocks, do it everywhere. 

(This one is an error because of shadowing, a problem the one in the block above doesn't have. But it's not easy to tell the difference, and we should always return errors on the line after getting one returned, not wait until after the branch logic merges together)

go/libraries/doltcore/sqle/dfunctions/merge.go
```diff
@@ -154,6 +154,10 @@ func checkForUncommittedChanges(root *doltdb.RootValue, parentRoot *doltdb.RootV
 }
 
 func getBranchCommit(ctx *sql.Context, ok bool, val interface{}, err error, ddb *doltdb.DoltDB) (*doltdb.Commit, hash.Hash, error) {
```
zachmu, line 156: This is just an error. This method should not take an error param at all.

go/libraries/doltcore/sqle/dfunctions/merge.go
```diff
@@ -182,6 +186,10 @@ func getBranchCommit(ctx *sql.Context, ok bool, val interface{}, err error, ddb
 }
 
 func getParent(ctx *sql.Context, err error, sess *sqle.DoltSession, dbName string) (*doltdb.Commit, hash.Hash, *doltdb.RootValue, error) {
```
zachmu, line 188: Same comment here.

zachmu (commented):
Thanks for the catches, little work to do on this one

zachmu (approved):

PR #1413, opened 2021-03-09
libraries/doltcore/sqle: fix dropped errors

alrs:
This fixes two dropped code `err` variables and a test `err` variable in `libraries/doltcore/sqle`.


go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1025,6 +1025,7 @@ func TestRenameColumn(t *testing.T) {
 			table, _, err := updatedRoot.GetTable(ctx, PeopleTableName)
 			assert.NoError(t, err)
 			sch, err := table.GetSchema(ctx)
+			assert.NoError(t, err)
```
zachmu, line 1028: Needs to be require

alrs: Done.

alrs:
The failure looks like a flake, to me. Could you kick-off a retest, @zachmu?

PR #1414, opened 2021-03-09
store/types: fix dropped errors

alrs:
This fixes `err` variables dropped in the `store/types` tests.


go/store/types/blob_test.go
```diff
@@ -273,6 +273,7 @@ func TestBlobConcat(t *testing.T) {
 
 	b7, b8 := split(b, 1e6/2)
 	b78, err := b7.Concat(context.Background(), b8)
+	assert.NoError(err)
```
zachmu, line 276: All of these asserts should be requires (so the test halts). Otherwise the following lines can die with nil pointer panics.

alrs: I made this change throughout the package, save for functions where `testing.T` wasn't available.

zachmu (approved):

PR #1415, opened 2021-03-09
store/datas: fix dropped errors

alrs:
This fixes dropped `err` variables in the `store/datas` package.

### DIFF
--- a/go/store/datas/tag_test.go
+++ b/go/store/datas/tag_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/dolthub/dolt/go/store/chunks"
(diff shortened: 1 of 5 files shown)


go/store/datas/tag_test.go
```diff
@@ -43,6 +43,7 @@ func TestNewTag(t *testing.T) {
 	cmRef, err := types.NewRef(commit, types.Format_7_18)
 	assert.NoError(err)
 	tag, err := NewTag(context.Background(), cmRef, types.EmptyStruct(types.Format_7_18))
+	assert.NoError(err)
```
zachmu, line 46: Needs to be require

alrs: Changed it. Want me to swap all over `tag_test.go` over to `require` instead of assert?

zachmu (approved):

PR #1421, opened 2021-03-09
Replace assert.NoError with require.NoError in tests

alrs:
This replaces most instances of `assert.NoError()` with `require.NoError()`, and picks up several dropped `err` variables in code and tests.


zachmu (approved):
Thanks for the contribution!

PR #1426, opened 2021-03-10
Max/active branch

max-hoffman:
access current_branch() through SQL


max-hoffman:
I'm having some trouble with the bats -- the queries work manually in an empty repo, but I'm not getting any error messages to point me in the right direction for the bats tests.

VinaiRachakonda:
For bats one thing you can do is run `echo $output` before the failing test to get a better sense of the output.

PR #1442, opened 2021-03-15
Added limited self referential foreign key support

Hydrocharged:
This only adds support for self referential foreign keys that consist of a single column, and also where the column with the constraint comes after the referenced column (in schema order).

### DIFF
--- a/go/libraries/doltcore/sqle/table_editor_fk_test.go
+++ b/go/libraries/doltcore/sqle/table_editor_fk_test.go
@@ -47,20 +47,32 @@ func init() {
 CREATE TABLE one (
   pk BIGINT PRIMARY KEY,
   v1 BIGINT,
-  v2 BIGINT
+  v2 BIGINT,
+  INDEX secondary (v1)
 );
 CREATE TABLE two (
   pk BIGINT PRIMARY KEY,
   v1 BIGINT,
-  v2 BIGINT
+  v2 BIGINT,
+  INDEX secondary (v1,v2)
 );
 CREATE TABLE three (
   pk BIGINT PRIMARY KEY,
   v1 BIGINT,
   v2 BIGINT
 );
-ALTER TABLE one ADD INDEX secondary (v1);
-ALTER TABLE two ADD INDEX secondary (v1,v2);
+CREATE TABLE parent (
+  id BIGINT PRIMARY KEY,
+  v1 BIGINT,
+  v2 BIGINT,
+  INDEX v1 (v1),
+  INDEX v2 (v2)
+);
+CREATE TABLE child (
+  id BIGINT PRIMARY KEY,
+  v1 BIGINT,
+  v2 BIGINT
+);
 `)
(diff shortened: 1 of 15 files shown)


Hydrocharged:
Regarding the `bats` tests, I decided to not add a `skip` for the currently unsupported ones, as the logic for those is neither straightforward nor what one would expect using the standard rules of foreign keys. Skipping tests leaves room for forgetting to unskip them when the relevant functionality is added, and I don't agree that we should take the chance as implementing it by "what makes sense" will 100% cause incorrect behavior as defined by MySQL.

go/libraries/doltcore/sqle/table_editor_fk_test.go
```diff
@@ -359,7 +371,266 @@ ALTER TABLE three ADD FOREIGN KEY (v1, v2) REFERENCES two(v1, v2) ON DELETE CASC
...
+			[]sql.Row{{1, 1, 1}, {2, 2, 1}},
+			false,
+		},
+		{
+			"DELETE FROM parent WHERE v1 = 1;",
+			[]sql.Row{},
+			true,
+		},
+		{
+			"UPDATE parent SET v1 = 2;",
```
zachmu, line 414: What about a test where you are setting the child and parent column both to the same value in the same statement?

zachmu (approved):
LG, nice work.

bats/foreign-keys.bats
```diff
@@ -1511,3 +1510,277 @@ SQL
...
+    [ "$status" -eq "1" ]
+    [[ "$output" =~ "violation" ]] || false
+    run dolt sql -q "REPLACE INTO parent VALUES (1, 1, 1);"
+    [ "$status" -eq "1" ]
+    [[ "$output" =~ "violation" ]] || false
+}
+
+@test "foreign-keys: self-referential child column precedes parent RESTRICT (SKIPPED)" {
+    # Once we implement this functionality then we need to delete this statement to let the rest of the test run, and edit the title.
+    # This ensures that the test is not ignored, unlike with a skipped test.
```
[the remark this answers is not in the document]
Hydrocharged: The idea was that these would fail if we ever implemented the logic but didn't "_unskip_" the test, so this makes it an explicit failure. We have tests that are/were skipped and they now work, and in some cases the tests are flat out wrong OR the logic implemented doesn't work according to the test because the implementor either didn't know that the test existed or forgot about it.

With this error and return 0, the PR will fail if the logic is touched and the implementor doesn't change it, which I feel is a FAR superior solution to skipped tests. Skipped tests make them optional, and assuming the tests are correct, I don't like optional things.

It's a non issue though cause I'mma rip them out anyway.

go/libraries/doltcore/sqle/table_editor_fk_test.go
Hydrocharged: In an `UPDATE`? I can add that.

PR #1443, opened 2021-03-15
[no-release-notes] /integration-tests: move mysql-client-tests

coffeegoddd:


timsehn:
We should also move `/bats` in there as well. Not sure if that's coming in another PR or forgotten. Just making sure ;-)

PR #1444, opened 2021-03-15
Updates to error handling for unset configs.

VinaiRachakonda:


bats/config.bats
```diff
@@ -164,3 +164,22 @@ teardown() {
     regex='John Doe <john@doe.com>'
     [[ "$output" =~ "$regex" ]] || false
 }
+
+@test "sql-server: starting test without server throws error" {
```
timsehn, line 168: This test name makes no sense? How about?

`sql-server: dolt sql commit() function errors appropriately when user.name or user.email is unset`

bats/config.bats
```diff
@@ -164,3 +164,22 @@ teardown() {
...
+       pk int primary key
+    )"
+
+    dolt config --global --unset user.name
+    dolt config --global --unset user.email
+
+    run dolt sql -q "SET @@dolt_repo_$$_head = COMMIT('-a', '-m', 'updated stuff')"
+    echo $output
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "Aborting commit due to empty committer name. Is your config set" ]] || false
```
timsehn, line 184: I think you should also add a test case in this test if `user.name` is set but `user.email` is not.

integration-tests/bats/config.bats
```diff
@@ -164,3 +164,27 @@ teardown() {
     regex='John Doe <john@doe.com>'
     [[ "$output" =~ "$regex" ]] || false
 }
+
+@test "COMMIT correctly errors when user.name or user.email is unset." {
```
timsehn, line 168: Start bats test with file name.

integration-tests/bats/config.bats
```diff
@@ -164,3 +164,27 @@ teardown() {
...
+    )"
+
+    dolt config --global --unset user.name
+    dolt config --global --unset user.email
+
+    run dolt sql -q "SET @@dolt_repo_$$_head = COMMIT('-a', '-m', 'updated stuff')"
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "Aborting commit due to empty committer name. Is your config set" ]] || false
+
+
```
timsehn, line 185: Don't need two spaces here.

timsehn (approved):
Lokks good!

PR #1445, opened 2021-03-15
[no-release-notes] Db/root refactor

coffeegoddd:


integration-tests/mysql-client-tests/mysql-client-tests-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+echo "Updating dolt config for tests:"
+dolt config --global --add metrics.disabled true
+dolt config --global --add metrics.host localhost
+dolt config --global --add user.name mysql-test-runner
+dolt config --global --add user.email mysql-test-runner@liquidata.co
+
+echo "Running mysql-client-tests:"
+bats /mysql-client-tests/mysql-client-tests.bats
```
zachmu, line 10: Probably need a relative path here right?

zachmu (approved):

integration-tests/mysql-client-tests/mysql-client-tests-entrypoint.sh
coffeegoddd: nah, inside the container that is the correct path

PR #1452, opened 2021-03-16
Added DELIMITER support to batch input

Hydrocharged:


integration-tests/bats/sql.bats
```diff
@@ -792,6 +792,51 @@ SQL
     rm -rf doltsql
 }
 
+@test "sql: batch delimiter" {
```
zachmu, line 795: Make sure this works for delimiters embedded in strings, e.g.

```
mysql> delimiter //
mysql> insert into mytable values (4, "//")
    -> //
```

zachmu: If it doesn't work for this use case (not clear if the quote matching code above is good enough), then fine to add a skipped test instead

zachmu (approved):
LGTM

integration-tests/bats/sql.bats
Hydrocharged: Ohh that's a good idea!

PR #1454, opened 2021-03-16
Bump go-mysql-server

andrew-wm-arthur:


go/libraries/doltcore/sqle/tables.go
```diff
@@ -106,7 +106,7 @@ func NewDoltTable(name string, sch schema.Schema, tbl *doltdb.Table, db SqlDatab
 var _ sql.Table = (*DoltTable)(nil)
 var _ sql.IndexedTable = (*DoltTable)(nil)
 var _ sql.ForeignKeyTable = (*DoltTable)(nil)
-var _ sql.StatisticsTable = (*DoltTable)(nil)
+//var _ sql.StatisticsTable = (*DoltTable)(nil)
```
zachmu, line 109: Let's wait to get Vinai's fixes merged in. We don't want to lose this interface, affects query performance

@VinaiRachakonda work with Andy to combine these two changes

PR #1455, opened 2021-03-17
Vinai/more dbeaver stuff

VinaiRachakonda:
This pr unblocks basic dbeaver functionality.


VinaiRachakonda:
This pr needs a bumped gms before review.

go/go.mod
```diff
@@ -78,4 +78,6 @@ require (
 
 replace github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi => ./gen/proto/dolt/services/eventsapi
 
+replace github.com/dolthub/go-mysql-server => /Users/vinairachakonda/go/src/dolthub/go-mysql-server
```
zachmu, line 81: Remove this, `go get github.com/dolthub/go-mysql-server@master` instead

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -115,20 +115,19 @@ func TableCacheFromSess(sess sql.Session, dbName string) TableCache {
 	return sess.(*DoltSession).caches[dbName]
 }
 
-func (sess *DoltSession) CommitTransaction(ctx *sql.Context) error {
-	currentDb := sess.GetCurrentDatabase()
-	if currentDb == "" {
-		return sql.ErrNoDatabaseSelected.New()
+func (sess *DoltSession) CommitTransaction(ctx *sql.Context, dbName string) error {
+	if dbName == "" {
```
zachmu, line 119: When does this happen?

zachmu: Leave a comment

integration-tests/bats/sql-server.bats
```diff
@@ -552,4 +552,57 @@ SQL
      "
 
      server_query 1 "SELECT * FROM test" "pk,c1,c2,c3,c4,c5\n0,1,2,3,4,5\n1,1,2,3,4,5"
+}
+
+@test "sql-server: Run queries on database without ever selecting it" {
+     skiponwindows "Has dependencies that are missing on the Jenkins Windows installation."
+
+     cd repo1
+     start_sql_server repo1
```
zachmu, line 561: Make sure this is actually testing what you think it is. These test helper methods definitely `use` a particular database one way or another. 

Would be better to pass `repo2` here, since you're operating on repo1 below.

integration-tests/bats/sql-server.bats
```diff
@@ -552,4 +552,57 @@ SQL
...
+     insert_query 1 "INSERT INTO repo1.one_pk VALUES (0), (1), (2)"
+     unselected_server_query 1 "SELECT * FROM repo1.one_pk" "pk\n0\n1\n2"
+
+     unselected_update_query 1 "UPDATE repo1.one_pk SET pk=3 WHERE pk=2"
+     unselected_server_query 1 "SELECT * FROM repo1.one_pk" "pk\n0\n1\n3"
+
+     unselected_update_query 1 "DELETE FROM repo1.one_pk WHERE pk=3"
+     unselected_server_query 1 "SELECT * FROM repo1.one_pk" "pk\n0\n1"
+
+     # We don't have transactions but some editors will throws this commit statement along
```
zachmu, line 582: ?

integration-tests/bats/helper/query-server-common.bash
```diff
@@ -161,3 +161,29 @@ update_query() {
 insert_query() {
     server_query $1 "$2" ""
 }
+
+# unselected_server_query connects to a running mysql server, but not to a particular database, executes a query and
+# compares the results against what is expected.
+# In the event that the results do not match expectations, the python process will exit with an exit code of 1
+#  * param1 is 1 for autocommit = true, 0 for autocommit = false
+#  * param2 is the query_str
+unselected_server_query() {
```
zachmu, line 170: How is this different from server_query?

integration-tests/bats/helper/query-server-common.bash
```diff
@@ -161,3 +161,29 @@ update_query() {
...
+    let PORT="$$ % (65536-1024) + 1024"
+    PYTEST_DIR="$BATS_TEST_DIRNAME/helper"
+    echo Executing server_query
+    python3 -c "$PYTHON_QUERY_SCRIPT" -- "$PYTEST_DIR" "" "$PORT" "$1" "$2" "$3"
+}
+
+# unselected_update_query runs an update query and should be called with 2 parameters
+#   * param1 is 1 for autocommit = true, 0 for autocommit = false
+#   * param2 is the query string
+unselected_update_query() {
```
zachmu, line 180: See above question, not sure what these methods are doing different

zachmu (approved):
Mostly good, just a few comments about the test logic

go/libraries/doltcore/sqle/dolt_session.go
VinaiRachakonda: This was a bug that was happening when dbeaver would run a query and throw a "commit" statement when no database was selected. For some reason even with autocommit off, dbeaver forces these transaction statements on you.

integration-tests/bats/helper/query-server-common.bash
VinaiRachakonda: This does not connect a database. The normal server_query uses $DEFAULT_DB variable which translates to `USE DB`.  Since we're not connecting to a database the engine does not have a currentDB set

integration-tests/bats/sql-server.bats
VinaiRachakonda: So each of these queries form a new connection to the server. So if start_sql_server does in fact connect to the db, the unselected_sql_query will only connect to the serve but not the db. I can add an additional statement validating that no database is selected to give more clarity

integration-tests/bats/sql-server.bats
VinaiRachakonda: Yeah that was just a sanity check that someone typing "commit" into the server doesn't mess anything up

zachmu: I mean the comment doesn't make sense, read it again

integration-tests/bats/helper/query-server-common.bash
zachmu: Ah ok, i missed that bit

integration-tests/bats/sql-server.bats
zachmu: Just use repo2.one_pk etc. I know what you're doing is right, but you don't know how the underlying code might change and make your test worthless. If want the test to be durable, run the statements on a different database than the one you're giving as an argument to start_sql_server

zachmu (approved):

PR #1456, opened 2021-03-17
store/marshal: dropped error and test refactor

alrs:
This picks up a dropped error in`store/marshal`, changes the tests to use `require.NoError` instead of `assert.NoError`, and prunes some dead code.


zachmu (approved):
Thanks for the contibution!

PR #1462, opened 2021-03-19
New select expression behavior

zachmu:


VinaiRachakonda:
Leaving a comment here so I get pinged when this is merged

max-hoffman (commented):
LGTM, sorry for late response

PR #1465, opened 2021-03-21
libraries/doltcore/doltdb: fix dropped errors

alrs:
This fixes dropped errors in `libraries/doltcore/doltdb`.

### DIFF
--- a/go/libraries/doltcore/doltdb/root_val.go
+++ b/go/libraries/doltcore/doltdb/root_val.go
@@ -274,7 +274,7 @@ func (root *RootValue) getOrCreateSuperSchemaMap(ctx context.Context) (types.Map
 	} else {
 		ssm, err = types.NewMap(ctx, root.vrw)
 	}
-	return ssm, nil
+	return ssm, err
 }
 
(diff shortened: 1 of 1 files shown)


alrs:
Am I looking at an actual Bats-failure, or a flake?

timsehn:
Seems like a flake. Sorry for not being on top of this.

@coffeegoddd do you mind kicking these off again?

timsehn:
Kicked it in the pants. @zachmu will review when/if it passes...

timsehn:
I think this may be failing for a valid reason. We've tried three times.

timsehn:
Looks like a true bug.

```
not ok 77 back-compat: table operations
# (in test file ./back-compat.bats, line 73)
#   'dolt table cp abc copy' failed
# Successfully initialized dolt data repository.
# error on line 1 for query CREATE TABLE `copy` LIKE `abc`: tag collision for columns pk and doc_name, different definitions (tag: 0)
# Error processing batch
```

zachmu:
Thanks for the catch, Lars. You caught a bug!

I skipped the test and cleaned up the error return style in https://github.com/dolthub/dolt/pull/1481. I'll merge that in once it passes CI.

Thanks again!

PR #1472, opened 2021-03-24
Vinai/merge squash

VinaiRachakonda:
Introduces SQUASH() which merges a branch into the working set.

### DIFF
--- a/go/libraries/doltcore/sqle/dfunctions/squash.go
+++ b/go/libraries/doltcore/sqle/dfunctions/squash.go
@@ -85,16 +85,16 @@ func (s SquashFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 
 	hashStr := h.String()
 
-	// Update the session and editor with the new root.
-	sess.SetRoot(dbName, hashStr, mergeRoot)
-
-	err = sess.SetEditorRoot(ctx, dbName, root)
-	if err != nil {
-		return nil, err
-	}
-
-	// Clear the cache associated with the DB.
-	sess.ClearCache(dbName)
+	//// Update the session and editor with the new root.
+	//sess.SetRoot(dbName, hashStr, mergeRoot)
+	//
+	//err = sess.SetEditorRoot(ctx, dbName, root)
+	//if err != nil {
+	//	return nil, err
+	//}
+	//
+	//// Clear the cache associated with the DB.
+	//sess.ClearCache(dbName)
 
 	return hashStr, nil
 }
--- a/go/libraries/doltcore/sqle/dfunctions/squash.go
+++ b/go/libraries/doltcore/sqle/dfunctions/squash.go
@@ -1,0 +1,114 @@
+// Copyright 2021 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dfunctions
+
+import (
+	"fmt"
+	"github.com/dolthub/dolt/go/libraries/doltcore/merge"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle"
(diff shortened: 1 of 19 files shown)


VinaiRachakonda:
#benchmark

VinaiRachakonda:
Also possible to do something like MERGE(--squash) which could write a commit

integration-tests/bats/sql-server.bats
```diff
@@ -285,6 +285,91 @@ teardown() {
...
+
+    # Validate tables and data on master
+    server_query 0 "SET @@repo1_head=hashof('master');SHOW tables" ";Table\none_pk"
+    server_query 0 "SET @@repo1_head=hashof('master');SELECT * FROM one_pk ORDER by pk" ";pk,c1,c2\n0,None,None\n1,1,None\n2,10,2\n3,3,3\n4,4,4\n5,5,5\n6,6,6\n7,7,7"
+
+    # Validate that the squash operations resulted in one commit to master than before
+    server_query 0 "select COUNT(*) from dolt_log" "COUNT(*)\n4"
+
+    # Validate the a merge commit was written by making sure the hashes of the two branches don't match
+    server_query 0 "select COUNT(hash) from dolt_branches where hash IN (select hash from dolt_branches WHERE name = 'test_branch')" "COUNT(hash)\n1"
```
timsehn, line 370: Some non-happy path testing?

timsehn (commented):
Need more non-happy path testing

go/libraries/doltcore/sqle/dfunctions/squash.go
```diff
@@ -0,0 +1,128 @@
...
+	if !ok {
+		return nil, sql.ErrDatabaseNotFound.New(dbName)
+	}
+
+	root, ok := sess.GetRoot(dbName)
+	if !ok {
+		return nil, sql.ErrDatabaseNotFound.New(dbName)
+	}
+
+	parent, _, parentRoot, err := getParent(ctx, sess, dbName)
```
bheni, line 61: getParent and sess.GetParentCommit are terrible names.  Should be Head not Parent right?  I mean I know I wrote them... but I did it poorly.  Should fix these to be in line the the terminology we've adopted.

go/libraries/doltcore/sqle/dfunctions/squash.go
```diff
@@ -0,0 +1,128 @@
...
+	// Update the session and editor with the new root.
+	sess.SetRoot(dbName, hashStr, mergeRoot)
+
+	err = sess.SetEditorRoot(ctx, dbName, root)
+	if err != nil {
+		return nil, err
+	}
+
+	// Clear the cache associated with the DB.
+	sess.ClearCache(dbName)
```
bheni, line 97: Why does this function mutate the session?  Like if you `SELECT Squash("branch")` why does that change anything?  State should only be mutated when the result is assigned to `@@dbname_working`.  Shouldn't ever be mutating state from a read query.

I think all this should get deleted, and you should just return hashStr.

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -225,6 +225,25 @@ func (sess *DoltSession) GetRoot(dbName string) (*doltdb.RootValue, bool) {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// ClearCache clears the cache for a particular root
+func (sess *DoltSession) ClearCache(dbName string) {
+	sess.caches[dbName].Clear()
+}
```
bheni, line 245: I think all this stuff goes away if you make the changes above.

integration-tests/bats/sql-server.bats
```diff
@@ -285,6 +285,103 @@ teardown() {
...
+    SET @@repo1_head=commit('-m', 'Change c 1 to 10');
+    UPDATE dolt_branches SET hash = @@repo1_head WHERE name= 'master';
+
+    SET @@repo1_head=hashof('test_branch');
+    INSERT INTO one_pk (pk,c1,c2) VALUES (6,6,6);
+    SET @@repo1_head=commit('-m', 'add 6');
+    INSERT INTO one_pk (pk,c1,c2) VALUES (7,7,7);
+    SET @@repo1_head=commit('-m', 'add 7');
+    INSERT INTO dolt_branches (name,hash) VALUES ('test_branch', @@repo1_head);"
+
```
bheni, line 355: maybe validate `SELECT SQUASH('test_branch');` doesn't mutate state.

go/libraries/doltcore/sqle/dfunctions/squash.go
timsehn: This comment is going on @bheni 's tombstone.

bheni (approved):
lgtm

PR #1474, opened 2021-03-24
go/libraries/doltcore/sqle/async_indexed_lookups.go: Fix leaking of results from one index lookup to another.

reltuk:
When a query with an index lookup as a LIMIT, it's possible for the result
iterator to be closed before all the index results have been streamed. When
that happens, we can end up with queueRows goroutine which is enqueuing rows
into a ring buffer that is going to be used by a different query later.

This PR adds two mechanisms, one to stop the wrong results from being seen and
to cancel the unnecessary work.

1. RingBuffer gets an epoch, which is incremented on each Reset(). If a given
   Push does not match the current epoch, then nothing is added to the ring
   buffer.

2. queueRows is run with a cancelable (sub-)Context, and that Context is
   canceled when the corresponding iterator is Closed().

### DIFF
--- a/go/libraries/utils/async/ring_buffer.go
+++ b/go/libraries/utils/async/ring_buffer.go
@@ -26,6 +26,7 @@ type RingBuffer struct {
 	allocSize int
 
 	closed    bool
+	epoch     int
 	headPos   int
 	tailPos   int
 	headSlice int
@@ -47,7 +48,7 @@ func NewRingBuffer(allocSize int) *RingBuffer {
 }
 
 // Reset clears a ring buffer so that it can be reused
-func (rb *RingBuffer) Reset() {
+func (rb *RingBuffer) Reset() int {
 	rb.cond.L.Lock()
 	defer rb.cond.L.Unlock()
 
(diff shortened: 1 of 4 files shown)


go/libraries/utils/async/ring_buffer_test.go
```diff
@@ -39,7 +39,7 @@ func TestSingleThread(t *testing.T) {
 			rb := NewRingBuffer(test.allocSize)
 
 			for i := 0; i < test.numItems; i++ {
-				err := rb.Push(i)
+				err := rb.Push(i, rb.epoch)
 				assert.NoError(t, err)
```
bheni, line 43: also `rb.Push(i, rb.epoth+1)` to make sure they are being filtered properly?

go/libraries/utils/async/ring_buffer.go
```diff
@@ -81,13 +85,16 @@ func (rb *RingBuffer) Close() error {
...
-func (rb *RingBuffer) Push(item interface{}) error {
+func (rb *RingBuffer) Push(item interface{}, epoch int) error {
 	rb.cond.L.Lock()
 	defer rb.cond.L.Unlock()
 
 	if rb.closed {
 		return os.ErrClosed
 	}
+	if epoch != rb.epoch {
+		return nil
```
bheni, line 96: If this is general purpose, may want to return `ErrBadEpoch`.  In the case of index lookups can do a check and ignore `ErrBadEpoch`

bheni (approved):
couple suggestions you can take them or leave them.  LGTM

go/libraries/utils/async/ring_buffer_test.go
reltuk: Added a standalone test for some epoch behavior.

go/libraries/utils/async/ring_buffer.go
reltuk: Good idea. Added.

PR #1475, opened 2021-03-24
Vinai/log bad rows

VinaiRachakonda:
This pr adds the functionality of logging bad rows to the cli. Since these rows are logged StdErr the rows can be sent to a file as well.


go/cmd/dolt/pprint/prettyprint.go
```diff
@@ -26,13 +26,80 @@ import (
...
+		return err
+	}
+
+	if okResult, ok := row[0].(sql.OkResult); ok {
+		rowNoun := "row"
+		if okResult.RowsAffected != 1 {
+			rowNoun = "rows"
+		}
+
+		fmt.Fprintf(color.Output, "Query OK, %d %s affected\n", okResult.RowsAffected, rowNoun)
```
VinaiRachakonda, line 89: Converted from cli.Print to Fmt.Fprint

go/libraries/doltcore/mvdata/data_mover.go
```diff
@@ -114,11 +119,23 @@ func (imp *DataMover) Move(ctx context.Context) (badRowCount int64, err error) {
...
 		}
+		r := pipeline.GetTransFailureRow(trf)
+
+		if r != nil {
+			sqlRow, err := sqlutil.DoltRowToSqlRow(r, sch)
+			if err != nil {
+				return true
+			}
+			mu.Lock()
+			badRows = append(badRows, sqlRow)
```
VinaiRachakonda, line 136: @zachmu This is the offender right here?

integration-tests/bats/import-create-tables.bats
```diff
@@ -529,3 +529,20 @@ DELIM
...
+
+    run dolt table import -c --continue --pk=pk test 1pk5col-rpt-ints.csv
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Rows Processed: 1, Additions: 1, Modifications: 0, Had No Effect: 0" ]] || false
+    [[ "$output" =~ "The following rows were skipped:" ]] || false
+    [[ "$output" =~ "pk,c1,c2,c3,c4,c5" ]] || false
+    [[ "$output" =~ "1,1,2,3,4,7" ]] || false
+    [[ "$output" =~ "Lines skipped: 1" ]] || false
+    [[ "$output" =~ "Import completed successfully." ]] || false
+}
```
timsehn, line 548: Maybe make sure it doesn't log anything when you only send good rows?

timsehn (commented):
Additional Bats test suggestion

integration-tests/bats/import-create-tables.bats
VinaiRachakonda: Yeah I edited a previous test to validate that but I can add more

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -380,10 +380,10 @@ func (cmd ImportCmd) Exec(ctx context.Context, commandStr string, args []string,
 	skipped, verr := mvdata.MoveData(ctx, dEnv, mover, mvOpts)
 
 	if skipped > 0 {
-		cli.PrintErrln(color.YellowString("Lines skipped: %d", skipped))
+		cli.Println(color.YellowString("Lines skipped: %d", skipped))
```
zachmu, line 383: This sends text to STDOUT instead of STDERR. Intentional?

go/cmd/dolt/pprint/prettyprint.go
zachmu: color.Output is not what you want here, or elsewhere. You need to use cli.Print methods to print. Otherwise this code is going to behave differently from other printing code when we change things in the CLI print methods.

go/libraries/doltcore/mvdata/data_mover.go
```diff
@@ -170,6 +187,22 @@ func MoveDataToRoot(ctx context.Context, mover *DataMover, mvOpts DataMoverOptio
 		return nil, badCount, errhand.BuildDError("An error occurred moving data:\n").AddCause(err).Build()
 	}
 
+	if len(badRows) > 0 {
```
zachmu, line 190: If all you're doing with bad rows is printing them out, just do that in the callback. No need to return a slice of all of them here

zachmu (commented):
Couple comments on this one

go/cmd/dolt/commands/tblcmds/import.go
VinaiRachakonda: Yea, that way only the file contents themselves are sent to the stderr descriptor. Thoughts on this?

zachmu: This is the total opposite of what should happen

"main" program output gets printed to STDOUT and auxiliary status info / error messages get printed to STDERR

Are you claiming that MoveData prints table info to STDERR? Because I don't believe you

go/libraries/doltcore/mvdata/data_mover.go
```diff
@@ -144,20 +164,56 @@ func (imp *DataMover) Move(ctx context.Context) (badRowCount int64, err error) {
...
+	return nil
+}
+
 func MoveDataToRoot(ctx context.Context, mover *DataMover, mvOpts DataMoverOptions, root *doltdb.RootValue, updateRoot func(c context.Context, r *doltdb.RootValue) error) (*doltdb.RootValue, int64, errhand.VerboseError) {
 	var badCount int64
 	var err error
 	newRoot := &doltdb.RootValue{}
 
-	badCount, err = mover.Move(ctx)
+	badCount, err = mover.Move(ctx, mover.Wr.GetSchema())
```
zachmu, line 208: Still need this extra param? Probably shouldn't, right?

go/libraries/doltcore/sqle/sqlutil/sql_row.go
```diff
@@ -173,3 +175,47 @@ func keylessDoltRowFromSqlRow(ctx context.Context, vrw types.ValueReadWriter, sq
...
+			return typedCol
+		case bool:
+			if typedCol {
+				return "true"
+			} else {
+				return "false"
+			}
+		case time.Time:
+			return typedCol.Format("2006-01-02 15:04:05.999999 -0700 MST")
+		}
```
zachmu, line 217: Should have a default that does fmt.Sprintf("%v")

zachmu (approved):
Generally looks good, but I'm a little skeptical of the stdio changes

Brian should take a quick peek

go/libraries/doltcore/mvdata/data_mover.go
VinaiRachakonda: No it's needed to print out the actual row.

go/cmd/dolt/commands/tblcmds/import.go
VinaiRachakonda: No I'm claiming that bad rows that do not get imported into the database are sent into STDERR. I can also print them in STDOUT but then the user does not have the option to get a neat file piped from STDERR. Ultimately, either solution is fine with me. I think the users benefits more with this approach

PR #1476, opened 2021-03-24
bump dolt version

VinaiRachakonda:


zachmu (approved):
Can you fix the release script to do this automatically? Kind of important

VinaiRachakonda:
@zachmu Sure will do

PR #1479, opened 2021-03-25
Add few doltpy commands that broke with dolt compatibility changes

max-hoffman:
Some bats that if failed indicate a doltpy incompatibility. The tests could be more involved, but the goal here isn't row-output correctness.

This list will grow as I rewrite more dolt commands to use SQL features.

### DIFF
--- a/integration-tests/bats/doltpy.bats
+++ b/integration-tests/bats/doltpy.bats
@@ -26,7 +26,7 @@ teardown() {
 # "SELECT * FROM `{table}` ASOF '{commit}'"
 
 # verifying the output comparison to dolt log would be nice
-@test "doltpy: Dolt.log" {
+@test "doltpy: log query returns expected header names" {
     run dolt sql -r csv <<SQL
 SELECT
     dc.commit_hash as commit_hash,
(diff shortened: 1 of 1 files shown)


integration-tests/bats/doltpy.bats
```diff
@@ -0,0 +1,62 @@
...
+}
+
+# NOTE: if these break on a release that's OK, just tell Max/Oscar
+
+# already covered by other tests:
+# "select * from dolt_branches"
+# "SELECT * FROM `{table}` ASOF '{commit}'"
+
+# verifying the output comparison to dolt log would be nice
+@test "doltpy: Dolt.log" {
```
timsehn, line 29: I think you should put the assertion in the test name. Like:

"dolt log returns proper named headers"

integration-tests/bats/doltpy.bats
```diff
@@ -0,0 +1,62 @@
...
+    message as message
+FROM
+    dolt_log AS dc
+    LEFT OUTER JOIN dolt_commit_ancestors AS dca
+        ON dc.commit_hash = dca.commit_hash
+    WHERE dc.commit_hash = HASHOF('HEAD')
+    ORDER BY date DESC
+    LIMIT 1;
+SQL
+    echo $status
```
timsehn, line 46: Get rid of theses. They are trapped if the test passes and you get output if it fails. echos are useless.

timsehn (commented):
Couple comments.

integration-tests/bats/doltpy.bats
max-hoffman: fixing

timsehn:
LGTM

PR #1487, opened 2021-03-27
Properly reverts the version string on Dolt as automation does the fix

VinaiRachakonda:
One day I will get automation right on the first try


oscarbatori:
LGTM

PR #1494, opened 2021-03-30
Fix dropped errors in libraries/doltcore/doltdocs

alrs:
Fixes dropped errors, changes tests to use `require.NoError()`.


zachmu (approved):
Thanks for the contribution!

PR #1498, opened 2021-03-30
Vinai/dolt checkout period

VinaiRachakonda:
Adds `dolt checkout .` functionality.


integration-tests/bats/misc.bats
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
+
+# Used for misc. tests that don't really fix anywhere else.
```
zachmu, line 4: No misc files in bats, these are already kind of a mess

A good place for this might be status.bats

zachmu (approved):
LG

PR #1499, opened 2021-03-31
/go/libraries/doltcore/dtestutils: use Format_LD_1 in CreateTestEnv

andrew-wm-arthur:
Switches to using `Format_LD_1` instead of `Format_7_18` in `CreateTestEnv()`. This covers most `doltcore` tests, including `enginetests`. 

One major difference between `LD_1` and `7_18` is `types.Float` serialization format.


bheni (approved):
lgtm

PR #1502, opened 2021-04-01
libraries/doltcore/rowconv: error handling

alrs:
This fixes dropped errors in `libraries/doltcore/rowconv` and replaces `assert.NoError()` with `require.NoError()` in tests.


zachmu (approved):
LGTM

PR #1503, opened 2021-04-01
/go: Bump GMS

andrew-wm-arthur:
- bump go-mysql-server 
- skip 2 queries
- ~delete stale Noms perf tests~
- updated dependency `github.com/shirou/gopsutil`


zachmu (approved):
LG, need to look into the two new failing queries

PR #1504, opened 2021-04-01
Andy/json diff merge

andrew-wm-arthur:
Implements diff and merge on the JSON type. The JSON type is still behind a feature flag


go/cmd/dolt/commands/sql_print.go
```diff
@@ -276,8 +276,13 @@ func getJSONProcessFunc(sch sql.Schema) pipeline.StageFunc {
...
 
 					validCols++
-					colStr := sqlutil.SqlColToStr(col)
-					colStr = strings.Replace(colStr, "\"", "\\\"", -1)
+					colStr := sqlutil.SqlColToStr(ctx, col)
+
+					if _, ok := col.(sql.JSONValue); !ok {
+						// don't escape for JSONValue literals
+						colStr = strings.Replace(colStr, "\"", "\\\"", -1)
+					}
```
bheni, line 284: Might want to break this out as a utility function that others can reuse rather than repeating strings.Replace in other places

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -28,6 +28,7 @@ import (
 
 func init() {
 	sqle.MinRowsPerPartition = 2
+	json.FeatureFlag = true
```
bheni, line 31: If the default value is true and there is no way to set it to false it may be time to kill the flag.  That or default it to false and require --json

bheni (approved):

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
andrew-wm-arthur: It defaults to false [here](https://github.com/dolthub/dolt/blob/61356c11fbce93d1d225588f6ac4fa9bcc3b6f68/go/libraries/doltcore/sqle/json/noms_json_value.go#L27) and there is a BATS test to assert that it's not available to users.
It only gets set true here for the duration of these unit tests.

PR #1507, opened 2021-04-02
Added DELIMITER newline termination

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1495
`DELIMITER` had to be terminated by the previous delimiter, similar to other statements. However, MySQL does not have this requirement (a newline is adequate for termination). This PR removes this requirement.


zachmu (approved):
LGTM, nice catch

PR #1513, opened 2021-04-05
/go/libraries/doltcore/sqle/{integration_test, json}: Test large JSON docs

andrew-wm-arthur:
Final checks before releasing JSON: 
- Round trip large JSON documents
- Assert JSON values are structurally sharing


bheni (approved):
lgtm

PR #1514, opened 2021-04-05
/go/libraries/doltcore/schema/typeinfo: Removing feature flag for JSON type

andrew-wm-arthur:
Removes feature flag for `json` type, making it available to users.


bheni (approved):
lgtm

PR #1519, opened 2021-04-06
Added generated partial indexes for multi-column indexes

Hydrocharged:
This adds partial indexes to composite (multi-column) indexes. That means that `INDEX (v1, v2, v3)` now allows for partial matches, so accessing `v1` and `v1, v2` will make use of the index.

### DIFF
--- a/go/libraries/doltcore/sqle/tables.go
+++ b/go/libraries/doltcore/sqle/tables.go
@@ -155,7 +155,24 @@ func (t *DoltTable) GetIndexes(ctx *sql.Context) ([]sql.Index, error) {
 			tableName:    t.Name(),
 			tableSch:     sch,
 			unique:       true,
+			generated:    false,
 		})
+		for i := 1; i < len(cols); i++ {
+			sqlIndexes = append(sqlIndexes, &doltIndex{
+				cols:         cols[:i],
+				db:           t.db,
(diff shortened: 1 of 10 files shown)


go/libraries/doltcore/sqle/tables.go
```diff
@@ -179,7 +196,24 @@ func (t *DoltTable) GetIndexes(ctx *sql.Context) ([]sql.Index, error) {
...
+				cols:         cols[:i],
+				db:           t.db,
+				id:           fmt.Sprintf("%s_PARTIAL_%d", index.Name(), i),
+				indexRowData: indexRowData,
+				indexSch:     index.Schema(),
+				table:        tbl,
+				tableData:    rowData,
+				tableName:    t.Name(),
+				tableSch:     sch,
+				unique:       index.IsUnique(),
```
zachmu, line 212: Probably better to say these are all non-unique... it shouldn't matter one way or another right now, but we definitely can't guarantee there will be at most one row for any equality lookup, which is the case for normal unique keys.

zachmu (approved):
Looks great

go/libraries/doltcore/sqle/tables.go
Hydrocharged: Ohh that’s true. I’ll fix it.

PR #1525, opened 2021-04-08
Andy/faster tag validation

andrew-wm-arthur:
Previously we fully validated column tag uniqueness on every call to `doltdb.PutTable()`. This is a huge perf hit on the write path. We can safely eliminate this validation in the case where the old and new table schemas have the same `types.Ref`.


zachmu (approved):
LGTM

andrew-wm-arthur:
#benchmark

andrew-wm-arthur:
#benchmark

PR #1527, opened 2021-04-08
libraries/doltcore/fkconstrain: fix dropped errors

alrs:
This fixes two dropped error variables in `libraries/doltcore/fkconstrain`.


zachmu (approved):
LGTM

PR #1528, opened 2021-04-08
Fixed some VARBINARY strings reading their length incorrectly

Hydrocharged:
Some `VARBINARY` strings were misreading their length, leading to other values reading garbage and erroring. Not sure how some `VARBINARY` strings were reading correctly, but this fixes the issue. Was caught by the fuzzer, but I stripped out the other queries for the bats test (still reproduces without them).

### DIFF
--- a/go/store/types/tuple.go
+++ b/go/store/types/tuple.go
@@ -654,8 +654,15 @@ func (t Tuple) TupleLess(nbf *NomsBinFormat, otherTuple Tuple) (bool, error) {
 			dec.offset += 1
 			otherDec.offset += 1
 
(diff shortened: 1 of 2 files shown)


timsehn:
This is beautiful.

bheni (approved):
lgtm

Surprised this wasn't found earlier.  Seems like it would have led to some real nasty results.

Hydrocharged:
> lgtm
> 
> Surprised this wasn't found earlier. Seems like it would have led to some real nasty results.

I agree, it seems like this should have been caught way earlier!

PR #1532, opened 2021-04-08
couple changes for aws errors

bheni:


go/libraries/doltcore/dbfactory/aws.go
```diff
@@ -115,7 +115,7 @@ func (fact AWSFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat, u
 
 	db = datas.NewDatabase(cs)
 
-	return db, err
+	return db, nil
```
bheni, line 118: returning this error was already handled

Issue #1491, opened 2021-03-29
insert select auto increment error


richardji7:
insert into table select with auto increment receive an error.  Please see test case below.
```
test_db> desc test_old;
+-------+--------------+------+-----+---------+-------+
| Field | Type         | Null | Key | Default | Extra |
+-------+--------------+------+-----+---------+-------+
| name  | varchar(100) | NO   |     |         |       |
+-------+--------------+------+-----+---------+-------+
test_db> desc test;
+-------+--------------+------+-----+---------+----------------+
| Field | Type         | Null | Key | Default | Extra          |
+-------+--------------+------+-----+---------+----------------+
| id    | int          | NO   | PRI |         | auto_increment |
| name  | varchar(100) | NO   |     |         |                |
+-------+--------------+------+-----+---------+----------------+
test_db> select * from test_old;
+------+
| name |
+------+
| abc  |
| bbb  |
| ccc  |
+------+
test_db> insert into test (name) select name from test_old;
*expression.AutoIncrement is unsupported for inserts
```

VinaiRachakonda:
@richardji7 Great bug, we'll get on it!

PR #1534, opened 2021-04-09
/go/libraries/doltcore/sqle: Fix `SELECT INTO ...` bug for `AUTO_INCREMENT` columns

andrew-wm-arthur:
Fixes: #1491


zachmu (approved):
LG

PR #1536, opened 2021-04-12
libraries/doltcore/tables: fix two subpackage dropped errors

alrs:


alrs:
I don't have a Mac, so there's not much I can do to follow up. Could we re-run this once in case it's a flake?

zachmu (approved):

PR #1537, opened 2021-04-12
Store check constraints in schema

zachmu:
This is missing:

1) Bats tests
2) Feature version bump


zachmu:
Also this depends on interface changes on https://github.com/dolthub/go-mysql-server/pull/368, won't pass tests until then

go/libraries/doltcore/schema/check_coll.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type checkCollection struct {
+	checks []check
+}
+
+func (c *checkCollection) AddCheck(name, expression string, enforce bool) (Check, error) {
+	for _, chk := range c.checks {
+		if strings.ToLower(name) == strings.ToLower(chk.name) {
+			return nil, fmt.Errorf("name %s in use", name)
```
Hydrocharged, line 65: I'd add an error in `go-mysql-server` and return that instead

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -203,7 +203,7 @@ func TestAddColumn(t *testing.T) {
 }
 
 func TestModifyColumn(t *testing.T) {
-	t.Skip("Type changes aren't supported")
+	t.Skip("Type changes not supported")
```
Hydrocharged, line 206: These should work though? Removing this skip doesn't allow this to pass?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1422,3 +1444,129 @@ func (t *AlterableDoltTable) updateFromRoot(ctx *sql.Context, root *doltdb.RootV
...
+func (t *AlterableDoltTable) generateCheckName(ctx *sql.Context, check *sql.CheckDefinition) (string, error) {
+	var bb bytes.Buffer
+	bb.Write([]byte(check.CheckExpression))
+	hash := hash.Of(bb.Bytes())
+
+	hashedName := fmt.Sprintf("chk_%s", hash.String()[:8])
+	name := hashedName
+
+	var i int
+	for {
```
Hydrocharged, line 1533: You're not incrementing `i` in this loop

integration-tests/bats/sql.bats
```diff
@@ -1027,3 +1027,54 @@ SQL
...
+    run dolt sql -q "insert into t2 values (3, 4)"
+    [ $status -eq 1 ]
+    [[ "$output" =~ "constraint" ]] || false
+
+    run dolt sql -q "insert into t2 values (4, 2)"
+    [ $status -eq 1 ]
+    [[ "$output" =~ "constraint" ]] || false
+
+    dolt sql <<SQL
+ALTER TABLE t2 DROP CONSTRAINT chk1;
```
Hydrocharged, line 1069: Maybe drop one, then check, then drop the other and check? Otherwise a future change could make it where dropping one wipes them all.

Hydrocharged (approved):
Overall looks good! Just a few comments

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
zachmu: No, tried it. I think it changes a primary key column, but I didn't dig in much.

integration-tests/bats/sql.bats
zachmu: Good call

go/libraries/doltcore/schema/check_coll.go
zachmu: Server already does this check, this is just a failsafe

I'll add a comment

go/libraries/doltcore/sqle/tables.go
zachmu: Oof thanks bro

PR #1540, opened 2021-04-13
Fixed TINYBLOB errors with specific inputs

Hydrocharged:
Similar to https://github.com/dolthub/dolt/pull/1528, except dealing with the kind. Still surprised this was not caught earlier, seems like it should be completely broken. Some test repos have had thousands of rows with a `BLOB`-type column and they all worked without issue.


bheni (approved):
lgtm

PR #1545, opened 2021-04-14
tuple type changes

bheni:


bheni:
#benchmark

zachmu (approved):
image

Hydrocharged (approved):
Looks fine to me as well

timsehn:
This looks like a real nice write throughput win across the board.

andrew-wm-arthur (approved):
Seems like we should do this. My only concern is diff perf. Changing the types we serialize in meta-tuples will change the content-addresses of the internal nodes. You could have equal two subtrees that have different root addresses, one written with the old types and one written with the new. However, this will phase out over time and this change won't affect the addresses of leaf nodes, so we'll still skip over them when equal.

timsehn:
What are we doing to advance the ball here?

andrew-wm-arthur:
#benchmark

zachmu:
#benchmark

timsehn:
Let's ship this.

timsehn:
And worth a blog.

Hydrocharged:
@bheni So I've run this with merging on the fuzzer (rebased onto the `master`) and it seems good to go! I think we can finally get this into Dolt.

PR #1547, opened 2021-04-15
cmd/dolt/commands: fix dropped errors

alrs:
This fixes some dropped errors in `diff` and `ls`.


coffeegoddd:
Same here, if you can pick up `master` it should pass CI.

zachmu (approved):

Issue #1546, opened 2021-04-14
Drop then add a table with a foreign key with the same name gives spurious error


zachmu:
```sh
dolt commit -m "parent and child"
commit c409g12fguc3duci5etd5bqbjece9did
Author: Zach Musgrave <zach@liquidata.co>
Date:   Wed Apr 14 15:32:39 -0700 2021

        parent and child

zachmu@zachmu-laptop:zachmu/dolt-scratch/auto% dolt sql -q "show create table child";
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table
                                                              |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| child | CREATE TABLE `child` (
  `b` int NOT NULL,
  PRIMARY KEY (`b`),
  CONSTRAINT `rm24o4um` FOREIGN KEY (`b`) REFERENCES `parent` (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
zachmu@zachmu-laptop:zachmu/dolt-scratch/auto% dolt sql -q "drop table child";
zachmu@zachmu-laptop:zachmu/dolt-scratch/auto% dolt sql -q 'create table child (b int primary key, constraint 'rm24o4um' foreign key (b) references parent(a))'
zachmu@zachmu-laptop:zachmu/dolt-scratch/auto% dolt add .
Unknown error
cause: a foreign key with the name `rm24o4um` already exists
```

PR #1548, opened 2021-04-15
Fixed FK issue #1546

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1546


go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -257,6 +259,32 @@ func GetTableDeltas(ctx context.Context, fromRoot, toRoot *doltdb.RootValue) (de
 		})
 	}
 
+	sort.Slice(deltas, func(i, j int) bool {
```
andrew-wm-arthur, line 262: It seems somewhat fragile to have the staging logic in `MoveTablesBetweenRoots` depend on the sorting order here. See comment below

go/libraries/doltcore/env/actions/table.go
```diff
@@ -48,6 +48,7 @@ func MoveTablesBetweenRoots(ctx context.Context, tbls []string, src, dest *doltd
 			}
```
andrew-wm-arthur, line 48: What if we just split this loop into two: one loop that processes all the drops, followed by another that processes the rest?

Hydrocharged: That seems fine, I'll change it.

andrew-wm-arthur (approved):
LGTM

PR #1553, opened 2021-04-15
Vinai/foreign key checks table creation

VinaiRachakonda:
This pr adds delayed fk resolution behavior


go/libraries/doltcore/doltdb/indexeditor.go
```diff
@@ -0,0 +1,355 @@
+package doltdb
```
VinaiRachakonda, line 1: Ignore this file. I got tired of dealing with import loops I just created a new indexeditor.go. Will def remove later

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -1000,13 +1002,39 @@ func (root *RootValue) ValidateForeignKeysOnSchemas(ctx context.Context) (*RootV
...
 			if err := foreignKey.ValidateTableSchema(tblSch); err != nil {
 				return nil, err
 			}
 			parentSch, existsInRoot := allTablesSet[foreignKey.ReferencedTableName]
 			if !existsInRoot {
 				return nil, fmt.Errorf("foreign key `%s` requires the referenced table `%s`", foreignKey.Name, foreignKey.ReferencedTableName)
 			}
+
+			// Check if the foreign key needs to be regenerated
+			if foreignKey.ReferencedTableColumns == nil || foreignKey.ReferencedTableIndex == "" {
```
VinaiRachakonda, line 1015: @Hydrocharged Is this the right place to start?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1179,6 +1188,68 @@ func (t *AlterableDoltTable) CreateForeignKey(
...
+	onUpdateRefOp, err := parseFkReferenceOption(onUpdate)
+	if err != nil {
+		return err
+	}
+	onDeleteRefOp, err := parseFkReferenceOption(onDelete)
+	if err != nil {
+		return err
+	}
+
+	foreignKey := doltdb.ForeignKey{
```
VinaiRachakonda, line 1201: @Hydrocharged What do you think of this empty FK definition?

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -134,6 +138,122 @@ func (fk ForeignKey) ValidateReferencedTableSchema(sch schema.Schema) error {
...
 }
 
+// HasDelayedResolution is true when a FK is created without a referenced table index and its relative tags.
+func (fk ForeignKey) HasDelayedResolution() bool {
+	return fk.ReferencedTableIndex == "" || fk.ReferencedTableColumns == nil
+}
+
+// ResolveReferencedIndexAndTags handles delayed foreign key resolution in the case that FOREIGN_KEY_CHECKS=0 when
+// a series of create table operations are done.
+func (fk ForeignKey) ResolveReferencedIndexAndTags(ctx context.Context, root *RootValue) (*RootValue, ForeignKey, error) {
```
Hydrocharged, line 148: The function signature seems confused. If we are modifying the called foreign key, then this should be a pointer receiver. As well, since we are modifying the called key, we don't need to return it again. Either create and return a new foreign key, leaving the called one untouched, or remove the returned foreign key. Additionally, I don't think this should be modifying the root. The caller may not want to write the updated key to the root (or may want to delay it for validation, etc.) and root calls are expensive.

```golang
func (fk *ForeignKey) ResolveReferencedIndexAndTags(ctx context.Context, root *RootValue) error
```
or
```golang
func (fk ForeignKey) ResolveReferencedIndexAndTags(ctx context.Context, root *RootValue) (ForeignKey, error)
```

go/libraries/doltcore/doltdb/index_editor.go
```diff
@@ -22,21 +22,42 @@ import (
...
-	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/row"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
-	"github.com/dolthub/dolt/go/libraries/doltcore/table"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/typed/noms"
 	"github.com/dolthub/dolt/go/store/hash"
 	"github.com/dolthub/dolt/go/store/types"
 )
 
+// TableReader is an interface for reading rows from a table
```
Hydrocharged, line 32: It doesn't make sense for these interfaces to be in the `index_editor` file.

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -1000,13 +1000,38 @@ func (root *RootValue) ValidateForeignKeysOnSchemas(ctx context.Context) (*RootV
...
 			if err := foreignKey.ValidateTableSchema(tblSch); err != nil {
 				return nil, err
 			}
 			parentSch, existsInRoot := allTablesSet[foreignKey.ReferencedTableName]
 			if !existsInRoot {
 				return nil, fmt.Errorf("foreign key `%s` requires the referenced table `%s`", foreignKey.Name, foreignKey.ReferencedTableName)
 			}
+
+			// Check if the foreign key needs to be resolved
+			if foreignKey.HasDelayedResolution() {
```
Hydrocharged, line 1013: We don't do any validation if foreign key checks are disabled right? We need that check in this function.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1062,18 +1061,39 @@ func (t *AlterableDoltTable) CreateForeignKey(
...
+		// TODO: Look here to see that then when the referenced table does not exist an error is thrown
 		refTbl, _, ok, err = root.GetTableInsensitive(ctx, refTblName)
 		if err != nil {
 			return err
 		}
+
+		checksDisabled := isForeignKeyChecksDisabled(ctx)
 		if !ok {
+			if checksDisabled {
+				return t.addEmptyForeignKeyToCollection(ctx, root, fkName, tableIndex.Name(), colTags, refTblName, refColumns, onUpdate, onDelete)
```
Hydrocharged, line 1095: Can't link to it as it's not edited, but we're duplicating quite a bit of work in `ForeignKey.ResolveReferencedIndexAndTags`. Instead, you can just create the foreign key here as though it's going to be an empty one every time, and if foreign key checks are _enabled_, then we immediately resolve them.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1201,6 +1207,70 @@ func (t *AlterableDoltTable) CreateForeignKey(
...
+}
+
+func isForeignKeyChecksDisabled(ctx *sql.Context) bool {
+	_, val := ctx.Get("foreign_key_checks")
+
+	if val == nil {
+		return false
+	}
+
+	asInt := int(val.(int8))
```
Hydrocharged, line 1265: Better to use `sql.ConvertToBool` instead.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1201,6 +1207,70 @@ func (t *AlterableDoltTable) CreateForeignKey(
...
+		TableColumns:               colTags,
+		ReferencedTableName:        refTblName,
+		ReferencedTableIndex:       "",
+		ReferencedTableColumns:     nil,
+		ReferencedTableColumnNames: refTblColNames,
+		OnUpdate:                   onUpdateRefOp,
+		OnDelete:                   onDeleteRefOp,
+	}
+
+	foreignKeyCollection, err := root.GetForeignKeyCollection(ctx)
```
Hydrocharged, line 1235: With my earlier comment about first writing all foreign keys as "empty", you should remove writing to the root here and do it in the calling function.

Hydrocharged (changes requested):
I don't see why a lot of this code was deleted or moved. The index editor is now in a separate package than the other editors, which doesn't make sense, along with deleted tests and moved interfaces. You need to put the code back in its original place, unless there's a good reason for why it was moved. If there's an import cycle somewhere that I missed, then duplicate the bit of code causing the import cycle or break that specific bit into its own package. The rest of it looks good with some comments here and there, but we definitely don't need to change this many lines for this feature.

Hydrocharged:
I haven't reviewed the `bats` tests yet, as I'll wait for the changes. Once the PR has been reworked, ping me again and I'll give it another look.

VinaiRachakonda:
Closing this pr due to changing priorities

PR #1560, opened 2021-04-19
store/perf/suite: errors and tests

alrs:
This fixes a dropped error and replaces most `assert.NoError()` with `require.NoError()`.


alrs:
@andrew-wm-arthur this removes a `Skip()` that was added pretty recently. Locally the tests were passing with the skip removed. I'm happy to pull the skip-removal commit off the PR if it's there for a good reason.

zachmu (approved):

PR #1573, opened 2021-04-21
/go/libraries/doltcore/doltdb: Consolidated valid identifier regexes

andrew-wm-arthur:
now matching https://dev.mysql.com/doc/refman/8.0/en/identifiers.html


go/libraries/doltcore/doltdb/table.go
```diff
@@ -36,30 +36,23 @@ const (
...
 
-	// TableNameRegexStr is the regular expression that valid tables must match.
-	TableNameRegexStr = `^[a-zA-Z]{1}$|^[a-zA-Z]+[-_0-9a-zA-Z]*[0-9a-zA-Z]+$`
-	// ForeignKeyNameRegexStr is the regular expression that valid foreign keys must match.
-	// From the unquoted identifiers: https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
-	ForeignKeyNameRegexStr = `^[-$_0-9a-zA-Z]+$`
+	// UnquotedIdentifierRegexStr is the regular expression that valid identifiers must match.
+	// https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
+	// TODO: move this to the engine
+	UnquotedIdentifierRegexStr = `^[a-zA-Z$_]+[0-9a-zA-Z$_]*$|^[0-9a-zA-Z$_]*[a-zA-Z$_]+$`
```
zachmu, line 42: Just call this ValidSqlIdentifierRegex. It's the same for every schema object. Whether it needs to be quoted in SQL or not is not our concern at this layer.

The original ForeignKeyNameRegexStr was correct. A table name can be all numerals, just needs backtick quoting in MySQL at that point.

zachmu: Some of the tests above are incorrect in light of this

zachmu (commented):
Too strict, see comments

zachmu (approved):
Discussed offline

PR #1575, opened 2021-04-22
cmd/dolt/commands: fix dropped errors

alrs:
This fixes the straightforward-looking dropped errors in `cmd/dolt/commands` and its subpackages.

### DIFF
--- a/go/cmd/dolt/commands/fetch.go
+++ b/go/cmd/dolt/commands/fetch.go
@@ -203,11 +203,12 @@ func fetchRefSpecs(ctx context.Context, mode ref.RefUpdateMode, dEnv *env.DoltEn
 					return verr
 				}
 
+				var ok bool
 				switch mode {
 				case ref.ForceUpdate:
 					err = dEnv.DoltDB.SetHeadToCommit(ctx, remoteTrackRef, srcDBCommit)
 				case ref.FastForwardOnly:
-					ok, err := dEnv.DoltDB.CanFastForward(ctx, remoteTrackRef, srcDBCommit)
+					ok, err = dEnv.DoltDB.CanFastForward(ctx, remoteTrackRef, srcDBCommit)
 					if !ok {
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/fetch.go
```diff
@@ -203,11 +203,12 @@ func fetchRefSpecs(ctx context.Context, mode ref.RefUpdateMode, dEnv *env.DoltEn
 					return verr
 				}
 
+				var ok bool
 				switch mode {
 				case ref.ForceUpdate:
 					err = dEnv.DoltDB.SetHeadToCommit(ctx, remoteTrackRef, srcDBCommit)
 				case ref.FastForwardOnly:
-					ok, err := dEnv.DoltDB.CanFastForward(ctx, remoteTrackRef, srcDBCommit)
+					ok, err = dEnv.DoltDB.CanFastForward(ctx, remoteTrackRef, srcDBCommit)
```
zachmu, line 211: This isn't quite right

We need to handle errors in each switch case independently, because they have different semantics.

The fast forward case needs to special-case  `err == doltdb.ErrUpToDate` because that error is used as a status code by CanFastForward, and isn't a user-face error.

zachmu: This is the reason for the bats test failure.

What was here before was correct but looked wrong. You fixing it to make it look correct actually introduced a change in semantics that looked like it could have been accidental due to variable shadowing

The goal is to end up with something that both looks and is correct. Easiest way to get there is to make each switch case do its own error handling, rather than coalesce at the end.

timsehn:
Shall we close this?

alrs:
@timsehn thanks for the reminder. There aren't any conflicts yet, so I can pick this back up as-is.

timsehn:
@alrs Got a few minutes to close this out?

zachmu:
Closing in favor of https://github.com/dolthub/dolt/pull/2012

PR #1578, opened 2021-04-22
include row when duplicate key err

bheni:


go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -188,34 +195,42 @@ func (tea *tableEditAccumulator) NewFromCurrent() *tableEditAccumulator {
...
+		affectedKeys: make(map[hash.Hash]*doltKVP),
 	}
 }
 
-// Has returns whether the current tableEditAccumulator contains the given key. This assumes that the given hash is for the given
-// key.
-func (tea *tableEditAccumulator) Has(ctx context.Context, keyHash hash.Hash, key types.Value) (bool, error) {
+// doltKVPIfHas returns a *doltKVP if the current tableEditAccumulator contains the given key, or it exists in the row data.
+// This assumes that the given hash is for the given key.
+func (tea *tableEditAccumulator) doltKVPIfHas(ctx context.Context, keyHash hash.Hash, key types.LesserValuable) (*doltKVP, bool, error) {
```
zachmu, line 204: Probably just change to Get or MaybeGet

zachmu (commented):
General idea seems good, but rather than change all the signatures to include the error func, just define an error creation interface and set it on the relevant editors

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -188,34 +195,42 @@ func (tea *tableEditAccumulator) NewFromCurrent() *tableEditAccumulator {
...
-	}
-	if _, ok := tea.removedKeys[keyHash]; !ok {
-		// When rowData is updated, prevTea is set to nil. Therefore, if prevTea is non-nil, we use it.
-		if tea.prevTea != nil {
-			pkExists, err := tea.prevTea.Has(ctx, keyHash, key)
-			if err != nil {
-				return false, err
-			}
-			return pkExists, nil
+	if kvp, ok := tea.affectedKeys[keyHash]; ok {
```
Hydrocharged, line 206: I think we still need these `addedKeys/removedKeys` checks. If you add then remove then the key isn't really in the `tea`, but will still be logged in `affectedKeys`, so this will return it.

I'd like to see a test confirming that `ON DUPLICATE KEY` isn't triggered when you insert a new, then delete it, then insert it again.

Hydrocharged (approved):
LGTM, just with the single question/comment

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -46,11 +46,13 @@ func init() {
 	}
 }
 
+type PKDuplicateErrFunc func(keyString string, k, v types.Tuple) error
```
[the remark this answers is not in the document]
bheni: In this case, we are only calling this function based on a duplicated PK.

go/libraries/doltcore/table/editor/pk_table_editor.go
bheni: done

go/libraries/doltcore/table/editor/pk_table_editor.go
bheni: So I added this to the bats tests:

```bash
    dolt sql -b -q "INSERT INTO test VALUES ('d','d','d');DELETE FROM test WHERE col_a='d';INSERT INTO test VALUES ('d', '1', '1') ON DUPLICATE KEY UPDATE col_b = '2', col_c='2';"
    run dolt sql -r csv -q "SELECT * from test where col_a = 'd'"
    [ $status -eq 0 ]
    [[ "$output" =~ "d,1,1" ]] || false
```

and it works.  I think there are 2 things here.  First inserts and deletes will not be run as part of the same batch, but I think they could be now and I think this will still work because in the case of a delete we do:

`te.tea.affectedKeys[oldHash] = &doltKVP{k: dOldKeyVal}`

leaving `v` as nil.  Then in this check we do:

```go
	// No locks as all calls and modifications to tea are done from a lock that the caller handles
	if kvp, ok := tea.affectedKeys[keyHash]; ok {
		if kvp.v == nil {
			// if the row has been deleted then return that the row doesn't exist even if it's in the row data
			return nil, false, nil
		} else {
			return kvp, true, nil
		}
	}
```

I've added that comment there to make it clearer as this is easy to overlook.

go/libraries/doltcore/table/editor/pk_table_editor.go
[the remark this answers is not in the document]
VinaiRachakonda: Shouldn't this all be wrapped under a DuplicateKey error 

"Error number: 1062; Symbol: ER_DUP_ENTRY; SQLSTATE: 23000

Message: Duplicate entry '%s' for key %d"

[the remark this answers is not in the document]
VinaiRachakonda: Technically via the SQL standard Unique and Pk violations fit under the same error code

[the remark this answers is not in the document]
bheni: However you translate it up the stack is up to you.  I'm just adding more information to existing errors.

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -46,11 +46,13 @@ func init() {
...
+
 type TableEditor interface {
-	InsertKeyVal(ctx context.Context, key, val types.Tuple, tagToVal map[uint64]types.Value) error
+	InsertKeyVal(ctx context.Context, key, val types.Tuple, tagToVal map[uint64]types.Value, errFunc PKDuplicateErrFunc) error
 
-	InsertRow(ctx context.Context, r row.Row) error
-	UpdateRow(ctx context.Context, old, new row.Row) error
+	InsertRow(ctx context.Context, r row.Row, errFunc PKDuplicateErrFunc) error
```
zachmu, line 54: Rather than change all these interfaces, just add a DuplicateKeyErrorGenerator as a field to the relevant classes and make it settable. Kind of onerous to pass this around everywhere

zachmu (approved):
Implementation is sound, but would really prefer a new field instead of a param passed everywhere

PR #1591, opened 2021-04-25
create skipped test for dolt_branches with remotes

VinaiRachakonda:
Create a skipped bats test that defines the behavior we want for dolt_branches as described in this issue: #1592


timsehn (approved):
Love it. Thank you.

PR #1595, opened 2021-04-27
Vinai/fk errors

VinaiRachakonda:
Updates error logic for INSERT IGNORE INTO and adds bats


go/libraries/doltcore/table/editor/sessioned_table_editor.go
```diff
@@ -18,11 +18,12 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/dolthub/dolt/go/store/hash"
+	gsql "github.com/dolthub/go-mysql-server/sql"
```
Hydrocharged, line 21: Why add the `gsql`? Should just be the standard import directive, I'm not seeing another `sql` package here

integration-tests/bats/foreign-keys.bats
```diff
@@ -1708,3 +1708,66 @@ SQL
...
+    color VARCHAR(32),
+
+    PRIMARY KEY(id),
+    CONSTRAINT color_fk FOREIGN KEY (color) REFERENCES colors(color)
+);
+INSERT INTO colors (id,color) VALUES (1,'red'),(2,'green'),(3,'blue'),(4,'purple');
+INSERT INTO objects (id,name,color) VALUES (1,'truck','red'),(2,'ball','green'),(3,'shoe','blue');
+SQL
+
+    # Run a query and assert that no changes were made
```
Hydrocharged, line 1760: Better to assert this by checking the contents of the table. More important that the data is correct than the return message. This goes for all of your tests (including in `index.bats`), always check the data if you want to ensure nothing was changed.

Hydrocharged (approved):
Looks good, but the tests need some updating.

PR #1597, opened 2021-04-27
Rewrote index editor + other changes

Hydrocharged:


go/libraries/doltcore/row/row.go
```diff
@@ -114,20 +114,51 @@ func GetFieldByNameWithDefault(colName string, defVal types.Value, r Row, sch sc
...
-func ReduceToIndex(idx schema.Index, r Row) (Row, error) {
-	newRow := nomsRow{
-		key:   make(TaggedValues),
-		value: make(TaggedValues),
-		nbf:   r.Format(),
-	}
+// ReduceToIndexKeys creates a full key and a partial key from the given row (first tuple being the full key). As all
+// non-key tuples in an index map are empty, the full key is equivalent to the entire row of an index. The partial key
+// contains only the stated indexed columns, allowing for matches across duplicate values (for non-unique indexes).
+func ReduceToIndexKeys(idx schema.Index, r Row) (types.Tuple, types.Tuple, error) {
```
zachmu, line 120: I don't get the "partial key" abstraction, probably because I haven't been familiar with the index implementation code for a while now. But I can say this is a confusing abstraction. I don't know what a partial key is after reading this description, or why I would want it.

Does this need to be a single method? Maybe two methods, one for unique keys and one for non-unique, would make this a simpler read.

go/libraries/doltcore/row/row.go
```diff
@@ -114,20 +114,51 @@ func GetFieldByNameWithDefault(colName string, defVal types.Value, r Row, sch sc
...
+	}
+	return fullKey, partialKey, nil
+}
 
-	return newRow, nil
+// ReduceToIndexKeysFromTagMap creates a full key and a partial key from the given map of tags (first tuple being the
+// full key). As all non-key tuples in an index map are empty, the full key is equivalent to the entire row of an index.
+// The partial key contains only the stated indexed columns, allowing for matches across duplicate values (for
+// non-unique indexes).
+func ReduceToIndexKeysFromTagMap(nbf *types.NomsBinFormat, idx schema.Index, tagToVal map[uint64]types.Value) (types.Tuple, types.Tuple, error) {
```
zachmu, line 144: Same comment as above

zachmu (commented):
So basically this rewrites the index editor to more closely resemble the table editor? Seems pretty reasonable, no objection.

I didn't read the logic of the editor very closely, going to defer to Brian who was just in there making other changes.

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1511,7 +1511,7 @@ INSERT INTO child_non_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5'
 	require.NoError(t, err)
 	_, err = ExecuteSql(dEnv, root, "INSERT INTO child_unq VALUES ('6', 5)")
 	if assert.Error(t, err) {
-		assert.Contains(t, err.Error(), "UNIQUE constraint violation")
+		assert.Contains(t, err.Error(), "duplicate unique key")
```
bheni, line 1514: Can this be something like

    assert.True(errors.Is(err, ErrDuplicateKey))

So this doesn't fail if the str changes

PR #1598, opened 2021-04-27
gc changes for import

bheni:
New `dolt table import` flag --no-gc, and parameter tuning to run gcs less during import.


andrew-wm-arthur (approved):
lgtm

PR #1663, opened 2021-04-30
Fixed index editors getting to an invalid state on error

Hydrocharged:
Previously, if you had two or more indexes on a table and one of the indexes threw an error (such as a `UNIQUE` violation), the other index(es) would have whatever it was written to it still, leaving the table and index in an inconsistent state. This adds a small stack to index editors that allow one to undo the last few operations. In addition, the pk table editor tracks how many successful operations have been made per index editor and calls the equivalent number of undo statements, so that they're reset.

I chose to panic rather than error as, if these fail for whatever reason, we need to completely start over since things are broken and it's not really recoverable without rebuilding the entire index.

### DIFF
--- a/go/libraries/doltcore/table/editor/index_operation_stack.go
+++ b/go/libraries/doltcore/table/editor/index_operation_stack.go
@@ -1,0 +1,59 @@
+// Copyright 2021 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/table/editor/index_operation_stack.go
```diff
@@ -0,0 +1,63 @@
...
+	if ios.numOfItems == 0 {
+		return indexOperation{}, false
+	}
+	// numOfItems may be arbitrarily large as we don't check on Push, so we fix it to the array length here.
+	if ios.numOfItems > uint64(len(ios.entries)) {
+		ios.numOfItems = uint64(len(ios.entries)) - 1
+	} else {
+		ios.numOfItems--
+	}
+	ios.currentIndex = (ios.currentIndex - 1) % uint64(len(ios.entries))
```
bheni, line 61: in go -1 % 4 = -1

https://play.golang.org/p/-IOav06KgYF

So this can lead to an indexing error

bheni (approved):
lgtm

1 suggestion and 1 bug that should be fixed.

PR #1667, opened 2021-05-04
Add dolt log <table> functionality

VinaiRachakonda:


go/cmd/dolt/commands/log.go
```diff
@@ -198,3 +211,132 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
 
 	return 0
 }
+
+func checkIfTableExists(ctx context.Context, commit *doltdb.Commit, tableName string) (bool, error) {
```
zachmu, line 215: tableExists

go/cmd/dolt/commands/log.go
```diff
@@ -198,3 +211,132 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
...
+
+	// Check that the table exists in the head commit
+	exists, err := checkIfTableExists(ctx, commit, tableName)
+	if err != nil {
+		cli.PrintErrln(color.HiRedString("Fatal error: failed to get commit info"))
+		return 1
+	}
+
+	if !exists {
+		cli.PrintErrln("error: Invalid commit spec or table given")
```
zachmu, line 244: Should be "table %s does not exist"

go/cmd/dolt/commands/log.go
```diff
@@ -125,14 +126,26 @@ func logWithLoggerFunc(ctx context.Context, commandStr string, args []string, dE
...
 	}
 
-	cs, err := parseCommitSpec(dEnv, apr)
-	if err != nil {
-		cli.PrintErr(err)
-		return 1
+	numLines := apr.GetIntOrDefault(numLinesParam, -1)
+
+	argIsRef := false
+	if apr.NArg() == 1 {
```
zachmu, line 132: This logic isn't right, you need to be able to do 

`dolt log -n 1 HEAD~ myTable`

The table arg is always the last one, and restricts the log to only a single table

go/cmd/dolt/commands/log.go
```diff
@@ -198,3 +211,132 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
...
+		return 1
+	}
+
+	h, err := commit.HashOf()
+	if err != nil {
+		cli.PrintErrln(color.HiRedString("Fatal error: failed to get commit hash"))
+		return 1
+	}
+
+	commits, err := commitwalk.GetTopNTopoOrderedCommits(ctx, dEnv.DoltDB, h, -1)
```
zachmu, line 254: Don't do this, this fetches the entire commit history one commit at a time, you don't want that. Use `GetTopologicalOrderIterator` and iterate

go/cmd/dolt/commands/log.go
```diff
@@ -198,3 +211,132 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
...
+			return 1
+		}
+
+		pHashes, err := commit.ParentHashes(ctx)
+		if err != nil {
+			cli.PrintErrln("error: failed to get parent hashes")
+			return 1
+		}
+
+		for _, pHash := range pHashes {
```
zachmu, line 285: Don't do parent logic here, you want the commit walk library above to sort everything from child to ancestor for you

Instead, iterate over the commiterator, and keep track of the hash of the table named at every point and the previously seen commit. Any time it changes from one iteration to the next, log `prevCommit`

zachmu: Make sure you have a test case for a table being deleted between two revisions, then added again

zachmu (commented):
Good start, needs a little work

go/cmd/dolt/commands/log.go
```diff
@@ -198,3 +200,135 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
...
+
+	_, ok, err := rv.GetTable(ctx, tableName)
+	if err != nil {
+		return false, err
+	}
+
+	return ok, nil
+}
+
+func logTableCommits(ctx context.Context, dEnv *env.DoltEnv, loggerFunc commitLoggerFunc, cs *doltdb.CommitSpec, tableName string, numLines int) int {
```
zachmu, line 218: In general, subroutines in cmd/ shouldn't return error codes, they should return errors that are handled by top-level cmd routines

So instead of logging errors and returning 1, just return an error and do the status code handling at the call site

integration-tests/bats/log.bats
```diff
@@ -90,3 +90,153 @@ teardown() {
     regex='Merge:.*MergeCommit.*'
     [[ "$output" =~ $regex ]] || false
 }
+
+@test "log: Properly throws an error when neither a valid commit hash nor a valid table are passed" {
```
zachmu, line 94: Good tests

zachmu (approved):
Much better, thanks

PR #1681, opened 2021-05-06
Fixed occasional panic from nil VRW on BLOB pks

Hydrocharged:


go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -770,6 +770,9 @@ func (te *pkTableEditor) flushEditAccumulator(ctx context.Context, teaInterface
 	defer func() {
 		//TODO: need some way to reset an index editor to a previous point as well
 		if encounteredErr {
+			if recoveredErr := recover(); recoveredErr != nil && err == nil {
```
bheni, line 773: we don't use a lot of recover statements.  A comment about why this is here would probably be good.

bheni (approved):
lgtm

PR #1688, opened 2021-05-06
integration-tests/bats: Add some simple bats tests to views in sql-server multi-head mode.

reltuk:


reltuk:
@zachmu Really simple start to these for now.

zachmu (approved):
LGTM

PR #1695, opened 2021-05-07
go/libraries/doltcore/remotestorage: chunk_store.go: Stream chunk download locations instead of batching.

reltuk:
Recreating #1686 which was accidentally merged.


go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -249,8 +249,7 @@ func (dcs *DoltChunkStore) GetManyCompressed(ctx context.Context, hashes hash.Ha
 }
 
 const (
-	getLocsBatchSize      = 32 * 1024
-	getLocsMaxConcurrency = 4
+	getLocsBatchSize = (4 * 1024) / 20
```
bheni, line 252: 20 because that's the hash size?  Would probably just change this to:

```suggestion
	getLocsBatchSize = 256
```

bheni (approved):

bheni:
lgtm

PR #1706, opened 2021-05-12
Got rid of the table cache, untangled some of the circular references in table state management

zachmu:


zachmu:
#banchmark

zachmu:
#benchmark

go/libraries/doltcore/sqle/tables.go
```diff
@@ -128,9 +134,34 @@ func (t *DoltTable) WithIndexLookup(lookup sql.IndexLookup) sql.Table {
 	}
 }
 
+// doltTable returns the underlying doltTable from the current session
+func (t *DoltTable) doltTable(ctx *sql.Context) (*doltdb.Table, error) {
+	root := t.lockedToRoot
+	if root == nil {
```
andrew-wm-arthur, line 140: I don't love this bit of state to support `AS OF` queries. I think this would be better as a separate type that has an immutable root value

andrew-wm-arthur (approved):
LGTM

This is a definite improvement. I think we should keep looking for simplifications here, as it's still not easy to follow

go/libraries/doltcore/sqle/tables.go
zachmu: Yeah, I'll look for ways to get rid of this as I keep digging

PR #1707, opened 2021-05-12
Initial tag-reuse within working

max-hoffman:
- If a table is not in working,  is in head, and has the same schema as the table in head, re-use the tags and add the empty table back.
- The diff doesn't show `TRUNCATE table` or `DELETE from table` immediately after the create. I would have to edit the SQL path to change that.
- Bats tests added

todo:
-  engine tests?
- more complicated replace workflows. ex: replace table with extra column is easy, because I can do a one-way length check on columns and use a map for name/kind checks. Replace with delete column would be the same. Change column maybe also the same? If you do a series of replaces with column tags disappearing and reappearing, is it OK to enforce the re-use? Table moves should be explicit, that isn't a concern of replace workflows.

### DIFF
--- a/go/libraries/doltcore/doltdb/root_val.go
+++ b/go/libraries/doltcore/doltdb/root_val.go
@@ -173,7 +173,7 @@ func (root *RootValue) GenerateTagsForNewColColl(ctx context.Context, tableName 
 		return false, nil
 	})
 
-	newTags, err := root.GenerateTagsForNewColumns(ctx, tableName, newColNames, newColKinds)
+	newTags, err := root.GenerateTagsForNewColumns(ctx, tableName, newColNames, newColKinds, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -188,7 +188,7 @@ func (root *RootValue) GenerateTagsForNewColColl(ctx context.Context, tableName 
 
 // GenerateTagsForNewColumns deterministically generates a slice of new tags that are unique within the history of this root. The names and NomsKinds of
 // the new columns are used to see the tag generator.
-func (root *RootValue) GenerateTagsForNewColumns(ctx context.Context, tableName string, newColNames []string, newColKinds []types.NomsKind) ([]uint64, error) {
+func (root *RootValue) GenerateTagsForNewColumns(ctx context.Context, tableName string, newColNames []string, newColKinds []types.NomsKind, headRoot *RootValue) ([]uint64, error) {
 	if len(newColNames) != len(newColKinds) {
 		return nil, fmt.Errorf("error generating tags, newColNames and newColKinds must be of equal length")
 	}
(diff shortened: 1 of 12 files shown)


go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -215,6 +215,45 @@ func (root *RootValue) GenerateTagsForNewColumns(ctx context.Context, tableName
...
+		}
+
+		if found {
+			sch, err := tbl.GetSchema(ctx)
+			if err != nil {
+				return nil, err
+			}
+			var existingColNames []string
+			var existingColTags []uint64
+			_ = sch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
```
zachmu, line 231: Probably want a schemasEqual method for this

integration-tests/bats/foreign-keys.bats
```diff
@@ -1693,7 +1693,7 @@ SQL
     dolt commit -m "committed"
 }
 
-@test "foreign-keys: deleting and readding" {
+@test "foreign-keys: deleting and reading" {
```
zachmu, line 1696: think this is re-adding

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -215,6 +215,45 @@ func (root *RootValue) GenerateTagsForNewColumns(ctx context.Context, tableName
 		})
 	}
 
+	if headRoot != nil {
```
zachmu, line 218: Move this whole block above the rest of the logic, it's a totally separate code path

zachmu (approved):
LGTM

go/libraries/doltcore/doltdb/root_val.go
max-hoffman: yeah, if we do re-use on replace with slight differences (extra column, drop column, etc) that method wouldn't be used though

max-hoffman: maybe `schemaOverlap -> map[colName->tag]`? that would be useful now and later

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -188,11 +188,34 @@ func (root *RootValue) GenerateTagsForNewColColl(ctx context.Context, tableName
...
+			}
+
+			colOverlap := schema.GetSharedCols(sch, newColNames, newColKinds)
+			if len(colOverlap) == len(newColNames) {
+				var reusedTags []uint64
+				for _, k := range newColNames {
+					reusedTags = append(reusedTags, colOverlap[k])
+				}
+				return reusedTags, nil
+			}
```
max-hoffman, line 215: @zachmu lmk if this is similar to what you wanted. this has the added benefit of extending to partial re-use later

PR #1714, opened 2021-05-14
Vinai/gms bump fixes

VinaiRachakonda:
This pr addresses three things.

1. Bumping GMS 
2. Fixing compatibility tests due to GMS updates
3. Fixing index lookup equals to unskip a query planning test.
4.


go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -273,3 +274,59 @@ func (di *doltIndex) keysToTuple(keys []interface{}) (types.Tuple, error) {
...
+func (di *doltIndex) Equals(oIdx DoltIndex) bool {
+	if !expressionsAreEquals(di.Expressions(), oIdx.Expressions()) {
+		return false
+	}
+
+	if !(di.Database() == oIdx.Database()) {
+		return false
+	}
+
+	if !(di.ID() == oIdx.ID()) {
```
zachmu, line 287: Pretty sure we generate unique IDs for indexes, so you can probably just check this

zachmu (approved):
LG, nice work

Pretty sure you can do the Equals operation more simply, dig around a bit and see if you agree.

go/libraries/doltcore/sqle/dolt_index.go
VinaiRachakonda: I see what you're saying

PR #1715, opened 2021-05-14
/go/store/nbs/table_set.go: Maintain table spec order to enable Manifest Appendix remote optimization

coffeegoddd:


go/store/nbs/table_set.go
```diff
@@ -433,10 +433,12 @@ func (ts tableSet) Rebase(ctx context.Context, specs []tableSpec, stats *Stats)
...
 	// Create a list of tables to open so we can open them in parallel.
-	tablesToOpen := map[addr]tableSpec{}
+	tablesToOpen := []tableSpec{} // keep specs in order to play nicely with manifest appendix optimization
+	presents := map[addr]tableSpec{}
 	for _, spec := range specs {
-		if _, present := tablesToOpen[spec.name]; !present { // Filter out dups
-			tablesToOpen[spec.name] = spec
+		if _, present := presents[spec.name]; !present { // Filter out dups
+			tablesToOpen = append(tablesToOpen, spec)
+			presents[spec.name] = spec
```
reltuk, line 441: My one suggestion is to change further down to:

```go
	wg := &sync.WaitGroup{}
	wg.Add(len(tablesToOpen))
	for i, spec := range tablesToOpen {
		go func(idx int, spec tableSpec) {
			...
		}(i, spec)
	}
```

which is maybe a little bit more clear than what's down there currently.

reltuk (commented):
Looks great! Just one comment :).

go/store/nbs/table_set.go
coffeegoddd: added this change

PR #1716, opened 2021-05-14
proto,go: Add the ability to refresh signed URLs when fetching chunks from doltremoteapi.

reltuk:
Currently for a given GetMany call we get all signed URLs and chunk locations
from the API and then start making fetches against blob storage. If the signed
URLs expire during the fetch, we did not go back to the API to get new ones so
we could continue making progress.

This adds the for the API to send a refresh request and a refresh after
timestamp on a provided DownloadLoc in a response. All downloads going to the
same resource path now rendezvous through a check against the expiration time,
and we will refresh the URL against the API if it is expired.


go/libraries/doltcore/remotestorage/chunk_store.go
```diff
@@ -323,12 +323,22 @@ func (gr *GetRange) ChunkByteRange(i int) (uint64, uint64) {
 	return start, end
 }
 
-func (gr *GetRange) GetDownloadFunc(ctx context.Context, fetcher HTTPFetcher, chunkChan chan nbs.CompressedChunk) func() error {
+func (gr *GetRange) GetDownloadFunc(ctx context.Context, fetcher HTTPFetcher, chunkChan chan nbs.CompressedChunk, pathToUrl func(context.Context, string) (string, error)) func() error {
```
zachmu, line 326: named funcType?

zachmu (approved):
LGTM

PR #1717, opened 2021-05-14
[no-release-notes] /go/store/types/map.go: return the chunk and byte count from VisitMapLevelOrder

coffeegoddd:


reltuk (approved):
LGTM!

PR #1724, opened 2021-05-17
[no-release-notes] go/libraries/doltcore/diff: async_differ.go: Small refactoring to reduce duplication between GetDiff variants.

reltuk:


bheni (approved):
lgtm

PR #1735, opened 2021-05-20
[no-release-notes] /go/{store,libraries}: prevent appendix file clone

coffeegoddd:
I think its prudent to block clients from cloning appendix files for the time being... since they are so large and are stateful files (generated at HEAD, maybe on different branches), and client might clone a repo, unknowingly run `dolt gc` at some point, and i dont _think_ we want that.


reltuk (approved):
LGTM!

PR #1740, opened 2021-05-21
Removed flushing on foreign key checks for tables with pks

Hydrocharged:
Currently foreign keys necessitate a flush whenever you want to check a table editor's indexes for a partial key. With the recent index editor rewrite adding an efficient way to check partial keys, we can extend that and use it for foreign key checks as well.


bheni (approved):
dope

PR #1741, opened 2021-05-21
fixes infinite loop

bheni:
A deadlock can occur when the chunk size is greater than the maximum chunk size and a chunk with children gets removed, but it isn't removed from the withBufferedChildren map.


reltuk (approved):
LGTM!

PR #1744, opened 2021-05-22
Mark the session state clean upon init

zachmu:
I don't love this... it's meant to address this failure in dolthub:

https://github.com/dolthub/ld/pull/6605

Read-only is also kind of a bad name, it still writes to the chunk store if you issue a write query, it just won't call WriteRootValue when a transaction is committed

Thoughts?


zachmu:
Simpler solution: don't mark the session dirty upon init

PR #1746, opened 2021-05-24
Added mini rollback functionality for statements that error

Hydrocharged:
This fixes https://github.com/dolthub/dolt/issues/1738
Previously, whenever we failed a statement we never wrote a new root, however with the addition of transactions, we now update the root on both success and failure. As a result, an incorrect behavior is now surfacing in more interfaces.

The bug is that any core data manipulation statement (`INSERT`, `REPLACE`, `UPDATE`, `DELETE`) writes to the table editor as rows are modified. However, if one of the rows throws an error (such as from a duplicate key or foreign key violation), the entire statement's work is supposed to be discarded, but that wasn't happening. This adds that functionality by updating the interfaces (https://github.com/dolthub/go-mysql-server/pull/429) to denote the start and end points of a statement, along with alteration of the edit accumulators. There is a performance penalty, but it seems to be relatively small (<2% regression on local comparisons).

My true preference would be a more major rewrite to the table editor, reducing the operations to just two (`Insert` and `Delete`, similar to the current `IndexEditor`), and changing the edit accumulator system to a branching model (with one "central" editor accumulator and smaller accumulators focused on only that specific statement, where changes are "merged" back to the "central" accumulator). It would also be much clearer and cleaner than the PR code, and perhaps a tad faster than the current code.

But that's for the future.


Hydrocharged:
#benchmark

Hydrocharged:
#benchmark

integration-tests/bats/sql.bats
```diff
@@ -39,6 +39,83 @@ teardown() {
     teardown_common
 }
 
+@test "sql: errors do not write incomplete rows" {
```
zachmu, line 42: Would like to see equivalents of these tests in the engine unit tests as well, it's pretty important

zachmu (commented):
Seems like a reasonable approach but will leave a close reading to Brian

go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -93,16 +94,16 @@ type indexEditAccumulator struct {
...
-	ed      types.EditAccumulator
-	opCount uint64
+	opCount int64
 
 	// addedPartialKeys is a map of partial keys to a map of full keys that match the partial key
 	addedPartialKeys map[hash.Hash]map[hash.Hash]types.Tuple
-	addedKeys        map[hash.Hash]struct{}
-	removedKeys      map[hash.Hash]struct{}
+	addedKeys        map[hash.Hash]hashedTuple // These hashes represent the hash of the partial key, with the tuple being the full key
+	removedKeys      map[hash.Hash]hashedTuple // These hashes represent the hash of the partial key, with the tuple being the full key
```
bheni, line 102: I worry about the size of these two maps.

go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -390,10 +387,91 @@ func (ie *IndexEditor) Index() schema.Index {
 	return ie.idx
 }
 
-// Close ensures that all goroutines that may be open are properly disposed of. Attempting to call any other function
-// on this editor after calling this function is undefined behavior.
+// StatementStarted is analogous to the TableEditor implementation, but specific to the IndexEditor.
+func (ie *IndexEditor) StatementStarted(ctx context.Context) {
+	ie.flushMutex.Lock()
```
bheni, line 392: defer ie.flushMutex.Unlock() is more idiomatic rather than putting at the end of the function.

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -723,10 +696,92 @@ func (te *pkTableEditor) Format() *types.NomsBinFormat {
 	return te.nbf
 }
 
+// StatementStarted implements TableEditor.
+func (te *pkTableEditor) StatementStarted(ctx context.Context) {
+	te.flushMutex.Lock()
```
bheni, line 701: defer unlock

bheni (approved):
Couple comments but LGTM

go/libraries/doltcore/table/editor/index_editor.go
Hydrocharged: It shouldn't be any different than before since I removed `affectedKeys`, and that contained all of the entries anyway.

PR #1753, opened 2021-05-26
Transaction tests for dolt, and couple small bug fixes for transactions

zachmu:


go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -42,13 +42,15 @@ const (
 
 const EnableTransactionsEnvKey = "DOLT_ENABLE_TRANSACTIONS"
 
-var transactionsEnabled = false
+// TransactionsEnabled controls whether to use SQL transactions
+// Exported only for testing
+var TransactionsEnabled = true
```
Hydrocharged, line 47: I think you meant to set this back to `false`. Probably why the test is failing too.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -299,3 +299,13 @@ func TestStoredProcedures(t *testing.T) {
 
 	enginetest.TestStoredProcedures(t, newDoltHarness(t))
 }
+
+func TestTransactions(t *testing.T) {
+	if !sqle.TransactionsEnabled {
+		sqle.TransactionsEnabled = true
+		defer func() {
+			sqle.TransactionsEnabled = false
```
Hydrocharged, line 307: `defer` should probably restore the state rather than force a `false`.

Hydrocharged (approved):
LGTM!

go/libraries/doltcore/sqle/dolt_session.go
zachmu: yuuuuup

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
zachmu: We only do this if it's false to begin with, so that's equivalent

PR #1755, opened 2021-05-27
Added session var that matches dolt and transaction commits

Hydrocharged:
This adds a new session variable named `dolt_transaction_commit` that, when enabled, will cause a dolt commit of all changes in the working set when a transaction commit is completed.

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_session.go
+++ b/go/libraries/doltcore/sqle/dolt_session.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
 	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
 	"github.com/dolthub/dolt/go/store/hash"
@@ -40,7 +41,10 @@ const (
 	WorkingKeySuffix = "_working"
 )
 
-const EnableTransactionsEnvKey = "DOLT_ENABLE_TRANSACTIONS"
+const (
+	EnableTransactionsEnvKey      = "DOLT_ENABLE_TRANSACTIONS"
+	DoltCommitOnTransactionCommit = "dolt_transaction_commit"
+)
 
 // TransactionsEnabled controls whether to use SQL transactions
 // Exported only for testing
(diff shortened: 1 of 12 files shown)


Hydrocharged:
I know the tests are ugly and horribly inefficient from a coding perspective, I apologize.

Also, this is built on top of Zach's https://github.com/dolthub/dolt/pull/1753, so this will fail for now as that PR is currently failing.

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -187,6 +201,31 @@ func (sess *DoltSession) CommitTransaction(ctx *sql.Context, dbName string, tx s
 		return err
 	}
 
+	if commitBool, err := sess.Session.GetSessionVariable(ctx, DoltCommitOnTransactionCommit); err != nil {
```
zachmu, line 204: Method for this?

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -0,0 +1,270 @@
...
+	require.NoError(t, err)
+	initialCommit, err := commit.GetAncestor(context.Background(), as)
+	require.NoError(t, err)
+	icm, err := initialCommit.GetCommitMeta()
+	require.NoError(t, err)
+	require.Equal(t, "Initialize data repository", icm.Description)
+}
+
+func TestDoltTransactionCommitTwoClients(t *testing.T) {
+	// In this test, we're setting only both clients to match transaction commits to dolt commits.
```
zachmu, line 103: only both?

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/dolthub/go-mysql-server/enginetest"
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle"
+)
+
+func TestDoltTransactionCommitOneClient(t *testing.T) {
```
zachmu, line 15: These tests should also check that the db_head variable moves after a commit

This is subtle and probably doesn't work correctly right now. The variable has to move for the client that commits, but not for the other client until they begin a new tx.

And at that point, they need to see the new head value

zachmu (approved):
LG, just a couple things you can address as a followup

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
Hydrocharged: Ohh yeah I’ll add that

go/libraries/doltcore/sqle/dolt_session.go
andrew-wm-arthur: +1

andrew-wm-arthur (approved):
lgtm

PR #1766, opened 2021-05-28
Vinai/create temp tables

VinaiRachakonda:
This pr allows for the creation of temporary tables.


go/libraries/doltcore/sqle/database.go
```diff
@@ -614,6 +673,67 @@ func (db Database) createDoltTable(ctx *sql.Context, tableName string, root *dol
...
+		return sql.ErrTableAlreadyExists.New(tableName)
+	}
+
+	var conflictingTbls []string
+	_ = doltSch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
+		_, tbl, exists, err := root.GetTableByColTag(ctx, tag)
+		if err != nil {
+			return true, err
+		}
+		if exists && tbl != tableName {
```
zachmu, line 718: This condition should be impossible, a panic is more appropriate

go/libraries/doltcore/sqle/database.go
```diff
@@ -657,7 +777,18 @@ func (db Database) Flush(ctx *sql.Context) error {
...
+		return nil
+	}
+
+	// Flush any changes made to temporary tables
+	newTempTableRoot, err := dsess.tempTableEditSessions[db.Name()].Flush(ctx)
+	if err != nil {
+		return nil
+	}
+
+	return dsess.SetTempTableRoot(ctx, db.Name(), newTempTableRoot)
```
zachmu, line 791: This is a problem... we need to update either both roots or neither of them, atomically

I think a TODO is fine for now

go/libraries/doltcore/sqle/database.go
```diff
@@ -915,10 +1046,40 @@ func (db Database) dropFragFromSchemasTable(ctx *sql.Context, fragType, name str
...
-func (db Database) TableEditSession(ctx *sql.Context) *editor.TableEditSession {
+func (db Database) TableEditSession(ctx *sql.Context, isTemporary bool) *editor.TableEditSession {
+	if isTemporary {
+		return DSessFromSess(ctx.Session).tempTableEditSessions[db.name]
+	}
 	return DSessFromSess(ctx.Session).editSessions[db.name]
 }
 
+// GetAllTemporaryTables returns all temporary tables
+func (db Database) GetAllTemporaryTables(ctx *sql.Context) ([]sql.Table, error) {
```
zachmu, line 1057: Delete as discussed in gms PR

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -622,6 +643,25 @@ func (sess *DoltSession) AddDB(ctx *sql.Context, db sql.Database, dbData env.DbD
 	// After setting the initial root we have no state to commit
 	sess.dirty[db.Name()] = false
 
+	err = sess.createTemporaryTablesRoot(ctx, db.Name(), ddb)
```
zachmu, line 646: This is expensive, it does IO

Rather than do this on session init, do it on demand the first time that a temp table is created

Don't maintain a temp root for sessions with no temp tables

integration-tests/bats/sql-create-tables.bats
```diff
@@ -410,3 +410,323 @@ SQL
...
+    run dolt status
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "On branch master" ]] || false
+    [[ "$output" =~ "nothing to commit, working tree clean" ]] || false
+}
+
+@test "sql-create-tables: You can drop temp tables" {
+    run dolt sql <<SQL
+CREATE TEMPORARY TABLE mytemptable(pk int PRIMARY KEY);
+DROP TABLE mytemptable
```
zachmu, line 517: Need another query to confirm it was actually dropped

integration-tests/bats/sql-server.bats
```diff
@@ -764,4 +764,24 @@ SQL
...
+    skiponwindows "Has dependencies that are missing on the Jenkins Windows installation."
+
+    cd repo1
+    start_sql_server repo1
+
+    # check no tables on master
+    server_query 1 "SHOW Tables" ""
+
+    # Create a temporary table with some indexes
+    server_query 1 "CREATE TEMPORARY TABLE one_pk (
```
zachmu, line 779: Need to verify that it still exists in this session in server mode, might need changes to the helper script for that

go/libraries/doltcore/sqle/dtables/branches_table.go
```diff
@@ -76,6 +76,10 @@ func (bt *BranchesTable) PartitionRows(sqlCtx *sql.Context, part sql.Partition)
 	return NewBranchItr(sqlCtx, bt.ddb)
 }
 
+func (bt *BranchesTable) IsTemporary() bool {
```
zachmu, line 79: I'm thinking that IsTemporary should be on a separate TemporaryTable interface in gms

zachmu (commented):
Looks pretty good, just needs a few things

go/libraries/doltcore/sqle/database.go
```diff
@@ -485,13 +508,30 @@ func (db Database) GetRoot(ctx *sql.Context) (*doltdb.RootValue, error) {
 	return currRoot.root, nil
 }
 
+func (db Database) GetTemporaryTablesRoot(ctx *sql.Context) (*doltdb.RootValue, error) {
+	dsess := DSessFromSess(ctx.Session)
+	currRoot, dbRootOk := dsess.tempTableRoots[db.name]
```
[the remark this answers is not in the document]
VinaiRachakonda: I didn't think it would be correct to change the method header for the accessor

go/libraries/doltcore/sqle/database.go
VinaiRachakonda: Resolved that this will continue to exist

integration-tests/bats/sql-server.bats
VinaiRachakonda: Can this be moved to a separate pr? I'd love to take some time to scope out better testing for multiple queries in the same session. Disconnecting on every query isn't good enough.

go/libraries/doltcore/sqle/database.go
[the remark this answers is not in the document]
VinaiRachakonda: Ideally I would pass in a bool param (isTemporary) but that would mean a lot of changes around the codebase

go/libraries/doltcore/sqle/database.go
```diff
@@ -515,14 +555,33 @@ func (db Database) GetHeadRoot(ctx *sql.Context) (*doltdb.RootValue, error) {
...
 
+	// Temporary Tables Get Precedence over schema tables
+	tempTableRoot, err := db.GetTemporaryTablesRoot(ctx)
 	if err != nil {
 		return err
 	}
 
-	if doltdb.IsReadOnlySystemTable(tableName) {
-		return ErrSystemTableAlter.New(tableName)
+	tempTableExists, err := tempTableRoot.HasTable(ctx, tableName)
```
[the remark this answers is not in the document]
VinaiRachakonda: Seems to work fine with case insensitive matches. It pretty much follows the same pattern as below.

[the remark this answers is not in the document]
VinaiRachakonda: It works because the planner passes in the correct case sensitive name. I'll leave a comment

go/libraries/doltcore/sqle/dtables/branches_table.go
VinaiRachakonda: I think there is less code with that approach. Although a separate interface feels wrong because every single table type can be a temporary table. I can see us in the future even allowing systems tables to be manipulated in temporary table formats to allow greater experimentation. 

Integrators should also be aware that they need to consider temporary table support.

go/libraries/doltcore/sqle/database.go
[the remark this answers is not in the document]
zachmu: I mean this:

dsess.GetTempTableRootValue(ctx, db.Name())

go/libraries/doltcore/sqle/database.go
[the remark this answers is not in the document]
zachmu: As long as there are tests, I'm happy

go/libraries/doltcore/sqle/database.go
```diff
@@ -635,6 +697,62 @@ func (db Database) createDoltTable(ctx *sql.Context, tableName string, root *dol
 	return db.SetRoot(ctx, newRoot)
 }
 
+// CreateTemporaryTable creates a table that only exists the length of a schema.
```
zachmu, line 700: length of a session

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -454,6 +460,33 @@ func (sess *DoltSession) SetRoot(ctx *sql.Context, dbName string, newRoot *doltd
 	return nil
 }
 
+func (sess *DoltSession) GetTempTableRootValue(ctx *sql.Context, dbName string) (*doltdb.RootValue, error) {
```
zachmu, line 463: This still isn't good enough. You need to only create a temp root / edit session when a temp table is created. The way this is written now, you are doing this on every table lookup, so you'll create a temp root for every session still

integration-tests/bats/sql-server.bats
zachmu: SGTM, must add a TODO in the meantime

zachmu (commented):
Looks better, but you're still creating a temp root for every session, which is expensive. Needs to happen only if a temp table is created.

go/libraries/doltcore/sqle/dolt_session.go
```diff
@@ -695,6 +716,21 @@ func (sess *DoltSession) AddDB(ctx *sql.Context, db sql.Database, dbData env.DbD
...
+// CreateTemporaryTablesRoot creates an empty root value and a table edit session for the purposes of storing
+// temporary tables. This should only be used on demand. That is only when a temporary table is created should we
+// create the root map and edit session map.
+func (sess *DoltSession) CreateTemporaryTablesRoot(ctx *sql.Context, dbName string, ddb *doltdb.DoltDB) error {
+	newRoot, err := doltdb.EmptyRootValue(ctx, ddb.ValueReadWriter())
+	if err != nil {
+		return err
+	}
+
+	sess.tempTableRoots[dbName] = newRoot
```
zachmu, line 728: Should call session.SetTempTableRoot here instead of direct asssignment (after you create the edit session below)

go/libraries/doltcore/sqle/tables.go
```diff
@@ -156,6 +159,21 @@ func (t *DoltTable) doltTable(ctx *sql.Context) (*doltdb.Table, error) {
 	return table, nil
 }
 
+// getRoot returns the correct root value that needs to updated with the new tale info. The only controlling factor
```
zachmu, line 162: table info?

zachmu: Should say "returns the appropriate root value for this table from the session"

zachmu (approved):
LG, thanks for the changes

PR #1786, opened 2021-06-04
Test support for transactions, and changed the behavior of dolt sql in batch mode when it encounters an error

zachmu:
These are changes necessary to make all dolt tests pass with transactions enabled. It moves transaction enablement to the session, and runs all engine tests with and without transactions enabled.

Batch mode configuration is moved out of the database into the session.

The behavior of `dolt sql` running in batch mode is changed to write to disk before exit if it encounters an error. Previously, batch mode would not change the repo state if it encountered an error in any statement.


go/libraries/doltcore/sqle/testutil.go
```diff
@@ -29,12 +29,14 @@ import (
...
+// ExecuteSql executes all the SQL non-select statements given in the string against the root value given and returns
+// the updated root, or an error. Statements in the input string are split by `;\n`
 func ExecuteSql(dEnv *env.DoltEnv, root *doltdb.RootValue, statements string) (*doltdb.RootValue, error) {
-	db := NewBatchedDatabase("dolt", dEnv.DbData())
-	engine, ctx, err := NewTestEngine(context.Background(), db, root)
+	db := NewDatabase("dolt", dEnv.DbData())
 
+	engine, ctx, err := NewTestEngine(context.Background(), db, root)
+	DSessFromSess(ctx.Session).EnableBatchedMode()
+	err = ctx.Session.SetSessionVariable(ctx, sql.AutoCommitSessionVar, uint8(0))
```
Hydrocharged, line 39: Similar to the comment earlier, for all of the `SetSessionVariable` calls you can just pass in `false` as it's clearer.

Hydrocharged (approved):
LGTM! Just some small nits

zachmu:
Thanks for the feedback!

PR #1811, opened 2021-06-11
Patched random ACCESS DENIED error on Windows

Hydrocharged:
This patches the random `ACCESS DENIED` error that may occur on Windows when noms attempts to rename the manifest.

### DIFF
--- a/go/store/nbs/file_manifest.go
+++ b/go/store/nbs/file_manifest.go
@@ -377,7 +377,7 @@ func parseIfExistsWithParser(_ context.Context, dir string, parse manifestParser
 		return false, manifestContents{}, err
 	}
 
-	// !exists(lockFileName) => unitialized store
+	// !exists(lockFileName) => uninitialized store
 	if locked {
 		var f *os.File
 		err = func() (ferr error) {
@@ -503,7 +503,7 @@ func updateWithParseWriterAndChecker(_ context.Context, dir string, write manife
 			defer func() {
 				closeErr := f.Close()
 
-				if ferr != nil {
+				if ferr == nil {
 					ferr = closeErr
 				}
 			}()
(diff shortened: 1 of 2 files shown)


Hydrocharged:
Left a massive comment in the code, so refer to that. But I'll add that I tried many, many different solutions. I cannot find the root cause of the issue. However, my patch does appear to fix it. To test, I imported a 10,000,000 statement file, and I set GC to run after every statement. On average I would get the `ACCESS DENIED` error every 1,500 renames, although I've seen it occur after as few as 200 renames and as many as 7,000 renames. With the patch, I was able to import the entire 10,000,000 statement file, so I'm fairly confident that this "fixes" it, although it's impossible to truly know without knowing the root cause.

Also, I've got no idea how to test this. Including a 1GB file isn't feasible, and the variability just means we're acquiring a confidence factor rather than a true test.

Hydrocharged:
Changed this to instead use exponential backoff. Didn't run it the whole 10,000,000 statements but I feel confident enough that it would have passed it. Whatever is causing the failure seems to disappear after even a 1ms delay for the retry, so the additional delays are a safety net. This doesn't leave the repo in a potentially invalid state that would require manual fixing.

reltuk (approved):
Love it!

go/store/nbs/file_manifest.go
```diff
@@ -546,9 +546,22 @@ func updateWithParseWriterAndChecker(_ context.Context, dir string, write manife
...
 	err = os.Rename(tempManifestPath, manifestPath)
-
 	if err != nil {
-		return manifestContents{}, err
+		// On Windows, renaming the temporary manifest file to the current manifest file overwrites the current file.
+		// This can occasionally cause an "ACCESS DENIED" error, aborting the entire operation. The cause is not clear,
+		// as it does not appear to be a dangling file handle (observed through Process Explorer). The error is also
+		// hard to reproduce and inconsistent, as it seems to occur between 200-7,000 renames, but averages around 1,500
+		// renames. This adds a delay before retring the rename, increasing the wait time by a factor of 10 after each
+		// failure, up to a max of 10 seconds. If an error still occurs after that time, then we just fail. It is
```
timsehn, line 555: retrying

timsehn (commented):
Spelling mistake in comment

PR #1814, opened 2021-06-11
[no-release-notes] added broken queries

andrew-wm-arthur:


go/libraries/doltcore/sqle/enginetest/dolt_system_table_queries.go
```diff
@@ -0,0 +1,40 @@
+package enginetest
+
+import (
+	"github.com/dolthub/go-mysql-server/enginetest"
+	"github.com/dolthub/go-mysql-server/sql/analyzer"
+)
+
+var BrokenQueries = []enginetest.QueryErrorTest{
```
zachmu, line 8: Can you structure these so that they register as skips in the engine test suite?

PR #1815, opened 2021-06-11
Vinai/elixir smoke test

VinaiRachakonda:


integration-tests/MySQLDockerfile
```diff
@@ -96,4 +103,4 @@ COPY mysql-client-tests /mysql-client-tests
 COPY mysql-client-tests/mysql-client-tests-entrypoint.sh /mysql-client-tests/entrypoint.sh
 
 WORKDIR /mysql-client-tests
-ENTRYPOINT ["/mysql-client-tests/entrypoint.sh"]
+ENTRYPOINT ["/mysql-client-tests/entrypoint.sh"]
```
coffeegoddd, line 106: newline

integration-tests/mysql-client-tests/mysql-client-tests.bats
```diff
@@ -100,6 +100,17 @@ cmake ..
     ruby $BATS_TEST_DIRNAME/ruby/ruby-mysql-test.rb $USER $PORT $REPO_NAME
 }
 
+@test "elixir myxql test" {
+    cd $BATS_TEST_DIRNAME/elixir/
+    # install some mix dependencies
```
coffeegoddd, line 105: can we do this in the dockerfile with the rest of the deps?

coffeegoddd (approved):
LG, couple comments

integration-tests/mysql-client-tests/mysql-client-tests.bats
VinaiRachakonda: I think elixir expects dependencies to be local so I have to leave it like this :/ Tried the other way around and it didn't work

PR #1828, opened 2021-06-15
Fixed #1824: Bug with adding index on merge

### LINKED ISSUE #1824: Merging a branch with a new index into branch without can error
If you create an index on a branch, then try to merge that back into a branch that has since added rows to the table with the new index, we get the following error:

```
   error: Failed to commit changes.
   cause: Bad merge
          cause: index `abc` is missing its data
```

A workaround is to merge in the other direction first, so that the index has a value for every row in the merge target table.

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/1824

### DIFF
--- a/go/libraries/doltcore/merge/merge.go
+++ b/go/libraries/doltcore/merge/merge.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


zachmu (approved):
LG thanks for the quick turnaround

PR #1831, opened 2021-06-15
--merges, --parents, --min-parents options for dolt log

bheni:


timsehn:
BATS tests?

go/cmd/dolt/commands/log.go
```diff
@@ -168,7 +200,7 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, cs *doltdb.CommitSpec, l
 		return 1
 	}
 
-	commits, err := commitwalk.GetTopNTopoOrderedCommits(ctx, dEnv.DoltDB, h, numLines)
+	commits, err := commitwalk.GetTopNTopoOrderedCommits(ctx, dEnv.DoltDB, h, opts.numLines)
```
andrew-wm-arthur, line 203: combining options `-n=x` and `--min-parents=y` will return incorrect results if we only fetch `-n` commits here

andrew-wm-arthur (approved):
LGTM, could use some BATS, octopus merges gonna be so great

PR #1832, opened 2021-06-15
ctx to WithFilters

bheni:


zachmu (approved):
Add some type assertions if they don't already have it, otherwise LGTM

Issue #1838, opened 2021-06-16
cat / diff leave output colored after SIGINT 


CaptainStabs:
After killing a `dolt diff` or `dolt conflicts cat` output using SIGINT, the terminal remains the last color used. 
![image](https://user-images.githubusercontent.com/40151222/122301120-3d2ee380-cece-11eb-8018-fe8a2aa5f62f.png)

Tested on:

Microsoft Windows 10 Home, Version 10.0.19042 Build 19042
@timsehn [says](https://discord.com/channels/746150696465727668/746152112169287760/854844182228893706) he's seen it on MacOs

VinaiRachakonda:
Closing since pr was merged in

PR #1842, opened 2021-06-17
reset output colors before exiting when interrupt encountered

bheni:


CaptainStabs:
Related to issue #1838

CaptainStabs:
Just tested, works on windows, but with a severe loss of responsivity (Nearly 16 seconds after SIGINT is first pressed)

go/cmd/dolt/cli/stdio.go
```diff
@@ -19,13 +19,27 @@ import (
...
 
+var outputClosed uint64
+
+func CloseOutput() {
+	if atomic.CompareAndSwapUint64(&outputClosed, 0, 1) {
+		fmt.Fprintln(CliOut)
+	}
+}
+
+func outputIsCLosed() bool {
```
zachmu, line 38: CLosed => Closed

zachmu (approved):
LGTM

bheni:
@CaptainStabs I've merged this to master.

In order to write colored output to the command line, we use [ansi escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code).  These codes are interpreted by the terminal window that renders the output.  When an ansi escape code is encountered the terminal window will change the color of the text that is written until more ansi escape codes are encountered changing the color to something new, or back to the default.

As Dolt writes text to standard out it is buffered.  Depending on the os and setup there could be a lot of text buffered to be written to standard out, and at the point that ctrl-c is hit.  If the program terminates immediately you are left in a state where your terminal window may still be actively writing colored text.  This change prevents further text from being written to standard out, and writes the ansi control codes to change the text back to the default.  The issue is that for these ansi control codes to be interpreted all the text that's already been buffered needs to be flushed.  That can lead to it feeling less responsive when you hit ctrl-c as it flushes and resets the text color.

With this change you have to options:
 1. Send sigint (ctrl+c) and know that your terminal colors will be reset but it may take a little time to flush
 2. Send sigterm (pkill / kill command in linux, not sure in windows) to kill dolt immediately, but your text color may not be properly reset as your terminal window will be left writing whatever color it wrote last.

PR #1861, opened 2021-06-25
Db/visit level order

coffeegoddd:


go/store/types/map_test.go
```diff
@@ -2109,3 +2109,34 @@ func TestMapWithNil(t *testing.T) {
...
+	d.PanicIfError(err)
+
+	actualChunkHashes := make([]hash.Hash, 0)
+	_, _, err = VisitMapLevelOrderSized([]Map{m}, 37624, func(h hash.Hash) (int64, error) {
+		actualChunkHashes = append(actualChunkHashes, h)
+		return 0, nil
+	})
+	d.PanicIfError(err)
+
+	assert.Equal(expectedChunkHashes, actualChunkHashes)
```
reltuk, line 2141: Might want to sort these, and add another case where you test `VisitMapLevelOrderSized` with a real small limit like `2`, just to make sure everything is proceeding as expected.

reltuk (approved):
LGTM!

PR #1866, opened 2021-06-28
Remote system table

max-hoffman:
todo:
- delete remote test
- maybe engine tests
- fix copied docstrings
- maybe fix the janky repo state reader, seems like i should use `dEnv`


integration-tests/bats/system-tables.bats
```diff
@@ -137,6 +143,25 @@ teardown() {
...
+
+    run dolt sql -q "select count(*) from dolt_remotes"
+    [ $status -eq 0 ]
+    [[ "$output" =~ 1 ]] || false
+
+    regex='file://.*/remote'
+    run dolt sql -q "select * from dolt_remotes" -r csv
+    [ $status -eq 0 ]
+    [[ "${lines[0]}" = name,url,fetch_specs,params ]] || false
+    [[ "${lines[1]}" =~ origin,file://.*/remote,[refs/heads/*:refs/remotes/origin/*,map[] ]] || false
```
max-hoffman, line 162: @zachmu keep fetch specs and params? reformat the string output? drop one or both?

timsehn:
Make sure when you ship this to add a section in the documentation.

PR #1873, opened 2021-06-29
Fix csv export types

max-hoffman:
- JSON strings were not serialized in CSV writing
- convert Dolt rows to SQL Rows, and lean on the `SqlColToStr` to correctly parse values to strings.
- manually handle nil values to match bats tests CSV export format


Hydrocharged (approved):
LGTM!

PR #1876, opened 2021-06-29
Vinai/ai transactions

VinaiRachakonda:


go/libraries/utils/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,72 @@
+package autoincr
+
+type AutoIncrementTracker interface {
+	Get(dbname, tableName string) (uint64, bool)
```
zachmu, line 4: Rather than Get and Set, just have one method, Reserve. Returns true or false, whether the reservation was successful (false if taken). Then clients try again with the next value as necessary

Instead of this interface being responsible for giving out next values, it just needs to remember what has been given out so integrators can adjust accordingly

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -684,6 +708,45 @@ func (te *pkTableEditor) Table(ctx context.Context) (*doltdb.Table, error) {
 	return te.t, nil
 }
 
+func (te *pkTableEditor) updateAutoIncrementTracker(val interface{}) error {
```
zachmu, line 711: This is suspect. Why do we need to manage the state explicitly like this? It should happen automatically as auto inc keys are handed out, no need to tell anyone about it

go/libraries/utils/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,72 @@
...
+	Set(dbname, tableName string, val uint64) bool
+}
+
+type AutoIncrementTrackerSubscriber interface {
```
zachmu, line 8: Get rid of this if possible, just use the tracker interface

zachmu (commented):
Generally a good start, comments on how to simplify this design

go/libraries/doltcore/table/editor/pk_table_editor.go
VinaiRachakonda: What happens if someone skips ahead on the current reserved value. Something still needs to tell the tracker to update

VinaiRachakonda:
@zachmu I almost think we should do get and set now

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,120 @@
+package autoincr
+
+import (
+	"fmt"
+	"sync"
+)
+
+type AutoIncrementTracker interface {
```
zachmu, line 8: This is the wrong interface. It's combining two different patterns.

Either this thing should be dumb, in which case its only responsibility is to tell people which IDs have been passed out already, and let them tell it "hey I passed out this ID"

Or it should be smart, and handle passing out the Next ID automatically.

What you have here combines these two modes of operation, and it's kind of a mess.

I would recommend going with option 2, and making this thing smart. Make it responsible for handing out all auto increment keys for each table.

That leaves you with just two interface methods here: InitTable(tableName, initialVal) and Next(tableName). Next should return a specific error type if that table hasn't been initialized yet in this session, and then the client can call InitTable on it.

The above pattern means you have to be tolerant of multiple init calls. You need to make InitTable forgiving if it gets called by more than one thread (probably silently succeeding if gets called more than once). The other option is to add Lock and Unlock methods to the interface so that callers can do the above two calls in a single mutex.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -608,12 +608,50 @@ func (t *WritableDoltTable) AutoIncrementSetter(ctx *sql.Context) sql.AutoIncrem
...
+		return nil, err
+	}
+
+	ok, err := autoIncTracker.Request(t.tableName, stored)
+	if err != nil {
+		return nil, err
+	}
+
+	// If the current session cannot requests its current key than ask for the next one and request again.
+	if !ok {
```
zachmu, line 633: This isn't right, it only works if exactly one other thread has requested a key. This needs to be a loop.

(or, per above comment, make the tracker smart so that Next() always just works)

zachmu (commented):
Needs some work, hit me up on chat if you want to talk about it

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
zachmu: Also you don't need to worry about skipping or burning auto increment IDs that you handed out to transactions that eventually fail. MySQL does this too, and application developers already have to deal with it.

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+
+type autoIncrementTracker struct {
+	tables  map[string]interface{}
+	written map[string]bool
+	mu      sync.Mutex
+}
+
+var _ AutoIncrementTracker = (*autoIncrementTracker)(nil)
+
+// Request is used by an integrator to request whether the wanted auto increment key is valid for use. Requests do not
```
zachmu, line 30: This is a garbled description.

Start with what the method returns, then move on to how it's used

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+// It is primarily used for concurrent transactions on an auto incremented table.
+func NewAutoIncrementTracker() AutoIncrementTracker {
+	return &autoIncrementTracker{
+		tables:  make(map[string]interface{}),
+		written: make(map[string]bool),
+	}
+}
+
+type autoIncrementTracker struct {
+	tables  map[string]interface{}
```
zachmu, line 23: Is it fair to say this is highestForTable?

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,100 @@
...
+	return true, nil
+}
+
+// Confirm is used when an auto increment has actually been written to a table.
+func (a *autoIncrementTracker) Confirm(tableName string, key interface{}) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	a.tables[tableName] = key
+	a.written[tableName] = true
```
zachmu, line 70: How is this supposed to work? We're setting this to true for all values? What if you have some sessions requesting new keys at the same time that other sessions are writing their old keys?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -608,12 +609,55 @@ func (t *WritableDoltTable) AutoIncrementSetter(ctx *sql.Context) sql.AutoIncrem
...
+	if !foundAit || autoIncTracker == nil {
+		return t.getAutoIncrementValue(ctx)
+	}
+
+	stored, err := t.getAutoIncrementValue(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	ok, err := autoIncTracker.Request(t.tableName, stored)
```
zachmu, line 628: Just return a best guess here, with the understanding that the table editor will replace it with the true value before committing.

zachmu (commented):
There is no way this works under real concurrency. Needs another pass.

Request() gets called multiple times because of analysis in the engine, then solve that problem separately. Don't even use the global state in WritableDoltTable. Return a best guess placeholder, without consulting any session state. Then in the table editor, when it comes time to actually put a value in storage, replace the placeholder with the global state.

You could also consider restructuring the go-mysql-server insert operation to get an auto inc value at runtime, rather than analysis time.

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (a *autoIncrementTracker) InitTable(tableName string, val interface{}) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	a.valuePerTable[tableName] = val
+}
+
+func (a *autoIncrementTracker) Peek(tableName string) interface{} {
+	return a.valuePerTable[tableName]
```
zachmu, line 63: Need a lock here

go/libraries/doltcore/sqle/tables.go
```diff
@@ -611,10 +620,34 @@ func (t *WritableDoltTable) GetAutoIncrementValue(ctx *sql.Context) (interface{}
...
+
+func (t *WritableDoltTable) getAutoIncrementValue(ctx *sql.Context) (interface{}, error) {
+	ed, err := t.getTableEditor(ctx)
+	if err != nil {
+		return nil, err
 	}
-	return t.DoltTable.GetAutoIncrementValue(ctx)
+
+	trackerVal, err := ed.aiTracker.Next(t.tableName)
+	if errors.Is(err, autoincr.ErrTableNotInitialized) {
```
zachmu, line 634: This has a race

Consider two sessions both hitting this line and error at the same time

The first one calls Init, Next

Then the second calls Init, Next

Now they both have the same key

zachmu (commented):
Basic idea is sound, I think, but you definitely have data races here.

In general the InitTable() method isn't safe to call as written.

You need additional information than what you have here, just the max value is not enough. You need to track what values have been passed out as well to avoid handing them out again

go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -434,15 +434,26 @@ func (te *pkTableEditor) InsertKeyVal(ctx context.Context, key, val types.Tuple,
...
+			}
+
 			if err != nil {
 				return err
 			}
-			if less {
+
+			if !less {
 				te.autoIncVal = types.Round(insertVal)
+				te.autoIncVal = types.Increment(te.autoIncVal)
```
zachmu, line 455: You should try to kill this field. I don't know what it could be possibly used for at this point.

Once you have, you can kill off a bunch of related methods that read and set it too.

go/libraries/doltcore/sqle/autoincr/auto_increment_tracker.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+type autoIncrementTracker struct {
+	valuePerTable map[string]interface{}
+	mu            sync.Mutex
+}
+
+var _ AutoIncrementTracker = (*autoIncrementTracker)(nil)
+
+func (a *autoIncrementTracker) Next(tableName string, insertVal interface{}, diskVal interface{}) (interface{}, error) {
```
zachmu, line 31: Do me a favor and write a unit test that spins up 100 goroutines that all call Next with nil 10 times, and asserts that none of them get the same value back

zachmu (commented):
This looks pretty good, but it needs to go farther. My claim is that the auto increment logic on the table editors themselves needs to be ripped out. It can't be correct after this change. That might not be possible, but you should try.

go/libraries/doltcore/table/editor/pk_table_editor.go
VinaiRachakonda: Used for merge logic on auto increments which still works. It'll get cleaned up in a subsequent pr

go/libraries/doltcore/sqle/sglobal/session_global.go
```diff
@@ -0,0 +1,49 @@
+package sglobal
+
+import (
+	"sync"
+)
+
+type RefInMemStore interface {
+	GetRefMemStore(refhash string) *refMemStore
+	SetRefMemStore(refhash string, store *refMemStore)
+}
```
zachmu, line 10: Don't try to expand this interface beyond the AI tracker for now. Just a single getter methods for the AI tracker on this interface, GetAutoIncrementTracker(ref.WorkingSetRef). You don't need a setter: just return a new AI tracker as needed when one isn't registered for the ref given

Call the interface GlobalState for now

go/libraries/doltcore/sqle/sglobal/session_global.go
```diff
@@ -0,0 +1,49 @@
+package sglobal
+
+import (
+	"sync"
+)
+
+type RefInMemStore interface {
+	GetRefMemStore(refhash string) *refMemStore
```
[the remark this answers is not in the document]
zachmu: "refhash" is not correct. These are not hashes as you're using them. Just pass a ref.WorkingSetRef, which is what you are actually doing

go/libraries/doltcore/sqle/sglobal/auto_increment_tracker.go
```diff
@@ -0,0 +1,135 @@
+package sglobal
+
+import (
+	"fmt"
+	"sync"
+)
+
+type AutoIncrementTracker interface {
```
zachmu, line 8: This probably should go in the sqle package, no reason to put it in the global state package

zachmu (commented):
Needs more work.

You need to do a little more due diligence with these before asking for a review. A lot of the problems I'm pointing out could have been caught by examining test behavior. And in fact I don't think this even builds, with or without your go.mod override, because of the exported / non-exported problem I pointed out

go/libraries/doltcore/sqle/sglobal/auto_increment_tracker.go
VinaiRachakonda: I honestly prefer it here, because it signals better its intended use.

integration-tests/bats/auto_increment.bats
```diff
@@ -336,3 +336,217 @@ SQL
...
+    [[ "${lines[0]}" =~ "pk,c0" ]] || false
+    [[ "${lines[1]}" =~ "1,1" ]] || false
+    [[ "${lines[2]}" =~ "2,2" ]] || false
+    [[ "${lines[3]}" =~ "3,3" ]] || false
+    [[ "${lines[4]}" =~ "4,4" ]] || false
+    [[ "${lines[5]}" =~ "5,5" ]] || false
+    [[ "${lines[6]}" =~ "6,6" ]] || false
+}
+
+@test "auto_increment: dolt_merge() with auto increment #4" {
```
zachmu, line 525: dolt_merge() lesser keys

zachmu (approved):
LGTM, thanks for the changes

integration-tests/bats/auto_increment.bats
```diff
@@ -336,3 +336,217 @@ SQL
...
+    [[ "${lines[1]}" =~ "1,1" ]] || false
+    [[ "${lines[2]}" =~ "2,2" ]] || false
+    [[ "${lines[3]}" =~ "3,3" ]] || false
+    [[ "${lines[4]}" =~ "4,4" ]] || false
+    [[ "${lines[5]}" =~ "5,5" ]] || false
+    [[ "${lines[6]}" =~ "6,6" ]] || false
+    [[ "${lines[7]}" =~ "7,7" ]] || false
+}
+
+@test "auto_increment: dolt_merge() with auto increment #2" {
```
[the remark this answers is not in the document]
VinaiRachakonda: It's to test a jump in auto increment values

PR #1885, opened 2021-07-02
Bh/push pull temptf

bheni:


go/store/datas/puller.go
```diff
@@ -166,37 +167,105 @@ func NewPuller(ctx context.Context, tempDir string, chunksPerTF int, srcDB, sink
...
 		tempDir:       tempDir,
 		wr:            wr,
 		chunksPerTF:   chunksPerTF,
 		eventCh:       eventCh,
-	}, nil
+		pushLog:       pushLogger,
+	}
+
+	lcs, ok := sinkDBCS.(interface {
+		SetLogFunc(func(string, ...interface{}))
```
zachmu, line 194: Name this interface?

go/store/datas/puller.go
```diff
@@ -166,37 +167,105 @@ func NewPuller(ctx context.Context, tempDir string, chunksPerTF int, srcDB, sink
 		return nil, err
 	}
 
-	return &Puller{
+	logFilePath := filepath.Join(tempDir, "push.log")
+	f, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.ModePerm)
+
+	var pushLogger *log.Logger
```
zachmu, line 173: Probably better to have a no-op implementation than a nil one

go/store/datas/puller.go
```diff
@@ -166,37 +167,105 @@ func NewPuller(ctx context.Context, tempDir string, chunksPerTF int, srcDB, sink
...
-func (p *Puller) processCompletedTables(ctx context.Context, ae *atomicerr.AtomicError, completedTables <-chan FilledWriters) {
-	type tempTblFile struct {
-		id          string
-		path        string
-		numChunks   int
-		contentLen  uint64
-		contentHash []byte
+func (p *Puller) log(fmt string, args ...interface{}) {
+	if p.pushLog != nil {
+		if len(args) > 0 {
```
zachmu, line 206: You don't need to check if the slice is empty for vararg expansion like this, works fine

https://play.golang.org/p/pB9g1NB6jll

go/store/datas/puller.go
```diff
@@ -211,55 +280,32 @@ func (p *Puller) processCompletedTables(ctx context.Context, ae *atomicerr.Atomi
...
 
-		if ae.SetIfError(err) {
-			return
-		}
+		fileIdToNumChunks[id] = ttf.numChunks
+	}
 
-		details.TableFilesUploaded++
-		p.eventCh <- NewTFPullerEvent(EndUpdateTableFile, details)
+	if err != nil {
```
zachmu, line 303: Add a comment about why you're returning here, looks like a typo otherwise

go/store/nbs/store.go
```diff
@@ -1299,20 +1299,25 @@ func (nbs *NomsBlockStore) WriteTableFile(ctx context.Context, fileId string, nu
...
 
 		return writeTo(f, rd, copyTableFileBufferSize)
-	}()
-
-	if err != nil {
-		return err
 	}
 
-	fileIdHash, ok := hash.MaybeParse(fileId)
+	return f()
```
zachmu, line 1304: No reason to assign this func to a var if you're just going to immediately call it

zachmu (approved):
Looks pretty good, but are you convinced this has sufficient test coverage?

PR #1886, opened 2021-07-02
keyless secondary index

max-hoffman:
- add secondary indexes to keyless tables
- biggest change adds cardinality to index value (`CardinalityCounter` in range reader)
- unit and bats tested

ex:
```
dolt sql -q "create table kl (a int, b int)"
dolt sql -q "insert into kl values (0,0), (1,1)"
dolt sql -q "create index idx on kl (a)"
```

```
> dolt sql -q "show create table kl1"
+-------+--------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                         |
+-------+--------------------------------------------------------------------------------------------------+
| kl1   | CREATE TABLE `kl1` (
  `a` int,
  `b` int,
  KEY `idx` (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 |
+-------+--------------------------------------------------------------------------------------------------+
```

```
> dolt sql -q "select * from kl where a = 0"
+---+---+
| a | b |
+---+---+
| 0 | 0 |
+---+---+
```


andrew-wm-arthur (approved):
LGTM
This does a nice job of reusing the pieces that were already in place. I would lookout for places where we assume Primary Keys. We also need to add engine tests that test index functionality and assert that indexes are used in query plans

andrew-wm-arthur:
Just realized that this doesn't touch the keyless table editor at all, have you tested this for inserts?

go/libraries/doltcore/sqle/dolt_map_iter.go
```diff
@@ -162,7 +162,8 @@ func (conv *KVToSqlRowConverter) processTuple(cols []interface{}, valsToFill int
 		}
 
 		tag64 := primReader.ReadUint()
-		if tag64 > maxTag {
+
+		if tag64 > maxTag && tag64 != schema.KeylessRowCardinalityTag && tag64 != schema.KeylessRowIdTag {
```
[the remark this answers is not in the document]
max-hoffman: Keyless index lookups get called with the other path, which sets the largest tag in the column values. So this fails to account for keyless tags that are bigger.  This is also true for your suggested change in index_row_iter.go: `if inPK || (tag == schema.KeylessRowIdTag) {`. Row lookups silently fail in a confusing way without these extra checks. An alternative solution would refactor upstream logic to choose better defaults for keyless rows, but this felt like a smaller change when i was walking through the code. Open to other suggestions, but much of the function stack leading up to row reads make assumptions that aren't true for keyless and I was trying to avoid a bunch of `if sch,IsKeyless { callNewSpecialFunc }`

go/libraries/doltcore/table/editor/keyless_table_editor.go
```diff
@@ -343,30 +354,85 @@ func applyEdits(ctx context.Context, tbl *doltdb.Table, acc keylessEditAcc) (*do
...
 		} else {
 			v, ok, err = modifyCardinalityWithDelta(v, delta.delta)
+
 		}
 		if err != nil {
 			return nil, err
 		}
 
+		indexOpsToUndo := make([]int, len(indexEds))
+		defer func() {
```
Hydrocharged, line 371: Be wary of putting defers inside of for loops. They are called at function boundaries and not at block boundaries, so you have to use an anonymous function as the body of your for loop to get the desired behavior (or break out the for loop body into its own function).

Hydrocharged (changes requested):
Overall this looks good, but I do have a few points that I want addressed. When they're done ping me and I'll review the new changes. We can also discuss them if you disagree with any of them.

go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -100,6 +102,7 @@ type indexEditAccumulator struct {
 	addedPartialKeys map[hash.Hash]map[hash.Hash]types.Tuple
 	addedKeys        map[hash.Hash]hashedTuple // These hashes represent the hash of the partial key, with the tuple being the full key
 	removedKeys      map[hash.Hash]hashedTuple // These hashes represent the hash of the partial key, with the tuple being the full key
+	addedValues      map[hash.Hash]types.Tuple
```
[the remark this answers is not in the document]
max-hoffman: :+1 yeah that's a good idea

max-hoffman:
@Hydrocharged There's a trade-off between re-writing a lot of code and re-using existing code with a couple extra if's to account for keyless tags. This dichotomy is at the heart of most of your comments. I'm a bit conflicted given how simple the changes with just re-using the pk code most of the time. Idk how difficult duplicating everything for keyless will be, also. Either way may switches are not self-explanatory, which is a problem. Maybe we can talk through tomorrow, also considering that Brian broke a bunch of this code in a PR merged today.

Hydrocharged:
I don't necessarily mean to duplicate the majority of the code that's present, but that I think it'd be far clearer where each if is:
```go
if keylessRow, ok := r.(row.KeylessRow); ok {
	// stuff
} else if keyedRow, ok := r.(row.KeyedRow); ok {
	// stuff
} else {
	// stuff
}
```
In addition to this, most of the code assumes a primary key (as mentioned in some of Andy's earlier feedback), where it may even just be the variable names that need changing. Having a variable named `pkTags` and then having logic for a keyless row is not clear, where as just a variable named `selectedTags` or something is clearer in the context of two different row types using the same pre-existing logic.

What I am suggesting would increase the amount of code needed _in some places_, but I kind of point to my earlier comment that we'll _"inevitably end up having split logic dealing with the different paths"_. The split may be as small as my example above, or as large as splitting `IndexEditor` into two just like we have with the table editors. I can imagine that optimizations further down the line may necessitate splitting `IndexEditor` or some other construct, and it'll be far cleaner if we've got if or switch statements already in place rather than attempting to see exactly what is necessary and what isn't.

go/libraries/doltcore/table/editor/keyless_table_editor.go
max-hoffman: subtle! Are you saying that it's preferable to put it in a separate closure? or will it still perform correctly if I move it outside of the for loop and `Undo` all of `applyEdits` at the same time?

max-hoffman: After re-reading, I suppose we want the granularity of a different closure in this case

Hydrocharged: Yep! Separate closure for the `for` body, or at least all of the logic following the defer up until the closing block of the `for` body.

go/libraries/doltcore/row/keyless_row.go
```diff
@@ -264,3 +264,52 @@ func (r keylessRow) TaggedValues() (TaggedValues, error) {
 func (r keylessRow) Format() *types.NomsBinFormat {
 	return r.val.Format()
 }
+
+// ReduceToIndexKeys creates a full key, a partial key, and a cardinality value from the given row
+// (first tuple being the full key). Please refer to the note in the index editor for more information
+// regarding partial keys.
+func (r keylessRow) ReduceToIndexKeys(idx schema.Index) (types.Tuple, types.Tuple, types.Tuple, error) {
```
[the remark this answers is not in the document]
max-hoffman: cool if I punt on this for now? This will need to be rewritten w/ taggedValues as the main interface for extracting indexes. I also think keyed rows being explicit about an empty value tuple is clear and correcy, I wasted a day on a strange bug b/c `types.Tuple{}` is not nbf compatible, but `types.EmptyTuple()` is.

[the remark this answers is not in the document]
Hydrocharged: It's fine to punt on `ReduceToIndexKeys`, and thinking about it more it's fine that keyed rows are explicit in returning an empty value. If possible I'd still like to have the different interfaces for them both, even if it's through the implementation of a private function to mark the explicit distinction between the two.

integration-tests/bats/keyless.bats
```diff
@@ -31,15 +31,15 @@ teardown() {
...
 
-    run dolt sql -q "CREATE TABLE bad (a int, b int, INDEX (b));"
-    [ $status -ne 0 ]
+    run dolt sql -q "CREATE TABLE fine (a int, b int, INDEX (b));"
+    [ $status -eq 0 ]
     [[ ! "$output" =~ "panic" ]] || false
 
     run dolt sql -q "CREATE TABLE worse (a int, b int, FOREIGN KEY (b) REFERENCES keyless(c1));"
-    [ $status -ne 0 ]
+    [ $status -eq 0 ]
```
Hydrocharged, line 42: Foreign keys don't work properly. I pulled this branch and tried out a deletion with a `CASCADE` reference option and it's throwing an error. We should probably get foreign keys solved before putting this in, just in case we have to change the way keyless index maps are stored.

Hydrocharged (changes requested):
Foreign keys aren't working. Normally I'd say to just check if the table is keyless and prevent foreign keys so we can punt, but as foreign keys are fundamentally tied to how indexes are implemented, it's possible that the index map on keyless tables may have to change. It's far harder to have to migrate and bump the feature version than to fix it now. Hopefully it's something quick and easy.

PR #1888, opened 2021-07-02
Max/dsess

max-hoffman:
- DoltSession moved to new package, `sqle/dsess`
- Renamed `dsess.Session` at most points of reference
- Exposes necessary Session and State variables as public
- Exposes certain constats (like `batched -> Batched`) as constant


zachmu:
Merged into my branch

PR #1890, opened 2021-07-02
Enable SQL transactions by default, and begin writing working set to database instead of repo_state.json file

zachmu:
This change includes migration code to initialize the working set of existing databases that lack a working set db value at runtime.

It begins to remove the repo state writer / reader abstraction for setting working set changes, but doesn't completely eliminate it. An additional refactoring pass will eliminate it completely once this is in.

It introduces two new doltdb domain objects:

WorkingSet: contains the root values and merge state for a working set
Roots: a simple container object to wrap up the head, working, and staged roots to enable library code that modifies the working set to be purely functional


zachmu:
#benchmark

reltuk (approved):
I've taken a look at the general approach, the new domain models and their interactions with `datas`, `doltdb` and `actions`. I think we can roll forward here and smooth out the still-rough edges against while the merge is fresh.

PR #1891, opened 2021-07-02
dolt roots command

bheni:


reltuk (approved):
LGTM! Doesn't look too hard to make nbs.IterChunks take a `ReaderAt` instead and avoid having to have the entire table file in memory, but that's not a show stopper for this first pass I don't think.

timsehn:
Brian is gonna make me write the bats tests for this isn't he?

reltuk (approved):
LGTM!

PR #1892, opened 2021-07-02
fix: infer_schema treat leading zero numbers as strings

coffeegoddd:


go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -224,6 +224,12 @@ func leastPermissiveNumericType(strVal string, floatThreshold float64) (ti typei
 		if err != nil {
 			return typeinfo.UnknownType
 		}
+
+		// handle leading zero case
+		if strVal[0] == '0' {
+			return typeinfo.StringDefaultType
```
coffeegoddd, line 230: @bheni returning string here since returning `typeinfo.UnknownType` will cause `strVal` to fall through to the chrono type https://github.com/dolthub/dolt/blob/master/go/libraries/doltcore/env/actions/infer_schema.go#L157

bheni: I think you need to check for a following decimal point.  Like 0.5 should be a float.  I'd add that, and then add a test.

bheni (changes requested):

bheni (approved):
After further discussion I'm fine with this.  For the record the issue is with strings like "05.78" which will be inferred to be a float with the value 5.78 which may not be what we want... but we'll address this later if it turns out we want it to be encoded as a string.

coffeegoddd:
@bheni added a test that shows our current behavior with leading zero floats like `05.78` for now https://github.com/dolthub/dolt/pull/1892/files#diff-c698c52fc13b45586221717e0cde3b5b8c1c17fa7434984b4248ddbbb892f47bR83

PR #1893, opened 2021-07-02
go/libraries/doltcore/doltdb: Prohibit deleting the last branch ref in the dataset.

reltuk:


go/libraries/doltcore/rebase/rebase.go
```diff
@@ -154,11 +154,10 @@ func rebaseRefs(ctx context.Context, dbData env.DbData, replay ReplayCommitFn, n
 
 		switch dRef.(type) {
 		case ref.BranchRef:
-			err = ddb.DeleteBranch(ctx, dRef)
+			err = ddb.NewBranchAtCommit(ctx, dRef, newHeads[i])
```
bheni, line 157: on line 167 below I don't think that error check is needed anymore.

bheni (approved):
lgtm

PR #1913, opened 2021-07-14
Additional Constraint Violation Fixes

Hydrocharged:
This PR has a few things. Most notably, we no longer allow fast forward merges if a merge may cause a constraint violation. This decision was discussed internally. Additionally, I've modified the command line messages to properly respond to the presence of constraint violations, which includes blocking some functionality until they have been resolved. Lastly are general fixes and whatnot.

### DIFF
--- a/go/cmd/dolt/commands/merge.go
+++ b/go/cmd/dolt/commands/merge.go
@@ -117,25 +117,30 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string, 
 			}
 
 			// If there are any conflicts or constraint violations then we disallow the merge
-			if has, err := root.HasConflicts(ctx); err != nil {
+			hasCnf, err := root.HasConflicts(ctx)
+			if err != nil {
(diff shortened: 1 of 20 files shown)


go/cmd/dolt/commands/merge.go
```diff
@@ -224,6 +229,19 @@ func mergeCommitSpec(ctx context.Context, apr *argparser.ArgParseResults, dEnv *
...
+		if err != nil {
+			return errhand.VerboseErrorFromError(err)
+		}
+		mergedRoot, err := cm2.GetRootValue()
+		if err != nil {
+			return errhand.VerboseErrorFromError(err)
+		}
+		if cvPossible, err := merge.MayHaveConstraintViolations(ctx, ancRoot, mergedRoot); err != nil {
+			return errhand.VerboseErrorFromError(err)
+		} else if cvPossible {
```
bheni, line 242: this could result in an ff when apr.Contains(cli.NoFFParam) is true.

bheni (approved):
looks good.  just one item that looks like an issue.

PR #1914, opened 2021-07-14
Vinai/remove table editor value

VinaiRachakonda:
Removes the autoInc val from the table editor and reads directly from disk.


VinaiRachakonda:
**10,000 inserts**

Before: 1.07m
After: 1.04m

**100,000 inserts**
Before: 18:25
After: 15:40.0

PR #1919, opened 2021-07-14
[no-release-notes] go/libraries/doltcore/doltdb: Add utility method for getting refs with their target hashes.

reltuk:


bheni (commented):
lgtm

PR #1920, opened 2021-07-15
Further attempts to fix ACCESS_DENIED error on Windows

Hydrocharged:
This attempts to fix the additional errors seen in the following issue:

https://github.com/dolthub/dolt/issues/1774

We've previously released a PR (https://github.com/dolthub/dolt/pull/1811) that seemed successful at _"fixing"_ an `ACCESS DENIED` error that would occur on table import when GC was running and the manifest was being renamed. That PR, however, did not touch any other rename within the codebase, nor any deletions (which can also trigger this bug). The above mentioned issue points to deletions, therefore this PR attempts to fix those plus any others by implementing the same strategy employed by the manifest renames and applying it to every deletion and rename in the project.

I also investigated the linked issues inside of the above mentioned issue (https://github.com/golang/go/issues/32088 and https://github.com/golang/go/issues/34681). The sharemode issue does not seem to be the exact issue we are seeing here, as adapting the fix repository (https://github.com/alexbrainman/goissue34681) into our code and having every file creation and opening use the updated syscalls did not result in any improvement.

### DIFF
--- a/go/cmd/dolt/system_checks.go
+++ b/go/cmd/dolt/system_checks.go
@@ -19,9 +19,9 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/dolthub/dolt/go/store/util/tempfiles"
-
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/utils/file"
+	"github.com/dolthub/dolt/go/store/util/tempfiles"
 )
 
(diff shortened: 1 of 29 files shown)


Hydrocharged:
Also I did test this the same way that I tested the previous PR (https://github.com/dolthub/dolt/pull/1811), which was by forcing GC to run after every line on a super long import.

reltuk (approved):
LGTM!

PR #1924, opened 2021-07-16
[no-release-notes] go/store/datas: FindCommonAncestor uses heap.

reltuk:


reltuk:
Local benchmarking for the bounties use case shows this is much faster than the current implementation. Took 160 seconds on 7973 branches against the current `menus/master`. Against the same dataset, the old implementation completes 262 lookups in 160 seconds.

bheni (approved):
dope as hell

PR #1930, opened 2021-07-19
No-op bad alter table auto increment

VinaiRachakonda:
Pr does 2 things

1. Fix panic on linked issue
2. Actually allow you to modify a column to add an AI


go/libraries/doltcore/sqle/table_editor.go
```diff
@@ -139,6 +139,11 @@ func (te *sqlTableEditor) GetAutoIncrementValue() (interface{}, error) {
 }
 
 func (te *sqlTableEditor) SetAutoIncrementValue(ctx *sql.Context, val interface{}) error {
+	// In the case that someone runs an alter table auto increment on a table with no existing auto increment key, nop
+	if te.autoIncCol.TypeInfo == nil {
+		return nil
+	}
```
andrew-wm-arthur, line 145: we should no-op in the SQL engine, before we get here

go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -149,7 +150,9 @@ func updateTableWithModifiedColumn(ctx context.Context, tbl *doltdb.Table, oldSc
 	var autoVal types.Value
 	if schema.HasAutoIncrement(newSch) {
 		autoVal, err = tbl.GetAutoIncrementValue(ctx)
-		if err != nil {
+		// In the case that the alter table is adding a new auto increment value, then it will return a
+		// ErrorNoAutoIncrement value. We can ignore that.
+		if err != nil && !errors.Is(err, doltdb.ErrNoAutoIncrementValue) {
```
andrew-wm-arthur, line 155: inspect the old/new schemas and catch the case where `auto_increment` has been added to a column

go/libraries/doltcore/schema/alterschema/modifycolumn.go
```diff
@@ -147,7 +147,7 @@ func updateTableWithModifiedColumn(ctx context.Context, tbl *doltdb.Table, oldSc
 		return nil, err
 	}
 	var autoVal types.Value
-	if schema.HasAutoIncrement(newSch) {
+	if schema.HasAutoIncrement(newSch) && schema.HasAutoIncrement(oldSch) {
```
andrew-wm-arthur, line 150: this looks much better, could you add a comment explaining the alter table case? Also, what happens if the `AUTO_INCREMENT` is removed from the schema?

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/schema/alterschema/modifycolumn.go
VinaiRachakonda: The new schema would not have auto increment an a nil value would be passed for autoVal

timsehn:
Change the Title from "Nop" to "No-op" please

PR #1931, opened 2021-07-20
[no-release-notes] List portion of refs

tbantle22:


reltuk:
We will revisit this and thread it through service interfaces if it becomes necessary. As written I think a client with a query would have trouble distinguishing exhausting iteration budget vs. there being no results...so implementation could be improved slightly. But as is, non-paginated interfaces should scale to pretty large branch/dataset sets...

PR #1935, opened 2021-07-21
fix initial gcGen and exit gc early if there is nothing to do

bheni:


go/cmd/dolt/commands/gc.go
```diff
@@ -122,7 +126,11 @@ func (cmd GarbageCollectionCmd) Exec(ctx context.Context, commandStr string, arg
 
 		err = dEnv.DoltDB.GC(ctx, keepers...)
```
andrew-wm-arthur, line 127: we should remove the `keepers` parameter if we don't use it

andrew-wm-arthur (approved):
LGTM

PR #1938, opened 2021-07-22
Bh/gen cs

bheni:


andrew-wm-arthur:
#benchmark

go/store/nbs/generational_chunk_store.go
```diff
@@ -0,0 +1,302 @@
...
+	return gcs.newGen
+}
+
+func (gcs *GenerationalNBS) OldGen() chunks.ChunkStoreGarbageCollector {
+	return gcs.oldGen
+}
+
+// Get the Chunk for the value of the hash in the store. If the hash is absent from the store EmptyChunk is returned.
+func (gcs *GenerationalNBS) Get(ctx context.Context, h hash.Hash) (chunks.Chunk, error) {
+	c, err := gcs.oldGen.Get(ctx, h)
```
andrew-wm-arthur, line 56: I wonder about looking in `oldGen` before `newGen`. Presumably, `newGen` is much smaller than `oldGen`. 

A lookup miss in `newGen` followed by a successful lookup in `oldGen` would be faster than a lookup miss in `oldGen` followed by a successful lookup in `newGen`. What's not clear is the relative lookup frequency in new vs old. If the `newGen` contains any chunks, one of them must be the dataset map, and in that case chunks in `oldGen` must be accessed through refs stored in chunks in `newGen`.

My guess is that it's probably faster overall to check `newGen` first

go/store/nbs/store.go
```diff
@@ -1406,22 +1412,42 @@ func (nbs *NomsBlockStore) MarkAndSweepChunks(ctx context.Context, last hash.Has
...
-	err = nbs.swapTables(ctx, specs)
+	specs, err := nbs.copyMarkedChunks(ctx, keepChunks, destNBS)
 	if err != nil {
 		return err
 	}
 	if ctx.Err() != nil {
 		return ctx.Err()
 	}
 
+	if destNBS == nbs {
```
andrew-wm-arthur, line 1433: should add some clarifying comments here. The first half of the block is doing an old-style full-copy GC, the second half is moving chunks from the newGen to oldGen

go/store/nbs/store_test.go
```diff
@@ -263,7 +263,7 @@ func TestNBSCopyGC(t *testing.T) {
 	wg := &sync.WaitGroup{}
 	wg.Add(1)
 	go func() {
-		msErr = st.MarkAndSweepChunks(ctx, r, keepChan)
+		msErr = st.MarkAndSweepChunks(ctx, r, keepChan, nil)
```
andrew-wm-arthur, line 266:
```suggestion
		msErr = st.MarkAndSweepChunks(ctx, r, keepChan, st)
```

andrew-wm-arthur: seems like it shouldn't be nil?

go/store/nbs/generational_chunk_store.go
```diff
@@ -0,0 +1,302 @@
...
+	})
+
+	if putErr != nil {
+		return putErr
+	}
+
+	return err
+}
+
+func (gcs *GenerationalNBS) MarkAndSweepChunks(ctx context.Context, last hash.Hash, keepChunks <-chan []hash.Hash, dest chunks.ChunkStore) error {
```
andrew-wm-arthur, line 227: this type doesn't need to satisfy the `GarbageCollector` interface, we can delete this

go/store/types/value_store.go
```diff
@@ -658,6 +670,48 @@ func (lvs *ValueStore) GC(ctx context.Context) error {
...
+				return errors.New("dangling reference found in chunk store")
+			}
+
+			hashes, err := walker.GetRefSet(visited, vals)
+			if err != nil {
+				return err
+			}
+
+			if oldGen != nil {
+				hashes, err = oldGen.HasMany(ctx, hashes)
```
andrew-wm-arthur, line 702: comment here to explain that we stop traversing if our path leads into the old generation

andrew-wm-arthur (approved):
LGTM!

go/libraries/doltcore/dbfactory/file.go
```diff
@@ -52,19 +52,50 @@ func (fact FileFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat,
...
-	st, err := nbs.NewLocalStore(ctx, nbf.VersionString(), path, defaultMemTableSize)
+	oldgenPath := filepath.Join(path, "oldgen")
+	err = validateDir(oldgenPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, err
+		}
+
+		err = os.Mkdir(oldgenPath, os.ModePerm)
+		if err != nil {
```
reltuk, line 74: If we're racing, check for already exists and just continue here?

```suggestion
		if err != nil && !errors.Is(err, os.ErrExist) {
```

go/libraries/doltcore/dbfactory/file.go
```diff
@@ -52,19 +52,50 @@ func (fact FileFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat,
...
 		return nil, err
-	} else if !info.IsDir() {
-		return nil, filesys.ErrIsFile
 	}
 
-	st, err := nbs.NewLocalStore(ctx, nbf.VersionString(), path, defaultMemTableSize)
+	oldgenPath := filepath.Join(path, "oldgen")
+	err = validateDir(oldgenPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
```
reltuk, line 69: I think

```suggestion
		if !errors.Is(err, os.ErrNotExist) {
```

is slightly preferred.

go/store/nbs/generational_chunk_store.go
```diff
@@ -0,0 +1,302 @@
...
+
+	return c, nil
+}
+
+// GetMany gets the Chunks with |hashes| from the store. On return, |foundChunks| will have been fully sent all chunks
+// which have been found. Any non-present chunks will silently be ignored.
+func (gcs *GenerationalNBS) GetMany(ctx context.Context, hashes hash.HashSet, found func(*chunks.Chunk)) error {
+	notInOldGen := hashes.Copy()
+	err := gcs.oldGen.GetMany(ctx, hashes, func(chunk *chunks.Chunk) {
+		delete(notInOldGen, chunk.Hash())
```
reltuk, line 74: Are `GetMany` callbacks serialized or can they happen concurrently?

go/store/nbs/generational_chunk_store.go
```diff
@@ -0,0 +1,302 @@
...
+// Returns the NomsVersion with which this ChunkSource is compatible.
+func (gcs *GenerationalNBS) Version() string {
+	return gcs.newGen.Version()
+}
+
+// Rebase brings this ChunkStore into sync with the persistent storage's
+// current root.
+func (gcs *GenerationalNBS) Rebase(ctx context.Context) error {
+	oErr := gcs.oldGen.Rebase(ctx)
+	nErr := gcs.newGen.Rebase(ctx)
```
reltuk, line 156: How do we think about the order here? We opened `new` before `old` previously, and here we're rebasing `old` and then `new`. What's the actual order we want? What conditions and invariants can we reason about so we can be certain we have a consistent unified view of old+new after we open or Rebase?

go/store/types/value_store.go
reltuk, line 702: I would change the interface here and make it the responsibility of the caller to pass in something that could filter hashes as we're talking them to short circuit the walk. GenerationalChunkStore or NBSStore with a GenerationalChunkStore would end up passing an implementation of that thing that basically did `oldGen.HasMany`, but this layer wouldn't explicitly know about old gen.

reltuk (commented):
Generally looks great. Some comments.

go/store/nbs/generational_chunk_store.go
bheni: serialized

go/store/nbs/generational_chunk_store.go
bheni: good call

go/store/nbs/store_test.go
bheni: ?  it's not nil.

timsehn:
Can I get a better PR title on this before we merge it?

go/store/nbs/store_test.go
andrew-wm-arthur: the last argument to `MarkAndSweepChunks` is `nil`

PR #1947, opened 2021-07-28
Change force commit to bypass conflicts and constraint violations

Hydrocharged:
Previously `--force` only bypassed invalid foreign keys. With the addition of constraint violations, this check has been rendered redundant, and `--force` seems like it should bypass all blockers on commit anyway.


Hydrocharged:
No tests were updated, nor has this been extensively tested. Just some local tests, primarily to unblock.

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -86,6 +86,14 @@ func init() {
 			Type:              sql.NewSystemBoolType(TransactionsDisabledSysVar),
 			Default:           int8(0),
 		},
+		{
```
andrew-wm-arthur, line 89: Somewhere we should put a comment explaining what this var does, this seems like a good place

andrew-wm-arthur (approved):
LGTM!

PR #1957, opened 2021-07-29
Allow setting head ref session var, e.g. @@mydb_head_ref to change branches

zachmu:
Also fixed a bug where it wasn't being updated on a branch change


andrew-wm-arthur (approved):
LGTM

PR #1960, opened 2021-07-31
Find common ancestor using transitive ref closure

andrew-wm-arthur:


reltuk:
This is quite great. For the workload we have, this makes the work to get through all 40,666 branches basically trivial. The ~16 second run is spent almost completely resolving the head commits of the branches themselves. Not actually clear if the parent+decoded ref cache is even worth it, although those will still help in other less batch-y use cases...

PR #1964, opened 2021-08-02
Add bats for adding and dropping pks

VinaiRachakonda:


VinaiRachakonda:
TODO: Needs a test for foreign keys

PR #1965, opened 2021-08-03
Fixed panic when merging a branch that still contains a table that was deleted in the target

zachmu:
Also some other panics, and added more merge tests. Cleaned up merge short circuit code, and clarified the behavior when one branch edits a table another deletes during a merge.

### DIFF
--- a/go/libraries/doltcore/merge/merge.go
+++ b/go/libraries/doltcore/merge/merge.go
@@ -877,9 +877,9 @@ func MergeRoots(ctx context.Context, ourRoot, theirRoot, ancRoot *doltdb.RootVal
 			if err != nil {
 				return nil, nil, err
 			}
-		} else if has, err := newRoot.HasTable(ctx, tblName); err != nil {
+		} else if newRootHasTable, err := newRoot.HasTable(ctx, tblName); err != nil {
 			return nil, nil, err
-		} else if has {
+		} else if newRootHasTable {
 			tblToStats[tblName] = &MergeStats{Operation: TableRemoved}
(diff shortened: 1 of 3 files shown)


zachmu:
This fixes #1966

go/libraries/doltcore/merge/merge.go
```diff
@@ -892,7 +901,11 @@ func MergeRoots(ctx context.Context, ourRoot, theirRoot, ancRoot *doltdb.RootVal
 				return nil, nil, err
 			}
 		} else {
-			panic("?")
+			// This is a deleted table that the merge root still has
+			if stats.Operation != TableRemoved {
+				panic("Invalid merge state. This is a bug.")
```
andrew-wm-arthur, line 906: let's put the table name into this panic message

go/libraries/doltcore/merge/merge.go
```diff
@@ -859,7 +868,6 @@ func MergeRoots(ctx context.Context, ourRoot, theirRoot, ancRoot *doltdb.RootVal
 	// need to validate merges can be done on all tables before starting the actual merges.
 	for _, tblName := range tblNames {
```
andrew-wm-arthur, line 869: One lingering issue is we still match tables by name, rather than by tags, across merge roots. So renamed tables will not be handled correctly

andrew-wm-arthur (approved):
lgtm

go/libraries/doltcore/merge/merge.go
zachmu: Added a note and TODO

PR #1967, opened 2021-08-03
Edit dolt merge ret type again 

VinaiRachakonda:


andrew-wm-arthur:
We should add some integrations tests for this. Take a look at https://github.com/dolthub/dolt/blob/76212aae15ac47447626cbab8fb1b2cebe8aeb12/go/libraries/doltcore/sqle/integration_test/history_table_test.go#L79

PR #1976, opened 2021-08-04
Cleanup from transaction migration

zachmu:
First batch of cleanup, getting rid of uses of RepoStateWriter as much as possible.

To push this farther requires new abstractions that can handle updating the HEAD, head commit, and working roots at the same time, which is the next step.


go/libraries/doltcore/dtestutils/testcommands/command.go
```diff
@@ -48,7 +48,10 @@ func (a StageAll) Exec(t *testing.T, dEnv *env.DoltEnv) error {
 	roots, err := dEnv.Roots(context.Background())
```
andrew-wm-arthur, line 48: we need to kill off this package, I added `testcommands` before realizing we could just use the actual commands in tests

go/libraries/doltcore/env/actions/staged.go
```diff
@@ -16,69 +16,44 @@ package actions
...
-		return err
-	}
-
+func StageTables(ctx context.Context, roots doltdb.Roots, docs doltdocs.Docs, tbls []string) (doltdb.Roots, error) {
 	if len(docs) > 0 {
+		var err error
 		roots.Working, err = doltdocs.UpdateRootWithDocs(ctx, roots.Working, docs)
 		if err != nil {
-			return err
+			return doltdb.Roots{}, nil
```
andrew-wm-arthur, line 29:
```suggestion
			return doltdb.Roots{}, err
```

andrew-wm-arthur: we should added Lars' linter

go/libraries/doltcore/env/environment.go
```diff
@@ -1016,3 +944,59 @@ func (dEnv *DoltEnv) GetUserHomeDir() (string, error) {
 func (dEnv *DoltEnv) TempTableFilesDir() string {
 	return mustAbs(dEnv, dEnv.GetDoltDir(), tempTablesDir)
 }
+
+// GetGCKeepers returns the hashes of all the objects in the environment provided that should be perserved during GC.
+func GetGCKeepers(ctx context.Context, env *DoltEnv) ([]hash.Hash, error) {
```
andrew-wm-arthur, line 949: I think that we don't need this anymore. Working and stage are stored in a `dataset` now, which should protect them from GC. That said, I think we should make that a separate change

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/dtestutils/testcommands/command.go
zachmu: added TODO

go/libraries/doltcore/env/actions/staged.go
zachmu: fixed

go/libraries/doltcore/env/environment.go
zachmu: Added a todo

PR #1978, opened 2021-08-05
[no-release-notes] go/store/types: value_store.go: Add some sanity checks on GC asserting that bufferedChunks is empty at beginning and end.

reltuk:


zachmu (approved):
LGTM

PR #1980, opened 2021-08-05
[no-release-notes] go/store/nbs: Improve manifest lock handling.

reltuk:
Currently `NBSStore.UpdateManifest` and `NBSStore.UpdateManifestWithAppendix` are not updating the manifest lock when they change the contents of the manifest. This causes concurrency control race-y writes in doltremoteapi and we can drop manifest table file appends in some cases.

The lock contents should always reflect the full mutable contents of the manifest, so we added the appendix spec to the contents that are used to generate the lock.

Also takes a pass at improving `MarkAndSweep` implementation to respect `NBSStore.mu` and correctly handle the case where upstream manifest contents changed while the GC was running.


bheni (approved):
lgtm

PR #1984, opened 2021-08-06
integration-tests/bats/runner: Start sketching a small harness to run bats tests in parallel.

reltuk:


VinaiRachakonda:
This was fun to read!

PR #1986, opened 2021-08-06
Andy/auto increment refactor

andrew-wm-arthur:


zachmu (commented):
I think this is fine as long as we re-use the same database object for every HEAD, which we currently do.

@VinaiRachakonda do you have thoughts on this?

zachmu (approved):

PR #1988, opened 2021-08-07
Vinai/alter drop pks

VinaiRachakonda:


go/libraries/doltcore/merge/merge_schema.go
```diff
@@ -28,6 +28,7 @@ type conflictKind byte
 const (
 	TagCollision conflictKind = iota
 	NameCollision
+	PkCollision
```
VinaiRachakonda, line 31: Note to self: Come back to this

go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -349,6 +370,76 @@ func getKeylessDeltas(ctx context.Context, fromRoot, toRoot *doltdb.RootValue) (
...
+		return nil, nil, err
+	}
+
+	err = toRoot.IterTables(ctx, func(name string, tbl *doltdb.Table, sch schema.Schema) (stop bool, err error) {
+		delta := TableDelta{
+			ToName:  name,
+			ToTable: tbl,
+		}
+
+		// TODO: is this the most efficient way to to do this
```
VinaiRachakonda, line 401: Note to self: Come back to this

go/libraries/doltcore/schema/column.go
```diff
@@ -149,8 +149,7 @@ func (c Column) Equals(other Column) bool {
 // kinds of values at the storage layer, but may have different constraints or type parameters.
 func (c Column) Compatible(other Column) bool {
 	return c.Tag == other.Tag &&
-		c.Kind == other.Kind &&
-		c.IsPartOfPK == other.IsPartOfPK
```
andrew-wm-arthur, line 153: This is suspect, what's the reasoning for this?

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -368,6 +373,11 @@ func (dp diffPartition) getRowIter(ctx *sql.Context, ddb *doltdb.DoltDB, ss *sch
 	toCmInfo := commitInfo{types.String(dp.toName), dp.toDate, toCol.Tag, toDateCol.Tag}
 
 	rd := diff.NewRowDiffer(ctx, fromSch, toSch, 1024)
+	if _, ok := rd.(*diff.EmptyRowDiffer); ok {
+		ctx.Warn(PrimaryKeyChanceWarningCode, fmt.Sprintf(PrimaryKeyChangeWarning, dp.fromName, dp.toName))
+		return sql.RowsToRowIter(), nil
+	}
+
```
andrew-wm-arthur, line 380: Why do we check this here? Where do we return `EmptyRowDIffer`?

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -551,6 +579,23 @@ func (dp *diffPartitions) Next() (sql.Partition, error) {
 		}
 
 		if next != nil {
+			// We want to render the dolt_diff table from a top down commit log traversal until we reach a point
+			// where we cannot diff two commits due to a schema difference. This logic is similar to whats in
+			// dp.getRowIter() for two reasons.
+			// 1. The dolt_commit_diff table exclusively uses getRowIter and not the diffPartitions object
+			// 2. The GMS partition exchange does not process partitions in the order they are sent in. This can allow
+			// commits with the computable diffs before the schema change to be rendered, convoluting the UX of the diff
+			// table.
```
andrew-wm-arthur, line 588: It's a good idea to document this facet of the diff table, but this comment is too implementation-oriented and won't make sense to the next person reading this code. See comment suggestions.

andrew-wm-arthur:
```suggestion
// if we can't diff this commit with its parent, don't traverse any lower
```

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -383,6 +393,24 @@ func (dp diffPartition) getRowIter(ctx *sql.Context, ddb *doltdb.DoltDB, ss *sch
 	}, nil
 }
 
+func (dp *diffPartition) isDiffablePartition(ctx *sql.Context) (bool, error) {
```
andrew-wm-arthur, line 396:
```suggestion
// isDiffablePartition checks if the commit pair for this partition is "diffable".
// If the primary key sets changed between the two commits, it may not be
// possible to diff them.
func (dp *diffPartition) isDiffablePartition(ctx *sql.Context) (bool, error) {
```

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1720,3 +1722,255 @@ func (t *AlterableDoltTable) constraintNameExists(ctx *sql.Context, name string)
 
 	return false, nil
 }
+
```
andrew-wm-arthur, line 1725: the rest of the schema change logic is located in `schema/alterschema` we should move this logic there and call to it

andrew-wm-arthur: We should also add some golang integration test for rewriting primary and secondary indexes

integration-tests/bats/schema-changes.bats
```diff
@@ -254,3 +254,452 @@ SQL
...
+
+    dolt sql -q "INSERT INTO t values (2,2,2)"
+    dolt commit -am "cm3"
+
+    run dolt merge test
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ 'different primary key definitions for our column pk and their column pk' ]] || false
+}
+
+@test "schema-changes: same as the previous test but in the opposite direction" {
```
andrew-wm-arthur, line 502: needs a more descriptive title, tests can be inserted between these two

integration-tests/bats/schema-changes.bats
```diff
@@ -254,3 +254,452 @@ SQL
     [[ "$output" =~ '>   `v2` VARCHAR(1024)' ]] || false
     [[ "$output" =~ 'PRIMARY KEY' ]] || false
 }
+
```
andrew-wm-arthur, line 257: nit: great test coverage here, but there are so many new ones, you should split them out into their own file 
`primary-key-change.bats`

andrew-wm-arthur (commented):
Generally looks good. 
This covers the major dimensions of the product surface area and has thorough test coverage.
We should discuss what this means for SQL diff and import, and see what's worth pursuing there.

go/libraries/doltcore/schema/column.go
VinaiRachakonda: The superschema will complain about a tag collision otherwise during. [cc:](https://github.com/dolthub/dolt/blob/83b4aaaf7413f9e49ff8c6c037b12888fe8b95c1/go/libraries/doltcore/schema/super_schema.go#L67)

I also don't think it's correct anymore since primary key changes are possible.

go/libraries/doltcore/sqle/dtables/diff_table.go
VinaiRachakonda: We return empty row differ async_differ.go when we call `NewRowDiffer`. We check to primarily throw a warning for users. The emptyRowDiffer is useful in reducing the conditional logic checks of schema changes since it just returns an empty diff. Cc. dolt constraints verify

go/libraries/doltcore/sqle/tables.go
VinaiRachakonda: 1. Agree with that refactor
2. Not sure what we're looking for in the integration test. Is it to better assert contents in the index?

integration-tests/bats/schema-changes.bats
VinaiRachakonda: Fair enough haha

integration-tests/bats/schema-changes.bats
VinaiRachakonda: I actually prefer less file clutter in our bats directories. This is technically another type of schema-change.

go/libraries/doltcore/schema/column.go
andrew-wm-arthur: We should adjust the super schema logic instead of changing this. changing this here could have unexpected consequences

go/libraries/doltcore/sqle/dtables/diff_table.go
andrew-wm-arthur: My reading of this is that `isDiffablePartition()` will prevent us from ever creating an `async_differ` for two roots that can't be diffed due to incompatible PK sets. Is this not true? 
 We definitely want to catch these cases earlier, higher up the stack

integration-tests/bats/schema-changes.bats
andrew-wm-arthur: Aesthetically, it's subjective, but we've started parallelizing BATS by passing each file to to its own github action. It makes CI quicker to have smaller files

go/libraries/doltcore/sqle/tables.go
andrew-wm-arthur: I'll sync in person

integration-tests/bats/primary-key-changes.bats
```diff
@@ -0,0 +1,539 @@
+#!/usr/bin/env bats
```
andrew-wm-arthur, line 1: need to add a test here for failing to add a PK because the column has duplicates

andrew-wm-arthur (approved):
LGTM

PR #2013, opened 2021-08-14
Andy/alter drop pks

andrew-wm-arthur:


go/libraries/doltcore/merge/integration_test.go
```diff
@@ -32,6 +32,7 @@ func TestMerge(t *testing.T) {
 
 	setupCommon := []testCommand{
 		{cmd.SqlCmd{}, args{"-q", "CREATE TABLE test (pk int PRIMARY KEY, c0 int);"}},
+		{cmd.AddCmd{}, args{"."}},
```
VinaiRachakonda, line 35: @andrew-wm-arthur This is a bug right? Commit -am is right beneath it. Do we need this Add to make the test pass

andrew-wm-arthur: After a little digging... this is somewhat of a bug with the `-a` flag of `dolt commit`. Currently passing `-a` stages all tables, tracked and untracked, but this does not match Git semantics for `-a`. 

As it's written, the test should fail to commit because table test is untracked and the commit is empty. However, `commit -am` just stages everything, so the tests pass with or without the preceding `dolt add .`.

VinaiRachakonda: Sounds good, are you comfortable merging this into my pr?

andrew-wm-arthur: yep, feel free to merge

PR #2014, opened 2021-08-16
[DRAFT] In-memory BATS Go runner

Hydrocharged:
Heavy WIP, may not even be a worthwhile approach (code is trash, don't look pls).

Right now this changes all file handling to work with the [afero file system](https://github.com/spf13/afero), which basically allows the file system to be swapped out. By default we're using the operating system's file system (which is the same as before), but we can change it to a memory-backed file system for testing. This should (maybe/hopefully) allow for a Go runner to run BATS tests using this in-memory file system only for tests, making it _significantly_ easier (and faster) to run BATS tests while developing.

Also uses https://github.com/dolthub/mmap-go/tree/daylon/mmap-mock

This is just the file system change for now, no Go runner has been built yet.


Hydrocharged:
This approach won't work.

PR #2019, opened 2021-08-16
Skipped bats test for non-deterministic table import behavior 

timsehn:
Skipped bats test for duplicate rows in a table import -u --continue producing non-deterministic imports when repeated


VinaiRachakonda:
This is fixed with #2426

PR #2021, opened 2021-08-16
Changed TEXT types to use BlobKind

Hydrocharged:
Previously we were using `types.String` at the noms layer to handle our `TEXT` types. This works for relatively small strings, but causes issues when using larger strings (such as a 2GB string). I'm not sure of the point at which performance tanks/crashes occur using `types.String`, but `types.Blob` can handle a string of arbitrary length (up to our allowed maximum of 4GB for sure). This also matches how our `BLOB` types are implemented.

We do make use of the `StringDefaultType` variable all over the codebase, which was a `LONGTEXT` field. This has been changed to a `VARCHAR(16383)` field to preserve the expected behavior (passing in a `types.String` to the default string type for example). This does mean that our length limit is no longer 4GB, however I don't _think_ we were ever reaching the `VARCHAR` limit to begin with, so this switch should be fine for all of our pre-existing logic. Additionally, the original string implementation remains (along with a `LegacyStringDefaultType` for system tables), so old repositories will continue to reference the old type as expected. This will only affect new tables. It was bad practice that we always used `LONGTEXT` to begin with.

Our Go and bats tests make extensive use of `LONGTEXT` already (it was our only supported string type for a while), so minimal testing needed to be written.

This PR also modifies `types.Blob` to properly sort, as previously it sorted by hash (which is the default for types that are an extension of `sequence`). Adding a proper `Less` function has broken some expected behaviors in noms for blobs, but they're not being used at all in Dolt, therefore it seems safe? Fixing these behaviors would entail a lower-level change, one which I didn't deem worth.

Lastly, besides the large collection of small bug fixes here and there, I bumped the feature version. Any new tables that contain a `TEXT` type (of which there will probably be numerous) will be unable to be read by the previous version of Dolt (as they're missing the new `blobStringType`), and will throw an `unknown` error.


reltuk (approved):
This generally looks fine to me. I don't love the backwards-incompatible change to `varbinary`, but I'm not sure if it's worth avoiding if there are no existing uses of `varbinary`.

Hydrocharged:
If there are any users that have a private repo using a `BLOB` type then we can always do a `SELECT ... INSERT` with the substring.

PR #2023, opened 2021-08-18
global system var for sql-server session branch

max-hoffman:
setting `@@GLOBAL.dolt_sql_server_branch_ref` overwrites the branch for new sql-server connections. Setting the variable does not change the current connection's branch. If the variable is set to an invalid ref (ex: branch does not exist), the behavior is undefined for new connections.

```
> set GLOBAL dolt_sql_server_branch_ref = 'refs/heads/new'
mysql> select @@GLOBAL.dolt_sql_server_branch_ref;
+-------------------------------------+
| @@GLOBAL.dolt_sql_server_branch_ref |
+-------------------------------------+
| refs/heads/new                      |
+-------------------------------------+
1 row in set (0.00 sec)
mysql> select active_branch();
+-----------------+
| active_branch() |
+-----------------+
| master          |
+-----------------+
mysql> exit
Bye
$ mysql --user root --host=0.0.0.0 -p tmp1

mysql> select active_branch();
+-----------------+
| active_branch() |
+-----------------+
| new             |
+-----------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/server.go
+++ b/go/cmd/dolt/commands/sqlserver/server.go
@@ -16,6 +16,7 @@ package sqlserver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -33,13 +33,30 @@ import (
...
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 	dsqle "github.com/dolthub/dolt/go/libraries/doltcore/sqle"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dfunctions"
 	_ "github.com/dolthub/dolt/go/libraries/doltcore/sqle/dfunctions"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
 	"github.com/dolthub/dolt/go/libraries/utils/tracing"
 )
 
+const SqlServerBranchRefKey = "dolt_sql_server_branch_ref"
```
zachmu, line 44: Would call this dolt_default_branch

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -260,7 +276,18 @@ func getDbStates(ctx context.Context, mrEnv env.MultiRepoEnv, dbs []dsqle.Databa
 			return nil, fmt.Errorf("couldn't find environment for database %s", db.Name())
 		}
 
+		if _, val, ok := sql.SystemVariables.GetGlobal(SqlServerBranchRefKey); ok {
```
zachmu, line 279: This is definitely not what you want.

This is using the repo state writer in doltEnv to write a new branch to the repo state, but without any of the business logic in `dolt checkout`.

Think of the command line env as a separate, persistent session. You're changing the branch of that session as a side effect to this logic.

You need to add the default branch to the session, and add logic to newSessionBuilder to set it there, initialized from the value of the global sys var at connection time

zachmu (commented):
Needs work, see comments

go/cmd/dolt/commands/sqlserver/server.go
max-hoffman: i still don't know how to test this thing for weird cases

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -261,12 +278,47 @@ func getDbStates(ctx context.Context, mrEnv env.MultiRepoEnv, dbs []dsqle.Databa
 		}
 
 		head := dEnv.RepoStateReader().CWBHeadSpec()
-		headCommit, err := dEnv.DoltDB.Resolve(ctx, head, dEnv.RepoStateReader().CWBHeadRef())
+
+		var ws *doltdb.WorkingSet
+		var headCommit *doltdb.Commit
+		var err error
+		if _, val, ok := sql.SystemVariables.GetGlobal(DoltDefaultBranchKey); ok {
+			if refStr, ok := val.(string); ok && ref.IsRef(refStr) {
```
zachmu, line 286: Better to throw an error here if this isn't a valid ref string

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -261,12 +278,47 @@ func getDbStates(ctx context.Context, mrEnv env.MultiRepoEnv, dbs []dsqle.Databa
 		}
 
 		head := dEnv.RepoStateReader().CWBHeadSpec()
```
zachmu, line 280: Would replace this with the definition from repo state:

head, _ := doltdb.NewCommitSpec("HEAD")

Makes it more obvious what's going on here. The CWBHeadSpec() method is pretty dumb, needs to go

go/cmd/dolt/commands/sqlserver/server.go
zachmu: Encapsulate this whole chunk of logic in a method

go/cmd/dolt/commands/sqlserver/server.go
zachmu: Also the conditional above (dEnv == nil) is dead code

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -261,12 +278,47 @@ func getDbStates(ctx context.Context, mrEnv env.MultiRepoEnv, dbs []dsqle.Databa
...
+					return nil, err
+				}
+
+				dbStates = append(dbStates, dsess.InitialDbState{
+					Db:         db,
+					HeadCommit: headCommit,
+					WorkingSet: ws,
+					DbData:     dEnv.DbData(),
+				})
+				continue
```
zachmu, line 312: The continues are getting a little out of hand, making this method hard to read.

You should rewrite this to be simple conditional logic that gets a db state one of 3 ways

```go
if (denv == nil) {
   dbState = getRevisionDbState(...)
} else if  (_, val, ok := sql.SystemVariables.GetGlobal(DoltDefaultBranchKey); ok) {
  dbState = getDbStateForDefaultBranch(...)
} else {
  dbstate = getDbStateForDEnv(...)
} 

dbStates.append...
```

integration-tests/bats/sql-server.bats
```diff
@@ -850,6 +850,51 @@ SQL
...
+    start_sql_server repo1
+
+    multi_query repo1 1 '
+    select dolt_checkout("new");
+    CREATE TABLE t (a int primary key, b int);
+    INSERT INTO t VALUES (2,2),(3,3);' ""
+
+    server_query repo1 1 "SHOW tables" "" # no tables on master
+
+    server_query repo1 1 "set GLOBAL dolt_default_branch = 'refs/heads/new';" ""
```
zachmu, line 872: You should make this also work with just a branch name, not a full ref. Add a test for that

integration-tests/bats/sql-server.bats
```diff
@@ -850,6 +850,51 @@ SQL
...
+    multi_query repo1 1 '
+    select dolt_checkout("new");
+    CREATE TABLE t (a int primary key, b int);
+    INSERT INTO t VALUES (2,2),(3,3);' ""
+
+    server_query repo1 1 "SHOW tables" "" # no tables on master
+
+    multi_query repo1 1 '
+    set GLOBAL dolt_default_branch = "refs/heads/new";
+    show tables;' "" # SET GLOBAL does not affect current connection
```
zachmu, line 895: I could be wrong, but I think that multi_query doesn't actually do any checks with the result string given. Try it and see

zachmu (commented):
Looks pretty good, just a few small changes needed

integration-tests/bats/sql-server.bats
max-hoffman: yeah multi_query doesn't pass that arg

max-hoffman: i'll just delete, idk how to test within current session

zachmu (approved):
LGTM

PR #2032, opened 2021-08-18
Method to add a commit and update working set together atomically

zachmu:


go/store/datas/database_common.go
```diff
@@ -547,6 +557,16 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
+	if err != nil {
+		return err
+	}
+
+	wsValRef, err := types.ToRefOfValue(workingSetRef, db.Format())
+	if err != nil {
+		return err
+	}
+
 	// This could loop forever, given enough simultaneous writers. BUG 2565
```
VinaiRachakonda, line 570: Which issue is 2565? Not seeing that as a valid issue number

go/libraries/doltcore/env/actions/commit.go
```diff
@@ -157,6 +157,81 @@ func CommitStaged(ctx context.Context, roots doltdb.Roots, mergeActive bool, mer
 	return c, nil
 }
 
+// CommitStaged adds a new commit to HEAD with the given props. Returns the new commit's hash as a string and an error.
```
VinaiRachakonda, line 160: Should be GetCommitStaged not CommitStaged

go/store/datas/database_common.go
zachmu: It was an issue in the noms repo.

I'm just deleting the comment

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1042,6 +1040,58 @@ func (ddb *DoltDB) UpdateWorkingSet(
...
+	}
+
+	// logrus.Tracef("Updating working set with root %s", workingSet.RootValue().DebugString(ctx, true))
+
+	workingRootRef, stagedRef, mergeStateRef, err := workingSet.writeValues(ctx, ddb)
+	if err != nil {
+		return err
+	}
+
+	// While we still have places that need user info threaded through, we're lenient on providing the meta
```
reltuk, line 1071: Should we keep a counter of these things and log it or something? Or should we create a special value at the call sites similar to `context.TODO()` so that we can grep for it easily?

go/store/types/value_store.go
```diff
@@ -460,101 +460,99 @@ func (lvs *ValueStore) Rebase(ctx context.Context) error {
...
 // rebased. Until Commit() succeeds, no work of the ValueStore will be visible
 // to other readers of the underlying ChunkStore.
 func (lvs *ValueStore) Commit(ctx context.Context, current, last hash.Hash) (bool, error) {
-	return func() (bool, error) {
-		lvs.bufferMu.Lock()
-		defer lvs.bufferMu.Unlock()
-
-		put := func(h hash.Hash, chunk chunks.Chunk) error {
-			err := lvs.cs.Put(ctx, chunk)
+	lvs.bufferMu.Lock()
```
zachmu, line 470: This diff looks complicated, but all I did was remove the anonymous func call

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1042,6 +1040,58 @@ func (ddb *DoltDB) UpdateWorkingSet(
...
+	var metaSt types.Struct
+	if meta != nil {
+		metaSt, err = meta.toNomsStruct(types.Format_Default)
```
reltuk, line 1074: Unless I'm mistaken, this is better as the DoltDB NBF. (`ddb.db.Format()` probably)

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1042,6 +1040,58 @@ func (ddb *DoltDB) UpdateWorkingSet(
...
+		if err != nil {
+			return err
+		}
+	} else {
+		metaSt, err = datas.NewWorkingSetMeta(types.Format_Default, "incomplete", "incomplete", uint64(time.Now().Unix()), "incomplete")
```
reltuk, line 1079: Is this better as the `CommitNowFunc()`? Any need to mock this or for it to track test time sometimes?

go/store/datas/database_common.go
```diff
@@ -547,67 +555,192 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
+			return db.tryCommitChunks(ctx, currentDatasets, currentRootHash)
+		}()
+	}
+
+	return tryCommitErr
+}
+
+// assertDatasetHash returns true if the hash of the dataset matches the one given. Use an empty hash for a dataset you
+// expect not to exist.
+// Typically this is called using pessimistic locking by the caller in order to implement atomic test-and-set semantics.
```
reltuk, line 608:
```suggestion
// Typically this is called using optimistic locking by the caller in order to implement atomic test-and-set semantics.
```

go/store/datas/database_common.go
```diff
@@ -547,67 +555,192 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
+		if err != nil {
+			return false, err
+		}
+
+		if h != currHash {
+			return false, err
+		}
+	} else {
+		if !currHash.IsEmpty() {
+			panic("No ref found for workspace " + datasetID)
```
reltuk, line 636: Is this right? Isn't this just, caller expected dataset to exist but it does not? Caller failed the optimistic lock check?

Another way to think of the question: Is there a way to remove a dataset that's going to be asserted on without risking a panic?

go/store/types/value_store.go
```diff
@@ -460,101 +460,99 @@ func (lvs *ValueStore) Rebase(ctx context.Context) error {
...
+		root, err := lvs.Root(ctx)
 
 		if err != nil {
 			return false, err
 		}
 
-		if !success {
-			return false, nil
+		if (current != hash.Hash{} && current != root) {
+			if _, ok := lvs.bufferedChunks[current]; !ok {
```
reltuk, line 530: What's happening here? Isn't `lvs.bufferedChunks` necessarily empty here? Where we supposed to check this somewhere else / further up in the function?

reltuk (commented):
A few comments!

go/store/types/value_store.go
zachmu: I didn't change this, just removed the anonymous func() invocation it was wrapped in for no reason.

Very possible this has bugs, but they were pre-existing conditions.

go/store/datas/database_common.go
zachmu: Going to make some changes as we discussed

zachmu (commented):
Going to get another draft out with some changes

go/store/datas/database_common.go
```diff
@@ -547,67 +563,186 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
+		if err != nil {
+			return false, err
+		}
+
+		if h != currHash {
+			return false, err
+		}
+	} else if !currHash.IsEmpty() {
+		return false, nil
+	}
```
reltuk, line 643: This seems a tiny bit cleaner (avoids the panic, for example?) as:

```go
if head, ok := ds.MaybeHead(); ok {
    h, err := head.Hash(db.Format())
    if err != nil {
        return false, err
    }
    if h != currHash {
        return false, err
    }
} else if !currHash.IsEmpty() {
    return false, nil
}
```

go/store/datas/database_common.go
```diff
@@ -547,67 +563,186 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
+			if hasHead {
+				head, err := r.(types.Ref).TargetValue(ctx, db)
 				if err != nil {
 					return err
 				}
 
-				if h != currHash {
-					testSetFailed = true
-					return ErrOptimisticLockFailed
+				currentHeadRef, err := types.NewRef(head, db.Format())
```
reltuk, line 727: This is weird to me...after

```go
v, err := ref.TargetValue(ctx, db)
if err != nil {
    return err
}
r, err := types.NewRef(v, db.Format())
if err != nil {
    return err
}
```

doesn't `r == ref`. Why did I do TargetValue and then re-ref the value?

go/store/datas/database_common.go
```diff
@@ -616,14 +751,27 @@ func (db *database) doUpdateWorkingSet(ctx context.Context, datasetID string, wo
...
-	return tryCommitErr
+	commitDS, err = db.GetDataset(ctx, commitDS.ID())
+	if err != nil {
+		return Dataset{}, Dataset{}, err
+	}
+
+	workingSetDS, err = db.GetDataset(ctx, workingSetDS.ID())
+	if err != nil {
+		return Dataset{}, Dataset{}, err
+	}
```
reltuk, line 762: Is this potentially subtly racy in the same way that the commit logic used to be? Could it theoretically get a smeared view of HEAD vs. workingset and should it work with the same root map across both calls instead?

reltuk (approved):
Everything looks great. A few questions / suggestions, but nothing big at all. Thanks for taking this on 👍

go/store/datas/database_common.go
zachmu: Done

go/store/datas/database_common.go
zachmu: Good call, fixed.

go/store/datas/database_common.go
zachmu: Sometimes you just gotta round trip to the DB for literally no reason

PR #2035, opened 2021-08-19
Added skipped bats test for some dolt conflicts and associated warnings I was seeing

timsehn:
This test tests the conflict resolution workflow where you get conflicts, resolve using --ours or --theirs but still want to modify the working set. It also tests the output of `dolt conflicts cat` when there are no conflicts.


zachmu (approved):
LGTM

PR #2040, opened 2021-08-20
drop primary key does not fail of foreign key constraint if fallback index exists

max-hoffman:
this is what I need for MLflow, which errored in the original implementation:
```bash
    dolt sql -q "CREATE table parent (pk int, val int, primary key (pk, val), key `backup` (val))"
    dolt sql -q "CREATE table child (pk int, val int, foreign key (val) references parent (val))"
    dolt sql -q "alter table parent drop primary key"
```

### DIFF
--- a/go/libraries/doltcore/schema/alterschema/droppk.go
+++ b/go/libraries/doltcore/schema/alterschema/droppk.go
@@ -48,6 +48,9 @@ func DropPrimaryKeyFromTable(ctx context.Context, table *doltdb.Table, nbf *type
 	}
 
 	newSchema.Indexes().AddIndex(sch.Indexes().AllIndexes()...)
+	if err != nil {
+		return nil, err
+	}
 
 	table, err = table.UpdateSchema(ctx, newSchema)
 	if err != nil {
--- a/go/libraries/doltcore/schema/alterschema/droppk_test.go
+++ b/go/libraries/doltcore/schema/alterschema/droppk_test.go
@@ -124,3 +124,152 @@ func TestDropPk(t *testing.T) {
 		assert.True(t, ok)
 	})
 }
+
+func TestDropPks(t *testing.T) {
+	var dropTests = []struct {
+		name      string
+		setup     []string
+		check     []string
+		want      []row.Row
+		exit      int
+		fkIdxName string
+	}{
+		{
+			name: "no error on drop pk",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id))",
+				"insert into parent values (1,1,1),(2,2,2)",
+			},
+			exit: 0,
+		},
+		{
+			name: "no error if backup index",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id), key `backup` (id))",
+				"create table child (id int, name varchar(1), age int, primary key (id), constraint `fk` foreign key (id) references parent (id))",
+				"insert into parent values (1,1,1),(2,2,2)",
+				"insert into child values (1,1,1),(2,2,2)",
+			},
+			exit:      0,
+			fkIdxName: "backup",
+		},
+		{
+			name: "no error if backup index for single FK on compound pk drop",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `backup` (age))",
+				"create table child (id int, name varchar(1), age int, primary key (id), constraint `fk` foreign key (age) references parent (age))",
+				"insert into parent values (1,1,1),(2,2,2)",
+				"insert into child values (1,1,1),(2,2,2)",
+			},
+			exit:      0,
+			fkIdxName: "backup",
+		},
+		{
+			name: "no error if compound backup index for compound FK",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `backup` (id, age))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (id, age) references parent (id, age))",
+				"insert into parent values (1,1,1),(2,2,2)",
+				"insert into child values (1,1,1),(2,2,2)",
+			},
+			exit:      0,
+			fkIdxName: "backup",
+		},
+		{
+			name: "no error if both invalid and valid backup index",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `bad_backup1` (id, age), key `bad_backup2` (id), key `backup` (age, id))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (age) references parent (age))",
+				"insert into parent values (1,1,1),(2,2,2)",
+				"insert into child values (1,1,1),(2,2,2)",
+			},
+			exit:      0,
+			fkIdxName: "backup",
+		},
+		{
+			name: "error if FK ref but no backup index for single pk",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id))",
+				"create table child (id int, name varchar(1), age int, primary key (id), constraint `fk` foreign key (id) references parent (id))",
+				"insert into parent values (1,1,1),(2,2,2)",
+				"insert into child values (1,1,1),(2,2,2)",
+			},
+			exit:      1,
+			fkIdxName: "id",
+		},
+		{
+			name: "error if FK ref but bad backup index",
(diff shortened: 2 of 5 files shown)


go/libraries/doltcore/schema/alterschema/droppk.go
```diff
@@ -48,6 +48,9 @@ func DropPrimaryKeyFromTable(ctx context.Context, table *doltdb.Table, nbf *type
 	}
 
 	newSchema.Indexes().AddIndex(sch.Indexes().AllIndexes()...)
+	if err != nil {
```
zachmu, line 51: This block is a no-op, `err` has not changed since the last return statement

go/libraries/doltcore/schema/alterschema/droppk_test.go
```diff
@@ -124,3 +124,134 @@ func TestDropPk(t *testing.T) {
...
+	for _, tt := range dropTests {
+		childName := "child"
+		parentName := "parent"
+		childFkName := "fk"
+
+		t.Run(tt.name, func(t *testing.T) {
+			dEnv := dtestutils.CreateTestEnv()
+			ctx := context.Background()
+			for _, c := range tt.setup {
+				exitCode := commands.SqlCmd{}.Exec(ctx, "sql", []string{fmt.Sprintf("-q %s", c)}, dEnv)
```
zachmu, line 226: This is a really cumbersome way to execute SQL statements. Why don't you use the SQL test utility functions here?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1803,3 +1794,69 @@ func (t *AlterableDoltTable) DropPrimaryKey(ctx *sql.Context) error {
 
 	return t.updateFromRoot(ctx, newRoot)
 }
+
+// canDropPk checks whether there are conflicting foreign key relationships referencing this
+// table's primary key tags
+func (t *AlterableDoltTable) verifyCanDropPk(ctx *sql.Context, root *doltdb.RootValue) (*doltdb.ForeignKeyCollection, error) {
```
zachmu, line 1800: Don't mix business logic and basic data access functionality like this. The fact that this method is called "verifyX" but returns a FKC (and in fact replaces a method called GetForeignKeyCollection in the usage above) is a huge red flag. This is a bad factoring, find a way to cleanly separate these two concerns. May be as simple as only returning a bool or err here instead of a FKC, but I haven't read very closely

zachmu (commented):
Daylon should give final OK, just a couple drive by comments from me

go/libraries/doltcore/schema/alterschema/droppk_test.go
max-hoffman: i agree i hate this, which utility functions? when i asked last week you pointed me towards enginetests

go/libraries/doltcore/sqle/tables.go
max-hoffman: yeah makes sense. the data structures i have to build for this seem expensive, so not sure we want to do it twice. `updateFkForNewSchema`, `updateFkForPkDrop`, `fkBackupsForPkDrop`, or something. happy to split search and modify into two steps if need be, seems equivalent on my end

go/libraries/doltcore/schema/alterschema/droppk_test.go
```diff
@@ -124,3 +124,134 @@ func TestDropPk(t *testing.T) {
...
+			fkIdxName: "backup",
+		},
+		{
+			name: "no error if both invalid and valid backup index",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `bad_backup1` (id, age), key `bad_backup2` (id), key `backup` (age, id))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (age) references parent (age))",
+			},
+			exit:      0,
+			fkIdxName: "backup",
```
Hydrocharged, line 177: This is a good test, but I'd also add a case where there's just the single compound index and the foreign key only references the first key.

go/libraries/doltcore/schema/alterschema/droppk_test.go
```diff
@@ -124,3 +124,134 @@ func TestDropPk(t *testing.T) {
...
+		},
+		{
+			name: "error if incomplete compound backup index for FK",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, constraint `primary` primary key (age, id), key `backup` (age, name))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (age, id) references parent (age,  id))",
+			},
+			exit:      1,
+			fkIdxName: "ageid",
+		},
```
Hydrocharged, line 214: Add one more test for a 3-key composite index with a foreign key over the first 2 keys.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1803,3 +1794,69 @@ func (t *AlterableDoltTable) DropPrimaryKey(ctx *sql.Context) error {
...
+		return nil, err
+	}
+
+	// pkBackups is a mapping from the table's PL tags to potentially compensating indexes
+	pkBackups := make(map[uint64][]schema.Index, len(t.sch.GetPKCols().TagToIdx))
+	for tag, _ := range t.sch.GetPKCols().TagToIdx {
+		pkBackups[tag] = nil
+	}
+	for _, idx := range indexes {
+		if !idx.IsUserDefined() || !!idx.IsUnique() {
```
Hydrocharged, line 1817: `!!`?

go/libraries/doltcore/schema/alterschema/droppk_test.go
```diff
@@ -124,3 +124,134 @@ func TestDropPk(t *testing.T) {
...
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `backup` (id, age))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (id, age) references parent (id, age))",
+			},
+			exit:      0,
+			fkIdxName: "backup",
+		},
+		{
+			name: "no error if both invalid and valid backup index",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, primary key (id, age), key `bad_backup1` (id, age), key `bad_backup2` (id), key `backup` (age, id))",
```
Hydrocharged, line 173: We need tests of multiple valid keys as well.

go/libraries/doltcore/schema/alterschema/droppk_test.go
```diff
@@ -124,3 +124,134 @@ func TestDropPk(t *testing.T) {
...
+		{
+			name: "error if incomplete compound backup index for FK",
+			setup: []string{
+				"create table parent (id int, name varchar(1), age int, constraint `primary` primary key (age, id), key `backup` (age, name))",
+				"create table child (id int, name varchar(1), age int, constraint `fk` foreign key (age, id) references parent (age,  id))",
+			},
+			exit:      1,
+			fkIdxName: "ageid",
+		},
+	}
```
Hydrocharged, line 215: Is there any priority or a specific order that governs which index is chosen? Now we don't allow for indexes to cover the same exact set of keys, but you can declare composite indexes that are partial keys of other indexes. Is the smallest index always chosen? The index that was declared first? A unique index takes priority over all others if one is available and the data doesn't have any duplicate rows? Some combination of the aforementioned?

We don't have to implement all of these (and can't for some like index declaration order since we don't preserve that information), but it's good to know, and perhaps even drop a comment in the source. But for something like unique key preference, I think we should definitely implement that.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1803,3 +1794,69 @@ func (t *AlterableDoltTable) DropPrimaryKey(ctx *sql.Context) error {
...
+			continue
+		}
```
Hydrocharged, line 1819: I'm assuming that we're choosing to not use `UNIQUE` keys as we'd have to check the data, but that's fine. We can short-circuit checks when the key length and fk length are the same as a primary key is a unique key, so we'd only have to check when the primary key is longer than the foreign key.

If unique keys aren't supported then I'd add a comment here stating such. I can clearly see why we only use user-defined indexes, but no idea on why uniques are thrown out.

Hydrocharged (commented):
Looks good, just a few comments. Once you address my feedback and Zach's feedback I'll take another look.

PR #2041, opened 2021-08-24
Added a global commit lock

zachmu:


go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -112,62 +117,117 @@ func (tx *DoltTransaction) Commit(ctx *sql.Context, workingSet *doltdb.WorkingSe
...
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			mergedWorkingSet := workingSet.WithWorkingRoot(mergedRoot)
+			err = tx.dbData.Ddb.UpdateWorkingSet(ctx, tx.workingSetRef, mergedWorkingSet, hash, tx.getWorkingSetMeta(ctx))
 			if err == datas.ErrOptimisticLockFailed {
-				continue
+				return nil, nil
```
andrew-wm-arthur, line 186: same question here

go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -112,62 +117,117 @@ func (tx *DoltTransaction) Commit(ctx *sql.Context, workingSet *doltdb.WorkingSe
...
-				// TODO: surface duplicate key errors as appropriate
-				return nil, fmt.Errorf("conflict in table %s", table)
-			}
+	// TODO: different error type for retries exhausted
+	return nil, datas.ErrOptimisticLockFailed
+}
+
+// stompConflicts resolves the conflicted tables in the root given by blindly accepting theirs, and returns the
+// updated root value
+func (tx *DoltTransaction) stompConflicts(ctx *sql.Context, mergedRoot *doltdb.RootValue, tablesWithConflicts []string) (*doltdb.RootValue, error) {
```
andrew-wm-arthur, line 207: is stomping conflicts the correct behavior for `REPEATABLE READ`? last write wins?

go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -112,62 +117,117 @@ func (tx *DoltTransaction) Commit(ctx *sql.Context, workingSet *doltdb.WorkingSe
...
+			hash, err := ws.HashOf()
+			if err != nil {
+				return nil, err
+			}
+
+			if newWorkingSet || rootsEqual(existingWorkingRoot, tx.startState.WorkingRoot()) {
+				// ff merge
+				err = tx.dbData.Ddb.UpdateWorkingSet(ctx, tx.workingSetRef, workingSet, hash, tx.getWorkingSetMeta(ctx))
+				if err == datas.ErrOptimisticLockFailed {
+					return nil, nil
```
andrew-wm-arthur, line 148: why do we throw away this error? we should avoid this error entirely with serialized commits

zachmu: This is essentially a continue, I'll add a comment

go/libraries/doltcore/sqle/dsess/transactions.go
zachmu: It's defensible, but for it to be correct you need to support select for update with row-level locks.

Really the only reason to turn this on is for benchmarks, an application would want to catch this error and retry the transaction

andrew-wm-arthur (approved):
LGTM

PR #2047, opened 2021-08-25
Bh/disk edits refactor

bheni:


go/store/types/edits/edit_prodiver_merger_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2021 Dolthub, Inc.
```
bheni, line 1: this file was moved.  It was previously tuple_stream_merger.go before the refactor.

go/store/types/edits/edit_prodiver_merger.go
```diff
@@ -0,0 +1,228 @@
+// Copyright 2021 Dolthub, Inc.
```
bheni, line 1: this file was moved. It was previously tuple_stream_merger.go before the refactor.

PR #2050, opened 2021-08-25
Bh/tea refactor

bheni:


go/libraries/doltcore/table/editor/table_edit_accumulator.go
```diff
@@ -0,0 +1,341 @@
...
+		ea.AddEdit(key, nil)
+	}
+
+	tea.uncommittedEaIds.Add(tea.accumulatorIdx)
+	tea.flusher.Flush(ea, tea.accumulatorIdx)
+	tea.accumulatorIdx++
+
+	tea.uncommittedFlushed.MergeIn(tea.uncommitted)
+	tea.uncommitted = NewInMemModifications()
+}
```
Hydrocharged, line 176: It looks like we're now handling the auto-flushing behavior on the same thread that writes are being done, where as previously it was asynchronous. I decided to run `master` against this branch on a 1-million-row SQL file and:

```
+---------+-----------------+
| master  | bh/tea-refactor |
+---------+-----------------+
| 27.843s | 34.109s         |
+---------+-----------------+
```

Takes 22% longer so it's not a massive difference, but it's primarily masked by how slow our SQL analyzer is, so I'm sure that as we optimize the analyzer, this gap will become larger and larger. I remember way back when making the first iteration of the `TableEditAccumulator`, the SQL path showed a 5% increase where as the import path was like 70-80% because there was just so much analyzer overhead in the SQL path.

Do you have any near-at-hand ideas on how to make this asynchronous? If not it's fine for now, but we may be optimizing this later if not.

Hydrocharged (approved):
LGTM! Just some nits and that `TEA` async question.

go/libraries/doltcore/table/editor/table_edit_accumulator.go
bheni: so the flushes happen asynchronously in the `tea.flusher.Flush(ea, tea.accumulatorIdx)` call.  I played with your test for a bit.  We are 20% slower in your case.  This only happens when the inserts are already ordered, the number of inserts isn't massive, and the number of columns isn't huge.  If any of these isn't true the perf is better and there is less garbage to collect. Been playing with other things to make it even better but haven't found a way to improve this special case.

PR #2064, opened 2021-08-29
create empty tuples at start

bheni:
creates empty tuples per format in advance so we don't create all this garbage

Screen Shot 2021-08-29 at 9 05 53 AM


reltuk (approved):
LGTM

Issue #2048, opened 2021-08-25
Dolt fetch on unknown remote formatted with `remotes/` panics


VinaiRachakonda:
Repro

1. Clone any repository
2. Run `dolt fetch remotes/asdasdasd`

Panic:

```
panic: runtime error: slice bounds out of range [6:5]

goroutine 1 [running]:
github.com/dolthub/dolt/go/libraries/doltcore/ref.NewRemoteRefFromPathStr(0x7ffeefbff973, 0x5, 0x8, 0x1, 0x0, 0xc000601200)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/libraries/doltcore/ref/remote_ref.go:80 +0x368
github.com/dolthub/dolt/go/libraries/doltcore/ref.Parse(0x7ffeefbff96b, 0xd, 0x286cd60, 0x1, 0x0, 0x0)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/libraries/doltcore/ref/ref.go:137 +0xa5b
github.com/dolthub/dolt/go/libraries/doltcore/ref.ParseRefSpecForRemote(0x0, 0x0, 0x7ffeefbff96b, 0xd, 0x50007d0, 0x2c442ca8, 0x28, 0x28)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/libraries/doltcore/ref/ref_spec.go:89 +0x6c9
github.com/dolthub/dolt/go/libraries/doltcore/ref.ParseRefSpec(...)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/libraries/doltcore/ref/ref_spec.go:55
github.com/dolthub/dolt/go/cmd/dolt/commands.parseRSFromArgs(0x2871bc6, 0x6, 0xc000188600, 0x1, 0x10, 0x286e500, 0x4, 0x3726250, 0x1, 0x1)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/cmd/dolt/commands/fetch.go:138 +0xbc
github.com/dolthub/dolt/go/cmd/dolt/commands.getRefSpecs(0xc000188600, 0x1, 0x10, 0xc000328370, 0xc0004406f0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/vinairachakonda/go/src/dolthub/dolt/go/cmd/dolt/commands/fetch.go:122 +0x31c
[...]

PR #2067, opened 2021-08-30
 Addresses different types of panics with remotes/fetches Issue: #2048

VinaiRachakonda:
This pr addresses three issues with fetching remotes. 

1. A panic with fetching a remote in the form "remotes/.." that is invalid 
2. A panic when an added remote that is then fetched is invalid
3. A lack of error when fetching an invalid remote


timsehn (commented):
BATS looks good

andrew-wm-arthur (approved):

PR #2074, opened 2021-08-30
Auto Increment Values should reset for a table after its been dropped

VinaiRachakonda:
This table fixes a bug where a table is dropped and then readded but its next auto increment value didn't reset.


VinaiRachakonda:
#benchmark

VinaiRachakonda:
#benchmark

go/libraries/doltcore/sqle/database.go
```diff
@@ -562,9 +562,29 @@ func (db Database) DropTable(ctx *sql.Context, tableName string) error {
...
 
+	err = db.dropTableFromAiTracker(ctx, tableName)
+	if err != nil {
+		return err
+	}
+
 	return db.SetRoot(ctx, newRoot)
 }
 
+// dropTableFromAiTracker grabs the auto increment tracker and removes the table from it.
```
VinaiRachakonda, line 573: I should change this to "removes the table name tableName from it"

andrew-wm-arthur (approved):

PR #2075, opened 2021-08-31
go/cmd/dolt: Handle SIGINT, SIGTERM with a context cancel.

reltuk:
Also install a handler for sql-server so that the server gets stopped cleanly
when we receive SIGINT or SIGTERM.


bheni (approved):
LGTM but probably breaks color reset.

PR #2078, opened 2021-08-31
Better case handling for system tables

Hydrocharged:


go/libraries/doltcore/sqle/dtables/conflicts_tables.go
```diff
@@ -43,22 +43,22 @@ type RootSetter interface {
 
 // NewConflictsTable returns a new ConflictsTableTable instance
 func NewConflictsTable(ctx *sql.Context, tblName string, root *doltdb.RootValue, rs RootSetter) (sql.Table, error) {
-	tbl, ok, err := root.GetTable(ctx, tblName)
-
+	var tbl *doltdb.Table
+	var ok bool
+	var err error
+	tbl, tblName, ok, err = root.GetTableInsensitive(ctx, tblName)
```
zachmu, line 49: Seems silly to declare all these vars just to reuse one variable here

zachmu: Same comment in all these files

zachmu (approved):
Needs a bats test or two, otherwise LGTM

go/libraries/doltcore/sqle/dtables/conflicts_tables.go
Hydrocharged: This `:=` syntax is not my favorite

PR #2083, opened 2021-09-01
Remove doltpy from CI

max-hoffman:
This should shave 1-6 minutes off of every bats job, including the ~90 windows ones which were 8-9 minutes.


coffeegoddd (approved):
LGTM!

Issue #2084, opened 2021-09-01
Data doesn't transfer to Dolt from R using the RMariaDB client


noamross:
When connecting to a Dolt database from R Using the RMariaDB package, I am unable to write data.  Sending SQL text works fine, so the commands generated to create a table and set up column types and times works.  However, transferring data doesn't work.  In this case, `dbAppendTable()` creates a parameterized SQL statement, sends it, then attaches data to it with the `dbBind()` method.  But no rows are attached to the table.  

I believe on the client end, binding is handled here: https://github.com/r-dbi/RMariaDB/blob/main/src/MariaBinding.cpp

Tagging @krlmlr, RMariaDB's maintainer.  

Here's a reproducible example:

``` r
library(sys)
library(RMariaDB)
library(DBI)
library(withr)

# Clean up from previous session
unlink("doltdb", recursive = TRUE)
try(dbDisconnect(conn), silent = TRUE)


# Running latest Dolt Compiled from source, create dolt db and start server
dir.create("doltdb", showWarnings = FALSE)
with_dir("doltdb", exec_wait("~/go/bin/dolt", "init"))
#> Successfully initialized dolt data repository.
#> [1] 0
dolt_server_pid <- with_dir("doltdb", sys::exec_background(
  "~/go/bin/dolt", c("sql-server", "--port 3333", "--host 127.0.0.1", "--user user", "--password pwd")
))
Sys.sleep(1) # Wait for startup to connect

conn <- dbConnect(RMariaDB::MariaDB(), host = "127.0.0.1", port = 3333,
                  username = "user", password = "pwd", dbname = "doltdb")

# Create the table from the data frame but do not add data, this works fine
dbCreateTable(conn, "mtcars", mtcars)
dbReadTable(conn, "mtcars")
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
#> <0 rows> (or 0-length row.names)

# Adding Data to the table doesn't work
dbAppendTable(conn, "mtcars", mtcars)
#> [1] 0
dbReadTable(conn, "mtcars")
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
#> <0 rows> (or 0-length row.names)

# Try to add data using low-level commands
[...]

PR #2086, opened 2021-09-01
go/go.mod,integration-tests/mysql-client-tests: Bump go-mysql-server to pick up fix for prepared statement writes resultset metadata.

reltuk:
Fixes: #2084.


zachmu (approved):
LG

PR #2093, opened 2021-09-02
RMariaDB mysql-integration-test

max-hoffman:
executes the standard query/response set, and then tests prepared statements separately.

### DIFF
--- a/integration-tests/mysql-client-tests/r/rmariadb-test.r
+++ b/integration-tests/mysql-client-tests/r/rmariadb-test.r
@@ -26,7 +26,7 @@ for(i in 1:length(queries)) {
     want = responses[[i]]
     if (!is.null(want)) {
         got <- dbGetQuery(conn, q)
-        if (!isTRUE(all.equal(rows, exp))) {
+        if (!all(want == got)) {
             print(q)
             print(want)
             print(got)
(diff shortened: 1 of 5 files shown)


timsehn:
Once you get this merged, don't forget to update the docs.

https://docs.dolthub.com/interfaces/sql/clients

integration-tests/mysql-client-tests/r/rmariadb-test.r
```diff
@@ -0,0 +1,65 @@
...
+            quit(1)
+        }
+    } else {
+        dbExecute(conn, q)
+    }
+}
+
+
+# check prepared statements
+statement <- enc2utf8(new("SQL", .Data = "INSERT INTO test (pk, value) values (?, ?)"))
```
reltuk, line 42: I think we should move to:

```r
stmt <- dbSendStatement(conn, 'INSERT INTO `test` VALUES (?, ?)')
rs <- dbBind(stmt, list(-16, '2020-09-10'))
rowsAff <- dbGetRowsAffected(rs)
dbClearResult(rs)
```

and do the same thing in the MySQL test as well.

integration-tests/mysql-client-tests/r/rmariadb-test.r
```diff
@@ -0,0 +1,65 @@
...
+                 username = user, dbname = db)
+
+# check standard queries
+queries = list("create table test (pk int, value int, primary key(pk))",
+               "describe test",
+               "insert into test (pk, value) values (0,0)",
+               "select * from test")
+
+responses = list(NULL,
+                 data.frame(Field = c("pk", "value"), Type = c("int", "int"), Null = c("NO", "YES"), Key = c("PRI", ""), Default = c("", ""), Extra = c("", "")),
```
reltuk, line 20: As discussed offline, I guess we need `stringsAsFactors = FALSE` as parameters to these data.frame calls if we're going to run on older R versions and we want the data frames coming back from the database layer to look as similar as possible to the data frames we are creating here.

reltuk (approved):
Looks great. Thanks for taking this on. Just a couple suggestions for improvements :).

PR #2098, opened 2021-09-07
/go/store/types: Don't compare `v` against itself

andrew-wm-arthur:


Hydrocharged (approved):
LGTM

Hydrocharged:
Variable shadowing should either be disabled or explicit using some special syntax.

PR #2102, opened 2021-09-08
sql server dolt_pull function first pass

max-hoffman:
- no concurrency management, missing mutexes `FetchRemoteBranch` and `FetchTag`. Multiple server connections calling `dolt_pull` will overwrite each others tempfiles, but not disrupt the workingset
- a lot of the more involved `dolt_pull` bats tests are tightly coupled to `push` and `fetch`. I'll clean those up when I add `dolt_push` and `dolt_fetch`
- afaik there are no special errors for `dolt_pull`, underlying fetch and merge steps are responsible for themselves


go/cmd/dolt/commands/merge.go
```diff
@@ -158,7 +158,17 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+			if err != nil {
+				return handleCommitErr(ctx, dEnv, err, usage)
+			}
+
+			name, email, err := env.GetNameAndEmail(dEnv.Config)
+			if err != nil {
+				return handleCommitErr(ctx, dEnv, err, usage)
+			}
+
+			spec, ok, err := merge.ParseMergeSpec(ctx, dEnv.RepoStateReader(), dEnv.DoltDB, roots, name, email, msg, commitSpecStr, apr.Contains(cli.SquashParam), apr.Contains(cli.NoFFParam), apr.Contains(cli.ForceFlag), t)
```
zachmu, line 171: This is a bad name for this method, definitely no parsing going on here

NewMergeSpec is better

go/cmd/dolt/commands/pull.go
```diff
@@ -87,7 +87,7 @@ func (cmd PullCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		remoteName = apr.Arg(0)
 	}
 
-	pullSpec, err := env.ParsePullSpec(ctx, dEnv, remoteName, apr.Contains(cli.SquashParam), apr.Contains(cli.NoFFParam), apr.Contains(cli.ForceFlag))
+	pullSpec, err := env.ParsePullSpec(ctx, dEnv.RepoStateReader(), remoteName, apr.Contains(cli.SquashParam), apr.Contains(cli.NoFFParam), apr.Contains(cli.ForceFlag))
```
zachmu, line 90: Same comment here:

NewPullSpec

go/libraries/doltcore/env/actions/branch.go
```diff
@@ -160,7 +160,7 @@ func DeleteBranchOnDB(ctx context.Context, ddb *doltdb.DoltDB, dref ref.DoltRef,
 		}
 
 		isMerged, _ := master.CanFastReverseTo(ctx, cm)
-		if err != nil && err != doltdb.ErrUpToDate {
+		if err != nil && errors.Is(err, doltdb.ErrUpToDate) {
```
zachmu, line 163: You inverted the logic here, purposeful?

go/libraries/doltcore/env/remotes.go
```diff
@@ -275,10 +275,10 @@ type PullSpec struct {
 	Branch     ref.DoltRef
 }
 
-func ParsePullSpec(ctx context.Context, dEnv *DoltEnv, remoteName string, squash, noff, force bool) (*PullSpec, error) {
-	branch := dEnv.RepoStateReader().CWBHeadRef()
+func ParsePullSpec(ctx context.Context, rsr RepoStateReader, remoteName string, squash, noff, force bool) (*PullSpec, error) {
```
zachmu, line 278: Same comment

NewPullSpec

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -102,108 +94,110 @@ func (d DoltMergeFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error)
...
+
+	err = sess.SetWorkingSet(ctx, dbName, ws, nil)
+	if err != nil {
+		return conflicts, err
 	}
 
+	return conflicts, nil
+}
+
+func mergeHelper(ctx *sql.Context, sess *dsess.Session, roots doltdb.Roots, ws *doltdb.WorkingSet, dbName string, spec *merge.MergeSpec) (*doltdb.WorkingSet, interface{}, error) {
```
zachmu, line 116: mergeIntoWorkingSet

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -239,112 +233,113 @@ func executeMerge(ctx *sql.Context, squash bool, head, cm *doltdb.Commit, ws *do
...
+	})
+	if err != nil {
+		return nil, err
 	}
 
-	// TODO: refactor, redundant
+	return ws, dSess.SetWorkingSet(ctx, dbName, ws.ClearMerge(), nil)
+}
+
+func sqlMergeSpec(ctx *sql.Context, sess *dsess.Session, dbName string, apr *argparser.ArgParseResults, commitSpecStr string) (*merge.MergeSpec, error) {
```
zachmu, line 306: createMergeSpec

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
zachmu: I know you didn't create these methods originally, but returning `interface{}` for the two constants `hasConflicts` and `noConflicts` is not great. Define an integer type alias for them and declare the consts in terms of that type.

zachmu: Also add a method doc here, pay attention to the return values

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if len(dbName) == 0 {
+		return noConflicts, fmt.Errorf("empty database name.")
+	}
+
+	sess := dsess.DSessFromSess(ctx.Session)
+	dbData, ok := sess.GetDbData(ctx, dbName)
+
+	if !ok {
+		return noConflicts, fmt.Errorf("Could not load database %s", dbName)
```
zachmu, line 75: Use the not found error you corrected in the dolt_merge function

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
```diff
@@ -0,0 +1,200 @@
...
+
+func progFunc(progChan chan datas.PullProgress) {
+	done := false
+	for !done {
+		select {
+		case _, ok := <-progChan:
+			if !ok {
+				done = true
+			}
+		case <-time.After(250 * time.Millisecond):
```
zachmu, line 170: You probably don't want this. If the pull slows down enough, you'll stop reading from the prog channel, which means it could fill up and additional sends could block.

If you want to do this right, then you need to include the *sql.Context in your select here, and check to see if it's been canceled.

Or you could just do a range on the progChan and don't do any timeout logic, that's fine for a first pass.

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
```diff
@@ -0,0 +1,200 @@
...
+		return noConflicts, fmt.Errorf("failed to get remote db; %w", err)
+	}
+
+	ws, err := sess.WorkingSet(ctx, dbName)
+	if err != nil {
+		return noConflicts, err
+	}
+
+	var conflicts interface{}
+	for _, refSpec := range pullSpec.RefSpecs {
```
zachmu, line 111: Structuring this in a loop is misleading, because the pull logic in the body only happens once (right?)

Use a loop to determine the remoteTrackRef, then outside the loop do a single pull action.

integration-tests/bats/sql-remotes-fs.bats
```diff
@@ -0,0 +1,126 @@
...
+    cd dolt-repo-$$
+    mkdir "dolt-repo-clones"
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+@test "sql-remotes-fs: Add a file system based remote" {
```
zachmu, line 16: The test body doesn't seem to be using SQL. Just remove this for now I think

integration-tests/bats/sql-remotes-fs.bats
```diff
@@ -0,0 +1,126 @@
...
+    skip "todo dolt_remote function?"
+    mkdir remote
+    dolt remote add origin file://remote/
+    run dolt remote -v
+    [ $status -eq 0 ]
+    regex='file://.*/remote'
+    [[ "$output" =~ $regex ]] || false 
+}
+
+@test "sql-remotes-fs: Add a file system remote with a bad path" {
```
zachmu, line 26: Same comment

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+}
+
+@test "sql-remotes-server: push with only one argument" {
+    skip "todo implement SQL push, pull, fetch"
+    dolt remote add test-remote http://localhost:50051/test-org/test-repo
+    run dolt sql -q "select dolt_push('test-remote')"
+    [ "$status" -eq 1 ]
+}
+
+@test "sql-remotes-server: push and pull master branch from a remote" {
```
zachmu, line 30: Isolate the SQL push and pull aspects. Fine to have a test that does both, but you definitely wants tests that only exercise one function (especially at this stage).

So do setup with the CLI, then make assertions about the output of the SQL func.

Your tests below already do this, they just need tweaked names and remove TODO and commented lines

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    dolt remote add test-remote http://localhost:50051/test-org/test-repo
+    # TODO dolt_push
+    #run dolt sql -q "select dolt_push('test-remote', 'master')"
+    run dolt push test-remote master
+    [ "$status" -eq 0 ]
+    [ -d "$BATS_TMPDIR/remotes-$$/test-org/test-repo" ]
+    run dolt sql -q "select dolt_pull('test-remote')"
+    [ "$status" -eq 0 ]
```
zachmu, line 38: Need more verification here. What should the data look like after a successful pull?

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    [[ ! "$output" =~ "panic:" ]] || false
+    dolt sql -q "create table test (pk int, c1 int, primary key(pk))"
+    dolt add .
+    dolt commit -m "Added test table"
+    run dolt sql -q "select dolt_push()"
+    [ "$status" -eq 0 ]
+}
+
+@test "sql-remotes-server: push and pull with docs from remote" {
+    skip "todo pull docs?"
```
zachmu, line 69: We don't touch the filesystem when running the SQL server. A SQL session doesn't really have a notion of files the way the CLI does.

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    run dolt push test-remote master
+    [ "$status" -eq 0 ]
+    [ -d "$BATS_TMPDIR/remotes-$$/test-org/test-repo" ]
+    run dolt sql -q "select dolt_pull('test-remote')"
+    [ "$status" -eq 0 ]
+    #[[ "$output" =~ "Everything up-to-date" ]] || false
+}
+
+@test "sql-remotes-server: push and pull non-master branch from remote" {
+    skip "todo implement SQL push, pull, fetch"
```
zachmu, line 43: The skip messages on these are too general, they all say the same thing.

You've already implemented pull, e.g. and not all the tests need both push and fetch

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    dolt fetch
+    run dolt tag -v
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "v1" ]] || false
+    [[ "$output" =~ "tag message" ]] || false
+    [[ "$output" =~ "other_tag" ]] || false
+    [[ "$output" =~ "other message" ]] || false
+}
+
+@test "sql-remotes-server: dolt fetch" {
```
zachmu, line 184: This test body doesn't do what the label indicates

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    [[ "$output" =~ "unknown remote" ]] || false
+    run dolt fetch test-remote refs/heads/master:refs/remotes/test-remote/poop
+    [ "$status" -eq 0 ]
+    [ "$output" = "" ]
+    run dolt branch -v -a
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "remotes/test-remote/poop" ]] || false
+}
+
+@test "sql-remotes-server: dolt fetch with docs" {
```
zachmu, line 205: Don't worry about docs

integration-tests/bats/sql-remotes-server.bats
```diff
@@ -0,0 +1,556 @@
...
+    dolt add other
+    dolt commit -m "added other table"
+    dolt fetch
+    # TODO dolt_push
+    #dolt sql -q "select dolt_push('-f', 'origin', 'master')"
+    dolt push -f origin master
+    cd ../../
+    run dolt fetch test-remote
+    [ "$status" -ne 0 ]
+    run dolt sql -q "select dolt_pull()"
```
zachmu, line 500: Need more validation of the result of the pull for these tests to be useful

zachmu (approved):
Overall this looks good, I like the new abstractions and the refactoring. Mostly suggestions about naming and the contents of the bats tests.

go/libraries/doltcore/env/actions/branch.go
max-hoffman: `func Is(err error, target error) bool`

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
max-hoffman: i think multiple remote tracking branches per remote is a thing https://git-scm.com/book/en/v2/Git-Internals-The-Refspec

go/libraries/doltcore/env/actions/branch.go
zachmu: no, I mean the previous version had !=, and you changed it to == (effectively)

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
```diff
@@ -0,0 +1,213 @@
...
+	err = actions.FetchFollowTags(ctx, dbData.Rsw.TempTableFilesDir(), srcDB, dbData.Ddb, runProgFuncs, stopProgFuncs)
+	if err != nil {
+		return noConflicts, err
+	}
+
+	return noConflicts, nil
+}
+
+func pullerProgFunc(ctx context.Context, pullerEventCh <-chan datas.PullerEvent) {
+	for {
```
zachmu, line 157: I don't understand the point of having two selects in this loop like this. What am I missing?

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
```diff
@@ -0,0 +1,213 @@
...
+			return
+		case <-pullerEventCh:
+		default:
+		}
+	}
+}
+
+func progFunc(ctx context.Context, progChan <-chan datas.PullProgress) {
+	for {
+		select {
```
zachmu, line 174: Same question

go/cmd/dolt/commands/push.go
```diff
@@ -228,7 +228,7 @@ func pullerProgFunc(pullerEventCh chan datas.PullerEvent) {
 	}
 }
 
-func progFunc(progChan chan datas.PullProgress) {
+func progFunc(ctx context.Context, progChan chan datas.PullProgress) {
```
zachmu, line 231: You changed this interface, but didn't change the behavior here (still relies on timeouts or channel close to exit the loop)

Add some TODOs?

zachmu (approved):
LG, just a couple comments

go/cmd/dolt/commands/push.go
max-hoffman: i don't understand how to properly drain channels if we short circuit. maybe doesn't matter in the cli because the process won't be long-running, i'll leave a commented block with the `<-ctx.Done(): return`

go/libraries/doltcore/sqle/dfunctions/dolt_pull.go
max-hoffman: i've only seen short circuits implemented this way in the docs i've read. granted i don't understand the tradeoffs of short circuiting (the extra select) vs a 50/50 chance of exiting if we can both read from the channel and exit (second select)

PR #2104, opened 2021-09-08
Modify bad row error format

### LINKED ISSUE #2103: Column names are not shown in a readable way when table imports fail
`
doltcli.dolt.DoltException: ('table import businesses -u --pk=name,business_type,state_registered /tmp/tmpu4lfpowa.csv', b'Rows Processed: 2, Additions: 2, Modifications: 0, Had No Effect: 0', b'\n\nA bad row was encountered while moving data.\nBad Row: \n"985011267" is not valid for "VarString(utf8mb4_0900_ai_ci, 5, SQL: Char)"\nThese can be ignored using the \'--continue\'\n', 1)
`

It would be helpful to instead see the column name that `utf8mb4_0900_ai_ci` represents.

abmyii:
Closes #2103.

Before:
```python
A bad row was encountered while moving data.
Bad Row: 
"03AGDBQ24TPRIFU0LRPRHLHU8KBOK5SQSMCT8BMVFJNANULFUG4RDHNZPOV4PJNMN0DOD1PM2V5AQQTFP1HZXZG5J5BJTP7BN0A8WLBZYMBJIRU7BWTYCNXTCQUD1UYWITEJIZR5EWWP1NM4OR40P6FTQL4NHKIA870UAFXSPRF0AIK_45YNRU98ZQOZQNJFKNF89FW9" is not valid for "VarString(utf8mb4_0900_ai_ci, 180, SQL: VarChar)"
These can be ignored using the '--continue'
```

After:
```python
A bad row was encountered while moving data.
Bad Row: 
"03AGDBQ24TPRIFU0LRPRHLHU8KBOK5SQSMCT8BMVFJNANULFUG4RDHNZPOV4PJNMN0DOD1PM2V5AQQTFP1HZXZG5J5BJTP7BN0A8WLBZYMBJIRU7BWTYCNXTCQUD1UYWITEJIZR5EWWP1NM4OR40P6FTQL4NHKIA870UAFXSPRF0AIK_45YNRU98ZQOZQNJFKNF89FW9" is not valid for column "street_physical" (type "VarString(utf8mb4_0900_ai_ci, 180, SQL: VarChar)")
These can be ignored using the '--continue'
```

WIP: BATS test still needs to be written

### DIFF
--- a/integration-tests/bats/import-update-tables.bats
+++ b/integration-tests/bats/import-update-tables.bats
@@ -16,6 +16,14 @@ CREATE TABLE test (
 );
 SQL
 
+    cat <<SQL > 1pk1col-char-sch.sql
+CREATE TABLE test (
+  pk BIGINT NOT NULL COMMENT 'tag:0',
+  c char(5) COMMENT 'tag:1',
+  PRIMARY KEY (pk)
+);
+SQL
+
     cat <<DELIM > 1pk5col-ints.csv
 pk,c1,c2,c3,c4,c5
 0,1,2,3,4,5
@@ -184,6 +192,20 @@ SQL
     [[ "${lines[4]}" =~ "title" ]]      || false
     [[ "${lines[5]}" =~ "start date" ]] || false
     [[ "${lines[6]}" =~ "end date" ]]   || false
+}
+
+@test "import-update-tables: update table with incorrect length char throws bad row error" {
+    cat <<DELIM > 1pk1col-rpt-chars.csv
+pk,c
+1,"123456"
+DELIM
+
+    dolt sql < 1pk1col-char-sch.sql
+    run dolt table import -u test 1pk1col-rpt-chars.csv
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "A bad row was encountered while moving data" ]] || false
+    [[ "$output" =~ "Bad Row:" ]] || false
+    [[ "$output" =~ '"123456" is not valid for column "c" (type "CHAR(5)")' ]] || false
 }
 
 @test "import-update-tables: update table with repeat pk in csv throws error" {
--- a/integration-tests/bats/import-update-tables.bats
+++ b/integration-tests/bats/import-update-tables.bats
@@ -19,7 +19,7 @@ SQL
     cat <<SQL > 1pk1col-char-sch.sql
 CREATE TABLE test (
   pk BIGINT NOT NULL COMMENT 'tag:0',
-  c char(5) COMMENT 'tag:1',
+  c CHAR(5) COMMENT 'tag:1',
   PRIMARY KEY (pk)
 );
 SQL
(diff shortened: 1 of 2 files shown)


timsehn:
I'm not sure @abmyii should do this but part of the problem is `col.TypeInfo.String()` is `"VarString(utf8mb4_0900_ai_ci, 180, SQL: VarChar)"`. Is there not a way to make this `varchar(5)` like is in the schema?

abmyii:
@timsehn This is where it is formatted: https://github.com/dolthub/dolt/blob/e69b8123990ed6edf57fb3196150fd4141330be6/go/libraries/doltcore/schema/typeinfo/varstring.go#L248

I expect there was a reason for it being formatted in that way, so I won't change it unless given the go-ahead.

timsehn:
@Hydrocharged why did you format it this way?

Hydrocharged:
That string output is more for debugging the actual typeinfo rather than for display. We should be calling `cop.TypeInfo.SqlType().String()` for anything that we’re displaying that’s public-facing or will reasonably be seen by a user, such as this error.

timsehn:
Cool. @abmyii, you mind changing?

abmyii:
Yup, will do!

abmyii:
BATS test added - hopefully it is a sufficient test.

@Hydrocharged I had to use `ToSqlType()` as apparently `SqlType` isn't an attribute/member.

integration-tests/bats/import-update-tables.bats
```diff
@@ -14,6 +14,14 @@ CREATE TABLE test (
   c5 BIGINT COMMENT 'tag:5',
   PRIMARY KEY (pk)
 );
+SQL
+
+    cat <<SQL > 1pk1col-char-sch.sql
+CREATE TABLE test (
+  pk BIGINT NOT NULL COMMENT 'tag:0',
+  c char(5) COMMENT 'tag:1',
```
Hydrocharged, line 22: Just a small nitpick: let's uppercase `char` here since we're uppercasing all of the other type declarations.

Hydrocharged (approved):
This looks good!

`ToSqlType()` is the correct function, I was typing from memory rather than referencing the code. I was close (:

Hydrocharged:
I'll approve and run CI against your PR after addressing my small nit (CI takes a while so I'd rather run it once), and assuming all of the tests pass you're free to merge!

integration-tests/bats/import-update-tables.bats
abmyii: Ah true - thanks.

abmyii:
@Hydrocharged How can I fix the failing CI task on this and PR #2112?

Hydrocharged:
It looks like this is affecting all of our PRs, we'll get a fix out for this in a bit.

abmyii:
OK, thanks!

Hydrocharged:
We've got a fix in for the failing tests, you may need to merge or rebase off of `master`.

abmyii:
Rebased.

Hydrocharged:
I'm going to go ahead and merge this in. Although CI is failing, the actual failures are all unrelated to your changes, and they're just being flaky (failed host resolution, etc.)

abmyii:
Thank you very much for your help @Hydrocharged.

Hydrocharged:
Anytime (:

PR #2109, opened 2021-09-09
go/libraries/doltcore/dbfactory: Change our GRPC remote DB factory to take the GRPC dialer through a param.

reltuk:
Prior to this change, we used to set a global variable in the dbfactory package
which referred to whichever DoltEnv was loaded most recently. It was racey and
it has pretty terrible semantics.

### DIFF
--- a/go/cmd/dolt/dolt.go
+++ b/go/cmd/dolt/dolt.go
@@ -227,7 +227,7 @@ func runMain() int {
 	warnIfMaxFilesTooLow()
 
 	ctx, cancelF := context.WithCancel(context.Background())
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 	go func() {
 		<-c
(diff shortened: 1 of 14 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -227,7 +227,7 @@ func runMain() int {
 	warnIfMaxFilesTooLow()
 
 	ctx, cancelF := context.WithCancel(context.Background())
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
reltuk, line 230: This is spuriously caught up in here due to a change in CI govet.

go/utils/remotesrv/main.go
```diff
@@ -73,9 +73,8 @@ func main() {
 }
 
 func waitForSignal() {
-	c := make(chan os.Signal)
-	signal.Notify(c, os.Interrupt)
-	signal.Notify(c, os.Kill)
+	c := make(chan os.Signal, 1)
```
reltuk, line 76: Same as channel change in dolt.go.

bheni (approved):
LGTM

PR #2112, opened 2021-09-09
Only show key name, not column tag number

### LINKED ISSUE #2107: Bad error message for foreign key constraint violation in import
Sample error message:

```
An error occurred moving data:

cause: cannot add or update a child row - Foreign key violation on fk: `8n27906l`, table: `businesses`, referenced table: `naics`, key: `(10772,null_value)`
```

The column tag number should not be included in the error message, it's confusing.

abmyii:
Closes #2107.

### DIFF
--- a/go/libraries/doltcore/table/editor/sessioned_table_editor.go
+++ b/go/libraries/doltcore/table/editor/sessioned_table_editor.go
@@ -299,7 +299,7 @@ func (ste *sessionedTableEditor) onDeleteHandleRowsReferencingValues(ctx context
 				}
 			}
 		case doltdb.ForeignKeyReferenceOption_DefaultAction, doltdb.ForeignKeyReferenceOption_NoAction, doltdb.ForeignKeyReferenceOption_Restrict:
-			indexKeyStr, _ := types.EncodedValue(ctx, indexKey)
+			indexKeyStr, _ := formatKey(ctx, indexKey)
 			return sql.ErrForeignKeyChildViolation.New(foreignKey.Name, foreignKey.TableName, foreignKey.ReferencedTableName, indexKeyStr)
 		default:
 			return fmt.Errorf("unknown ON DELETE reference option on `%s`: `%s`", foreignKey.Name, foreignKey.OnDelete.String())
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/table/editor/sessioned_table_editor.go
```diff
@@ -299,7 +299,8 @@ func (ste *sessionedTableEditor) onDeleteHandleRowsReferencingValues(ctx context
 				}
 			}
 		case doltdb.ForeignKeyReferenceOption_DefaultAction, doltdb.ForeignKeyReferenceOption_NoAction, doltdb.ForeignKeyReferenceOption_Restrict:
-			indexKeyStr, _ := types.EncodedValue(ctx, indexKey)
+			value, _ := indexKey.Get(1)
```
zachmu, line 302: You should use the formatKey() function for this, here and in the other places you do it

zachmu (commented):
Thanks for the contribution, one suggestion

go/libraries/doltcore/table/editor/sessioned_table_editor.go
abmyii: Sorry, I am not very familiar with Go - how would I use it?

abmyii:
```go
indexKeyStr, _ := formatKey(ctx, indexKey)
return sql.ErrForeignKeyChildViolation.New(foreignKey.Name, foreignKey.TableName, foreignKey.ReferencedTableName, indexKeyStr)
```

I am using it like this, but in the output it shows as `["green"]`, not `"green"` for some reason. Indexing it (`indexKeyStr[0]`) doesn't give the first value as expected - instead `%!s(uint8=91)` is output.

zachmu: You should use formatKey, and we'll evolve how it prints keys over time. Right now it will put keys in square brackets, so make the test expectations match

zachmu (approved):
LG, thanks!

PR #2114, opened 2021-09-09
Bump Go version to 1.17

Hydrocharged:


reltuk (approved):
Why don't we have to change `.github/workflows/ci-compatibility-tests.yaml`?

Otherwise LGTM!

Hydrocharged:
Because I missed it XD

PR #2116, opened 2021-09-09
Make hasNull true when val is types.NullValue

### LINKED ISSUE #2108: NULL entries treated as foreign key constraint violations on import
A NULL value never violates a foreign key constraint, by definition. But when using `dolt table import`, we treat null values for columns with this constraint as an error.

abmyii:
Closes #2108. I believe my fix is a hack and that `ok` should be `false` when `value` is `null_value` but isn't being set due to some underlying bug.


abmyii:
https://github.com/dolthub/dolt/issues/2108#issuecomment-916451170

I tested my assumption that this may break NOT NULL FK fields - it does not.

zachmu (commented):
@bheni The fact that this change is necessary points to a larger problem in import.

We don't ever intentionally put types.NullValue in a row, right? Why is it showing up here?

abmyii:
Superseded by https://github.com/dolthub/dolt/pull/2143.

PR #2117, opened 2021-09-09
go: remotestorage: hedge_test.go: Fix these tests to be less flakey, use determinstic interleaving of goroutines and hedge requests by instrumenting hedger's time.After.

reltuk:


andrew-wm-arthur (approved):
LGTM

PR #2125, opened 2021-09-10
Support GetDiffsWithFilter for keyless tables

tbantle22:
On dolthub we're using `diff.NewAsyncDiffer` instead of `diff.NewRowDiffer` to list row diffs (see [here](https://github.com/dolthub/ld/blob/master/go/services/dolthubapi/pkg/domain/repositorydata.go#L1534)), which doesn't have support for diffing keyless tables (https://www.dolthub.com/repositories/ttest/keyless/pulls/1/compare). In order to use `NewRowDiffer`, we need to add a `GetDiffsWithFilter` method for keyless tables.

I tried writing tests for this, but was having a hard time figuring out what the rows map for keyless tables should look like

### DIFF
--- a/go/libraries/doltcore/diff/async_differ.go
+++ b/go/libraries/doltcore/diff/async_differ.go
@@ -35,6 +35,9 @@ type RowDiffer interface {
 	// GetDiffs returns the requested number of diff.Differences, or times out.
 	GetDiffs(numDiffs int, timeout time.Duration) ([]*diff.Difference, bool, error)
 
+	// GetDiffsWithFilter returns the requested number of filtered diff.Differences, or times out.
+	GetDiffsWithFilter(numDiffs int, timeout time.Duration, filterByChangeType types.DiffChangeType) ([]*diff.Difference, bool, error)
+
 	// Close closes the RowDiffer.
 	Close() error
 }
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/diff/async_differ.go
```diff
@@ -214,13 +215,30 @@ func (kd *keylessDiffer) GetDiffs(numDiffs int, timeout time.Duration) (diffs []
...
 
-			kd.df, kd.copiesLeft, err = convertDiff(d)
+			df, copiesLeft, err := convertDiff(d)
 			if err != nil {
 				return nil, false, err
 			}
+			if pred(&df) {
+				kd.df = df
+			}
+			kd.copiesLeft = copiesLeft
```
tbantle22, line 225: Not sure if this is right/the right place for this

go/libraries/doltcore/diff/async_differ.go
```diff
@@ -214,13 +215,30 @@ func (kd *keylessDiffer) GetDiffs(numDiffs int, timeout time.Duration) (diffs []
...
 			}
 
-			kd.df, kd.copiesLeft, err = convertDiff(d)
+			df, copiesLeft, err := convertDiff(d)
 			if err != nil {
 				return nil, false, err
 			}
+			if pred(&df) {
+				kd.df = df
+			}
```
andrew-wm-arthur, line 224: If `&df` doesn't satisfy the predicate, we need to keep looking for a difference that does

```suggestion
			ok := false
			for !ok {
				kd.df, kd.copiesLeft, err = convertDiff(d)
				if err != nil {
					return nil, false, err
				}
				ok = pred(&df)
			}
```
(couldn't highlight 225 because Github sux)

andrew-wm-arthur (approved):
LGTM

PR #2131, opened 2021-09-12
Create Dolt Database

andrew-wm-arthur:


go/libraries/doltcore/env/memory.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/dolthub/dolt/go/store/chunks"
+	"github.com/dolthub/dolt/go/store/hash"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdocs"
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
+)
+
+func NewMemoryDbData(ctx context.Context) (DbData, error) {
```
zachmu, line 31: Why is this in the env package?

zachmu: I mean all these functions, not just this one

go/libraries/doltcore/env/memory.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (m MemoryRepoState) CWBHeadSpec() *doltdb.CommitSpec {
+	spec, err := doltdb.NewCommitSpec(m.CWBHeadRef().GetPath())
+	if err != nil {
+		panic(err)
+	}
+	return spec
+}
+
+func (m MemoryRepoState) UpdateStagedRoot(ctx context.Context, newRoot *doltdb.RootValue) error {
```
zachmu, line 121: Mind adding a TODO to this and the other root update method? I'm in the middle of killing them in the interface, want to make sure to get them here too

go/libraries/doltcore/env/memory.go
```diff
@@ -0,0 +1,229 @@
...
+func (m MemoryRepoState) AddRemote(name string, url string, fetchSpecs []string, params map[string]string) error {
+	return fmt.Errorf("cannot insert a remote in a memory database")
+}
+
+func (m MemoryRepoState) GetBranches() (map[string]BranchConfig, error) {
+	return make(map[string]BranchConfig), nil
+}
+
+func (m MemoryRepoState) UpdateBranch(name string, new BranchConfig) error {
+	return nil
```
zachmu, line 212: Should throw an error

go/libraries/doltcore/sqle/database.go
```diff
@@ -140,6 +153,41 @@ func NewDatabase(name string, dbData env.DbData, editOpts editor.Options) Databa
 	}
 }
 
+// GetInitialDBState returns the InitialDbState for |db|.
+func GetInitialDBState(ctx context.Context, db Database) (dsess.InitialDbState, error) {
```
zachmu, line 157: Add a note here that this is only to support the creation of in-memory databases.

The method of using repo state to initialize these things is pretty janky

integration-tests/mysql-client-tests/elixir/lib/simple.ex
```diff
@@ -26,7 +26,7 @@ defmodule SmokeTest do
     {:ok, result} = MyXQL.query(pid, "UPDATE test SET pk = pk where pk = 0")
     myTestFunc(result.num_rows, 1)
 
-    {:ok, result} = MyXQL.query(pid, "INSERT INTO test VALUES (0, 0) ON DUPLICATE KEY UPDATE value = value")
+    {:ok, result} = MyXQL.query(pid, "INSERT INTO test VALUES (0, 0) ON DUPLICATE KEY UPDATE `value` = `value`")
```
zachmu, line 29: This seems like a bad regression. File an issue?

zachmu (approved):
I don't love it, but fine moving forward with this for now.

Particularly, the coupling of repo state and creation of in-memory databases is pretty nasty, should clean that up soon

VinaiRachakonda:
This is pretty dope. Think the memory implementations will help new contributors a lot with understanding the fundamentals

PR #2132, opened 2021-09-13
Fixed VRW reference issues

Hydrocharged:
Previously we were using memory value stores in some places, and everything seemed to be working, but I think our garbage collection changes exposed some inconsistencies that were causing bugs. This PR threads values stores to where they need to be and removes the majority of the places where we were using memory value stores.

This also solves a multitude of fuzzer issues, including the `corrupted database` error:
https://github.com/dolthub/fuzzer/issues/337
And also the cause of some rows not being found and silently causing some operations to no-op (bats test included):
https://github.com/dolthub/fuzzer/issues/597


bheni (approved):
lgtm

PR #2135, opened 2021-09-13
/go/libraries/doltcore/{env,doltdb,ref}: Add `init.defaultBranch` to Dolt Config

andrew-wm-arthur:
mirrors Git config functionality: 
```
git init
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint:   git config --global init.defaultBranch <name> 
```


go/libraries/doltcore/env/environment.go
```diff
@@ -155,6 +160,11 @@ func Load(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys, urlStr,
 	return dEnv
 }
 
+func GetDefaultInitBranch(dEnv *DoltEnv) string {
```
bheni, line 163: when used in other packages usage looks like:

`env.GetDefaultInitBranch(dEnv)`

If it were a member of dEnv it would be:

`dEnv.GetDefaultInitBranch()`

which feels a bit better.

go/libraries/doltcore/env/environment.go
```diff
@@ -412,8 +421,8 @@ func (dEnv *DoltEnv) InitDBWithTime(ctx context.Context, nbf *types.NomsBinForma
 
 // InitializeRepoState writes a default repo state to disk, consisting of a master branch and current root hash value.
 func (dEnv *DoltEnv) InitializeRepoState(ctx context.Context) error {
-	cs, _ := doltdb.NewCommitSpec(doltdb.MasterBranch)
-	commit, err := dEnv.DoltDB.Resolve(ctx, cs, nil)
+	init := GetDefaultInitBranch(dEnv)
```
bheni, line 424: branchName is a better variable name than init.

integration-tests/bats/config.bats
```diff
@@ -186,4 +186,25 @@ teardown() {
...
+@test "config: Set default init branch" {
+    dolt config --global --add user.name "bats tester"
+    dolt config --global --add user.email "joshn@doe.com"
+
+    dolt config --global --add init.default_branch "main"
+    run dolt config --list
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "init.default_branch = main" ]]
+
+    dolt init
```
bheni, line 200: an argument to init that lets you specify the branch makes this more accessible for users.

bheni (approved):
lgtm. couple comments.

go/libraries/doltcore/env/config.go
```diff
@@ -38,6 +38,10 @@ const (
 
 	DoltEditor = "core.editor"
 
+	// todo: analogous Git param is "init.defaultBranch", but
+	//   Dolt config params are case-insensitive
+	InitBranchName = "init.default_branch"
```
Hydrocharged, line 43: If we're case insensitive, perhaps we should just allow `init.defaultbranch`? That way people can still type `init.defaultBranch` and it'll match.

go/libraries/doltcore/env/environment.go
andrew-wm-arthur: I changed the param to be a `config.Config`

go/libraries/doltcore/env/config.go
andrew-wm-arthur: good call

PR #2137, opened 2021-09-14
Made SQL commits involving a dolt commit atomic

zachmu:


go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -40,6 +42,15 @@ func TestDoltTransactionCommitOneClient(t *testing.T) {
 				Query:    "/* client a */ SET @@autocommit=0;",
 				Expected: []sql.Row{{}},
 			},
+			// start transaction implicitly commits the current transaction, so we have to do so before we turn on dolt commits
```
Hydrocharged, line 45: Perhaps we should also put this in some documentation somewhere, or perhaps as a comment above the `dolt_transaction_commit` session var.

Hydrocharged (approved):
🎆

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
zachmu: It's already in the MySQL docs, but we should mention it when we do our next doc sweep for transaction stuff

PR #2139, opened 2021-09-14
Fetch refactor

max-hoffman:
- consolidate `getRefSpecs` functions from cli/env
- fetch argument switching matches git more closely -- specifying the remote is necessary, no default to origin

### DIFF
--- a/go/cmd/dolt/commands/fetch.go
+++ b/go/cmd/dolt/commands/fetch.go
@@ -17,13 +17,12 @@ package commands
 import (
 	"context"
 
-	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
-
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
 	eventsapi "github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi/v1alpha1"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
 	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 	"github.com/dolthub/dolt/go/libraries/utils/filesys"
 )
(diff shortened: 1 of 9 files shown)


go/libraries/doltcore/env/remotes.go
```diff
@@ -241,20 +241,18 @@ func NewFetchOpts(args []string, rsr RepoStateReader) (Remote, []ref.RemoteRefSp
...
-			remoteOK = true
-			args = args[1:]
-		}
+	var remName string
+	if len(args) == 0 {
+		remName = "origin"
+	} else {
+		remName = args[0]
+		args = args[1:]
 	}
```
max-hoffman, line 250: if there is a first argument, must be valid remote (like git)

go/cmd/dolt/commands/fetch.go
```diff
@@ -77,164 +76,15 @@ func (cmd FetchCmd) Exec(ctx context.Context, commandStr string, args []string,
...
-			args = args[1:]
-		}
-	}
-
-	if !remoteOK {
-		return env.NoRemote, nil, errhand.BuildDError("error: unknown remote").SetPrintUsage().Build()
-	}
-
-	var rs []ref.RemoteRefSpec
 	var verr errhand.VerboseError
```
zachmu, line 79: In general, this is an anti-pattern: very easy to just not handle an error when it's declared outside where it's used.

Handle each branch's error separately, returning immediately.

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -358,3 +358,68 @@ func FetchRemoteBranch(ctx context.Context, tempTablesDir string, rem env.Remote
 
 	return srcDBCommit, nil
 }
+
+func FetchRefSpecs(ctx context.Context, dbData env.DbData, refSpecs []ref.RemoteRefSpec, remote env.Remote, mode ref.UpdateMode, progStarter ProgStarter, progStopper ProgStopper) error {
```
zachmu, line 362: Method docs please

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -358,3 +358,68 @@ func FetchRemoteBranch(ctx context.Context, tempTablesDir string, rem env.Remote
...
+			if remoteTrackRef != nil {
+				rsSeen = true
+				srcDBCommit, err := FetchRemoteBranch(ctx, dbData.Rsw.TempTableFilesDir(), remote, srcDB, dbData.Ddb, branchRef, remoteTrackRef, progStarter, progStopper)
+				if err != nil {
+					return err
+				}
+
+				switch mode {
+				case ref.ForceUpdate:
+					err := dbData.Ddb.SetHeadToCommit(ctx, remoteTrackRef, srcDBCommit)
```
zachmu, line 388: This is unfortunate, should add a TODO here (can't be used safely in SQL context)

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -358,3 +358,68 @@ func FetchRemoteBranch(ctx context.Context, tempTablesDir string, rem env.Remote
...
+						return fmt.Errorf("%w: %s", ErrCantFF, err.Error())
+					}
+					if !ok {
+						return ErrCantFF
+					}
+
+					switch err {
+					case doltdb.ErrUpToDate:
+					case doltdb.ErrIsAhead, nil:
+						err = dbData.Ddb.FastForward(ctx, remoteTrackRef, srcDBCommit)
```
zachmu, line 404: Same issue here

zachmu: This is fine, just needs to be documented and we need to come up with a reasonable set of behaviors for open connections

integration-tests/bats/sql-fetch.bats
```diff
@@ -217,7 +217,7 @@ teardown() {
 @test "sql-fetch: dolt_fetch unknown remote fails" {
     cd repo2
     dolt remote remove origin
-    run dolt sql -q "select dolt_fetch('unknown', 'master')"
+    run dolt sql -q "select dolt_fetch('unknown')"
```
zachmu, line 220: Is this not a valid test?

zachmu (approved):
Looks fine, I like the refactor

integration-tests/bats/sql-fetch.bats
max-hoffman: both are valid, but remotes are resolved before refspecs, so i thought the addition of a branch name distracted from what was actually being tested

go/libraries/doltcore/env/actions/remotes.go
max-hoffman: lots of server locking todos

PR #2143, opened 2021-09-16
Import fix to not put NullValue into types.Tuples for map values

bheni:


bheni:
Includes changes submitted by @abmyii from https://github.com/dolthub/dolt/pull/2116

zachmu (approved):
LGTM, should tag this on the PR

https://github.com/dolthub/dolt/issues/1138

integration-tests/bats/foreign-keys.bats
```diff
@@ -1651,3 +1651,37 @@ SQL
     dolt add .
     dolt commit -m 'this should not break'
 }
+
+@test "foreign-keys: dolt table import with null in nullable FK field should work (issue #2108)" {
```
timsehn, line 1655: Don't think we should have issue #s in bats

timsehn (commented):
Remove issue number please

integration-tests/bats/foreign-keys.bats
abmyii: My bad - I assumed it would make it easier to understand the context. I'll remember this for next time.

PR #2148, opened 2021-09-17
no panic if HOME does not exist

max-hoffman:
https://github.com/dolthub/dolt/issues/2147

### DIFF
--- a/go/cmd/dolt/dolt.go
+++ b/go/cmd/dolt/dolt.go
@@ -236,7 +236,7 @@ func runMain() int {
 	root, err := env.GetCurrentUserHomeDir()
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -236,6 +236,7 @@ func runMain() int {
 	root, err := env.GetCurrentUserHomeDir()
 
 	if err != nil {
+		cli.PrintErrln(color.RedString("Failed to load the HOME dir: %v", err))
```
timsehn, line 239: I think we go with directory.

timsehn (commented):
Nit on error message

Hydrocharged (approved):
LGTM

PR #2151, opened 2021-09-18
replicate HEAD on commit

max-hoffman:
- prototype commit replication
- only works for cli, will work for SQL after https://github.com/dolthub/dolt/pull/2137/files is merged
- because the init state is in `cli`, the target dataset in `datas`, and logic for manipulating datasets is constrained to `doltdb`, this is what fell out without rearranging dependencies or exposing private fields
- `type CommitHook func(ctx context.Context, ds Dataset, db Database) error` isn't super flexible


zachmu (commented):
This seems like a good first draft, my main comment is that commit hooks should be configured on the stores themselves rather than provided as params in interface methods.

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func pullerProgFunc(ctx context.Context, pullerEventCh <-chan datas.PullerEvent) {
+	for {
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
max-hoffman: the golang book says a switch with multiple available channels will choose equally between them. i've seen the same pattern in the golang source code -- check for done on its own to avoid a case where we have 50/50 chance between being terminating or continuing to consume from the channel. i'm not sure it makes that big of a difference, so i can remove if that explanation doesn't make sense to you.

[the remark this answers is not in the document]
reltuk: Especially since we don’t do anything at all with the result from the non-done channel, I think foregoing it is fine. With contexts in particular, just checking `ctx.Err() == nil` before going into the select might be clearer/syntactically cleaner.

go/libraries/doltcore/env/environment.go
```diff
@@ -56,6 +58,20 @@ const (
 	tempTablesDir = "temptf"
 )
 
+func initializeCommitHooks(ctx context.Context, dEnv *DoltEnv) (postCommitHooks []datas.CommitHook, err error) {
+	backupUrl := os.Getenv(doltdb.ReplicateToRemoteKey)
+	if backupUrl != "" {
+		var destDB *doltdb.DoltDB
+		_, destDB, err = CreateRemote(ctx, "backup", backupUrl, nil, dEnv)
```
zachmu, line 65: This isn't the right interface here. Let the user name the remote, use another env var (or put it in config as a followup)

go/libraries/doltcore/env/environment.go
```diff
@@ -158,6 +174,15 @@ func Load(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys, urlStr,
 		}
 	}
 
+	if dbLoadErr == nil {
+		postCommitHooks, dbLoadErr := initializeCommitHooks(ctx, dEnv)
+		if dbLoadErr != nil {
+			dEnv.DBLoadError = dbLoadErr
+		} else {
+			dEnv.DoltDB = ddb.WithCommitHooks(ctx, postCommitHooks)
```
zachmu, line 182: This is phrased badly: you are using a With method on a different variable than the one you are assigning it to. They point to the same object in this case, but that could change and then this code would break. Use:

```go
dEnv.DoltDB = dEnv.DoltDB.WithCommitHooks(...)
```

go/libraries/doltcore/env/remotes.go
```diff
@@ -57,6 +57,17 @@ type Remote struct {
 	dialer     dbfactory.GRPCDialProvider
 }
 
+func CreateRemote(ctx context.Context, remoteName, remoteUrl string, params map[string]string, dialer dbfactory.GRPCDialProvider) (Remote, *doltdb.DoltDB, error) {
```
zachmu, line 60: This isn't really CreateRemote, it's more like GetRemote

go/store/datas/database_common.go
```diff
@@ -36,14 +37,21 @@ import (
...
+	rt              rootTracker
+	postCommitHooks []CommitHook
 }
 
 var (
 	ErrOptimisticLockFailed = errors.New("optimistic lock failed on database Root update")
 	ErrMergeNeeded          = errors.New("dataset head is not ancestor of commit")
 )
 
+type CommitHook interface {
```
zachmu, line 49: Need docs on this type and methods

zachmu (approved):
Generally looks good, just some comments on naming and interfaces

go/libraries/doltcore/env/environment.go
max-hoffman: leaning towards having a user indicate a named repo-state remote, and passing `ReplicateToRemoteKey = <remote-name>` as the env variable

max-hoffman: if you think that's a bad idea and backups should be namespace separated lmk before standup

go/libraries/doltcore/doltdb/commit_hooks_test.go
```diff
@@ -0,0 +1,149 @@
...
+	meta, err = NewCommitMeta(committerName, committerEmail, "Sample data")
+	if err != nil {
+		t.Error("Failed to commit")
+	}
+
+	// setup hook
+	hook := NewReplicateHook(destDB, tmpDir)
+	ddb.WithCommitHooks(ctx, []datas.CommitHook{hook})
+
+	t.Run("replicate to backup remote", func(t *testing.T) {
```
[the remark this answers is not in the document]
max-hoffman: i agree but that will be easier with the test helper i'm adding the read replica PR. until then `CommitWithWorkingSet` is bats tested in both CLI and SQL

PR #2152, opened 2021-09-20
Fix update on web issue

bheni:
BulkImportIEA and BulkImportTEA only support Commit, and not Rollback.  Dolt uses these for import operations which do not get an explicit Commit or Rollback call.  When used on the web for the in memory implementation Commit is used, and should not error.


bheni:
Fixes https://github.com/dolthub/ld/issues/7444

PR #2154, opened 2021-09-20
Rebase the database on transaction start

### LINKED ISSUE #2092: dolt sql-server must be restarted to see changes from other processes
Repro: start up a dolt server, connect to it.

In a separate process, run `dolt table import -u ` to add rows to a table (or use `dolt sql -q`, any command that modifies data works). The running server can't see this change, even after starting a new transaction.

If you attempt to modify data using the running server, it fails with "context canceled," and every subsequent query on this session (even SELECT statements) gets the same error.

In the debugger, I can see that `Ddb.ResolveWorkingSet(ctx, wsRef)` returns a stale working set value, indicating it must have a stale noms root.

Expectation is that either:

1) The running server can tolerate such writes from other processes, and sees them, like it would from another connected SQL client
2) The server locks the database and prevents such edits from other processes until it terminates

Option 2) is nearer to hand, but breaks a pattern we commonly see in the python world, something like this:

```python
import pandas as pd

from doltpy.cli import Dolt
from doltpy.cli.write import write_pandas


df = pd.read_sql_table('schools', 'mysql+pymysql://root@localhost:4000/us_schools')

df.loc[df.name == "PATHWAYS SCHOOL--EDGEWOOD", 'category'] = 'TEST'

dolt = Dolt('.')
write_pandas(
    dolt,
    'schools',
    df,
    import_mode='update',
    primary_key=['name', 'city', 'state']
)
```

This pattern uses SQL to read data, then `write_pandas` to write new data (which uses `dolt table import` under the hood).

Ideally we make all command line write paths tolerant of a running server, treating them like any other connected SQL client.

zachmu:
This fixes #2092

As part of this work, also cleaned up manifest reading and writing code. We now always read the manifest flexibly (any known version), and always write the most recent version.


andrew-wm-arthur (approved):
LGTM

PR #2182, opened 2021-09-25
FF head on read replica

max-hoffman:
- adds `dsqle.ReadReplicaDatabase` which implements `dsqle.SqlDatabase`
- expands uses of `dsqle.Database` to accommodate all `dsqle.SqlDatabase` implementations
- Read replica DB decorates `db.StartTransaction` to execute a fetch on `DOLT_READ_REPLICA_REMOTE`
- `ReadReplicaDatabase` is -not- read only right now, need another decorator or field
- cli and sql bats tests
- ~~sql dolt tests~~ this is skipped due to in process server conflicts with multiple ongoing tests
- bunch of sql shadow functions (create table, stage, commit, push) to allow for testing multiple dEnv's side-by-side


zachmu (approved):
Tests need a lot of work, but this is good enough to get in and iterate on

go/libraries/doltcore/sqle/database.go
```diff
@@ -1205,3 +1233,122 @@ func RegisterSchemaFragments(ctx *sql.Context, db Database, root *doltdb.RootVal
...
 }
+
+type ReadReplicaDatabase struct {
+	Database
+	headRef        ref.DoltRef
+	remoteTrackRef ref.DoltRef
+	remote         env.Remote
+	srcDB          *doltdb.DoltDB
+	tmpDir         string
+	wsMeta         *doltdb.WorkingSetMeta
```
zachmu, line 1244: This can't be set at construction

PR #2193, opened 2021-09-28
Bug fix for SQL export of enum columns

### LINKED ISSUE #2187: SQL export doesn't quote enums
Doesn't work on any database with string values

zachmu:
Fixes #2187


max-hoffman (approved):
LGMT, as far as i can tell the only functional change is the Enum Identifier?

PR #2196, opened 2021-09-28
config, clone, backups refactor

max-hoffman:
- previously env variables are now dolt config variables for better locality
- move clone logic into library functions
- multi env test setup is more like a builder and implements clone so that we can have repos with common ancestors


bheni (approved):
a few comments

PR #2201, opened 2021-09-30
Fix SQL Updates to only include changed columns

VinaiRachakonda:


integration-tests/bats/diff.bats
```diff
@@ -489,3 +489,15 @@ SQL
...
+    dolt sql -q "create table t(pk int primary key, val1 int, val2 int)"
+    dolt sql -q "INSERT INTO t VALUES (1, 1, 1)"
+
+    dolt commit -am "cm1"
+
+    dolt sql -q "UPDATE t SET val1=2 where pk=1"
+    run dolt diff -r sql
+    [ $status -eq 0 ]
+    [[ "$output" = 'UPDATE `t` SET `val1`=2 WHERE (`pk`=1);' ]] || false
+}
```
timsehn, line 503: I would definitely test more than one column and all columns.

timsehn (commented):
Small testing suggestion.

PR #2204, opened 2021-09-30
go/store/nbs: table_reader.go: canReadAhead: Limit how large a read can get from coalescing individual chunk reads.

reltuk:


zachmu (approved):
LGTM!

PR #2205, opened 2021-09-30
Unblock Keyless SQL Diffs and Minimize SQL Diffs

VinaiRachakonda:
Allows for dolt diff -sql to be used for keyless tables and add alter pk support for sql diffs.

### DIFF
--- a/go/libraries/doltcore/diff/sql_diff.go
+++ b/go/libraries/doltcore/diff/sql_diff.go
@@ -16,14 +16,13 @@ package diff
 
 import (
 	"errors"
-	"io"
-
 	"github.com/dolthub/dolt/go/libraries/doltcore/row"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/sqlfmt"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/pipeline"
 	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
 	"github.com/dolthub/dolt/go/store/types"
+	"io"
 )
 
 type SQLDiffSink struct {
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/diff/sql_diff.go
```diff
@@ -95,8 +95,13 @@ func (sds *SQLDiffSink) ProcRowWithProps(r row.Row, props pipeline.ReadableMap)
 			case DiffModifiedOld:
 				return nil
 			case DiffModifiedNew:
-				// TODO: minimize update statement to modified rows
-				stmt, err := sqlfmt.RowAsUpdateStmt(r, sds.tableName, sds.sch)
+				// Update Col Diffs to interface{} to avoid import loop
```
andrew-wm-arthur, line 98: Hacking the type system to get around import loops is generally a bad idea. Think a bit about what the types mean and how you want to modify their meaning. 

Here we want to provide `RowAsUpdateStmt` with more information in order to filter out portions of the SQL update statement that would have no effect. Specifically, we need a set of columns. The keys of `colDiffs` happen to be that set, but that object itself has a different purpose. A better approach here would be to construct a `set.StrSet` from the keys in `colDiffs` and pass that to `RowAsUpdateStmt`. 

Your current approach is pretty close to that as it is, but it's using the wrong types. `colDIffsInter` (does this mean "column diff intersection"?) doesn't need the values of `colDiffs`.

go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
```diff
@@ -120,7 +121,7 @@ func RowAsDeleteStmt(r row.Row, tableName string, tableSch schema.Schema) (strin
 	return b.String(), nil
 }
 
-func RowAsUpdateStmt(r row.Row, tableName string, tableSch schema.Schema) (string, error) {
+func RowAsUpdateStmt(r row.Row, tableName string, tableSch schema.Schema, colDiffs map[string]interface{}) (string, error) {
```
andrew-wm-arthur, line 124:
```suggestion
func RowAsUpdateStmt(r row.Row, tableName string, tableSch schema.Schema, colDiffs set.StrSet) (string, error) {
```

See above comment

VinaiRachakonda: Huh, had no idea this util existed...

andrew-wm-arthur (approved):
LGTM

PR #2211, opened 2021-10-04
refactor: move from io/ioutil to io and os package

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since dolt has upgraded to Go 1.17 (#2114), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.


zachmu (approved):
LGTM assuming tests pass

Thanks for the contribution!

PR #2221, opened 2021-10-05
/go: Add replace statement for forked jsonpath package so we use dolthub's…

andrew-wm-arthur:
… version


bheni (approved):
I assume this is good.

PR #2228, opened 2021-10-06
[no-release-notes] /go/doltcore/libraries/diff: Make a copy of keyless diff.Difference to avoid reference errors

andy-wm-arthur:


tbantle22 (approved):
LGTM! Thanks for fixing

PR #2236, opened 2021-10-07
Prevent panic when iea capacity is reached and undo operations are attemped

Hydrocharged:
In DoltHub we are setting memory capacities for the edit accumulators, and we're running into a bug where an index edit accumulator is at capacity and another accumulator is over capacity and errors, which causes the first accumulator to attempt an undo. This undo fails as it puts it over capacity, causing a panic. This makes it so that undo operations are not affected by capacity (and it should never make a difference as the undo stack is only 4 elements deep at max).


reltuk:
This seems likely to be fine…

Is it the case that we need this panic here? Why can’t an index editor enter a terminal failure case and just return errors? In the case that there was an I/O error in temp external sort file creation, for example, it seems better to fail the operation and let the failure propagate than to panic?

PR #2238, opened 2021-10-07
Use a Tuple Factory for inserts to decrease memory usage.

bheni:
Improvement related to https://github.com/dolthub/dolt/issues/2237

peek resident before: 4.5GB
peek resident after: 2.1GB


abmyii:
Confirmed that this fixes the issue.

PR #2254, opened 2021-10-12
Prevent pull from erroring on config with no email or name:

VinaiRachakonda:


timsehn:
I( think we still need to validate if it's not a fast-forward. Pull can create a merge commit.

VinaiRachakonda:
@timsehn I believe the user is left in a merged state which requires a `dolt commit`. `dolt commit` would then require configs for the merge to actually be finalized

timsehn:
Then add that case to the test?

PR #2314, opened 2021-10-30
Fixed docs markdown generation and fixed a number of related formatting errors

zachmu:


go/cmd/dolt/commands/read_tables.go
```diff
@@ -72,7 +72,7 @@ func (cmd ReadTablesCmd) createArgParser() *argparser.ArgParser {
 		{"commit", "Branch or commit hash representing a point in time to retrieve tables from"},
 		{"table", " Optional tables to retrieve.  If omitted, all tables are retrieved."},
 	}
-	ap.SupportsString(dirParamName, "d", "directory", "directory to create and put retrieved table data.")
+	ap.SupportsString(fileParamName, "d", "directory", "directory to create and put retrieved table data.")
```
max-hoffman, line 75: should the flags change to `--file, -f`?

go/cmd/dolt/commands/reset.go
```diff
@@ -36,20 +35,19 @@ const (
...
-	This means that {{.EmphasisLeft}}dolt reset <tables>{{.EmphasisRight}} is the opposite of {{.EmphasisLeft}}dolt add <tables>{{.EmphasisRight}}.
-
-	After running {{.EmphasisLeft}}dolt reset <tables>{{.EmphasisRight}} to update the staged tables, you can use {{.EmphasisLeft}}dolt checkout{{.EmphasisRight}} to check the
-	contents out of the staged tables to the working tables.
-
-dolt reset .
-	This form resets {{.EmphasisLeft}}all{{.EmphasisRight}} staged tables to their values at HEAD. It is the opposite of {{.EmphasisLeft}}dolt add .{{.EmphasisRight}}`,
-
+	LongDesc: "Sets the state of a table in the staging area to be that table's value at HEAD\n\n" +
+		"{{.EmphasisLeft}}dolt reset <tables>...{{.EmphasisRight}}" +
```
max-hoffman, line 39: kind of awkward formatting, is there a way to do with with a backtick block or something?

go/cmd/dolt/cli/documentation_helper.go
```diff
@@ -199,31 +190,21 @@ var CliFormat = docFormat{"<", ">", "<b>", "</b>"}
 // Special format for the synopsis which is rendered inside raw HTML in markdown
 var SynopsisMarkdownFormat = docFormat{"&lt;", "&gt;", "`", "`"}
 
-func transformSynopsisToHtml(commandStr string, synopsis []string) string {
+func transformSynopsisToMarkdown(commandStr string, synopsis []string) string {
```
max-hoffman, line 193: i don't understand the context of why you're switching from HTML to markdown here, I'm assuming you know what you're doing.  afaik most of this PR just looks like `dolt --help` printing, were we gatsby formatting those before?

max-hoffman: tests would have made it cleared where this is used

max-hoffman (approved):
LGMT, a couple questions about things I didn't understand

go/cmd/dolt/cli/documentation_helper.go
zachmu: It was always markdown, just markdown with embedded HTML

go/cmd/dolt/commands/read_tables.go
zachmu: This is a good catch, turns out this command was reusing the same const and I broke it by renaming it

PR #2319, opened 2021-11-02
First Iteration of Application Server Deploy Script

VinaiRachakonda:


docs/deploy-server.sh
```diff
@@ -0,0 +1,117 @@
...
+  if ! command -v dolt &> /dev/null
+  then
+    sudo bash -c 'curl -L https://github.com/dolthub/dolt/releases/latest/download/install.sh | bash'
+  fi
+}
+
+setup_configs() {
+   # Set up the dolt user along with core dolt configurations
+  echo "Setting up Configurations..."
+  useradd -r -m -d /var/lib/doltdb dolt
```
zachmu, line 25: Want to check if this user exists first, err if so

docs/deploy-server.sh
```diff
@@ -0,0 +1,117 @@
...
+
+  sudo -u dolt dolt config --global --add user.email $email
+  sudo -u dolt dolt config --global --add user.name $username
+}
+
+# Database creation
+database_configuration() {
+  echo "Setting up the dolt database..."
+
+  read -e -p "Input the name of your database: " -i "doltdb" db_name
```
zachmu, line 39: Too many doltdb references, just call this mydb or something

zachmu (approved):
LGTM

docs/deploy-server.sh
VinaiRachakonda: Should we make this configurable? Might be annoying if someone runs the script multiple times/.

PR #2326, opened 2021-11-02
CREATE DATABASE now creates a new dolt repository in the working directory

zachmu:
This is a first implementation of durable CREATE DATABASE support. It always creates a new database in the same directory that the server was run from, as a separate repository. It works in both server and command line mode.

Future extensions:
* Configurable DOLT_DATA_DIR variable
* Create database in the same repository (same commit graph)


timsehn (approved):
This is very cool. 

I think next up would be some sort DOLT_ROOT environment variable but this is good MVP.

timsehn:
As discussed I think the makes the `--multi-db-dir` optionwhen running `dolt sql-server` the default.

PR #2330, opened 2021-11-02
/go/store/val: New Storage Format for Tuples

andrew-wm-arthur:


go/store/val/offsets.go
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package val
+
+import "encoding/binary"
+
+type Offsets []byte
```
bheni, line 19: Do any of the Offsets methods need to be exported?  Should they all be internal?

bheni (commented):
Great start

go/store/val/offsets.go
andrew-wm-arthur: In the scope of this PR they're all internal, but I reused this type for the offset array in prolly tree nodes

PR #2340, opened 2021-11-03
dolt_fetch error when working directory != server directory

max-hoffman:


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -155,6 +155,7 @@ type DatabaseSessionState struct {
 	readReplica          *env.Remote
 	TempTableRoot        *doltdb.RootValue
 	TempTableEditSession *editor.TableEditSession
+	tmpdir               string
```
andrew-wm-arthur: Do we need this extra member? We have access to a `RepoStateReader` through `dbData`

andrew-wm-arthur (commented):
LGTM, one comment

go/libraries/doltcore/sqle/dsess/session.go
max-hoffman: yeah that's what i thought originally, but the session state adaptor is the RsR so you get an infinite recursion if you `dbData.RsR.TempTableFilesDir()`

PR #2342, opened 2021-11-04
go/libraries/doltcore/table/editor: Thread some TupleFactories to places that allocate tuples.

reltuk:
Makes types.TupleFactory thread-safe. Allocates a new TupleFactory when we construct a SessionTableEditor, and threads that TupleFactory through to individual table editors that can use it to do their editing. Makes that TupleFactory available through an accessor that sqlTableEditor uses for a few tuple manipulations.


bheni (approved):
lgtm

PR #2346, opened 2021-11-04
DOLT_REPLICATE_HEADs_STRATEGY option, fetch all branches from remote before every read

max-hoffman:


go/libraries/doltcore/dtestutils/testcommands/multienv.go
```diff
@@ -146,6 +146,22 @@ func (mr *MultiRepoTestSetup) NewRemote(remoteName string) {
 	mr.Remotes[remoteName] = rem
 }
 
+func (mr *MultiRepoTestSetup) NewBranch(dbName, branchName string) {
+	dEnv := mr.MrEnv[dbName]
```
zachmu, line 150: You need to merge main

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -100,6 +100,14 @@ func init() {
 			Type:              sql.NewSystemBoolType(ForceTransactionCommit),
 			Default:           int8(0),
 		},
+		{
+			Name:              doltdb.ReplicateHeadsStrategy,
+			Scope:             sql.SystemVariableScope_Session,
+			Dynamic:           true,
+			SetVarHintApplies: false,
+			Type:              sql.NewSystemStringType(doltdb.ReplicateHeadsStrategy),
+			Default:           "one",
```
zachmu, line 109: Default should be many, only replicating one head is an advanced setup and probably not what you want

And the configuration should be "all" or "the ones I name specifically"

zachmu: I think you want two system vars:

`replicate_all_heads` (boolean)
`replicate_heads` (comma separated list of heads)

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -99,12 +99,31 @@ func (rrd ReadReplicaDatabase) StartTransaction(ctx *sql.Context, tCharacteristi
...
 
 func (rrd ReadReplicaDatabase) pullFromReplica(ctx *sql.Context) error {
+	if _, val, ok := sql.SystemVariables.GetGlobal(doltdb.ReplicateHeadsStrategy); ok {
+		switch val {
+		case doltdb.ReplicateHeads_MANY:
+			err := fetchBranches(ctx, rrd)
+			if err != nil {
+				return err
+			}
+			return fetchRef(ctx, rrd, rrd.headRef)
```
zachmu, line 109: Every case ends with this statement, should pull it outside the switch

zachmu (commented):
Some design notes, not a complete review

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -217,14 +218,20 @@ func (p DoltDatabaseProvider) databaseForRevision(ctx context.Context, revDB str
...
+	srcDb, ok := candidate.(SqlDatabase)
 	if !ok {
 		return nil, dsess.InitialDbState{}, false, nil
 	}
 
-	if isBranch(ctx, srcDb.ddb, revSpec) {
-		// if the requested revision is a br we can
-		// write to it, otherwise make read-only
+	if replicaDb, ok := srcDb.(ReadReplicaDatabase); ok {
+		err := switchAndFetchReplicaHead(ctx, revSpec, replicaDb)
```
zachmu, line 228: I think this is fine for now, but it's not going to work longer term.

The issue is that this method is called primarily during analysis, *before* any IO should in principle happen. We want the fetch and pull to happen when the query begins to execute, not here.

Would add a TODO for now

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -233,7 +240,11 @@ func (p DoltDatabaseProvider) databaseForRevision(ctx context.Context, revDB str
 	}
 
 	if doltdb.IsValidCommitHash(revSpec) {
-		db, init, err := dbRevisionForCommit(ctx, srcDb, revSpec)
+		srcDb, ok = srcDb.(Database)
+		if !ok {
+			return nil, dsess.InitialDbState{}, false, ErrCannotCreateReplicaRevisionDbForCommit
```
zachmu, line 245: No reason to error here that I can see. The replica DB may have the commit hash named, or it may not. Shouldn't err out here by default.

Add a TODO to do better error management.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -262,6 +273,49 @@ func (p DoltDatabaseProvider) Function(name string) (sql.Function, error) {
...
+	branchRef := ref.NewBranchRef(branch)
+
+	// check whether remote branch exists before creating local tracking branch
+	_, err := actions.FetchRemoteBranch(ctx, destDb.tmpDir, destDb.remote, destDb.srcDB, destDb.DbData().Ddb, branchRef, nil, actions.NoopRunProgFuncs, actions.NoopStopProgFuncs)
+	if err != nil {
+		return err
+	}
+
+	// TODO is localBranches var really necessary? we're not going to call this that often
+	if _, ok := destDb.localBranches[branch]; !ok {
```
zachmu, line 291: I don't think you want to track this in memory at all. Ask the DB if it the branch exists or not.

zachmu: Especially since you're attempting to create the branch below as a consequence of this answer.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -262,6 +273,49 @@ func (p DoltDatabaseProvider) Function(name string) (sql.Function, error) {
...
+	}
+
+	// update ReadReplicaRemote with new HEAD and remote tracking branch
+	destDb, err = destDb.SetHeadRef(branchRef)
+	if err != nil {
+		return err
+	}
+
+	// we fetched, now formalize working set update if branch is configured to replicate
+	err = destDb.PullFromReplica(ctx)
```
zachmu, line 311: Shouldn't this be PullFromRemote?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -67,88 +72,175 @@ func NewReadReplicaDatabase(ctx context.Context, db Database, remoteName string,
...
+			return err
+		}
+	default:
+		return fmt.Errorf("%w: dolt_replicate_heads not set", ErrInvalidReplicateHeadSetting)
+	}
+	return nil
+}
+
+func (rrd ReadReplicaDatabase) SetHeadRef(head ref.DoltRef) (ReadReplicaDatabase, error) {
+	//rtRef, err := remoteTrackingRef(rrd.rsr, head, rrd.remote.Name)
```
zachmu, line 158: Delete?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -67,88 +72,175 @@ func NewReadReplicaDatabase(ctx context.Context, db Database, remoteName string,
...
+		if err != nil {
+			return err
+		}
+	case allHeads == int8(1):
+		err := rrd.srcDB.Rebase(ctx)
+		if err != nil {
+			return err
+		}
+
+		refs, err := rrd.srcDB.GetBranches(ctx)
```
zachmu, line 137: This doesn't work to get a new branch that was pushed, right? Unless you did a fetch to include branches on the remote?

integration-tests/bats/replication.bats
```diff
@@ -100,9 +101,161 @@ teardown() {
...
+    cd repo2
+    dolt branch new_feature
+    dolt push origin new_feature
+
+    cd ../repo1
+    dolt config --local --add sqlserver.global.dolt_replicate_all_heads 1
+    dolt config --local --add sqlserver.global.dolt_read_replica_remote remote1
+    dolt config --list
+    dolt sql -q "show tables"
+    dolt checkout new_feature
```
zachmu, line 222: For these tests, it would be good see data changes on the new branch as well

zachmu (approved):
Looks pretty good, nice work

Mostly small comments and requests for more rigorous tests

go/libraries/doltcore/sqle/database_provider.go
max-hoffman: kk. i used the `USE db/branch` syntax to test, wasn't sure whether or not that's wrapped in a transaction

max-hoffman: it was just returning nil before, l'll set it back to that

PR #2347, opened 2021-11-04
fix fetch newline issue

jennifersp:
dolt fetch: prints many empty newline to terminal issue - fixed

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -1280,7 +1280,7 @@ func (ddb *DoltDB) PullChunks(ctx context.Context, tempDir string, srcDB *DoltDB
 		puller, err := datas.NewPuller(ctx, tempDir, defaultChunksPerTF, srcDB.db, ddb.db, stRef.TargetHash(), pullerEventCh)
 
 		if err == datas.ErrDBUpToDate {
-			return nil
+			return datas.ErrDBUpToDate
 		} else if err != nil {
 			return err
 		}
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -1278,10 +1278,7 @@ func (ddb *DoltDB) PushChunksForRefHash(ctx context.Context, tempDir string, src
 func (ddb *DoltDB) PullChunks(ctx context.Context, tempDir string, srcDB *DoltDB, stRef types.Ref, progChan chan datas.PullProgress, pullerEventCh chan datas.PullerEvent) error {
 	if datas.CanUsePuller(srcDB.db) && datas.CanUsePuller(ddb.db) {
 		puller, err := datas.NewPuller(ctx, tempDir, defaultChunksPerTF, srcDB.db, ddb.db, stRef.TargetHash(), pullerEventCh)
-
-		if err == datas.ErrDBUpToDate {
-			return datas.ErrDBUpToDate
-		} else if err != nil {
+		if err != nil {
 			return err
 		}
 
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -260,7 +260,7 @@ func FetchCommit(ctx context.Context, tempTablesDir string, srcDB, destDB *doltd
 	return destDB.PullChunks(ctx, tempTablesDir, srcDB, stRef, progChan, pullerEventCh)
 }
 
-// FetchCommit takes a fetches a commit tag and all underlying data from a remote source database to the local destination database.
+// FetchTag takes a fetches a commit tag and all underlying data from a remote source database to the local destination database.
 func FetchTag(ctx context.Context, tempTableDir string, srcDB, destDB *doltdb.DoltDB, srcDBTag *doltdb.Tag, progChan chan datas.PullProgress, pullerEventCh chan datas.PullerEvent) error {
 	stRef, err := srcDBTag.GetStRef()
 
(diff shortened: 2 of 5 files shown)


timsehn:
Can we write a bats test so it doesn't come back? Is it easy to construct a fetch that would do this?

jennifersp:
I'll check bats tests with 'dolt fetch' use.

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -352,6 +354,11 @@ func FetchRemoteBranch(ctx context.Context, tempTablesDir string, rem env.Remote
 	wg, progChan, pullerEventCh := progStarter(newCtx)
 	err = FetchCommit(ctx, tempTablesDir, srcDB, destDB, srcDBCommit, progChan, pullerEventCh)
 	progStopper(cancelFunc, wg, progChan, pullerEventCh)
+	if err == nil {
+		cli.Println()
```
zachmu, line 358: I don't think we want any output here at all, right?

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1280,7 +1280,7 @@ func (ddb *DoltDB) PullChunks(ctx context.Context, tempDir string, srcDB *DoltDB
 		puller, err := datas.NewPuller(ctx, tempDir, defaultChunksPerTF, srcDB.db, ddb.db, stRef.TargetHash(), pullerEventCh)
 
 		if err == datas.ErrDBUpToDate {
-			return nil
+			return datas.ErrDBUpToDate
```
zachmu, line 1283: Changing this will change the behavior at its call sites. I would expect one or more tests to die from this change.

Also, this condition should just collapse to:

```go
if err != nil {
    return err
}
```

zachmu (commented):
Would like to see some tests for this, it's not obvious to me that this fix or works, or that it doesn't break something else

go/libraries/doltcore/env/actions/remotes.go
jennifersp: it is for printing newline after printing output from changed branch

integration-tests/bats/remotes.bats
```diff
@@ -429,6 +429,82 @@ SQL
     [[ "$output" =~ "remotes/test-remote/poop" ]] || false
 }
 
+@test "remotes: dolt fetch check for printing many newlines" {
```
zachmu, line 432: call this "remotes: fetch output"

integration-tests/bats/remotes.bats
```diff
@@ -429,6 +429,82 @@ SQL
...
+    dolt sql -q 'insert into test (id) values (10), (11), (12);'
+    dolt add .
+    dolt commit -m 'add some values to  branch 4.'
+    dolt push --set-upstream origin branch4
+
+    cd ..
+    cd test-repo
+    run dolt fetch
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 4 ]
```
zachmu, line 505: Does this test what you want it to?

The bug is dolt fetch producing many blank lines when it should not. I think you need another test case for that behavior (where the number of lines of output is one, and it's just a newline). Here's you're testing that there are 4 lines of output. If those are legitimate, you should make assertions about their content.

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1278,10 +1278,7 @@ func (ddb *DoltDB) PushChunksForRefHash(ctx context.Context, tempDir string, src
 func (ddb *DoltDB) PullChunks(ctx context.Context, tempDir string, srcDB *DoltDB, stRef types.Ref, progChan chan datas.PullProgress, pullerEventCh chan datas.PullerEvent) error {
 	if datas.CanUsePuller(srcDB.db) && datas.CanUsePuller(ddb.db) {
 		puller, err := datas.NewPuller(ctx, tempDir, defaultChunksPerTF, srcDB.db, ddb.db, stRef.TargetHash(), pullerEventCh)
-
-		if err == datas.ErrDBUpToDate {
-			return nil
-		} else if err != nil {
+		if err != nil {
```
zachmu, line 1281: I'm worried this introduces problems. I haven't checked, but you need to examine every place that calls this method and make sure they have equivalent error checking code to what you removed, somewhere in the call stack. Otherwise this error will interfere with other operations.

zachmu: Looks like FetchTag has this problem.

integration-tests/bats/remotes.bats
jennifersp: added uptodate fetch output check

zachmu (approved):
LGTM!

integration-tests/bats/remotes.bats
```diff
@@ -429,6 +429,138 @@ SQL
...
+    dolt clone http://localhost:50051/test-org/test-repo
+    cd test-repo
+    dolt status
+    run dolt status
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "On branch main" ]] || false
+    [[ "$output" =~ "nothing to commit, working tree clean" ]] || false
+
+    cd ..
+    cd test-repo
```
timsehn, line 558: I think this can be moved up into the other test at the end. Basically a second fetch (ie. everything is updated) should produce no output. This saves a bunch of code.

timsehn (commented):
Bats comment

integration-tests/bats/remotes.bats
```diff
@@ -429,6 +429,93 @@ SQL
...
+    run dolt fetch
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 4 ]
+    [ "${lines[0]}" != "" ]
+    [ "${lines[1]}" != "" ]
+    [ "${lines[2]}" != "" ]
+    [ "${lines[3]}" != "" ]
+}
```
timsehn, line 517: I think what you want to add right here is
```
# Make sure fetching return nothing when there is nothing to fetch
run dolt fetch
[ "$status" -eq 0 ]
[ $output = "" ]
```

timsehn (commented):
This is what I meant :-)

integration-tests/bats/remotes.bats
jennifersp: oh i see. got it. thank you!

timsehn:
Ship it :-)

PR #2352, opened 2021-11-05
/go/store/prolly: New Storage Format for Maps

andrew-wm-arthur:


andrew-wm-arthur:
Edges cases dealing with limits for tuple size, key size, and chunk size are minimally handled here. Specifically, there is a constraint that internal nodes have at least two meta pairs, but are also within the maximum node/chunk size. This case could be solved be placing restrictions on key size. Some cursory research shows that similar limits exist in MySQL/InnoDB:
 https://dev.mysql.com/doc/refman/8.0/en/innodb-limits.html

go/store/prolly/node.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (nd Node) level() int {
+	sl := nd[nd.size()-treeLevelSize:]
+	return int(val.ReadUint8(sl))
+}
+
+func (nd Node) nodeCount() int {
+	stop := nd.size() - treeLevelSize
+	start := stop - nodeCountSize
+	return int(val.ReadUint16(nd[start:stop]))
```
andrew-wm-arthur, line 180: the need for (de)serialization here and below motivated me to make the codec functions public. I considered even splitting them out to their own package.

go/store/prolly/tree_chunker.go
```diff
@@ -0,0 +1,444 @@
...
+	if overflow && degenerate {
+		// Constraints (2) and (3) are in conflict
+		panic("impossible node")
+	}
+
+	if overflow {
+		// Enforce constraints (1) and (2):
+		//  |key| and |value| won't fit in this chunk,
+		//  force a boundary here and pass them as leftovers
+		err := sc.handleChunkBoundary(ctx, key, value)
```
andrew-wm-arthur, line 229: In the "overflow" case, we're adding `key` and `value` as "leftovers", and adding them again on line 235.
Needs a fix

PR #2369, opened 2021-11-10
Async push for replication

max-hoffman:
Async hook overview:
- Calling the commit hook (Execute) pushes a (dataset, database) tuple into a channel
- go routine 1 is responsible for reading from this channel, and setting the dataset id -> pull args in a shared map. Multiple commits to the same dataset override old values in this map.
- goroutine 2 has two phases that are run iteratively. Phase 1 copies the current state of the commit map into a local value. Phase 2 iterates through daatsets in the local commit map, pushing the most recent dataset commit if its hash is not the deduplication map's value. The deduplication map is private to gouroutine 2, and saved the pushed commit hash for each dataset.

lifecycle management:
- new engine interface maintains `BackgroundThreads`, finalizing when we are done with the engine

Initial perf comparison:
- PushOnWrite: 7.9 sec / op
- AsyncPushOnWrite: 6.5 sec / op
- diff ~ 20% speedup

The perf comparison depends on many factors (the transaction throughput, commit size, network latency, etc), but this isolated test supports the idea that amortizing commits is better than individually pushing every commit. In practice, the time/memory overhead of coordinating the amortization with threads should always be smaller that the round trip of pushing every individual dataset.

edge cases:
- commits to a dataset move backwards in time or delete values. The older commits contains chunks that fail to push (? but push synchronizes ancestors commits? so this should be OK unless we reset hard?)
- every commit creates a unique dataset id, and the amortization map blows up


go/cmd/dolt/commands/sql.go
```diff
@@ -555,8 +555,11 @@ func getPushOnWriteHook(ctx context.Context, dEnv *env.DoltEnv) (*doltdb.PushOnW
 		return nil, err
 	}
 
-	pushHook := doltdb.NewPushOnWriteHook(ddb, dEnv.TempTableFilesDir())
-	return pushHook, nil
+	if _, val, ok := sql.SystemVariables.GetGlobal(dsqle.AsyncReplicationKey); ok && val == int8(1) {
```
bheni, line 558: You shouldn't need the int8 cast for comparison with a constant.  The compiler should handle that.  Probably better to use a named constant

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -93,6 +94,97 @@ func pushDataset(ctx context.Context, destDB, srcDB datas.Database, tempTableDir
...
+var _ datas.CommitHook = (*AsyncPushOnWriteHook)(nil)
+
+// NewAsyncPushOnWriteHook creates a AsyncReplicateHook
+func NewAsyncPushOnWriteHook(ctx context.Context, destDB *DoltDB, tmpDir string) *AsyncPushOnWriteHook {
+	ch := make(chan PushArg, asyncPushBufferSize)
+
+	var newHeads = make(map[string]PushArg, asyncPushBufferSize)
+	go func() error {
+		defer close(ch)
+		for {
```
bheni, line 119: This loop will only exit when ch is closed.  It won't exit if the context is cancelled, and this routine itself is responsible for closing the channel on exit.  So I don't think this routine will ever exit.

In most cases the writer should be responsible for closing the channel.

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -93,6 +94,97 @@ func pushDataset(ctx context.Context, destDB, srcDB datas.Database, tempTableDir
...
+				if len(newHeads) == 0 {
+					continue
+				}
+				// TODO do i need to explicitly copy newHeads?
+				for id, newCm := range newHeads {
+					if latest, ok := latestHeads[id]; !ok || latest != newCm.hash {
+						err := pushDataset(ctx, destDB.db, newCm.db, tmpDir, newCm.ds)
+						if err != nil {
+							// TODO what to do with errors?
+							return err
```
bheni, line 144: should probably do something with them.

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -93,6 +94,97 @@ func pushDataset(ctx context.Context, destDB, srcDB datas.Database, tempTableDir
...
+		for {
+			p, ok := <-ch
+			if !ok {
+				return ctx.Err()
+			}
+			newHeads[p.ds.ID()] = p
+		}
+	}()
+
+	go func() error {
```
bheni, line 128: Not sure when this routine is gonna exit.  It is dependent on ctx.Done to exit.  Is that what we want?  What context is this using?

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -93,6 +94,97 @@ func pushDataset(ctx context.Context, destDB, srcDB datas.Database, tempTableDir
...
+		}
+	}()
+
+	return &AsyncPushOnWriteHook{ch: ch}
+}
+
+// Execute implements datas.CommitHook, replicates head updates to the destDb field
+func (ah *AsyncPushOnWriteHook) Execute(ctx context.Context, ds datas.Dataset, db datas.Database) error {
+	rf, ok, err := ds.MaybeHeadRef()
+	if !ok {
```
bheni, line 159: should handle errors before other return parameters as the other return parameters are often times not valid when an error is returned.  Doesn't really matter in this case, but still a better practice.

timsehn:
What's up here? @max-hoffman who you waiting on?

max-hoffman:
@timsehn brian's planning on rereviewing

timsehn:
@bheni planning?

timsehn:
How we looking here?

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -93,6 +95,61 @@ func pushDataset(ctx context.Context, destDB, srcDB datas.Database, tempTableDir
...
+	db   datas.Database
+	hash hash.Hash
+}
+
+type AsyncPushOnWriteHook struct {
+	out io.Writer
+	ch  chan PushArg
+}
+
+const asyncPushBufferSize = 20
```
bheni, line 109: I think you may want to bump this up much higher.  At some point Dolt will be running with many simultaneous connections each of which may be writing.  Don't want to block when writing to the channel in this case.

bheni: 2048?

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -127,3 +184,69 @@ func (lh *LogHook) SetLogger(ctx context.Context, wr io.Writer) error {
...
 }
+
+func RunAsyncReplicationThreads(ctx context.Context, wg *sync.WaitGroup, ch chan PushArg, destDB *DoltDB, tmpDir string, logger io.Writer) {
+	mu := &sync.Mutex{}
+	var newHeads = make(map[string]PushArg, asyncPushBufferSize)
+
+	// stop first go routine before second
+	newCtx, stop := context.WithCancel(context.Background())
+	wg.Add(1)
+	go func() error {
```
bheni, line 195: not actually doing anything with the error.  may want to at least make it clear that that is intentional

`_ = go func() error {...}()`

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -127,3 +184,69 @@ func (lh *LogHook) SetLogger(ctx context.Context, wr io.Writer) error {
...
+				mu.Unlock()
+			case <-ctx.Done():
+				stop()
+				return ctx.Err()
+			}
+		}
+	}()
+
+	wg.Add(1)
+	go func() error {
```
bheni, line 214: same as above.  Either don't return the error, do something with it, or assign it to _.

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -127,3 +184,69 @@ func (lh *LogHook) SetLogger(ctx context.Context, wr io.Writer) error {
...
+			}
+		}
+
+		for {
+			select {
+			case <-newCtx.Done():
+				flush()
+				return newCtx.Err()
+			default:
+				flush()
```
bheni, line 247: probably want to sleep for at least a couple hundred millis

go/libraries/doltcore/doltdb/commit_hooks.go
```diff
@@ -127,3 +184,69 @@ func (lh *LogHook) SetLogger(ctx context.Context, wr io.Writer) error {
...
+		flush := func() {
+			mu.Lock()
+			if len(newHeads) == 0 {
+				mu.Unlock()
+				return
+			}
+			for k, v := range newHeads {
+				newHeadsCopy[k] = v
+			}
+			mu.Unlock()
```
bheni, line 228: I think I would move this logic into a function that can use a defer block to unlock. something like:

```
getHeadsCopy() map[string]PushArg {
	mu.Lock()
	defer mu.Unlock()
	if len(newHeads) == 0 {
		return nil
	}
		
	var newHeadsCopy = make(map[string]PushArg, asyncPushBufferSize)
	for k, v := range newHeads {
		newHeadsCopy[k] = v
	}
	return newHeadsCopy
}
```

bheni: This also clears out the newHeadsCopy each loop... which seems like what you would want.

bheni (commented):
I think this mostly looks good... but I'm not a fan of the waitgroup changes. There has to be a better way to accomplish this part.  We should chat about it tomorrow.

bheni (approved):
lgtm

PR #2371, opened 2021-11-10
dolt push - no newline after result printed bug

jennifersp:
added newline printing statements for dolt push command

### DIFF
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -468,7 +468,7 @@ func SyncRoots(ctx context.Context, srcDb, destDb *doltdb.DoltDB, tempTableDir s
 
 	newCtx, cancelFunc := context.WithCancel(ctx)
 	wg, progChan, pullerEventCh := progStarter(newCtx)
-	defer func(){
+	defer func() {
 		progStopper(cancelFunc, wg, progChan, pullerEventCh)
 		if err == nil {
 			cli.Println()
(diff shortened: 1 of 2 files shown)


timsehn (approved):
LGTM

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -191,6 +191,9 @@ func PushToRemoteBranch(ctx context.Context, rsr env.RepoStateReader, tempTableD
 	wg, progChan, pullerEventCh := progStarter(newCtx)
 	err = Push(ctx, tempTableDir, mode, destRef.(ref.BranchRef), remoteRef.(ref.RemoteRef), localDB, remoteDB, cm, progChan, pullerEventCh)
 	progStopper(cancelFunc, wg, progChan, pullerEventCh)
+	if err == nil {
```
zachmu, line 194: Instead of separate logic, you should handle this as a case of the switch below (and remove the `err != nil` check above the switch):

```go
switch err {
    case nil:
        cli.PrintLn()
        return nil
}
```

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -215,6 +218,9 @@ func pushTagToRemote(ctx context.Context, tempTableDir string, srcRef, destRef r
 	wg, progChan, pullerEventCh := progStarter(newCtx)
 	err = PushTag(ctx, tempTableDir, destRef.(ref.TagRef), localDB, remoteDB, tg, progChan, pullerEventCh)
 	progStopper(cancelFunc, wg, progChan, pullerEventCh)
+	if err == nil {
```
zachmu, line 221: Reverse the logic here

```go
if err != nil {
    return err
}
cli.Println()

return nil

zachmu (approved):
LGTM, couple suggestions

Also you should write a more descriptive PR title

PR #2372, opened 2021-11-11
added for M1 machines

bheni:


timsehn:
I defer to Aaron.

timsehn (approved):

timsehn:
@reltuk. Should we do this?

Don't we also have to change the dolt install script?

timsehn:
This now seems credible to me.

PR #2379, opened 2021-11-12
Command to generate a zsh autocomplete file

zachmu:


zachmu:
I would back out your last commit. Not sure what your local version of format did, but it disagrees with what our CI expects.

zachmu:
I can format the code once it's otherwise good to go

PR #2382, opened 2021-11-12
Added support for indexes to use mixed operators, bug fix for returning duplicate rows

Hydrocharged:
Tests covering mixing operators will come in a separate PR.


go/libraries/doltcore/sqle/dolt_index.go
```diff
@@ -70,63 +68,84 @@ func (di *doltIndex) NewLookup(ctx *sql.Context, ranges ...sql.Range) (sql.Index
...
+			Reverse:   false,
+			Check:     rangeCheck,
+		})
 	}
 
 	return &doltIndexLookup{
-		idx:       idx,
-		ranges:    lookupRanges,
+		idx:       di,
+		ranges:    readRanges,
```
zachmu, line 148: Possible for readRanges to be empty at this point? Does that just mean an empty lookup, working as intended?

go/libraries/doltcore/sqle/dolt_index_test.go
```diff
@@ -910,7 +951,13 @@ func TestDoltIndexBetween(t *testing.T) {
...
 			[]interface{}{3, 4},
 			[]interface{}{4, 4},
-			[]sql.Row{{2, 2, 4, 3}, {1, 2, 3, 4}, {2, 1, 4, 4}},
+			[]sql.Row{{2, 1, 4, 4}, {2, 2, 4, 3}},
+		},
+		{
+			"twopk:idx_v2v1",
+			[]interface{}{3, 3},
+			[]interface{}{4, 4},
+			[]sql.Row{{1, 1, 3, 3}, {1, 2, 3, 4}, {2, 1, 4, 4}, {2, 2, 4, 3}},
```
zachmu, line 960: Why aren't there new tests here for mixed operators? All of these functions do this with one of the operators:

```go
testDoltIndex(t, test.keys, test.expectedRows, index, indexComp_Eq)
```

Which then does this:

```go
	for i, key := range keys {
		switch cmp {
		case indexComp_Eq:
			builder = builder.Equals(ctx, exprs[i], key)
```

So all these tests necessarily use the same operator for every key column. Seems like a pretty big gap in testing for an increase in functionality this big.

zachmu (commented):
This looks pretty clean, good refactorings.

My main comment is I don't see where the tests are. They aren't in dolt_index_test.go. Am I missing something?

go/libraries/doltcore/sqle/dolt_index.go
Hydrocharged: Correct, nil ranges will be an empty lookup returning no rows as intended.

go/libraries/doltcore/sqle/dolt_index_test.go
Hydrocharged: Someone didn't read my comment lol

PR #2402, opened 2021-11-16
Refactor the cli's SQL Engine to its own package

VinaiRachakonda:


go/cmd/dolt/commands/filter-branch.go
```diff
@@ -190,8 +191,7 @@ func processFilterQuery(ctx context.Context, dEnv *env.DoltEnv, cm *doltdb.Commi
 			return nil, fmt.Errorf("error parsing DDL: %v", err.Error())
 		}
 		// ddl returns a nil itr
```
VinaiRachakonda, line 193: Get rid of this comment

andrew-wm-arthur (approved):
In general this looks good, but this package is in the wrong place. It should live inside `go/cmd/dolt/commands/...`

Also, I think `engine` would be a more appropriate name as this provides a SQL engine for both the CLI and `dolt sql-server`

PR #2411, opened 2021-11-17
Drop database support

zachmu:
Also improved support for creating and dropping dbs when using the --multi-db-dir flag


integration-tests/bats/sql-create-database.bats
```diff
@@ -8,94 +8,185 @@ teardown() {
...
+    dolt sql -q "drop database mydb"
+
+    [ ! -d mydb ]
 }
 
-@test "sql-create-database: create database that already exists throws an error" {
-    run dolt sql << SQL
-CREATE DATABASE mydb;
-CREATE DATABASE mydb;
+@test "sql-create-database: with multi-db-dir" {
```
max-hoffman, line 55: this feels like 4 tests

integration-tests/bats/sql-server.bats
```diff
@@ -1114,11 +1113,60 @@ while True:
...
+
+    mkdir no_dolt && cd no_dolt
+    mkdir db_dir
+    start_sql_server_with_args --host 0.0.0.0 --user dolt --multi-db-dir=db_dir
+
+    server_query "" 1 "create database test1"
+    server_query "" 1 "show databases" "Database\ninformation_schema\ntest1"
+    server_query "test1" 1 "create table a(x int)"
+    server_query "test1" 1 "insert into a values (1), (2)"
+    # not bothering to check the results of the commit here
```
max-hoffman, line 1144: is this preferred or multi_query, which doesn't expect a result?

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -99,14 +105,37 @@ func Serve(ctx context.Context, version string, serverConfig ServerConfig, serve
 	dbNamesAndPaths := serverConfig.DatabaseNamesAndPaths()
 
 	if len(dbNamesAndPaths) == 0 {
-		var err error
-		mrEnv, err = env.DoltEnvAsMultiEnv(ctx, dEnv)
-		if err != nil {
-			return err, nil
+		if len(serverConfig.DataDir()) > 0 && serverConfig.DataDir() != "." {
```
max-hoffman, line 108: i'm not sure i understand why the logic is different between the two outer cases. A noop `fs.WithWorkingDir` that just sets it to "." seems fine? If we call `MultiEnvForDirectory` when we could have called `DoltEnvAsMultiEnv` that also seems ok?

max-hoffman (commented):
looks good overall but small comments; probably leave the sub dEnvs until we want to do another config refactor pass?

integration-tests/bats/sql-server.bats
zachmu: I prefer it

go/cmd/dolt/commands/sqlserver/server.go
zachmu: DoltEnvAsMultiEnv includes the dolt DB in the current working directory, as well as any in subdirectories of it. It's a subtle difference, but specifying data_dir = "." should exclude the dolt DB in the current working directory.

DoltEnvAsMultiEnv is really only there to support quick start, it makes some compromises in the data model.

zachmu (commented):
Thanks for feedback

Issue #2390, opened 2021-11-15
Table with new columns not detected as modified


knutwannheden:
Given the following *old* and *new* table definitions Dolt isn't able to work out that it is the same table. Instead it wants to delete and recreate the table (as shown by `dolt status`):

Old definition:
```sql
CREATE TABLE `organisation` (
  `id` int NOT NULL AUTO_INCREMENT COMMENT 'Technischer Schlüssel der Lobbyorganisation',
  `name_de` varchar(160) collate utf8mb4_unicode_ci NOT NULL COMMENT 'Name der Organisation. Sollte nur juristischem Namen entsprechen, ohne Zusätze, wie Adresse.',
  `name_fr` varchar(160) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Französischer Name',
  `name_it` varchar(160) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Italienischer Name',
  `uid` varchar(15) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'UID des Handelsregisters; Schweizweit eindeutige ID (http://www.bfs.admin.ch/bfs/portal/de/index/themen/00/05/blank/03/02.html); Format: CHE-999.999.999',
  `in_handelsregister` tinyint NOT NULL DEFAULT "0" COMMENT 'Ist die Organisation im Handelsregister (Zefix) eingetragen?',
  `inaktiv` tinyint DEFAULT "0" COMMENT 'Gibt es die Organisation noch?',
  `ort` varchar(100) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Ort der Organisation',
  `abkuerzung_de` varchar(30) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Abkürzung der Organisation, kann in der Anzeige dem Namen nachgestellt werden, z.B. Schweizer Kaderorganisation (SKO)',
  `alias_namen_de` varchar(255) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Strichpunkt-getrennte Aufzählung von alternative Namen für die Organisation; bei der Suche werden für ein einfacheres Finden auch in den Alias-Namen gesucht.',
  `abkuerzung_fr` varchar(30) collate utf8mb4_unicode_ci DEFAULT NULL COMMENT 'Französische Abkürzung der Organisation',
[...]

PR #2423, opened 2021-11-19
More intelligent drop / create for tables

zachmu:
This addresses #2390 and other issues.

Previously we were only reusing dropped tables if they had exactly the same schema upon recreation. This PR loosens that constraint.


integration-tests/bats/drop-create.bats
```diff
@@ -0,0 +1,258 @@
...
+
+    dolt sql -q "drop table test"
+
+    run dolt sql -q "show tables"
+    [ "$status" -eq 0 ]
+    [[ ! "$output" =~ "test" ]] || false
+
+    dolt sql --disable-batch <<SQL
+create table test(a bigint primary key, b tinyint null, c int null);
+insert into test(a,b) values (1,1), (2,2);
```
timsehn, line 238: Maybe a few other alters like adding defaults, constraints or triggers.

timsehn (commented):
A bit more test coverage

integration-tests/bats/drop-create.bats
zachmu: Take another look

andrew-wm-arthur (approved):
LGTM

PR #2431, opened 2021-11-22
Fix issue with not clearing conflicts correctly

JCOR11599:
Added change in `conflict_source.go` to just return `ErrNoConflicts` when there are no conflicts rather than printing an empty table.

Added extra check in `resolve.go` to call `ClearConflicts()` when there are zero rows in conflict after resolving table.

Removed skip statement from bats test that checked for these behaviors.

Edited case-sensitivity test for selecting `dolt_conflicts_test`, as that now returns a `ErrNoConflicts`.

### DIFF
--- a/go/libraries/doltcore/doltdb/table.go
+++ b/go/libraries/doltcore/doltdb/table.go
@@ -155,6 +155,7 @@ func (t *Table) SetConflicts(ctx context.Context, schemas Conflict, conflictData
 	return &Table{t.vrw, updatedSt}, nil
 }
 
+// GetConflicts returns a map built from ValueReadWriter when there are no
 func (t *Table) GetConflicts(ctx context.Context) (Conflict, types.Map, error) {
 	schemasVal, ok, err := t.tableStruct.MaybeGet(conflictSchemasKey)
 	if err != nil {
--- a/go/libraries/doltcore/doltdb/table.go
+++ b/go/libraries/doltcore/doltdb/table.go
@@ -161,7 +161,8 @@ func (t *Table) GetConflicts(ctx context.Context) (Conflict, types.Map, error) {
 		return Conflict{}, types.EmptyMap, err
 	}
 	if !ok {
-		return Conflict{}, types.EmptyMap, ErrNoConflicts
+		return Conflict{}, types.EmptyMap, nil
+		//return Conflict{}, types.EmptyMap, ErrNoConflicts
 	}
 
 	schemas, err := ConflictFromTuple(schemasVal.(types.Tuple))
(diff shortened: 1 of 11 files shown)


integration-tests/bats/sql-conflicts.bats
```diff
@@ -31,7 +31,8 @@ teardown() {
 }
 
 @test "sql-conflicts: read from empty table" {
-    dolt sql -q "SELECT * FROM dolt_conflicts_empty"
+    run dolt sql -q "SELECT * FROM dolt_conflicts_empty"
+    [ $status -eq 1 ] || false
```
timsehn, line 35: Why does this error now?

timsehn (commented):
One comment

go/libraries/doltcore/doltdb/table.go
```diff
@@ -161,7 +161,8 @@ func (t *Table) GetConflicts(ctx context.Context) (Conflict, types.Map, error) {
 		return Conflict{}, types.EmptyMap, err
 	}
 	if !ok {
-		return Conflict{}, types.EmptyMap, ErrNoConflicts
+		confMap, _ := types.NewMap(ctx, t.ValueReadWriter())
```
zachmu, line 164: Add method docs detailing what is returned if there are no conflicts

go/libraries/doltcore/merge/conflict_source.go
```diff
@@ -47,7 +47,10 @@ type ConflictReader struct {
 // NewConflictReader returns a new conflict reader for a given table
 func NewConflictReader(ctx context.Context, tbl *doltdb.Table) (*ConflictReader, error) {
 	base, sch, mergeSch, err := tbl.GetConflictSchemas(ctx)
-	if err == doltdb.ErrNoConflicts {
+	if err != nil {
```
zachmu, line 50: Callers of this method are still checking for ErrNoConflicts

Need to delete that error type entirely to make sure you have all references

zachmu (commented):
What you have looks good, but you need to be more thorough with cleaning up ErrNoConflicts. Should delete that error type entirely.

go/libraries/doltcore/merge/conflict_source.go
JCOR11599: I think I removed all references to this error since I deleted `ErrNoConflicts` from `go/libraries/doltcore/doltdb/errors.go`

go/libraries/doltcore/doltdb/table.go
```diff
@@ -155,13 +155,15 @@ func (t *Table) SetConflicts(ctx context.Context, schemas Conflict, conflictData
 	return &Table{t.vrw, updatedSt}, nil
 }
 
+// GetConflicts returns a map built from ValueReadWriter when there are no
```
zachmu, line 158: Incomplete thought here

zachmu (approved):
LG, sorry about the red herring comment about error handling earlier

PR #2435, opened 2021-11-24
Adding columns to primary key now adds not null constraint

JCOR11599:
Also added a test case to ensure adding not null constraint to primary key column does nothing.


zachmu (commented):
You need tests on CreateTable as well, same rules apply

zachmu (approved):
This looks good, but it needs tests for CREATE TABLE as well. Need to assert that primary key columns get a not null constraint on them (that remains there after the primary key is dropped).

This should fail, but passes in dolt today:

```sql
create table mpks (a int null, b int, primary key (a,b));
```

Try it in mysql and confirm. We should have bats tests for this.

zachmu (approved):
LGTM

PR #2437, opened 2021-11-29
parquet reader for import update and replace

jennifersp:
Added parquet reader for import update and import replace only, for import create, schema file is required.


go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -715,6 +720,10 @@ func transformToDoltRow(row row.Row, rdSchema schema.Schema, wrSchema sql.Schema
 	for i, col := range wrSchema {
 		switch col.Type {
 		case sql.Boolean, sql.Int8, sql.MustCreateBitType(1): // TODO: noms bool wraps MustCreateBitType
+			switch doltRow[i].(type) {
+			case int8:
+				doltRow[i] = strconv.Itoa(int(doltRow[i].(int8)))
```
VinaiRachakonda, line 725: Have two cases. One from int8 and the other for string. No need to do an additional string conversion. The int8 can be left alone. The boolean type in the sqlengine should process it accordingly.

go/libraries/doltcore/schema/typeinfo/time.go
```diff
@@ -63,10 +63,18 @@ func (ti *timeType) ConvertValueToNomsValue(ctx context.Context, vrw types.Value
 	if v == nil {
 		return types.NullValue, nil
 	}
-	val, err := ti.sqlTimeType.Marshal(v)
-	if err != nil {
-		return nil, err
+	var val int64
+	switch v.(type) {
+	case int64:
+		val = v.(int64)
```
VinaiRachakonda, line 69: This line of code doesn't make sense. Even if v is already an int64 it should still be marshalled appropriately.

go/libraries/doltcore/table/typed/parquet/reader.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, err
+	}
+
+	columns := sche.GetAllCols().GetColumns()
+	num := pr.GetNumRows()
+
+	data := make(map[string][]interface{})
+	var colName []string
+	for _, col := range columns {
+		// table.Values, table.RepetitionLevels, table.DefinitionLevels, nil
```
VinaiRachakonda, line 69: Either give a more detailed comment on what these things mean or delete.

go/libraries/doltcore/table/typed/parquet/reader.go
```diff
@@ -0,0 +1,145 @@
...
+func NewParquetReader(vrw types.ValueReadWriter, fr source.ParquetFile, sche schema.Schema) (*ParquetReader, error) {
+	pr, err := reader.NewParquetColumnReader(fr, 4)
+	if err != nil {
+		return nil, err
+	}
+
+	columns := sche.GetAllCols().GetColumns()
+	num := pr.GetNumRows()
+
+	data := make(map[string][]interface{})
```
VinaiRachakonda, line 66: Having all the column in memory like this could be a problem in the future. Please leave a TODO

go/libraries/doltcore/table/typed/parquet/writer.go
```diff
@@ -103,12 +102,20 @@ func (pwr *ParquetWriter) WriteRow(ctx context.Context, r row.Row) error {
...
-			switch val.(type) {
-			case time.Time:
+			// convert datetime and time types to int64
+			switch colT.TypeInfo.GetTypeIdentifier() {
+			case typeinfo.DatetimeTypeIdentifier:
 				val = val.(time.Time).Unix()
+			case typeinfo.TimeTypeIdentifier:
+				colVal, ok := r.GetColVal(colT.Tag)
+				if !ok {
+					return fmt.Errorf("could not get row value")
```
VinaiRachakonda, line 116: need a more expressive error

VinaiRachakonda (changes requested):
I want to review this pr again once the Time changes are done

andrew-wm-arthur (approved):

PR #2438, opened 2021-11-29
Export method needed for dolthub fix

VinaiRachakonda:


zachmu (commented):
Un-export these, otherwise LGTM

zachmu (approved):

PR #2439, opened 2021-11-30
add force flag to dolt checkout

jennifersp:
added -f 'force' flag to dolt checkout. with force flag, the current changes will be wiped

### DIFF
--- a/integration-tests/bats/checkout.bats
+++ b/integration-tests/bats/checkout.bats
@@ -104,7 +104,7 @@ SQL
 
 @test "checkout: with -f flag without conflict" {
     dolt sql -q 'create table test (id int primary key);'
-    dolt sql -q 'insert into test (id) values (10);'
+    dolt sql -q 'insert into test (id) values (8);'
     dolt add .
     dolt commit -m 'create test table.'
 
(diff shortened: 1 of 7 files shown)


VinaiRachakonda:
The tests aren't correct. You need to make sure this works in between branches locally not just with remotes. For example this is the control flow i'm looking for

1) Make a change on branch b
2) Try to checkout to another branch. Assert correct error
3) Force checkout to the other branch. Assert data is unchanged
4) Checkout back to branch b. Assert original change was dropped

integration-tests/bats/checkout.bats
```diff
@@ -97,8 +97,74 @@ SQL
...
+    dolt checkout -b branch1
+    dolt sql -q 'insert into test (id) values (1), (2), (3);'
+    dolt add .
+    dolt commit -m 'add some values to branch 1.'
+
+    run dolt checkout -f main
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Switched to branch 'main'" ]] || false
+
+    run dolt table export test test1.sql
```
VinaiRachakonda, line 120: Get rid of these exports and just use selects

VinaiRachakonda (approved):

PR #2446, opened 2021-11-30
default PrimaryKeySchema, maintain primary key ordinals

max-hoffman:
Use the new GMS `sql.PrimaryKeySchema`, implement new `PrimaryKeyAlterableTable` methods, add new `TestPkOrdinals` enginetests.

Other datails:
- `schemaImpl` and the serialized schema has a new field for PK ordinals
- support backwards compatibility by using the column definition order for PKs by default
- addPk, dropPk, addColumn, dropColumn, and modifyColumn all track pkOrdinal changes

Future:
- some methods still return `sql.Schema` for printing/vizualization
- diffing/merging are not changed (partially because `SHOW CREATE TABLE` does not print the correct order still)


go/libraries/doltcore/schema/alterschema/addcolumn.go
```diff
@@ -46,20 +46,20 @@ type ColumnOrder struct {
 //
 // Returns an error if the column added conflicts with the existing schema in tag or name.
 func AddColumnToTable(ctx context.Context, root *doltdb.RootValue, tbl *doltdb.Table, tblName string, tag uint64, newColName string, typeInfo typeinfo.TypeInfo, nullable Nullable, defaultVal, comment string, order *ColumnOrder) (*doltdb.Table, error) {
```
zachmu, line 48: These alterschema methods need unit tests for these operations

zachmu (approved):
Think I'm fine with this, with the proviso that we need to dramatically increase test coverage for different foreign key orderings.

Make sure that the schema alteration operations you're doing here are well covered by the tests in the alterschema package.

PR #2454, opened 2021-12-01
go/cmd/dolt: commands/sqlserver: Add configuration for disabling MULTI_STATEMENTS support on the server.

reltuk:
Currently sql-server parses some MULTI_STATEMENTS incorrectly, in particular
statements that include substatements get split up incorrectly. This adds
server configuration for the yaml config, so that multi-statement handling can
be disabled server-side. This will allow things like CREATE TRIGGERs submitted
with a client that advertises MULTI_STATEMENTS support to work correctly with
Dolt sql-server. However, if the client actually sends a multi-statement in
this mode, even two SELECTs separated by a ';', for example, while the server
is running in this mode, it will not work.

To enable, place:

```yaml
behavior:
  disable_client_multi_statements: false
```

in the config.yaml for the server.


integration-tests/bats/sql-server.bats
```diff
@@ -1011,6 +1011,103 @@ while True:
...
+log_level: debug
+behavior:
+  disable_client_multi_statements: true
+user:
+  name: dolt
+listener:
+  host: "0.0.0.0"
+  port: $PORT
+EOF
+    dolt sql-server --config ./config.yml &
```
zachmu, line 1029: You can use `start_sql_server_with_config` for this

zachmu (approved):
LGTM

PR #2457, opened 2021-12-03
Add import benchmarks and add in import performance fix

VinaiRachakonda:
Two noticeable problems.

1) No longer using BulkImportTableEditor. Needs to be threaded through in case the batch mode is set in the session.
2) Performance overhead from DoltRow to SqlRow conversion throughout the insert process. This should be reduced in the future as row.Row is deprecated

Current results
```
generating test file
Running the current version of import
Rows Processed: 1000000, Additions: 1000000, Modifications: 0, Had No Effect: 0
Import completed successfully.

real	0m47.759s
user	1m16.795s
sys	0m25.848s
Running version 0.34.5
Rows Processed: 1000000, Additions: 1000000, Modifications: 0, Had No Effect: 0
Import completed successfully.

real	0m37.736s
user	1m12.005s
sys	0m30.281s
```

### DIFF
--- a/benchmark/import_tester/csv_gen.py
+++ b/benchmark/import_tester/csv_gen.py
@@ -1,0 +1,136 @@
+import sys
+import json
+import uuid
+import string
+import random
+
+random.seed(0)
+
+# this is dramatically faster than generating a random string for every row. Generate a buffer
+# then select random portions of the buffer.
+random_string_buffer_size = 128*1024
+letters = string.ascii_lowercase + string.ascii_uppercase
+random_string_buffer = ''.join(random.choice(letters) for i in range(random_string_buffer_size))
+
+def sequential_int(row_count, col):
+    i = 0
+    def f():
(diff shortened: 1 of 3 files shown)


benchmark/import_tester/csv_gen.py
```diff
@@ -0,0 +1,136 @@
...
+        start = random.randint(0, random_string_buffer_size-length)
+        return '"' + random_string_buffer[start:start+length] + '"'
+    return f
+
+
+generator_methods = {
+    "int": {"random": random_int, "sequential": sequential_int, "shuffled_sequential": shuffled_sequential_int},
+    "uuid": {"random": random_uuid},
+    "string": {"random": random_string},
+    "float": {"random": random_float},
```
zachmu, line 77: Fine for v1 but you should also cover timestamp types

zachmu (approved):
LGTM

PR #2459, opened 2021-12-03
fix: userNameRequiredError message in CLI

peter-gy:
Running `dolt init` before configuring `user.email` and `user.name` results (correctly) in error; however, the CLI hint itself is incorrect:

```text
Author identity unknown

*** Please tell me who you are.

Run

  dolt config --global user.email "you@example.com"
  dolt config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name not allowed
```

The proposed commands for identity configuration do not include the `--add` flag, resulting in a new error message: `Exactly one of the -add, -get, -unset, -list flags must be set.`

This PR fixes the CLI message by adding the `--add` flag to the proposed commands.

### DIFF
--- a/go/cmd/dolt/cli/command.go
+++ b/go/cmd/dolt/cli/command.go
@@ -259,8 +259,8 @@ const (
 
 Run
 
-  dolt config --global user.email "you@example.com"
-  dolt config --global user.name "Your Name"
+  dolt config --global --add user.email "you@example.com"
+  dolt config --global --add user.name "Your Name"
 
 to set your account's default identity.
 Omit --global to set the identity only in this repository.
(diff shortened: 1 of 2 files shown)


timsehn:
Thank you so much for the contribution. @jennifersp will review this and add a bats test so the behavior does not regress.

jennifersp (approved):
Hi Peter, I updated a bats test for checking correct CLI hint. Thank you!

timsehn:
Looks good. I'll merge. Thanks for the contribution.

peter-gy:
Thanks a lot for the friendly & smooth review process!

Issue #2484, opened 2021-12-09
Index breaks query when using lower()


timsehn:
Tip of master

```
test-lower $ dolt init
Successfully initialized dolt data repository.
test-lower $ dolt sql -q "create table test (id int, c1 varchar(255), c2 varchar(255), primary key(id))"
test-lower $ dolt sql -q "insert into test (1, 'Aaaa', 'Bbbb'), (2, 'aAaa', 'bBbb')"
test-lower $ dolt sql -q "insert into test values (1, 'Aaaa', 'Bbbb'), (2, 'aAaa', 'bBbb')"
Query OK, 2 rows affected
test-lower $ dolt sql -q "select * from test"
+----+------+------+
| id | c1   | c2   |
+----+------+------+
| 1  | Aaaa | Bbbb |
| 2  | aAaa | bBbb |
+----+------+------+
test-lower $ dolt sql -q "select * from test where lower(c1)='aaaa' and lower(c2)='bbbb'"
+----+------+------+
| id | c1   | c2   |
+----+------+------+
| 1  | Aaaa | Bbbb |
| 2  | aAaa | bBbb |
+----+------+------+
test-lower $ dolt sql -q "create index test on test (c1,c2)"
test-lower $ dolt sql -q "select * from test where lower(c1)='aaaa' and lower(c2)='bbbb'"
+----+----+----+
| id | c1 | c2 |
+----+----+----+
+----+----+----+
```

PR #2485, opened 2021-12-10
Skipped test for Index overuse issue

timsehn:
Added skipped bats test for the issue where an index is used inappropriately when a function modifies the column producing incorrect results.


timsehn:
Skipped test for issue #2484

PR #2488, opened 2021-12-10
Fixed bug with incorrectly applied index lookups for multi column indexes

zachmu:


Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/2484

PR #2490, opened 2021-12-10
Fix auto fwt transform to not drop a row on flush

VinaiRachakonda:


zachmu (approved):
Nice fix

integration-tests/bats/diff.bats
```diff
@@ -626,3 +626,27 @@ SQL
...
+    do
+      if [ $i -eq 1 ]
+      then
+        VALUES="${VALUES}($i,$i)"
+      else
+        VALUES="${VALUES},($i,$i)"
+      fi
+    done
+
+    dolt sql -q "INSERT INTO t values $VALUES"
```
timsehn, line 644: You could be cool here and make sure this diff is 1007 lines :-)

timsehn (approved):

PR #2497, opened 2021-12-13
/go/libraries/doltcore/doltdb: Extracted Persistent Object from `doltdb.Table`

andrew-wm-arthur:
This refactor is the first step in supporting a new storage format. The primary focus of this change is to modify `doltdb.Table` to extract its backing `types.Struct`. 

Throughout the `doltdb` package, in-memory objects are coupled with the persistent `types` objects that are used to de/serialize them to/from disk. The intention of this refactor and subsequent work is to decouple storage details from persistent objects. For example, `schema.Schema` is an in-memory abstraction that has a separate utility in the `encoding` package to read/write schemas to/from disk. This has created a clean separation of concerns for `schema.Schema`, and makes it easy to swap out serialization formats.

### DIFF
--- a/go/cmd/dolt/commands/commit.go
+++ b/go/cmd/dolt/commands/commit.go
@@ -312,7 +312,7 @@ func docCnfsOnWorkingRoot(ctx context.Context, dEnv *env.DoltEnv) (bool, error) 
 		return false, nil
 	}
 
-	return docTbl.HasConflicts()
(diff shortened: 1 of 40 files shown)


andrew-wm-arthur:
#benchmark

zachmu (approved):
Suggest adding a package doc README to the new durable package explaining its purpose

Otherwise full steam ahead

PR #2505, opened 2021-12-15
metrics in sql-server command

bheni:


bheni:
already merged in #2511

PR #2512, opened 2021-12-15
/go/libraries/doltcore/sqle: Refactor Mergable Index Tests

andrew-wm-arthur:
Uses `plan.Inspect` to extract `*noms.ReadRange`s from the test queries, rather than creating a test-specific database, table, etc.

This is part of a larger refactor to move `sqle` index code to its own package


Hydrocharged (approved):
LGTM! I like the changes

PR #2513, opened 2021-12-15
go: remotestorage: hedge_test: Fix transient flake on wg.Done() usage in a test context.

reltuk:


andrew-wm-arthur (approved):
looks good

PR #2514, opened 2021-12-15
Better tests for alter table, and added the ability to modify a table with AUTO_INCREMENT

zachmu:


integration-tests/bats/auto_increment.bats
```diff
@@ -570,14 +570,81 @@ SQL
...
+    [[ "${lines[0]}" =~ "pk" ]] || false
+    [[ "${lines[1]}" =~ "1" ]] || false
+    [[ "${lines[2]}" =~ "2" ]] || false
+    [[ "${lines[3]}" =~ "3" ]] || false
+    [[ "${lines[4]}" =~ "4" ]] || false
+    [[ "${lines[5]}" =~ "5" ]] || false
+    [[ "${lines[6]}" =~ "6" ]] || false
+}
+
+@test "auto_increment: alter table modify column to auto inc works" {
```
timsehn, line 589: Maybe test a remove auto increment using alter

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1507,7 +1595,20 @@ func (t *AlterableDoltTable) updateFromRoot(ctx *sql.Context, root *doltdb.RootV
...
+	if err != nil {
+		return err
+	}
+
+	updatedTable, _, err := root.GetTable(ctx, t.tableName)
+	if err != nil {
+		return err
+	}
+
+	sch, err := updatedTable.GetSchema(ctx)
```
max-hoffman, line 1608: i'm not sure i understand why we're going table -> root -> schema in some of these cases

integration-tests/bats/auto_increment.bats
max-hoffman: also do we already test alter add primary key auto_increment?

go/libraries/doltcore/sqle/tables.go
zachmu: ALTER TABLE statements can have any number of clauses, which means that CreateCheck could be called after AddColumn for example. `t.sch` is out of date in this case (having been set at table creation time). So we need to get the updated version of the table and schema in the session root.

integration-tests/bats/auto_increment.bats
zachmu: You can't specify auto_increment when adding a primary key, you have to add the key, then modify the column

go/libraries/doltcore/schema/col_coll.go
```diff
@@ -143,6 +143,29 @@ func (cc *ColCollection) Append(cols ...Column) *ColCollection {
 	return NewColCollection(allCols...)
 }
 
+// IndexOf returns the index of the column with the name given (case-insensitive) or -1 if it's not found
+func (cc *ColCollection) IndexOf(colName string) (int, error) {
+	idx := -1
```
max-hoffman, line 148: in the past you've recommended leaning on the `NameToCol` and `TagToIndex` maps

max-hoffman (approved):
LGTM

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1007,12 +1016,91 @@ func (t *AlterableDoltTable) ModifyColumn(ctx *sql.Context, columnName string, c
...
+		}
+
+		rowData, err := updatedTable.GetRowData(ctx)
+		if err != nil {
+			return err
+		}
+
+		// Note that we aren't calling the public PartitionRows, because it always gets the table data from the session
+		// root, which hasn't been updated yet
+		rowIter, err := partitionRows(ctx, updatedTable, t.projectedCols, sqlutil.SinglePartition{RowData: rowData})
```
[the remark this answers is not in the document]
zachmu: At Andy's advice, decided to do this entirely in the SQL layer as much as possible. First draft inspected row data directly.

go/libraries/doltcore/schema/col_coll.go
zachmu: Those both use case-sensitive match unfortunately, makes them more trouble than they're worth from the SQL layer

max-hoffman: you added `LowerNameToCol` also

PR #2529, opened 2021-12-17
/go/libraries/doltcore/sqle/writer: Extracted Interface for WriteSession

andrew-wm-arthur:
Extracted an interface from TableEditSession to allow for multiple implementations for different formats


go/libraries/doltcore/sqle/writer/table_editor.go
```diff
@@ -41,17 +40,19 @@ type TableEditor interface {
 	NextAutoIncrementValue(potentialVal, tableVal interface{}) (interface{}, error)
 }
 
-// sqlTableEditor is a wrapper for *doltdb.SessionedTableEditor that complies with the SQL interface.
+type SessionRootSetter func(ctx *sql.Context, dbName string, root *doltdb.RootValue) error
```
zachmu, line 43: Probably should go higher in the file, with a type comment

zachmu: These days I'm wary of defining top-level callback types like this, and especially embedding them in other structs like you do below. In general it seems like a sign that we're missing some other load-bearing abstraction. My rule of thumb is that function entities belong as params to methods only, almost never as fields.

go/libraries/doltcore/table/editor/bulk_import_tea.go
```diff
@@ -51,6 +51,19 @@ func (cap *fixedCapacity) capacityExceeded(size int) bool {
 
 var _ TableEditAccumulator = (*BulkImportTEA)(nil)
 
+// Options are properties that define different functionality for the tableEditSession.
+type Options struct {
```
zachmu, line 55: This seems misplaced, is there a better file for this to live in?

zachmu (approved):
Generally like the direction this moves us in, just a few comments

go/libraries/doltcore/sqle/writer/table_editor.go
reltuk: In general I think I agree that that often naked function types can point to some missing abstraction that would actual help structure things better. But I did want to point out that embedding function fields in structs to provide pluggable functionality is common throughout the go standard library and I think it's a legitimate and reasonable approach. See `http.Client`, `http.Transport`, `tls.Config`, etc.

andrew-wm-arthur: In this case it points to a flaw in abstractions. I'm still working out what exactly is missing. 

`SessionRootSetter` is set to `dsess.Session.SetRoot()`, which currently needs to be called each time we `Close()` a sqlTableEditor.

PR #2551, opened 2022-01-03
/go/gen/fb/serial: Flatbuffers Serialization Format

andrew-wm-arthur:
Proposal for a new serialization format based on [Flatbuffers](https://github.com/google/flatbuffers). Flatbuffers has a number of features that make it appealing as a serialization/storage format for Dolt:
- Forward/Backward compatibility for format evolution
- Client library code generation for many languages
- Stable binary format
- Zero-copy deserialization, good performance
- (g)RPC support

However, it's not completely free. Serialized leaf nodes for indexes are roughly 1.6% bigger than the current candidate format in `prolly/Node.go`. Serialization overhead for a leaf node with 200 key-value tuple pairs:
```
key:   (int32)
value: (int32, int32, int32)

data        3200 = 800 + 2400
offsets     796  = sizeof(uint16) * (199 + 199)
metadata    11   = TupleFormat * 2, tree_count, tree_level
flatbuffers 65   = (1.6% overhead)
total size  4072
```

This overhead estimate is after "externalizing" the format for tuples, meaning that key/value tuples in Prolly tree nodes are stored in raw byte buffers and cannot be parsed/deserialized with Flatbuffers alone. Encoding each tuple in an individual Flatbuffers `struct` or `table` would be prohibitively expensive in both space and time.


bheni (approved):
lgtm

PR #2578, opened 2022-01-07
/go/doltcore/{doltdb, sqle}: Extracted `durable.Index` interface for `types.Map`

andrew-wm-arthur:
This PR adds the `durable.Index` interface to abstract `types.Map` out of primary and secondary indexes. Currently, this interface has limited functionality, and in almost all cases will need to be unwrapped to be useful. This is the first step is a large refactor to decouple Index logic from a Noms-specific implementation.


go/libraries/doltcore/doltdb/durable/index.go
```diff
@@ -40,53 +83,69 @@ func NewIndexSet(ctx context.Context, vrw types.ValueReadWriter) IndexSet {
...
 type nomsIndexSet struct {
 	indexes types.Map
 	vrw     types.ValueReadWriter
 }
 
 var _ IndexSet = nomsIndexSet{}
 
-func (c nomsIndexSet) GetIndex(ctx context.Context, name string) (types.Map, error) {
-	v, ok, err := c.indexes.MaybeGet(ctx, types.String(name))
+// HashOf implementes IndexSet.
```
zachmu, line 93: Implementes, is that spanish

go/libraries/doltcore/doltdb/durable/index.go
```diff
@@ -40,53 +83,69 @@ func NewIndexSet(ctx context.Context, vrw types.ValueReadWriter) IndexSet {
...
+func (s nomsIndexSet) HashOf() (hash.Hash, error) {
+	return s.indexes.Hash(s.vrw.Format())
+}
+
+// GetIndex implementes IndexSet.
```
zachmu, line 98: Do you have a generator for this or something

go/libraries/doltcore/sqle/rows.go
```diff
@@ -128,13 +128,11 @@ func newKeyedRowIter(ctx context.Context, tbl *doltdb.Table, projectedCols []str
...
 }
 
-func iterForPartition(ctx context.Context, partition *doltTablePartition) (types.MapTupleIterator, error) {
-	rowData := partition.rowData
+func iterForPartition(ctx context.Context, partition doltTablePartition) (types.MapTupleIterator, error) {
 	if partition.end == NoUpperBound {
-		return rowData.RangeIterator(ctx, 0, rowData.Len())
-	} else {
-		return partition.IteratorForPartition(ctx, rowData)
+		partition.end = partition.rowData.Count()
```
zachmu, line 133: This appears to be a no-op, this value is never read in this function

go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -21,11 +21,21 @@ import (
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/typed/noms"
 	"github.com/dolthub/dolt/go/store/types"
 )
 
+type DoltIndex interface {
```
zachmu, line 30: Needs some docs on this type

go/libraries/doltcore/doltdb/durable/index.go
```diff
@@ -18,20 +18,63 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/dolthub/dolt/go/store/hash"
+
 	"github.com/dolthub/dolt/go/store/types"
 )
 
+type Index interface {
```
zachmu, line 26: Presumably this is a stub? Would be nice to understand what the intended final shape is

zachmu (approved):
No objections

go/libraries/doltcore/doltdb/durable/index.go
andrew-wm-arthur: > Presumably this is a stub? Would be nice to understand what the intended final shape is

That would be nice...

go/libraries/doltcore/doltdb/durable/index.go
andrew-wm-arthur: multi-cursor!

PR #2580, opened 2022-01-07
Modify Add Column to apply defaults in engine

VinaiRachakonda:
This pr removes the direct row map access in addcolumn.go when adding a new column and pushes the logic to the engine. The engine uses the Updater interface to apply the default value.


go/libraries/doltcore/schema/alterschema/addcolumn_test.go
```diff
@@ -116,7 +115,7 @@ func TestAddColumnToTable(t *testing.T) {
 			defaultVal: mustStringToColumnDefault(`("default")`),
 			expectedSchema: dtestutils.AddColumnToSchema(dtestutils.TypedSchema,
 				schemaNewColumn("newCol", dtestutils.NextTag, types.StringKind, false, `("default")`, schema.NotNullConstraint{})),
-			expectedRows: dtestutils.AddColToRows(t, dtestutils.TypedRows, dtestutils.NextTag, types.String("default")),
+			expectedRows: dtestutils.AddColToRows(t, dtestutils.TypedRows, dtestutils.NextTag, types.NullValue),
```
zachmu, line 118: This is fine to check in, but you need to fast followup with additional tests, probably in `DoltScripts`. This code used to exercise the "fill in the default value of existing rows with new column" logic, and it doesn't do that anymore.

So 1) reproduce these tests as SQL scripts, and 2) pare these down so they're just unit tests for the AddColumnToTable method

zachmu (approved):
Change looks fine, needs more tests

Also go through and make sure nothing is using the AddColumn codepath outside SQL, as it won't give the right results outside the engine anymore. Also add a comment about this limitation

zachmu (approved):
LGTM

PR #2584, opened 2022-01-07
/go/store/types/format.go: Added `NomsBinFormat` for new storage format

andrew-wm-arthur:


go/store/types/format.go
```diff
@@ -36,9 +46,11 @@ type formatTag struct{}
 
 var formatTag_7_18 *formatTag = nil
 var formatTag_LD_1 = &formatTag{}
+var formatTag_DOLT_1 = &formatTag{}
 
 var Format_7_18 = &NomsBinFormat{}
 var Format_LD_1 = &NomsBinFormat{formatTag_LD_1}
+var Format_DOLT_1 = &NomsBinFormat{tag: formatTag_DOLT_1}
```
reltuk, line 53: Nit: I guess the expression to construct this value and the value just above should look the same. Either name the field above, or drop the field name here?

reltuk (approved):
LGTM!

PR #2592, opened 2022-01-10
sql-server: Fix regression: Respect autocommit config.

reltuk:
In the migration to *SqlEngine in 0.34.4 to 0.34.5, the server config's respect
for the default value for autocommit regressed. In 0.34.5 - 0.35.3,
`autocommit` is always turned on by default for incoming connections,
regardless of the setting in the config.yaml file or the flag
`--no-auto-commit` being passed at the command line.

### DIFF
--- a/go/cmd/dolt/commands/engine/sqlengine.go
+++ b/go/cmd/dolt/commands/engine/sqlengine.go
@@ -115,7 +115,7 @@ func NewSqlEngine(
 	return &SqlEngine{
 		dbs:            nameToDB,
 		contextFactory: newSqlContext(sess, initialDb),
-		dsessFactory:   newDoltSession(pro, mrEnv.Config()),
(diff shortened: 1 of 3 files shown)


zachmu (commented):
Think we need a test in sql-server.bats for this too, using `start_sql_server_with_config` and making sure the @@autocommit value matches what's in the config.yaml

reltuk:
> Think we need a test in sql-server.bats for this too, using `start_sql_server_with_config` and making sure the @@autocommit value matches what's in the config.yaml

That's in there...the default config has `autocommit: false` and the mysql script is now asserting that `0 |` comes back for `select @@autocommit`. AFAICT, the python client can't be used to assert the new behavior because it always sets autocommit for the session.

zachmu (approved):

PR #2609, opened 2022-01-13
Andy/prolly indexes

andrew-wm-arthur:
This PR adds a second implementation of `durable.Index` and attempts to refactor pkg `sqle` to use `durable.Index` in place of `types.Map`. This is a major refactor in the SQL engine with the aim of supporting both the existing storage format and the new-and-improved™ storage format. As a measure of progress roughly 70% of the engine tests are passing against `sqle` while using the new format. 

Various changes were made in pkg `val` in order to support the SQL types/encodings need for sysbench. Some of these encodings are experimental and will change in the future.


andrew-wm-arthur:
#benchmark

andrew-wm-arthur:
#benchmark

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -0,0 +1,296 @@
...
+	if err := m.mut.Delete(ctx, oldKey); err != nil {
+		return err
+	}
+
+	for to, from := range m.keyMap {
+		m.keyBld.PutField(to, newRow[from])
+	}
+	newKey := m.keyBld.Build(sharePool)
+
+	ok, err := m.mut.Has(ctx, newKey)
```
zachmu, line 241: This tracks unapplied mutations?

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+		return
+	})
+
+	s := m.keyBld.Desc.Format(key)
+
+	return sql.NewUniqueKeyErr(s, true, existing)
+}
+
+type colMapping []int
```
zachmu, line 277: could use a comment

go/store/prolly/node_store.go
```diff
@@ -33,36 +41,235 @@ type NodeStore interface {
...
 }
+
+func (ns nodeStore) Format() *types.NomsBinFormat {
+	// todo(andy): read from |ns.store|
+	return types.Format_DOLT_1
+}
+
+type centry struct {
+	c    chunks.Chunk
+	i    int
```
zachmu, line 107: ?

go/store/val/codec.go
```diff
@@ -55,27 +61,26 @@ type Encoding uint8
...
-	Int16Enc   Encoding = 3
-	Uint16Enc  Encoding = 4
-	Int24Enc   Encoding = 5
-	Uint24Enc  Encoding = 6
+	NullEnc   Encoding = 0
+	Int8Enc   Encoding = 1
+	Uint8Enc  Encoding = 2
+	Int16Enc  Encoding = 3
+	Uint16Enc Encoding = 4
+	// Int24Enc   Encoding = 5
```
zachmu, line 69: ?

zachmu (approved):
Overall looks good, just a few comments.

Seems like the new prolly store files could use some unit tests, e.g. node_store_test.go. Probably a good idea to add those as soon as the interface stabilizes.

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
andrew-wm-arthur: yeah buddy, got a sweet skip list in there

go/store/prolly/node_store.go
andrew-wm-arthur: "cache entry" stole this from Aaron's implementation in `dolthubapi`

go/store/val/codec.go
andrew-wm-arthur: still undecided if we're going to support a 3 byte integer encoding

PR #2611, opened 2022-01-13
Spatial types actually stored as spatial types

JCOR11599:
The ReadPoint, ReadLine, and ReadPolygon methods just use ReadString and convert to correct types.
Able to call st_methods on column types now.


go/store/types/linestring.go
```diff
@@ -0,0 +1,200 @@
...
+func (v Linestring) Hash(nbf *NomsBinFormat) (hash.Hash, error) {
+	return getHash(v, nbf)
+}
+
+func (v Linestring) isPrimitive() bool {
+	return true
+}
+
+func (v Linestring) WalkValues(ctx context.Context, cb ValueCallback) error {
+	return nil
```
andrew-wm-arthur, line 110:
```suggestion
	for _, p := range v.Points {
		if err := cb(p); err != nil {
			return err
		}
	}
```

go/libraries/doltcore/schema/typeinfo/linestring.go
```diff
@@ -0,0 +1,180 @@
...
+	return ConvertSQLLinestringToTypesLinestring(line.(sql.Linestring)), nil
+}
+
+// Equals implements TypeInfo interface.
+func (ti *linestringType) Equals(other TypeInfo) bool {
+	if other == nil {
+		return false
+	}
+	if ti2, ok := other.(*linestringType); ok {
+		return ti.sqlLinestringType.Type() == ti2.sqlLinestringType.Type()
```
Hydrocharged, line 105: This may technically work, but generally we're comparing the parameters assuming the `TypeInfo`s line up. If there are specific parameters that need to be checked, then we should compare those directly (see `blobstring.go`). If there aren't any parameters to check, then the `TypeInfo` check itself is fine (see `time.go`).

go/libraries/doltcore/schema/typeinfo/linestring.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Promote implements TypeInfo interface.
+func (ti *linestringType) Promote() TypeInfo {
+	return &linestringType{ti.sqlLinestringType.Promote().(sql.LinestringType)}
+}
+
+// String implements TypeInfo interface.
+func (ti *linestringType) String() string {
+	return "Linestring()"
```
Hydrocharged, line 174: These don't have any parentheses in the name if there are no parameters to put in there.

go/libraries/doltcore/schema/typeinfo/linestring.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// ToSqlType implements TypeInfo interface.
+func (ti *linestringType) ToSqlType() sql.Type {
+	return ti.sqlLinestringType
+}
```
Hydrocharged, line 180: Need to add `linestringTypeConverter` for this type (and the relevant function for your other added types too), along with the relevant entries in every other existing type (and the `GetTypeConverter` function in `typeconverter.go`). It's not the most efficient implementation but it's what we have to do for now.

These functions handle the conversion from any one type to any other type, hence why every type has its own `TypeConverter` function, and those functions list every other type.

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -158,6 +164,18 @@ func FromSqlType(sqlType sql.Type) (TypeInfo, error) {
...
+		// TODO: bad, but working way to determine which specific geometry type
+		switch sqlType.String() {
+		case sql.PolygonType{}.String():
+			return &polygonType{sqlType.(sql.PolygonType)}, nil
+		case sql.LinestringType{}.String():
+			return &linestringType{sqlType.(sql.LinestringType)}, nil
+		case sql.PointType{}.String():
+			return &pointType{sqlType.(sql.PointType)}, nil
+		default:
+			return nil, fmt.Errorf(`expected "PointTypeIdentifier" from SQL basetype "Point"`)
```
Hydrocharged, line 177: SQL base type Geometry?

Hydrocharged (commented):
Left feedback on `linestring.go`, but the same feedback also applies to `point.go` and `polygon.go`, so make sure to do those changes on those files as well.

go/libraries/doltcore/schema/typeinfo/linestring.go
JCOR11599: Does this mean that all other TypeConverter functions now need to list pointType, linestringType, and polygonType?

Hydrocharged: Correct!

Hydrocharged:
@JCOR11599 In `typeconverter.go`, under the function `wrapConvertValueToNomsValue`, you'll also need to add your new noms types as well. It's also worthwhile to at least check the behavior in MySQL for converting to the `VARCHAR` and `TEXT` types from your geometry types, as they may hold special significance. The rest of the types just wrapping `ConvertValueToNomsValue` should be fine for now.

Here's an example of what I mean using the `BIT` type. If we convert from `BIT` to `VARCHAR` this is what we get:
```sql
DROP TABLE IF EXISTS test;
CREATE TABLE test (pk BIGINT PRIMARY KEY, v1 BIT(24));
INSERT INTO test VALUES (1, 5001036);
ALTER TABLE test MODIFY COLUMN v1 VARCHAR(10);
SELECT * FROM test;
```
Which outputs:
```
+----+---------+
| pk | v1      |
+----+---------+
| 1  | 5001036 |
+----+---------+
```
However, when we convert from `BIT` to `TEXT` this is what we get:
```sql
CREATE TABLE test (pk BIGINT PRIMARY KEY, v1 BIT(24));
INSERT INTO test VALUES (1, 5001036);
ALTER TABLE test MODIFY COLUMN v1 TINYTEXT;
SELECT * FROM test;
```
Which gives you:
```
+----+-----+
| pk | v1  |
+----+-----+
| 1  | LOL |
+----+-----+
```
You can see some of these tests under `go/libraries/doltcore/sqle/altertests`, but _do not_ worry about making your own. These were automatically generated against MySQL. Someday they'll be finished, but no time soon. Also this can come in a separate PR if you want to get this in now (assuming there's any changes to begin with).

go/store/types/linestring.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (v Linestring) Less(nbf *NomsBinFormat, other LesserValuable) (bool, error) {
+	// Compare types
+	v2, ok := other.(Linestring)
+	if !ok {
+		return LinestringKind < other.Kind(), nil
+	}
+	// TODO: should I even take this into account?
+	// Compare SRID
+	if v.SRID != v2.SRID {
```
Hydrocharged, line 74: Regarding the TODO, I'd make linestrings in MySQL with their only difference being their SRID. Insert them in some random order, then query with an `ORDER BY` and see if MySQL takes it into account. If it does order them, then do another one but with the SRID in descending order and the linestring values in ascending order (or vice versa, as long as they're opposite). Then do another `ORDER BY` and it will tell you the priority (whether we should compare points first or the SRID first), as right now you have the SRID first (which may be correct).

Test this (and the other ones if needed, don't see a TODO on those) and make a new PR with the changes (if any). We 100% need to know before we remove the feature flag though.

integration-tests/bats/sql-spatial-types.bats
```diff
@@ -0,0 +1,23 @@
...
+    run dolt sql -q "create table point_tbl (p point)"
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "cannot be made" ]] || false
+}
+
+@test "sql-spatial-types: can make spatial types with flag" {
+    DOLT_ENABLE_SPATIAL_TYPES=true run dolt sql -q "create table point_tbl (p point)"
+    [ "$status" -eq 0 ]
+    [[ "$output" = "" ]] || false
+}
```
Hydrocharged, line 23: Need some more bats tests using the new types. Add those in a new PR.

Hydrocharged (approved):
Looks good! Get this merged in for now, and you can address my feedback in a separate PR.

PR #2639, opened 2022-01-19
Fixed bug with unresolved foreign keys being resolved in transactions

Hydrocharged:
The logic that handled foreign key merging during transaction commits did not take unresolved foreign keys into consideration, therefore errors were produced. This fixes that bug.


VinaiRachakonda (approved):
LGTM

PR #2652, opened 2022-01-20
dump_docs.go now includes verifyConstraintsDocs

alecstein:
`dump_docs` doesn't dump the docs for `verify constraints`, which means they're not visible in the online documentation. Hopefully this fixes that.

After building I checked that it worked locally and seemed to produce the right output.

Context: this is my first PR and my first time editing go code. Entering the great unknown here.  :-)


zachmu (approved):
LGTM!

PR #2653, opened 2022-01-20
std out and err redirects

bheni:
Allow for stdout and stderr to be redirected to a file via a command line argument.  When running outside of a terminal window it can be really inconvenient to redirect io.  These parameters were added to make that easier.


zachmu (approved):
LGTM

PR #2654, opened 2022-01-21
Removed Parse from typeinfo, some type conversion fixes

Hydrocharged:
Back when `GetTypeConverter` was added, `Parse` was deprecated. This PR now fully removes it. In addition, a few bugs regarding type conversions were fixed, a new test enforcing the inclusion of all types in `GetTypeConverter` was added, and a few more comments were added to clarify the different string type implementations.


andrew-wm-arthur (approved):
LGTM!

PR #2662, opened 2022-01-24
Add ordinal to merge schema after merge

VinaiRachakonda:
Still need to add test cases here

### DIFF
--- a/go/libraries/doltcore/merge/merge_schema.go
+++ b/go/libraries/doltcore/merge/merge_schema.go
@@ -118,6 +118,12 @@ func SchemaMerge(ourSch, theirSch, ancSch schema.Schema, tblName string) (sch sc
 	if err != nil {
 		return nil, sc, err
 	}
+
+	err = sch.SetPkOrdinals(ourSch.GetPkOrdinals())
+	if err != nil {
+		return nil, sc, err
+	}
+
 	_ = mergedIdxs.Iter(func(index schema.Index) (stop bool, err error) {
 		sch.Indexes().AddIndex(index)
 		return false, nil
--- a/go/libraries/doltcore/merge/merge_schema.go
+++ b/go/libraries/doltcore/merge/merge_schema.go
@@ -119,6 +119,8 @@ func SchemaMerge(ourSch, theirSch, ancSch schema.Schema, tblName string) (sch sc
 		return nil, sc, err
 	}
 
+	// TODO: We should enforce the existing primary key set pk ordinals as out of order primary
+	// key ordinal merge is not supported yet.
 	err = sch.SetPkOrdinals(ourSch.GetPkOrdinals())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/merge/merge_schema.go
```diff
@@ -118,6 +118,12 @@ func SchemaMerge(ourSch, theirSch, ancSch schema.Schema, tblName string) (sch sc
 	if err != nil {
 		return nil, sc, err
 	}
+
+	err = sch.SetPkOrdinals(ourSch.GetPkOrdinals())
```
max-hoffman, line 122: what if the three schemas have different pk ordinals? how to tell which one to keep?

VinaiRachakonda: I believe we do not support merge between tables with competing ordinals.

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
```diff
@@ -738,4 +738,57 @@ var DoltTransactionTests = []enginetest.TransactionTest{
...
+			{
+				Query:    "/* client a */ select * from test order by x",
+				Expected: []sql.Row{{1, 1, 1}, {2, 3, 2}, {2, 5, 2}},
+			},
+			{
+				Query:    "/* client b */ select * from test order by x",
+				Expected: []sql.Row{{1, 1, 1}, {2, 3, 2}, {2, 5, 2}},
+			},
+			{
+				Query:       "/* client b */ insert into test values (4,3,2)",
```
max-hoffman, line 789: doesn't strike me as the most direct way to test the change, but if it mimics the original customer issue then it make sense

go/libraries/doltcore/merge/merge_schema.go
```diff
@@ -118,6 +118,14 @@ func SchemaMerge(ourSch, theirSch, ancSch schema.Schema, tblName string) (sch sc
 	if err != nil {
 		return nil, sc, err
 	}
+
+	// TODO: We should enforce the existing primary key set pk ordinals as out of order primary
+	// key ordinal merge is not supported yet.
```
max-hoffman, line 123: "merge conflict should have blocked any ordinal change"?

max-hoffman (approved):
LGTM, a simpler unit test would have been helpful, the script test seems sufficient

max-hoffman (approved):
LGTM, a simpler unit test would have been helpful, the script test seems sufficient

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
VinaiRachakonda: Yeah without the change, this becomes an incorrect primary key issue and the ordering of the select changes. This indicates that primary key set has changed.

PR #2681, opened 2022-01-25
Minor updates to schema export/import command documentation

fulghum:


go/cmd/dolt/commands/init.go
```diff
@@ -43,7 +43,7 @@ Running dolt init in an already initialized directory will fail.
 `,
 
 	Synopsis: []string{
-		//`[{{.LessThan}}options{{.GreaterThan}}] [{{.LessThan}}path{{.GreaterThan}}]`,
+		"",
```
zachmu, line 46: Should this be an empty slice?

zachmu (approved):
LGTM!

go/cmd/dolt/commands/init.go
fulghum: Making it an empty slice will cause help and markdown to omit the synopsis section. Probably okay for this simple command, but it's nice for consistency in the docs for all commands to have the same sections.

PR #2685, opened 2022-01-25
Disable --batch mode for `dolt sql` by default

zachmu:
--batch mode for `dolt sql` has issues in when it flushes to disk which make it incompatible with some SQL scripts. It's now only enabled when asked for explicitly.

As part of this work, also made multiple statements in a single -q block work by default (used to require --batch or --disable-batch, the latter of which is retired)

Finally, added a --file option to `dolt sql`, useful for environments where it's hard to redirect STDIN.


timsehn:
I'm in favor of this.

VinaiRachakonda (approved):
LGTM

go/cmd/dolt/commands/sql.go
```diff
@@ -135,10 +136,10 @@ func (cmd SqlCmd) ArgParser() *argparser.ArgParser {
...
 	ap.SupportsFlag(listSavedFlag, "l", "Lists all saved queries")
 	ap.SupportsString(messageFlag, "m", "saved query description", "Used with --query and --save, saves the query with the descriptive message given. See also --name")
-	ap.SupportsFlag(BatchFlag, "b", "batch mode, to run more than one query with --query, separated by ';'. Piping input to sql with no arguments also uses batch mode")
-	ap.SupportsFlag(disableBatchFlag, "", "When issuing multiple statements, used to override more efficient batch processing to give finer control over session")
+	ap.SupportsFlag(BatchFlag, "b", "Use to enable more efficient batch processing for large SQL import scripts consisting of only INSERT statements. Other statements types are not guaranteed to work in this mode.")
 	ap.SupportsString(multiDBDirFlag, "", "directory", "Defines a directory whose subdirectories should all be dolt data repositories accessible as independent databases within ")
 	ap.SupportsFlag(continueFlag, "c", "continue running queries on an error. Used for batch mode only.")
+	ap.SupportsString(fileInputFlag, "", "input file", "Execute statements from the file given")
```
VinaiRachakonda, line 142: Don't see any test cases for this flag

go/cmd/dolt/commands/sql.go
```diff
@@ -207,30 +206,37 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
 		} else {
-			runInBatchMode = fi.Mode()&os.ModeCharDevice == 0
+			isTty = fi.Mode()&os.ModeCharDevice != 0
+		}
+
+		_, continueOnError := apr.GetValue(continueFlag)
+
+		input := os.Stdin
+		if fileInput, ok := apr.GetValue(fileInputFlag); ok {
+			input, err = os.OpenFile(fileInput, os.O_RDONLY, os.ModePerm)
```
VinaiRachakonda, line 225: missing error check

go/cmd/dolt/commands/sql.go
zachmu: Good call, will add a couple

PR #2695, opened 2022-01-27
Bug fix for dropped table check constraints when altering a table.

fulghum:
Resolves: https://github.com/dolthub/dolt/issues/2672

Confirmed that alterschema tests now cover all schema properties.

### DIFF
--- a/integration-tests/bats/schema-changes.bats
+++ b/integration-tests/bats/schema-changes.bats
@@ -68,6 +68,9 @@ teardown() {
     dolt sql -q "alter table test add constraint test_check CHECK (c2 < 12345);"
     dolt sql -q "alter table test rename column c2 to c0"
 
+    # TODO: This test can be enabled/finished once the go-mysql-server validation for checks
+    #       has been merged in:
+    #       https://github.com/dolthub/go-mysql-server/pull/774
     skip "dolt incorrectly allows a column used in a table check expression to be renamed/removed"
 }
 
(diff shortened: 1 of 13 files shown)


timsehn (commented):
Since we f'ed it up, should we have a bats test as well?

zachmu:
Column name integrity for ALTER should be enforced in the engine, not in dolt. Should be doable.

fulghum:
Added a BATS test; I'll keep digging into how we can enforce integrity for table constraints on schema alter.

integration-tests/bats/schema-changes.bats
```diff
@@ -37,6 +37,16 @@ teardown() {
     [[ ! "$ouput" =~ "Merge failed" ]] || false
 }
 
+@test "schema-changes: dolt schema alter column preserves table checks" {
+    run dolt sql -q "alter table test add constraint test_check CHECK (c2 < 12345);"
+    [ "$status" -eq 0 ]
```
timsehn, line 42: So if all you want to do is check that the command returns 0 just omit `run`. For example 41 and 42 can just be:
`dolt sql -q "alter table test add constraint test_check CHECK (c2 < 12345);"`

`run` populates $status, $output, @lines, etc

fulghum:
Also resolves: https://github.com/dolthub/dolt/issues/2708

fulghum:
Related go-mysq-server PR that adds additional validation to preserve table check constraint integrity when adding or removing columns: https://github.com/dolthub/go-mysql-server/pull/774

zachmu (approved):
LGTM, but you should also add engine tests in go-mysql-server for the operations where this bug occurred in a follow-up PR.

PR #2699, opened 2022-01-27
Add CreateTagOnDB and DeleteTagsOnDB

### LINKED ISSUE #2693: actions.CreateTag and actions.DeleteTag can't be used from dolthubapi
`actions.CreateTag` and `actions.DeleteTag` cannot be accessed from dolthubapi since both require passing `env.DoltEnv`. 

Please change function signatures to match [actions.CreateBranchOnDB](https://github.com/dolthub/dolt/blob/5e37076540e5942bdc6f2a833e55bc1170abb8fe/go/libraries/doltcore/env/actions/branch.go#L205) where `ddb *doltdb.DoltDB` and `headRef ref.DoltRef` (if required) are passed instead.

Proposed signatures:
```go
func CreateTag(ctx context.Context, ddb *doltdb.DoltDB, tagName, startPoint string, props TagProps, headRef ref.DoltRef)

func DeleteTags(ctx context.Context, ddb *doltdb.DoltDB, tagNames ...string) error
```

druvv:
Closes #2693

### DIFF
--- a/go/libraries/doltcore/env/actions/tag.go
+++ b/go/libraries/doltcore/env/actions/tag.go
@@ -31,9 +31,13 @@ type TagProps struct {
 }
 
 func CreateTag(ctx context.Context, dEnv *env.DoltEnv, tagName, startPoint string, props TagProps) error {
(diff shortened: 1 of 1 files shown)


andrew-wm-arthur (approved):
lgtm

PR #2714, opened 2022-01-31
BATS tests for Spatial Types

JCOR11599:


integration-tests/bats/sql-spatial-types.bats
```diff
@@ -20,4 +20,31 @@ teardown() {
...
+    [[ "$output" = "" ]] || false
+    DOLT_ENABLE_SPATIAL_TYPES=true run dolt sql -q "insert into point_tbl () values (point(1,2))"
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Query OK" ]] || false
+}
+
+@test "sql-spatial-types: create linestring table and insert linestring" {
+    DOLT_ENABLE_SPATIAL_TYPES=true run dolt sql -q "create table line_tbl (l linestring)"
+    [ "$status" -eq 0 ]
+    [[ "$output" = "" ]] || false
```
timsehn, line 37: This is a useless test. Your regexing an empty string. All strings contain an empty string :-)

timsehn (dismissed):

integration-tests/bats/sql-spatial-types.bats
timsehn: Then you need to do `[ "$output" = "" ]`. Not sure what  the regex version does in this context.

Hydrocharged (changes requested):
For each of these tests, insert a few more values that just one, and `SELECT` the contents to make sure we're getting back what we expect. Right now this is only testing that the `INSERT` statement succeeds, not that anything is actually persisted, let alone correctly. Multiple values would surface other potential issues, such as incorrect value boundaries in noms (since these also add new noms types).

integration-tests/bats/sql-spatial-types.bats
timsehn: https://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash#:~:text=Double%20Brackets%20i.e.%20%5B%5B%5D%5D,%2C%20literally.

A bit of a deep dive if you're curious. Turns out they are equivalent in this case but I prefer the single brackets fro equality because you don't need the || false for the ancient mac version of bash to work consistently.

JCOR11599: Ok, will change

Hydrocharged (approved):
LGTM! What happens if a Polygon doesn’t have the same start and end? Is that illegal? If it is may as well add that extra line testing. If it’s legal then no need to test.

PR #2715, opened 2022-01-31
go/store/datas: Decouple pull/push/clone from store/types, and move it into its own package.

reltuk:


go/store/types/ref.go
```diff
@@ -220,3 +222,19 @@ func (r Ref) String() string {
 func (r Ref) HumanReadableString() string {
 	panic("unreachable")
 }
+
+// Returns a function that can be used to walk the hash and height of all the
+// Refs of a given Chunk.  This function is meant to decouple callers from the
+// types package itself, and so the callback itself does not take |types.Ref|
+// values.
+func WalkRefsForChunkStore(cs chunks.ChunkStore) (func(chunks.Chunk, func(h hash.Hash, height uint64) error) error, error) {
```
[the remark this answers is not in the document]
andrew-wm-arthur: nvm, you define `WalkRefs` in pull.go

go/store/datas/pull/pull.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+		updateProgress(1, 0, sampleSize/uint64(math.Max(1, float64(sampleCount))))
+	})
+	if err != nil {
+		return nil, err
+	}
+	return neededChunks, nil
+}
+
+type WalkRefs func(chunks.Chunk, func(hash.Hash, uint64) error) error
```
andrew-wm-arthur, line 190: Do we ever need the `height uint64` param? It's not currently available in the new format

andrew-wm-arthur (approved):
LGTM!

go/store/datas/pull/pull.go
reltuk: AFAICT, the existing stuff could be adapted to not need it. It's currently used for reporting how many levels of the tree we have left to process and for keeping track of leaves so that we don't WalkRefs them.

In the new format, I think both of those use cases can be satisfied basedon the self-reported height of the chunk itself.

In the past, `jankyfetch` used the height to aggressively attack the highest refs before going after the lower portions of the DAG. A similar effect could probably be had by using the height of the node the reference is in as the heuristic.

PR #2716, opened 2022-01-31
go/store/prolly: Prefix Range Scans for Prolly Trees

andrew-wm-arthur:
Expands range-scan functionality for pkg `prolly` to include ranges defined on a prefix of key fields. This is necessary to support partial indexes and secondary index in Dolt. Various refactors were made to support this change:

- removed support for "reverse" ranges, which aren't used anywhere
- refactored range iters for `prolly.Map` to use a `nodeCursor` to track the end of the range


go/store/prolly/map.go
```diff
@@ -210,26 +264,35 @@ func (m Map) compareKeys(left, right val.Tuple) int {
...
 }
 
-type mapTupleCursor struct {
-	cur *nodeCursor
+type prollyRangeIter struct {
+	// current tuple location
+	curr *nodeCursor
+	// non-inclusive range stop
+	stop *nodeCursor
+	rng  Range
```
reltuk, line 272: Is this used?

go/store/prolly/node_cursor.go
```diff
@@ -341,7 +327,7 @@ func (cur *nodeCursor) retreatInBounds(ctx context.Context) (bool, error) {
 			return true, nil
 		}
 		// if not |ok|, then every parent, grandparent, etc.,
-		// failed to retreatInBounds(): we're before the start.
+		// failed to retreatInBounds(): we're before the start	.
```
reltuk, line 330: nit.

reltuk (approved):
LGTM!

PR #2728, opened 2022-02-01
Fixing BATS test for table check constraint validation errors

fulghum:


zachmu (approved):
LGTM

zachmu:
You should run `go get github.com/dolthub/go-mysql-server@main` to upgrade dolt to gms-main at the same time

fulghum:
thanks for the tip on go get @ main. I had a replace with a workspace override, but @ main seems like a cleaner/safer way to do it.

PR #2745, opened 2022-02-03
Retain not null constraint on primary key columns when dropping primary key

jennifersp:
- Fixes adding primary keys using db.table syntax and unskipped bats test for it
- Fixes issue of making all columns NOT NULL when adding pk with alter table
(With the above issue, dropping pk bats test was not catching retaining of null constraint)
- Fixes retaining NOT NULL constraint when dropping pk with alter table


timsehn:
Should you update the bats test so that it would catch it if it happened again?

jennifersp:
oh, I updated the bats test for it, but for some reason it did not get pushed. I'm pushing it now.

integration-tests/bats/primary-key-changes.bats
```diff
@@ -617,7 +617,7 @@ SQL
     dolt sql -q "alter table t drop primary key"
     run dolt sql -q "show create table t"
     [ $status -eq 0 ]
-    [[ "$output" =~ "NOT NULL" ]] || false
+    [[ "$output" =~ "`pk` int NOT NULL" ]] || false
```
zachmu, line 620: Should have a test for a table with two primary key columns and additional non-primary key columns as well

zachmu (commented):
Code looks good but needs more tests

Remember that your PR titles show up in release notes, they should be a one-sentence summary of what you fixed

In this case suggest "Retain not null constraint on primary key columns when dropping primary key"

timsehn:
I think the description should also say:

"Also fixes adding primary keys using db.table syntax"

zachmu (approved):
LGTM

VinaiRachakonda (approved):
LGTM

PR #2750, opened 2022-02-04
Fix --set-upstream bug when feature branch already exists on remote

max-hoffman:
We were missing a test for push --set-upstream where
the feature branch had already been pushed from another
clone.


go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -136,7 +132,7 @@ func DoPush(ctx context.Context, rsr env.RepoStateReader, rsw env.RepoStateWrite
 		})
 	}
 
-	return nil
+	return err
```
zachmu, line 135: This is a little overbroad, I think.

You should check for a specific error type you want to defer before continuing onward. Other error types should be treated as unrecoverable and terminate the method early.

zachmu (approved):
LG, just one chang

timsehn:
Prompted @zachmu to review this because it has come up a few times for one of our customers. Please get it in.

PR #2754, opened 2022-02-04
DOLT_BLAME system view support

fulghum:
Initial support for the new DOLT_BLAME system view.


timsehn:
Those are some nice bats...

integration-tests/bats/blame-system-view.bats
```diff
@@ -0,0 +1,208 @@
...
+    setup_common
+    setup_repository
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+setup_repository() {
```
zachmu, line 14: These are great tests, but going forward we want bats tests of SQL features to mostly be smoke tests. Most of these would be better expressed as engine tests.

go/libraries/doltcore/schema/super_schema.go
```diff
@@ -69,6 +69,14 @@ func (ss *SuperSchema) AddColumn(col Column) (err error) {
 			return fmt.Errorf("tag collision for columns %s and %s, different definitions (tag: %d)",
 				ecName, col.Name, col.Tag)
 		}
+
+		// If the column's primary key metadata has changed, update SuperSchema's
+		// representation of the column to match.
+		if existingCol.IsPartOfPK != col.IsPartOfPK {
```
zachmu, line 75: Seems like this is unrelated to the primary purpose of this PR, right? Could it be its own PR, or is this a bug that affects tests you're writing?

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -395,6 +395,11 @@ func (dp *diffPartition) isDiffablePartition(ctx *sql.Context) (bool, error) {
 		return false, err
 	}
 
+	// dp.to can be nil if a table was deleted and then recreated
+	if dp.to == nil {
```
zachmu, line 399: Same comment, is this bug related to this PR>

zachmu (commented):
This is definitely in the right direction, just a couple comments about tests and the scope of the PR

integration-tests/bats/blame-system-view.bats
fulghum: Cool; happy to move these to engine tests.

go/libraries/doltcore/schema/super_schema.go
fulghum: Yup, it came out as part of testing the DOLT_BLAME view. It isn't directly related to the DOLT_BLAME functionality, but it was causing my tests to fail because of how they were written (altering PKs to test different schemas). This definitely felt like it needed more testing, so I put it in its own commit; it'd be easy to pull it out to another branch/change/PR and look at it separately.

go/libraries/doltcore/sqle/dtables/diff_table.go
fulghum: This one is related to the functionality of DOLT_BLAME (for some cases at least), since it's built on top of DOLT_DIFF_<TABLE>. It also felt like it needed a little bit more thorough vetting to make sure this was the right change though.

go/libraries/doltcore/schema/super_schema.go
fulghum: Removed this commit from this PR and pushed to a new branch (fulghum/superschema-primarkey-bugfix). I'll circle back to that after getting the DOLT_BLAME view out.

go/libraries/doltcore/sqle/dtables/diff_table.go
fulghum: To keep this PR simple... I went ahead and reverted this commit, and pushed it onto a new local branch (fulghum/diff-table-bugfix-2496). I'll follow up on that one after landing the DOLT_BLAME system view.

integration-tests/bats/blame-system-view.bats
fulghum: Actually... how do you feel about leaving the BATS test around for this initial commit to get this view out to customers faster, then I can loop back to converting them to enginetests? Having a DOLT_CONFIG() function to change the user for the blame test data would make writing the enginetests cleaner. Happy to push them into enginetests now if you feel strongly about it though.

zachmu: Fine for now, sure

zachmu (approved):
LGTM

Next steps:

* Move bats tests to be in Go (keep one as a smoke test)
* Reimplement `dolt blame` to select from this new view

PR #2766, opened 2022-02-07
go/store/prolly: diffs for new prolly tree

andrew-wm-arthur:


go/store/prolly/map.go
```diff
@@ -71,6 +72,28 @@ func NewMapFromTuples(ctx context.Context, ns NodeStore, keyDesc, valDesc val.Tu
...
+		return err
+	}
+
+	for {
+		var diff Diff
+		if diff, err = differ.Next(ctx); err != nil {
+			break
+		}
+
+		if err = cb(ctx, diff); err != nil {
```
reltuk, line 87: Nit(?): Is it weird to let this return `io.EOF` and not forward it along to the caller?

go/store/prolly/tree_diff_test.go
```diff
@@ -0,0 +1,206 @@
...
+		10000,
+	}
+
+	for _, s := range scales {
+		name := fmt.Sprintf("test proCur map at scale %d", s)
+		t.Run(name, func(t *testing.T) {
+			prollyMap, tuples := makeProllyMap(t, s)
+			require.Equal(t, s, len(tuples))
+
+			t.Run("empty map diff", func(t *testing.T) {
```
reltuk, line 44: Nit: this seems more like "equal map diff"...agreed the set of diffs is empty, but it sounds like we're diffing an empty map.

Might be good to add: diff empty vs. prollyMap; diff prollyMap vs. empty; diff where the callback returns an error forwards the error along.

reltuk (approved):
LGTM! A few tiny comments.

PR #2769, opened 2022-02-08
New DOLT_DIFF system table 

fulghum:
Adds support for a new DOLT_DIFF system table that is not scoped to a specific user table; for each commit, this table includes one row for every table changed in that commit.

Sample Output: 
```
> SELECT * FROM DOLT_DIFF;
+----------------------------------+-----------+-------------------------+-----------------------------------+-------------------------+------------+
| commit_hash                      | committer | email                   | date                              | message                 | table_name |
+----------------------------------+-----------+-------------------------+-----------------------------------+-------------------------+------------+
| edr1ichcj8vpve9lofv31e7taiajv3uu | jfulghum  | jason@dolthub.com       | 2022-02-07 13:39:47.717 -0800 PST | Creating tables z       | z          |
| 91ff8so7alcnuiq0qa1qq9e7o58i7jdb | jfulghum  | jason@dolthub.com       | 2022-02-07 13:39:26.143 -0800 PST | Creating tables x and y | x          |
| 91ff8so7alcnuiq0qa1qq9e7o58i7jdb | jfulghum  | jason@dolthub.com       | 2022-02-07 13:39:26.143 -0800 PST | Creating tables x and y | y          |
+----------------------------------+-----------+-------------------------+-----------------------------------+-------------------------+------------+

```

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
+++ b/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
@@ -1,0 +1,227 @@
+// Copyright 2022 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dtables
+
+import (
+	"io"
+	"sort"
+
+	"github.com/dolthub/go-mysql-server/sql"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/diff"
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
+)
+
+// UnscopedDiffTable is a sql.Table implementation of a system table that shows which tables have
+// changed in each commit, across all branches.
+type UnscopedDiffTable struct {
+	ddb  *doltdb.DoltDB
+	head *doltdb.Commit
+}
+
+// NewUnscopedDiffTable creates an UnscopedDiffTable
+func NewUnscopedDiffTable(_ *sql.Context, ddb *doltdb.DoltDB, head *doltdb.Commit) sql.Table {
+	return &UnscopedDiffTable{ddb: ddb, head: head}
+}
+
+// Name is a sql.Table interface function which returns the name of the table which is defined by the constant
+// LogTableName
+func (dt *UnscopedDiffTable) Name() string {
+	return doltdb.DiffTableName
+}
+
+// String is a sql.Table interface function which returns the name of the table which is defined by the constant
+// DiffTableName
+func (dt *UnscopedDiffTable) String() string {
+	return doltdb.DiffTableName
+}
+
+// Schema is a sql.Table interface function that returns the sql.Schema for this system table.
+func (dt *UnscopedDiffTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{Name: "commit_hash", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: true},
+		{Name: "committer", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: false},
+		{Name: "email", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: false},
+		{Name: "date", Type: sql.Datetime, Source: doltdb.LogTableName, PrimaryKey: false},
+		{Name: "message", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: false},
+		{Name: "table_name", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: false},
+	}
+}
+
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -466,6 +466,229 @@ func TestSystemTableQueries(t *testing.T) {
 	enginetest.RunQueryTests(t, newDoltHarness(t), BrokenSystemTableQueries)
 }
 
+func TestUnscopedDoltDiffSystemTable(t *testing.T) {
+	// There's a bug in queries with where clauses that compare column equality with a
+	// variable. These tests use "commit_hash in (@Commit1)" to work around that bug.
+	// https://github.com/dolthub/go-mysql-server/issues/790
+
+	var tests = []enginetest.ScriptTest{
```
fulghum, line 474: I wanted to pull these out into a separate file... is extracting the []ScriptTest var into a separate file the preferred pattern?

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -466,6 +466,229 @@ func TestSystemTableQueries(t *testing.T) {
...
+		{
+			Name: "unscoped-diff-system-table: basic case",
```
zachmu, line 476: No need to do name scoping like this (like we do in bats), because golang tests have built in nested scope via the t.Run methods. So just call them a simple name. Like this one would be something like "three tables"

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -0,0 +1,227 @@
...
+	ddb          *doltdb.DoltDB
+	commits      []*doltdb.Commit
+	commitIdx    int
+	tableNames   []string
+	tableNameIdx int
+}
+
+// NewUnscopedDiffTableItr creates a UnscopedDiffTableItr from the current environment.
+func NewUnscopedDiffTableItr(ctx *sql.Context, ddb *doltdb.DoltDB, head *doltdb.Commit) (*UnscopedDiffTableItr, error) {
+	commits, err := actions.TimeSortedCommits(ctx, ddb, head, -1)
```
zachmu, line 87: Fine for a first pass, but you probably want an iterator that fetches commits one at a time, rather than this which gets them all at once

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -0,0 +1,227 @@
...
+// String is a sql.Table interface function which returns the name of the table which is defined by the constant
+// DiffTableName
+func (dt *UnscopedDiffTable) String() string {
+	return doltdb.DiffTableName
+}
+
+// Schema is a sql.Table interface function that returns the sql.Schema for this system table.
+func (dt *UnscopedDiffTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{Name: "commit_hash", Type: sql.Text, Source: doltdb.LogTableName, PrimaryKey: true},
```
zachmu, line 56: It doesn't matter because this is only a result set, but for doc purposes, you need another key col than just commit_hash. I think it's actually (commit_hash, table_name), right?

I would also put table name as the second column for this reason

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -0,0 +1,227 @@
...
+
+// calculateChangedTables calculates the tables that changed in the specified commit, by comparing that
+// commit with its immediate ancestor commit.
+func (itr *UnscopedDiffTableItr) calculateChangedTables(commit *doltdb.Commit) ([]string, error) {
+	toRootValue, err := commit.GetRootValue()
+	if err != nil {
+		return nil, err
+	}
+
+	parent, err := itr.ddb.ResolveParent(itr.ctx, commit, 0)
```
zachmu, line 177: Nit: you're doing more work than necessary here, since you're already going to resolve the parent when you iterate over it. Fine for a first pass though.

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -0,0 +1,227 @@
...
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/diff"
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
+)
+
+// UnscopedDiffTable is a sql.Table implementation of a system table that shows which tables have
+// changed in each commit, across all branches.
+type UnscopedDiffTable struct {
```
zachmu, line 31: On a second pass, you probably want to implement sql.FilteredTable, which will let you only examine commit hashes in a where clause instead of the entire history.

zachmu (approved):
Very nice! No major comments, some suggestions about how to make this performant that could be tackled in another revision.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
fulghum: ✅

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
fulghum: Ah! Good catch! Fixed now. ✅

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
fulghum: Sounds good; I've got the two perf improvements noted in my queue and will follow up with a separate PR for them: 1)don't load all commits up front, and 2) implemented sql.FilteredTable interface.

PR #2771, opened 2022-02-08
Bug fix for an index out of bounds panic in DOLT_HISTORY_<TABLE>

fulghum:
Panic occurs when selecting from `DOLT_HISTORY_<TABLE>` using a filter expression and the underlying table that doesn't have a primary key.

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/filtered_reader.go
+++ b/go/libraries/doltcore/sqle/dtables/filtered_reader.go
@@ -249,18 +249,20 @@ func CreateReaderFuncLimitedByExpressions(nbf *types.NomsBinFormat, tblSch schem
 	pkCols := tblSch.GetPKCols()
 	var keySet setalgebra.Set = setalgebra.UniversalSet{}
 	var err error
-	for _, filter := range filters {
-		var setForFilter setalgebra.Set
-		setForFilter, err = getSetForKeyColumn(nbf, pkCols.GetByIndex(0), filter)
 
-		if err != nil {
-			break
-		}
+	if len(pkCols.GetColumns()) > 0 {
+		for _, filter := range filters {
+			var setForFilter setalgebra.Set
 
-		keySet, err = keySet.Intersect(setForFilter)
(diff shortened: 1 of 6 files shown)


timsehn (commented):
Maybe add a keyless table diff and history test in bats?

timsehn:
https://github.com/dolthub/dolt/blob/main/integration-tests/bats/system-tables.bats

fulghum:
> Maybe add a keyless table diff and history test in bats?

Just added a new commit with BATS coverage for selecting from DOLT_HISTORY_$TABLE with a filter expression for a table with no pks. The dolt_diff table currently requires there be a primary key on the underlying user table.

go/libraries/doltcore/sqle/dtables/filtered_reader.go
```diff
@@ -249,18 +249,20 @@ func CreateReaderFuncLimitedByExpressions(nbf *types.NomsBinFormat, tblSch schem
...
 	var keySet setalgebra.Set = setalgebra.UniversalSet{}
 	var err error
-	for _, filter := range filters {
-		var setForFilter setalgebra.Set
-		setForFilter, err = getSetForKeyColumn(nbf, pkCols.GetByIndex(0), filter)
 
-		if err != nil {
-			break
-		}
+	if len(pkCols.GetColumns()) > 0 {
```
zachmu, line 253: This is fine for a bug fix but this logic ultimately needs to be ripped out and replaced with the unified indexing path that all other tables use. This was implemented back before we had a reasonable way to apply multiple filter conditions to an indexed table scan, and we do now. Just add a TODO.

zachmu (approved):
LGTM

go/libraries/doltcore/sqle/dtables/filtered_reader.go
fulghum: Makes sense; I added a TODO in history_table for it to stop using ThreadSafeCRFuncCache and upgrade to the new unified indexing path. ✅

PR #2772, opened 2022-02-08
Index Transformer

druvv:
IndexTransformer is an io.Reader that takes an io.Reader of the table file index. It specifically transforms the lengths into offsets.


reltuk (approved):
LGTM!

coffeegoddd (approved):
LGTM...nothing immediately sticks out about why `go test -race` takes longer than 30 mins...thoughts?

PR #2773, opened 2022-02-08
go/store/datas: Remove ValueReadWriter from Database interface; make Database interface smaller.

reltuk:
Moving VRW out of the Database interface means users need to keep a VRW and a Database reference in sync and call the Database once the correct VRW has the data in it.


andrew-wm-arthur (approved):
LGTM!

PR #2777, opened 2022-02-08
Fix auto increment imports

VinaiRachakonda:
Adds functionality and tests that allow `dolt table import` to correctly import auto increment data.


timsehn:
Good bug fix and test.

PR #2778, opened 2022-02-09
go/store/prolly: 3-way Merge for Prolly Trees

andrew-wm-arthur:


go/store/prolly/utils_test.go
```diff
@@ -77,23 +77,34 @@ func randomTuplePairs(count int, keyDesc, valDesc val.TupleDesc) (items [][2]val
...
+			break
+		}
+
+		// replace duplicates and validate again
+		for _, d := range dupes {
+			items[d][0] = randomTuple(keyBuilder)
 		}
+		dupes = dupes[:0]
 	}
 	return
```
bheni, line 101: not a fan of naked returns.  Might `return items` while you're here

go/store/prolly/tree_merge.go
```diff
@@ -0,0 +1,229 @@
...
+		return Map{}, err
+	}
+
+	eg, ctx := errgroup.WithContext(ctx)
+	buf := newPatchBuffer(patchBufferSize)
+
+	// iterate |ld| and |rd| in parallel, populating |buf|
+	eg.Go(func() (err error) {
+		defer func() {
+			if cerr := buf.close(); err != nil {
```
bheni, line 57: I think this should be if `err == nil` so it only overwrites the error when there is no error already existing.  This currently overwrites the error only if there is a pre-existing error.

go/store/prolly/tree_merge.go
```diff
@@ -0,0 +1,229 @@
...
+					err = buf.sendPatch(ctx, resolved)
+				}
+				if err != nil {
+					return err
+				}
+			}
+
+			left, err = l.Next(ctx)
+			if err == io.EOF {
+				err, lok = nil, false
```
bheni, line 182: wow, such line savings
much concise

bheni (approved):
lgtm.  One bug. few comments/suggestions.

go/store/prolly/map_test.go
```diff
@@ -20,14 +20,15 @@ import (
...
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/dolthub/dolt/go/store/val"
 )
 
-var testRand = rand.New(rand.NewSource(1))
+var testRand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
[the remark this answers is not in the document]
andrew-wm-arthur: For now I'm going to revert this back to a constant seed. I'll take another pass at this to randomize test data

go/store/prolly/tree_merge.go
andrew-wm-arthur: dangerous!

PR #2782, opened 2022-02-09
Fixing schema ordering to unbreak tests

fulghum:
I added the table_name column as part of the primary key, then changed the order, not realizing that would mess up the test data.


andrew-wm-arthur:
LGTM!

PR #2800, opened 2022-02-14
[no-release-notes] Tweaked interfaces for GMS

zachmu:
Also deleted a noms perf test that doesn't build with gcc 11+


Hydrocharged (approved):
LGTM

PR #2801, opened 2022-02-14
Reduce onHeapTableIndex memory overhead

druvv:
Changes the way onHeapTableIndex is implemented. Instead of parsing the given byte buffer into new prefixes, ordinals, lengths, and suffixes slices, we now compute directly on the given buffer taking ownership of it.

Creating an `onHeapTableIndex` before this PR:
- Put the table file index bytes at the end of a byte slice and pass it to parseTableIndex
- Footer is read to determine chunkCount
- Byte slice is used to compute offsets from the lengths. Then parsed prefixes, ordinals, lengths, offsets, and suffixes are stored as new slices on the heap.
- Byte slice reference is dropped

Creating an `onHeapTableIndex `after this PR:
- Put the table file index bytes at the end of a byte slice and pass it to `parseTableIndex`
- Footer is read to determine chunkCount and chunkCount is used to determine total size of the index. The total size is used to slice away any additional bytes given at the front of the buffer. `onHeapTableIndex` retains a slice of just the index data (without the footer).
- New space is allocated to store offsets. OffsetsReader is used to convert the provided chunk lengths into chunk offsets.
- Byte slice reference is retained

As a result of this change, `mmapTableReader` was broken. Since the new index takes ownership of the buffer, as soon as mmapTableReader unmapped the index file from memory, a segfault was thrown. `mmapTableReader` has instead been changed to copy index bytes into the heap.


go/store/nbs/mmap_table_reader.go
```diff
@@ -105,29 +105,32 @@ func newMmapTableReader(dir string, h addr, chunkCount uint32, indexCache *index
...
+					if unmapErr != nil {
+						err = unmapErr
+					}
+				}()
+				copy(buff, mm[indexOffset-aligned:])
 			}()
 
-			buff := []byte(mm)
-			ti, err = parseTableIndex(buff[indexOffset-aligned:])
+			ti, err = parseTableIndex(buff)
```
reltuk, line 133: Add the `if err != nil` here.

go/store/nbs/table_index.go
```diff
@@ -0,0 +1,506 @@
...
+func parseTableIndex(buff []byte) (onHeapTableIndex, error) {
+	chunkCount, totalUncompressedData, err := ReadTableFooter(bytes.NewReader(buff))
+	if err != nil {
+		return onHeapTableIndex{}, err
+	}
+
+	iS := indexSize(chunkCount)
+	buff = buff[:len(buff)-footerSize]
+	// Trim away any extra bytes
+	buff = buff[uint64(len(buff))-iS:]
```
reltuk, line 103: Can we add a test that we don't actually do this in the case of mmapTableReader here?

The backing array is going to be retained in these cases, and we would rather not waste overallocated bytes...the caller should have no reason to give us a bunch of bytes that aren't part of the index here.

go/store/nbs/table_index.go
```diff
@@ -0,0 +1,506 @@
...
+	lengths := b[prefixTupleSize*chunkCount : prefixTupleSize*chunkCount+lengthSize*chunkCount]
+	suffixes := b[prefixTupleSize*chunkCount+lengthSize*chunkCount:]
+
+	lR := bytes.NewReader(lengths)
+	offsets := make([]byte, chunkCount*offsetSize)
+	_, err := io.ReadFull(NewOffsetsReader(lR), offsets)
+	if err != nil {
+		return onHeapTableIndex{}, err
+	}
+
```
reltuk, line 155: Probably worth adding a TODO here about how `len(lengths)` in the backing array is over retained and we could technically squeeze offsets into that space + another allocation of size `len(lengths)`.

go/store/nbs/table_index.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func (ti onHeapTableIndex) Prefixes() ([]uint64, error) {
+	p := make([]uint64, ti.chunkCount)
+	for i, off := uint32(0), 0; i < ti.chunkCount; i, off = i+1, off+prefixTupleSize {
+		b := ti.tupleB[off : off+addrPrefixSize]
+		p[i] = binary.BigEndian.Uint64(b)
+	}
+	return p, nil
+}
```
reltuk, line 296: Where are these used? Conjoin? They seem less-than-ideal for an API from a memory consumption point of view...

reltuk (approved):
Looks great. A few small comments.

go/store/nbs/table_index.go
druvv: Correct. Conjoin. Agreed, although I haven't thought too deeply about what the tableIndex interface should look like. It depends on how we reimplement conjoin without loading indexes into memory.

go/store/nbs/table_index.go
druvv: @reltuk I changed it to throw an error if the buff was the wrong size and broke a bunch of tests. Traced one of them back to here:
https://github.com/dolthub/dolt/blob/76bff36c632e15980b73f99ec1262a4031b79106/go/store/nbs/file_table_persister.go#L93

In this case, do I copy the data needed data to a new slice to ensure we don't retain unnecessary memory?

reltuk: Yeah, make fsTablePersister copy it before it passes it in.

druvv: I created a new function called parseTableIndexByCopy and updated functions that were failing tests. parseTableIndex by default will now throw an error if the length or capacity of the given buff isn't correct.

Commits are organized to highlight the non-testing changes

reltuk (approved):
LGTM!

PR #2807, opened 2022-02-15
Making `dolt_transaction_commit` a global variable

JCOR11599:


integration-tests/bats/sql-server-config.bats
```diff
@@ -54,6 +54,18 @@ teardown() {
     [[ "$output" =~ "sqlserver.global.max_connections = 1000" ]] || false
 }
 
+@test "sql-server-config: dolt_transaction_commit is global variable" {
+    cd repo1
+    start_sql_server repo1
+
+    insert_query repo1 1 "SET @@PERSIST.dolt_transaction_commit = 1"
```
zachmu, line 61: This only really tests the PERSIST aspect of it. To see if it's GLOBAL you should add a test in sql-server.bats that makes sure that the value persists across sessions.

zachmu (approved):
LG, thanks for the fast turnaround

But see comment about additional test

PR #2810, opened 2022-02-15
Added privilege persistence

Hydrocharged:
This adds the ability for users and roles to be persisted, through a YAML option or a CLI argument, to a JSON file.


andrew-wm-arthur (approved):
LGTM!

PR #2822, opened 2022-02-17
go/cmd/dolt: Add download statistics to dolt clone

druvv:
Previously, it was difficult to understand the progress of `dolt clone`, especially when you are cloning a db with a small number of large table files. `dolt clone` now lists which table files are being concurrently downloaded and shows the progress and download rate for each. 

An example:
Screen Shot 2022-02-16 at 4 45 29 PM

@reltuk Brought up the concern for compatibility between the new `uilive` dep and `github.com/faith/color`. In my testing there doesn't seem to be any issues:

Screen_Shot_2022-02-16_at_3 51 45_PM

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -1216,7 +1216,12 @@ func newTableFile(cs chunkSource, info tableSpec) tableFile {
 				return nil, 0, err
 			}
 
-			return io.NopCloser(r), 0, nil
+			s, err := cs.size()
+			if err != nil {
+				return nil, 0, err
+			}
+
+			return io.NopCloser(r), s, nil
 		},
 	}
 }
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -1216,12 +1216,7 @@ func newTableFile(cs chunkSource, info tableSpec) tableFile {
 				return nil, 0, err
 			}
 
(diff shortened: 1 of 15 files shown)


go/store/nbs/store.go
```diff
@@ -1210,13 +1210,13 @@ func getTableFiles(css map[addr]chunkSource, contents manifestContents, numSpecs
...
-		open: func(ctx context.Context) (io.ReadCloser, error) {
+		open: func(ctx context.Context) (io.ReadCloser, uint64, error) {
 			r, err := cs.reader(ctx)
 			if err != nil {
-				return nil, err
+				return nil, 0, err
 			}
 
-			return io.NopCloser(r), nil
+			return io.NopCloser(r), 0, nil
```
reltuk, line 1219: Should this return the actual file size?

This is going to break reporting on `dolt clone file:///....`, but not sure that's a huge priority.

reltuk (approved):
Looks great. One question.

go/store/nbs/store.go
druvv: I just added `size() (uint64, error)` to the `chunkSource` interface and then in the `tableReader` implementation, it returns the table file size as reported by the index.

I also updated the test cases to check that the reported size is correct.

PR #2823, opened 2022-02-17
[no-release-notes] go/store/datas: Database: SetHead, FastForward, Tag take target hash instead of types.Ref.

reltuk:


andrew-wm-arthur (approved):
LGTM

PR #2825, opened 2022-02-17
Removing Duplicate Index Constraint

JCOR11599:


zachmu (commented):
This is good but incomplete. You also need to test that the engine behaves correctly in the presence of duplicate indexes. Write some additional query tests, including at least one that asserts the query plan uses an index when two identical ones are present.

JCOR11599:
added query plan tests in GMS PR: https://github.com/dolthub/go-mysql-server/pull/809

zachmu (approved):

PR #2826, opened 2022-02-17
Add the ignore skipped row param

VinaiRachakonda:
Adds a flag to skip rows that are printed by --continue.


go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -505,6 +509,13 @@ func move(ctx context.Context, rd table.SqlRowReader, wr *mvdata.SqlEngineTableW
 			return true
 		}
 
+		atomic.AddInt64(&badCount, 1)
+
+		// We can ignore the out of skipped rows when the ignore-skipped-rows param is specified.
```
andrew-wm-arthur, line 514:
```suggestion
		// don's log skipped rows when the ignore-skipped-rows param is specified.
```

andrew-wm-arthur (approved):

PR #2828, opened 2022-02-17
Remove index filters

max-hoffman:


max-hoffman:
#benchmark

timsehn:
This is beautiful.

integration-tests/bats/keyless.bats
```diff
@@ -843,14 +843,6 @@ SQL
     [[ "${lines[1]}" = "1" ]] || false
     [[ "${lines[2]}" = "1" ]] || false
     [ "${#lines[@]}" -eq 3 ]
-
-    run dolt sql -q "describe select c0 from keyless where c1 = 1" -r csv
```
zachmu, line 847: Why removed test?

zachmu: Same for below

zachmu (commented):
@Hydrocharged can you please give this a once-over? Looks fine to me

integration-tests/bats/keyless.bats
max-hoffman: bats is a bad place for query query plans that break often

go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -258,17 +258,22 @@ RangeLoop:
...
+			}
 			rangeCheck[i] = cb
 		}
 
 		// If the suffix checks will always succeed (both bounds are infinity) then they can be removed to reduce the
-		// number of checks that are called per-row.
-		for i := len(rangeCheck) - 1; i >= 0; i-- {
+		// number of checks that are called per-row. Always leave one check to skip NULLs.
+		i := len(rangeCheck) - 1
+		for i > 0 && len(rangeCheck) > 1 {
```
Hydrocharged, line 270: I'd rather have all of the variables in the `for` statement:
`for i := len(rangeCheck) - 1; i > 0 && len(rangeCheck) > 1; i-- {`

go/libraries/doltcore/sqle/index/index_lookup.go
```diff
@@ -244,6 +245,19 @@ func (il *doltIndexLookup) Ranges() sql.RangeCollection {
 // Between returns whether the given types.Value is between the bounds. In addition, this returns if the value is outside
 // the bounds and above the upperbound.
 func (cb columnBounds) Between(ctx context.Context, nbf *types.NomsBinFormat, val types.Value) (ok bool, over bool, err error) {
+	// Only boundCase_isNull matches NULL values,
+	// otherwise we terminate the range scan.
+	// This is checked early to bypass unpredictable
+	// null type comparisons.
+	if val.Kind() == types.NullKind {
+		switch {
+		case cb.boundsCase == boundsCase_isNull:
```
Hydrocharged, line 254: I'd rather this be an `if-else` since it's just two cases, and we're not using the switch as a type switch or anything, but it's nbd either way

Could even do
```
isNullCase := cb.boundsCase == boundsCase_isNull
return isNullCase, !isNullCase, nil
```
Which would remove that branch altogether and should be a tad faster

Hydrocharged (approved):
LGTM! Just small nits

PR #2829, opened 2022-02-17
unskip fixed tests

jennifersp:
- un-skipped some tests that are resolved
- fixed some test typos
- added test for add foreign key constraint bug


timsehn:
LGTM.

PR #2830, opened 2022-02-18
/go/store/prolly: Add Prolly Node array with subtree sizes

andrew-wm-arthur:
- adds new field to prolly tree nodes to store a varint-encoded array of subtree cardinalities 
- alters tree chunker to track subtree cardinalities during tree construction
- splits out node construction logic into `nodeBuilder`

the varint encoding scheme should probably be replaced by something more performant like [streaming vbyte](https://github.com/dolthub/stream-vbyte-go)


reltuk (approved):
LGTM! A few comments :)

PR #2832, opened 2022-02-18
Simplify DOLT_DIFF_$TABLE system tables by removing tag suffixes

fulghum:
When the `DOLT_DIFF_$TABLE` system tables show schema history with column name conflicts, the column names are disambiguated by adding their unique tags as suffixes. This makes it difficult to work with these system tables. This change simplifies the output by restricting the output schema to be based on the current table schema and avoids any column name conflicts. 

Resolves: https://github.com/dolthub/dolt/issues/1910 (for `dolt_diff_$table`, but does not change `dolt_commit_diff_$table yet)


go/libraries/doltcore/sqle/tables.go
```diff
@@ -960,6 +965,66 @@ func (t *AlterableDoltTable) DropColumn(ctx *sql.Context, columnName string) err
 	return t.updateFromRoot(ctx, newRoot)
 }
 
+// dropColumnData drops values for the specified column from the underlying storage layer
```
fulghum, line 968: I could use some extra scrutiny here since this is my first time working with the storage layer.

While testing diffs, I noticed that when you drop a column, there was no diff reported, even though that operation logically modifies your data by removing that column. I wanted that change to show up in diff history, so as part of DropColumn, I added this code to update the stored data to remove the column.

go/libraries/doltcore/diff/diff_source.go
```diff
@@ -36,14 +38,16 @@ type RowDiffSource struct {
 	joiner     *rowconv.Joiner
 	oldRowConv *rowconv.RowConverter
 	newRowConv *rowconv.RowConverter
+	sqlCtx     *sql.Context
 }
 
-func NewRowDiffSource(ad RowDiffer, joiner *rowconv.Joiner) *RowDiffSource {
+func NewRowDiffSource(ad RowDiffer, joiner *rowconv.Joiner, ctx *sql.Context) *RowDiffSource {
```
zachmu, line 44: Always want to put the ctx param (if any) first in golang methods, it's like a biblical law

go/libraries/doltcore/rowconv/row_converter.go
```diff
@@ -68,9 +77,34 @@ func NewRowConverter(ctx context.Context, vrw types.ValueReadWriter, mapping *Fi
...
+
+// Convert takes a row and maps its columns to their destination columns, and performs any type conversion needed to
+// create a row of the expected destination schema.
 func (rc *RowConverter) Convert(inRow row.Row) (row.Row, error) {
+	return rc.convert(nil, inRow, false)
+}
+
+// convert takes a row and maps its columns to their destination columns, automatically performing any type conversion
+// needed, and optionally logging SQL warnings on type conversion errors if the warnOnCoercionError flag is set to true
+func (rc *RowConverter) convert(ctx *sql.Context, inRow row.Row, warnOnCoercionError bool) (row.Row, error) {
```
zachmu, line 107: Little weird that ctx can be nil, but only if warn is false.

Another way to structure this would be for it to accept a `warnFn` which is optionally nil.

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -239,7 +239,7 @@ type commitInfo struct {
 }
 
 // Next returns the next row
-func (itr *diffRowItr) Next(*sql.Context) (sql.Row, error) {
+func (itr *diffRowItr) Next(sqlCtx *sql.Context) (sql.Row, error) {
```
zachmu, line 242: ctx is idiomatic

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -369,7 +376,60 @@ func (dp diffPartition) getRowIter(ctx *sql.Context, ddb *doltdb.DoltDB, ss *sch
...
+		sch:            joiner.GetSchema(),
+		fromCommitInfo: fromCmInfo,
+		toCommitInfo:   toCmInfo,
+	}, nil
+}
+
+// TODO: This is reused from commit_diff_table and we have likely changed it's behavior by not
+//       passing in super schema any more and changing this logic. Need to test and resolve this
+//       before releasing these changes.
+func (dp diffPartition) getRowIter(ctx *sql.Context, ddb *doltdb.DoltDB, joiner *rowconv.Joiner) (sql.RowIter, error) {
```
zachmu, line 395: dolt_diff and dolt_commit_diff should work the same with respect to their generated schemas. So you should just change it to match, either as part of this work or as a fast followup

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -452,9 +520,10 @@ type diffPartitions struct {
 	cmItr           doltdb.CommitItr
 	cmHashToTblInfo map[hash.Hash]tblInfoAtCommit
 	selectFunc      partitionSelectFunc
+	targetSch       schema.Schema
 }
 
-func newDiffPartitions(ctx *sql.Context, cmItr doltdb.CommitItr, wr *doltdb.RootValue, tblName string, selectFunc partitionSelectFunc) (*diffPartitions, error) {
+func newDiffPartitions(ctx *sql.Context, cmItr doltdb.CommitItr, wr *doltdb.RootValue, tblName string, selectFunc partitionSelectFunc, targetSch schema.Schema) (*diffPartitions, error) {
```
zachmu, line 526: Probably want schema param before iter param

zachmu: Best order here: ctx, tbl, sch, itr, root

The general principle (after context being first) is that when you are passing both schema and data, pass them in that order. If you have metadata (like table name), that goes first.

You might consider a little more of a refactor here. The only reason `wr` is passed is to get the underlying dolt table from it. So you could just pass that into this method in place of the table name and omit the root param entirely.

go/libraries/doltcore/sqle/dtables/history_table.go
```diff
@@ -446,3 +446,27 @@ func calcSuperSchema(ctx context.Context, wr *doltdb.RootValue, tblName string)
 
 	return ss, nil
 }
+
+// creates a RowConverter for transforming rows with the given schema to this super schema.
```
zachmu, line 450: Begin golang method comments with the method name

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -565,6 +565,408 @@ var DoltMerge = []enginetest.ScriptTest{
...
+			},
+			{
+				Query: "SELECT to_pk, to_c2, from_pk, from_c2 FROM DOLT_DIFF_t WHERE TO_COMMIT=@Commit1 ORDER BY to_pk;",
+				Expected: []sql.Row{
+					{1, 3, nil, nil},
+					{4, 6, nil, nil},
+				},
+			},
+			{
+				Query: "SELECT to_pk, to_c2, from_pk, from_c2 FROM DOLT_DIFF_t WHERE TO_COMMIT=@Commit2 ORDER BY to_pk;",
```
zachmu, line 689: I would expect this test case to select c1

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -565,6 +565,408 @@ var DoltMerge = []enginetest.ScriptTest{
...
+				},
+			},
+			{
+				Query: "SELECT to_pk, to_c, from_pk, from_c, diff_type FROM DOLT_DIFF_t WHERE TO_COMMIT=@Commit3 ORDER BY to_pk;",
+				Expected: []sql.Row{
+					{100, 101, nil, nil, "added"},
+				},
+			},
+			{
+				// TODO: What's the best way to test show warnings? With a custom/non-scripted test?
```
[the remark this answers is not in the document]
zachmu: You can clear warnings in between different test cases so that you know which warnings came from which statements.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -565,6 +565,408 @@ var DoltMerge = []enginetest.ScriptTest{
...
+			{
+				Query: "SELECT to_pk, to_c2, from_pk, from_c2, diff_type FROM DOLT_DIFF_t WHERE TO_COMMIT=@Commit4 ORDER BY to_pk;",
+				Expected: []sql.Row{
+					{100, 101, nil, nil, "added"},
+				},
+			},
+		},
+	},
+	{
+		Name: "primary key change",
```
zachmu, line 932: These are great tests, but you're missing a large and important bit of surface area: asserting that `dolt_diff AS OF @commit1` has the schema the table at that commit. And should include data changes after that schema change, but in terms of the old schema.

go/libraries/doltcore/sqle/tables.go
zachmu: @andrew-wm-arthur We have talked about this off and on. You ok with dropping a column touching every row in storage?

zachmu (commented):
This looks pretty good. Main comments are related to some missing tests. Nice work!

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: Interesting... should `select * from dolt_diff_$table as of @Commit1` work? I tried this while testing and skipped it because it didn't seem like it was supported. I just tested with my non-dev dolt build from homebrew (version 0.35.10) and couldn't get as of to work with `dolt_diff_$table` either. 

I had assumed that using `as of @Commit1` would work as if @Commit1 were checked out – i.e. none of the diff data more recent than @Commit1 would be included. It sounds like you think it should just use the old schema version, but try to fit all the current data up to HEAD into it?

go/libraries/doltcore/rowconv/row_converter.go
fulghum: oooh... great idea about `warnFn`. That made the code much cleaner! Thanks!

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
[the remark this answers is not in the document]
fulghum: I extended the ScriptTest test utils in go-mysql-server so that we can express more expectations about warnings: 
https://github.com/dolthub/go-mysql-server/pull/810

Once that GMS PR is merged in and available to Dolt, I'll push the update for these tests to use those new members.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
zachmu: Right, the AS OF clause only specifies the schema for the diff, the diff table's data always has the entire diff history of the repository. Otherwise you can only work backwards from an older schema, which is very limiting.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
zachmu: Right, see my other comment about AS OF for diff tables.

We want tests that demonstrate that dropped columns are shown as NULL in the diff table.

go/libraries/doltcore/sqle/tables.go
andrew-wm-arthur: We've changed the implementation here a few times, people have different opinions. It makes sense that it's causes incorrect diff output here.

Paying the table scan cost here isn't strictly necessary, but this is the direction we're moving toward with the new format, so I think we should do it here too

andrew-wm-arthur:
LGTM!

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: Gotcha. Currently no system tables support `as of`. I'm happy to pick up that work – I see how the `GetTableInsensitiveAsOf` code is going around the code where we instantiate system tables. I'd prefer to tackle that in a separate PR if you're cool with that though.

fulghum: Chatted with Zach yesterday and confirmed this plan – I'll tackle `as of` support for system tables in a separate PR and come back to covering those cases in these tests.

PR #2836, opened 2022-02-19
go/store/nbs: Add UpdateManifestAndClose

druvv:
In some cases, it may be desirable to add tables files to the manifest without opening the new table files as chunkSources. For large clone operations, it can be problematic to load the table file chunkSources as this causes their indices to be loaded into memory. In order to reduce the memory overhead of a clone, this PR adds UpdateManifestAndClose.

UpdateManifestAndClose adds table files to the manifest file, persists it, and then closes the store as it now outdated. It will not call `Rebase`, unlike `UpdateManifest`.

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -203,6 +203,30 @@ func (nbs *NomsBlockStore) GetChunkLocations(hashes hash.HashSet) (map[hash.Hash
 	return ranges, nil
 }
 
+// UpdateManifestAndClose updates the manifest and closes the block store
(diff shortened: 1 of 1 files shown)


reltuk (commented):
This generally looks OK to me, but I think we need a good understanding of what changes we need to make to dolthubapi and doltremoteapi in order to make changes here that will be fruitful.

As currently designed, I think it will be hard to coordinate the `Close()` of the cached-and-shared nbs.Store() instance in doltremoteapi, for example. And the current CloneRepository code path has a bigger API surface area than this method (UpdateManifestWithAppendix, SetRootChunk...).

druvv:
Will open separate PR with new changes

PR #2840, opened 2022-02-21
go/dolt/cmd: Use download language when dolt pulls from a remote

### LINKED ISSUE #2798: Dolt fetch, pull, etc. use "upload" language
All dolt remote operations currently output status as "uploading" data, regardless of the actual direction of the transfer.

druvv:
Fixes #2798

For downloads (pulling chunks from a remote db to a local db), dolt will now report the number of tables files written to disk as `Files Written`. It will no longer report an `Upload Rate` for a download as that is the speed at which the file is copied from a temp directory.

For uploads (pulling chunks from a local db to a remote db), dolt will now report the number of table files created during the pull process as `Files Created` and the number of these tables files that have been uploaded as `Files Uploaded`.

### DIFF
--- a/go/cmd/dolt/commands/push.go
+++ b/go/cmd/dolt/commands/push.go
@@ -112,7 +112,7 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
 	}
 
 	var verr errhand.VerboseError
-	err = actions.DoPush(ctx, dEnv.RepoStateReader(), dEnv.RepoStateWriter(), dEnv.DoltDB, dEnv.TempTableFilesDir(), opts, runProgFuncs, stopProgFuncs)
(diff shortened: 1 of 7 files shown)


go/cmd/dolt/commands/push.go
```diff
@@ -175,11 +175,11 @@ func (ts *TextSpinner) next() string {
...
 }
 
-func pullerProgFunc(ctx context.Context, pullerEventCh chan pull.PullerEvent) {
+func pullerProgFunc(ctx context.Context, pullerEventCh chan pull.PullerEvent, language progLanguage) {
 	var pos int
 	var currentTreeLevel int
 	var percentBuffered float64
-	var tableFilesBuffered int
+	var tableFilesClosed int
 	var filesUploaded int
```
zachmu, line 183: Suggest renaming this to filesTranferred

zachmu (approved):
LGTM

PR #2843, opened 2022-02-21
Expand DOLT_DIFF system table to indicate the type of change

fulghum:
This change adds two new boolean columns to the unscoped `DOLT_DIFF` system table to enable customers to determine if a change to a table was a schema change or a data change (or both). 

Example Usage: 
```
> create table x (a int primary key, b int, c int);
> create table y (a int primary key, b int, c int);
> insert into x values (1, 2, 3), (2, 3, 4);
> select DOLT_COMMIT('-am', 'Creating tables x and y');
> select * from dolt_diff;
+----------------------------------+------------+-----------+-------------------------+----------------------------------+-------------------------+-------------+---------------+
| commit_hash                      | table_name | committer | email                   | date                             | message                 | data_change | schema_change |
+----------------------------------+------------+-----------+-------------------------+----------------------------------+-------------------------+-------------+---------------+
| 1blnkur3m1hla2a4got9t513982a4ad1 | x          | jfulghum  | jason.fulghum@gmail.com | 2022-02-21 14:39:37.01 -0800 PST | Creating tables x and y | true        | true          |
| 1blnkur3m1hla2a4got9t513982a4ad1 | y          | jfulghum  | jason.fulghum@gmail.com | 2022-02-21 14:39:37.01 -0800 PST | Creating tables x and y | false       | true          |
+----------------------------------+------------+-----------+-------------------------+----------------------------------+-------------------------+-------------+---------------+
```

Resolves: https://github.com/dolthub/dolt/issues/2834

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
+++ b/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
@@ -34,6 +34,14 @@ type UnscopedDiffTable struct {
 	head *doltdb.Commit
 }
 
+// tableChange is an internal data structure used to hold the results of processing
+// a diff.TableDelta structure into the output data for this system table.
+type tableChange struct {
+	tableName    string
+	dataChange   bool
+	schemaChange bool
+}
+
 // NewUnscopedDiffTable creates an UnscopedDiffTable
 func NewUnscopedDiffTable(_ *sql.Context, ddb *doltdb.DoltDB, head *doltdb.Commit) sql.Table {
 	return &UnscopedDiffTable{ddb: ddb, head: head}
@@ -169,12 +177,6 @@ func (itr *UnscopedDiffTableItr) loadTableChanges(commit *doltdb.Commit) error {
 	}
 
 	return nil
-}
-
-type tableChange struct {
-	tableName    string
-	dataChange   bool
-	schemaChange bool
 }
 
 // calculateTableChanges calculates the tables that changed in the specified commit, by comparing that
--- a/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
+++ b/go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
@@ -15,10 +15,9 @@
 package dtables
 
 import (
+	"github.com/dolthub/go-mysql-server/sql"
 	"io"
 	"sort"
-
-	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/diff"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -97,9 +108,19 @@ func NewUnscopedDiffTableItr(ctx *sql.Context, ddb *doltdb.DoltDB, head *doltdb.
...
 	//   1) there is more than one commit left to process, or
-	//   2) the tableNames array isn't nilled out and has data to process
+	//   2) the tableChanges array isn't nilled out and has data left to process
+	return itr.commitIdx+1 < len(itr.commits) || itr.tableChanges != nil
+}
 
-	return itr.commitIdx+1 < len(itr.commits) || itr.tableNames != nil
+// IncrementIndexes increments the table changes index, and if it's the end of the table changes array, moves
+// to the next commit, and resets the table changes index so that it can be populated when Next() is called.
+func (itr *UnscopedDiffTableItr) IncrementIndexes() {
```
zachmu, line 117: No reason to export this, right?

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -189,39 +202,158 @@ func (itr *UnscopedDiffTableItr) calculateChangedTables(commit *doltdb.Commit) (
 		return nil, err
 	}
 
-	tablesMap := make(map[string]bool)
+	tableChanges := make([]tableChange, 0)
```
zachmu, line 205: When you know the length of the slice ahead of time, it's more idiomatic to declare it here (`len(deltas)`) and then do direct assignment in the loop (`tableChanges[i] = ...`) rather than use append. This avoids the cost of growing the slice and backing array several times.

Also with golang range loops, if you're using `i` you just want to use the simpler form and not declare `delta` as well.

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -189,39 +202,158 @@ func (itr *UnscopedDiffTableItr) calculateChangedTables(commit *doltdb.Commit) (
...
+	if len(tableChanges) == 0 {
 		return nil, nil
 	}
 
-	tables := make([]string, len(tablesMap))
-	i := 0
-	for key := range tablesMap {
-		tables[i] = key
-		i++
+	sort.Slice(tableChanges, func(i, j int) bool {
```
zachmu, line 220: Generally shouldn't do this unless the order is unstable. We should generally guarantee stable order for multiple executions of the same query but no particular defined order. Let the customer do an ORDER BY if they care about the order of results. If not this is wasted work.

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -189,39 +202,158 @@ func (itr *UnscopedDiffTableItr) calculateChangedTables(commit *doltdb.Commit) (
...
 }
 
 // Close closes the iterator.
 func (itr *UnscopedDiffTableItr) Close(*sql.Context) error {
 	return nil
 }
+
+// isTableDataEmpty return true if the table does not contain any data
+func (itr *UnscopedDiffTableItr) isTableDataEmpty(table *doltdb.Table) (bool, error) {
+	rowData, err := table.GetNomsRowData(itr.ctx)
```
zachmu, line 296: Probably want to use GetRowData here (storage-agnostic version of this) if possible

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -517,8 +517,8 @@ func TestUnscopedDoltDiffSystemTable(t *testing.T) {
 	for _, test := range UnscopedDiffTableTests {
 		databases := harness.NewDatabases("mydb")
 		engine := enginetest.NewEngineWithDbs(t, harness, databases)
-		engine.Analyzer.Debug = true
-		engine.Analyzer.Verbose = true
+		//engine.Analyzer.Debug = true
```
zachmu, line 520: Probably should delete these lines

zachmu (approved):
LGTM nice work

PR #2844, opened 2022-02-21
go/store/prolly: Faster table scans

andrew-wm-arthur:


go/store/prolly/node_cursor.go
```diff
@@ -390,6 +382,33 @@ func (cur *nodeCursor) copy(other *nodeCursor) {
 	}
 }
 
+func compareCursors(left, right *nodeCursor) int {
```
zachmu, line 385: Could use a method comment (as well as one on compare)

go/store/prolly/node_store.go
```diff
@@ -38,6 +38,8 @@ type NodeStore interface {
 	// Write writes a prolly tree Node to the store.
 	Write(ctx context.Context, nd Node) (hash.Hash, error)
 
+	FetchMany(ctx context.Context, refs hash.HashSet) error
```
zachmu, line 41: Method doc

zachmu (approved):
LG, just a couple funcs that were non obvious

PR #2857, opened 2022-02-22
Removing use of SuperSchema from Dolt_Blame system view.

fulghum:
Quick change to remove the SuperSchema reference from `Dolt_Blame` system view, now that we have migrated `Dolt_DIff_$tablename` off of SuperSchema's tag-suffixed column names.


zachmu (approved):
LGTM

PR #2862, opened 2022-02-22
[no-release-notes] go/cmd/dolt: Properly seed the global source in math/rand at dolt start.

reltuk:


zachmu (approved):
I GUESS this is kind of important

PR #2864, opened 2022-02-22
doltcore/sqle: Optimize Delete

andrew-wm-arthur:
Fetching and updating auto increment values is relatively expensive given the current implementation. Conditionally skipping this step makes deleted 2x faster


zachmu (approved):
LGTM

PR #2867, opened 2022-02-23
go/store/nbs: Add OverwriteStoreManifest to nbs package

druvv:


go/store/nbs/store.go
```diff
@@ -390,6 +393,66 @@ func fromManifestAppendixOptionNewContents(upstream manifestContents, appendixSp
...
+	updatedContents, err := nbs.mm.Update(ctx, originalLock, contents, nbs.stats, nil)
+	if err != nil {
+		return manifestContents{}, err
+	}
+
+	if updatedContents.lock != contents.lock {
+		return manifestContents{}, ErrConcurrentManifestWriteDuringOverwrite
+	}
+
+	err = nbs.Close()
```
reltuk, line 448: Why does this method need to close this Store? Doesn't this interact poorly with nbs cache in doltremoteapi?

go/store/nbs/store.go
```diff
@@ -390,6 +393,66 @@ func fromManifestAppendixOptionNewContents(upstream manifestContents, appendixSp
...
+	defer nbs.mu.Unlock()
+
+	var originalLock addr
+	{
+		_, c, err := nbs.mm.Fetch(ctx, nbs.stats)
+		if err != nil {
+			return manifestContents{}, err
+		}
+		originalLock = c.lock
+	}
```
reltuk, line 419: This seems wrong to me. If we are overwriting the manifest, not appending to it, then we should concurrency control on `nbs.upstream.lock`. We don't need to fetch the current manifest from manifest manager; we need to use our upstream's view to know if we failed the optimistic write.

go/store/nbs/store.go
druvv: While doltremoteapi's `ChunkStoreServiceServer` implementation does use an nbs cache, the `RespositoryServiceServer` does not. We can see [here](https://github.com/dolthub/ld/blob/90beb700bba591afcbc8a9872cabc1223971fa85/go/services/doltremoteapi/pkg/services/remoteapi_internal/repository.go#L233) that the repository service directly calls the factory when it retrieves an NBS while the chunkStore service uses the nbs cache [here](https://github.com/dolthub/ld/blob/90beb700bba591afcbc8a9872cabc1223971fa85/go/services/doltremoteapi/pkg/services/storesrv/chunk_store_service.go#L733-L735).

If this is also an idiomatic code concern, I can just mention that NBS will be out of date and leave it up to the caller to close the NBS.

reltuk: Thanks for pointing that out. I thought getStore went through nbs cache. This method still shouldn't close the store.

I will out a separate PR to fix doltremoteapi not closeing its stores in CloneRepository.

druvv:
Just pushed code that resolves the above. 

OverwriteManifest will not retry if it fails the optimistic write because we don't want any rebase to occur during OverwriteManifest. I can't think of situation where the manifest would be changed out of process for a clone, but just wanted to bring this up in case there is one.

go/store/nbs/store.go
reltuk: Is it possible to write this method as a function in the `nbs` package instead?

```go
// OverwriteStoreManifest is a low level interface to completely replace the manifest contents
// of |store| with the supplied |root|, |tableFiles| and |appendixTableFiles|. It performs concurrency
// control on the existing |store| manifest, and can fail with |ErrConcurrentManifestWriteDuringOverwrite|
// if the |store|'s view is stale. If contents should be unconditionally replaced without regard for the existing
// contents, run this in a loop, rebasing |store| after each failure.
//
// Regardless of success or failure, |OverwriteStoreManifest| does *not* Rebase the |store|. The persisted
// manifest contents will have been updated, but nothing about the in-memory view of the |store| will reflect
// those updates. If |store| is Rebase'd, then the new upstream contents will be picked up.
//
// Extreme care should be taken when updating manifest contents through this interface. Logic typically
// assumes that stores grow monotonically unless the |gcGen| of a manifest changes. Since this interface
// cannot set |gcGen|, callers must ensure that calls to this function grow the store monotonically.
func OverwriteStoreManifest(ctx context.Context, store *NomsBlockStore, root hash.Hash, tableFiles map[hash.Hash]uint32, appendixTableFiles map[hash.Hash]uint32) error {
...
}

druvv: fixed

reltuk (approved):
LGTM!

PR #2868, opened 2022-02-23
[no-release-notes] go/store/datas: Make passed in Dataset for Commit assert the head of the current ref.

reltuk:


andrew-wm-arthur (approved):
LGTM!

PR #2875, opened 2022-02-24
fix null value being exported for any types into .sql file

jennifersp:
the issue covered all typeInfo types where null values were not getting exported correctly. 
the fix is to add `NULL` whenever the value is nil and cannot be converted into sql types.


zachmu (approved):
LGTM, thanks for the quick turnaround

PR #2876, opened 2022-02-24
[no-release-notes] go/libraries/doltcore/dbfactory: file.go: Fix the default NBF Format used when constructing the old gen to match the new gen.

reltuk:
This fixes the panic when using very old repositories (7.18) or experimental formats in new ones.


zachmu (approved):
You da man

PR #2888, opened 2022-02-28
don't loop through all fks, if change doesn't impact any fk columns

JCOR11599:


go/libraries/doltcore/sqle/writer/sessioned_table_editor.go
```diff
@@ -362,14 +362,14 @@ func (ste *sessionedTableEditor) handleReferencingRowsOnUpdate(ctx context.Conte
...
 		}
-		referencingRows, err := editor.GetIndexedRows(ctx, referencingSte.tableEditor, indexKey, foreignKey.TableIndex, idxSch)
-		if err != nil {
-			return err
-		}
-		if len(referencingRows) == 0 {
-			continue
+
+		// make dictionary of primary key tags
+		pkTags := make(map[uint64]bool)
```
zachmu, line 367: Remove this block, now unused

go/libraries/doltcore/sqle/writer/sessioned_table_editor.go
```diff
@@ -362,14 +362,14 @@ func (ste *sessionedTableEditor) handleReferencingRowsOnUpdate(ctx context.Conte
...
+		for _, tag := range idxSch.GetPKCols().Tags {
+			pkTags[tag] = true
 		}
 
+		// Detect if there are changes, and if those changes impact a primary key columns
```
zachmu, line 372: Second part of comment no longer accurate

zachmu (approved):
Just a couple changes otherwise good to check in

PR #2894, opened 2022-03-01
Adds FixedStrategy which is a hedging strategy with a fixed hedge duration

druvv:
`FixedStrategy` can be used to hedge requests that should be hedged based on a fixed duration.


druvv:
will put in LD instead

PR #2901, opened 2022-03-01
Edit Stats display in Import and fix stats computation

VinaiRachakonda:
This pr

1. Fixes the formatting of stats and allows commas to be printed
2. Fixes an error with stats computation where many queries were getting marked as modifications.
3. Bumps GMS to add in a fix with dropped insert ignore errors


andrew-wm-arthur (approved):
LGTM

PR #2930, opened 2022-03-04
Prevent Multiple Not-NULL Constraints

JCOR11599:
Removed code where we would add an additional `NOT NULL` constraint when dropping primary keys.

Should be impossible to add multiple `NOT NULL` constraints to a single column, as there are checks right before anywhere we append a `NOT NULL` for if there already is `NOT NULL` constraint.

Added a method in schema marshalling that'll remove all duplicate not null constraints.


go/libraries/doltcore/schema/alterschema/addpk.go
```diff
@@ -52,10 +52,11 @@ func AddPrimaryKeyToTable(ctx context.Context, table *doltdb.Table, tableName st
 	for ord, col := range sch.GetAllCols().GetColumns() {
 		if i, ok := pkColOrdering[col.Name]; ok {
 			pkOrdinals[i] = ord
-			col.IsPartOfPK = true
+			// Can't assign IsPartOfPK true, before checking if IsNullable, since it'll always be false
```
andrew-wm-arthur, line 55: good catch here, I would go a step further and add a helper method like `addSingleNotNullConstraint()` that adds not null constraints iff it doesn't exist, and then validates that only one exists. 
Schema modification operations have proven to be very fragile, so I think we should be extra explicit when we make them and add extra validation as backup

go/libraries/doltcore/schema/alterschema/droppk.go
```diff
@@ -38,10 +38,11 @@ func DropPrimaryKeyFromTable(ctx context.Context, table *doltdb.Table, nbf *type
 
 	// Modify the schema to convert the primary key cols into non primary key cols
 	newCollection := schema.MapColCollection(sch.GetAllCols(), func(col schema.Column) schema.Column {
-		if col.IsPartOfPK {
-			col.Constraints = append(col.Constraints, schema.NotNullConstraint{})
-		}
 		col.IsPartOfPK = false
+		// TODO: should be impossible for column to have been a primary key, but not have a NOT NULL constraint, do I still bother to check?
```
andrew-wm-arthur, line 42: you could apply the method here as well, even just as a sanity check

go/libraries/doltcore/schema/constraint.go
```diff
@@ -92,15 +92,47 @@ func IndexOfConstraint(constraints []ColConstraint, constraintType string) int {
 
 // ColConstraintsAreEqual validates two ColConstraint slices are identical.
 func ColConstraintsAreEqual(a, b []ColConstraint) bool {
-	if len(a) != len(b) {
+	// TODO: do something with maps to avoid all duplicate constraint issues, and order issues too
```
andrew-wm-arthur, line 95: I would revert this change. We should have a well-formed schema by the time we're working with it. Let's apply the validation/cleaning earlier

go/libraries/doltcore/schema/encoding/schema_marshaling.go
```diff
@@ -316,6 +316,9 @@ func (sd schemaData) addChecksIndexesAndPkOrderingToSchema(sch schema.Schema) er
 func MarshalSchemaAsNomsValue(ctx context.Context, vrw types.ValueReadWriter, sch schema.Schema) (types.Value, error) {
 	// Anyone calling this is going to serialize this to disk, so it's our last line of defense against defective schemas.
 	// Business logic should catch errors before this point, but this is a failsafe.
+
+	schema.RemoveDuplicateNotNullColumnConstraints(sch.GetAllCols())
```
andrew-wm-arthur, line 320: Let's validate the schema as we write it, and clean the schema as we read it. Two reasons
- Validate-on-write forces us to make the correct operations as we're modifying the schema, rather than catching errors here at the last step. 
- Clean-on-read allows us to correct the mistakes that currently exists in any Dolt databases "in the wild".

andrew-wm-arthur (commented):
Good find. Added some suggestions to create an extra layer of protection against future schema change bugs

PR #2933, opened 2022-03-04
go/utils/remotesrv: use io.Copy in the http file server to write table file data to http.ResponseWriter

druvv:
## Problem

This PR fixes an issue when pulling chunks from Dolt `remotesrv`. The symptom is a `throughput below minimum allowable` error. The problem is that the dolt client can't handle large http chunks without failing to meet its minimum required download throughput. 

Dolt `remotesrv` hosts a simple http file server that can be used to write new table files and read existing table files. When `remotesrv` receives a request to read a file it will currently read all of the requested range into memory. It then, writes the 
entire range to `http.ResponseWriter` in a single `Write` call. 

This causes a very large http chunk ([see chunked transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding)) to be sent over the wire. When the Dolt client tries to read the http response using `Read`, it takes a long time to return. Due to the way the Dolt client measures `Read` throughput, if a single `Read` takes long enough, the minimum required throughput will not be met.

## Solution
We fixed this issue by using `io.Copy` to read from the file and simultaneously write to the `http.ResponseWriter`. `io.Copy` produces small enough writes to the response writer. This reduces the number of bytes read during a singular `Read` call on the Dolt client and allows the throughput measurement to be accurate.

We could have fixed this problem by changing the way the minimum required throughput was measured, but the above change reduces the memory overhead of `remotesrv` and uses an idiomatic Golang pattern. 

## Why doesn't this issue appear when pulling chunks from S3?
In Golang, `http.Response.Body` maximum `Read` sizes are determined by the `Content-Length` and whether chunked transfer encoding is being used. The exact semantics can be investigated by reading the sources here:
- [net/http parseTransferEncoding](https://cs.opensource.google/go/go/+/master:src/net/http/transfer.go;drc=master;bpv=1;bpt=1;l=624?gsn=parseTransferEncoding&gs=kythe%3A%2F%2Fgolang.org%3Flang%3Dgo%3Fpath%3Dnet%2Fhttp%23method%2520transferReader.parseTransferEncoding)
- [net/http chunkedReader](https://github.com/golang/go/blob/master/src/net/http/internal/chunked.go#L70)
- [net/http chunkedWriter](https://github.com/golang/go/blob/master/src/net/http/server.go#L368)

Since S3 returns a Content-Length and does not use chunked transfer encoding each `Read` call returns in a small enough size.

### DIFF
--- a/go/utils/remotesrv/http.go
+++ b/go/utils/remotesrv/http.go
@@ -90,14 +90,6 @@ func readTableFile(logger func(string), org, repo, fileId string, respWr http.Re
 			r, fileErr = getFileReaderAt(path, offset, length)
 		}
 	}
-	defer func() {
-		err := r.Close()
-		if err != nil {
-			err = fmt.Errorf("failed to close file at path %s: %w", path, err)
-			logger(err.Error())
-		}
-	}()
-
 	if fileErr != nil {
 		logger(fileErr.Error())
 		if errors.Is(fileErr, os.ErrNotExist) {
@@ -107,6 +99,13 @@ func readTableFile(logger func(string), org, repo, fileId string, respWr http.Re
 		}
 		return http.StatusInternalServerError
 	}
+	defer func() {
+		err := r.Close()
+		if err != nil {
+			err = fmt.Errorf("failed to close file at path %s: %w", path, err)
+			logger(err.Error())
+		}
+	}()
 
 	logger(fmt.Sprintf("opened file at path %s, going to read %d bytes", path, readSize))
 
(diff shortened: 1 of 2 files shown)


go/utils/remotesrv/http.go
```diff
@@ -70,6 +68,63 @@ func ServeHTTP(respWr http.ResponseWriter, req *http.Request) {
...
+		return http.StatusInternalServerError
+	}
+	if n != readSize {
+		logger(fmt.Sprintf("wanted to write %d bytes from file (%s) but only wrote %d", readSize, path, n))
+		return http.StatusInternalServerError
+	}
+
+	logger(fmt.Sprintf("wrote %d bytes", n))
+
+	return -1
```
reltuk, line 125:
```suggestion
	return http.StatusOK
```

reltuk (approved):
Looks great. One comment.

PR #2946, opened 2022-03-08
go/libraries/doltcore/dbfactory: aws.go: Support / as a character in database names for AWS remotes.

reltuk:
This was previously forbidden because of how table files were stored in the
destination bucket, but that was changed long ago and this is now well
supported. Our internal remotes serving infrastructure uses chunk stores
constructed in the same way.


andrew-wm-arthur (approved):
LGTM!

PR #2961, opened 2022-03-11
Fix error muting by defer and excessive memory allocation in `newBSChunkSource`. Fix `localbs` scheme using `file` scheme.

### LINKED ISSUE #2960: Runtime error: invalid memory address or nil pointer dereference error when cloning a Dolt repository
Seeing this issue with the most recent dolt versions, when downgrading to a previous version it is able to clone the repository without any issues.

Full error here:
Screenshot_2022-03-07_at_12 45 13

druvv:
Root issue is that `blobstore.GetBytes` returns a byte buffer whose capacity is too large. That buffer is passed to `parseTableIndex` which takes ownership of that buffer.

In Dolt v0.37.0, `parseTableIndex` was changed to throw an error if the capacity or length of the given buffer is too large. This was done to prevent large memory usage by Dolt. Since the above buffer was too large, parseTableIndex would throw an error and the chunkSource would fail to initialize. Two other issues prevented this error from being caught.

This bug was not caught prior to Dolt v0.37.0's release because errors returned in `newBSChunkSource` were being overridden by a defer. This muted any errors returned by `newBSChunkSource` including the error that `parseTableIndex` was throwing.

Additionally, the bats tests for the `localbs` scheme were using the `file` scheme under the hood. This prevented the tests from catching this bug.

This PR fixes the above issues and closes #2960


timsehn:
As discussed in standup, worth batsing this path?

reltuk (approved):
LGTM!

druvv:
> As discussed in standup, worth batsing this path?

Definitely. Was going to include as separate PR, but will include in this PR. We can bats test the local version of the blobstore easily which would have caught this bug.

timsehn:
Perfect. Works for me. Merge it and get a release out.

druvv:
@timsehn @reltuk 
Turns out we already have a bats test for `localbs`. Dolt was replacing the `localbs` scheme with the `file` scheme, so that bats test wasn't testing anything useful.

I fixed the scheme issue, checkout out main, and confirmed that the bats test failed. Then I checked out this branch and confirmed that the bats test completes successfully.

PR #2966, opened 2022-03-12
go/libraries/doltcore/sqle: Remove "Detached Head" mode

andrew-wm-arthur:
This PR removes support for "detached-head" mode in the SQL context and modifies surrounding functionality to restrict state changes in the SQL context. The central change here is that all SQL writes must be made against a working set and all writes must be made transactionally. With the exception of read-only sessions, all SQL sessions will have an associated working set.

**Currently**, SQL sessions can modify their state through special session variables:
- `myDatabase_head_ref`: essentially a branch change, changes the current head and also switches to the associated working set.
- `myDatabase_head`: similar to the `_head_ref` variable, but more general in that any commit can be used as a new head. Setting this variable puts the session into *detached-head* mode. Writes to this session are not transactionally written to a working set. Instead, they are written as free-floating values (eligible to be GC'd) and commits made in detached-head mode are "dangling" commits: they are not associated with a branch (and also eligible to be GC'd). A typical pattern for using this mode is to set the database head, make some writes and commits, and land this work by writing into the `dolt_branches` table. 
- `myDatabase_working`: setting this variable only affects the working set of session, essentially force-setting it to some other state.

**This PR** restricts the use of these variables in the following way: 
- `myDatabase_head_ref`: if the new value is a branch or working set ref, the session will switch working sets, other refs will error.
- `myDatabase_head`: ~~the new value will be resolved to a commit. If the commit corresponds to a single branch, the session will switch to the associated working set. If the commit cannot be resolved or doesn't correspond to a unique branch, an error will be returned.~~  **edit**: this variable is now read-only
- `myDatabase_working`: ~~similar to `_head`, if the new value can be resolved to a unique working set, the session will switch to it, otherwise it will error.~~ **edit**: this variable is now read-only


go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -836,6 +836,23 @@ func (ddb *DoltDB) GetRefsOfType(ctx context.Context, refTypeFilter map[ref.RefT
 	return refs, err
 }
 
+func (ddb *DoltDB) VisitWorkingSets(ctx context.Context, visit func(r ref.WorkingSetRef, addr hash.Hash) error) error {
```
zachmu, line 839: Method comment

go/libraries/doltcore/doltdb/workingset.go
```diff
@@ -91,6 +91,61 @@ func (m MergeState) PreMergeWorkingRoot() *RootValue {
...
+			werks = append(werks, r)
+		}
+		return nil
+	})
+	if err != nil {
+		return ref.WorkingSetRef{}, err
+	}
+
+	if len(werks) < 1 {
+		err = fmt.Errorf("could not resolve working set from root hash (%s)", h.String())
```
zachmu, line 108: Not a root hash

You might consider using ErrWorkingSetNotFound instead here

go/libraries/doltcore/doltdb/workingset.go
```diff
@@ -91,6 +91,61 @@ func (m MergeState) PreMergeWorkingRoot() *RootValue {
...
+		return ref.WorkingSetRef{}, err
+	}
+	if len(werks) > 1 {
+		err = fmt.Errorf("root hash (%s) corresponds to multiple working sets (%v)", h.String(), werks)
```
zachmu, line 112: Same comment, not a root hash

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -875,76 +845,82 @@ func (sess *Session) setForeignKeyChecksSessionVar(ctx *sql.Context, key string,
 	return sess.Session.SetSessionVariable(ctx, key, value)
 }
 
-func (sess *Session) setWorkingSessionVar(ctx *sql.Context, value interface{}, dbName string) error {
-	valStr, isStr := value.(string) // valStr represents a root val hash
-	if !isStr || !hash.IsValid(valStr) {
-		return doltdb.ErrInvalidHash
-	}
+func (sess *Session) resolveVersionKey(ctx *sql.Context, key, value string) (ws ref.WorkingSetRef, db string, err error) {
```
zachmu, line 848: I don't love the behavior here, where if there are two branches with the same HEAD commit (common), then setting db_head dies.

Needs to be fixed soon, preferably by making db_head read-only. Add some TODOs here

zachmu (approved):
I have some serious misgivings about the behavior when there is more than one working set / branch that matches a particular commit. This could actually break turbine on day 1. We need to walk them through how to migrate their code to using the head_ref variable and explain why we're doing it.

andrew-wm-arthur:
> I have some serious misgivings about the behavior when there is more than one working set / branch that matches a particular commit. This could actually break turbine on day 1. We need to walk them through how to migrate their code to using the head_ref variable and explain why we're doing it.

You make a good point. Setting these variables will fail depending on the state of a difference branch. It's safer to simply make them read only and (as you said) set the `_head_ref` variable instead.

PR #2967, opened 2022-03-12
Add Ephemeral Printer

### LINKED ISSUE #2964: Use a helper function to handle errors returned by functions in a defer statement?
I think Dolt should consider using some kind of linter or helper function to handle errors returned by functions in a defer statement. Debugging error logic when using defers that also return errors is tricky and can lead to found bugs such as #2961.

Consider the following function, can you spot the bug?
```go
func BuggyErrorHandling() (err error) {
	h, err := os.Open("test")
	defer func() {
		err = h.close()
	}()
	return err
}
```

The value of the returned err will always be the result of `h.Close()` even if `os.Open("test")` returns an error. So if `Open` returns an error, the returned error by `BuggyErrorHandling` will be nil if `Close` has no error! This is the correct code:
```go
func CorrectErrorHandling() (err error) {
	h, err := os.Open("test")
	defer func() {
		closeErr := h.Close()
		if err == nil {
			err = closeErr
		}
	}()
	return err
}
```

In #2961, the if statement inside the defer was mistyped as `if err != nil` which always overrode `os.Open`'s error! This lead to errors being dropped. Spotting this was tricky for me.

I propose the following helper:
```go
// MergeErrOf merges the error returned by |f| with |out|.
// Specifically, if |*out| is nil, set |*out| to be the error returned by |f|.
func MergeErrOf(f func() error, out *error) {
	fErr := f()
	if *out == nil {
		*out = fErr
	}
}
```

And usage:
```go
func EasyErrorHandling() (err error) {
	var h *os.File
	h, err = os.Open("test")
	defer MergeErrOf(h.Close, &err)
	return err
}
```

`MergeErrOf` needs to take a pointer to an error because the parameters of a deferred function are captured at the time of defer.

druvv:
`EphemeralPrinter` is tool that you can use to print temporary line(s) to the console. Every time `Display` is called, the previously written lines are cleared, and the new lines are flushed to the output.

Fixes #2964

### DIFF
--- a/go/cmd/dolt/cli/stdio.go
+++ b/go/cmd/dolt/cli/stdio.go
@@ -22,11 +22,11 @@ import (
 	"strings"
 	"sync/atomic"
 
-	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
-	"github.com/gosuri/uilive"
-
 	"github.com/fatih/color"
 	"github.com/google/uuid"
+	"github.com/gosuri/uilive"
+
+	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
 )
 
 var outputClosed uint64
--- a/go/cmd/dolt/cli/stdio.go
+++ b/go/cmd/dolt/cli/stdio.go
@@ -145,11 +145,12 @@ func PrintErrf(format string, a ...interface{}) {
 
 // EphemeralPrinter is tool than you can use to print temporary line(s) to the
 // console. Each time Display is called, the output is reset, and you can begin
-// writing anew. If you need to display multiple temporary lines, call Newline,
-// before writing each line.
+// writing anew.
 type EphemeralPrinter struct {
-	outW io.Writer
-	w    *uilive.Writer
+	outW         io.Writer
+	w            *uilive.Writer
+	wrote        bool
+	wroteNewline bool
 }
 
 // StartEphemeralPrinter creates a new EphemeralPrinter and starts it. You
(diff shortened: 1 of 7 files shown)


go/cmd/dolt/cli/stdio.go
```diff
@@ -141,6 +142,66 @@ func PrintErrf(format string, a ...interface{}) {
...
+func StartEphemeralPrinter() *EphemeralPrinter {
+	w := uilive.New()
+	w.Out = CliOut
+	e := &EphemeralPrinter{w, w}
+	e.start()
+	return e
+}
+
+// Printf formats and prints a string to the printer. You should not add
+// newlines in this string. If you need to display multiple lines, use Newline.
```
fulghum, line 165: How important is it to not add newlines? Does it screw up the display and line clearing? If so... you might want to consider returning an error or panic'ing if the caller has passed in a newline as an extra protection against than happening, instead of relying on people to read and follow the comments.

fulghum: Or... is it possible to detect the newline and automatically handle it correctly? i.e. by hiding the current Newline method so callers don't ever have to learn that interface?

fulghum: Continuing that thought about simplifying the caller's interface... is it helpful to combine the Printf method and Display? It seems like the calling code typically calls Printf, then (almost always) immediately calls Display. A small change, but might make the interface just a little bit easier to use if callers don't have to know there is a two step process to update text on screen.

fulghum:
Code looked fine to me; just a few thoughts on how we could simplify the interface and make it harder/impossible for callers to use it incorrectly. 

Any clever ways we can test this on linux and windows with BATS tests somehow?

fulghum (approved):

go/cmd/dolt/cli/stdio.go
druvv: It does screw up the display and line clearing! Good points, let me  add this.

druvv:
> Code looked fine to me; just a few thoughts on how we could simplify the interface and make it harder/impossible for callers to use it incorrectly.
> 
> Any clever ways we can test this on linux and windows with BATS tests somehow?

The trickiest part here is that we need to emulate terminal behavior to test this. I thought about mocking stdout with a file, but that won't emulate how a specific implementation responds to certain characters like backspace `\b`.  I couldn't come up with a trick unfortunately.

PR #2993, opened 2022-03-15
/go/libraries/doltcore/sqle: Remove temporary tables root value in favor of session-scoped tables

andrew-wm-arthur:


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -986,23 +960,31 @@ func (sess *Session) AddDB(ctx *sql.Context, dbState InitialDbState) error {
...
+			sess.tmpTables = append(sess.tmpTables[:i], sess.tmpTables[i+1:]...)
+			break
+		}
 	}
+}
 
-	dbState, _, err := sess.LookupDbState(ctx, dbName)
-	if err != nil {
-		return err
+func (sess *Session) GetTemporaryTable(ctx *sql.Context, name string) (sql.Table, bool) {
```
andrew-wm-arthur, line 976: According to the MySQL [docs](https://dev.mysql.com/doc/refman/8.0/en/create-temporary-table.html):
"TEMPORARY tables have a very loose relationship with databases (schemas)."

Should we add a database name param and check it?

andrew-wm-arthur: MySQL does, we should too

go/libraries/doltcore/sqle/temp_table.go
```diff
@@ -0,0 +1,410 @@
...
+
+func (t *TempTable) CreateForeignKey(
+	ctx *sql.Context,
+	fkName string,
+	columns []string,
+	referencedTable string,
+	referencedColumns []string,
+	onUpdate, onDelete sql.ForeignKeyReferenceOption,
+) error {
+	return errors.New("temporary tables do not support foreign keys")
```
zachmu, line 243: This probably belongs in the engine instead, add a TODO?

go/libraries/doltcore/sqle/temp_table.go
```diff
@@ -0,0 +1,410 @@
...
+}
+func (t *TempTable) DropForeignKey(
+	ctx *sql.Context,
+	fkName string,
+) error {
+	return errors.New("temporary tables do not support foreign keys")
```
zachmu, line 249: Same here

zachmu (approved):
LGTM

andrew-wm-arthur:
closing and reopening this PR to run CI :/

PR #2994, opened 2022-03-15
Change transactions per second to sql_transactions_per_second

VinaiRachakonda:
Fixes the column name of the result set for a particular benchmarking query.


coffeegoddd (approved):
LGTM

PR #2996, opened 2022-03-15
Add ORM Bats testing utilities

VinaiRachakonda:
This PR setups the basic bats setup for ORMs. Once we have an ORM passing end to end we can expose this to the CI with the relevant workflows.


coffeegoddd (approved):
Needs workflow?

fulghum (approved):
Looks like a good start to get into the ORM work. I'm still wondering how these tests scale out over time though... it seems like there are going to be a lot of ORMs we'll need to cover and keeping all their tests up to date and running could become pretty challenging. This is good work to keep finding gaps, but I think we should also keep brainstorming about other clever ways to test MySQL compatibility.

VinaiRachakonda:
@coffeegoddd I'll check in the workflow when one of these tests are correctly running end to end. That might take some time. I thought this was worth checking in now to get the ball rolling

VinaiRachakonda:
@fulghum I actually think this is the best way to get very thorough coverage. These ORMs are constantly adding MySQL features and created generated queries with correctness requirements. Seeing failures that happen in CI/CD can quickly inform us of missing feature gaps and compatibility gaps.

PR #2998, opened 2022-03-16
Added skipped test for buggy AS OF behavior in view definitions

zachmu:


timsehn (commented):
How would you be able to see the current version of the view with data from an old commit?

timsehn:
I think I like the current behavior. Otherwise, it's not easy to get it.

PR #3001, opened 2022-03-16
libraries/doltcore/dtestutils/testcommands: fix dropped errors

alrs:
This fixes several `err` variables that were set but then unhandled, as well as an `os.Chdir()` that looked a little risky without one.


zachmu (approved):

fulghum:
Thanks for the error handling contribution, Lars!

PR #3013, opened 2022-03-16
Apply PK ordinals to secondary index pk tag ordering

max-hoffman:
The original bug was found by stabs: https://github.com/dolthub/dolt/issues/2991
Refer to the GMS PR with a script test for testing the condition that triggers the bug: https://github.com/dolthub/go-mysql-server/pull/881

A summary of the bug is: Secondary indexes were caching old pk orderings. Schema changes created an inconsistency between secondary indexes and the primary keys they referred to, causing errors in index lookups. This only impacts DDL statements involving a new pk ordering and preexisting secondary indexes.

There are several layers of indirection between pk source row and secondary index row. When creating an index row, a higher layer of code reads and formats keys in the correct index ordering (for example, reading schema order (x,y,z) -> (y,z,x)). However, we deconstruct and rebuild that row before inserting values into a secondary index. The second constructor drops the ordering in favor of a hashmap of tags, and uses an index embedded tag ordering `pks` to recreate the row.

This fix extends `schema.SetPkOrdinals` to override the preexisting ordering of `pks` with an `indexCollection.SetPks` method.

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/dolthub/ishell v0.0.0-20220112232610-14e753f0f371
 	github.com/dolthub/mmap-go v1.0.4-0.20201107010347-f9f2a9588a66
 	github.com/dolthub/sqllogictest/go v0.0.0-20201107003712-816f3ae12d81
-	github.com/dolthub/vitess v0.0.0-20220314213812-0ccb112f20e4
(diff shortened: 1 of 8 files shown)


andrew-wm-arthur (approved):
Good work with the testing here. 
We might be too far down the current road, but I wonder if it wouldn't better better for the schema to reflect physical order and store display order out-of-band, rather than storing physical order out-of-band. In either case, we weren't taking this step and updating all primary key locations. Mirroring state is tricky!

PR #3019, opened 2022-03-16
[no-release-notes] go: store/datas,doltdb: Decouple some Commit accesses and implementations from types.

reltuk:
Here we start decoupling `doltdb.Commit` from `types.Struct` so that we can replace it with a different value in the new implementation. This PR includes:

1) Add datas.Dataset.MaybeHeadAddr() and move a lot of MaybeHeadRef() uses to it.
2) Remove some unused code, like types/ref_heap.go.
3) Create some accessors for Commit values in `datas`.
4) Have `doltdb.Commit` materialize all its fields when it is constructed, instead of lazily. Includes Parents(), RootValue(), StRef().
5) Do not store the `commitSt` in `doltdb.Commit` itself.

An alternative approach would be to store `commitVal` instead of eagerly fetching the fields from the passed in value. I'm a little partial to not carrying around the `types.Value`, but could be convinced otherwise.


reltuk:
#benchmark

go/libraries/doltcore/doltdb/commit.go
```diff
@@ -25,111 +24,49 @@ import (
...
 
 // Commit contains information on a commit that was written to noms
 type Commit struct {
-	vrw      types.ValueReadWriter
-	commitSt types.Struct
-	parents  []types.Ref
+	vrw     types.ValueReadWriter
+	meta    *datas.CommitMeta
+	parents []types.Ref
+	stref   types.Ref
```
andrew-wm-arthur, line 35: it seems a little weird for a commit to store a ref to itself, although it is a succinct way to check equality, etc

go/libraries/doltcore/doltdb/workingset.go
```diff
@@ -270,7 +273,17 @@ func (ws *WorkingSet) writeValues(ctx context.Context, db *DoltDB) (
...
+		if err != nil {
+			return types.Ref{}, types.Ref{}, nil, err
+		}
+		commitV, err := db.vrw.ReadValue(ctx, commitH)
+		if err != nil {
+			return types.Ref{}, types.Ref{}, nil, err
+		}
+
+		// TODO: Remove types.Struct coerce here.
+		mergeStateRefSt, err := datas.NewMergeState(ctx, preMergeWorking, commitV.(types.Struct))
```
andrew-wm-arthur, line 286: wow this is all so tangled!

andrew-wm-arthur (approved):
LGTM, getting commit val out of `Commit` makes sense to me

go/libraries/doltcore/doltdb/workingset.go
reltuk: It's a mess! :P

Issue #3029, opened 2022-03-17
Support server side cursors


VinaiRachakonda:
Recently the [MySQL FDW](https://github.com/EnterpriseDB/mysql_fdw) added the use of the [mysql_stmt_store_result](https://dev.mysql.com/doc/c-api/8.0/en/mysql-stmt-store-result.html) function. This requires [server side cursors](https://dev.mysql.com/doc/refman/8.0/en/cursor-restrictions.html#:~:text=A%20server%2Dside%20cursor%20enables,remaining%20rows%20are%20not%20transferred.).

Until then we will support MySQL FDW up till [REL-2_7_0](https://github.com/EnterpriseDB/mysql_fdw/releases/tag/REL-2_7_0)

timsehn:
This may be related to: #5441

JCOR11599:
The dolt server now partially implements server side cursors.

Some caveats:
 - We only allow one server side cursor to be open at a time.
 - The results are not stored in a temporary table like in MySQL, but rather we have a goroutine that blocks spooling the results from `COM_STMT_EXECUTE`, and waits for `COM_STMT_FETCH`.

Regardless it seems like dolt works well enough with MySQL FDW 2.9.0

PR #3030, opened 2022-03-17
Unblock the failing MySQL FDW integration test

VinaiRachakonda:
This pr:

1. Switches to an older version of MySQL FDW due to #3029 
2. Adds a comment in the c api to test an additional function
3. Adds logic in the mysql bats client tests that shuts the down the postgres server in the teardown. Will be useful in case new versions of the wrapper cause another breakage.


coffeegoddd (approved):
LGTM

PR #3041, opened 2022-03-18
go/cmd/dolt: Fix fetch printing newline for each progress update on Windows

druvv:
Fixes: https://github.com/dolthub/dolt/issues/2941

The above reported bug is caused by `EphemeralPrinter`. It's related to how the underlying library of `uilive` clears terminal lines on Windows. `uilive` will only clear terminal lines successfully if it detects that the given `io.Writer` is a terminal. The package does this by checking if the io.Writer has a `Fd` function and using the returned file descriptor for its `istty` check.

When color.Output is used as the `io.Writer`, `uilive`'s terminal check fails since that does not declare `Fd`. The problem is that color.Output does use the initial `os.StdOut` which is in fact the terminal... 

This was fixed by wrapping color.Output with a struct that reports the correct file descriptor.

This PR also improves the display output of ephemeral printer by controlling when `uilive` flushes its output.

### DIFF
--- a/go/cmd/dolt/cli/stdio.go
+++ b/go/cmd/dolt/cli/stdio.go
@@ -140,6 +140,11 @@ func PrintErrf(format string, a ...interface{}) {
 	fmt.Fprintf(CliErr, format, a...)
 }
 
+// DeleteAndPrint prints a new message and deletes the old one given the
+// previous messages length. It returns the length of the printed message that
+// should be passed as prevMsgLen on the next call of DeleteAndPrint.
+//
+// DeleteAndPrint does not work for multiline messages.
 func DeleteAndPrint(prevMsgLen int, msg string) int {
 	if outputIsClosed() {
 		return 0
@@ -186,6 +191,16 @@ type EphemeralPrinter struct {
 // NewEphemeralPrinter creates a new EphemeralPrinter.
 func NewEphemeralPrinter() *EphemeralPrinter {
 	w := uilive.New()
+	// How uilive needs to clear terminal output, depends on what StdOut is
+	// being used. It checks the running terminal by casting the provided writer
+	// to an internal interface that just defines a `Fd() uintptr` function. It
+	// uses the file descriptor returned by fd to check if the writer writes to
+	// a terminal.
+	//
+	// If we use |colorOutput|, the type cast will fail and uilive will fail to
+	// detect the output terminal even though colorOutput always points to
+	// StdOut. We provide the file descriptor by wrapping |colorOutput| in a
+	// struct.
 	if CliOut == colorOutput {
 		w.Out = fdProvider{CliOut, colorFd}
 	} else {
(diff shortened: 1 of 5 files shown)


go/cmd/dolt/commands/diff.go
```diff
@@ -956,8 +956,7 @@ func diffSummary(ctx context.Context, td diff.TableDelta, colLen int) errhand.Ve
 	acc := diff.DiffSummaryProgress{}
 	var count int64
 	var pos int
-	eP := cli.StartEphemeralPrinter()
-	defer eP.Stop()
+	eP := cli.NewEphemeralPrinter()
```
fulghum, line 959: Nice job removing the need to call `Start` and `Stop`. The EP interface looks cleaner now from that. Also looks nice to have the more standard "New<object>" pattern instead of "Start<object>".

go/cmd/dolt/cli/stdio.go
```diff
@@ -143,6 +140,37 @@ func PrintErrf(format string, a ...interface{}) {
 	fmt.Fprintf(CliErr, format, a...)
 }
 
+func DeleteAndPrint(prevMsgLen int, msg string) int {
```
fulghum, line 143: Add some godocs since this is an exported function

go/cmd/dolt/cli/stdio.go
```diff
@@ -143,6 +140,37 @@ func PrintErrf(format string, a ...interface{}) {
...
+	if outputIsClosed() {
+		return 0
+	}
+
+	msgLen := len(msg)
+	backspacesAndMsg := make([]byte, prevMsgLen+msgLen, 2*prevMsgLen+msgLen)
```
fulghum, line 149: Is `2*prevMsgLen+msgLen` the exact length or are just approximating to try and get close? It seems like it would be something more like prevMsgLen+msgLen+2*diff ?

fulghum (approved):
Code looks fine; just a couple of small suggestions to potentially tidy up a little more. I love that you were able to simplify the EphermeralPrinter interface a little bit more. 

I added a link to the associated issue in your PR description, too. Nice to have that linkage between the issue and PR. 

Have you already manually tested on both cygwin and standard terminal on Windows? I think it's worth creating a simple way to test this across platforms since we know this cross-platform terminal work can be finicky. The idea you suggested on friday with a hidden CLI command and BATS tests sounded pretty interesting.

druvv:
Thank you for the review! I have not tested on cygwin, but I did manually test the standard terminal. I will test on cygwin today.

I've been thinking about the tests a little more and I'm going to investigate further. I haven't found a solution that makes sense yet. I have some ideas on what to explore. I will get back to you!

go/cmd/dolt/cli/stdio.go
druvv: It's an estimate! That value is for the capacity of the buffer to prevent the `append` calls from needing to grow the capacity too often.

From go documentation:
> The append function appends the elements x to the end of the slice s, and grows the slice if a greater capacity is needed.

druvv:
Cygwin works as well!

PR #3042, opened 2022-03-19
Fix two bugs in foreign key creation

zachmu:
Fixed https://github.com/dolthub/dolt/issues/3023
Fixes https://github.com/dolthub/dolt/issues/3028


Hydrocharged (approved):
LGTM!

PR #3053, opened 2022-03-21
Simplify RebuildIndex

max-hoffman:
Two improvements with index rewriting:

1) Brian's `TupleFactory` optimizations were being skipped, and 70% of
the runtime was spent building `newBinaryNomsWriter` instances for tuple
construction. Removing this restores index inserts as the only memory
pressure.

2) Brian's `BulkEditAccumulator` lets the final Map sort skip
work materializing intermediate prolly trees.

On a 2000 row table, fix (1) reduces the runtime by 50% (80ms -> 40ms),
and fix (2) reduces the runtime by another 40% (40ms -> 10ms)%, for a
cumulative ~90% improvement.

### DIFF
--- a/go/libraries/doltcore/table/editor/index_editor.go
+++ b/go/libraries/doltcore/table/editor/index_editor.go
@@ -17,6 +17,7 @@ package editor
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"sync"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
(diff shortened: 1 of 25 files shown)


go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -287,7 +288,14 @@ func RebuildIndex(ctx context.Context, tbl *doltdb.Table, indexName string, opts
 		return types.EmptyMap, fmt.Errorf("index `%s` does not exist", indexName)
 	}
 
-	rebuiltIndexData, err := rebuildIndexRowData(ctx, tbl.ValueReadWriter(), sch, tableRowData, index, opts)
+	tf := tupleFactories.Get().(*types.TupleFactory)
+
+	tmpDir, err := ioutil.TempDir("", "tempif-*")
```
max-hoffman, line 293: pass down the appropriate tempDir from the sessionState? this fails on linux

go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -287,7 +288,14 @@ func RebuildIndex(ctx context.Context, tbl *doltdb.Table, indexName string, opts
 		return types.EmptyMap, fmt.Errorf("index `%s` does not exist", indexName)
 	}
 
-	rebuiltIndexData, err := rebuildIndexRowData(ctx, tbl.ValueReadWriter(), sch, tableRowData, index, opts)
+	tf := tupleFactories.Get().(*types.TupleFactory)
```
bheni, line 291: you are done with this TF instance when the functions exits right?  I'd put it back in the pool with a defer block here.

go/libraries/doltcore/table/editor/index_editor.go
```diff
@@ -314,8 +322,15 @@ func RebuildAllIndexes(ctx context.Context, t *doltdb.Table, opts Options) (*dol
 		return nil, err
 	}
 
+	tf := tupleFactories.Get().(*types.TupleFactory)
```
bheni, line 325: put back in the pool?

go/libraries/doltcore/table/editor/index_editor.go
bheni: or use the .dolt/temptf. I'd do whatever we do already for other bulk editors.  I can't remember exactly what we do.

bheni (approved):
Looks reasonable.

PR #3063, opened 2022-03-22
[no-release-notes] Adding tests for dolt_diff table function CheckPrivileges code

fulghum:
Following up on: https://github.com/dolthub/dolt/pull/2913 to add tests for CheckPrivileges logic

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -108,21 +108,21 @@ var DoltUserPrivTests = []enginetest.UserPrivilegeTest{
 		},
 		Assertions: []enginetest.UserPrivilegeTestAssertion{
 			{
-				// Without access to the database, dolt_diff should fail
+				// Without access to the database, dolt_diff should fail with a database access error
 				User:        "tester",
 				Host:        "localhost",
 				Query:       "SELECT * FROM dolt_diff('test', 'main~', 'main');",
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -94,6 +94,51 @@ var DoltScripts = []enginetest.ScriptTest{
...
+				Expected: []sql.Row{{1}},
+			},
+			{
+				// With access to the db, but not the table, dolt_diff should fail
+				User:        "tester",
+				Host:        "localhost",
+				Query:       "SELECT * FROM dolt_diff('test2', 'main~', 'main');",
+				ExpectedErr: sql.ErrPrivilegeCheckFailed,
+			},
+		},
```
Hydrocharged, line 138: I'd add a few more assertions. Let's revoke the `SELECT` privilege on `mydb.test`, and then grant a `SELECT` privilege on the `mydb` database and check that the function still works with the tables. Then we revoke the db-level privilege and grant a global `SELECT` privilege, just to make sure that each level works as expected.

Hydrocharged (approved):
Looks good! Just want a few more test assertions!

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: done deal; Thanks Daylon!

PR #3069, opened 2022-03-23
[no-release-notes] go/{store/prolly, libraries/doltcore/sqle}: Redefine `prolly.Range` to support all range queries

andrew-wm-arthur:
- Redefines `prolly.Range` to more closely match `sql.Range`. This is necessary to support a number of range lookup cases that cannot be defined using the current tuple bound implementation. 
- Implements a merge algorithm for `prolly.Range` to prevent duplicate rows/tuples being returned from range lookups. This is a notable difference from the current `NomsRangeReader` implementation which combines both range lookups and filtering. 
- Refactors utilities in pkg `val` to support value comparisons needed for `prolly.Range` logic.


Hydrocharged (approved):
LGTM!

PR #3073, opened 2022-03-23
Add Tests to ensure default values are correctly interpreted in the information_schema.columns table

jennifersp:
this PR fixes NULL and EMPTY values were not being converted correctly when converting from Dolt schema

previousy default value of NULL and "" (empty) was defined incorrectly
```
> dolt sql -q "SELECT column_name, is_nullable, column_default FROM information_schema.columns WHERE table_name = 'mytable'" -r csv
column_name,is_nullable,column_default
i,NO,""
s,YES,""""""
somecol,YES,""
col4,YES,""
``` 
for table with schema:
```
CREATE TABLE `mytable` (
  `i` bigint NOT NULL,
  `s` varchar(20) DEFAULT "",
  `somecol` int,
  `col4` bigint,
  PRIMARY KEY (`i`)
)
```


go/libraries/doltcore/sqle/sqlutil/convert.go
```diff
@@ -32,14 +33,26 @@ func FromDoltSchema(tableName string, sch schema.Schema) (sql.PrimaryKeySchema,
...
 		sqlType := col.TypeInfo.ToSqlType()
-		var extra string
-		if col.AutoIncrement {
-			extra = "auto_increment"
-		}
 
 		var deflt *sql.ColumnDefaultValue
 		if col.Default != "" {
-			deflt = sql.NewUnresolvedColumnDefaultValue(col.Default)
+			if col.Default == "NULL" {
```
VinaiRachakonda, line 39: I'm a little confused by this. Shouldn't this be in the engine in Eval() of an UnresolvedComment. Or another way to look at it is that are we doing the conversion of a SQL Schema to a Dolt Schema incorrectly?

go/libraries/doltcore/sqle/sqlutil/convert.go
```diff
@@ -32,14 +33,26 @@ func FromDoltSchema(tableName string, sch schema.Schema) (sql.PrimaryKeySchema,
...
 		sqlType := col.TypeInfo.ToSqlType()
-		var extra string
-		if col.AutoIncrement {
-			extra = "auto_increment"
-		}
 
 		var deflt *sql.ColumnDefaultValue
 		if col.Default != "" {
-			deflt = sql.NewUnresolvedColumnDefaultValue(col.Default)
+			// col.Default == "" represent nil value, and empty value is represented as empty-string value `""`
```
VinaiRachakonda, line 39: Change to 

```
The nil value can be represented in two ways. 1) "NULL" 2) "". Note that the default value of an empty string is double wrapped with `""`
```

go/libraries/doltcore/sqle/sqlutil/convert.go
```diff
@@ -32,14 +33,26 @@ func FromDoltSchema(tableName string, sch schema.Schema) (sql.PrimaryKeySchema,
...
+			// nil value also can be represented as string `"NULL"`
+			if col.Default == "NULL" {
+				deflt = nil
+			} else if col.Default == `""` {
+				deflt, err = sql.NewColumnDefaultValue(expression.NewLiteral("", sql.LongText), sqlType, true, col.IsNullable())
```
VinaiRachakonda, line 44: I don't Text is better than LongText

integration-tests/bats/default-values.bats
```diff
@@ -487,3 +487,25 @@ DELIM
...
+    dolt sql -q "CREATE TABLE test(pk BIGINT PRIMARY KEY, c1 BIGINT, c2 BIGINT, c3 INT)"
+    run dolt sql -q "SELECT column_name, is_nullable, column_default FROM information_schema.columns WHERE table_name = 'test'" -r=csv
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "column_name,is_nullable,column_default" ]] || false
+    [[ "$output" =~ "pk,NO," ]] || false
+    [[ "$output" =~ "c1,YES," ]] || false
+    [[ "$output" =~ "c2,YES," ]] || false
+    [[ "$output" =~ "c3,YES," ]] || false
+
+    dolt sql -q "ALTER TABLE test CHANGE c1 c1 INT NULL DEFAULT NULL"
```
VinaiRachakonda, line 501: Let's add more types here.

VinaiRachakonda (approved):
Good job catching this issue. Fwiw, I think @zachmu  should take a look at this PR as well. 

Overall, you are touching a very sensitive portion of the code base so I'd like to see more types being tested here.

VinaiRachakonda:
Attached bug with Tableplus: https://github.com/TablePlus/TablePlus/issues/2650

PR #3080, opened 2022-03-24
go/libraries/doltcore/dbfactory: Thread memory quota provider through store layer to onHeapTableIndex

druvv:
This PR allows creators of noms block stores to manage memory resources related to them.


go/store/nbs/table_set.go
```diff
@@ -474,6 +491,16 @@ func (ts tableSet) Rebase(ctx context.Context, specs []tableSpec, stats *Stats)
 						return
 					}
 					if spec.name == h {
+						memSize, err := getCSMemSize(existing)
+						if err != nil {
+							ae.SetIfError(err)
+							return
+						}
+						err = ts.q.AcquireQuota(ctx, memSize)
```
reltuk, line 499: This might work fine for now. But I'm concerned about the double-counting overhead when we're rebasing a large repository..AFAIU, the interaction is basically

```
newtableset, err := tableset.Rebase(...)
if err ...
tableset.Close()
tableset = newtableset
```

Also AFAIU, the mmap machinery already requires that explicitly manage the lifecycle of these index (`.Clone()`). So...one way to get around this double counting is by moving the ReleaseQuota call into the tableIndex `Close()` function (`Close(ctx, MemoryQuotaProvider) error`) and adding reference counting to `onHeapTableIndex` as well.

Any thoughts?

reltuk (approved):
Looks great. One comment :)

go/store/nbs/table_set.go
druvv: I think this makes a lot of sense. Intuitively, we are only trying to manage the memory of the table indices. 

I'll move it down another level.

druvv:
tested these Dolt changes against LD tests just to make sure I didn't break something

PR #3095, opened 2022-03-25
Fix bad upload table file retries

druvv:


bheni (approved):
Seems totally reasonable

PR #3114, opened 2022-03-29
[no-release-notes] Andy/prolly partitions and point lookups

andrew-wm-arthur:
- Implemented a special range lookup case for point-lookups. This is both more performant and necessary for join correctness
- Refactored new-format table partitions to use the same ordinal bounds as the current format. 
- Implemented ordinal range scans for `prolly.Map`
- Skipped panicking engine tests


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -43,15 +43,20 @@ func TestSingleQuery(t *testing.T) {
...
+			{0, 0, 1, 1},
+		},
 	}
 
 	harness := newDoltHarness(t)
 	engine := enginetest.NewEngine(t, harness)
 	enginetest.CreateIndexes(t, harness, engine)
-	engine.Analyzer.Debug = true
-	engine.Analyzer.Verbose = true
+	//engine.Analyzer.Debug = true
```
zachmu, line 58: Revert these two lines

go/store/prolly/map.go
```diff
@@ -168,9 +168,54 @@ func (m Map) IterAll(ctx context.Context) (MapRangeIter, error) {
...
 func (m Map) IterRange(ctx context.Context, rng Range) (MapRangeIter, error) {
-	return m.iterFromRange(ctx, rng)
+	if rng.isPointLookup(m.keyDesc) {
+		return m.pointLookupFromRange(ctx, rng)
+	} else {
+		return m.iterFromRange(ctx, rng)
+	}
+}
+
+func (m Map) pointLookupFromRange(ctx context.Context, rng Range) (*pointLookup, error) {
```
zachmu, line 204: Bothers me a little that we're doing this IO effectively at compile time rather than at runtime. This is a very fat constructor. It would be better to move the IO into the call to `iter.Next`

zachmu (approved):
LGTM, just a couple comments

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
andrew-wm-arthur: I refuse!

go/store/prolly/map.go
andrew-wm-arthur: That's a good shout, but I'm not sure it's true in practice. 
Essentially all table access is wrapped in a partition iterator which lazily constructs `RowIter`s for table scans, range queries, and joins. After the partitions row iter is requested, we typically perform IO eagerly.

PR #3121, opened 2022-03-29
Adding tests to validate as of support for describe table.

fulghum:
Enables Dolt to support `describe table as of <commit>` or `show columns from table as of <commit>`.

Depends on:
* Vitess support: https://github.com/dolthub/vitess/pull/150
* Go-mysql-server support: https://github.com/dolthub/go-mysql-server/pull/923


zachmu (approved):
These look great, what a cool new feature!

PR #3141, opened 2022-04-01
Adding tests for 'show create table as of <version>'

fulghum:
Adds support for `show create table as of <version>`.

Depends on:
* https://github.com/dolthub/go-mysql-server/pull/932
* https://github.com/dolthub/vitess/pull/151


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -510,14 +510,15 @@ func TestDoltScripts(t *testing.T) {
 	}
 }
 
-func TestDescribeTableAsOf(t *testing.T) {
-	// These tests rely on altering schema, and the new storage format doesn't support that yet,
-	// so we need to skip them. Once the new storage format supports altering schema, we can move
-	// these ScriptTests back into the DoltScripts var so they get picked up by the TestDoltScripts
-	// method above and then remove this method.
+func TestShowTableAsOf(t *testing.T) {
```
zachmu, line 513: Rather than combine these into the same test method, introduce a new one.

zachmu (approved):

PR #3167, opened 2022-04-06
fix panic from nil VRW

jennifersp:
Fixes avoiding panics when `readSequence()` functions is called on valueDecoder with nil VRW


go/store/types/value_decoder.go
```diff
@@ -181,7 +181,11 @@ func (r *valueDecoder) readSequence(nbf *NomsBinFormat, kind NomsKind, leafSkipp
 		return newMetaSequence(r.vrw, r.byteSlice(start, end), offsets, length), nil
 	}
 
-	return newLeafSequence(r.vrw, r.byteSlice(start, end), offsets, length), nil
+	vrw := r.vrw
```
Hydrocharged, line 184: IIRC, this did not produce the correct output right?

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -75,6 +74,27 @@ func (f *ForeignKeyViolationError) Error() string {
...
+func getRowValuesAsString(key types.Value, val types.Value) string {
+	valSlice, _ := val.(types.Tuple).AsSlice()
+	all, _ := key.(types.Tuple).Append(valSlice...)
+	allSlice, _ := all.AsSlice()
+	var rowVals []string
+	for i, s := range allSlice {
+		if i%2 == 1 {
+			switch v := s.(type) {
+			case types.Blob:
+				rowVals = append(rowVals, v.DebugText())
```
Hydrocharged, line 88: I'm not sure this is the best way to handle this. What does MySQL return when setting up this same scenario and using a MySQL client? `DebugText` seems like the wrong thing to use here.

jennifersp: MySQL client returns 
```
ERROR 1452 (23000): Cannot add or update a child row: a foreign key constraint fails (`tempdb`.`#sql-577f_a`, CONSTRAINT `fk_b_a_id_refs_a` FOREIGN KEY (`a_id`) REFERENCES `a` (`id`))
```

timsehn:
What's going on with this one @jennifersp ?

jennifersp:
> What's going on with this one @jennifersp ?

I dropped this last week to focus on dumps. I'll get this in this week. The issue I'm at is that not sure if it should match mysql error output, which does not include any detailed violating rows information only foreign key constraint that is getting violated, but our current error output prints all violating rows.

Hydrocharged:
We will keep the key information in the error message. But the key representation should match that of MySQL. If the error does not return the key, then compare the key from a `SELECT` statement instead.

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -61,18 +61,41 @@ func (f *ForeignKeyViolationError) Error() string {
...
 	} else {
-		return fmt.Sprintf("foreign key violations on `%s`.`%s`:%s", f.ForeignKey.Name, f.ForeignKey.TableName, sb.String())
+		return fKeyErrorMsg
+	}
+}
+
+// getRowValuesAsString returns row values as string format.
+// Variables 'key' and 'val' have both 'tags and values' added in them, so tags are removed
+func getRowValuesAsString(key types.Value, val types.Value) string {
+	valSlice, _ := val.(types.Tuple).AsSlice()
```
Hydrocharged, line 89: Rather than use `AsSlice` here, we should use `IterFields` instead, and just do one loop over the key and another over the value.

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -61,18 +61,41 @@ func (f *ForeignKeyViolationError) Error() string {
...
-		valSlice, _ := val.(types.Tuple).AsSlice()
-		all, _ := key.(types.Tuple).Append(valSlice...)
-		str, _ := types.EncodedValue(context.Background(), all)
+
+		str := getRowValuesAsString(key, val)
 		sb.WriteRune('\n')
 		sb.WriteString(str)
 	}
+
+	var refCols []string
```
Hydrocharged, line 70: These aren't referenced columns, these are just standard columns as you're using the `TableColumns` field. Referenced columns would be the `ReferencedTableColumns` field.

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -61,18 +61,41 @@ func (f *ForeignKeyViolationError) Error() string {
...
+	for _, tct := range f.ForeignKey.TableColumns {
+		col, _ := f.Schema.GetAllCols().GetByTag(tct)
+		refCols = append(refCols, col.Name)
+	}
+
+	fKeyErrorMsg := fmt.Sprintf("a foreign key constraint fails (CONSTRAINT `%s` FOREIGN KEY (`%s`) REFERENCES `%s` (`%s`)) on:%s", f.ForeignKey.Name, strings.Join(refCols, ", "), f.ForeignKey.ReferencedTableName, f.ForeignKey.ReferencedTableIndex, sb.String())
```
Hydrocharged, line 76: The index for the referenced table shouldn't be used here. That position should actually have the referenced columns. To get those though, we'd need access to the referenced table schema, which isn't being passed here. For now, I'd suggest to just back out all of these changes for this function (still use the new `getRowValuesAsString` function though).

integration-tests/bats/foreign-keys.bats
```diff
@@ -1977,3 +1977,30 @@ SQL
...
+INSERT INTO b (id, b, a_id) VALUES (1, 'b', NULL);
+INSERT INTO b (id, b, a_id) VALUES (2, 'b2', 31337);
+INSERT INTO b (id, b, a_id) VALUES (3, 'b2', 31337);
+INSERT INTO b (id, b, a_id) VALUES (4, 'b2', 31337);
+SQL
+
+    run dolt sql -q "ALTER TABLE b ADD CONSTRAINT fk_b_a_id_refs_a FOREIGN KEY (a_id) REFERENCES a (id)";
+    [ "$status" -eq 1 ]
+    [[ ! "$output" =~ "panic:" ]] || false
+    [[ "$output" =~ "foreign key constraint fails (CONSTRAINT \`fk_b_a_id_refs_a\` FOREIGN KEY (\`a_id\`) REFERENCES \`a\` (\`id\`))" ]] || false
```
Hydrocharged, line 2005: This test is failing, have to make sure that you at least run any new tests that you add to make sure they pass.

Hydrocharged (changes requested):
Let me know once the feedback has been addressed and I'll do another review!

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -75,6 +74,21 @@ func (f *ForeignKeyViolationError) Error() string {
 	}
 }
 
+// getRowValuesAsString returns row values as string format.
+// Variables 'key' and 'val' have both 'tags and values' added in them, so tags are removed
```
Hydrocharged, line 78: Are we removing the tags? Unless `IterFields` removes tags, they'll be included. This can be caught in the test you've added, by expanding the message that we're expecting, as right now it's just checking for `"violation"`

Hydrocharged (changes requested):

go/libraries/doltcore/doltdb/foreign_key_coll.go
jennifersp: `IterFields` includes tags.

Hydrocharged (approved):

PR #3173, opened 2022-04-06
Implemented RowIter2 for relevant types to enable indexed point-lookup with new storage format

zachmu:


go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -297,7 +297,7 @@ RangeLoop:
 func (di doltIndex) HandledFilters(filters []sql.Expression) []sql.Expression {
 	if types.IsFormat_DOLT_1(di.vrw.Format()) {
 		// todo(andy): handle first column filters
-		return nil
+		return filters
```
andrew-wm-arthur, line 300: need to revert this change, we can't handle all filters

andrew-wm-arthur (approved):
LGTM

PR #3176, opened 2022-04-07
Moved the bulk of foreign key processing into the SQL engine

Hydrocharged:
This removes a lot of the local code dealing with foreign keys, as it has been moved to the engine.


andrew-wm-arthur:
Generally looks good. 
Unfortunately, bats tests were broken/not-running when you opened this PR, and it looks like there's a few failures when I run locally. No immediate issues with current code

PR #3193, opened 2022-04-08
Adding auto increment support for non-primary key columns

JCOR11599:


andrew-wm-arthur (approved):
You should remove the `ColumnNames()` methods, see GMS pr for details

PR #3199, opened 2022-04-11
[no-release-notes] Killed alterschema package, since it was only used by sqle package

zachmu:
Moved all functionality to sqle/alterschema.go. This should be a strict refactoring, no change in functionality.

This is a first step in supporting DDL operations with the new table format. Next step is moving format-specific parts of schema management down into that layer.

Getting this in now to avoid further merge conflicts while I continue working.


Hydrocharged (approved):
LGTM

PR #3208, opened 2022-04-12
Allow local Dolt Clone

JCOR11599:
Fix for: https://github.com/dolthub/dolt/issues/3184
For some reason we use `EmptyWorkingSet()` (which has a TODO for it to be deleted/modified), instead of `srcDB.ResolveWorkingSet()`.

can't include `/c/...`  before path on windows or it'll translate it to `C:\c\Users\...`


integration-tests/bats/replication.bats
```diff
@@ -393,3 +393,13 @@ teardown() {
     [ "${#lines[@]}" -eq 2 ]
     [[ "$output" =~ "t1" ]] || false
 }
+
+@test "replication: local clone" {
+    run dolt clone file://./repo1/.dolt/noms repo2
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "chunks complete. 0 chunks being downloaded currently." ]] || false
```
zachmu, line 400: this message is overspecified, will break when we change things

go/libraries/doltcore/env/actions/clone.go
```diff
@@ -264,7 +264,17 @@ func CloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
 		return err
 	}
 
-	ws := doltdb.EmptyWorkingSet(wsRef)
+	// Retrieve working set from srcDB
+	ws, err := srcDB.ResolveWorkingSet(ctx, wsRef)
```
zachmu, line 268: Definitely don't want to do this -- this will pull in uncommitted changes in the case of a local file system remote

zachmu (commented):
Not quite the right approach, try this when there are uncommited changes on the remote

integration-tests/bats/remotes.bats
```diff
@@ -1463,3 +1463,24 @@ setup_ref_test() {
     dolt push --set-upstream origin feature
     dolt push
 }
+
+@test "remotes: clone local repo and push" {
```
zachmu, line 1467: Name of this test should be "clone local repo with file url"

Couple things: 

1) Make sure no uncommitted changes show up in the cloned repo.
2) Make sure push works as expected. Push on a different branch than the one that has working set changes.

zachmu: Also make sure this works with an absolute path

zachmu (approved):
LGTM, couple test improvements to make

PR #3209, opened 2022-04-12
add bats tests for importing from mysqldump

jennifersp:
Added bats tests for importing from mysqldump of currently tested simple cases including empty database and databases with view, trigger, procedure. 
Added skipped bats tests for currently known issue, character set introducer support.


integration-tests/bats/import-mysqldump.bats
```diff
@@ -0,0 +1,241 @@
...
+
+/*!50001 DROP VIEW IF EXISTS myview*/;
+/*!50001 CREATE ALGORITHM=UNDEFINED */
+/*!50013 DEFINER=root@localhost SQL SECURITY DEFINER */
+/*!50001 VIEW myview AS select mytable.id AS id,mytable.col2 AS col2,mytable.col3 AS col3 from mytable */;
+SQL
+    [ "$status" -eq 0 ]
+
+    dolt sql -q "INSERT INTO mytable (id, col3) VALUES (1, TIMESTAMP('2003-12-31'));"
+    run dolt sql -q "SELECT * FROM myview;"
```
VinaiRachakonda, line 154: Run `show_create_view` as well

integration-tests/bats/import-mysqldump.bats
```diff
@@ -0,0 +1,241 @@
...
+LOCK TABLES mytable WRITE;
+/*!40000 ALTER TABLE mytable DISABLE KEYS */;
+INSERT INTO mytable VALUES (0,2),(1,3),(2,44);
+/*!40000 ALTER TABLE mytable ENABLE KEYS */;
+UNLOCK TABLES;
+
+/*!50003 SET @saved_sql_mode       = @@sql_mode */ ;
+/*!50003 SET sql_mode              = 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION' */ ;
+DELIMITER ;;
+/*!50003 CREATE*/ /*!50017 DEFINER=root@localhost*/ /*!50003 TRIGGER tt BEFORE INSERT ON mytable FOR EACH ROW SET NEW.v1 = NEW.v1 * 11 */;;
```
VinaiRachakonda, line 177: Can you add backticks to `root` and `localhost` as part of the definer. cc: https://github.com/dolthub/dolt/issues/3091

integration-tests/bats/import-mysqldump.bats
```diff
@@ -0,0 +1,241 @@
...
+CREATE DEFINER=root@localhost PROCEDURE new_proc(x DOUBLE, y DOUBLE)
+SELECT x*y ;;
+DELIMITER ;
+/*!50003 SET sql_mode              = @saved_sql_mode */ ;
+SQL
+    [ "$status" -eq 0 ]
+
+    run dolt sql -q "CALL new_proc(2, 3);"
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "| 6       |" ]] || false
```
VinaiRachakonda, line 204: You can also csv export to make testing easier

integration-tests/bats/import-mysqldump.bats
```diff
@@ -0,0 +1,241 @@
...
+  engine_name varchar(64) NOT NULL,
+  device_type int NOT NULL,
+  cost_name varchar(64) NOT NULL,
+  cost_value float DEFAULT NULL,
+  last_update timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  comment varchar(1024) DEFAULT NULL,
+  default_value float GENERATED ALWAYS AS ((case cost_name when _utf8mb3'io_block_read_cost' then 1.0 when _utf8mb3'memory_block_read_cost' then 0.25 else NULL end)) VIRTUAL,
+  PRIMARY KEY (cost_name,engine_name,device_type)
+) /*!50100 TABLESPACE mysql */ ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 STATS_PERSISTENT=0 ROW_FORMAT=DYNAMIC;
+SQL
```
VinaiRachakonda, line 239: Would love a show create table here in the future. I know this doesn't work right now.

VinaiRachakonda (commented):
This is a great start. I would love to see these tests moved to the mysql-client-tests directory (I can help you with that) so we can actually call the mysqldump command in these tests. 

I'd also like to see a couple sample tables created in a mysql server that we can run mysqldump against. These should then be loaded into Dolt.

integration-tests/mysql-client-tests/import-mysqldump.bats
```diff
@@ -0,0 +1,210 @@
...
+    dolt init
+}
+
+teardown() {
+    cd ..
+    rm -rf $REPO_NAME
+}
+
+@test "import mysqldump: empty database dump" {
+    service mysql start
```
VinaiRachakonda, line 17: I think you can move start and stop to setup and teardown

VinaiRachakonda (approved):
Nice work! Let's get the one-experience database you created in here as well to make sure we can correctly dump it. Let's also get the wacky table with all sorts of types in here as well.

integration-tests/mysql-client-tests/import-mysqldump.bats
jennifersp: I kept this inside the test since we use mysql-server in only two tests, others I test on parts of mysqldump code without using mysql server or mysqldump

PR #3223, opened 2022-04-14
go/store: changes to support memory management of NomsBlockStores

druvv:
Various changes to support doltremoteapi NBS memory management.

**Changes:**
- Acquires quota with a singular call during a Rebase
- Table file indices loaded from an AWS or Blob chunk source load their contents into an individually mmapped region of memory. When an index is closed, it's associated mmapped region is unmapped.
- When loading indices from a disk file, stop using mmap pointlessly
- Optimizes index memory usage. Instead of allocating an additional `n` sized buffer to hold the offsets, allocate a `n - n/2` sized buffer and recycle the unused lengths region of memory.
- Table reader no longer copies all the prefixes when reading

Also renames the following noms copyrighted files:
- `mmap_table_reader.go` -> `file_table_reader.go`
- `mmap_table_reader_test.go` -> `file_table_reader_test.go`


druvv:
#deploy [doltremoteapi]

druvv:
#deploy [doltremoteapi]

go/store/nbs/table_index.go
```diff
@@ -479,188 +556,138 @@ func (ti onHeapTableIndex) Clone() (tableIndex, error) {
...
+	used uint64
 }
 
-func (i mmapTableIndex) Clone() (tableIndex, error) {
-	cnt := atomic.AddInt32(i.refCnt, 1)
-	if cnt == 1 {
-		panic("Clone() called after last Close(). This index is no longer valid.")
+func mmapWithStats(f *os.File, length int, prot, flags int, offset int64) (mmapWStat, error) {
+	GlobalMmapStats.mu.Lock()
+	GlobalMmapStats.WillMmap(uint64(length), GlobalMmapStats.totalUsed)
```
reltuk, line 672: `defer GlobalMmapStats.mu.Unlock()` (or Unlock it in the error early return for MapRegion)

go/store/nbs/table_index.go
```diff
@@ -479,188 +556,138 @@ func (ti onHeapTableIndex) Clone() (tableIndex, error) {
...
-		if i.prefixes[h] < prefix {
-			idx = h + 1 // preserves f(i-1) == false
-		} else {
-			j = h // preserves f(j) == true
-		}
+func (m mmapWStat) Unmap() error {
+	GlobalMmapStats.mu.Lock()
+	err := m.m.Unmap()
+	if err != nil {
+		return err
```
reltuk, line 687: Also here, don't leak the lock.

reltuk (approved):
LGTM. Two minor comments :).

PR #3234, opened 2022-04-14
/go/cmd/dolt/commands/login.go: make `dolt login` command configurable, can authenticate against different remotes

coffeegoddd:
This PR adds the `--auth-endpoint`, `--login-url`, and `--insecure` arguments to the `dolt login` command so that it can authenticate against remotes beyond `doltremoteapi.dolthub.com:443`. This supports `dolt login`'s use with DoltLab instances. It also adds a dolt config option `doltlab.insecure`, which if `true` will establish an insecure gRPC connection to the remote authentication server (currently required for DoltLab's remote auth server).

`dolt login` will continue to use `doltremoteapi.dolthub.com:443` as the default remote `authEndpoint` and `https://www.dolthub.com/settings/credentials` as the `loginUrl` opened by the browser. 

If a local or global dolt config file sets `remotes.default_host` and `remotes.default_port`, the values specified in the config will be used over the defaults. If `creds.add_url` is set in the config file, its value will be used instead of the default `loginUrl`. All values supplied with the command line arguments take precedence over the config file values.

Examples:
```bash
# login to dolthub.com
dolt login
```

```bash
# login to doltlab.dolthub.com
dolt login -i -e doltlab.dolthub.com:50051 -url http://doltlab.dolthub.com/settings/credentials
```

```bash
# login to doltlab.dolthub.com with config
dolt config --global --add remotes.default_host doltlab.dolthub.com
dolt config --global --add remotes.default_port 50051
dolt config --global --add creds.add_url http://doltlab.dolthub.com/settings/credentials
dolt config --global --add doltlab.insecure true
dolt login
```

### DIFF
--- a/go/cmd/dolt/commands/login.go
+++ b/go/cmd/dolt/commands/login.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strconv"
 	"time"
 
 	"github.com/skratchdot/open-golang/open"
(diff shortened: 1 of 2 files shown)


reltuk (approved):
LGTM!

PR #3245, opened 2022-04-17
Enable unique indexes on keyless tables

VinaiRachakonda:
This pr

1. Unblocks unique indexes for keyless tables
2. Fixes index/cat.go for keyless rows
3. Adds additional test and verification of unique key indexes


go/libraries/doltcore/table/editor/keyless_table_editor.go
```diff
@@ -493,10 +494,10 @@ func applyEdits(ctx context.Context, tbl *doltdb.Table, acc keylessEditAcc, inde
 			return nil, err
 		}
 
-		func(k, v types.Tuple) (*doltdb.Table, error) {
+		err = func(k, v types.Tuple) (localErr error) {
 			indexOpsToUndo := make([]int, len(indexEds))
 			defer func() {
-				if retErr != nil {
+				if localErr != nil {
```
VinaiRachakonda, line 500: @max-hoffman TBH i still haven't found an edge case that break this code. Would love some insight on why you wrote and whether it should be salvaged or not.

integration-tests/bats/keyless.bats
```diff
@@ -984,3 +984,191 @@ SQL
...
+
+    run dolt sql -q "INSERT INTO gis VALUES (1, POINT(1,1))"
+    [ $status -eq 1 ]
+    [[ "$output" =~ "duplicate unique key given: [1]" ]] || false
+
+    run dolt sql -r csv -q "SELECT count(*) FROM gis where pk = 1"
+    [ $status -eq 0 ]
+    [[ "${lines[1]}" = "1" ]] || false
+
+    run dolt sql -q "INSERT INTO gis VALUES (NULL, POINT(1,1))"
```
VinaiRachakonda, line 1040: @max-hoffman This is standard Auto Increment logic. You can put in NULL and it will get resolved with the next auto increment value.

go/libraries/doltcore/table/editor/keyless_table_editor.go
```diff
@@ -493,10 +494,10 @@ func applyEdits(ctx context.Context, tbl *doltdb.Table, acc keylessEditAcc, inde
 			return nil, err
 		}
 
-		func(k, v types.Tuple) (*doltdb.Table, error) {
+		err = func(k, v types.Tuple) (localErr error) {
```
VinaiRachakonda, line 497: Fyi: https://stackoverflow.com/questions/37248898/how-does-defer-and-named-return-value-work

max-hoffman (approved):
Summarizing this morning, 1) add a TODO next to the dead code to mark for simplification in the same vein as the pk_table_editor rewrite, and 2) enginetests to better enforce the desired behavior and provide a testbed for inevitable keyless index bugs in the future Otherwise LGTM, thanks for your patience.

PR #3246, opened 2022-04-18
Added stored procedure versions of all dolt functions

Hydrocharged:


go/libraries/doltcore/sqle/dfunctions/dolt_merge_base.go
```diff
@@ -27,11 +27,13 @@ import (
...
 const DoltMergeBaseFuncName = "dolt_merge_base"
 
+// Deprecated: please use the version in the dprocedures package
 type MergeBase struct {
 	expression.BinaryExpression
 }
 
 // NewMergeBase returns a MergeBase sql function.
+// Deprecated: please use the version in the dprocedures package
 func NewMergeBase(left, right sql.Expression) sql.Expression {
```
zachmu, line 37: This one should not be deprecated: it's read-only (an actual function, not a procedure)

go/libraries/doltcore/sqle/dfunctions/hashof.go
```diff
@@ -30,11 +30,13 @@ import (
...
 const HashOfFuncName = "hashof"
 
+// Deprecated: please use the version in the dprocedures package
 type HashOf struct {
 	expression.UnaryExpression
 }
 
 // NewHashOf creates a new HashOf expression.
+// Deprecated: please use the version in the dprocedures package
 func NewHashOf(e sql.Expression) sql.Expression {
```
zachmu, line 40: Same comment, this is fine as a function

go/libraries/doltcore/sqle/dfunctions/verify_constraints.go
```diff
@@ -33,6 +33,7 @@ const (
 )
 
 // ConstraintsVerifyFunc represents the sql functions "verify_constraints" and "verify_constraints_all".
+// Deprecated: please use the version in the dprocedures package
```
zachmu, line 36: Same, no need to convert to stored procedure

go/libraries/doltcore/sqle/dfunctions/version.go
```diff
@@ -20,9 +20,11 @@ const VersionFuncName = "dolt_version"
 
 var VersionString = "SET_BY_INIT"
 
+// Deprecated: please use the version in the dprocedures package
 type Version struct{}
```
zachmu, line 24: Same

go/libraries/doltcore/sqle/dfunctions/active_branch.go
```diff
@@ -25,33 +25,39 @@ import (
 
 const ActiveBranchFuncName = "active_branch"
 
+// Deprecated: please use the version in the dprocedures package
 type ActiveBranchFunc struct {
 }
 
 // NewActiveBranchFunc creates a new ActiveBranchFunc expression.
+// Deprecated: please use the version in the dprocedures package
```
zachmu, line 33: No need to deprecate / convert to a procedure, this is strictly functional with no side effects

zachmu (commented):
Two main comments:

1) Several of the migrated functions don't need to be. Anything that's read-only is actually functional and should be left as a function.

2) The tests you've added would be better expressed as script tests in the engine. We do want bats coverage just as a smoke test, but generally speaking this is engine functionality and should be tested in golang. The only reason this wasn't the case from the beginning is that it wasn't possible to test custom functions in engine tests at first.

Other than that this is straightforward and seems fine.

go/libraries/doltcore/sqle/dfunctions/verify_constraints.go
Hydrocharged: This writes to the constraint violations table, so it's not necessarily read-only.

Hydrocharged:
Failed test is flaking? Only failing on MacOS, and none of my changes should have any impact on any of the code used in the failing test. Could rerun it until it passes, but going to go ahead and merge.

PR #3249, opened 2022-04-18
Rewrite table rows when adding a column anywhere but the end of the schema

zachmu:


go/libraries/doltcore/schema/schema.go
```diff
@@ -45,6 +46,10 @@ type Schema interface {
 
 	// SetPkOrdinals specifies a primary key column ordering
 	SetPkOrdinals([]int) error
+
+	// AddColumn adds a column to this schema in the order given and returns the resulting Schema.
+	// The new column cannot be a primary key. To alter primary keys, create a new schema with those keys.
+	AddColumn(column Column, order *sql.ColumnOrder) (Schema, error)
```
andrew-wm-arthur, line 52: I don't think we want `sql` abstractions at this layer of the codebase

andrew-wm-arthur (commented):
This looks correct, but you should have Aaron give this a pass and see what his thoughts are on abstracting tables between the two formats. 
Adding another method to the `durable.Table` interface seems like the wrong path to take

go/libraries/doltcore/doltdb/durable/index.go
```diff
@@ -37,6 +40,10 @@ type Index interface {
 
 	// Format returns the types.NomsBinFormat for this index.
 	Format() *types.NomsBinFormat
+
+	// AddColumnToRows adds the column given to the rows data and returns the resulting rows.
+	// The |newCol| is present in |newSchema|.
+	AddColumnToRows(ctx context.Context, newCol string, newSchema schema.Schema) (Index, error)
```
andrew-wm-arthur, line 46: I think this would be better as a top-level function rather than on this interface, but it's nbd

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/schema/schema.go
zachmu: fixed

PR #3251, opened 2022-04-19
[no-release-notes] go/{serial,libraries/doltcore/doltdb/root_val.go}: Implement RootValue as a flatbuffers message.

reltuk:


andrew-wm-arthur (approved):
LGTM!

PR #3271, opened 2022-04-21
Fixing bad error message for `ErrBranchNotFound`

JCOR11599:
Fix error message for ErrBranchNotFound to include the branch that wasn't found.

helps for: https://github.com/dolthub/dolt/issues/3270

### DIFF
--- a/go/libraries/doltcore/doltdb/errors.go
+++ b/go/libraries/doltcore/doltdb/errors.go
@@ -17,6 +17,7 @@ package doltdb
 import (
 	"errors"
 	"fmt"
+
 	errors2 "gopkg.in/src-d/go-errors.v1"
 )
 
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/doltdb/errors.go
```diff
@@ -17,6 +17,8 @@ package doltdb
 import (
 	"errors"
 	"fmt"
+
+	errors2 "gopkg.in/src-d/go-errors.v1"
```
fulghum, line 21: Is there any existing naming pattern we've used when we need to import both `errors` and `"gopkg.in/src-d/go-errors.v1`? 

`errors2` just looked a little funny with the "errors.v1" part of the import. 

I see `goerrors` used for the newer go-errors.v1 module used in a couple places in the codebase and would be more consistent here.

fulghum (approved):
Fix looks good to me. Just one small request to keep the `go-errors.v1` module import name consistent with other places in the codebase where it looks like we're using `goerrors` as the imported name.

PR #3272, opened 2022-04-21
Add first-hour-db and bats tests for it

jennifersp:
Added dump file of `first-hour-db` database, which is slightly modified version of MySQL Sakila DB to fit current Dolt version.
Added bats tests on first-hour-db, to test current supported functionalities. 
Added both the dump file and tests as more functionalities are supported/fixed in Dolt.
Currently tests basics of importing the dump file.


integration-tests/mysql-client-tests/first-hour-tests.bats
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bats
+
+setup() {
```
fulghum, line 3: I'd recommend naming the BATS file something like `sakilla-data-dump-load.bats` just to be explicit about what's in there. That'll probably make more sense in a year or two. I'd put these bats files into a subdirectory called something like `data-dump-loading` in the mysql-client-tests directory, too. Or... you might even consider a separate directory, since these tests seem different from the other tests in mysql-client-tests that are really testing the mysql client interfaces across multiple languages.

fulghum:
It would be helpful to have a way to track how far off we are from the official Sakilla db dump. You and I had talked about diff'ing the contents before, but adding comments to the dump file that we could easily find and commenting out portions could be an easier way to do that. Ultimately, we just want a quick/easy way to know how far off we are from being able to load the official Sakilla db.

integration-tests/data-dump-loading-tests/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Running the built container will produce output like:
+```bash
+$ docker run data-dump-loading-tests:latest
+updating dolt config for tests:
+Config successfully updated.
+Config successfully updated.
+Config successfully updated.
+Config successfully updated.
+Running mysql-client-tests:
```
coffeegoddd, line 22: This is the wrong output, should be `data-dump-loading-tests`

coffeegoddd (approved):
LGTM! needs to be in `main` to start running, but we can iterate on any brokenness missed in my review.

coffeegoddd:
Oh and please change the `FROM` in your docker file to `FROM --platform=linux/amd64 ubuntu:18.04` for M1 users

integration-tests/MySQLDockerfile
```diff
@@ -19,7 +19,6 @@ RUN apt update -y && \
         wget \
         pkg-config \
         mysql-client \
-        mysql-server \
```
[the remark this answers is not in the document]
jennifersp: yes, before I added it for import-mysql-dump.bats tests, but moving that out of mysql-client-tests now.

fulghum (approved):
Nice work getting this mysqldump test together. 

I love that we are so close to running the full Sakila mysqldump file! Good job clearly commenting the unsupported features, making sure we had GitHub tracking issues for each, and linking those issues in the unsupported comments. 👍  Looks like there are only three features missing for full support of this mysqldump!

PR #3298, opened 2022-04-25
Fix ErrBranchNotFound

JCOR11599:
This reverts commit 8cd5f16de103f92b40a65c099b82caef558e7608, reversing
changes made to f857be6c4a4d51422f038a87706f08b019179897.

### DIFF
--- a/go/libraries/doltcore/doltdb/errors.go
+++ b/go/libraries/doltcore/doltdb/errors.go
@@ -17,8 +17,6 @@ package doltdb
 import (
 	"errors"
 	"fmt"
-
-	goerrors "gopkg.in/src-d/go-errors.v1"
 )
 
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/doltdb/errors.go
```diff
@@ -82,6 +82,9 @@ func IsNotFoundErr(err error) bool {
 }
 
 func IsNotACommit(err error) bool {
+	if errors.Is(err, ErrBranchNotFound) {
```
zachmu, line 85: If this is necessary for this type, do it for the other types in the case as well, otherwise it will break when we begin wrapping those too

zachmu (approved):
LGTM

PR #3322, opened 2022-04-28
[no-release-notes] go/store/prolly/tree: node_splitter.go: Make the key splitter a bit more principled and able to target any CDF.

reltuk:


andrew-wm-arthur (approved):
looks great!

PR #3329, opened 2022-04-29
Changed default behavior for handling merge conflicts in sessions

zachmu:
Changed the default of `@@dolt_allow_commit_conflicts` to false, and changed how merge conflicts that happen incidentally on commit are handled.

A commit that results in a merge conflict from another client updating the same head will always roll back the transaction, since many connectors will be unable to recover from such a condition otherwise, regardless of the value of `@@dolt_allow_commit_conflicts`.

A commit that has merge conflicts from a `dolt_merge()` can be committed by setting `@@dolt_allow_commit_conflicts = on` to allow others to collaborate on resolving the conflicts. By default, attempting to commit a working set with merge conflicts will result in a rollback, as above.


go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -36,6 +37,8 @@ const (
 	maxTxCommitRetries = 5
 )
 
+var ErrRetryTransaction = errors.New("another client has modified the working set incompatibly. retry the transaction")
```
andrew-wm-arthur, line 40: maybe something more like:
"this transaction conflicts with a committed transaction from another client, please retry"

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
```diff
@@ -795,9 +728,400 @@ var DoltTransactionTests = []enginetest.TransactionTest{
...
+			},
+			{
+				Query:    "/* client b */ select count(*) from dolt_conflicts",
+				Expected: []sql.Row{{1}},
+			},
+			{
+				Query:    "/* client b */ select * from test order by 1",
+				Expected: []sql.Row{{0, 0}, {1, 2}},
+			},
+			{ // no error because of our session settings
```
andrew-wm-arthur, line 930: we want to be able to commit here even if `client a` has done something (unrelated) concurrently

andrew-wm-arthur (approved):
LGTM!

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
zachmu: I added a TODO about this, I want to get this in

PR #3332, opened 2022-04-29
Fix composite index merge bug

druvv:
A bug was discovered where a composite index is left in an inconsistent state after a merge. This PR adds a bats test to catch this bug, and fixes it.

In preparation for the new storage format, this PR also rewrites the merge tests in a universal way. This will allow us to test both the old format and new format functionality with the same test cases.

### DIFF
--- a/go/libraries/doltcore/merge/merge.go
+++ b/go/libraries/doltcore/merge/merge.go
@@ -296,7 +296,8 @@ func calcTableMergeStats(ctx context.Context, tbl *doltdb.Table, mergeTbl *doltd
 	return ms, nil
 }
 
-type rowMerger func(ctx context.Context, nbf *types.NomsBinFormat, sch schema.Schema, r, mergeRow, baseRow types.Value) (types.Value, bool, error)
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/merge/merge.go
```diff
@@ -296,7 +296,8 @@ func calcTableMergeStats(ctx context.Context, tbl *doltdb.Table, mergeTbl *doltd
 	return ms, nil
 }
 
-type rowMerger func(ctx context.Context, nbf *types.NomsBinFormat, sch schema.Schema, r, mergeRow, baseRow types.Value) (types.Value, bool, error)
+// returns merged roow, bool indicating if a cell-wise merge was performed, bool indicating a conflict, and error
+type rowMerger func(ctx context.Context, nbf *types.NomsBinFormat, sch schema.Schema, r, mergeRow, baseRow types.Value) (types.Value, bool, bool, error)
```
andrew-wm-arthur, line 300: 4 return values is too many. You should either separate `rowMerger` logic into multiple steps or create a new type to hold the returned information

andrew-wm-arthur (approved):
lgtm, one comment

PR #3345, opened 2022-05-04
Implemented foreign keys for the new format

Hydrocharged:
This adds foreign key support to the new format.


go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -153,25 +162,72 @@ func (w *prollyTableWriter) Close(ctx *sql.Context) error {
...
+	idx := index.IndexFromIndexLookup(lookup)
+	return prollyFkIndexer{
+		writer: w,
+		index:  idx,
+		pRange: index.ProllyRangesFromIndexLookup(lookup)[0],
+	}
+}
+
+// Reset puts the writer into a fresh state, updating the schema and index writers according to the newly given table.
+func (w *prollyTableWriter) Reset(ctx context.Context, sess *prollyWriteSession, tbl *doltdb.Table, sch schema.Schema) error {
```
andrew-wm-arthur, line 203: I'm missing some detail about why this is necessary. Do we do this for DDL / adding foreign keys? 
In those cases we should probably be discarding the write session and creating a new one

PR #3346, opened 2022-05-04
[no-release-notes] implement cell-wise merges for new storage format

druvv:
Implements cell-wise merges. If a cell-wise merge occurs, we need to update the secondary indexes before we merge them so that after the merge the indexes are consistent. We accomplish this by resetting the entry on the left to the ancestor, and updating the entry on the right to a value consistent with the cell merge. 

Updates cell-wise merge tests to work across the storage formats.


druvv:
I just checked what the conflict reports for the above and it shows this:
```
➜ dolt conflicts cat .
+-----+--------+---+---+
|     |        | a | b |
+-----+--------+---+---+
|  +  | ours   | 1 | 1 |
|  +  | theirs | 1 | 1 |
+-----+--------+---+---+
```

I need to fix the conflict or diff reporting to also show column c, it's unclear why it is missing.

andrew-wm-arthur (commented):
General approach looks reasonable. Needs some reorganization

go/libraries/doltcore/merge/merge.go
```diff
@@ -320,84 +379,202 @@ func mergeProllyRowData(ctx context.Context, postMergeSchema, rootSch, mergeSch,
...
-//		}
-//		mergedValues[i] = v
-//	}
-//
-//	return val.NewTuple(syncPool, mergedValues...), true
-//}
+type valueMerger struct {
+	numCols                                int
+	vD, lVD, rVD, bVD                      val.TupleDesc
+	leftMapping, rightMapping, baseMapping map[int]int
```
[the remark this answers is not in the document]
druvv: I don't think I can use `val.OrdinalMapping` because there may not be a match for column in the merged schema. For example, if right adds a column, the merged schema will contain the added column, but left will not. 

I could store `-1` in the val.OrdinalMapping if necessary and check for it.

go/libraries/doltcore/merge/merge_prolly.go
```diff
@@ -0,0 +1,466 @@
...
+	var mergedData durable.Index
+
+	group.Go(func() error {
+		var err error
+		// TODO (dhruv): update this function definition to return any conflicts
+		updatedTable, mergedData, err = mergeProllyRowData(gCtx, postMergeSchema, rootSchema, mergeSchema, ancSchema, tbl, mergeTbl, ancTbl, tableToUpdate, cellWiseMerges)
+		if err != nil {
+			return err
+		}
+		close(cellWiseMerges)
```
andrew-wm-arthur, line 66: defer this close

go/libraries/doltcore/merge/merge_prolly.go
```diff
@@ -0,0 +1,466 @@
...
+//
+// Each cell-wise merge reverts the corresponding index entries in the root
+// branch, and modifies index entries in the merge branch. The merge branch's
+// entries are set to values consistent the cell-wise merge result. When the
+// root and merge secondary indexes are merged, they will produce entries
+// consistent with the primary row data.
+func mergeTableData(ctx context.Context, vrw types.ValueReadWriter, postMergeSchema, rootSchema, mergeSchema, ancSchema schema.Schema, tbl, mergeTbl, ancTbl, tableToUpdate *doltdb.Table) (*doltdb.Table, *MergeStats, error) {
+	group, gCtx := errgroup.WithContext(ctx)
+
+	cellWiseMerges := make(chan cellWiseMerge)
```
andrew-wm-arthur, line 55: buffer this channel

PR #3348, opened 2022-05-04
Max/versioned enginetests

max-hoffman:
Engine test `CheckpointHarness` and parallelize tests:

Before
```
> go test github.com/dolthub/dolt/go/libraries/doltcore/sqle/enginetest -count=1
ok  	github.com/dolthub/dolt/go/libraries/doltcore/sqle/enginetest	246.408s
```

After:
```
> go test github.com/dolthub/dolt/go/libraries/doltcore/sqle/enginetest -count=1
ok  	github.com/dolthub/dolt/go/libraries/doltcore/sqle/enginetest	30.254s
```

edit: new seems to vary between 90-150sec on my machine

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -68,7 +68,7 @@ require (
 )
 
 require (
-	github.com/dolthub/go-mysql-server v0.11.1-0.20220518032013-ac288f7475eb
+	github.com/dolthub/go-mysql-server v0.11.1-0.20220518220116-fbd5e8b1e49e
 	github.com/google/flatbuffers v2.0.5+incompatible
 	github.com/gosuri/uilive v0.0.4
 	github.com/kch42/buzhash v0.0.0-20160816060738-9bdec3dec7c6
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/writer/prolly_write_session.go
```diff
@@ -193,5 +193,25 @@ func (s *prollyWriteSession) setWorkingSet(ctx context.Context, ws *doltdb.Worki
...
+
+	root := ws.WorkingRoot()
+	if err := s.updateAutoIncrementSequences(ctx, root); err != nil {
+		return err
+	}
+
 	return nil
 }
+
+func (s *prollyWriteSession) updateAutoIncrementSequences(ctx context.Context, root *doltdb.RootValue) error {
```
andrew-wm-arthur, line 205: I don't think this is the behavior we want for auto increment. In particular, rolling back a transaction does not reset the auto increment sequence to its previous state

max-hoffman: @zachmu

PR #3354, opened 2022-05-05
Fix `dolt conflicts cat` failing to show conflicted rows in the union-ed schema between left, right, and base.

druvv:
Fixes a one-liner bug where the conflicted rows wouldn't show with the union schema between left, right, and base. Added a bats test as well.

### DIFF
--- a/integration-tests/bats/conflict-cat.bats
+++ b/integration-tests/bats/conflict-cat.bats
@@ -1,0 +1,55 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
+
+setup() {
+    setup_common
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "conflicts should show using the union-schema (new schema on right)" {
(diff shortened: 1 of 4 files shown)


integration-tests/bats/conflict-cat.bats
```diff
@@ -0,0 +1,55 @@
...
+
+setup() {
+    setup_common
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "conflict-cat: conflicts should show using the union-schema (new schema on right)" {
```
andrew-wm-arthur, line 12: what's the behavior if the column exists on both branches but is NULL on one?

andrew-wm-arthur (approved):
LGTM!

andrew-wm-arthur:
Don't merge this yet

druvv:
We decided that the conflict behavior for the old storage format is correct. I have reverted the changes to the old storage format, and instead added a todo for the new storage format. 

Kept the changes for the conflict cat issue.

PR #3359, opened 2022-05-05
[no-release-notes] go/store/types: TupleRowStorage: Implement walkRefs.

reltuk:


go/store/types/tuplerowstorage.go
```diff
@@ -58,10 +61,39 @@ func (v TupleRowStorage) isPrimitive() bool {
 }
 
 func (v TupleRowStorage) WalkValues(ctx context.Context, cb ValueCallback) error {
-	return nil
+	return errors.New("unsupported WalkValues on TupleRowStorage. Use types.WalkValues.")
 }
 
 func (v TupleRowStorage) walkRefs(nbf *NomsBinFormat, cb RefCallback) error {
+	switch serial.GetFileID([]byte(v)) {
```
andrew-wm-arthur, line 68: I think we should use [`message.WalkAddresses()`](https://github.com/dolthub/dolt/blob/ea91de8c7f9c268a1bbdd0f6b061f1fdc3066dcf/go/store/prolly/message/message.go#L56) here

reltuk: Yeah, that's clearly right. Just fixed it :).

PR #3365, opened 2022-05-06
[no-release-notes] column_statistics table in mysql db

JCOR11599:


go/libraries/doltcore/sqle/tables.go
```diff
@@ -400,22 +407,176 @@ func (t *DoltTable) DataLength(ctx *sql.Context) (uint64, error) {
 	return numBytesPerRow * numRows, nil
 }
 
-// TODO: Have actual implementations for AnalyzeTable and Statistics; this is just a quick fix to unblock others.
-
 // AnalyzeTable implements the sql.StatisticsTable interface.
 func (t *DoltTable) AnalyzeTable(ctx *sql.Context) error {
+	table, err := t.DoltTable(ctx)
```
max-hoffman, line 412: now might be a good time to make either a `HistogramBuilder` or `TableStatisticsBuilder` in GMS to avoid the duplication here. A couple ways of doing that: i) builder for each table, consumes rows; ii) constructor is passed an `sql.RowIter`, it exhausts the row iter and returns `map[string]sql.TableStatistics`

go/libraries/doltcore/sqle/tables.go
```diff
@@ -98,10 +99,16 @@ func (ds *DoltTableStatistics) NullCount() uint64 {
 }
 
 func (ds *DoltTableStatistics) Histogram(colName string) (*sql.Histogram, error) {
```
max-hoffman, line 101: why do we need `DoltTableStatistics`? just to wrap `map[string]sql.TableStatistics`?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -400,22 +407,176 @@ func (t *DoltTable) DataLength(ctx *sql.Context) (uint64, error) {
...
+
+				hist.Mean += v
+				hist.Min = math.Min(hist.Min, v)
+				hist.Max = math.Max(hist.Max, v)
+				hist.Count++
+			}
+		}
+	}
+
+	// TODO: logic to determine ranges for buckets
```
max-hoffman, line 510: merged comments should either be: 1) consolidated into a docstring to describe what the function does, or 2) explain tricky edge cases that future readers would stumble on

go/libraries/doltcore/sqle/tables.go
```diff
@@ -400,22 +407,176 @@ func (t *DoltTable) DataLength(ctx *sql.Context) (uint64, error) {
...
-	if err != nil {
-		return nil, err
+	if t.doltStats != nil {
+		return t.doltStats, nil
 	}
-	return &DoltTableStatistics{
-		rowCount: numRows,
-	}, nil
+
+	// Load stats from the session
```
max-hoffman, line 558: same as above about docstrings

go/libraries/doltcore/sqle/tables.go
max-hoffman: (maybe fold with info schema fix?)

max-hoffman (commented):
Pretty good, some nits for docstrings, worth refactoring the histogram building to only write it once on the GMS side

go/libraries/doltcore/sqle/tables.go
JCOR11599: it implements `sql.TableStatistics`

PR #3369, opened 2022-05-06
short dolt procs

max-hoffman:
```SQL
> call dbranch('-b', 'new')
> call dadd('.');
> call dcommit('-am', 'message');
> call dreset('--hard', 'main')
```


zachmu (approved):
Fine with this but please no more bats tests, these should be engine tests only

Hydrocharged:
Something to consider:
We can guard the `dolt_` namespace for new procedures, but that’s not the case for this `d` prefix. I know they’re supplementary rather than replacements for the originals, but it’s something to consider.

max-hoffman:
thanks for the comments guys! actively working on an escape hatch for bats

PR #3370, opened 2022-05-06
[no-release-notes] adds merge conflicts for the new storage format.

druvv:
## Summary
- New storage format passes `merge.bats` after this PR except on one test which relies on `MergeStats`. MergeStats is a TODO.

- Adds and implements `durable.ConflictIndex`. 
- Adds conflict handling logic to `mergeTableData`. 
- Fixes a new storage format bug that occurs when the same two tables are added in left and right.
- Moves secondary index merges into its own file for clarity.

## Updating secondary indexes when a conflict occurs
When a conflict occurs, we also need to update the secondary indexes so that the left value's index entires are kept post-merge. The edit simply resets the right value.

Resetting values is the inverse of the original operation (Update -> Update, Insert -> Delete, Delete -> Insert). Previously, MutableSecondaryIndex only supported the Update operation, but with conflicts we needed to add all three. In order to send these edits over a single channel to the goroutine that processes these index updates, we added the interface `indexEdit`.

There are two implementations of `indexEdit`: a `conflictEdit` and a `cellWiseMergeEdit`. `conflictEdit` simply resets the right value. `cellWiseMergeEdit` resets the left value and updates the right value to the cell-wise merge result.

### DIFF
--- a/go/libraries/doltcore/merge/merge_prolly_indexes.go
+++ b/go/libraries/doltcore/merge/merge_prolly_indexes.go
@@ -1,3 +1,17 @@
+// Copyright 2022 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package merge
 
 import (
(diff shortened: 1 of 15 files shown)


go/libraries/doltcore/merge/merge_prolly_indexes.go
```diff
@@ -0,0 +1,315 @@
...
+)
+
+// indexEdit is an edit to a secondary index based on the primary row's key and value.
+// The members of tree.Diff have the following meanings:
+// - |Key| = the key to apply this edit on
+// - |From| = the previous value of this key (If nil, an insert is performed.)
+// - |To| = the new value of this key (If nil, a delete is performed.)
+//
+// It is invalid for |From| and |To| to be both be nil.
+type indexEdit interface {
```
andrew-wm-arthur, line 38: I'm not super excited about this interface if its only motivation is to use a common channel for `cellWiseMergeEdit` and `conflictEdit`. The allocation overhead of boxing these types as interfaces will likely outweigh the benefits of asynchronously updating secondary indexes. 

That being said, I don't think it's worth changing within this PR, we can take another pass at it in the future

andrew-wm-arthur (approved):
LGTM!

PR #3379, opened 2022-05-10
Only skip engine, not inverse

max-hoffman:
Enginetest -race has it's own CI job that is only run on merge main. As part of that PR, I tried to skip enginetest -race during routine PR tests. I mistakenly did the opposite, running only enginetest -race on ubuntu.


andrew-wm-arthur (approved):
LGTM!

PR #3385, opened 2022-05-10
Support add and modify column via table rewrites (new format only for now)

zachmu:


zachmu:
This relies on https://github.com/dolthub/go-mysql-server/pull/1004 to actually do anything useful

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -470,7 +470,9 @@ func (root *RootValue) GenerateTagsForNewColumns(
 	if headRoot != nil {
 		for _, col := range existingCols {
 			for i := range newColNames {
-				if strings.ToLower(newColNames[i]) == strings.ToLower(col.Name) {
+				// Only re-use tags if the noms kind didn't change
+				if strings.ToLower(newColNames[i]) == strings.ToLower(col.Name) &&
```
andrew-wm-arthur, line 474: you should add a todo here to revisit this

go/libraries/doltcore/sqle/globalstate/auto_increment_tracker.go
```diff
@@ -46,15 +46,30 @@ func CoerceAutoIncrementValue(val interface{}) (uint64, error) {
 	return val.(uint64), nil
 }
 
-func NewAutoIncrementTracker(ctx context.Context, ws *doltdb.WorkingSet) (ait AutoIncrementTracker, err error) {
-	ait = AutoIncrementTracker{
+// AutoIncrementTracker knows how to track the latest auto-increment values for every table across all sessions.
+type AutoIncrementTracker interface {
```
andrew-wm-arthur, line 50: we don't need this interface, it only has one implementation

andrew-wm-arthur (approved):
LGTM!

PR #3398, opened 2022-05-11
go/libraries/doltcore/sqle: Change Index and IndexLookup implementations to not capture table rows. Resolve them from the table at RowIter/Partitions time.

reltuk:


andrew-wm-arthur (approved):
LGMT!

PR #3400, opened 2022-05-12
Reworked keyless indexes for new format + keyless fk

Hydrocharged:


go/store/prolly/map.go
```diff
@@ -196,6 +196,21 @@ func (m Map) iterFromRange(ctx context.Context, rng Range) (*orderedTreeIter[val
 	return treeIterFromRange(ctx, m.tuples.root, m.tuples.ns, rng)
 }
 
+// ConvertToKeylessIndex converts the given map to a keyless index map.
+func ConvertToKeylessIndex(m Map) Map {
```
andrew-wm-arthur, line 200: let's move this to `shim.go`, we probably want to revisit how we construct these tuple descriptors

go/libraries/doltcore/sqle/writer/prolly_fk_indexer.go
```diff
@@ -161,9 +165,10 @@ func (iter prollyFkKeylessRowIter) Next(ctx *sql.Context) (sql.Row, error) {
 	if k == nil {
 		return nil, io.EOF
 	}
+	hashId := k.GetField(k.Count() - 1)
 
 	nextRow := make(sql.Row, len(iter.primary.valMap))
-	err = iter.primary.mut.Get(ctx, k, func(tblKey, tblVal val.Tuple) error {
+	err = iter.primary.mut.Get(ctx, val.NewTuple(sharePool, hashId), func(tblKey, tblVal val.Tuple) error {
```
andrew-wm-arthur, line 171: would prefer to use a TupleBuilder here to construct the key tuple

andrew-wm-arthur (approved):
Looks Good! 
Some of the details that "special case" keyless tuples should be handled in `val/keyless.go`, but that can be cleaned up in follow up PRs

PR #3411, opened 2022-05-13
mysql db

JCOR11599:
Will no longer persist changes to privileges-file, and will ignore privileges file entirely if there is a `mysql.db` file present in current directory. 

TODO:
 - ~`User.Attributes` aren't being loaded/stored~
 - ~`global_dynamic` privileges aren't being loaded/stored~


go/libraries/doltcore/sqle/privileges/file_handler.go
```diff
@@ -81,12 +87,33 @@ func LoadPrivileges() ([]*grant_tables.User, []*grant_tables.RoleEdge, error) {
...
+
+var _ mysql_db.PrivilegePersistCallback = SavePrivileges
+var _ mysql_db.DataPersistCallback = SaveData
 
-// SavePrivileges implements the interface grant_tables.PersistCallback. This is used to save privileges to disk. If the
+// SavePrivileges implements the interface mysql_db.PrivilegePersistCallback. This is used to save privileges to disk. If the
 // file path has not been previously set, this returns without error. This is so that the logic path can retain the
 // calls regardless of whether a user wants privileges to be loaded or persisted.
-func SavePrivileges(ctx *sql.Context, users []*grant_tables.User, roles []*grant_tables.RoleEdge) error {
+func SavePrivileges(ctx *sql.Context, users []*mysql_db.User, roles []*mysql_db.RoleEdge) error {
```
Hydrocharged, line 116: Looking at this interface a bit more, I think the best way to do this is to have GMS actually handle all of the FlatBuffer logic. So we can have `SaveData` take a byte slice and write it to disk, and `LoadData` load the byte slice from a file and send it to GMS. We also would need that `LoadPrivilegesData` function in GMS as well.

Hydrocharged (commented):
Looking at this a bit more, I think we should probably change a few things on the GMS side to make it a bit more sensible on the Dolt side.

Hydrocharged (commented):
Along with the other comments, we should add some bats tests. Easiest way to do this would be to use `expect`, and run the server using `dolt sql-client -d`, which will spawn a server and automatically connect the client for you to use.

You can see some examples of how `expect` works from bats tests like `sql.bats::"use syntax on shell"`. We should test the default path for both, setting a path for both, that `privs.json` is ignored when `mysql.db` is present, that `privs.json` is used when `mysql.db` isn't present (and then is ignored afterward), and that no files present will properly create a `mysql.db` with the given user credentials.

Do note that `expect` doesn't work on Windows so you'll need to add that skip. You can also make a `privs.json` file with a version off the `main` branch so you can test with it. We should always generate a `mysql.db` in the test since we may change the FlatBuffers schema in the future and it'll be annoying to update the tests then.

integration-tests/bats/sql-privs.bats
```diff
@@ -0,0 +1,258 @@
...
+    [ "${lines[2]}" = '# "exit" or "quit" (or Ctrl-D) to exit.' ]
+    [ "${lines[3]}" = '+------------+' ]
+    [ "${lines[4]}" = '| User       |' ]
+    [ "${lines[5]}" = '+------------+' ]
+    [ "${lines[6]}" = '| dolt       |' ]
+    [ "${lines[7]}" = '| mysql_user |' ]
+    [ "${lines[8]}" = '| new_user   |' ]
+    [ "${lines[9]}" = '+------------+' ]
+
+    # check that mysql.db exists, and privs.json doesn't
```
Hydrocharged, line 187: For good measure, stop and restart the server and verify that new user still exists.

Hydrocharged (approved):
A few adjustments and it's ready to commit!

PR #3412, opened 2022-05-13
Make import work with FKs and add disableFks check

VinaiRachakonda:
This pr does the following:

1. Add back FK support for dolt table import
2. Add a --disable-fk-checks flag to table import
3. Skips some of new import update tests due to issues with the new storage format
4. Integrates GMS changes that address Insert on Duplicate inconsistencies with Fks


Hydrocharged (approved):
I thought this just did not work, as when I initially tried it, it failed to use foreign keys and I added the block. Strange that removing the block now just...works. Perhaps I added additional logic after this check that somehow allows foreign keys to work, but I truly don't know what I would have changed to enable it. But if it works, it works.

PR #3429, opened 2022-05-16
[no-release-notes] Moving a Dolt-specific transaction test from GMS into Dolt repo.

fulghum:
Moving a test case from GMS to Dolt, as requested in: https://github.com/dolthub/go-mysql-server/pull/1013

Depends on bug fix in: https://github.com/dolthub/go-mysql-server/pull/1013


zachmu (approved):
LGTM

PR #3437, opened 2022-05-17
[no-release-util] Added concurrent query util

andrew-wm-arthur:


zachmu (approved):
A little readme doc above main would be good, otherwise LGTM

PR #3439, opened 2022-05-17
Grant access to `mysql` database when in dolt shell

JCOR11599:
Creating a new SQL engine will load privileges by default.


go/cmd/dolt/commands/sql.go
```diff
@@ -377,12 +379,25 @@ func execShell(
...
+	}}
+
+	se, err := engine.NewSqlEngine(ctx, mrEnv, format, initialDb, false, tempUsers, true)
 	if err != nil {
 		return errhand.VerboseErrorFromError(err)
 	}
 	defer se.Close()
 
+	// Load MySQL db
+	se.GetUnderlyingEngine().Analyzer.Catalog.GrantTables.SetPersistCallback(privileges.SavePrivileges)
```
zachmu, line 395: All of this setup code should live in NewEngine

integration-tests/bats/sql-shell.bats
```diff
@@ -42,6 +42,13 @@ teardown() {
     [[ "$output" =~ "+---------------------" ]] || false
 }
 
+@test "sql-shell: dolt sql shell has mysql db" {
```
zachmu, line 45: Need tests of adding grants via `-q`, listing them again

zachmu (commented):
Good start but I want to consolidate this setup code into one place

integration-tests/bats/sql-shell.bats
JCOR11599: Tests for `sql -q` are in `integration-tests/bats/sql.bats`, but they won't persist right now. They will after my refactor changes though.

go/cmd/dolt/commands/engine/sqlengine.go
```diff
@@ -53,7 +55,10 @@ func NewSqlEngine(
 	format PrintResultFormat,
 	initialDb string,
 	isReadOnly bool,
-	tempUsers []gms.TemporaryUser,
+	mysqlDbFilePath string,
```
zachmu, line 58: This method now officially has too many params. Time to introduce a config struct to hold them.

You can do that on a second pass after you get this in

go/libraries/doltcore/sqle/mysql_file_handler/file_handler.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/dolthub/go-mysql-server/sql/mysql_db"
 )
 
+const defaultMySQLFilePath = "mysql.db"
```
zachmu, line 28: This should go in a .dolt directory

Do that in a follow-up

zachmu (approved):
LGTM

PR #3444, opened 2022-05-18
[no-release-notes] Changing Dolt Harness to track a MultiRepoEnv 

fulghum:
DoltHarness was tracking a single DoltEnv for all databases, meaning the DoltDB, working set, etc were shared. After fixing a transaction initialization bug, that fix triggered the tests to start failing because of the DoltDB sharing. 

This change moves DoltHarness to use a MultiRepoEnv for tracking test databases, instead of a single DoltEnv.

### DIFF
--- a/go/libraries/doltcore/dtestutils/environment.go
+++ b/go/libraries/doltcore/dtestutils/environment.go
@@ -28,20 +28,28 @@ import (
 )
 
 const (
-	TestHomeDir = "/user/bheni"
-	WorkingDir  = "/user/bheni/datasets/states"
+	TestHomeDirPrefix = "/user/dolt/"
+	WorkingDirPrefix  = "/user/dolt/datasets/"
 )
 
-func testHomeDirFunc() (string, error) {
-	return TestHomeDir, nil
+// CreateTestEnv creates a new DoltEnv suitable for testing. The CreateTestEnvWithName
+// function should generally be preferred over this method, especially when working
+// with tests using multiple databases within a MultiRepoEnv.
+func CreateTestEnv() *env.DoltEnv {
+	return CreateTestEnvWithName("test")
 }
 
-func CreateTestEnv() *env.DoltEnv {
+// CreateTestEnvWithName creates a new DoltEnv suitable for testing and uses
+// the specified name to distinguish it from other test envs. This function
+// should generally be preferred over CreateTestEnv, especially when working with
+// tests using multiple databases within a MultiRepoEnv.
+func CreateTestEnvWithName(envName string) *env.DoltEnv {
 	const name = "billy bob"
 	const email = "bigbillieb@fake.horse"
-	initialDirs := []string{TestHomeDir, WorkingDir}
-	fs := filesys.NewInMemFS(initialDirs, nil, WorkingDir)
-	dEnv := env.Load(context.Background(), testHomeDirFunc, fs, doltdb.InMemDoltDB, "test")
+	initialDirs := []string{TestHomeDirPrefix + envName, WorkingDirPrefix + envName}
+	homeDirFunc := func() (string, error) { return TestHomeDirPrefix + envName, nil }
+	fs := filesys.NewInMemFS(initialDirs, nil, WorkingDirPrefix+envName)
+	dEnv := env.Load(context.Background(), homeDirFunc, fs, doltdb.InMemDoltDB+envName, "test")
 	cfg, _ := dEnv.Config.GetConfig(env.GlobalConfig)
 	cfg.SetStrings(map[string]string{
 		env.UserNameKey:  name,
--- a/go/libraries/doltcore/env/multi_repo_env.go
+++ b/go/libraries/doltcore/env/multi_repo_env.go
@@ -69,6 +69,21 @@ func (mrEnv *MultiRepoEnv) AddEnv(name string, dEnv *DoltEnv) {
 	})
 }
 
+// AddOrReplaceEnvs adds the specified DoltEnv to this MultiRepoEnv, replacing
+// any existing environment in the MultiRepoEnv with the same name.
+func (mrEnv *MultiRepoEnv) AddOrReplaceEnv(name string, dEnv *DoltEnv) {
+	// TODO: Modeling NamedEnvs as a map could probably simplify this file
+	newNamedEnvs := make([]NamedEnv, 0, len(mrEnv.envs))
+	for _, namedEnv := range mrEnv.envs {
+		if namedEnv.name != name {
+			newNamedEnvs = append(newNamedEnvs, namedEnv)
+		}
+	}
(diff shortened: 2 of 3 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_harness.go
```diff
@@ -219,13 +241,15 @@ func (d *DoltHarness) NewReadOnlyDatabases(names ...string) (dbs []sql.ReadOnlyD
 }
 
 func (d *DoltHarness) NewDatabaseProvider(dbs ...sql.Database) sql.MutableDatabaseProvider {
-	if d.env == nil {
-		d.env = dtestutils.CreateTestEnv()
+	// Because NewDatabases is called before NewDatabaseProvider, we need to re-add the
+	// test DoltEnvs created in NewDatabases to the new MultiRepo created here.
```
max-hoffman, line 245: or make the multienv here? or add the envs when we make the database? i know you are constrained by the existing interfaces, but it seems like one of these methods should be doing more work. Like should `NewDatabaseProvider` even take databases as an argument if they are already in `d.databases`? My guess is that either 1) new databases should not persist the databases, and everything should happen here, or 2) NewDatabases does the multienv setup work and NewDatabaseProvider uses `d.databases` instead of the passed dbs. Leaning towards (2), zach will have better ideas.

go/libraries/doltcore/env/multi_repo_env.go
```diff
@@ -69,6 +69,21 @@ func (mrEnv *MultiRepoEnv) AddEnv(name string, dEnv *DoltEnv) {
 	})
 }
 
+// AddOrReplaceEnvs adds the specified DoltEnv to this MultiRepoEnv, replacing
+// any existing environment in the MultiRepoEnv with the same name.
+func (mrEnv *MultiRepoEnv) AddOrReplaceEnv(name string, dEnv *DoltEnv) {
+	// TODO: Modeling NamedEnvs as a map could probably simplify this file
```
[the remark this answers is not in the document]
max-hoffman: If you do refactor this probably simplify the getters/settors.

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
```diff
@@ -187,19 +196,32 @@ func (d *DoltHarness) NewDatabase(name string) sql.Database {
...
+
+func (d *DoltHarness) NewDatabases(names ...string) []sql.Database {
+	if d.multiRepoEnv == nil {
+		d.initializeNewMultiRepoEnv()
+	}
 
 	d.databases = nil
 	d.databaseGlobalStates = nil
 	for _, name := range names {
+		dEnv := dtestutils.CreateTestEnv()
```
max-hoffman, line 214: It makes me uncomfortable that all of the databases having the same working directory `/user/bheni/datasets/states`. This was a preexisting problem and doesn't appear to break anything right now, so maybe not a big deal.

max-hoffman: For example, `CreateTestEnv` could take a single in-memory filesystem and working directory for the new db. Again, maybe this is just a TODO.

max-hoffman (approved):
Thanks for doing this! I added a lot of thoughts because multidb is underpowered right now and really important for testing transactions and replication. Some suggestions on low hanging fruit which you already acknowledged and I agree can benefit from refactors.

go/libraries/doltcore/env/multi_repo_env.go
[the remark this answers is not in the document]
fulghum: I took another look at this – I think it's straightforward and worth changing. I'll follow up in a separate PR though, just to keep the scope of this bug fix as tight as possible.

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
fulghum: Yup, I was fairly surprised this worked actually. When I was first testing a fix I had the directory paths and the in-mem URL set up to be distinct for different envs. I tested without it and was able to take that out, but I'll take another look and see if there's a little bit more I can clean up here as part of this PR.

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
```diff
@@ -75,6 +76,7 @@ func newDoltHarness(t *testing.T) *DoltHarness {
 		t:              t,
 		session:        session,
 		skippedQueries: defaultSkippedQueries,
+		createdEnvs:    make(map[string]*env.DoltEnv),
```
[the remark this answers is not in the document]
fulghum: Initializing `multiRepoEnv` here does help tidy things up a little bit. Good suggestion. 

We still need to reset the MultiRepoEnv when NewDatabaseProvider is called though, in order to ensure the tests have access to only that set of databases. For example... if we don't reset MultiRepoEnv, any tests that create databases will have those databases visible to other tests and this causes some tests to fail. Generally, it feels like a good/fair assumption that setting the databases in the provider should mean that only those databases are available. 

I played with some other ideas to try and work around this, but it felt like a lot of edge cases were coming up and I didn't want to expand the scope too much. For example... I looked at exposing a `RemoveEnv` to MultiRepoEnv to make sure only the specified dbs were present, but I realized that databases created through tests don't reflect directly in MultiRepoEnv until it is loaded from disk. So I looked at adding support for reloading, but it felt like going a little too far and pulling on too many loose threads for this PR bug fix. You're right that there's definitely more we should keep tidying up in here, though.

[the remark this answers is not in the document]
max-hoffman: good point! I am pretty close to being able to remove most of these helper interface methods, but until then I guess the ordering is mandatory.

go/libraries/doltcore/sqle/enginetest/dolt_harness.go
fulghum: I think isolating the working directories is a good call. I added a new form for `CreateTestEnv` and started using that within DoltHarness' `NewDatabase` function. This could definitely be pushed farther, but this felt like a good first step to help prevent test databases from interfering with each other in the future.

go/libraries/doltcore/dtestutils/environment.go
```diff
@@ -28,20 +28,28 @@ import (
 )
 
 const (
-	TestHomeDir = "/user/bheni"
-	WorkingDir  = "/user/bheni/datasets/states"
+	TestHomeDirPrefix = "/user/dolt/"
```
zachmu, line 31: End of an era @bheni

zachmu: We'll always have fake.horse

zachmu (approved):
LGTM

PR #3448, opened 2022-05-18
go/store/types: add types.UnionMaps

druvv:


go/store/types/map.go
```diff
@@ -821,3 +821,88 @@ func indexForKeyWithinSubtree(ctx context.Context, key orderedKey, metaSeq metaS
 
 	return idx, nil
 }
+
+type MapUnionConflictCB func(key Value, aValue Value, bValue Value) error
```
andrew-wm-arthur, line 825: this should probably just return the chosen values

andrew-wm-arthur (approved):
LGTM

PR #3454, opened 2022-05-19
Fixed panic when dropping a database with active branches in use

zachmu:


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -198,8 +198,14 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
 	p.mu.Lock()
 	defer p.mu.Unlock()
 
+	// get the case-sensitive name for case-sensitive file systems
+	// TODO: there are still cases (not server-first) where we rename databases because the directory name would need
+	//  quoting if used as a database name, and that breaks here. We either need the database name to match the directory
+	//  name in all cases, or else keep a mapping from database name to directory on disk.
+	db := p.databases[strings.ToLower(name)]
```
max-hoffman, line 205: lowercasing the map key when we add a database might make sense

PR #3457, opened 2022-05-19
support SRID value in column definition

jennifersp:


integration-tests/data-dump-loading-tests/sakila_dump.sql
```diff
@@ -80,7 +80,7 @@ CREATE TABLE `address` (
   `city_id` smallint unsigned NOT NULL,
   `postal_code` varchar(10) DEFAULT NULL,
   `phone` varchar(20) NOT NULL,
-  `location` geometry NOT NULL /* UNSUPPORTED SYNTAX (https://github.com/dolthub/dolt/issues/3229): SRID 0 */,
+  `location` geometry NOT NULL /*!80003 SRID 0 */,
```
fulghum, line 83: Whoohoo! Awesome to get rid of one of the last remaining unsupported syntax comments in the Sakila dump file!

go/libraries/doltcore/schema/typeinfo/linestring.go
```diff
@@ -214,3 +222,18 @@ func linestringTypeConverter(ctx context.Context, src *linestringType, destTi Ty
...
+	if err != nil {
+		return nil, err
+	}
+
+	dSRID, err := strconv.ParseBool(params["DefinedSRID"])
+	if err != nil {
+		return nil, err
+	}
+
+	LinestringType = &linestringType{sqlLinestringType: sql.LinestringType{SRID: uint32(sridVal), DefinedSRID: dSRID}}
```
fulghum, line 237: Same comment here as with `Geometry` – I'm not sure we want to cache the last instantiated Linestring TypeInfo and share it globally. Let me know if I'm missing something.

go/libraries/doltcore/schema/typeinfo/point.go
```diff
@@ -45,13 +46,17 @@ func (ti *pointType) ConvertNomsValueToValue(v types.Value) (interface{}, error)
 	if _, ok := v.(types.Null); ok || v == nil {
 		return nil, nil
 	}
-
+	var err = fmt.Errorf(`"%v" cannot convert NomsKind "%v" to a value`, ti.String(), v.Kind())
 	// Expect a types.Point, return a sql.Point
 	if val, ok := v.(types.Point); ok {
-		return ConvertTypesPointToSQLPoint(val), nil
+		sqlVal := ConvertTypesPointToSQLPoint(val)
+		err = ti.sqlPointType.MatchSRID(sqlVal)
```
fulghum, line 53: I'm wondering if we actually **need** to validate the SRID here... we definitely need to validate any new data coming into the database from the caller (of course) and we need to validate any schema changes to make sure they don't invalidate the SRID matching constraint, but I'm wondering if we can be more lenient when we're loading data into and out of the storage layer interface. The impact of this validation failing is that customer won't be able to load their (already entered into the system) data any more, which could be worse than it being loaded and having inconsistent SRIDs that don't fit the constraint of the column. 

Are we testing this constraint in enough other places where the data is coming into the system so that we don't need to test this again as the data goes into the storage layer? I'm not confident yet, but wanted to bring this up to consider.

go/libraries/doltcore/row/noms_row.go
```diff
@@ -78,6 +80,12 @@ func pkRowFromNoms(sch schema.Schema, nomsKey, nomsVal types.Tuple) (Row, error)
 		if col.IsPartOfPK {
 			return false, errors.New("writing columns that are part of the primary key to non-pk values. col:" + col.Name)
 		} else if !types.IsNull(val) {
+			if err = validateSpatialTypeSRID(col, val); err != nil {
```
fulghum, line 83: It feels odd to be doing this spatial type validation at this level. This file is really all about noms io interface and feels like not the right place to do validation on specific types.

fulghum (commented):
Looking pretty good. I think we can simplify the SRID validation quite a bit and potentially avoid doing it at this level and rely on the validation at higher levels.

go/libraries/doltcore/schema/typeinfo/point.go
```diff
@@ -208,3 +209,25 @@ func pointTypeConverter(ctx context.Context, src *pointType, destTi TypeInfo) (t
...
 	}
 }
+
+func CreatePointTypeFromParams(params map[string]string) (TypeInfo, error) {
+	var (
+		err     error
+		sridVal uint64
+		def     bool
+	)
+	if s, ok := params["SRID"]; ok {
```
jennifersp, line 219: adding check here because the previously created tables with geometry type will not have value in it, so starting server fails.

fulghum: smart; good job thinking about backwards compatibility with existing databases.

go/libraries/doltcore/sqle/alterschema.go
```diff
@@ -272,6 +275,8 @@ func updateTableWithModifiedColumn(ctx context.Context, tbl *doltdb.Table, oldSc
 			val, ok := r.GetColVal(modifiedCol.Tag)
 			if !ok || val == nil || val == types.NullValue {
 				return true, fmt.Errorf("cannot change column to NOT NULL when one or more values is NULL")
+			} else if err = validateSpatialTypeSRID(modifiedCol, val); err != nil {
```
fulghum, line 278: Should this validation check be inside the `if !modifiedCol.IsNullable()` block? 

It seems like we only need to run this validation when a column type is changing, and you've got that covered in the if block right above this. 

This block seems like it's specific for checking that a modified column that does not allow nulls doesn't have any existing nulls in the data. I'm not sure we need to do anything for SRID here, but let me know if I'm missing something.

fulghum (commented):
Code is looking great! I just had one question I wanted to make sure I understood about the validation logic in alterschema.go. 

Thanks again for making the `LineString` capitalization change to be consistent with MySQL!

fulghum (approved):
Nice work on these!

PR #3460, opened 2022-05-19
go/libraries/doltcore/merge: allow merges in an inconsistent state

druvv:
Merges are now allowed in an inconsistent state, i.e. conflicts and constraint violations can exist in the working root. 

## Design
Conflicts are stashed from the left root, then applied to the merged root. If the merged root and stash contain conflicts, then we abort the merge.

Constraint violations have to be stashed from the ancestor, then merged with any violations generated during the merge. We currently generate violations by diffing tables involved in foreign keys between the merged root and the ancestor root. This may result in constraint violations that already existed in left to be regenerated during the merge process. This is why we have to stash from the ancestor.

## Summary of changes
- `CALL DOLT_MERGE` returns `0` if conflicts OR constraint violations exist. Previously it only returned `0` if conflicts existed, although it seems as though the intention was to also return `0` for constraint violations.
- Adds bats and sql engine tests for new behavior
- Updates broken tests

## Future work
- Update `CALL DOLT_MERGE` and `dolt merge` to only report conflicts / constraint violations if those inconsistencies where generated in the merge. Existing inconsistencies shouldn't be reported. (We may show a warning in the CLI case).
- Allow multiple merges in the working root, currently if a merge produces conflicts / constraint violations, they must be committed to perform another merge.


andrew-wm-arthur:
> Constraint violations have to be stashed from the ancestor, then merged with any violations generated during the merge. We currently generate violations by diffing tables involved in foreign keys between the merged root and the ancestor root. This may result in constraint violations that already existed in left to be regenerated during the merge process. This is why we have to stash from the ancestor.

Wondering about the necessity of this logic. What is the issue if we regenerate violations that already existed? Can we not reconcile this on violation merge?

andrew-wm-arthur (approved):
LGTM!

druvv:
> > Constraint violations have to be stashed from the ancestor, then merged with any violations generated during the merge. We currently generate violations by diffing tables involved in foreign keys between the merged root and the ancestor root. This may result in constraint violations that already existed in left to be regenerated during the merge process. This is why we have to stash from the ancestor.
> 
> Wondering about the necessity of this logic. What is the issue if we regenerate violations that already existed? Can we not reconcile this on violation merge?

We can definitely reconcile this on violation merge... let me revisit in the next PR to clean it up.

PR #3465, opened 2022-05-20
Refactoring MultiRepoEnv to use a map to store envs by name

fulghum:
As discussed in https://github.com/dolthub/dolt/pull/3444, this is a follow up to simplify MultiRepoEnv by moving to track envs in a map.


fulghum:
Removing the ordering guarantee for MultiRepoEnv's Iteration method resulted in some of the BATS tests to have flakey failures around diff output. I was able to fix the failing tests by adding back in a list to ensure iteration in the order environments were added to the MultiRepoEnv, but having to track the order means it makes less sense to refactor this from a list of NamedEnvs to a map of environments by name, so closing this one out.

PR #3466, opened 2022-05-20
Zachmu/ddl3

zachmu:


go/libraries/doltcore/sqle/alterschema.go
```diff
@@ -778,6 +779,67 @@ func dropColumn(ctx context.Context, tbl *doltdb.Table, colName string) (*doltdb
 	return tbl.UpdateSchema(ctx, newSch)
 }
 
+// backupFkcIndexesForKeyDrop finds backup indexes to cover foreign key references during a primary
+// key drop. If multiple indexes are valid, we sort by unique and select the first.
+// This will not work with a non-pk index drop without an additional index filter argument.
+func backupFkcIndexesForPkDrop(ctx *sql.Context, sch schema.Schema, fkc *doltdb.ForeignKeyCollection) ([]doltdb.FkIndexUpdate, error) {
```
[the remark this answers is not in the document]
Hydrocharged: After some more thought, I think this kind of backup-searching functionality belongs in GMS on the PK drop path. We have too much stuff in Dolt that should be in GMS, and we can unintentionally break things because of it. If we later adjust GMS and introduce a new interface or something dealing with PK alteration, then it's not going to be immediately clear that we're also affecting foreign keys (and also that foreign keys should even be considered in the first place). This is very similar to the issues with the recent foreign key breakage in the new format as well (from the `RewriteInserter` addition, Dolt just does too much and GMS too little).

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -123,6 +123,9 @@ func getSecondaryKeylessProllyWriters(ctx context.Context, t *doltdb.Table, sqlS
 
 // Insert implements TableWriter.
 func (w *prollyTableWriter) Insert(ctx *sql.Context, sqlRow sql.Row) error {
+	if err := w.primary.Insert(ctx, sqlRow); err != nil {
```
Hydrocharged, line 126: Why was this moved?

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -324,9 +325,6 @@ func (w *prollyTableWriter) table(ctx context.Context) (t *doltdb.Table, err err
 }
 
 func (w *prollyTableWriter) flush(ctx *sql.Context) error {
-	if !w.primary.HasEdits(ctx) {
```
Hydrocharged, line 327: I noticed that we call `flush` multiple times when we should only be calling it once. Was this causing errors?

Hydrocharged (approved):
LGTM! Did you try removing the skips on the `dolt_engine_test.go` foreign key tests?

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
Hydrocharged: Ahh I see

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
Hydrocharged: That's a good point. It's worth adding a comment.

PR #3471, opened 2022-05-23
Read replica update working set fixes

max-hoffman:
Read replica pull updates the session working set after pulling a
filtered set of branches from the tracking database. The original
(buggy) implementation updates the working set to the branch specified
at server-start time. The identity of that branch was fixed for the
duration of the server, so dolt_checkout would (appear to) have no
effect on the new branch's working set. What actually happened was more
pernicious: the working set was updated to the value of the
incorrect branch.

The fix no longer statically sets the active branch for a read replica
database. The active branch is pulled from the `*sql.Context`, so the
correct working set will be updated.

Note:
- Changing the remote name or remote endpoint would have a similar static state issue. We could read remote name and remote target from the session vars, repo state reader at pull time.

### DIFF
--- a/integration-tests/bats/replication.bats
+++ b/integration-tests/bats/replication.bats
@@ -255,7 +255,6 @@ teardown() {
     dolt config --local --add sqlserver.global.dolt_replicate_all_heads 1
     dolt config --local --add sqlserver.global.dolt_read_replica_remote origin
     # repo2 pulls on read
-    dolt branch
     run dolt sql -r csv -b <<SQL
 call dolt_checkout('feat');
 show tables;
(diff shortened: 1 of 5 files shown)


integration-tests/bats/replication.bats
```diff
@@ -232,6 +232,41 @@ teardown() {
...
+    dolt sql -q "create table t2 (a int)"
+    dolt commit -am "cm"
+    # remote1 has tables
+    dolt push remote1 feat
+    # repo2 replication config
+    cd ../repo2
+    dolt config --local --add sqlserver.global.dolt_replicate_all_heads 1
+    dolt config --local --add sqlserver.global.dolt_read_replica_remote origin
+    # repo2 pulls on read
+    dolt branch
```
zachmu, line 258: What's this line for?

zachmu (approved):
LGTM

integration-tests/bats/replication.bats
max-hoffman: thanks, good catch

PR #3475, opened 2022-05-23
[no-release-notes] Fix chunk corruption for keyless tables

andrew-wm-arthur:
Fixes a bug in `prollyKeylessWriter` where a cached chunk was being corrupted by an in-place edit to its buffer.


zachmu (approved):
LGTM

PR #3486, opened 2022-05-25
[no-release-notes] fbs schema encoding

andrew-wm-arthur:
This proposal lays out a serialization format for table schemas in `__DOLT_1__`. The primary changes from the existing schema serialization format are:
- Storing SQL type definitions as strings rather than `typeinfo` params.
- Removing columns tags and reconciling column identity by name. This implies that column renames will be identified heuristically. 
- Making the serialized schema better represent how rows/indexes are store. Ie storing a column for keyless table "id columns" (and other columns not visible to the SQL layer).


go/serial/schema.fbs
```diff
@@ -12,73 +12,55 @@
...
+  columns:[Column];
+  indexes:[Index];
+  checks:[string];
 
-table ColumnDefault {
-  expression:string (required);
 }
 
-table ColumnConstraint {
+table Column {
```
zachmu, line 27: Where's the extended type info?

PR #3487, opened 2022-05-25
Adding support for deleting and renaming branches from SQL with `dolt_branch`

fulghum:
Also added more go tests to cover `dolt_branch` functionality, cleaned up the code a little bit, and fixed a few other gaps in functionality I came across (e.g. --force option wasn't always being honored).

The only snag I hit is that the interfaces for the CLI functionality required a `DoltEnv` struct and we don't have a good way to get that from within a SQL function/procedure. I was able to change the interfaces a bit to not require `DoltEnv`, but I did still have to do one weird thing in order to get a `DoltCLIConfig` struct, but I was able to keep it pretty isolated and clearly documented in one place in the code. Happy to hear feedback if there's a better way to tackle that. 

Depends on: https://github.com/dolthub/go-mysql-server/pull/1034 ✅
Doc updates: https://github.com/dolthub/docs/pull/625


go/libraries/doltcore/sqlserver/dolt_server.go
```diff
@@ -0,0 +1,45 @@
...
+	"sync"
+
+	"github.com/dolthub/go-mysql-server/server"
+)
+
+var mySQLServer *server.Server
+var mySQLServerMutex sync.Mutex
+
+// RunningInServerMode returns true if the current process is running a SQL server.
+func RunningInServerMode() bool {
```
zachmu, line 27: This is kind of gross although I can see why it's necessary.

We need to do a pass and figure out how to incorporate session management and other server concepts into the engine better.

zachmu (approved):
LGTM, but we need to think about sane interfaces for giving the runtime access to session management

This is also all incredibly racy, but we can't really address that until we fix the above, so we can ignore for now

PR #3493, opened 2022-05-26
[no-release-notes] Prolly Encodings for `Enum`, `Set` and `Bit`

andrew-wm-arthur:


go/libraries/doltcore/sqle/index/prolly_fields.go
```diff
@@ -22,12 +22,62 @@ import (
...
 	"github.com/shopspring/decimal"
 
-	"github.com/dolthub/dolt/go/libraries/doltcore/schema/typeinfo"
 	geo "github.com/dolthub/dolt/go/store/geometry"
-	"github.com/dolthub/dolt/go/store/types"
 	"github.com/dolthub/dolt/go/store/val"
 )
 
+// todo(andy): this should go in GMS
+func DenormalizeRow(sch sql.Schema, row sql.Row) (sql.Row, error) {
```
Hydrocharged, line 30: These feel like the worst things ever but I guess it's okay for now. GMS should absolutely be taking care of this.

Hydrocharged (approved):
I can live with it, those row normalization functions don't make me happy lol

go/libraries/doltcore/sqle/index/prolly_fields.go
andrew-wm-arthur: Yeah I agree. We need another pass to move this logic to GMS

PR #3498, opened 2022-05-28
format databases map key for case-insensitive db name and case-sensitive branch name

jennifersp:
Database name is case-insensitive, but branch name is case sensitive.


fulghum (approved):
Looks good. Just a couple minor comments.

PR #3499, opened 2022-05-28
Add filter option for diff command

### LINKED ISSUE #1430: `dolt diff` should support `--filter` option
`--filter` could be arbitrability complex but I think we start with `<inserts | updates | deletes>` as the first options accepted. This option filters the diff down to modifications of those types.

This is particularly useful in human review of changes. It is common for inserts to be ok but want to add extra scrutiny to updates or deletes. If the diff is large, this becomes unwieldy. `| grep` for `+` or` -` does not help because updates have an insert and delete.

ksankeerth:
This PR closes #1430 

Description : 

The following assumptions were taken  for diff filter.

**For docs** 

* adding new doc will be considered as "insert"
* modifying existing file will be considered as "update"
* removing a file will be considered as "update" and "delete" (Removing lines = update and removing docs table = delete)

**For Table**
* create a table = insert
* adding rows = insert
* modifying row = update
* delete row = delete
* adding a column = update
* changing data type of a column = update
* deleting a column = update 
* renaming a column = update 

The `filter` flag  is not allowed with `--summary`

**Tests** 

* added tests for docs and tables in diff.bats

Please check the changes and let me know your feedback.
Thanks


go/cmd/dolt/commands/diff.go
```diff
@@ -145,6 +153,7 @@ func (cmd DiffCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(whereParam, "", "column", "filters columns based on values in the diff.  See {{.EmphasisLeft}}dolt diff --help{{.EmphasisRight}} for details.")
 	ap.SupportsInt(limitParam, "", "record_count", "limits to the first N diffs.")
 	ap.SupportsFlag(CachedFlag, "c", "Show only the unstaged data changes.")
+	ap.SupportsString(filterParam, "", "diff_type", "filters(inserts|updates|deletes) based on type of modifications.")
```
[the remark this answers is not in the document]
timsehn: I'm fine with "added/modified/removed".

go/cmd/dolt/commands/diff.go
```diff
@@ -480,30 +507,48 @@ func printShowCreateTableDiff(ctx context.Context, td diff.TableDelta) errhand.V
 	}
 
 	if fromCreateStmt != toCreateStmt {
-		cli.Println(textdiff.LineDiff(fromCreateStmt, toCreateStmt))
+		if fromCreateStmt == "" && toSch.GetAllCols().Size() > fromSch.GetAllCols().Size() {
```
fulghum, line 510: If we go with diff_types of added, modified, removed... this could probably be simplified a bit...

You could use `td.IsAdd() && (filter == FilterAdded || filter == NoFilter)` to identify added tables, `td.Drop()` to identify dropped tables, and just rely on everything else being a table modification (e.g. renames, column adds/deletes, etc). Slightly different than the insert/add/delete mapping you proposed, but it would match up better with the existing Dolt diff_types. 

Also... it might be worth creating a small private struct in this file for the filter. Then you could clean up some of the filter checking logic a bit, make the code read a little smoother, and not have to pass around a string as the filter. For example, I'm imagining usage like: 
> if (td.Add() && filter.includeAdds()) { 

That would give a little more typing/structure and also removes the need to check for `NoFilter` directly in many places.

go/cmd/dolt/commands/diff.go
```diff
@@ -480,30 +507,48 @@ func printShowCreateTableDiff(ctx context.Context, td diff.TableDelta) errhand.V
...
-func sqlSchemaDiff(ctx context.Context, td diff.TableDelta, toSchemas map[string]schema.Schema) errhand.VerboseError {
+func sqlSchemaDiff(ctx context.Context, td diff.TableDelta, toSchemas map[string]schema.Schema, filter string) errhand.VerboseError {
 	fromSch, toSch, err := td.GetSchemas(ctx)
 	if err != nil {
 		return errhand.BuildDError("cannot retrieve schema for table %s", td.ToName).AddCause(err).Build()
 	}
 
 	if td.IsDrop() {
-		cli.Println(sqlfmt.DropTableStmt(td.FromName))
+		if filter == NoFilter || filter == FilterDeletes {
```
fulghum, line 534: (minor) If you go with the `diffTypeFilter` struct idea above, this could be simple enough to roll into the existing if statement above; something like `if td.IsDrop() && diffTypeFilter.includeDrops() {`

go/cmd/dolt/commands/diff.go
```diff
@@ -927,6 +1070,7 @@ func printTableDiffSummary(td diff.TableDelta) {
 		_, _ = bold.Printf("diff --dolt a/%s b/%s\n", td.ToName, td.ToName)
 		_, _ = bold.Println("added table")
 	} else {
+
```
fulghum, line 1073: (minor) This extra newline can be trimmed out please.

go/cmd/dolt/commands/diff_filter.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/dolthub/dolt/go/store/types"
+)
+
+type DiffFilterTransform struct {
+	Pipeline     *pipeline.Pipeline
+	filterBy     string
+	newColsCount int
+}
+
+func NewDiffFilterTrans(filterBy string, newColsCount int) *DiffFilterTransform {
```
fulghum, line 29: (minor) Might as well add the "form" at the end just to be complete and consistent with other method naming. Four extra chars isn't too big of a deal anyway.

go/cmd/dolt/commands/diff_filter.go
```diff
@@ -0,0 +1,60 @@
...
+	taggedVals, err := inRow.TaggedValues()
+	if err != nil {
+		return nil, ""
+	}
+
+	joinedRowAsMap := map[uint64]types.Value(taggedVals)
+
+	v, ok := joinedRowAsMap[0]
+	if v != nil && ok {
+		if df.newColsCount < len(joinedRowAsMap) {
```
fulghum, line 44: This should also simplify nicely if we look at `diff_type` as added/modified/removed and treat any schema change as modified, table create as add, and table delete as remove. I suspect this _might_ be easier/cleaner if you do this check when you have a `TableDelta` object (instead of in this pipeline step).

ksankeerth:
@fulghum Thanks for the feedback and review. I'll go through them and push the changes. Yeah, If We consider added/modified/removed instead of inserts/updates/deletes, it going to be very simple

ksankeerth:
Hi @fulghum,

Could you please have look at the [latest commit](https://github.com/dolthub/dolt/pull/3499/commits/9fdfebc4f94ed8390f37a8f14444e38da5f96c08) when you have time. Let me know your feedback.

Previously, I filtered each table row based on inserts/updates/deletes/. Now I use td.IsAdd() and td.IsDrop() to identity the diff type,  so it seems that not necessary to have the transform.

timsehn:
Jason is on vacation today so likely, he'll get to this Monday.

timsehn:
I'm going to close this. Reopen if you ever revisit the feedback.

PR #3506, opened 2022-05-31
go/libraries/doltcore/dbfactory: aws.go: Load shared config by default, instead of requiring the user to set AWS_SDK_LOAD_CONFIG=1.

reltuk:


integration-tests/bats/remotes-aws.bats
```diff
@@ -52,6 +52,14 @@ skip_if_no_aws_tests() {
     dolt sql -q 'show tables'
 }
 
+@test "remotes-aws: can clone an existing aws remote with AWS_SDK_LOAD_CONFIG=1 set." {
```
fulghum, line 55: did you mean "without AWS_SDK_LOAD_CONFIG=1 set" ?

reltuk, line 55:
```suggestion
@test "remotes-aws: can clone an existing aws remote without AWS_SDK_LOAD_CONFIG=1 set." {
```

PR #3514, opened 2022-06-01
Apply index lookups on diff tables

JCOR11599:
`diff_table` now implements `sql.IndexedTable` interface and uses dolt lookups.
Currently treats `to_` columns as primary key, so index lookups on work on those.

Fix for: https://github.com/dolthub/dolt/issues/3438


go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -127,6 +134,7 @@ func (dt *DiffTable) Schema() sql.Schema {
 }
 
 func (dt *DiffTable) Partitions(ctx *sql.Context) (sql.PartitionIter, error) {
+	// TODO: if this were in parallel it can't always start at HEAD
```
zachmu, line 137: This call doesn't happen in parallel. We get the partitions, then iterate them in parallel.

zachmu (approved):
LGTM

PR #3528, opened 2022-06-03
Fixed foreign key parent referencing + keyless referencing

Hydrocharged:
We, apparently, don't have any tests for referencing the primary key in a foreign key, so those have been failing in the new format. In addition, some changes ended up breaking new format keyless, which was also not caught due to lack of tests. Both of those issues have been fixed.

Tests have been added to GMS: https://github.com/dolthub/go-mysql-server/pull/1038

### DIFF
--- a/go/libraries/doltcore/sqle/writer/prolly_table_writer.go
+++ b/go/libraries/doltcore/sqle/writer/prolly_table_writer.go
@@ -39,6 +39,7 @@ type prollyTableWriter struct {
 
 	primary   indexWriter
 	secondary []indexWriter
+	secNames  map[string]int
 
 	tbl    *doltdb.Table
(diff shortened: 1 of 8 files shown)


go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -39,6 +39,7 @@ type prollyTableWriter struct {
 
 	primary   indexWriter
 	secondary []indexWriter
+	secNames  map[string]int
```
andrew-wm-arthur, line 42: let's collapse these into one collection
```suggestion
	secondary map[string]indexWriter
```

andrew-wm-arthur (approved):

PR #3535, opened 2022-06-03
change dolt_diff system table to coerce schema by name only

druvv:
The dolt_diff_{table_name} system table now coerces schema by name only. The dolt_diff table uses the schema at head to interpret the tables at every prior commit. Previously the schema of the table at each prior commit was matched to the head schema using tags with a name fallback. Now it is matched by name only.

Let's consider an example between the new and old behaviors. Define table `t` with the following schema:
```sql
CREATE TABLE t (
    pk int PRIMARY KEY,
    a text
);
```

In commit1 we insert the following row:
```
INSERT INTO T VALUES (1, 'one');
```

In commit2 we rename col1 to col2:
```
ALTER TABLE T RENAME col1 TO col2
```

In the old behavior, we would see the following diffs. 

Between init and commit1:
```
+ (1, 'one')
```

No diff between commit1 and commit2.

In the new behavior, we would see the following diffs.

Between init and commit1:
```
+ (1, NULL)
```

No diff between commit1 and commit2. (We actually do expect to see `~ (1, 'one')` here but this is a TODO and known issue.)

Also adds warnings to new storage format when a column value coercion fails.

### DIFF
--- a/go/libraries/doltcore/rowconv/field_mapping.go
+++ b/go/libraries/doltcore/rowconv/field_mapping.go
@@ -261,6 +261,36 @@ func TagMappingWithNameFallback(srcSch, destSch schema.Schema) (*FieldMapping, e
 	return NewFieldMapping(srcSch, destSch, srcToDest)
 }
 
+// TagMappingByName takes a source schema and a destination schema and maps
+// columns by matching names.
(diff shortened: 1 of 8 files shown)


zachmu (approved):
LGTM

PR #3544, opened 2022-06-03
[no-release-notes] add immutable chunker

max-hoffman:


go/store/prolly/tree/immutable_tree.go
```diff
@@ -0,0 +1,151 @@
...
+	"io"
+
+	"github.com/dolthub/dolt/go/store/prolly/message"
+)
+
+const defaultFixedChunkLength = 4000
+
+var ErrInvalidChunkSize = errors.New("invalid chunkSize; value must be > 1")
+
+// buildImmutableTree writes the contexts of |reader| as an append-only
```
andrew-wm-arthur, line 29: contents

go/store/prolly/tree/immutable_tree.go
```diff
@@ -0,0 +1,151 @@
...
+// tree, returning the root node or an error if applicable. |chunkSize|
+// fixes the split size of leaf and intermediate node chunks.
+func buildImmutableTree(ctx context.Context, r io.Reader, ns NodeStore, S message.Serializer, chunkSize int) (Node, error) {
+	if chunkSize <= 1 {
```
andrew-wm-arthur, line 33: 1 byte is invalid?

go/store/prolly/tree/immutable_tree_test.go
```diff
@@ -0,0 +1,198 @@
...
+				}
+				return nil
+			})
+
+			require.Equal(t, expLevel, root.Level())
+			require.Equal(t, expSum, sum)
+			require.Equal(t, tt.inputSize, byteCnt)
+			require.Equal(t, expUnfilled, unfilledCnt)
+			require.Equal(t, expSubtrees, root.getSubtreeCounts())
+		})
```
andrew-wm-arthur, line 135: you should add a validation step here that asserts we can round-trip `buf`

andrew-wm-arthur (approved):
Generally looks good, tree construction logic appears correct. 

The `ProllyTreeNode` flatbuffers message, and the `Serialize` interface are both oriented towards `Tuple` indexes. You should follow up on this PR to create a dedicated message and serialization path. It might also be worthwhile to move blobs into their own package.

go/store/prolly/tree/immutable_tree.go
max-hoffman: 1 byte is OK and tested, chunkSize 1 causes infinite recursion (can't be tree)

go/store/prolly/tree/immutable_tree_test.go
max-hoffman: `sum` validates the contents were written as expected, I was planning on round-tripping in the next PR that adds a struct for read/writing blobs at the tuple layer

PR #3548, opened 2022-06-06
[no-release-notes] only create privilege file, when specified with option

JCOR11599:
TODO (next pr): have `show grants on user` work


integration-tests/bats/sql.bats
```diff
@@ -41,7 +41,7 @@ teardown() {
 
 @test "sql: dolt sql -q has mysql db and can create users" {
```
zachmu, line 42: Write a version of this test where you run `dolt sql` in a directory without having done `dolt init` first, then run `create database` and grant some permissions on that new database.

(Want to make sure this works when the `.dolt` directory doesn't exist already)

zachmu (approved):
LGTM

zachmu (approved):
LGTM

PR #3549, opened 2022-06-06
Bug fix for not saving explicit name and email settings to local config

fulghum:
Fixes: https://github.com/dolthub/dolt/issues/3546

### DIFF
--- a/go/cmd/dolt/commands/init_test.go
+++ b/go/cmd/dolt/commands/init_test.go
@@ -16,18 +16,21 @@ package commands
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
 )
 
+type initTest struct {
+	Name          string
+	Args          []string
+	GlobalConfig  map[string]string
+	ExpectSuccess bool
+}
+
 func TestInit(t *testing.T) {
-	tests := []struct {
-		Name          string
-		Args          []string
-		GlobalConfig  map[string]string
-		ExpectSuccess bool
-	}{
+	tests := []initTest{
 		{
 			"Command Line name and email",
 			[]string{"-name", "Bill Billerson", "-email", "bigbillieb@fake.horse"},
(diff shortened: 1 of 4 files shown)


go/cmd/dolt/commands/init_test.go
```diff
@@ -96,3 +101,37 @@ func TestInitTwice(t *testing.T) {
...
+	for i := 0; i <= len(test.Args)-2; i = i + 2 {
+		if test.Args[i] == "-"+argKey {
+			found = true
+			expectedValue = test.Args[i+1]
+		}
+	}
+
+	actualValue, err := localConfig.GetString(envKey)
+	if found {
+		if err != nil {
```
zachmu, line 125: Nit: try not to write any new tests without testify

integration-tests/bats/init.bats
```diff
@@ -0,0 +1,142 @@
...
+  [[ "$output" =~ "foo" ]] || false
+
+  run dolt config --local --get user.email
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "foo@bar.com" ]] || false
+
+  assert_valid_repository
+}
+
+@test "init: no explicit or implicit configuration for name and email" {
```
zachmu, line 79: Seems like we're missing a test where `user.name` and `user.email` aren't set, but are given to `dolt init`?

zachmu (approved):
LGTM, thanks for fixing!

go/cmd/dolt/commands/init_test.go
fulghum: ooooh... good call. I migrated this unit test over to stretchr/testify and it's cleaner/leaner now.

integration-tests/bats/init.bats
fulghum: I was covering a similar case where global config was set and tested that explicit config was saved to local config, but you're right that it's worth also covering when no global config is present at all. Added now.

PR #3554, opened 2022-06-06
add bats test for procedure alias not showing in information_schema.routines table

jennifersp:
Added bats tests for procedure aliases.
Depends on https://github.com/dolthub/go-mysql-server/pull/1042


timsehn (approved):
LGTM on both

PR #3557, opened 2022-06-07
checkout proc startpoint

max-hoffman:


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -1433,6 +1433,31 @@ var DoltBranchScripts = []queries.ScriptTest{
 			},
 		},
 	},
+	{
+		Name: "Create branch from startpoint",
+		SetUpScript: []string{
+			"create table a (x int)",
+			"CALL DOLT_COMMIT('--allow-empty', '-am', 'add table a')",
```
zachmu, line 1440: Do this with a commit hash too. You can store a commit hash in a user var and then use it later in the script

zachmu (approved):
LGTM nice catch

PR #3563, opened 2022-06-07
Adding working set changes to unscoped `dolt_diff` system table 

fulghum:
Doc updates: https://github.com/dolthub/docs/pull/631

### DIFF
--- a/go/libraries/doltcore/diff/table_deltas.go
+++ b/go/libraries/doltcore/diff/table_deltas.go
@@ -234,10 +234,12 @@ func (td TableDelta) IsRename() bool {
 }
 
 // HasHashChanged returns true if the hash of the table content has changed between
-// the fromRoot and toRoot. Note that before calling HasHashChanged, you need to
-// make sure the TableDelta doesn't represent a table add or drop change, since
-// those cases won't have both FromTable and ToTable set and will return an error.
+// the fromRoot and toRoot.
 func (td TableDelta) HasHashChanged() (bool, error) {
+	if td.IsAdd() || td.IsDrop() {
+		return true, nil
+	}
+
 	toHash, err := td.ToTable.HashOf()
 	if err != nil {
 		return false, err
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -247,6 +251,24 @@ func (td TableDelta) HasHashChanged() (bool, error) {
 	return !toHash.Equal(fromHash), nil
 }
 
+// HasSchemaChanged returns true if the table schema has changed between the
```
zachmu, line 254: Is it more useful to define the add / drop case as true? That seems like a schema change to me

zachmu: Same question for above method

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
```diff
@@ -74,18 +80,101 @@ func (dt *UnscopedDiffTable) Schema() sql.Schema {
...
+	} else if bytes.Equal(partition.Key(), commitHistoryPartitionKey) {
+		return dt.newCommitHistoryRowItr(ctx)
+	} else {
+		return nil, fmt.Errorf("unexpected partition: %v", partition)
+	}
+}
+
+func (dt *UnscopedDiffTable) newWorkingSetRowItr(ctx *sql.Context) (sql.RowIter, error) {
+	sess := dsess.DSessFromSess(ctx.Session)
+	rootsProvider := dsess.NewSessionStateAdapter(sess.Session, ctx.GetCurrentDatabase(), map[string]env.Remote{}, map[string]env.BranchConfig{}, map[string]env.Remote{})
```
zachmu, line 105: Don't think you need to go to all this trouble, just call `sess.GetRoots(dbName)`

zachmu (approved):
LGTM!

go/libraries/doltcore/sqle/dtables/unscoped_diff_table.go
fulghum: ooooh! that's **soo** much better! thanks!

go/libraries/doltcore/diff/table_deltas.go
fulghum: good call; changed `HasSchemaChanged` and `HasHashChanged` so that they behave correctly even if you don't check for `IsAdd` and `IsDrop` first.

PR #3564, opened 2022-06-07
CALL DOLT_MERGE now returns fast_forward and conflicts columns

### LINKED ISSUE #3560: `call dolt_merge()` needs to return two things: fast-forward (yes/no) and conflicts (yes/no)
Right now it only returns `no_conflicts`. That should be changed to `conflicts`.

druvv:
The double negative `no_conflicts` column returned by `CALL DOLT_MERGE` and `SELECT DOLT_MERGE` has been replaced with a simpler `conflicts` column. `CALL DOLT_MERGE` additionally returns a `fast_forward` column which indicates if the merge was a fast-forward. If `fast_forward` is `0`, then you must perform a commit to reach a clean working set state. 

Closes #3560 

Replaces uses of `SELECT DOLT_MERGE` with `CALL DOLT_MERGE` in the engine tests. Both the sql function and procedure have similar test coverage. We currently don't support multiple return columns with sql functions, so the change was made to test the fast_forward column.


go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -47,101 +47,110 @@ type DoltMergeFunc struct {
...
 )
 
 func (d DoltMergeFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
 	args, err := getDoltArgs(ctx, row, d.Children())
 	if err != nil {
 		return noConflictsOrViolations, err
 	}
-	return DoDoltMerge(ctx, args)
+	hasConflicts, _, err := DoDoltMerge(ctx, args)
+	return hasConflicts, err
```
andrew-wm-arthur, line 65: so we only get the `fast_forward` info through the store procedure variant?

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -47,101 +47,110 @@ type DoltMergeFunc struct {
...
 const (
-	hasConflictsOrViolations int = 0
-	noConflictsOrViolations  int = 1
+	noConflictsOrViolations  int = 0
+	hasConflictsOrViolations int = 1
+)
+
+const (
+	threeWayMerge = 0
+	ff            = 1
```
andrew-wm-arthur, line 56: let's spell this out as `fastForwardMerge`

andrew-wm-arthur (approved):
LGTM, nice update of tests

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
druvv: correct

Issue #3567, opened 2022-06-08
`dolt table import` fails on `bit` columns


kretes:
Following repro case shows the problem:
```
># dolt init
Successfully initialized dolt data repository.

># dolt sql -q "CREATE TABLE bitted (id int PRIMARY KEY, b bit)"

># dolt sql -q "INSERT INTO bitted VALUES (1, 0), (3, 1)"
Query OK, 2 rows affected

># dolt table export bitted bitted.csv
Successfully exported data.

># cat bitted.csv 
id,b
1,0
3,1

># dolt table import -u bitted bitted.csv 

A bad row was encountered while moving data.
Bad Row: [3,1]
49 is beyond the maximum value that can be held by 1 bits
These can be ignored using '--continue'
```

VinaiRachakonda:
We have identified the issue and working on a fix. We will get out a fix by EOD and release accordingly. Thank you!

VinaiRachakonda:
This is now fixed and merged in main! I will get out a release shortly

PR #3570, opened 2022-06-08
Support bit types on import

VinaiRachakonda:
This pr addresses issue #3567 and additional tests for different types

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/import.go
+++ b/go/cmd/dolt/commands/tblcmds/import.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 
 	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/vitess/go/sqltypes"
 	"github.com/fatih/color"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/text/message"
@@ -784,8 +785,8 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema, 
 
 // detectAndConvertToBoolean determines whether a column is potentially a boolean and converts it accordingly.
 func detectAndConvertToBoolean(columnVal interface{}, columnType sql.Type) (bool, bool) {
-	switch columnType {
-	case sql.Boolean, sql.Int8, sql.MustCreateBitType(1): // TODO: noms bool wraps MustCreateBitType
+	switch columnType.Type() {
+	case sqltypes.Int8, sqltypes.Bit: // TODO: noms bool wraps MustCreateBitType
 		switch columnVal.(type) {
 		case int8:
 			return stringToBoolean(strconv.Itoa(int(columnVal.(int8))))
--- a/go/cmd/dolt/commands/tblcmds/import.go
+++ b/go/cmd/dolt/commands/tblcmds/import.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


integration-tests/bats/import-update-tables.bats
```diff
@@ -1017,3 +1017,149 @@ DELIM
...
     [[ "$output" =~ "All constraints are not satisfied" ]] || false
 }
+
+@test "import-update-tables: bit types" {
+    dolt sql -q "CREATE TABLE bitted (id int PRIMARY KEY, b bit)"
+    dolt sql -q "INSERT INTO bitted VALUES (1, 0), (3, 1)"
+
+    dolt table export bitted bitted.csv
+
+    run dolt table import -u bitted bitted.csv
```
fulghum, line 1027: (minor) is it worth using `-cf` to force recreation of the table instead of updating it in place? Just wondering if the data could be skipped and this test still pass.

integration-tests/bats/import-update-tables.bats
```diff
@@ -1017,3 +1017,149 @@ DELIM
...
+    run dolt table import -u bitted2 bitted.csv
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Rows Processed: 1, Additions: 0, Modifications: 0, Had No Effect: 1" ]] || false
+
+    run dolt sql -r csv -q "select * from bitted2 order by id"
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "id,b" ]] || false
+    [[ "$output" =~ "1,4" ]] || false
+
+    # Try with a binary value like 0x04
```
fulghum, line 1051: (minor) technically `0x04` is hex, which is great to test. Would be good to include a binary representation, too.

fulghum: you *might* even consider using a hex number that doesn't map directly to it's base-10 number, like `0xa` or `0x10` (I know that doesn't currently fit in the current bit width though.) That would just test a _little_ bit more of the hex case and give us a sanity check that we are handling it as hex.

integration-tests/bats/import-update-tables.bats
```diff
@@ -1017,3 +1017,149 @@ DELIM
...
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "Bad Row: [3,b'1001]" ]] || false
+}
+
+@test "import-update-tables: binary and varbinary types" {
+    # Varbinary column
+    dolt sql -q "create table t(pk int primary key, val varbinary(100))"
+    cat <<DELIM > binary.csv
+pk,val
+1,0xdsdd
```
fulghum, line 1091: isn't this an invalid hex string? I was surprised we saved this. MySQL has the same behavior here?

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -724,26 +725,47 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema,
...
+			// Check if the column can be parsed an uint64
+			val, err := strconv.ParseUint(colAsString, 10, 64)
+			if err == nil {
+				row[i] = val
+				continue
+			}
+
+			// Check if the column is of type b'110'
+			re := regexp.MustCompile(`(?m)b\'(\d+)\'`)
+			groups := re.FindStringSubmatch(colAsString)
```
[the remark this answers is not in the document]
fulghum: Could the regex be declared in a const? Table import may not be on the query execution path, but might be worth having that compiled regex cached somewhere so we don't recreate it over and over for large imports that use `BitType`.

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -724,26 +725,47 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema,
...
+					continue
+				}
+			}
+
+			// Check if the column can be parsed as a hex string
+			numberStr := strings.Replace(colAsString, "0x", "", -1)
+			val, err = strconv.ParseUint(numberStr, 16, 64)
+			if err == nil {
+				row[i] = val
+			}
```
fulghum, line 768: if none of these worked and `row[i]` still isn't set, we should probably return an error instead of continuing execution and making the problem harder to find.

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -724,26 +725,47 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema,
...
+		// Bit types need additional verification due to the differing values they can take on. "4", "0x04", b'100' should
+		// be interpreted in the correct manner.
+		case sql.BitType:
+			colAsString, ok := row[i].(string)
+			if !ok {
+				return nil, fmt.Errorf("error: column value should be of type string")
+			}
+
+			// Check if the column can be parsed an uint64
+			val, err := strconv.ParseUint(colAsString, 10, 64)
```
fulghum, line 746: Should this conversion logic be put somewhere more shared? For example, should `sql/bit.go::Convert` know these rules for converting a string to a bit type?

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -724,26 +725,47 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema,
...
-			}
+		// Check if this a string that can be converted to a boolean
+		val, ok := detectAndConvertToBoolean(row[i], col.Type)
+		if ok {
+			row[i] = val
+			continue
 		}
 
 		switch col.Type.(type) {
 		case sql.StringType:
```
fulghum, line 736: I know this line didn't change, but it still looked odd... do we need this, or should string types get picked up in the default case?  i.e. do we want string types to have empty strings converted to nils?  I suspect we may want to leave any literal empty string string types as empty string, but it would be good to add a short comment if that's the intent to make it explicit.

fulghum (commented):
Logic looked correct, but I had a few questions/suggestions to make sure I understood things correctly.

go/cmd/dolt/commands/tblcmds/import.go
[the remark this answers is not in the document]
VinaiRachakonda: Yea definitely.

integration-tests/bats/import-update-tables.bats
VinaiRachakonda: This was modeling the customer issue flow

integration-tests/bats/import-update-tables.bats
VinaiRachakonda: Sure! I did add binary representation in the form b'100' below.

go/cmd/dolt/commands/tblcmds/import.go
VinaiRachakonda: Hmmm I see your point. Eventually the error does get caught at the insert node. I could see that being a potential UX problem.

go/cmd/dolt/commands/tblcmds/import.go
VinaiRachakonda: I don't think so. This is pretty unique to the import path. Usually when inserting this data through the sql shell, the vitess parser will do string/type parsing especially when special characters like `b' '` are used

go/cmd/dolt/commands/tblcmds/import.go
VinaiRachakonda: Yea I can add a comment

integration-tests/bats/import-update-tables.bats
VinaiRachakonda: I see your point, I can add a better example. Generally varbinary refers to a particular encoding more than what is inserted: https://dev.mysql.com/doc/refman/8.0/en/binary-varbinary.html

integration-tests/bats/import-update-tables.bats
VinaiRachakonda: Added 0xa

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -752,6 +782,23 @@ func NameAndTypeTransform(row sql.Row, rowOperationSchema sql.PrimaryKeySchema,
 	return row, nil
 }
 
+// detectAndConvertToBoolean determines whether a column is potentially a boolean and converts it accordingly.
+func detectAndConvertToBoolean(columnVal interface{}, columnType sql.Type) (bool, bool) {
+	switch columnType {
+	case sql.Boolean, sql.Int8, sql.MustCreateBitType(1): // TODO: noms bool wraps MustCreateBitType
```
Hydrocharged, line 788: I think it makes more sense to use `columnType.Type()` here and the case statement should be `case querytypes.INT8, querytypes.BIT` (or whatever the actual variable names are)

Hydrocharged (approved):
LGTM!

PR #3571, opened 2022-06-08
support `dolt cherry-pick` - version 1

jennifersp:
This PR is the first version of `cherry-pick` support.
- Only a single commit is allowed with cherry-picking
- No flags or subcommands are supported
- No conflicts, leaves the working set clean if conflicts occur
- Only available on CLI
- No renaming or dropping a table in cherry-pick commit
- Cherry-picking a merge commit is not supported
- Always creates a new commit except there is no changes.


integration-tests/bats/cherry-pick.bats
```diff
@@ -0,0 +1,439 @@
...
+    run dolt cherry-pick branch1
+    [ "$status" -eq "1" ]
+    [[ "$output" =~ "duplicate key error" ]] || false
+}
+
+@test "cherry-pick: insert, update, delete rows and schema changes on non existent table in working set" {
+    dolt sql -q "CREATE TABLE branch1table (id int primary key, col1 int)"
+    dolt sql -q "INSERT INTO branch1table VALUES (9,8),(7,6),(5,4)"
+    dolt add -A
+    dolt commit -m "create table with rows"
```
fulghum, line 151: (minor) consider just using `dolt commit -am "<message>"` to shorten this test file. If that pattern appears 10 times, it's worth it to save those 10 lines of test code.

integration-tests/bats/cherry-pick.bats
```diff
@@ -0,0 +1,439 @@
...
+
+    dolt sql -q "INSERT INTO branch1table VALUES (1,2)"
+    dolt add -A
+    dolt commit -m "Insert a row"
+
+    dolt checkout main
+    run dolt cherry-pick branch1
+    [ "$status" -eq "0" ]
```
fulghum, line 159: If a patch cannot be applied, we should error out. Try this out in git and you'll see something like:
```
❯ git cherry-pick 24be215f406059113a13605ea1e2d864e637d6cf
CONFLICT (modify/delete): bar.txt deleted in HEAD and modified in 24be215 (editing bar.txt). Version 24be215 (editing bar.txt) of bar.txt left in tree.
error: could not apply 24be215... editing bar.txt
hint: after resolving the conflicts, mark the corrected paths
hint: with 'git add <paths>' or 'git rm <paths>'
hint: and commit the result with 'git commit'
```

The general principal should be that if a user asks to cherry pick a commit, we only return success if we were able to apply that commit successfully. 

For v1, it would be good to identify that there is a conflict, briefly explain it, then return a non-zero exit code, but leave the working set clean.

integration-tests/bats/cherry-pick.bats
```diff
@@ -0,0 +1,439 @@
...
+    dolt sql -q "INSERT INTO test VALUES (3, 'c')"
+    dolt add -A
+    dolt commit -m "Inserted 3"
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
```
fulghum, line 27: We should add a test for trying to cherry pick a merge commit. We should follow git's lead and error out if a merge commit is specified, since we don't know which of the two parents to use for calculating the diff. Git provides an option in the CLI to select which parent for a merge commit, but we can just mention that we don't support cherry-picking merge commits yet in the CLI docs somewhere for now.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+	workingRoot, err := dEnv.WorkingRoot(ctx)
+	if err != nil {
+		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+	}
+	headHash, err := headRoot.HashOf()
+	if err != nil {
+		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+	}
+	workingHash, err := workingRoot.HashOf()
```
fulghum, line 109: Seems like we also need to check `dEnv.StagedRoot(ctx)`? Would be good to have two separate bats tests – one for a working root change and one for a staged root change.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,466 @@
...
+	authorStr, ok := apr.GetValue(cli.AuthorParam)
+	if ok {
+		commitParams = append(commitParams, "--author", authorStr)
+	}
+
+	return CommitCmd{}.Exec(ctx, "commit", commitParams, dEnv)
+}
+
+// CherryPick returns updated Root value for current HEAD and commit message of cherry-picked commit.
+func CherryPick(ctx context.Context, dEnv *env.DoltEnv, wRoot *doltdb.RootValue, cherryStr string) (*doltdb.RootValue, string, error) {
```
fulghum, line 158: (minor) Does this function need to be exported yet? Better to start off as package private unless we need it exported.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,466 @@
...
+	updatedRoot, err := cherryPickACommit(ctx, dEnv.DoltDB, wRoot, cherryCommit, opts)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return updatedRoot, commitMsg, nil
+}
+
+// cherryPickACommit takes working root and cherry-pick commit and returns updated root with changes from cherry-pick commit applied.
+func cherryPickACommit(ctx context.Context, ddb *doltdb.DoltDB, headRoot *doltdb.RootValue, cherryCM *doltdb.Commit, opts editor.Options) (*doltdb.RootValue, error) {
```
fulghum, line 185: we've got a couple of very similar function names here... `cherryPick` and `cherryPickACommit`. Would be good to rename this second one to something more self-descriptive that helps readers easily understand the difference between the two methods.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,466 @@
...
+	var err error
+	// parentCommitRoot, cherryCommitRoot
```
fulghum, line 187: (minor) I couldn't figure out what this comment was for. Were they previous variable names that could be pulled out, or was it trying to express something?

fulghum (commented):
Good progress. I took a pretty close look over the tests and left some comments about behavior in different cases. I also skimmed over the command code, and will take a closer look at that in the next revision. It would be good to sync up with Dhruv and see if there are any common pieces of code between this work and his work that we could share. 

I'm still wondering if we should keep it simpler and error out on any schema changes for v1. My biggest thought there is that it could be confusing to customers and hard to explain that some schema change types are supported, but others are not. Let me know what you think. 

One last thought on testing... what's the behavior for keyless tables? I'm assuming the row differ throws an error? Would be good to have a test case covering that scenario.

fulghum (commented):
Looking great! I only had a few minor comments looking at the code. Moving over to the same merge logic *really* simplified the code! Nice job with that. 

I just wanna check out your branch and play with the code a little bit tomorrow and look over the tests a little closer and then I'll be happy to give my approval.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
+	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+)
+
+var cherryPickDocs = cli.CommandDocumentationContent{
+	ShortDesc: `Apply the changes introduced by an existing commit.`,
+	LongDesc: `
+Updates tables in the clean working set with changes introduced in cherry-picked commit and creates a new commit with applied changes.
```
fulghum, line 35:
```suggestion
Applies the changes from an existing commit and creates a new commit from the current HEAD. This requires your working tree to be clean (no modifications from the HEAD commit).
```

Just some minor word smithing. I cribbed from `git cherry-pick` a little bit and split the clean working tree requirement into a second sentence to make it read a little smoother.

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,262 @@
...
+
+Currently, schema changes introduced in cherry-pick commit are not supported. Row data changes are allowed with a table schema in working set
```
fulghum, line 37:
```suggestion
Cherry-picking merge commits or commits with schema changes is not currently supported. Row data changes are allowed as long as the two table schemas are exactly identical. 
```

(Could remove that next line, too)

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,262 @@
...
+
+var cherryPickDocs = cli.CommandDocumentationContent{
+	ShortDesc: `Apply the changes introduced by an existing commit.`,
+	LongDesc: `
+Updates tables in the clean working set with changes introduced in cherry-picked commit and creates a new commit with applied changes.
+
+Currently, schema changes introduced in cherry-pick commit are not supported. Row data changes are allowed with a table schema in working set 
+matching a table schema in cherry-pick commit with the same name. 
+
+If there is a conflict, the working state stays clean. Cherry-picking a merge commit is not supported.
```
fulghum, line 40:
```suggestion
If applying the row data changes from the cherry-picked commit results in a data conflict, the cherry-pick operation is aborted and no changes are made to the working tree or committed. 
```

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -0,0 +1,262 @@
...
+
+dolt cherry-pick {{.LessThan}}commit{{.GreaterThan}}
```
fulghum, line 42: I'm not sure you need this section in the LongDesc. Looking at some other command documentation (e.g. status, commit, merge), I didn't see something similar. I think the

fulghum (commented):
Left some more fine tuning feedback on the help documentation. 

I also did some manual testing and it's a fun little command! Works very well. I noticed a couple small things:
* I was able to cherry-pick a commit with a schema change. I added a column to a table on a branch and then was able to successfully cherry pick that commit and alter the schema on another branch. Seems like the schema equality check may not be working? 
* I also noticed that there's a `fatal: cherry-pick failed` message printed in some error cases (e.g. working set changes). It would be a little cleaner to not print a fatal (feels kinda crashy). I didn't see other commands print that, but let me know if I'm missing something and that's common for other dolt commands.
* have you tested other, less-common commit types, like creating a stored procedure or a trigger? 
* Some of the error messages could be a little clearer. This may be partially to do with the schema equivalence checking. For example, when trying to cherry-pick a commit with a table rename, I got something like `conflict: table with same name deleted and modified`. A more direct message just saying that schema changes and renames aren't supported currently would be nice. Bonus points for sending people to GitHub issues to ask for us to prioritize support. 😉

go/libraries/doltcore/merge/merge.go
```diff
@@ -145,6 +145,14 @@ func (merger *Merger) MergeTable(ctx context.Context, tblName string, opts edito
 
 		// Deleted in root or in merge, either a conflict (if any changes in other root) or else a fast-forward
 		if ancHasTable && (!rootHasTable || !mergeHasTable) {
+			if isCherryPick && rootHasTable && !mergeHasTable {
+				// TODO : this is either drop table or rename table case
+				// We can delete only if the table in current HEAD and parent commit contents are exact the same (same schema and same data);
+				// otherwise, return ErrTableDeletedAndModified
+				// We need to track renaming of a table --> the renamed table could be added as new table
+				return nil, &MergeStats{Operation: TableModified}, errors.New(fmt.Sprintf("error: %s table was renamed or dropped", tblName))
```
fulghum, line 153: Maybe make this error message very similar to the no schema changes allowed message? 

```suggestion
				return nil, &MergeStats{Operation: TableModified}, errors.New(fmt.Sprintf("schema changes not supported: %s table was renamed or dropped in cherry-pick commit", tblName))
```

fulghum (approved):
Looks beautiful! Thanks for tightening up those final details today. Awesome new feature for customers! 🎉

PR #3577, opened 2022-06-08
[no-release-notes] doltcore/sqle/enginetest: Unskip some new format enginetests

andrew-wm-arthur:


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -254,12 +242,8 @@ func TestScripts(t *testing.T) {
 		skipped = append(skipped,
 			// Different error output for primary key error
 			"failed statements data validation for INSERT, UPDATE",
-			// missing FK violation
```
zachmu, line 257: There are duplicate skips in the TestScriptsPrepared method

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -254,12 +242,8 @@ func TestScripts(t *testing.T) {
 		skipped = append(skipped,
 			// Different error output for primary key error
 			"failed statements data validation for INSERT, UPDATE",
```
zachmu, line 244: Does your change below not address this discrepancy? Should be simple to make the two formats match their string output for this errror

zachmu (approved):
LGTM

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
andrew-wm-arthur: there was another error that failed this test. will circle back

PR #3582, opened 2022-06-09
[no-release-notes] /doltcore/sqle/enginetest: Added alternative query plans for `__DOLT_1__`

andrew-wm-arthur:
companion PR: https://github.com/dolthub/go-mysql-server/pull/1052


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3217,3 +3217,2359 @@ var DoltDiffPlanTests = []queries.QueryPlanTest{
 			"",
 	},
 }
+
+var NewFormatQueryPlanTests = []queries.QueryPlanTest{
```
zachmu, line 3221: Put these in their own file plz

zachmu (approved):

PR #3588, opened 2022-06-09
New format blobs

max-hoffman:


go/store/prolly/message/prolly_map.go
```diff
@@ -63,6 +65,10 @@ func (s ProllyMapSerializer) Serialize(keys, values [][]byte, subtrees []uint64,
 		valTups = writeItemBytes(b, values, valSz)
 		serial.ProllyTreeNodeStartValueOffsetsVector(b, len(values)-1)
 		valOffs = writeItemOffsets(b, values, valSz)
+		if len(s.ValDesc.Addrs) > 0 {
+			serial.ProllyTreeNodeStartValueOffsetsVector(b, len(values)*len(s.ValDesc.Addrs))
```
andrew-wm-arthur, line 69: we're using `TupleDesc.Addrs` to measure the size of `value_address_offsets`, but it looks like we never populate `TupleDesc.Addrs`.

Also, you want to use `ProllyTreeNodeStartValueAddressOffsetsVector` instead.

go/store/val/tuple_descriptor.go
```diff
@@ -32,6 +32,7 @@ type TupleDesc struct {
 	Types []Type
 	cmp   TupleComparator
 	fast  fixedAccess
+	Addrs []int
```
andrew-wm-arthur, line 35: We could probably go without this field and calculate the number of ref'd fields from the `Types` field

go/store/prolly/tree/merge.go
```diff
@@ -44,6 +45,7 @@ func ThreeWayMerge[S message.Serializer](
 	compare CompareFn,
 	collide CollisionFn,
 	serializer S,
+	valDesc val.TupleDesc,
```
andrew-wm-arthur, line 48: this is unused

go/store/prolly/message/serialize.go
```diff
@@ -58,6 +59,25 @@ func writeItemOffsets(b *fb.Builder, items [][]byte, sumSz int) fb.UOffsetT {
...
 
+// writeAddrOffests returns offsets into the values array that correspond to addr root
+// hashes that themselves have subtrees
+func writeValAddrOffsets(b *fb.Builder, items [][]byte, sumSz int, valDesc val.TupleDesc) fb.UOffsetT {
+	var cnt int
+	var off = sumSz
+	for i := len(items) - 1; i >= 0; i-- {
+		tup := val.Tuple(items[i])
+		off -= len(tup) // start of tuple
+		for j := range valDesc.Addrs {
```
andrew-wm-arthur, line 70: `valDesc.Addrs` is always nil

go/store/prolly/message/serialize.go
```diff
@@ -58,6 +59,25 @@ func writeItemOffsets(b *fb.Builder, items [][]byte, sumSz int) fb.UOffsetT {
...
+			// get index into value tuple pointing at address
+			o, _ := tup.GetOffset(j)
```
andrew-wm-arthur, line 72: you need to check if field `j` is NULL, otherwise we'll be reading a nonsense address from the next field

go/store/prolly/message/prolly_map.go
```diff
@@ -198,13 +206,12 @@ func estimateProllyMapSize(keys, values [][]byte, subtrees []uint64) (keySz, val
 	}
 
 	// todo(andy): better estimates
-	bufSz += keySz + valSz               // tuples
-	bufSz += refCntSz                    // subtree counts
-	bufSz += len(keys)*2 + len(values)*2 // offsets
```
andrew-wm-arthur, line 203: need to put this back to account for `[uint16]` offsets. Underestimating the buffer size slows us down a lot

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -47,6 +48,7 @@ func getPrimaryProllyWriter(ctx context.Context, t *doltdb.Table, sqlSch sql.Sch
 		keyMap:  keyMap,
 		valBld:  val.NewTupleBuilder(valDesc),
 		valMap:  valMap,
+		ns:      t.NodeStore(),
```
andrew-wm-arthur, line 51: you can access the `NodeStore` through the `prolly.Map` rather than added a utility method to `doltdb.Table`

andrew-wm-arthur (commented):
Needs some changes to handle blob addresses correctly

go/store/prolly/message/prolly_map.go
```diff
@@ -63,6 +65,10 @@ func (s ProllyMapSerializer) Serialize(keys, values [][]byte, subtrees []uint64,
 		valTups = writeItemBytes(b, values, valSz)
 		serial.ProllyTreeNodeStartValueOffsetsVector(b, len(values)-1)
 		valOffs = writeItemOffsets(b, values, valSz)
+		if len(s.ValDesc.Addrs) > 0 {
+			serial.ProllyTreeNodeStartValueAddressOffsetsVector(b, len(values)*len(s.ValDesc.Addrs))
```
andrew-wm-arthur, line 69: `len(values)*len(s.ValDesc.Addrs)` is only accurate if every blob field is non-null

go/store/prolly/message/serialize.go
andrew-wm-arthur: you need to account for null blob fields here

go/libraries/doltcore/schema/serial_encoding.go
```diff
@@ -86,6 +79,10 @@ func EncodingFromSqlType(typ query.Type) serial.Encoding {
 		return serial.EncodingJSON
 	case query.Type_GEOMETRY:
 		return serial.EncodingGeometry
+	case query.Type_BLOB:
+		return serial.EncodingBytesAddr
+	case query.Type_TEXT:
+		return serial.EncodingString
```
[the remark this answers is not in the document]
max-hoffman: text/json is my next follow-up

go/store/prolly/message/prolly_map.go
max-hoffman: isn't it the same issue with other null fields? or do you do null counting somewhere

go/store/prolly/message/serialize.go
max-hoffman: in a different way than i already did?

go/libraries/doltcore/schema/typeinfo/blobstring.go
```diff
@@ -134,7 +143,8 @@ func (ti *blobStringType) FormatValue(v types.Value) (*string, error) {
 		if err != nil {
 			return nil, err
 		}
-		return &resStr, nil
+		return (*string)(unsafe.Pointer(&resStr)), nil
```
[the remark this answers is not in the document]
max-hoffman: deleting, good catch

go/store/prolly/message/serialize.go
andrew-wm-arthur: sorry, I was confused by the `_`, LGTM

PR #3591, opened 2022-06-10
Add support for bulk_ea and reconfigure constructor to use config pat…

VinaiRachakonda:
…tern instead of 8 params


JCOR11599 (approved):
LGTM

PR #3605, opened 2022-06-13
[no-release-notes] foreign key constraint violations

druvv:
Implements foreign key constraint violations for the new format. Unskips related engine tests and bats tests. 

In general, we should move all of the bats tests testing merge functionality into the engine tests and we should write more engine tests to cover test coverage gaps.

TODO:
- Unique key constraint violations (currently not tested by engine tests), but will be a fast follow to this PR.


druvv:
TODO:
- Store their hash instead of our hash in the artifact table
- anywhere we reference an address should use hash.Hash instead of a byte slice

andrew-wm-arthur (approved):
Generally looks good, there's some room to cleanup up the Artifact API. 

Ideally we'd like to define `Artifact` such that we can construct them in a zero-allocation way:
```
type Artifact struct {
    k, v val.Tuple
}
```
and then use accessor methods to replicate the fields that are currently defined

PR #3608, opened 2022-06-13
Add a warning for imports where the file's schema is smaller than the table's schema

VinaiRachakonda:
We should warn the user when executing a partial import as it may have antagonistic behaviors when import files do not properly match a table schema. For example consider a table with the schema (pk, val). If I import a file with schema (pk, va) the second column will get dropped by the importer.

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/import.go
+++ b/go/cmd/dolt/commands/tblcmds/import.go
@@ -500,6 +500,12 @@ func newImportSqlEngineMover(ctx context.Context, dEnv *env.DoltEnv, rdSchema sc
 		return nil, &mvdata.DataMoverCreationError{ErrType: mvdata.SchemaErr, Cause: err}
 	}
 
+	// Leave a warning if the import operation is operating on fewer columns than the relevant table's schema.
+	// This can certainly be intentional, but it is often due to typos in the header of a csv file.
+	if rowOperationSchema.GetAllCols().Size() < tableSchema.GetAllCols().Size() {
+		cli.PrintErrln(color.YellowString("Warning: There are fewer columns in the import file's schema than the table's schema.\nIf unintentional, check for any typos in the import file's header."))
+	}
+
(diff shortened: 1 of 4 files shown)


go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -500,6 +500,12 @@ func newImportSqlEngineMover(ctx context.Context, dEnv *env.DoltEnv, rdSchema sc
 		return nil, &mvdata.DataMoverCreationError{ErrType: mvdata.SchemaErr, Cause: err}
 	}
 
+	// Leave a warning if the import operation is operating on fewer columns than the relevant table's schema.
+	// This can certainly be intentional, but it is often due to typos in the header of a csv file.
+	if rowOperationSchema.GetAllCols().Size() < tableSchema.GetAllCols().Size() {
+		cli.PrintErrln(color.YellowString("Warning: There are fewer columns in the import file's schema than the table's schema.\nIf unintentional, check for any typos in the import file's header."))
```
fulghum, line 506: Great check. I'm just wondering if we should be stricter and make people pass an extra flag to allow partial column coverage, if the columns don't match exactly, but this is a step in a good direction for sure. 

Might be interesting to think about the other side, too... do we already have a guard against the csv containing columns the table doesn't have? (I was going to dig through the code myself, but have a call in a couple minutes) Worth double-checking that we don't drop data in that case.

VinaiRachakonda: 1. My sense is no because we do allow users to pass in a csv file that acts as an update operation. Since update usually take the form of a partial I think we should make it as accessible as possible. 
2. We ignore any column that the csv has that isn't in the ultimate table schema. The only edge case is if a mapping file is passed in

VinaiRachakonda: 1. My sense is no because we do allow users to pass in a csv file that acts as an update operation. Since update usually take the form of a partial I think we should make it as accessible as possible. 
2. We ignore any column that the csv has that isn't in the ultimate table schema. The only edge case is if a mapping file is passed in

PR #3619, opened 2022-06-14
Fix setting `max_connections` in server yaml

jennifersp:
We use system variables depending on whether a variable is user-defined variable or not. The issue was variables defined in config variable was not used in starting new sql engine. When `max_connections` variable was defined in config file either through yaml file or on command line, it should use that value, if not it uses default value. We check for persistence behavior. If it's set to `load`, then use the value of global `max_connections` variable, otherwise use value set in server config, which can either be default or user defined value.


timsehn:
Beautiful :-)

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -171,11 +171,23 @@ func (cfg *commandLineServerConfig) Password() string {
 
 // ReadTimeout returns the read and write timeouts.
 func (cfg *commandLineServerConfig) ReadTimeout() uint64 {
+	if cfg.timeout == 0 && cfg.PersistenceBehavior() == loadPerisistentGlobals {
```
max-hoffman, line 174: this does two things -- 1) applies a partial default hierarchy for config values, 2) makes the config dynamic to session level variable changes. I think 1) is somewhat confusing, because it fails to take into account command line arguments in the hierarchy. 2) is an enhancement, but these configs are only ever used once immediately at Serve time, so it is not used. Perhaps it is better to apply the full hierarchy elsewhere?

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -289,5 +297,50 @@ func getYAMLServerConfig(fs filesys.Filesys, path string) (ServerConfig, error)
 		return nil, fmt.Errorf("Failed to parse yaml file '%s'. Error: %s", path, err.Error())
 	}
 
+	if cfg.PersistenceBehavior() == loadPerisistentGlobals {
```
max-hoffman, line 300: these changes look correct, but it might be simpler to structure the code in a way that naturally describes what's happening here. For every variable, 1) we try to set it from command line arg, 2) we try to set it from persisted globals, 3) we resort to a default value. The output config is considered final, because this function has taken care of all possible inputs and applied the correct ordering (? I don't actually know if it's true, but if not it would make sense to add those here, right?)

max-hoffman: if it is possible to structure the function like above I think it makes sense to add a table-test that runs `getCommandLineServerConfig` with different sets of arg and config inputs. You can test every variable rather than getting one case in a bats test.

max-hoffman (commented):
mostly LGMT, but I think if you write more tests you might catch some edge cases, which restructuring might clarify

go/cmd/dolt/commands/sqlserver/sqlserver.go
jennifersp: I wasn't able to make tests explicitly for `getCommandLineServerConfig`, but I added more bats test.

go/cmd/dolt/commands/sqlserver/serverconfig.go
jennifersp: I included getting persisted global values in `getCommandLineServerConfig` but it's kind of weird. I was not sure how to set timeout when there are two types of timeouts. At the same time, the timeout are set through `ReadTimeout()` and `WriteTimeout()` which both return the same value, `timeout`.

max-hoffman (approved):
LGTM, we still need to refactor config merging like i mentioned but we should ship something

Issue #3015, opened 2022-03-16
Dolt dump should support common bulk loading paradigms


VinaiRachakonda:
When [bulk loading](https://dev.mysql.com/doc/refman/8.0/en/optimizing-innodb-bulk-data-loading.html) data or dump data certain variables are set to ensure that data can loaded in as easily as possible.

Typically this leverages the following variables.

1) `foreign_key_checks` (Essential for loading in tables with foreign key references)
2) `autocommit`

We should toggle these variables accordingly in `dolt dump` as this ensures that a dumped database can correctly be loaded in other dolt databases

PR #3620, opened 2022-06-14
Add bulk export parameters to dump

VinaiRachakonda:
Addressed issue #3015 

Note that `mysqldump` uses the `--no-autocommit` flag which wraps each table with a `SET AUTOCOMMIT=0` and `COMMIT=1`

cc: https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html#option_mysqldump_no-autocommit


VinaiRachakonda:
Just realized that this should be added to dolt table export as well. Will followup and edit accordingly

VinaiRachakonda:
TODO: Update docummentation

jennifersp (approved):
LGMT!

PR #3625, opened 2022-06-15
[no-release-notes] Flatbuffer serialization for `schema.Schema`

andrew-wm-arthur:
This PR modifies `__DOLT_1__` to use Flatbuffers serialization for `schema.Schema`. The new encoding makes a number of changes with intention of providing more flexibility going forward:
- SQL type information is serialized as strings rather than key-value type parameters.
- The "primary" clustered row index is explicitly encoded.
- Hidden columns (eg for keyless tables) are explicitly encoded.
- Indexes encode the physical layout of their key and value tuples, in addition to the columns their defined on. This allows more flexibility when building new index storage layouts (eg spatial indexes, virtual columns).

This PR does include tags in column definitions, it is undecided whether they will remain in the final version.

`__DOLT_DEV__` is not modified in this PR. Two Noms types (`UUID` and `Bool`) have no corresponding SQL type and therefore can't be encoded as a SQL type string. This will be addressed in future changes.


go/libraries/doltcore/schema/encoding/serialization.go
```diff
@@ -0,0 +1,472 @@
...
+	n := s.ColumnsLength()
+	if n < 2 {
+		return false
+	}
+	// keyless id is the 2nd to last column
+	// in the columns vector (by convention)
+	// and the only field in key tuples of
+	// the clustered index.
+	id := serial.Column{}
+	s.Columns(&id, n-2)
```
reltuk, line 421: Nit: Is this access pattern better than `id := new(serial.Column)` for heap allocation / elision?

If so, should we be using it above in the functions like `deserializeSecondaryIndexes`?

reltuk (approved):
LGTM. A couple nit-y questions.

andrew-wm-arthur:
Addressing PR feedback in a subsequent PR

PR #3639, opened 2022-06-17
Improved error message for multistatement query failing during exec

fulghum:
While debugging https://github.com/dolthub/dolt/issues/3512 , realized that if a multistatement exec fails during query execution, Dolt doesn't tell you which line failed. Updated the error message to include the failing statement's line and added a bats test for the original customer issue. 

Depends on: https://github.com/dolthub/go-mysql-server/pull/1066


zachmu (approved):
LGTM

PR #3640, opened 2022-06-17
[no-release-notes] Flatbuffer encoding for Foreign Keys

andrew-wm-arthur:


go/serial/foreign_key.fbs
```diff
@@ -0,0 +1,56 @@
...
+  parent_table_index:string;
+  parent_table_columns:[uint64];
+
+  // reference options
+  on_update:ForeignKeyReferentialAction;
+  on_delete:ForeignKeyReferentialAction;
+
+  // unresolved details
+  unresolved_child_columns:[string];
+  unresolved_parent_columns:[string];
```
Hydrocharged, line 50: These used to represent unresolved columns, but now they're always set. Before the migration I am planning to fully remove the `child_table_columns:[uint64];` and `parent_table_columns:[uint64];` fields (in addition to the index fields too) and strictly rely on these fields instead. I'd just call these `string_` for now.

Hydrocharged (approved):
LGTM!

PR #3642, opened 2022-06-17
Update Describe tests in Dolt due to bug in GMS

VinaiRachakonda:
This pr modifies bats and enginetests due to some missing functionality in GMS.


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -103,23 +103,23 @@ var DescribeTableAsOfScriptTest = queries.ScriptTest{
 		{
 			Query: "describe a as of @Commit1;",
 			Expected: []sql.Row{
-				{"pk", "int", "NO", "PRI", "", ""},
-				{"c1", "int", "YES", "", "", ""},
+				{"pk", "int", "NO", "PRI", "NULL", ""},
```
fulghum, line 106: Should these be `nil` instead of `"NULL"`? 

I recently had a bug where I was returning `"NULL"` from Dolt code for a row instead of `nil` and my tests were explicitly checking for `"NULL"`, so they passed, but when the code was used over the wire, that `"NULL"` string couldn't be converted to a timestamp (or whatever datatype it was) and it was a runtime error. Probably good to standardize on `nil` in tests to avoid problems that like being able to hide.

fulghum: actually... nevermind... on a closer look, I see that this is the default value column and we do need to indicate a default value (not the lack of any value), so `nil` doesn't make sense. I just hit this test failure as well, trying to get a GMS change into Dolt.

integration-tests/bats/sql-merge.bats
```diff
@@ -621,7 +621,7 @@ SQL
 }
 
 @test "sql-merge: DOLT_MERGE detects merge conflicts, fails to commit and leaves working set clean when dolt_allow_commit_conflicts = 0" {
-    skip "This test is flaking and needs to be fixed"
+    skip "This test is flaky in ci/cd. Skipping for now"
```
VinaiRachakonda, line 624: Mistake from merge but don't want to run ci again :/

fulghum (approved):
Looks good!

PR #3644, opened 2022-06-17
Reimplemented dolt diff command to use SQL queries

zachmu:
Also:
* Re-enable push / pull, unskip related tests for new format
* Get rid of use of row printing pipeline in a few places, reimplemented output printing with SQL rows


go/cmd/dolt/commands/diff.go
```diff
@@ -438,6 +431,43 @@ func diffUserTables(ctx context.Context, fromRoot, toRoot *doltdb.RootValue, dAr
 	return nil
 }
 
+func newSqlEngine(ctx context.Context, dEnv *env.DoltEnv) (*engine.SqlEngine, error) {
```
druvv, line 434: lots of other cli commands will probably want to use the engine. Maybe this func should live in the commands/engine package?

andrew-wm-arthur (approved):
LGTM!

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -38,6 +38,13 @@ func init() {
 	types.CreateEditAccForMapEdits = edits.NewAsyncSortedEditsWithDefaults
 }
 
+// WORKING and STAGED identifiers refer to the working and staged roots in special circumstances where
+// we expect to resolve a commit spec, but need working or staged
+const (
+	Working = "WORKING"
+	Staged  = "STAGED"
+)
+
```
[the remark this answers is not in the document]
zachmu: No, because of cycles. They're used in sqle and dsession

go/cmd/dolt/commands/diff.go
zachmu: Good call, moved

PR #3653, opened 2022-06-21
selecting geometry types now show up in hex form, rather than raw ASCII

JCOR11599:


fulghum (approved):
Nice! Awesome to be able to move this non-dolt-specific code out of Dolt and down into GMS! 😄

PR #3654, opened 2022-06-21
Update `dolt merge` unique key violation behavior. Implement unique key violations for new format.

druvv:
## Both format changes 
**`CALL DOLT_MERGE` `conflicts` column returns 1 if the post-merge root contains conflicts or violations**:
Previously, the `conflicts` column would only return 1 if the merge produced new conflicts. Now it returns 1 regardless if whether the conflicts in the root are new or from a previous merge.

It also returns 1 if there are any constraint violations.

## Current format changes

**1. Unique key constraints during merge are documented for both the existing row and violating row**:

Sometimes a `dolt merge` or the stored-procedure equivalent `CALL DOLT_MERGE` may produce unique key violations even when those violations did not exist in either the target or source branches. 

Consider the following:
```sql
-- in the ancestor
CREATE TABLE t (pk int PRIMARY KEY, col1 int UNIQUE);
INSERT INTO t VALUES (1, 1);

-- on the right, this is valid
INSERT INTO t VALUES (2, 2);
/*
Query OK, 1 row affected
*/

-- on the left, this is valid
UPDATE t set col1 = 2 where pk = 1;
/*
Query OK, 1 row affected
Rows matched: 1  Changed: 1  Warnings: 0
*/

-- in the post merge state col1 has the duplicate value 2
SELECT * from t;
/*
+----+------+
| pk | col1 |
+----+------+
| 1  | 2    |
| 2  | 2    |
+----+------+
*/
```

Prior to this PR only PK 2 would be returned as a constraint violations:
```sql
SELECT * from dolt_constraint_violations_t;
/*
+----------------+----+------+---------------------------------------+
| violation_type | pk | col1 | violation_info                        |
+----------------+----+------+---------------------------------------+
| unique index   | 2  | 2    | {"Columns": ["col1"], "Name": "col1"} |
+----------------+----+------+---------------------------------------+
*/
```

After this PR both PK 1 and 2 are returned as constraint violations:
```sql
SELECT * from dolt_constriant_violations_t;
/*
+----------------+----+------+---------------------------------------+
| violation_type | pk | col1 | violation_info                        |
+----------------+----+------+---------------------------------------+
| unique index   | 1  | 2    | {"Columns": ["col1"], "Name": "col1"} |
| unique index   | 2  | 2    | {"Columns": ["col1"], "Name": "col1"} |
+----------------+----+------+---------------------------------------+
*/
```

**2. Rows that violate a unique key constraint during merge are no longer filtered. They are allowed to persist into the left**:

In the above example, prior to this PR we would not see PK 2 in our post-merge state:
```sql
-- post-merge state
SELECT * from t;
/*
+----+------+
| pk | col1 |
+----+------+
| 1  | 2    |
+----+------+
*/
```

After this PR, we do see PK 2:
```sql
-- post-merge state
SELECT * from t;
/*
+----+------+
| pk | col1 |
+----+------+
| 1  | 2    |
| 2  | 2    |
+----+------+
*/
```

## New format changes
The new format matches the current format behavior except for the following.

**1. Merge errors if a single row violates multiple UK or FK violations**:

The PKs of the artifact table are: PKs of the source table, a root-ish hash, and then the artifact type. If a row violates multiple UK or FK constraints in the same merge, there is a PK key collision and the previous CV entry is overwritten with the new one. In the future, we can store both constraint violations in the json metadata.

This is a problem in the current format as well but we don't produce errors.

**2. If the ancestor has duplicate values for a column and the right de-dupes them and adds a UK constraint, the merge will not abort**:

In the current format this case aborts the merge because on the left we have constraint violations. If the left is missing a unique index, it is built prior to the merge. During this process if a duplicate value is found the entire merge is aborted.

In the new format, if a unique index is missing in either the left, right, or ancestor we build the merged index with the post-merge primary row data. If we encounter duplicate values here we document them instead of aborting. 

Due to the way the current format updates indexes, even if we documented the constraint violations in the left we would not achieve consistent behavior. For example, we would document the violations that the right might have already resolved.


go/libraries/doltcore/table/editor/pk_table_editor.go
```diff
@@ -125,16 +122,19 @@ type pkTableEditor struct {
...
 
 	// Whenever any write operation occurs on the table editor, this is set to true for the lifetime of the editor.
 	dirty uint32
 
-	// This mutex blocks on each operation, so that map reads and updates are serialized
+	// This mutex blocks on each operation, so that map reads and updates are serialized. Protects the above data.
 	writeMutex *sync.Mutex
+
+	// protects cvEditor
+	cvMutex  *sync.Mutex
```
druvv, line 136: I created a new mutex to avoid a deadlock:
1. write lock is taken on `writeMutex` when InsertRow is called
2. Any duplicate rows are passed to `PkDuplicateCb` which calls `SetConstraintViolation` which tries to take a lock.
3. Deadlock.

andrew-wm-arthur (approved):
LGTM

PR #3662, opened 2022-06-22
go/libraries/doltcore/sqle/index: Invalidate schema-dependent cached durable state when root value changes.

reltuk:


go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -44,10 +44,10 @@ type DoltIndex interface {
 	Format() *types.NomsBinFormat
 	IsPrimaryKey() bool
 
-	GetDurableIndexes(*sql.Context, DoltTableable) (durable.Index, durable.Index, error)
-	coversColumns(cols []string) bool
-	lookupTags() map[uint64]int
-	sqlRowConverter() *KVToSqlRowConverter
+	getDurableState(*sql.Context, DoltTableable) (*durableIndexState, error)
+	coversColumns(s *durableIndexState, columns []string) bool
+	sqlRowConverter(s *durableIndexState) *KVToSqlRowConverter
```
max-hoffman, line 49: i can't think of why we'd want a monomorphic visitor interface method, rather than just attaching state to the index. do we use the state elsewhere? would other implementations do something other than call the visit method?

max-hoffman (approved):
LGTM, just confused about expectations for how you plan to expand/change durable state use

PR #3667, opened 2022-06-22
Clean up revision database metadata when renaming/deleting branches

fulghum:
When a Dolt sql-server connects to a branch-qualified database (e.g. `<database>/<branch>`) either through a connection or `use` statement, it tracks metadata for that database+branch and shows the entry as a database in `show databases`. If that branch is deleted or renamed, the metadata isn't cleaned up and can cause the server to stop working correctly in any query that causes all databases to be searched (e.g. using `information_schema`, calling a stored procedure, calling `show procedure status). 

This PR updates the branch delete/rename logic to clean up this database metadata and also fixes the safegaurd for deleting/renaming a branch in use by another session, since it didn't take revision databases into account previously. 

Fixes: https://github.com/dolthub/dolt/issues/3636

Depends on: https://github.com/dolthub/vitess/pull/170


go/libraries/doltcore/sqle/dfunctions/dolt_branch.go
```diff
@@ -196,6 +214,42 @@ func validateBranchNotActiveInAnySession(ctx *sql.Context, branchName string) er
...
+		return fmt.Errorf("unexpected session type: %T", ctx.Session)
+	}
+
+	provider := dsess.Session.Provider()
+	if provider, ok := provider.(sql.MutableDatabaseProvider); ok {
+		if provider.HasDatabase(ctx, revisionDbName) == false {
+			return nil
+		}
+
+		err := provider.DropDatabase(ctx, revisionDbName)
```
zachmu, line 244: Maybe should be different from the SQL interface method for the above reason

zachmu (approved):
LGTM with some caveats.

If you wanted to do a little more in this space as a followup, it seems like we should really go all-in on the database metaphor: make `drop` delete the branch, show all branches in `show databases` instead of just the ones they have connected to so far.

go/libraries/doltcore/sqle/dfunctions/dolt_branch.go
fulghum: I like that idea. I moved this code to a new method: `RevisionDbProvider.DropRevisionDb`

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -331,6 +337,12 @@ func (p DoltDatabaseProvider) TableFunction(ctx *sql.Context, name string) (sql.
 	return nil, sql.ErrTableFunctionNotFound.New(name)
 }
 
+// IsRevisionDatabase returns true if the specified dbName represents a database that is tied to a specific
+// branch or commit from a database (e.g. "dolt/branch1").
+func IsRevisionDatabase(dbName string) bool {
```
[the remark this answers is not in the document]
fulghum: Ah, thanks for pointing out that slashes are valid in db names. It would be a bummer to be more restrictive than MySQL here. That'll require a little more thinking about to handle safely, like you're suggesting. 

I've gone ahead and added a note in the code for now and happy to include this in the database/branch cleanup work this week, too.

PR #3670, opened 2022-06-23
Remove `SELECT CONSTRAINTS_VERIFY_ALL` and `DOLT_VERIFY_CONSTRAINTS_ALL`

druvv:
Removes the ALL versions of CONSTRAINTS_VERIFY() and DOLT_VERIFY_CONSTRAINTS() in favor of the --all flag. 

Also adds `--output-only` support for SQL


integration-tests/bats/verify-constraints.bats
```diff
@@ -189,169 +189,3 @@ SQL
     [[ ! "$output" =~ "child4,2" ]] || false
     [[ "${#lines[@]}" = "1" ]] || false
 }
-
-@test "verify-constraints: SQL no violations" {
```
max-hoffman, line 193: :)

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -218,6 +219,14 @@ func CreateBackupArgParser() *argparser.ArgParser {
 	return ap
 }
 
+func CreateVerifyConstraintsArgParser() *argparser.ArgParser {
+	ap := argparser.NewArgParser()
+	ap.SupportsFlag(AllFlag, "a", "Verifies constraints against every row.")
```
max-hoffman, line 224: All rows? Or all tables?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3568,3 +3568,370 @@ var CommitDiffSystemTableScriptTests = []queries.ScriptTest{
 		},
 	},
 }
+
+var verifyConstraintsSetupScript = []string{
```
max-hoffman, line 3572: i'm not sure this works for all of your tests, but using this as the harness setup -> `harness.Setup(verifyConstraintsSetu)`, and having the rest as regular `queries.QueryTest` or `queries.WriteQueryTest` is generally faster

max-hoffman (approved):
LGTM, nice improvements

go/cmd/dolt/cli/arg_parser_helpers.go
druvv: I'll improve the description, it is all rows. If you specify no tables then all tables are considered...

PR #3676, opened 2022-06-23
update and fix cherry-pick docs

jennifersp:


timsehn:
Looks like this won't fail but could use an update:

https://github.com/dolthub/dolt/blob/8a67c5326bff435f13b38809776404c274d0abe8/integration-tests/bats/no-repo.bats#L20

It makes sure the `dolt --help` is not busted because it's the front door.

timsehn (approved):
Ship it :-)

PR #3678, opened 2022-06-23
Fix broken link for hosted dolt in readme.md

illgitthat:
Without the HTTPS prefix, the link went to https://github.com/dolthub/dolt/blob/main/hosted.doltdb.com


timsehn:
Thanks! I added this on mobile this morning.

PR #3685, opened 2022-06-23
Speed up client tests by installing pre-compiled R packages

noamross:
https://packagemanager.rstudio.com maintains binary versions of R packages so one doesn't have to compile them all to install on Linux


fulghum:
Nice! Thanks Noam! 

I just kicked off the CI tests. After they pass we'll get this merged in. Thanks for the contribution! 🙏

PR #3687, opened 2022-06-23
Fix internal merges not rolling back on a constraint violation

druvv:
Bug fix for transactions that create constraint violations.

Prior to this PR, an internal merge (a merge called inside a transaction) or a transaction merge that produced constraint violations would not rollback the transaction. 

The desired behavior is that if `dolt_force_transaction_commit` is false, then rollback the transaction if any constraint violations are encountered.


go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -456,3 +456,244 @@ func TestDoltTransactionCommitLateFkResolution(t *testing.T) {
...
+			},
+			{
+				Query:    "/* client a */ COMMIT;",
+				Expected: []sql.Row{},
+			},
+			{
+				Query:          "/* client b */ COMMIT;",
+				ExpectedErrStr: "this transaction conflicts with a committed transaction from another client, please retry",
+			},
+			// TODO: uncommenting this query causes a weird reset script error @max??
```
zachmu, line 563: You should figure out what's going on here

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -456,3 +456,244 @@ func TestDoltTransactionCommitLateFkResolution(t *testing.T) {
...
+
+// Tests behavior of internal merges ( merges that occur inside a transaction ).
+func TestInternalMerges(t *testing.T) {
+	harness := newDoltHarness(t)
+	for _, script := range internalMergeScripts {
+		enginetest.TestTransactionScript(t, harness, script)
+	}
+}
+
+var transactionMergeScripts = []queries.TransactionTest{
```
zachmu, line 532: Same comment about test organization

I'm fine with having multiple blocks of tests that test different themes, but I don't know that this organizational scheme makes sense to me

zachmu (approved):
Mostly looks good, a couple things to clean up before you merge

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -456,3 +456,244 @@ func TestDoltTransactionCommitLateFkResolution(t *testing.T) {
 		},
 	})
 }
+
+var internalMergeScripts = []queries.TransactionTest{
```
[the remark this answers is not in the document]
druvv: How about explicit merge? It would be explicit because the user called the `DOLT_MERGE` procedure.

PR #3690, opened 2022-06-24
[no-release-notes] text and json written as byte trees

max-hoffman:


go/libraries/doltcore/doltdb/table.go
```diff
@@ -110,6 +110,9 @@ func (t *Table) ValueReadWriter() types.ValueReadWriter {
 
 // NodeStore returns the NodeStore for this table.
 func (t *Table) NodeStore() tree.NodeStore {
+	if t == nil {
```
zachmu, line 113: Is this a principled use of a nil receiver? Does the nil `*Table` have a semantic meaning where we need to support this (as opposed to it just kicking the error along silently)?

go/store/prolly/tree/immutable_tree.go
```diff
@@ -161,27 +166,72 @@ func NewByteArray(addr hash.Hash, ns NodeStore) *ByteArray {
...
 	}
-	return string(b.buf[:toShow]), nil
+	toShow := bytePeekLength
+	if len(buf) < toShow {
+		toShow = len(buf)
+	}
+	return string(buf[:toShow]), nil
+}
+
+type StringArray struct {
```
zachmu, line 217: Can we call this TextStorage or something? The type isn't an array, it's a string type that stores its data as an array

go/store/prolly/tree/immutable_tree.go
```diff
@@ -161,27 +166,72 @@ func NewByteArray(addr hash.Hash, ns NodeStore) *ByteArray {
...
+		return "", err
+	}
+	toShow := bytePeekLength
+	if len(buf) < toShow {
+		toShow = len(buf)
+	}
+	return string(buf[:toShow]), nil
+}
+
+type JSONArray struct {
```
zachmu, line 184: JsonDoc?

zachmu (approved):
LGTM, just some comments about names

PR #3696, opened 2022-06-27
[no-release-notes] log $output when $status is a mismatch in merge tests

druvv:
Will revert this commit when it's clear what is causing the tests to flake.

### DIFF
--- a/integration-tests/bats/helper/common.bash
+++ b/integration-tests/bats/helper/common.bash
@@ -78,6 +78,22 @@ teardown_common() {
     rm -rf "$BATS_TMPDIR/dolt-repo-$$"
 }
 
+status_eq() {
+    if ! [ "$status" -eq $1 ]; then
+        echo "status: expected $1, received $status"
+        printf "output:\n$output"
+        exit 1
+    fi
+}
+
+output_has() {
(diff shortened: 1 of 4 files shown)


integration-tests/bats/helper/common.bash
```diff
@@ -78,6 +78,22 @@ teardown_common() {
     rm -rf "$BATS_TMPDIR/dolt-repo-$$"
 }
 
+status_eq() {
```
zachmu, line 81: Call this log_status_eq

integration-tests/bats/helper/common.bash
```diff
@@ -78,6 +78,22 @@ teardown_common() {
...
+    if ! [ "$status" -eq $1 ]; then
+        echo "status: expected $1, received $status"
+        printf "output:\n$output"
+        exit 1
+    fi
+}
+
+output_has() {
```
zachmu, line 89: log_output_has

zachmu (commented):
LGTM

Are the tests in constraint-violations.bats also flakey?

druvv:
Yes I saw a flaky test in constraint-violation.bats since it uses merge.

PR #3714, opened 2022-06-29
Add missing semicolon

yubischiess:
Missing semicolon in docs


fulghum:
Good eye @yubischiess. Thanks for sending this fix over!

PR #3719, opened 2022-06-29
Update Dolt to handle the information_schema having capital column names

VinaiRachakonda:


fulghum (approved):
Nice job having the Dolt fixes ready to go.

PR #3722, opened 2022-06-29
Only rebase transaction start for read replica

max-hoffman:
Changes:
- Remove rebase on TransactionStart, unless the engine is a read replica, in which case we continue to poll for updates.
- Starting a server writes a lock file `.dolt/sql-server.lock` that is deleted when the server is terminated.
- A SqlEngine created after a server lock is written is limited to `ReadOnly` mode. Bulk table import does not appear to respect read only mode, so it has an additional check.


max-hoffman:
#newbenchmark

timsehn:
These still aren't bad perf wine ~5-10% on most metrics.

go/libraries/doltcore/env/environment.go
```diff
@@ -1278,3 +1280,26 @@ func (dEnv *DoltEnv) DbEaFactory() editor.DbEaFactory {
 func (dEnv *DoltEnv) BulkDbEaFactory() editor.DbEaFactory {
 	return editor.NewBulkImportTEAFactory(dEnv.DoltDB.Format(), dEnv.DoltDB.ValueReadWriter(), dEnv.TempTableFilesDir())
 }
+
+func (dEnv *DoltEnv) lockFile() string {
+	return filepath.Join(dbfactory.DoltDir, ServerLockFile)
+}
+
+// IsLocked returns true if this database's lockfile exists
+func (dEnv *DoltEnv) IsLocked() bool {
```
max-hoffman, line 1289: should these be mutex protected?

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -35,6 +36,8 @@ import (
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqlserver"
 )
 
+var ErrActiveServerLock = errors.New("database locked by another sql-server; either clone the database to run a second server, or delete the '.dolt/sql-server.lock' if no other sql-servers are active")
```
zachmu, line 39: Error message should have the absolute path to the lock file

go/libraries/doltcore/env/multi_repo_env.go
```diff
@@ -32,6 +33,8 @@ import (
 	"github.com/dolthub/dolt/go/libraries/utils/filesys"
 )
 
+var ErrActiveServerLock = errors.New("database locked by another sql-server; either clone the database to run a second server, or delete the '.dolt/sql-server.lock' if no other sql-servers are active")
```
zachmu, line 36: What's with the duplicate error def?

go/performance/scripts/local_benchmark.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 set -o pipefail
 
-SYSBENCH_TEST="oltp_point_select"
```
zachmu, line 5: Revert?

zachmu (approved):
LGTM

As a followup, you need to extend this to all `dolt` commands which can update db state.

PR #3725, opened 2022-06-30
fix inconsistencies of dolt checkout and pull commands

jennifersp:
This PR fixes inconsistent behaviors of Dolt CLI against Git commands AND of Dolt SQL against Dolt CLI commands.

Inconsistencies between Dolt CLI and Git:
- `dolt checkout new_branch`: checking out a new branch without `-b` flag. If there is a remote branch with matching name, sets upstream.
- `dolt checkout -b new_branch`: if there is a remote branch with matching name, it does not set upstream.
- `dolt pull` and `dolt pull remote_name`: each gives separate errors when run on a branch that does not have upstream

Inconsistencies between Dolt CLI and Dolt SQL:
- `dolt pull` should match `dolt_pull()`
- `dolt_checkout()` and `dolt_push()` should set upstream that persists outside of sql session (added skipped tests)

### DIFF
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -132,7 +132,6 @@ func DoPush(ctx context.Context, rsr env.RepoStateReader, rsw env.RepoStateWrite
 			if err != nil {
 				return err
 			}
-			// TODO : set upstream should be persisted outside of session
 		}
 	}
 
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -132,6 +132,7 @@ func DoPush(ctx context.Context, rsr env.RepoStateReader, rsw env.RepoStateWrite
 			if err != nil {
 				return err
 			}
+			// TODO : set upstream should be persisted outside of session
 		}
 	}
 
--- a/go/libraries/doltcore/env/actions/table.go
+++ b/go/libraries/doltcore/env/actions/table.go
@@ -131,28 +131,20 @@ func RemoveDocsTable(tbls []string) []string {
 	return result
 }
 
-// GetRemoteBranchRef returns the ref of a branch and ensures it matched with name. It will also return boolean value
-// representing whether there is not match or not and an error if there is one.
-func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) (ref.RemoteRef, bool, error) {
+// GetRemoteBranchRef returns a remote ref with matching name for a branch for each remotes.
+func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) ([]ref.RemoteRef, error) {
 	remoteRefFilter := map[ref.RefType]struct{}{ref.RemoteRefType: {}}
 	refs, err := ddb.GetRefsOfType(ctx, remoteRefFilter)
-
 	if err != nil {
-		return ref.RemoteRef{}, false, err
+		return nil, err
 	}
 
-	// This will only do this behavior of guessing the remote branch if there is exactly one remote tracking branch with the matching branch name.
-	var found = false
-	var remoteRef ref.RemoteRef
+	var remoteRef []ref.RemoteRef
 	for _, rf := range refs {
 		if remRef, ok := rf.(ref.RemoteRef); ok && remRef.GetBranch() == name {
-			if found {
-				return ref.RemoteRef{}, false, nil
-			}
-			found = true
-			remoteRef = remRef
+			remoteRef = append(remoteRef, remRef)
 		}
 	}
 
-	return remoteRef, found, nil
+	return remoteRef, nil
 }
--- a/go/libraries/doltcore/env/actions/table.go
+++ b/go/libraries/doltcore/env/actions/table.go
@@ -141,11 +141,18 @@ func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) (r
 		return ref.RemoteRef{}, false, err
 	}
 
+	// This will only do this behavior of guessing the remote branch if there is exactly one remote tracking branch with the matching branch name.
+	var found = false
+	var remoteRef ref.RemoteRef
 	for _, rf := range refs {
 		if remRef, ok := rf.(ref.RemoteRef); ok && remRef.GetBranch() == name {
-			return remRef, true, nil
+			if found {
+				return ref.RemoteRef{}, false, nil
+			}
+			found = true
+			remoteRef = remRef
 		}
 	}
 
-	return ref.RemoteRef{}, false, nil
+	return remoteRef, found, nil
 }
--- a/go/libraries/doltcore/env/actions/table.go
+++ b/go/libraries/doltcore/env/actions/table.go
@@ -133,19 +133,19 @@ func RemoveDocsTable(tbls []string) []string {
 
 // GetRemoteBranchRef returns the ref of a branch and ensures it matched with name. It will also return boolean value
 // representing whether there is not match or not and an error if there is one.
-func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) (ref.DoltRef, bool, error) {
+func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) (ref.RemoteRef, bool, error) {
 	remoteRefFilter := map[ref.RefType]struct{}{ref.RemoteRefType: {}}
 	refs, err := ddb.GetRefsOfType(ctx, remoteRefFilter)
 
 	if err != nil {
-		return nil, false, err
+		return ref.RemoteRef{}, false, err
 	}
 
 	for _, rf := range refs {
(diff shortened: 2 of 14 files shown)


fulghum:
Are you sure about this one?
> `dolt push origin new_branch`: if remote name and branch name is given, it sets upstream without --set-upstream flag

I didn't see anything in the git docs about automatically setting the upstream without having to supply `--set-upstream` and a quick test with git didn't record an upstream for me. Let me know if you can repro git setting the upstream this way.

go/libraries/doltcore/env/remotes.go
```diff
@@ -45,6 +45,7 @@ var ErrCannotPushRef = errors.New("cannot push ref")
 var ErrNoRefSpecForRemote = errors.New("no refspec for remote")
 var ErrInvalidSetUpstreamArgs = errors.New("invalid set-upstream arguments")
 var ErrInvalidFetchSpec = errors.New("invalid fetch spec")
+var ErrPullOnlyNoUpstream = errors.New("There is no tracking information for the current branch.\nPlease specify which branch you want to merge with.")
```
fulghum, line 48: How about naming this one `ErrPullWithNoRemoteAndNoUpstream`? I think that would be a little clearer. 

Also, it would be awesome to include the "next steps" users may need to take, like git is so good at doing. I included the git output below as an example. It would be great to tell people they can run git pull remote branch to make their pull work, or that they can set the upstream (we still need to implement dolt branch --set-upstream though, right?)
```
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=origin/<branch> branch3
```

go/libraries/doltcore/env/remotes.go
```diff
@@ -205,6 +207,9 @@ func NewPushOpts(ctx context.Context, apr *argparser.ArgParseResults, rsr RepoSt
 	switch src.GetType() {
 	case ref.BranchRefType:
 		remoteRef, err = GetTrackingRef(dest, remote)
+		if !setUpstream {
+			setUpstream = gotRemoteAndBranch
```
fulghum, line 211: I added a top-level comment for this, too, but wanted to mention it with the code, too. I don't think this is a behavior that Git performs (automatically setting upstream on push when remote and branch are specified), but let me know if you think I'm just missing something.

go/libraries/doltcore/env/remotes.go
```diff
@@ -372,6 +377,19 @@ func NewPullSpec(ctx context.Context, rsr RepoStateReader, remoteName string, sq
...
 	}
 
+	trackedBranches, err := rsr.GetBranches()
+	if err != nil {
+		return nil, err
+	}
+
+	if _, hasUpstream := trackedBranches[branch.GetPath()]; !hasUpstream {
+		if remoteOnly {
+			return nil, fmt.Errorf("You asked to pull from the remote '%s', but did not specify a branch. Because this is not the default configured remote for your current branch, you must specify a branch on the command line.", remoteName)
```
fulghum, line 387: (minor) Would be nice to throw this into a named Err var, too, just like the Err var below. Maybe called `ErrPullWithRemoteNoUpstream`?

go/libraries/doltcore/sqle/dfunctions/dolt_checkout.go
```diff
@@ -121,10 +122,28 @@ func checkoutRemoteBranch(ctx *sql.Context, dbName string, dbData env.DbData, ro
 		return ErrEmptyBranchName
 	}
 
-	if ref, refExists, err := actions.GetRemoteBranchRef(ctx, dbData.Ddb, branchName); err != nil {
+	if remoteRef, refExists, err := actions.GetRemoteBranchRef(ctx, dbData.Ddb, branchName); err != nil {
 		return errors.New("fatal: unable to read from data repository")
 	} else if refExists {
-		return checkoutNewBranch(ctx, dbName, dbData, roots, branchName, ref.String())
+		err = checkoutNewBranch(ctx, dbName, dbData, roots, branchName, remoteRef.String())
+		if err == nil {
```
fulghum, line 129: This is another good place where the code will be slightly cleaner if you deal with the `err != nil` case first, return the error, and then handle the happy path. Nice to get the err handling out of the way immediately and then you won't have the extra level of nesting.

integration-tests/bats/remotes.bats
```diff
@@ -1001,7 +1090,12 @@ SQL
     [ "$status" -eq 0 ]
 }
 
-@test "remotes: DOLT_CHECKOUT to checkout to a remote branch." {
+@test "remotes: DOLT_CHECKOUT new branch without '-b' to checkout to set upstream to a remote branch." {
```
fulghum, line 1093: some of these test names could be a little more descriptive/clear. 

For example, from the names, I have a hard time understanding the difference in intent between this test and the one on line 123, named: "remotes: 'dolt checkout remote_branch_name' sets upstream".

fulghum (commented):
Lots of good fixes in this PR!! Good job digging in deep and finding all these details. No major concerns from me; let me know if any comments don't make sense. 

Two larger comments/thoughts popped out at me while reviewing, too:
* there is a LOT of duplication between the CLI path and SQL path! 😊 It'll be hard to keep those exactly in sync, just like you found. There's an interesting question about when to refactor a little bit more while we're in here and see if we can chip away at the duplication. For example, with dolt pull, it seems like there might be some low hanging fruit around combining the code paths. Not necessary for this PR, but something to think about as we're in this code. 
* we should talk more about layers of testing between BATS and the enginetests. I both are super important, and could be a good conversation to talk through a mental framework for what to test where.

integration-tests/bats/remotes.bats
```diff
@@ -1030,6 +1133,12 @@ SQL
     [ "$status" -eq 0 ]
     [[ "$output" =~ "pk" ]] || false
     [[ "$output" =~ "1" ]] || false
+
+    skip # above checkout command should set upstream persisting outside of session
```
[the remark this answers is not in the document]
jennifersp: We store upstream tracking metadata in `repo_state.json` file, and we do not have access to that file from inside the SQL session. As long as, user is in the same session, any changes on upstream tracking works fine, but once out of the session, the changes made inside the session will not be saved since we cannot access the repo_state file

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -132,6 +132,7 @@ func DoPush(ctx context.Context, rsr env.RepoStateReader, rsw env.RepoStateWrite
 			if err != nil {
 				return err
 			}
+			// TODO : set upstream should be persisted outside of session
```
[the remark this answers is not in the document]
jennifersp: Ah yes, I meant to add the comment in `dolt_push.go` after `DoPush()` was called.

go/libraries/doltcore/env/remotes.go
jennifersp: Right! I'll add this with `dolt push --set-upstream <remote> <branch>` option instead of `branch --set-upstream` for now.

integration-tests/bats/remotes.bats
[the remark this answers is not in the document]
fulghum: Thanks for the extra context. If we don't have a tracking issue for this, could you create one? Seems like something that will confuse/surprise customers and would be great to close this gap before someone falls in it.

go/libraries/doltcore/env/actions/table.go
```diff
@@ -141,11 +141,18 @@ func GetRemoteBranchRef(ctx context.Context, ddb *doltdb.DoltDB, name string) (r
...
 	}
 
+	// This will only do this behavior of guessing the remote branch if there is exactly one remote tracking branch with the matching branch name.
+	var found = false
+	var remoteRef ref.RemoteRef
 	for _, rf := range refs {
 		if remRef, ok := rf.(ref.RemoteRef); ok && remRef.GetBranch() == name {
-			return remRef, true, nil
+			if found {
+				return ref.RemoteRef{}, false, nil
```
fulghum, line 150: Nice job adding this in to resolve any ambiguity when there are multiple remote tracking branches that match the new branch name. 

It would be nice to return an error message in this case, to match Git's behavior. Here's what Git sends back: 
```
hint: If you meant to check out a remote tracking branch on, e.g. 'origin',
hint: you can do so by fully qualifying the name with the --track option:
hint: 
hint:     git checkout --track origin/<name>
hint: 
hint: If you'd like to always have checkouts of an ambiguous <name> prefer
hint: one remote, e.g. the 'origin' remote, consider setting
hint: checkout.defaultRemote=origin in your config.
fatal: 'branch10' matched multiple (2) remote tracking branches
```

I think there are two good options for updating this method... 
1 - since this is only used by checkout, you could directly return an error, probably just the last line of the error message above, so people know why checkout failed, although if you can include a hint, that's even better of course. :-) 
2 – alternatively, you could have this method return all the matching remote tracking branches and have both callers check that only one is returned, and then send that error. 

Either way, it would be good to update the method comments to mention that aspect of the method's contract.

fulghum (approved):
Nice work! Just one comment on a small tweak for an error message to match Git's response a little closer.

PR #3727, opened 2022-06-30
Build and install for for Linux OS on Mac arm64/aarch64 hardware

fulghum:
I tested the build binaries script locally and confirmed it is producing the new `dolt-linux-arm64.tar.gz` artifact. I also ran the install script on an Ubuntu container and verified that it tries to download that same artifact from GitHub, then copied the new artifact over and verified that the dolt binary executes as expected. 

Anything else I need to do to get that new artifact to show up on GitHub in the next release? 

Resolves: https://github.com/dolthub/dolt/issues/3726


reltuk (approved):
Looks great! Thanks for fixing this :D.

PR #3742, opened 2022-07-01
Updating release workflow to include new Linux-ARM64 build

fulghum:
I previously updated the build and install scripts to support Linux OS running on Mac hardware (https://github.com/dolthub/dolt/pull/3727), but I missed this update to the GitHub workflow to copy the new artifact over and make it available on the GitHub release page.


coffeegoddd (approved):
LGTM!

PR #3743, opened 2022-07-01
fix for datetime and time types for exporting parquet file

jennifersp:
Fixes https://github.com/dolthub/dolt/issues/3517 

Issues for importing parquet file from Dolt export to pandas.
- Datetime type is exported as timestamp to avoid out of range issues.
- Time type is exported as timedelta to avoid being parsed as time and panics if val is negative

### DIFF
--- a/go/libraries/doltcore/table/typed/parquet/writer.go
+++ b/go/libraries/doltcore/table/typed/parquet/writer.go
@@ -42,7 +42,7 @@ var typeMap = map[typeinfo.Identifier]string{
 	typeinfo.EnumTypeIdentifier:       "type=BYTE_ARRAY, convertedtype=UTF8",
 	typeinfo.InlineBlobTypeIdentifier: "type=BYTE_ARRAY, convertedtype=UTF8",
 	typeinfo.SetTypeIdentifier:        "type=BYTE_ARRAY, convertedtype=UTF8",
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/table/typed/parquet/writer.go
```diff
@@ -114,10 +114,13 @@ func (pwr *ParquetWriter) WriteSqlRow(ctx context.Context, r sql.Row) error {
...
 			case typeinfo.DatetimeTypeIdentifier:
-				val = val.(time.Time).Unix()
+				val = val.(time.Time).UnixMicro()
 				sqlType = sql.Int64
 			case typeinfo.TimeTypeIdentifier:
-				val = int64(val.(sql.Timespan))
+				v := int64(val.(sql.Timespan))
+				// take abs(val) TODO : negative timespan goes out of range when imported into pandas
+				//shift := v >> 63
+				val = v //(v ^ shift) - shift
```
fulghum, line 123: Looks like this logic hasn't changed at all? I couldn't figure out what the comments were for. Is this still work in progress?

jennifersp: oops, I was doing more testing on negative value. I added abs(v) function to avoid negative value causing panic for now.

fulghum (approved):
Code changes look good. 

Any idea what's up with that failing sysbench runner test?

jennifersp:
Trying running sysbench runner test again

PR #3745, opened 2022-07-01
dolt_clone stored procedure

zachmu:
Adds support for a new `dolt_clone()` stored procedure. 

Removes most of the use of `os.Chdir` (except for in test code and one or two dolt-cli-only spots). 

Doc Updates: https://github.com/dolthub/docs/pull/804


zachmu:
This needs tests but I'm not going to get to them before taking off on vacation. I'm putting it up for review because Jennifer has already merged some of these changes into her own branch, so it seems wise to get it checked in before it diverges too much. I can write bats tests when I'm back in town.

zachmu:
Actually did add a bats test, but it needs more

fulghum:
I'll start digging into this one, review, and add some more tests so we can unblock Jennifer's dolt_remote change.

fulghum (approved):
Looks pretty good. I noticed a couple spots where some error handling is needed. I'll flesh out the BATS tests a little more and fill in the error handling, then get this merged in and add some docs.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -212,6 +215,87 @@ func (p DoltDatabaseProvider) CreateDatabase(ctx *sql.Context, name string) erro
 	return dsess.AddDB(ctx, dbstate)
 }
 
+// CloneDatabaseFromRemote implements DoltDatabaseProvider interface
+func (p DoltDatabaseProvider) CloneDatabaseFromRemote(ctx *sql.Context, dbName, branch, remoteName, remoteUrl string, remoteParams map[string]string) error {
+	p.mu.Lock()
```
fulghum, line 220: This is potentially a loooong time to be holding a lock! I think this is okay for v1, since `dolt_clone()` will not be frequently used, but it would be good to follow up and shrink down the code that executes with this lock.

andrew-wm-arthur (approved):
LGTM

PR #3748, opened 2022-07-02
Adds `our_diff_type` and `their_diff_type` to `dolt_conflicts_table_name`

druvv:
`our_diff_type` and `their_diff_type` reports the row diff between the base and our/their rows. It is one of `added`, `modified`, or `removed`.

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/conflicts_tables_prolly.go
+++ b/go/libraries/doltcore/sqle/dtables/conflicts_tables_prolly.go
@@ -18,11 +18,11 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/dolthub/dolt/go/libraries/doltcore/merge"
 	"github.com/dolthub/go-mysql-server/sql"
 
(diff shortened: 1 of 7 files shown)


go/libraries/doltcore/sqle/dtables/conflicts_tables_prolly.go
```diff
@@ -240,10 +242,22 @@ func (itr *prollyConflictRowIter) putConflictRowVals(ctx *sql.Context, c conf, r
...
 		}
 	}
+	r[itr.t+itr.kd.Count()+itr.theirsVD.Count()] = getDiffType(c.bV, c.tV)
 
 	return nil
 }
 
+func getDiffType(base val.Tuple, other val.Tuple) string {
+	if base == nil {
+		return "added"
```
zachmu, line 252: Aren't there constants for these strings defined somewhere? There should be, i would think either in the merge package or the doltdb package

zachmu (approved):
LGTM

PR #3755, opened 2022-07-05
CLI write commands server-locked

max-hoffman:


go/libraries/doltcore/env/environment.go
```diff
@@ -1281,25 +1281,25 @@ func (dEnv *DoltEnv) BulkDbEaFactory() editor.DbEaFactory {
 	return editor.NewBulkImportTEAFactory(dEnv.DoltDB.Format(), dEnv.DoltDB.ValueReadWriter(), dEnv.TempTableFilesDir())
 }
 
-func (dEnv *DoltEnv) lockFile() string {
+func (dEnv *DoltEnv) LockFile() string {
 	return filepath.Join(dbfactory.DoltDir, ServerLockFile)
```
zachmu, line 1285: Needs to be an absolute path

zachmu (approved):
Just one comment otherwise LGTM

PR #3759, opened 2022-07-05
Add `--track` flag to dolt checkout

jennifersp:
Adds `--track` flag for dolt checkout.
Fixes local branch tracking remote tracking branch with different names
- `dolt checkout --track origin/feature` will create new branch named `feature` and set its upstream to `origin/feature` remote branch
- `dolt checkout -b branch1 --track origin/feature` will create new branch named `branch1` and set its upstream to `origin/feature` remote branch
- `dolt checkout -b branch1 origin/feature` will guess remote branch. If remote branch `origin/feature` exists, it will set upstream to it.

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -95,6 +95,7 @@ const (
 	DeleteFlag       = "delete"
 	DeleteForceFlag  = "D"
 	OutputOnlyFlag   = "output-only"
+	TrackFlag        = "track"
 )
 
 const (
@@ -163,6 +164,7 @@ func CreateCheckoutArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParser()
 	ap.SupportsString(CheckoutCoBranch, "", "branch", "Create a new branch named {{.LessThan}}new_branch{{.GreaterThan}} and start it at {{.LessThan}}start_point{{.GreaterThan}}.")
 	ap.SupportsFlag(ForceFlag, "f", "If there is any changes in working set, the force flag will wipe out the current changes and checkout the new branch.")
+	ap.SupportsString(TrackFlag, "t", "direct|inherit", "When creating a new branch, set up 'upstream' configuration. Defaults to 'direct'.")
 	return ap
 }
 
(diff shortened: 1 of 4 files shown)


go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -163,6 +164,7 @@ func CreateCheckoutArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParser()
 	ap.SupportsString(CheckoutCoBranch, "", "branch", "Create a new branch named {{.LessThan}}new_branch{{.GreaterThan}} and start it at {{.LessThan}}start_point{{.GreaterThan}}.")
 	ap.SupportsFlag(ForceFlag, "f", "If there is any changes in working set, the force flag will wipe out the current changes and checkout the new branch.")
+	ap.SupportsString(TrackFlag, "t", "direct|inherit", "When creating a new branch, set up 'upstream' configuration. Defaults to 'direct'.")
```
fulghum, line 167: Probably better to not document `inherit` if it's not supported yet, otherwise customers might read the documentation, see it, try to use it, and then be frustrated that it doesn't work after we advertised it in the docs. If it's not documented, it's less noise, and we can always go back and add it. 

If you think this will be a really fast follow and already know how to implement inherit, then I'm okay with leaving it in, but I think it's better to leave it out until we have support for it.

go/cmd/dolt/commands/checkout.go
```diff
@@ -144,6 +142,74 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
+			return errhand.VerboseErrorFromError(fmt.Errorf("--track='inherit' is not supported yet"))
+		}
+		remoteName, remoteBranchName = parseRemoteBranchName(startPt)
+		remotes, err := dEnv.RepoStateReader().GetRemotes()
+		if err != nil {
+			return errhand.BuildDError(err.Error()).Build()
+		}
+		_, remoteOk := remotes[remoteName]
+		if !remoteOk {
+			return errhand.BuildDError(fmt.Errorf("'%s' is not a commit and a branch '%s' cannot be created from it", startPt, remoteBranchName).Error()).Build()
```
fulghum, line 169: Does this match the Git error message? "is not a commit" seemed a little confusing to me, since we know this should be a remote/branch in this block, since the track flag is set. I'm wondering if "is not a valid remote ref" or something.

go/cmd/dolt/commands/checkout.go
```diff
@@ -266,20 +316,21 @@ func checkoutBranch(ctx context.Context, dEnv *env.DoltEnv, name string, force b
 
 // setRemoteUpstreamForCheckout sets upstream for checked out branch. This applies `dolt checkout <bn>`,
 // if <bn> matches any remote branch name. This should not happen for `dolt checkout -b <bn>` case.
-func setRemoteUpstreamForCheckout(dEnv *env.DoltEnv, branchRef ref.RemoteRef) errhand.VerboseError {
-	refSpec, err := ref.ParseRefSpecForRemote(branchRef.GetRemote(), branchRef.GetBranch())
+func setRemoteUpstreamForCheckout(dEnv *env.DoltEnv, remName, remBranchName string) errhand.VerboseError {
```
fulghum, line 319: (minor/nit) `remote` and `remoteBranch` might be easier to read variable names. I stopped for a second to think about what `remName` meant. It's not a bad name, but calling it `remote` would be even more obvious I think.

fulghum (approved):
Great changes! Just a couple of minor comments.

PR #3760, opened 2022-07-05
[no-release-notes] go: Carry a NodeStore in DoltDB and datas.database, instead of shim'ing a new NodeStore from the VRW when we need one.

reltuk:


andrew-wm-arthur (approved):
LGTM 🧵

PR #3766, opened 2022-07-06
[no-release-notes]Blob internal chunksize different than leaf chunksize

max-hoffman:
Blob chunker panic fix. If we want to split at a chunk size of ex: 4000 bytes, then we can only fit 4000/20 hashes into internal nodes.


fulghum (approved):
Looks good! Thanks for grabbing this one and for unskipping the JSON tests, too.

PR #3774, opened 2022-07-07
add `dolt_tags` system table and `dolt_tag()` procedure

jennifersp:
Adds `dolt_tags` system table and `DOLT_TAG()` procedure.


go/libraries/doltcore/sqle/dfunctions/dolt_tag.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/dolthub/dolt/go/cmd/dolt/cli"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
+)
+
+const DoltTagFuncName = "dolt_tag"
+
+// DoltTagFunc runs a `dolt commit` in the SQL context, committing staged changes to head.
+// Deprecated: please use the version in the dprocedures package
```
fulghum, line 32: You can just move this code to the stored procedure package. No need to add something as deprecated in its first version.

go/libraries/doltcore/sqle/dfunctions/dolt_tag.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (d DoltTagFunc) String() string {
+	childrenStrings := make([]string, len(d.Children()))
+
+	for i, child := range d.Children() {
+		childrenStrings[i] = child.String()
+	}
+
+	return fmt.Sprintf("DOLT_BRANCH(%s)", strings.Join(childrenStrings, ","))
```
fulghum, line 50: copy/paste typo

go/libraries/doltcore/sqle/dfunctions/dolt_tag.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (d DoltTagFunc) Eval(ctx *sql.Context, row sql.Row) (interface{}, error) {
+	args, err := getDoltArgs(ctx, row, d.Children())
+	if err != nil {
+		return 1, err
+	}
+	return DoDoltTag(ctx, args)
+}
+
+func DoDoltTag(ctx *sql.Context, args []string) (int, error) {
```
fulghum, line 69: Add some godocs for any exported functions and consider it for any complex private functions.

fulghum (commented):
Looking good! Just a few comments. Biggest one is making sure we have some tests that deeply cover this code in the Go enginetest layer.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4634,3 +4634,148 @@ var DoltVerifyConstraintsTestScripts = []queries.ScriptTest{
...
+				Expected: []sql.Row{{0}},
+			},
+			{
+				Query:    "SELECT tag_name, message from dolt_tags",
+				Expected: []sql.Row{{"v2", "create tag v2"}},
+			},
+		},
+	},
+	{
+		Name: "dolt-tag: SQL delete multiple tags",
```
fulghum: (minor) looks like this name was copied; should probably be something about using a created tag or something?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4634,3 +4634,148 @@ var DoltVerifyConstraintsTestScripts = []queries.ScriptTest{
 		},
 	},
 }
+
+var DoltTagTestScripts = []queries.ScriptTest{
```
fulghum, line 4638: These go enginetests are awesome! Thanks for adding them. 

It would be nice to have some coverage over other columns in the `dolt_tags` table. tagger and email should be set to test values that don't change. date and tag_hash will change with each run of course, but you could still assert that they are not null.

fulghum (approved):
Beautiful code. Nice work Jennifer! 😁

PR #3779, opened 2022-07-07
Allow backticks in index names

fulghum:
Fixes: https://github.com/dolthub/dolt/issues/3776

Related to: https://github.com/dolthub/go-mysql-server/pull/1109


go/libraries/doltcore/doltdb/table.go
```diff
@@ -43,7 +43,7 @@ const (
 	// IndexNameRegexStr is the regular expression that valid indexes must match.
 	// From the unquoted identifiers: https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
 	// We also allow the '-' character from quoted identifiers.
-	IndexNameRegexStr = `^[-$_0-9a-zA-Z]+$`
+	IndexNameRegexStr = "^[-`$_0-9a-zA-Z]+$"
```
zachmu, line 46: Seems like this is going to keep coming up over and over with new characters.

MySQL's only rules for identifier names are for when they are unquoted. If they quote the identifier they can use any unicode character except null.

I don't see any reason to restrict names except that they must have at least one non-whitespace character in them. Is there a technical reason for being stricter than MySQL here? Something in noms?

```
Permitted characters in quoted identifiers include the full Unicode Basic Multilingual Plane (BMP), except U+0000:

ASCII: U+0001 .. U+007F

Extended: U+0080 .. U+FFFF

ASCII NUL (U+0000) and supplementary characters (U+10000 and higher) are not permitted in quoted or unquoted identifiers.
```

zachmu (approved):
LG but it seems like we should probably be much less strict about checking identifier names. If it parses, we should accept it.

go/libraries/doltcore/doltdb/table.go
fulghum: Good feedback. Captured this one in a tracking issue: https://github.com/dolthub/dolt/issues/3789

PR #3783, opened 2022-07-08
go: doltdb: commit_hooks,hooksdatabase: Ensure that replication hooks work for replicating HEADs when the commit itself is already on the remote.

reltuk:


max-hoffman (approved):
LGTM

PR #3793, opened 2022-07-11
doltcore/sqle/dsess: Return "serialization failure" `ER_LOCK_DEADLOCK (1213)` on transaction merge conflict, remove `DOLT_TRANSACTION_MERGE_STOMP` flag and session var

andrew-wm-arthur:
This change removes the ability to stomp merge conflicts stemming from transaction commit. Instead, transaction merge conflicts return [`ER_LOCK_DEADLOCK (1213)`](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_lock_deadlock) which signals to the client that the transaction has been rolled back due to failures in locking and/or transaction isolation guarantees. Part of this signal is the ANSI SQLSTATE code of `40001` that is associated with `ER_LOCK_DEADLOCK`. This code means "serialization failure", which is a more accurate description of the error being thrown by transaction merge conflicts. 

In this context `ER_LOCK_DEADLOCK (1213)` is a misnomer, and is somewhat misleading to the user. However, this is the only MySQL error code associated with transaction rollback due to serialization failures, and in particular it is recognized by the TPC-C benchmarking utility that we use. This change is partially motivated by Postgres's implementation of [`REPEATABLE READ`](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-REPEATABLE-READ) which is similar to snapshot isolation.

companion GMS PR: https://github.com/dolthub/go-mysql-server/pull/1112


andrew-wm-arthur:
#benchmark

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -685,11 +685,12 @@ func TestTransactions(t *testing.T) {
 	}
 }
 
-func TestTransactionStomping(t *testing.T) {
-	for _, script := range DoltTransactionMergeStompTests {
-		enginetest.TestTransactionScript(t, newDoltHarness(t), script)
-	}
-}
+// TODO(andy): convert to err tests
```
zachmu, line 688: Fix?

zachmu (approved):
LGTM, definitely in favor of eliminating the stomp setting and returning a standard error code clients know how means to retry the transaction.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
andrew-wm-arthur: definitely, need to update these, thx

PR #3796, opened 2022-07-11
support unix socket

jennifersp:
Depends on https://github.com/dolthub/go-mysql-server/pull/1114

Uses unix socket file to run local server for Unix and Unix-like systems.
Uses TCP for Windows.


go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -35,6 +36,8 @@ import (
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqlserver"
 )
 
+const SockAddr = "/tmp/mysql.sock"
```
fulghum, line 39: Following up on our chat yesterday... it would be ideal to mimic MySQL's behavior and put this in the expected location for different unix distros. That might require looking at the mysql source code to reverse engineer what they're doing.

go/cmd/dolt/commands/sqlserver/server_test.go
```diff
@@ -187,7 +187,7 @@ func TestServerGoodParams(t *testing.T) {
 
 func TestServerSelect(t *testing.T) {
 	env := dtestutils.CreateEnvWithSeedData(t)
-	serverConfig := DefaultServerConfig().withLogLevel(LogLevel_Fatal).WithPort(15300)
+	serverConfig := DefaultServerConfig().withLogLevel(LogLevel_Fatal).WithPort(15300).WithHost("127.0.0.1")
```
fulghum, line 190: Why did we need to add the 127.0.0.1 localhost address here?

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -263,16 +264,16 @@ func (cfg *commandLineServerConfig) DataDir() string {
...
-
-// withHost updates the host and returns the called `*commandLineServerConfig`, which is useful for chaining calls.
-func (cfg *commandLineServerConfig) withHost(host string) *commandLineServerConfig {
+// WithHost updates the host and returns the called `*commandLineServerConfig`, which is useful for chaining calls.
+func (cfg *commandLineServerConfig) WithHost(host string) *commandLineServerConfig {
 	cfg.host = host
 	return cfg
 }
 
+func (cfg *commandLineServerConfig) CfgDir() string {
```
fulghum, line 273: Can we get some godocs plz? 😄

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -392,6 +393,9 @@ func ValidateConfig(config ServerConfig) error {
 
 // ConnectionString returns a Data Source Name (DSN) to be used by go clients for connecting to a running server.
 func ConnectionString(config ServerConfig) string {
+	if runtime.GOOS != "windows" && config.Host() == "localhost" {
```
fulghum, line 396: Is `localhost` enough here? Should we also include `127.0.0.1`? I'm guessing you've probably already checked MySQL's behavior.

fulghum: We probably also want an option on `dolt sql-server` to disable unix socket support.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +350,23 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
...
 	return conn, sess
 }
+
+func TestDoltServerRunningUnixSocket(t *testing.T) {
+	// with default host = 'localhost', the server will use unix socket located in /tmp/mysql.sock
+	sc, serverConfig := startServerWithDefaultConfigOnly(t)
+	sc.WaitForStart()
+
+	t.Run("connecting mysql works", func(t *testing.T) {
+		conn, sess := newConnection(t, serverConfig)
```
fulghum, line 360: It would be good to somehow assert that we are actually using the unix socket. It could be verifying the DSN string contains "unix" or something, or maybe there's a way to query the connections and verify that it's a unix domain socket address or something. 

It might also be a good idea to test that we can connect over both a TCP socket and a unix socket at the same time, since they not mutually exclusive.

integration-tests/bats/sql-server.bats
```diff
@@ -1357,7 +1357,7 @@ databases:
     dolt commit -am "create three tables"
 
     cd ..
-    start_sql_server_with_args --user dolt -ltrace --no-auto-commit
+    start_sql_server_with_args --host 0.0.0.0 --user dolt -ltrace --no-auto-commit
```
fulghum, line 1360: What does the `0.0.0.0` address do for `dolt sql-server`? I looked in the docs and didn't see anything. Does this have a special meaning?

go/cmd/dolt/commands/sqlserver/server.go
fulghum: Where do we actually create this file? I didn't see that in here or in the GMS PR, but I could just be missing it.

fulghum (commented):
Looks like this is going in the right direction! I added some high-level comments, like we talked about yesterday, since you mentioned you're still working through more details. 

Mostly some small comments/questions, but two larger thoughts popped out at me:
* It would be awesome for us to start regularly including documentation update PRs linked as part of the final review when you get there. Thinking about the user docs from the start always helps me frame the feature and it's super helpful for the reviewer to see how things are supposed to work when they're reading the PR. Hopefully it'll also help us ensure we don't forget to follow up with docs if we can build it into our regular process, too. 
* I'm betting there are some interesting edge cases with the socket file (e.g. is MySQL already running and has a socket file created? Is another Dolt sql-server running already?) Seems like we'll need some flags/options on sql-server to potentially turn off socket file, maybe to specify an alternate location, and to define what happens if there are conflicts/problems. Looking at how MySQL handles these is probably a solid start.

integration-tests/bats/sql-server.bats
jennifersp: the client uses --host 0.0.0.0 to connect to this server in `sql-server-mysql.expect` file, so when I added unix socket support, it started failing, to match the host I added this here, the client host can be modified instead I think.

go/cmd/dolt/commands/sqlserver/serverconfig.go
jennifersp: If nothing is given along with running `dolt sql-server` it will use default host, https://github.com/dolthub/dolt/pull/3796/files#diff-e137b5485cf44698ce998c1e76e36fb0a608954e38f084323c87b04db6ba662fR39 , which is defined as `localhost` so it'll be always localhost in case of not defining host and port when running a server.

I agree we should add option for disable unix socket support, but this would require host and port number which it can fallback to default host `127.0.0.1` and port `3306` if none given I think.

go/cmd/dolt/commands/sqlserver/server_test.go
jennifersp: With no host defined, it will start local server with the new implementation which uses unix socket so there is no specific port being used that it defines here. So, the clients cannot connect using this port number to the server that started as local host.

go/cmd/dolt/commands/sqlserver/server.go
jennifersp: Ah I was creating this file from dolt, but found out that the `net` package that we use already handles creating this file when server starts and deleting it after server stops, https://github.com/dolthub/go-mysql-server/blob/main/server/listener.go#L28

go/cmd/dolt/commands/sqlserver/server_test.go
jennifersp: I think I misunderstood here, the docs says if port defined, it should be tcp connection, I'm fixing according to that

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -309,3 +305,32 @@ func getConfigFromServerConfig(serverConfig ServerConfig) (server.Config, error,
...
+	portAsString := strconv.Itoa(port)
+	hostPort := net.JoinHostPort(host, portAsString)
+	serverConf.Address = hostPort
+	if portInUse(hostPort) {
+		portInUseError := fmt.Errorf("Port %s already in use.", portAsString)
+		return server.Config{}, portInUseError
+	}
+	// if socket is defined with or without value -> unix
+	if serverConfig.Socket() != "" {
+		// unix
```
fulghum, line 324: (minor) I'm not sure this comment is adding anything. You could mention that this is a unix socket on line 322, but I dont' think this comment on this line is needed.

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -217,22 +222,29 @@ func (cfg *commandLineServerConfig) PersistenceBehavior() string {
...
 func (cfg *commandLineServerConfig) TLSCert() string {
 	return cfg.tlsCert
 }
 
+// RequireSecureTransport is true if the server should reject non-TLS connections.
 func (cfg *commandLineServerConfig) RequireSecureTransport() bool {
 	return cfg.requireSecureTransport
 }
 
+// DisableClientMultiStatements is true if we want the server to not
```
fulghum, line 240: AWESOME godocs! Thank you for helping fill these in while you were working in this file! 🎉

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -145,6 +146,7 @@ func (cmd SqlServerCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsInt(maxConnectionsFlag, "", "max-connections", fmt.Sprintf("Set the number of connections handled by the server (default `%d`)", serverConfig.MaxConnections()))
 	ap.SupportsString(persistenceBehaviorFlag, "", "persistence-behavior", fmt.Sprintf("Indicate whether to `load` or `ignore` persisted global variables (default `%s`)", serverConfig.PersistenceBehavior()))
 	ap.SupportsString(commands.PrivsFilePathFlag, "", "privilege file", "Path to a file to load and store users and grants. Defaults to $doltcfg-dir/privileges.db")
+	ap.SupportsString(socketFlag, "", "socket file", "Path to a unix socket file. Defaults to '/tmp/mysql.sock'")
```
fulghum, line 149:
```suggestion
	ap.SupportsString(socketFlag, "", "socket file", "Path for the unix socket file. Defaults to '/tmp/mysql.sock'")
```

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -329,10 +329,78 @@ func startServer(t *testing.T) (*sqlserver.ServerController, sqlserver.ServerCon
...
+		// Assertions
+		rows, err := sess.Query("select 1")
+		require.NoError(t, err)
+		assertResultsEqual(t, []sql.Row{{1}}, rows)
+		require.NoError(t, conn.Close())
+	})
+
+	t.Run("connecting to local server with both tcp and socket connections", func(t *testing.T) {
+		// default connection
```
fulghum, line 369: probably worth mentioning the unix socket in this comment

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -393,6 +426,9 @@ func ValidateConfig(config ServerConfig) error {
 
 // ConnectionString returns a Data Source Name (DSN) to be used by go clients for connecting to a running server.
 func ConnectionString(config ServerConfig) string {
+	if runtime.GOOS != "windows" && config.Host() == "localhost" {
```
fulghum, line 429: (minor) This could probably benefit from a helper method to make this more self-descriptive and easier to read. Maybe something like `shouldUseUnixSocket() boolean` or something similar? Seemed like there were a couple of places where that could be used to make the code look a little simpler.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -329,10 +329,78 @@ func startServer(t *testing.T) (*sqlserver.ServerController, sqlserver.ServerCon
...
+
 func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Connection, *dbr.Session) {
 	const dbName = "dolt"
 	conn, err := dbr.Open("mysql", sqlserver.ConnectionString(serverConfig)+dbName, nil)
 	require.NoError(t, err)
 	sess := conn.NewSession(nil)
 	return conn, sess
 }
+
+func TestDoltServerRunningUnixSocket(t *testing.T) {
```
fulghum, line 354: It would be good to assert that this code really is using the unix socket and not connecting over TCP. One easy way to do this is to ask your serverConfig struct for the ConnectionString, and then just assert that it has a unix socket in it.

fulghum: I think there are some other edge cases that we should be testing... 
* assert that the socket file is not used in specific configurations, such as when host or port are set.
* assert that the server fails to start up if there is already a file in the socket file path (and of course gives a helpful, actionable error message about how to proceed)

You might consider BATS for some of these higher-level, end-to-end tests.

fulghum: We should also make sure there's some indication to customers in the logs that the server is listening on the socket file. 

For example, this startup log message should mention the socket and socket file path:
```
Starting server with Config HP="localhost:1234"|T="28800000"|R="false"|L="info"
```

fulghum (commented):
Looking good; just a few more details to round out.

go/cmd/dolt/commands/sqlserver/serverconfig.go
jennifersp: Thank you for suggesting this! I think this is really helpful, I'm going to try and make this a habit :)

go/cmd/dolt/commands/sqlserver/yaml_config.go
```diff
@@ -368,44 +376,62 @@ func (cfg YAMLConfig) QueryParallelism() int {
...
+// CfgDir is the path to a directory to use to store the dolt configuration files.
 func (cfg YAMLConfig) CfgDir() string {
 	if cfg.CfgDirStr != nil {
 		return *cfg.CfgDirStr
 	}
 	return filepath.Join(cfg.DataDir(), defaultCfgDir)
 }
+
+// Socket is a path to the unix socket file
+func (cfg YAMLConfig) Socket() string {
```
fulghum, line 428: It would be good to have a test of this option from a YAML config file, but we can look into that as a follow-up.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -313,11 +314,21 @@ func assertResultsEqual(t *testing.T, expected []sql.Row, rows *gosql.Rows) {
 	}
 }
 
-func startServer(t *testing.T) (*sqlserver.ServerController, sqlserver.ServerConfig) {
+func startServer(t *testing.T, withPort bool, host string, unixSocketPath string) (*sqlserver.ServerController, sqlserver.ServerConfig) {
```
fulghum, line 317: godocs for this function would help, especially as we're adding many more arguments to this function.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +347,92 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
 	sess := conn.NewSession(nil)
 	return conn, sess
 }
+
+func TestDoltServerRunningUnixSocket(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("we define socket flag and unix socket cannot be created on Windows, so it will return error")
```
fulghum, line 353: You can keep this skip message short and sweet. Something like... 

```suggestion
		t.Skip("unix sockets not support on Windows")
```

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +347,92 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
...
+	}
+
+	// TODO : socket flag is defined in this startServer method for now
+	unixSocketPath := "/tmp/mysql.sock"
```
fulghum, line 357: (minor) To tidy up... you can:
* remove line 356 – I didn't think it was adding any extra info
* inline the `unixSocketPath` variable since it's only referenced once

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +347,92 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
...
+	// Running unix socket server
+	sc, serverConfig := startServer(t, false, "", unixSocketPath)
+	sc.WaitForStart()
+	require.True(t, strings.Contains(sqlserver.ConnectionString(serverConfig, "dolt"), "unix"))
+
+	t.Run("connecting mysql works", func(t *testing.T) {
```
fulghum, line 363: I think you could probably pull out this test block, since you've got unix sockets covered here and in the test block below. Let me know if I'm missing something unique that this is testing though.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +347,92 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
...
+		require.NoError(t, conn3.Close())
+		require.NoError(t, conn2.Close())
+		require.NoError(t, conn1.Close())
+		require.NoError(t, localConn.Close())
+	})
+
+	// Stopping unix socket server
+	sc.StopServer()
+	err := sc.WaitForClose()
+	require.NoError(t, err)
```
fulghum, line 411: (minor) Might be worth a quick test that the socket file doesn't exist any more.

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -336,3 +347,92 @@ func newConnection(t *testing.T, serverConfig sqlserver.ServerConfig) (*dbr.Conn
...
+
+	// Running TCP socket server
+	tcpSc, tcpServerConfig := startServer(t, true, "0.0.0.0", "")
+	tcpSc.WaitForStart()
+	require.False(t, strings.Contains(sqlserver.ConnectionString(tcpServerConfig, "dolt"), "unix"))
+
+	t.Run("host and port specified, there should not be unix socket created", func(t *testing.T) {
+		// unix socket connection should fail
+		localServerConfig := sqlserver.DefaultServerConfig()
```
fulghum, line 420: Wouldn't `DefaultServerConfig()` return a config with unix socket set to "", so the connection on the next line wouldn't try to use a unix socket? Let me know if I'm just missing something obvious...

fulghum (approved):
Code looks good. Thanks for switching this to be opt-in for the first release. 

Mostly just minor comments that should be quick and a potential follow-up or two that we can handle later.

go/cmd/dolt/commands/sqlserver/yaml_config.go
jennifersp: I'll add this to my note, so I can add them when I come back

jennifersp: I just added one :)

PR #3802, opened 2022-07-11
go/libraries/doltcore/sqle: read_replica_database.go: Fix race condition when we replicate the same head on more than one session.

reltuk:
Previously StartTransaction would pull the new head from the upstream on each
StartTransaction, but there was no concurrency control to ensure that pulls
that started earlier didn't attempt to overwrite pulls that started later. The
end result was that some transactions would fail spuriously because the old
commit was not a fast forward merge of the current commit.

This adds a concurrency limiter so that only one pull per branch is outstanding
at a time.


go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,3 +253,95 @@ func parseBranches(arg string) []string {
...
+		l.mu.Unlock()
+		select {
+		case r := <-ch:
+			return r.v, r.err
+		case <-ctx.Done():
+			go func() { <-ch }()
+			return nil, ctx.Err()
+		}
+	} else {
+		l.blocked[s] = new(blocked)
```
max-hoffman, line 321: would it be simpler if the initializing routine had its lifecycle managed the same as the others?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,3 +253,95 @@ func parseBranches(arg string) []string {
...
+	return res, err
+}
+
+func (l *limiter) finish(s string) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	b := l.blocked[s]
+	if len(b.waiters) != 0 {
+		go func() {
+			r, err := b.f()
```
max-hoffman, line 336: so f` is always the second callback? even if the first routine is non-nil and there are > 2 routines?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,3 +253,95 @@ func parseBranches(arg string) []string {
...
+		if b.f == nil {
+			b.f = f
+		}
+		b.waiters = append(b.waiters, ch)
+		l.mu.Unlock()
+		select {
+		case r := <-ch:
+			return r.v, r.err
+		case <-ctx.Done():
+			go func() { <-ch }()
```
max-hoffman, line 317: should we close all of the channels?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,3 +253,95 @@ func parseBranches(arg string) []string {
...
+	if len(b.waiters) != 0 {
+		go func() {
+			r, err := b.f()
+			for _, ch := range b.waiters {
+				ch <- res{r, err}
+				close(ch)
+			}
+			l.finish(s)
+		}()
+		l.blocked[s] = new(blocked)
```
max-hoffman, line 343: the `blocked` KV management is different in these two conditions and i can't immediately tell why? seems like either is fine, but might be simpler if you defer one default?

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,3 +253,95 @@ func parseBranches(arg string) []string {
...
+
+	res, err := f()
+	l.finish(s)
+	return res, err
+}
+
+func (l *limiter) finish(s string) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	b := l.blocked[s]
```
max-hoffman, line 333: if `b` is ever nil, the `b.waiters` will panic

max-hoffman (approved):
lgtm!

PR #3808, opened 2022-07-12
dolt_checkout and read replication working set fixes

max-hoffman:
Sql path `dolt_checkout` used to create a new branch from HEAD
if the checkout name did not have a remote tracking ref. The CLI,
on the other hand, will initialize a local branch if a remote tracking
ref with the same name in the default remote exists. The SQl path
now mirrors the CLI checkout: if we checkout a new branch, like
'new_feature', and we have previously fetched that branch from our
default remote, `dolt_checkout('new_feature')` will initialize a branch
using the head commit from the remote `new_feature`.

Second, read replication was creating a local tracking branch for
remote branches (sourced from the master remote), but not creating
corresponding working sets. This let users attach to new branches with
`use db/new_feature` syntax, but prevented users from
`dolt_checkout('new_feature')` both for the bug in dolt checkout, but also
because we created an intermediary state with a local branch, but
without a corresponding working set. SQL replication will now fetch,
create the new branch, and create the appropriate working set.

### DIFF
--- a/go/libraries/doltcore/sqle/read_replica_database.go
+++ b/go/libraries/doltcore/sqle/read_replica_database.go
@@ -182,11 +182,10 @@ func pullBranches(ctx *sql.Context, rrd ReadReplicaDatabase, branches []string) 
 		{
 			branchExists, err := rrd.ddb.HasBranch(ctx, branch.GetPath())
 			switch {
+			case err != nil:
 			case branchExists:
 				err = rrd.ddb.FastForward(ctx, branch, srcDBCommit)
-			case err != nil:
 			default:
-				//err = actions.CreateBranchOnDB(ctx, rrd.ddb, branch.GetPath(), rtRef.GetPath(), )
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -154,69 +154,78 @@ func pullBranches(ctx *sql.Context, rrd ReadReplicaDatabase, branches []string)
...
-			return err
+		// either ff or create local tracking branch
+		{
+			branchExists, err := rrd.ddb.HasBranch(ctx, branch.GetPath())
+			switch {
+			case branchExists:
+				err = rrd.ddb.FastForward(ctx, branch, srcDBCommit)
+			case err != nil:
+			default:
+				//err = actions.CreateBranchOnDB(ctx, rrd.ddb, branch.GetPath(), rtRef.GetPath(), )
```
jennifersp, line 189:
```suggestion
```

jennifersp (approved):
LGTM!

PR #3810, opened 2022-07-12
more privilege changes

JCOR11599:
Changes:
 - add `--user` to `dolt sql` and `dolt sql -q`
 - default user is persisted immediately iff `--user`  or `--password` option is given
 - default super user is `root@localhost` instead of `root@%`
 - only create `doltcfgdir` if we need to persist something
 - has more tests

Companion PR: https://github.com/dolthub/go-mysql-server/pull/1127
 
 Fix for: https://github.com/dolthub/dolt/issues/3794
 Also a fix for: https://github.com/dolthub/dolt/issues/3881


Hydrocharged:
This needs a few changes as per our earlier conversation offline (with some additional comments):
* `dolt sql` does not need a `--password` argument, as the CLI has full access and it does not make logical sense to add password checking or even the ability to specify one
* The default user is the super user which is **not** persisted at all, and should actually be invisible to queries on the `mysql.user` table
* For both `dolt sql` and `dolt sql-server`, if the user given via `--user` exists then we assume that user. What this means for `dolt sql-server` is that, since users are attached to connecting clients, the `--user` directive is just ignored if the user exists. This would also mean that `--password` is ignored for existing users.
* If no user is specified for `dolt sql` then we can default to `root`. Perhaps we should default to a unique user name to guarantee that you're always the super user if no user is specified, but defaulting to `root` is also fine.
* If no user is specified for `dolt sql-server` and the privilege file does not exist then we create the default `root` super user. If a privilege file exists then we only create a super user if one was specified through `--user`. This prevents the situation where an admin is unaware that a root super user exists that others can log into (assuming they can connect as `localhost`).

Of course this would all also need to be in docs too.

go/cmd/dolt/commands/sql.go
```diff
@@ -570,8 +546,13 @@ func execMultiStatements(
 		return errhand.VerboseErrorFromError(err)
 	}
 
-	// Add root client
-	sqlCtx.Session.SetClient(sql.Client{User: "root", Address: "%", Capabilities: 0})
+	// check if user exists, else add them as superuser; might need to do something else for sql-server
+	if user := se.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.GetUser(config.ServerUser, config.ServerHost, false); user == nil {
+		se.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, config.ServerPass)
+	}
```
Hydrocharged, line 552: Are these functions called from `sql-server`? I'm fairly confident that these are not, so we shouldn't have to worry about that context.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -149,6 +151,16 @@ func Serve(
 	if err != nil {
 		return err, nil
 	}
+
+	// check if user exists, else add them as superuser; might need to do something else for sql-server
+	if user := sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.GetUser(config.ServerUser, config.ServerHost, false); user == nil {
+		sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, config.ServerPass)
+	}
```
Hydrocharged, line 158: This should be a bit different. Refer to the bullet points on my comment:
* For both `dolt sql` and `dolt sql-server`, if the user given via `--user` exists then we assume that user. What this means for `dolt sql-server` is that, since users are attached to connecting clients, the `--user` directive is just ignored if the user exists. This would also mean that `--password` is ignored for existing users.
* If no user is specified for `dolt sql-server` and the privilege file does not exist then we create the default `root` super user. If a privilege file exists then we only create a super user if one was specified through `--user`. This prevents the situation where an admin is unaware that a root super user exists that others can log into (assuming they can connect as `localhost`).

And just to clarify it a bit: the `--user` argument should only be creating super users. If the user already exists, then the argument is ignored. The default value of `root` is **only** used when no privilege file exists, otherwise we require an explicit `--user` argument to construct a super user.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -149,6 +151,16 @@ func Serve(
...
+
+	// TODO: maybe delete
+	//if len(config.ServerUser) == 0 {
+	//	sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.Persist(nil)
+	//}
```
Hydrocharged, line 163: Seems like you meant to delete this?

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -132,13 +131,13 @@ func (cmd SqlServerCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(configFileFlag, "", "file", "When provided configuration is taken from the yaml config file and all command line parameters are ignored.")
 	ap.SupportsString(hostFlag, "H", "host address", fmt.Sprintf("Defines the host address that the server will run on (default `%v`)", serverConfig.Host()))
 	ap.SupportsUint(portFlag, "P", "port", fmt.Sprintf("Defines the port that the server will run on (default `%v`)", serverConfig.Port()))
-	ap.SupportsString(userFlag, "u", "user", fmt.Sprintf("Defines the server user (default `%v`)", serverConfig.User()))
+	ap.SupportsString(commands.UserFlag, "u", "user", fmt.Sprintf("Defines the server user (default `%v`)", serverConfig.User()))
```
Hydrocharged, line 134: From my previous comment, this should mention that it should be explicit if desired.

integration-tests/bats/sql-privs.bats
```diff
@@ -60,12 +60,12 @@ teardown() {
     start_sql_server_with_config test_db server.yaml
 
     server_query test_db 1 "select user from mysql.user order by user" "User\ndolt"
-    run ls -a
-    [[ "$output" =~ ".doltcfg" ]] || false
-
     server_query test_db 1 "create user new_user" ""
     server_query test_db 1 "select user from mysql.user order by user" "User\ndolt\nnew_user"
 
+    run ls -a
```
Hydrocharged, line 66: This is for the file in general. I looks like you haven't added any new tests for the new functionality. We'll also need tests for the additions I mentioned previously.

Hydrocharged (commented):
Needs more testing, plus the changes I mentioned. Looks good besides that!

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -147,6 +149,19 @@ func Serve(
...
+	if sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.UserTable().Data().Count() != 0 {
+		// privileges specified, only add if superuser specified is not an existing user
+		userSpecified := config.ServerUser != defaultUser || config.ServerHost != defaultHost || config.ServerPass != defaultPass
+		superuser := sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.GetUser(config.ServerUser, "%", false)
+		if userSpecified && superuser == nil {
+			sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, "%", config.ServerPass)
+		}
+	} else {
+		// no privileges, must add superuser; will already be defaulted to root
+		sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, "%", config.ServerPass)
```
Hydrocharged, line 162: Shouldn't we be using `config.ServerHost`? Otherwise it looks like it's getting completely ignored. I'm assuming the default value is `localhost`, so it should be `%` (if it's not possible to have an empty default).

Hydrocharged (commented):
Have to at least figure out why the tests are failing on CI. Current server behavior for `--user` restricts people from using `root` as their account name for the simplest use cases, which I'm confident someone will run into.

go/cmd/dolt/commands/sqlserver/server.go
JCOR11599: We should, but the Sysbench and Compatibility tests fail with permission denied.
Default is localhost.

Hydrocharged: We can modify both the sysbench and compatibility tests if necessary. They should be able to work with empty defaults though, as the resulting logic should end up equivalent unless they’re specifying a root user.

Hydrocharged: Functionally equivalent in this particular case I mean.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -148,6 +150,37 @@ func Serve(
...
+	//	// privileges specified, only add if superuser specified is not an existing user
+	//	userSpecified := config.ServerUser != defaultUser || config.ServerHost != defaultHost || config.ServerPass != defaultPass
+	//	superuser := sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.GetUser(config.ServerUser, serverHost, false)
+	//	if userSpecified && superuser == nil {
+	//		sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, serverHost, config.ServerPass)
+	//	}
+	//} else {
+	//	// no privileges, must add superuser; will already be defaulted to root
+	//	sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb.AddSuperUser(config.ServerUser, serverHost, config.ServerPass)
+	//}
```
Hydrocharged, line 182: Delete this comment

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -148,6 +150,37 @@ func Serve(
 	}
 	defer sqlEngine.Close()
 
+	// should user "%" if host is empty string, same for 0.0.0.0
+	serverHost := config.ServerHost
+	if serverHost == "" || serverHost == "0.0.0.0" {
+		serverHost = "%"
```
Hydrocharged, line 156: Add a TODO to move this into GMS

go/cmd/dolt/commands/sqlserver/serverconfig.go
```diff
@@ -432,9 +431,9 @@ func ValidateConfig(config ServerConfig) error {
 	if config.Port() < 1024 || config.Port() > 65535 {
 		return fmt.Errorf("port is not in the range between 1024-65535: %v\n", config.Port())
 	}
-	if len(config.User()) == 0 {
-		return fmt.Errorf("user cannot be empty")
-	}
+	//if len(config.User()) == 0 {
+	//	return fmt.Errorf("user cannot be empty")
+	//}
```
Hydrocharged, line 436: Delete the comment

integration-tests/bats/migrate.bats
```diff
@@ -2,6 +2,7 @@
 load $BATS_TEST_DIRNAME/helper/common.bash
 
 setup() {
+    skip "temporary skip"
```
Hydrocharged, line 5: Remove this skip so that we'll properly see it fail, but since it's not from your changes we should still commit this

integration-tests/bats/sql-server.bats
```diff
@@ -231,7 +231,7 @@ SQL
     [[ "$output" =~ "one_pk" ]] || false
 
     # Add rows on the command line
-    run dolt sql -q "insert into one_pk values (1,1,1)"
+    run dolt sql --user=dolt -q "insert into one_pk values (1,1,1)"
```
Hydrocharged, line 234: We need a test that show the user assumption if the privilege file already exists. We can do this by setting a password in the privilege file and providing a different password as an argument, then attempting to log in.

Hydrocharged (approved):
A few simple changes and we're set!

PR #3813, opened 2022-07-12
support dolt_remote() stored procedure

jennifersp:
Supported options:
`DOLT_REMOTE('add', ...)`
`DOLT_REMOTE('remove', ...)`
`DOLT_REMOTE('rm', ...)`

Got `dialer` out of `env.Remote` object definition, dialer will be set when `GetRemoteDB()`  method is called in dolt push and pull


go/libraries/doltcore/sqle/dprocedures/dolt_remote.go
```diff
@@ -0,0 +1,153 @@
...
+	fs := sess.Provider().FileSystem()
+	r := env.NewRemote(remoteName, absRemoteUrl, params, nil)
+
+	repoState, err := env.LoadRepoState(fs)
+	if err != nil {
+		return err
+	}
+	repoState.AddRemote(r)
+	err = repoState.Save(fs)
+	if err != nil {
```
fulghum, line 104: (minor/nit) you could save a few lines of code here and just call `return repoState.Save(fs)`

integration-tests/bats/system-tables.bats
```diff
@@ -174,52 +174,61 @@ teardown() {
...
 @test "system-tables: insert into dolt_remotes system table" {
-    skip "Remotes table not yet mutable in SQL session"
     run dolt sql -q "insert into dolt_remotes (name, url) values ('origin', 'file://remote')"
     [ $status -ne 0 ]
     [[ ! "$output" =~ panic ]] || false
 
     mkdir remote
-    dolt sql -q "insert into dolt_remotes (name, url) values ('origin1', 'file://remote')"
-    dolt sql -q "insert into dolt_remotes (name, url) values ('origin2', 'aws://[dynamo_db_table:s3_bucket]/repo_name')"
+    dolt sql -q "CALL DOLT_REMOTE('add', 'origin1', 'file://remote/')"
```
fulghum, line 182: Looks like we needed to end with a trailing slash to get this test working? If so, that's a sharp edge that's worth looking into fixing so customers don't have to remember to always end remotes in slashes.

integration-tests/bats/system-tables.bats
```diff
@@ -174,52 +174,61 @@ teardown() {
...
+    dolt sql -q "CALL DOLT_REMOTE('add', 'origin2', 'aws://[dynamo_db_table:s3_bucket]/repo_name')"
 
     run dolt sql -q "select count(*) from dolt_remotes" -r csv
     [ $status -eq 0 ]
     [[ "$output" =~ 2 ]] || false
 
+    run dolt sql -q "select name, fetch_specs, params from dolt_remotes" -r csv
```
fulghum, line 189: It would be good to have some Go enginetest tests for `dolt_remote()`, too.

fulghum (commented):
Looking very good. I skipped over the `dolt_clone()` files, since Zach is still working on those and I know we'll pick up changes from his branch. The `dolt_remote()` work generally looks good though. Biggest comment was that we should add some go engine tests for this one (even if we can't use the remotes from within tests, we should still be able to add/remove them, right?).

After taking care of those and pulling in Zach's final `dolt_clone()` changes, this should be ready to go. 

Don't forget to update the stored procedure docs for `dolt_remote()`, too. Might be nice if you wanted to offer to fill those in for `dolt_clone()` for Zach, too.

integration-tests/bats/system-tables.bats
jennifersp: the test works with or without trailing slash :)

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4755,3 +4755,43 @@ var DoltTagTestScripts = []queries.ScriptTest{
...
+			{
+				Query:    "SELECT name, IF(CHAR_LENGTH(url) < 0, NULL, 'not null'), fetch_specs, params FROM DOLT_REMOTES",
+				Expected: []sql.Row{{"origin", "not null", sql.MustJSON(`["refs/heads/*:refs/remotes/origin/*"]`), sql.MustJSON(`{}`)}},
+			},
+		},
+	},
+	{
+		Name: "dolt-remote: SQL remove remotes",
+		SetUpScript: []string{
+			"CALL DOLT_REMOTE('add','origin1','file://.')",
```
fulghum, line 4777: Add an error case like `call dolt_remote()` or `call dolt_remote('origin')` or something, somewhere in the tests, just to get some coverage over the error handling code.

fulghum (commented):
Looking very good! Just a few last minor comments for the `dolt_remote` work. The only reason I didn't approve is because we're still waiting on Zach's `dolt_clone` updates.

jennifersp:
Made new PR, https://github.com/dolthub/dolt/pull/3862 with dolt_remote related changes + FS access copied from Zach's PR, https://github.com/dolthub/dolt/pull/3745

PR #3816, opened 2022-07-12
replicate delete branch

max-hoffman:
Replicate will push deletes on write in syncronous and async mode. Read replicas will delete refs that don't exist on the replica. Both of these can have destructive effects if not used carefully.


integration-tests/bats/replication.bats
```diff
@@ -58,6 +58,32 @@ teardown() {
...
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 2 ]
+    [[ "$output" =~ "t1" ]] || false
+
+    cd ../repo1
+    dolt checkout main
+    dolt sql -q "call dolt_branch('-df', 'feature');"
+
+    cd ..
+    dolt clone file://./bac1 repo3
```
timsehn, line 82: I think you may need an additional test, potentially skipped, where the branch already exists on the replica.

max-hoffman: added more tests

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,6 +223,58 @@ func pullBranches(ctx *sql.Context, rrd ReadReplicaDatabase, branches []string)
 	return nil
 }
 
+func getReplicationBranches(ctx *sql.Context, rrd ReadReplicaDatabase) ([]string, []ref.DoltRef, error) {
```
zachmu, line 226: Should generally name return params when they are ambiguous like this

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -230,6 +223,58 @@ func pullBranches(ctx *sql.Context, rrd ReadReplicaDatabase, branches []string)
...
+			deletedBranches = append(deletedBranches, localRefs[j])
+			j++
+		}
+	}
+	for i < len(srcRefs) {
+		allBranches = append(allBranches, srcRefs[i].GetPath())
+		i++
+	}
+	for j < len(localRefs) {
+		deletedBranches = append(deletedBranches, localRefs[i])
```
zachmu, line 260: `i` is never being incremented here, this is suspicious

This algorithm could use a quick unit test (deleted branches at the beginning, at the end, in the middle)

zachmu (approved):
LGTM

go/libraries/doltcore/sqle/read_replica_database.go
max-hoffman: yeah that aint right, thanks

PR #3823, opened 2022-07-13
go/{commands,doltcore}: Rewrote Dolt docs to use SQL path

andrew-wm-arthur:


go/libraries/doltcore/sqle/sqlselect_test.go
```diff
@@ -1489,19 +1488,19 @@ func TestCaseSensitivity(t *testing.T) {
...
-			NewRowWithSchema(doltdocs.DocsSchema,
-				types.String("LICENSE.md"),
-				types.String("A license")),
-		),
-		Query: "select * from dolt_docs",
-		ExpectedRows: ToSqlRows(CompressSchema(doltdocs.DocsSchema),
-			NewRow(types.String("LICENSE.md"), types.String("A license"))),
-		ExpectedSchema: CompressSchema(doltdocs.DocsSchema),
-	},
+	//{
```
zachmu, line 1491: ?

go/libraries/doltcore/sqle/sqldelete_test.go
```diff
@@ -189,14 +189,6 @@ func TestExecuteDeleteSystemTables(t *testing.T) {
 }
 
 var systemTableDeleteTests = []DeleteTest{
-	//{
```
zachmu, line 192: These tests should confirm that you can delete  / insert / update from the docs table

go/cmd/dolt/commands/docscmds/read.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+)
+
+var readDocs = cli.CommandDocumentationContent{
+	ShortDesc: "Reads Dolt docs to stdout",
+	LongDesc:  "Reads Dolt docs to stdout",
+	Synopsis: []string{
+		"[{{.LessThan}}dolt doc{{.GreaterThan}}]",
```
zachmu, line 37: Incorrect synopsis

zachmu (approved):
Overall this looks good. A few top-level comments

1) You've got read / write backwards in the command terminology. They should also be symmetric in terms of whether they read / write to files or stdin / stdout. I kind of like this:

```
dolt docs write LICENSE.MD [filename]
dolt docs read LICENSE.MD [filename]
```

Where if `[filename]` isn't given then you read or write a file with the same name as the doc specified. Or you could make these use stdin / stdout and provide a `--file` option. But they should be symmetric.

2) We should really have line-by-line diffs for the docs table. Can add this in a second pass.

3) Engine tests for DML statements on the docs table

PR #3824, opened 2022-07-13
Moving LargeJsonObject tests from go-mysql-server into Dolt

fulghum:
Moved LargeJsonObjects out of go-mysql-server (https://github.com/dolthub/go-mysql-server/pull/1115) since they were causing `go test -race` to fail for Dolt. These tests are now executed from a new Dolt-specific enginetest that is skipped by default in CI environments, unless the `DOLT_TEST_RUN_NON_RACE_TESTS` env var is set (same process used to skip the `TestLargeJsonObjects` unit test in Dolt).

I also added the `-v` flag to the dolthub/dolt enginetest race detection GH workflow so future failures have better output on exactly what failed. 

I ran these changes through go test -race enginetests on Ubuntu and confirmed it now passes.


max-hoffman (approved):
lgtm

PR #3827, opened 2022-07-14
Added table, index, and view caching for performance

zachmu:


zachmu:
#benchmark

go/libraries/doltcore/sqle/dsess/session_cache.go
```diff
@@ -0,0 +1,164 @@
...
+	"strings"
+
+	"github.com/dolthub/go-mysql-server/sql"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+)
+
+// SessionCache caches various pieces of expensive to compute information to speed up future lookups in the session.
+// No methods are thread safe.
+type SessionCache interface {
```
andrew-wm-arthur, line 27: do we need an interface for this? we don't have any polymorphism here

go/libraries/doltcore/sqle/tables.go
```diff
@@ -248,7 +248,12 @@ func (t *DoltTable) DataCacheKey(ctx *sql.Context) (doltdb.DataCacheKey, bool, e
 	if err != nil {
 		return doltdb.DataCacheKey{}, false, err
 	}
-	return doltdb.NewDataCacheKey(r), true, nil
+	key, err := doltdb.NewDataCacheKey(r)
+	if err != nil {
+		return doltdb.DataCacheKey{}, false, err
+	}
+
+	return key, true, nil
```
andrew-wm-arthur, line 256: we do we have a boolean signal here if we're not using it? future use?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -267,12 +272,46 @@ func (t *DoltTable) getRoot(ctx *sql.Context) (*doltdb.RootValue, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !ok {
+		tbl, err := t.DoltTable(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return index.DoltIndexesFromTable(ctx, t.db.Name(), t.tableName, tbl)
```
andrew-wm-arthur, line 285: we don't populate the cache here, but we also never his this block

andrew-wm-arthur (approved):
LGTM, couple comments

zachmu:
#benchmark

go/libraries/doltcore/sqle/tables.go
zachmu: This implementation always returns true, but there are others that only return false

go/libraries/doltcore/sqle/tables.go
zachmu: Adding a comment to clarify

PR #3838, opened 2022-07-15
Make unique key errors more consistent and fix unique key bats for new format

druvv:
Companion GMS PR: https://github.com/dolthub/go-mysql-server/pull/1116


go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -78,13 +77,12 @@ type indexWriter interface {
 	Commit(ctx context.Context) error
 	Discard(ctx context.Context) error
 	HasEdits(ctx context.Context) bool
-	UniqueKeyError(ctx context.Context, sqlRow sql.Row) error
+	getMut() prolly.MutableMap
+	errForSecondaryUniqueKeyError(ctx context.Context, err secondaryUniqueKeyError) error
```
andrew-wm-arthur, line 81: adding this methods here muddles this interface a bit. Instead I would expose a `RangeIter()` in place of `getMut()` and move `errForSecondaryUniqueKeyError` out of this interface.

You can either extent the interface to include this method, assert past it, or refactor it out of the object

andrew-wm-arthur (commented):
Generally looks good, need to clean up the `IndexWriter` interface

PR #3851, opened 2022-07-18
[no-release-notes] go/{serial, store/prolly}: dedicated Flatbuffer message for merge artifacts 

andrew-wm-arthur:


go/gen/fb/serial/mergeartifacts.go
```diff
@@ -0,0 +1,397 @@
...
+
+package serial
+
+import (
+	"strconv"
+
+	flatbuffers "github.com/google/flatbuffers/go"
+)
+
+type ArtifactType byte
```
druvv, line 25: `ArtifactType` might be confusing. We use the name ArtifactType for the field the describes what kind of conflict / constraint violation an entry is. Maybe ArtifactFormat?

druvv (approved):
Can we unskip bats tests using dolt gc in this PR?

PR #3852, opened 2022-07-19
[no-release-notes] In __DOLT_1__, throw error if schemas have different primary key types

druvv:
Unskips schema-changes.bats


go/libraries/doltcore/schema/schema.go
```diff
@@ -194,6 +195,9 @@ func ArePrimaryKeySetsDiffable(fromSch, toSch Schema) bool {
 		if (c1.Tag != c2.Tag) || (c1.IsPartOfPK != c2.IsPartOfPK) {
 			return false
 		}
+		if types.IsFormat_DOLT_1(format) && !c1.TypeInfo.Equals(c2.TypeInfo) {
```
andrew-wm-arthur, line 198: I think we want to compare the `sqltypes` instead: https://github.com/dolthub/go-mysql-server/blob/main/sql/type.go#L68

andrew-wm-arthur (commented):
Needs another pass

PR #3861, opened 2022-07-19
Add a better error message for schema mismatch

VinaiRachakonda:
With csv file imports we used to throw an error if the a read row did not have the same numbers of columns as the original schema. 

For example:

```
Rows Processed: 0, Additions: 0, Modifications: 0, Had No Effect: 0
A bad row was encountered while moving data.
csv reader's schema expects 3 fields, but line only has 2 values.
line: '5,5,'
These can be ignored using '--continue'
```

This is not particularly useful output when the size of the row is massive. It has been a headache debugging this issue for the recent museums bounty. This pr is changing the error message to as follows: 

```
Rows Processed: 0, Additions: 0, Modifications: 0, Had No Effect: 0
A bad row was encountered while moving data.
CSV reader expected 3 values, but saw 2.
The row was interpreted to be: '{
	"pk": "5"
	"v1": "5"
	"v2": ""
}
'
```

Also improve printing for skipping rows


go/libraries/doltcore/table/untyped/csv/reader.go
```diff
@@ -187,42 +187,57 @@ func (csvr *CSVReader) ReadSqlRow(crx context.Context) (sql.Row, error) {
...
-		return nil, table.NewBadRow(nil,
-			fmt.Sprintf("csv reader's schema expects %d fields, but line only has %d values.", schSize, len(colVals)),
-			fmt.Sprintf("line: '%s'", out.String()),
+
+		badMpStr, unusedRowValues := interpretRowSizeError(csvr.sch, rowVals)
+
+		args := []string{
+			fmt.Sprintf("CSV reader expected %d values, but saw %d.", schSize, len(rowVals)),
+			fmt.Sprintf("The row was interpreted to be: '%s'", badMpStr),
+		}
```
fulghum, line 212: How hard would it be to include the line number of the bad row? That seems like the gold standard for reporting an error in a data file and what customers would likely expect.

go/libraries/doltcore/table/untyped/csv/reader.go
```diff
@@ -455,3 +470,43 @@ func (csvr *CSVReader) parseQuotedField(rs *recordState) (kontinue bool, err err
...
+		if i >= len(rowVals) {
+			keyValPairs[i] = []string{col.Name, ""}
+		} else {
+			keyValPairs[i] = []string{col.Name, *rowVals[i]}
+		}
+	}
+
+	// 2. Append any unused row values to print to the user
+	for i := len(cols); i < len(rowVals); i++ {
+		if rowVals[i] == nil {
```
fulghum, line 493: (minor) do you need that Sprintf? Seems simpler just to use a literal value and not call Sprintf.

fulghum (approved):
Looks good! Nice changes to make the error messages more useful. My biggest thought is that it would be ideal to report the line number of the bad row – I think this is what customers would expect for an error from a data import tool.

go/libraries/doltcore/table/untyped/csv/reader.go
VinaiRachakonda: %q quotes the string. felt helpful for a map output

go/libraries/doltcore/table/untyped/csv/reader.go
VinaiRachakonda: I like this feature idea. I'll put it in a separate pr. It's something that works across file types

PR #3863, opened 2022-07-19
[no-release-notes] Stop reading/writing super schema

andrew-wm-arthur:
`SuperSchema` was initially added to address unsafe tag reuse. Since it was added, Dolt switched from constant-time column deletion to full table rewrites to remove associated values. 
Because of this, it is now safe to reuse the tag of a column that has been deleted, and column tags only need to be unique for the current Root, not the entire history.


Hydrocharged (approved):
Because of all of the deletions

PR #3867, opened 2022-07-20
Sql server jwt auth

tbantle22:
GMS PR: https://github.com/dolthub/go-mysql-server/pull/1123


integration-tests/bats/sql-jwt-auth.bats
```diff
@@ -0,0 +1,84 @@
...
+    teardown_common
+}
+
+
+@test "sql-server: jwt auth from config" {
+    cd repo1
+    cp "$BATS_TEST_DIRNAME"/../../go/cmd/dolt/commands/sqlserver/testdata/chain_key.pem .
+    cp "$BATS_TEST_DIRNAME"/../../go/cmd/dolt/commands/sqlserver/testdata/chain_cert.pem .
+    let PORT="$$ % (65536-1024) + 1024"
+    TOKEN="`cat $BATS_TEST_DIRNAME/token.jwt`"
```
tbantle22, line 30: Is this the right way to get the token from the jwt file? Seems like I might be doing something wrong in the python code below. When I run this test it hangs for a long time but if I generate a jwt and pass it in directly as the password it passes.

Also should I gitignore the jwks and jwt files? Seems like it would be annoying if there were changes every times someone runs bats

tbantle22: Ignore this. Got test to pass and am removing jwt/jwks files at the end of the test now

go/cmd/dolt/commands/engine/testfiles/test_jwks.json
```diff
@@ -0,0 +1 @@
+{"keys":[{"use":"sig","kty":"RSA","kid":"e0606cd0-905d-4aba-920c-6e514a1cbb26","alg":"RS256","n":"3uhDGQLrA8ZJIPAixKoxzCMzWRHb5_UWNihyccCWUicUnCOuTr5caoP4pTVdvK8SudQDQJFXBcO8i3Y5zIHTafTjYZ_ofe07_E9Pz9hFI8pvfXLfL0PWaSO5Y2XGyCgC2ibLqAEc71VYIqruyVuaYqv4gqG__gmBcR3C4gULOjYkBTNs7jMT2XaqlPE0dWDWSfLC-Zz-Czxkp52fOxUVAYb1tZcuActo6DLwnYGf5cpJMU76BhEegEkJzvJqDr9DdS996cjl2vxhPtCtS2rx7z4fCk5gCnCefztu_JjageY5Lk9IG03xzW3A2TUAXm-t_KJbRKa4-DghWEfBr8zv1w","e":"AQAB"}]}
```
reltuk, line 1: Nit: move to `testdata`, which is the normal place to put this stuff in go land I think.

go/libraries/utils/jwtauth/jwks.go
```diff
@@ -0,0 +1,100 @@
...
+	mutex   *sync.Mutex
+}
+
+func newCachedJWKS() *cachedJWKS {
+	return &cachedJWKS{value: nil, expires: time.Now(), mutex: &sync.Mutex{}}
+}
+
+type fetchedJWKS struct {
+	URL          string
+	BrokenBase64 bool
```
reltuk, line 42: I don't think this is necessary here...it's a kludge in the code on the infra side to account for a bug in a particular JWKS which is published by a particular AWS system.

go/libraries/utils/jwtauth/jwks.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+
+			jwks := jose.JSONWebKeySet{}
+			err = json.Unmarshal(contents, &jwks)
+			if err != nil {
+				return nil, err
+			}
+			f.cache.value = &jwks
+			_, _, err = cachecontrol.CachableResponse(request, response, cachecontrol.Options{})
+			//TODO: expire the cache based on cachecontrol response.
```
reltuk, line 87: Go ahead and drop this. I don't think this is really something we'll do unless we get explicit feature requests from users.

integration-tests/bats/helper/gen_keys.bash
```diff
@@ -0,0 +1,38 @@
+python3 -c '
```
reltuk, line 1: Does this need new python dependencies installed when `bats` runs? Is it better/possible to use `go/libraries/jwtauth/gen_key/main.go` instead?

reltuk (commented):
This looks great so far. Just a few comments.

PR #3898, opened 2022-07-22
/.github/workflows/ci-performance-benchmarks-new-format.yaml: remove since #benchmark also does new format

coffeegoddd:


andrew-wm-arthur (approved):
LGTM!

PR #3899, opened 2022-07-22
go/store/types: SerialMessage: Unify TupleRowStorage and SerialMessage.

reltuk:
Get rid of TupleRowStorage and use SerialMessage everywhere. In turn, change
SerialMessage so that it carries its kind and size prefix as part of itself.


reltuk:
@andrew-wm-arthur One question here would be if we should push FinishMessage(), MessagePrefixSz, MessageTypesKind, etc., into the `serial` package. Then `serial.GetFileID()` would do the MessagePrefix offset itself, for example.

andrew-wm-arthur (approved):
LGTM, I think `GetFileID()` and `FinishMessage()` should definitely live in the same place, `message` or `serial` both seem fine

PR #3905, opened 2022-07-22
Fix for persisted database specific system variables failing to load

druvv:
Fix for https://github.com/dolthub/dolt/issues/3887

Persisting and loading `${branch_name}_default_branch` from config now works.

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/session.go
+++ b/go/libraries/doltcore/sqle/dsess/session.go
@@ -1143,8 +1143,11 @@ func (d *DoltSession) SystemVariablesInConfig() ([]sql.SystemVariable, error) {
 	if d.globalsConf == nil {
 		return nil, ErrSessionNotPeristable
 	}
-
-	return SystemVariablesInConfig(d.globalsConf, func(string) {})
+	sysVars, _, err := SystemVariablesInConfig(d.globalsConf)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


max-hoffman (approved):
LGMT, two comments are 1) `NewSqlEngine` might be the place to lazily load these values, and 2) if you can log without the callback it'd be cleaner

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -1234,8 +1234,10 @@ func setPersistedValue(conf config.WritableConfig, key string, value interface{}
 	}
 }
 
-// SystemVariablesInConfig returns system variables from the persisted config
-func SystemVariablesInConfig(conf config.ReadableConfig) ([]sql.SystemVariable, error) {
+// SystemVariablesInConfig returns system variables from the persisted config.
+// It ignores persisted keys that do not exist in |sql.SystemVariables|. The
+// ignored keys are sent to |cb|.
+func SystemVariablesInConfig(conf config.ReadableConfig, cb func(key string)) ([]sql.SystemVariable, error) {
```
max-hoffman, line 1240: i think you need to pick one, a callback that takes `func(sql.SystemVariable, error)` or a signature that returns `valid, inValid []sql.SystemVariable`. A combo of the two is more difficult for a generic user even when the doc comment is good.

PR #3928, opened 2022-07-26
[no-release-notes] make `dsess.SessionCache` threadsafe

andrew-wm-arthur:


fulghum (approved):
Thanks for fixing this Andy!

PR #3939, opened 2022-07-27
[no-release-notes] Added data validation to `dolt migrate`

andrew-wm-arthur:
Adds validation stages for `dolt migrate`. Patches schemas for `dolt_docs` and `dolt_schemas` tables

### DIFF
--- a/go/libraries/doltcore/migrate/transform.go
+++ b/go/libraries/doltcore/migrate/transform.go
@@ -204,6 +204,10 @@ func migrateRoot(ctx context.Context, root *doltdb.RootValue, new *doltdb.DoltDB
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/migrate/transform.go
```diff
@@ -228,6 +232,12 @@ func migrateTable(ctx context.Context, table *doltdb.Table, new *doltdb.DoltDB)
 	if err != nil {
 		return nil, err
 	}
+	if sch, err = patchMigrateSchema(ctx, sch); err != nil {
+		return nil, err
+	}
+	if err = validateSchema(sch); err != nil {
```
max-hoffman, line 238: can `validateSchema` ever fail? the patch code looks like it plugs the gaps with the same comparison condition

max-hoffman (approved):
LGTM

PR #3941, opened 2022-07-27
Fix some env loading panics

max-hoffman:


go/libraries/doltcore/env/environment.go
```diff
@@ -99,7 +100,7 @@ type DoltEnv struct {
 }
 
 // Load loads the DoltEnv for the .dolt directory determined by resolving the specified urlStr with the specified Filesys.
-func Load(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys, urlStr, version string) *DoltEnv {
+func Load(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys, urlStr, version, binFormat string) *DoltEnv {
```
zachmu, line 103: Any time you find yourself writing a new empty default value in the vast majority of call sites, it's a good indicator you should have a different method.

In this case you want to introduce a new LoadWithFormat method, and the existing Load method will dispatch to it with the default for the new param. And it should probably be a types.Format pointer rather than a string.

zachmu (approved):
Fine with this tactically, but this just masks the underlying error. We do need to be able to load multiple DBs of different formats side by side.

PR #3947, opened 2022-07-28
[no-release-notes] `__DOLT_1__` migration

andrew-wm-arthur:


go/libraries/doltcore/migrate/transform.go
```diff
@@ -195,6 +201,13 @@ func migrateRoot(ctx context.Context, root *doltdb.RootValue, new *doltdb.DoltDB
 			return true, err
 		}
 
+		ok, err := tbl.HasConflicts(ctx)
```
bheni, line 204: It might be better to check all the tables for conflicts upfront.  If you have 10 tables and the first 9 are good and the 10 has conflicts.  You can know this isn't going to work upfront, which is better than migrating 9 and then failing on the 10th.

bheni (approved):
LGTM.  1 comment

PR #3956, opened 2022-07-28
Add command line and SQL support for `dolt pull <remote> <remoteBranch>`

fulghum:
Adds support for explicitly specifying the remote branch to merge into the current working head. 

I was tempted to work on combining these two duplicated code paths, but wanted to keep scope small and get several bugs fixed before taking that on. 

Fixes: https://github.com/dolthub/dolt/issues/3937

Documentation updates: https://github.com/dolthub/docs/pull/811

### DIFF
--- a/go/libraries/doltcore/env/remotes.go
+++ b/go/libraries/doltcore/env/remotes.go
@@ -361,6 +361,20 @@ type PullSpec struct {
 }
 
 func NewPullSpec(_ context.Context, rsr RepoStateReader, remoteName, remoteRefName string, squash, noff, force, remoteOnly bool) (*PullSpec, error) {
+	refSpecs, err := GetRefSpecs(rsr, remoteName)
+	if err != nil {
+		return nil, err
+	}
+	if len(refSpecs) == 0 {
+		return nil, ErrNoRefSpecForRemote
+	}
+
+	remotes, err := rsr.GetRemotes()
+	if err != nil {
+		return nil, err
+	}
+	remote := remotes[refSpecs[0].GetRemote()]
+
 	var remoteRef ref.DoltRef
 	if remoteRefName == "" {
 		branch := rsr.CWBHeadRef()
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/env/remotes.go
```diff
@@ -360,45 +360,50 @@ type PullSpec struct {
...
 		}
-	}
 
-	remotes, err := rsr.GetRemotes()
-	if err != nil {
-		return nil, err
+		trackedBranch, hasUpstream := trackedBranches[branch.GetPath()]
+		if !hasUpstream {
+			if remoteOnly {
+				return nil, fmt.Errorf(ErrPullWithRemoteNoUpstream.Error(), remoteName)
```
max-hoffman, line 389: i probably wrote the wrong version, but i think this way works?
```suggestion
				return nil, fmt.Errorf("%w: '%s'", ErrPullWithRemoteNoUpstream, remoteName)
```

integration-tests/bats/sql-pull.bats
```diff
@@ -459,6 +459,27 @@ teardown() {
...
+    [[ ! "$output" =~ "t1" ]] || false
+
+    # Explicitly specifying the remote and remote ref will merge in that branch
+    run dolt sql -q "call dolt_pull('origin', 'main');"
+    [ "$status" -eq 0 ]
+    run dolt sql -q "show tables"
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "t1" ]] || false
+
+    # Specifying a non-existent remote ref gives an error
```
max-hoffman, line 477: suggested additional tests off the top of my head, your discretion ofc all of this looks good already
- pull a feature branch that exists on the remote, but not locally
- pull a refspec branch of the form `refs/remotes/origin/feature`
- pull a branch that is not the active branch but exists locally
- pull a remote branch with active changes in the working set (no conflicts)
- pull branch with conflicting changes in the working set (probably same as above in Git)
- pull a branch that ff-merges with active/non-active branch
- pull a branch that can't merge with active/non-active branch

max-hoffman (approved):
lgtm! i've been looking forward to this feature for awhile. at some point we'll need to update the dolthub docs. we might have to route the CLI through the SQL engine to truly merge these paths, the best i could do before was common arg parsing

go/libraries/doltcore/env/remotes.go
fulghum: I see what you mean... this line looks a little odd... I believe it's still correct though, because the error string already contains the `%s` to plug the remote name into it. It would be nice to switch over to `go-errors.v1` and then this could be cleaned up a little more, but I think that's a bigger change (to move all the errors in this file over to go-errors.v1) than is needed right now.

integration-tests/bats/sql-pull.bats
fulghum: Nice test case brainstorming! Some of these won't work yet, e.g. non-branch name parameters, using pull to merge a local branch (doesn't look like we support the '.' syntax yet that Git uses for this mode). I like the suggestion to test with different merge conditions. I'll poke around with that and add a little more coverage.

PR #3958, opened 2022-07-28
[no-release-notes] Store first and last offset in `ItemArray`

andrew-wm-arthur:
store 4 extra bytes per chunk to maybe make us less branchy accessing data


max-hoffman (approved):
lgtm!

PR #3960, opened 2022-07-28
When modifying a column, don't seed the new column tag with the old NomsKind

druvv:
Columns tags are generated in a psuedo-random manner. They are seeded with the NomsKinds of all the existing columns and the name of the column for which the tag is being generated. 

When modifying a column, we previously included the old column's NomsKind when seeding the new tag. This produced bad round-tripping behavior explained here:
https://github.com/dolthub/dolt/issues/3950#issuecomment-1198548171

This PR excludes the old column's NomsKind from the seed.


integration-tests/bats/column_tags.bats
```diff
@@ -236,3 +236,63 @@ DELIM
...
+@test "column_tags: Round-tripping a column type through different NomsKinds restores original tag" {
+    dolt sql -q "CREATE TABLE t (pk INT PRIMARY KEY, col1 int);"
+    run dolt schema tags
+    [ $status -eq 0 ]
+    [[ $output =~ "col1   | 10878" ]] || false
+
+    dolt sql -q "ALTER TABLE t MODIFY COLUMN col1 VARCHAR(100);"
+    run dolt schema tags
+    [ $status -eq 0 ]
+    echo $output
```
VinaiRachakonda, line 249: Prune these echos

integration-tests/bats/column_tags.bats
```diff
@@ -236,3 +236,63 @@ DELIM
...
+    [[ $output =~ "col1   | 16050" ]] || false
+
+     dolt sql -q "ALTER TABLE t MODIFY COLUMN col1 VARCHAR(100);"
+    run dolt schema tags
+    [ $status -eq 0 ]
+    echo $output
+    [[ $output =~ "col1   | 16050" ]] || false
+}
+
+# Tags aren't magic...
```
VinaiRachakonda, line 279: I don't get this comment

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1372,6 +1371,92 @@ func (t *AlterableDoltTable) RewriteInserter(
...
+		}
+		return false, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return newRoot, nil
+}
+
+func filterColumnFromSch(sch schema.Schema, colTag uint64) (schema.Schema, error) {
```
VinaiRachakonda, line 1443: Is there no remove by tag method for a column collection?

VinaiRachakonda (approved):
looks good

go/libraries/doltcore/sqle/tables.go
druvv: No, there isn't

PR #3964, opened 2022-07-29
improve error message for merge when two columns have the same tag

druvv:
See https://github.com/dolthub/dolt/issues/3963


VinaiRachakonda (approved):
lgtm

PR #3982, opened 2022-08-01
doltcore/sqle/index: Restructed concurrency for indirect secondary index (new format)

andrew-wm-arthur:


andrew-wm-arthur:
#benchmark

andrew-wm-arthur:
This change doesn't have much perf impact after the changes we made to remove BLOBs from the benchmarks. The last commit here removed concurrency entirely from indirect index lookups. This has marginal impact and allows us to avoid allocating a channel, but it's not much juice for the squeeze right now. Might revisit later

PR #3986, opened 2022-08-02
Lockfile should be ignored if encoded pid is different than the current process

VinaiRachakonda:


go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,71 @@ func (dEnv *DoltEnv) LockFile() string {
...
+		return false
+	}
+
+	if string(out) == "" {
+		return true // pid exist
+	}
+	return false
+}
+
+func (dEnv *DoltEnv) getProcessFromLockFile() (int, error) {
```
VinaiRachakonda, line 1299: Probably does not need to be a struct method

integration-tests/bats/sql-server.bats
```diff
@@ -1504,3 +1504,54 @@ behavior:
...
+
+    start_sql_server
+    server_query repo1 1 "SELECT 1" "1\n1"
+    stop_sql_server
+
+    # Add malicious text to lockfile and expect to fail
+    echo "iamamaliciousactor" > repo1/.dolt/sql-server.lock
+
+    run start_sql_server
+    [ "$status" -eq 1 ]
```
VinaiRachakonda, line 1556: Should validate the right output here

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,71 @@ func (dEnv *DoltEnv) LockFile() string {
 	return f
 }
 
-// IsLocked returns true if this database's lockfile exists
+// IsLocked returns true if this database's lockfile exists and that process written in lockfile is still active.
```
VinaiRachakonda, line 1257: grammar

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,71 @@ func (dEnv *DoltEnv) LockFile() string {
...
+}
+
+func (dEnv *DoltEnv) getProcessFromLockFile() (int, error) {
+	// validate that the pid on the lock file is still active
+	lockFile, err := dEnv.FS.OpenForRead(dEnv.LockFile())
+	if err != nil {
+		return -1, err
+	}
+
+	b := make([]byte, 100000)
```
VinaiRachakonda, line 1306: probably too much lol

integration-tests/bats/sql-server.bats
```diff
@@ -1504,3 +1504,55 @@ behavior:
...
+    # Try adding fake pid numbers. Could happen via debugger or something
+    echo "423423" > repo1/.dolt/sql-server.lock
+    echo "4123423" > repo2/.dolt/sql-server.lock
+
+    start_sql_server
+    server_query repo1 1 "SELECT 1" "1\n1"
+    stop_sql_server
+
+    # Add malicious text to lockfile and expect to fail
+    echo "iamamaliciousactor" > repo1/.dolt/sql-server.lock
```
max-hoffman, line 1553: is failing the desired behavior here? or would we want to ignore the noise and continue along?

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,71 @@ func (dEnv *DoltEnv) LockFile() string {
...
+}
+
+func getProcessFromLockFile(fs filesys.Filesys, lockFile string) (int, error) {
+	// validate that the pid on the lock file is still active
+	rd, err := fs.OpenForRead(lockFile)
+	if err != nil {
+		return -1, err
+	}
+
+	b := make([]byte, 1000)
```
max-hoffman, line 1306: is there a maximum process id const you can tie this magic number to?

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,71 @@ func (dEnv *DoltEnv) LockFile() string {
...
+		return false
+	}
+
+	return checkPid(p.Pid)
 }
 
-// Lock writes this database's lockfile or errors if it already exists
+// cc. https://stackoverflow.com/questions/15204162/check-if-a-process-exists-in-go-way
+func checkPid(pid int) bool {
+	out, err := exec.Command("kill", "-s", "0", strconv.Itoa(pid)).CombinedOutput()
```
max-hoffman, line 1288: suspicious of whether this will work with windows

max-hoffman (approved):
LGTM, thanks for adding this feature!

integration-tests/bats/sql-server.bats
VinaiRachakonda: My sense was to lean towards strictness.

go/libraries/doltcore/env/environment.go
VinaiRachakonda: I added some notes

zachmu: It for sure will not work

zachmu: You should probably just not do this check on windows

zachmu (commented):
Method for checking the running PID is pretty fragile, you should restrict it so it doesn't run on windows

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,75 @@ func (dEnv *DoltEnv) LockFile() string {
...
+	lockFilePid, err := getProcessFromLockFile(dEnv.FS, dEnv.LockFile())
+	if err != nil { // if there's any error assume that env is locked since the file exists
+		return true
+	}
+
+	// Check whether the pid that spawned the lock file is still running. Ignore it if not.
+	// For unix systems FindProcess will always return a nil err and a process object which is why we must use the checkPid
+	// function. For Windows systems FindProcess will correctly determine if a process exists or not.
+	p, err := os.FindProcess(lockFilePid)
+	if runtime.GOOS == "windows" {
```
zachmu, line 1277: Instead of this, you should have different versions of this function in different files that only run on one particular platform

call it `processExists`

You might also check out if there's a lightweight library that does this

go/libraries/doltcore/env/environment.go
```diff
@@ -1251,17 +1254,75 @@ func (dEnv *DoltEnv) LockFile() string {
...
+
+func getProcessFromLockFile(fs filesys.Filesys, lockFile string) (int, error) {
+	// validate that the pid on the lock file is still active
+	rd, err := fs.OpenForRead(lockFile)
+	if err != nil {
+		return -1, err
+	}
+
+	// Technically, the max pid is bounded by int types (~32 bits). That gets
+	// encoded to about 11 bytes. We'll round about just in case.
```
VinaiRachakonda, line 1309: Uhhhh double check this math lol

go/libraries/doltcore/env/environment.go
VinaiRachakonda: I used this library https://github.com/mitchellh/go-ps. Gotta be good since it was made by the Hashicopr founder. Also tested in on windows

go/libraries/doltcore/env/environment.go
VinaiRachakonda: Validate that 2^32 gets interpreted as 11 bytes (i.e n = 11)

PR #3996, opened 2022-08-02
Fix `dolt_remote` stored procedure to write to the correct database directory

### LINKED ISSUE #3936: Can't add a remote from a sql-server started outside of a dolt directory
Shell 1:
```
dolt $ dolt sql-server
Starting server with Config HP="localhost:3306"|T="28800000"|R="false"|L="info"
2022-07-26T16:21:08-07:00 INFO [conn 1] NewConnection {DisableClientMultiStatements=false}
2022-07-26T16:21:57-07:00 WARN [conn 1] error running query {connectTime=2022-07-26T16:21:08-07:00, connectionDb=pnp_bug, error=syntax error at position 19 near 'table', query=create table table (pk int, primary key(pk));}
2022-07-26T16:22:24-07:00 WARN [conn 1] error running query {connectTime=2022-07-26T16:21:08-07:00, connectionDb=pnp_bug, error=syntax error at position 20 near 'table', query=call dolt_add(table);}
2022-07-26T16:22:28-07:00 WARN [conn 1] error running query {connectTime=2022-07-26T16:21:08-07:00, connectionDb=pnp_bug, error=column "table1" could not be found in any table in scope, query=call dolt_add(table1);}
2022-07-26T16:27:47-07:00 WARN [conn 1] error running query {connectTime=2022-07-26T16:21:08-07:00, connectionDb=pnp_bug, error=open /Users/timsehn/liquidata/dolt/.dolt/repo_state.json: no such file or directory, query=call dolt_remote('add', 'origin', 'timsehn/pnp_bug');}
```

Shell 2:

```
git $ dolt sql-client
# Welcome to the Dolt MySQL client.
# "exit" or "quit" (or Ctrl-D) to exit.
mysql> create database pnp_bug;
mysql> use pnp_bug;
...
...
...
mysql> call dolt_remote('add', 'origin', 'timsehn/pnp_bug');
Error 1105: open /Users/timsehn/liquidata/dolt/.dolt/repo_state.json: no such file or directory
```

This looks like because it is looking at $CWD for a .dolt direectory but the use is like `cd`ing into the database directory.

fulghum:
Changes DoltDatabaseProvider to track a Filesystem location for each database it manages. This allows callers, like the `dolt_remote` stored procedure to find the correct root directory for a database and update repo state correctly. 

Fixes #3936

### DIFF
--- a/go/cmd/dolt/commands/engine/sqlengine.go
+++ b/go/cmd/dolt/commands/engine/sqlengine.go
@@ -95,7 +95,7 @@ func NewSqlEngine(
 	if err != nil {
 		return nil, err
 	}
-	pro.WithRemoteDialer(mrEnv.RemoteDialProvider())
+	pro = pro.WithRemoteDialer(mrEnv.RemoteDialProvider())
 
 	// Load in privileges from file, if it exists
 	persister := mysql_file_handler.NewPersister(config.PrivFilePath, config.DoltCfgDirPath)
--- a/go/cmd/dolt/commands/engine/sqlengine.go
+++ b/go/cmd/dolt/commands/engine/sqlengine.go
@@ -75,7 +75,7 @@ func NewSqlEngine(
 
 	parallelism := runtime.GOMAXPROCS(0)
 
-	dbs, err := CollectDBs(ctx, mrEnv, config.Bulk)
+	dbs, locations, err := CollectDBs(ctx, mrEnv, config.Bulk)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


max-hoffman (approved):
LGTM this all looks above board. The surface area that multidb broke makes me worry that this is going to continue for everything dEnv related, like tempfile folders and transaction related things. Should some of these properties be embedded in the database itself? Should we make an interface above DoltDatabaseProvider so that DBP doesn't continue to be thing we keep pushing features into?

fulghum:
Thanks for the quick review Max! Yeah... there are definitely some recurring problems (aka "greatest hits") that I encountered in this one... multi-repo environments still have sharp edges and our test setup has significant differences between production. For example, I got this fix working from bats/command-line, but had to debug through a problem in our test harness where multiple providers were being created and the session wasn't updated to use the latest one. Last but not least, DoltDatabaseProvider does tend to be a junk drawer for us as well. 

In this case, I briefly considered if sql.Database should know it's location, but decided against it and went with DoltDatabaseProvider. It felt like a smaller change and more appropriate for a provider to know the filesystem location than the Database itself.

I think all those issues are definitely worth a synchronous discussion sometime to help us figure out any longer term direction we want to move towards.

PR #3998, opened 2022-08-02
Root value cache key

max-hoffman:
from `index_join_scan`:
```
Latency (ms):
         min:                                    3.80
         avg:                                    4.35
         max:                                   10.85
         95th percentile:                        4.91
         sum:                                10000.42
```

to `index_join_scan`:
```
Latency (ms):
         min:                                    3.54
         avg:                                    3.92
         max:                                    8.84
         95th percentile:                        5.18
         sum:                                 9999.99
```


zachmu:
This breaks table session caching (the root values aren't equal). It's faster on data-heavy benchmarks like scan, but significantly worse on point lookups.

PR #4001, opened 2022-08-03
Fixing `dolt_clone` to clean up partially cloned database directories on failure 

fulghum:


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -227,13 +227,27 @@ func (p DoltDatabaseProvider) CloneDatabaseFromRemote(ctx *sql.Context, dbName,
...
 
-	var r env.Remote
-	var srcDB *doltdb.DoltDB
-	dialer := p.remoteDialer
-	if dialer == nil {
+	err := p.cloneDatabaseFromRemote(ctx, dbName, remoteName, branch, remoteUrl, remoteParams)
+	if err != nil {
+		// Make a best effort to clean up any artifacts on disk from a failed clone
+		// before we return the error
+		p.fs.Delete(dbName, true)
```
max-hoffman, line 234: maybe log if the delete errors?

max-hoffman (approved):
lgtm!

go/libraries/doltcore/sqle/database_provider.go
fulghum: Good suggestion. I captured the error from delete and if it's not nil, the code appends an extra note on the returned error about not being able to clean up the failed clone directory.

PR #4009, opened 2022-08-03
Removed skipped bats for modifying dolt_remotes system table. 

timsehn:
Went with dolt_remote() stored procedure instead.


max-hoffman:
These don't look like they were skipped, did you mean to delete these?

timsehn:
Good catch. Wrong tests :-)

timsehn:
Looks like Jennifer already killed these tests closing PR and deleting branch.

PR #4018, opened 2022-08-04
Bug fix to configure remote dialer when no databases exist yet

fulghum:
When creating a DoltDatabaseProvider, the GRPCDialProvider was being pulled from the first DoltEnv found in the mutli-repo environment. When no DoltEnvs existed yet, the GRPCDialProvider was being set to nil, which caused `dolt_clone` to fail (see https://github.com/dolthub/hosted-issues/issues/31). 

This change pulls out the GPRCDialProvider code from DoltEnv into a separate type that can work (without supporting custom user credentials) when there is no existing DoltEnv yet. 

Fixes: https://github.com/dolthub/hosted-issues/issues/31


zachmu (approved):
LGTM!

PR #4028, opened 2022-08-05
[no-release-notes] Adding Dolt engine tests to cover using views with as of expressions

fulghum:
I didn't see any Dolt engine tests using views with as of expressions, so wanted to add some coverage. 

Depends on: https://github.com/dolthub/go-mysql-server/pull/1159


zachmu (approved):
LGTM!

PR #4033, opened 2022-08-06
Implement --fun flag for dolt init

### LINKED ISSUE #4015: Implement `dolt init --fun`
The dolt logo is inspired by a commit hash `d01t`. 

It would be cool if we had an option to have a dolt initial commit be rerolled until the first four characters matched the following pattern:

1. d, o l, t in order 
2. any uppercase or lowercase d, o, l, t
3. 0 or 1 can replace l or t

I think this should be a hidden option not in the docs.

abmyii:
Closes #4015.

Hopefully not the most messy implementation. I'm not sure modifying the time was the best solution, but it seemed to work OK.

In all my tests it took ~3-5 mins to find a matching commit hash.

### DIFF
--- a/go/cmd/dolt/commands/init.go
+++ b/go/cmd/dolt/commands/init.go
@@ -32,6 +32,7 @@ const (
 	emailParamName      = "email"
 	usernameParamName   = "name"
 	initBranchParamName = "initial-branch"
+	fun                 = "fun"
 )
 
 var initDocs = cli.CommandDocumentationContent{
@@ -75,6 +76,7 @@ func (cmd InitCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(emailParamName, "", "email", fmt.Sprintf("The email address used. If not provided will be taken from {{.EmphasisLeft}}%s{{.EmphasisRight}} in the global config.", env.UserEmailKey))
 	ap.SupportsString(cli.DateParam, "", "date", "Specify the date used in the initial commit. If not specified the current system time is used.")
 	ap.SupportsString(initBranchParamName, "b", "branch", fmt.Sprintf("The branch name used to initialize this database. If not provided will be taken from {{.EmphasisLeft}}%s{{.EmphasisRight}} in the global config. If unset, the default initialized branch will be named '%s'.", env.InitBranchName, env.DefaultInitBranch))
+	ap.SupportsFlag(fun, "", "Fun!")
 	return ap
 }
 
--- a/go/cmd/dolt/commands/init.go
+++ b/go/cmd/dolt/commands/init.go
@@ -134,7 +134,7 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		}
 	}
 
-	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t)
+	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t, apr.Contains(fun))
 	if err != nil {
 		cli.PrintErrln(color.RedString("Failed to initialize directory as a data repo. %s", err.Error()))
 		return 1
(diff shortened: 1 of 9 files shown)


timsehn:
This is so cool! Thank you.

go/cmd/dolt/commands/init.go
```diff
@@ -132,7 +134,7 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		}
 	}
 
-	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t)
+	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t, apr.Contains(fun))
```
zachmu, line 137: Rather than introduce a new parameter to this method, you should introduce a new method that does this specific thing

zachmu (commented):
This is really cool, thanks :)

Here's my suggestion: rather than plumbing this new option all the way down into the deepest database library code, keep it at the application layer in the commands package. Just blow away and re-create create the new repo storage over and over until it matches the condition you want.

abmyii:
> Just blow away and re-create create the new repo storage over and over until it matches the condition you want.

Do you mean delete the repo and recreate it until the commit matches?

go/cmd/dolt/commands/init.go
abmyii: Ah, that makes sense! Wouldn't this be unnecessary if it's all done at the application/command layer though?

zachmu:
> > Just blow away and re-create create the new repo storage over and over until it matches the condition you want.
> 
> Do you mean delete the repo and recreate it until the commit matches?

Yup, that's right.

go/cmd/dolt/commands/init.go
zachmu: Yeah, I gave this feedback before reconsidering the overall approach.

abmyii:
Hmm, so I tried that before too (I had a similar idea but in the InitRepoWithTime function itself) but `dEnv.bestEffortDeleteAll` didn't seem to do anything, and it doesn't work now either. I'll try to dig around a bit and see if I can figure out why.

abmyii:
@zachmu Is there a way to clear the DoltCliConfig or reset dEnv? I'm running into https://github.com/dolthub/dolt/blob/20e4f9b015105a4827e0264a9fd844ec17dfabec/go/libraries/utils/config/config_hierarchy.go#L52 when trying to reinitialize the repo.

abmyii:
> Is there a way to clear the DoltCliConfig or reset dEnv? I'm running into

So I fixed this using `dEnv.Config, _ = env.LoadDoltCliConfig(env.GetCurrentUserHomeDir, dEnv.FS)` (probably a terrible hack, but worked). However, I left it running for over two hours and it didn't generate a matching commit in that time (vs 3/5 mins with the original version). Any ideas why?

```go
        commit, _ := dEnv.HeadCommit(context.Background())
		hash, _ := commit.HashOf()

		reg := regexp.MustCompile("^d[o0][1l]t")
		for !reg.MatchString(hash.String()) {
			dEnv.FS.Delete(dEnv.GetDoltDir(), true)

			dEnv.Config, _ = env.LoadDoltCliConfig(env.GetCurrentUserHomeDir, dEnv.FS)
			dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, time.Now())

			commit, _ = dEnv.HeadCommit(context.Background())
			hash, _ = commit.HashOf()
			println(hash.String())
		}
```
(My entire code, below line 135 - https://github.com/dolthub/dolt/blob/a8dacbf42fc31b0ed1cecada8df26392255a11e5/go/cmd/dolt/commands/init.go#L135)

abmyii:
@zachmu https://github.com/abmyii/dolt/tree/init-fun-new - cleaner implementation, but as before it often just runs in an infinite loop. I had one instance where it completed in around a minute but apart from that it just keeps running. It's also quite volatile - running `ls .dolt` whilst its running causes the `Delete` to fail and the program to stop.

abmyii:
> It's also quite volatile - running ls .dolt whilst its running causes the Delete to fail and the program to stop.

I just realised this is due to me having `dolt branch --show-current 2> /dev/null | sed -e 's/\(.*\)/(\1) /'` in my bashrc - it seems to make a tmp folder every time that is run, which is a bit annoying as now I've probably cluttered quite a few folders with .tmp dirs. Are they necessary?

timsehn:
I'm going to close this as we were never able to drive to resolution.

PR #4036, opened 2022-08-08
Fix indeterministic diff delta matching

druvv:
Before this fix, tables with different names would attempted to be matched even if there was a pair of tables with a matching name.


go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -144,6 +144,7 @@ func getFkParentSchs(ctx context.Context, root *doltdb.RootValue, fks ...doltdb.
 }
 
 func matchTableDeltas(fromDeltas, toDeltas []TableDelta) (deltas []TableDelta, err error) {
+	var intersectedNames []string
```
zachmu, line 147: matchedNames?

zachmu (commented):
Test?

PR #4050, opened 2022-08-10
Add amend option for commit

### LINKED ISSUE #2338: `dolt commit --amend` to amend commit messages
Being able to amend the previous commit's message would be a very useful feature to have. I will usually use `git commit --amend -m "<new message>"` rather than amending in an editor as I find that much faster, and I think that's sufficient for this issue.

tanyongzhi:
This PR adds the ability to amend the commit at `HEAD`. It accomplishes this by first soft resetting to `HEAD~1` before applying the new commit. Empty new amended commits are allowed to support the original use case of changing the commit messages of previous commits. 

Note that this PR does **not** add the amend feature to the SQL stored procedures.

Fixes #2338


go/cmd/dolt/commands/commit.go
```diff
@@ -150,7 +157,7 @@ func (cmd CommitCmd) Exec(ctx context.Context, commandStr string, args []string,
 	pendingCommit, err := actions.GetCommitStaged(ctx, roots, ws.MergeActive(), mergeParentCommits, dEnv.DbData(), actions.CommitStagedProps{
 		Message:    msg,
 		Date:       t,
-		AllowEmpty: apr.Contains(cli.AllowEmptyFlag),
+		AllowEmpty: apr.Contains(cli.AllowEmptyFlag) || apr.Contains(cli.AmendFlag),
```
fulghum, line 160: What was the reason this was needed? To account for amending a commit that was already an empty commit?

fulghum (commented):
This is looking good @tanyongzhi. Thank you for taking the time to send us this PR! 🙏 

Would you be able to write a of BATS tests function for this functionality? There's not a perfect existing BATS test file for this, but [checkout.bats](https://github.com/dolthub/dolt/blob/main/integration-tests/bats/checkout.bats) is probably a good enough place to start off a new test for `dolt commit --amend` for now. 

I'll also check out your branch later today and test out the experience on the command line and see if I notice any other feedback.

go/cmd/dolt/commands/commit.go
tanyongzhi: This is to account for commit amends where only the commit message is changed. For these cases, the new "commit" after the soft reset will not have any staged tables. This change is hence to prevent `GetCommitStaged` from returning an error.

tanyongzhi:
I added a couple of tests in `checkout.bats`. Please let me know if there is a more appropriate test to add these tests in!

integration-tests/bats/checkout.bats
```diff
@@ -178,3 +178,42 @@ SQL
...
+  dolt add .
+  dolt commit --amend -m "modified_commit_message"
+
+  commitmsg=$(dolt log --oneline | head -n 1)
+  [[ $commitmsg =~ "modified_commit_message" ]]
+
+  numcommits=$(dolt log --oneline | wc -l)
+  [[ $numcommits =~ "2" ]]
+
+  run dolt sql -q 'select * from test;'
```
fulghum, line 216: A little BATS trivia... 

using `run` won't error out if the command fails, so it's good to assert the status after using run, like this:
```
    [ "$status" -eq 0 ]
```

For the output matching, add a ``` || false ``` at the end of each output matching line, to make sure bash will error out on the right line. This one has bitten us before where not including that makes the failing output harder to debug. ex: 
```
    [[ "$output" =~ "8" ]] || false
```

fulghum (commented):
Thanks for adding the BATS test! I took this for a spin this morning and overall, this is looking very good. I noticed a couple of things we should fix up:
* (git consistency) we should pre-populate the editor with the previous commit message to match Git’s behavior. 
* (git consistency) when amending a merge commit, Git preserves the metadata about the two merge parents when it creates the new, amended commit. In my quick testing, it seems like the merge parent metadata was lost after amending a merge commit. This is a good test case to add to the BATS tests. 
* (bug) I exited the editor (without saving a commit message) to abort the command, but it seemed like the repo was still reset to HEAD~1, so my log was missing the most recent commit and I had changes staged. We should match Git’s behavior here where if you exit the editor without entering a commit message, your repo history doesn't get modified. This would be an awesome test case to cover in the tests, but I’m not sure exactly how we’d trigger that case from the BATS scripts. Bonus points for sure if you can come up with something clever. 

If you wanted to create a new `commit.bats` file for these new commit --amend BATS tests, that would be awesome, but I'm happy to refactor that later to tidy up and let you just focus on the main logic for this change. 

Thanks again for the time and effort you're putting into this one! It's very appreciated and the change is looking very close to done once we wrap up the details above.

integration-tests/bats/checkout.bats
tanyongzhi: I see, thanks for clearing this up!

tanyongzhi:
Thanks for the feedback! I added some changes addressing the pointers mentioned above, including an integration test that tests if the meta data of the parents of merge commits are preserved.

integration-tests/bats/checkout.bats
```diff
@@ -178,3 +178,76 @@ SQL
   cmd=$(echo "${lines[1]}" | cut -d ' ' -f 1,2,3)
   [[ $cmd =~ "dolt checkout $sha" ]]
 }
+
+@test "commit --amend: only commit message is changed" {
```
fulghum, line 182:
```suggestion
@test "checkout: commit --amend only changes commit message" {
```

integration-tests/bats/checkout.bats
```diff
@@ -178,3 +178,76 @@ SQL
...
+
+  numcommits=$(dolt log --oneline | wc -l)
+  [[ $numcommits =~ "2" ]] || false
+
+  run dolt sql -q 'select * from test;'
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "8" ]] || false
+}
+
+@test "commit --amend adds new changes to existing commit" {
```
fulghum, line 201:
```suggestion
@test "checkout: commit --amend adds new changes to existing commit" {
```

integration-tests/bats/checkout.bats
```diff
@@ -178,3 +178,76 @@ SQL
...
+  numcommits=$(dolt log --oneline | wc -l)
+  [[ $numcommits =~ "2" ]] || false
+
+  run dolt sql -q 'select * from test;'
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "8" ]] || false
+  [[ "$output" =~ "9" ]] || false
+}
+
+@test "commit --amend on merge commits does not modify metadata of merged parents" {
```
fulghum, line 223:
```suggestion
@test "checkout: commit --amend on merge commits does not modify metadata of merged parents" {
```

fulghum (commented):
Sorry for the slow response to the latest changes – I must have missed the GitHub notification that there were new commits on this PR! Thanks for the ping on Discord this morning.

Changes look awesome @tanyongzhi! Great work jumping into the Dolt codebase and contributing this. Amending a commit is a **really** nice feature for Dolt to have. 🎉 

The code looks great to me; I built it locally to try it out and it works great. I did find one very small issue where you can't amend the very first initial commit, but we can fill that small gap in later in a separate PR so we can get this out faster. 

I kicked off CI tests and fixed up a couple formatting changes to keep some of the CI checks happy. I'll keep an eye on the tests and when they all complete, I'll approve and get this change merged in to main!

tanyongzhi:
Thank you @fulghum!

Unfortunately, the current method of soft resetting to `HEAD~1` does not work for the initial commit, simply because `HEAD~1` does not exist. I've created an issue here #4160.

PR #4052, opened 2022-08-10
Remove background async index queue

max-hoffman:
Remove the global async iter queue that is not used by new format. This has little effect on a long-running set of tests or a server that runs multiple tests, but muddies the waters for individual benchmarking.


reltuk:
It's very surprising to me that this wouldn't regress perf on large indexed joins, at least on LD_1. Can we perf test this against some turbine queries?

max-hoffman:
It was written specifically to boost perf for old format index joins, so I'm not sure now is the time to merge this. It is pure drag on the new format though.

max-hoffman:
#benchmark

timsehn:
Gonna close this.

PR #4061, opened 2022-08-11
Index Builder more efficient for secondary lookups

max-hoffman:


max-hoffman:
#benchmark

go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -379,7 +379,8 @@ type doltIndex struct {
 	ns     tree.NodeStore
 	keyBld *val.TupleBuilder
 
-	cache cachedDurableIndexes
+	cache  cachedDurableIndexes
+	ranges []prolly.Range
```
zachmu, line 383: This seems kind of fraught, it means we can't cache these indexes anywhere or use them in parallel.

We need some separation between the index definition (this object) and its use in lookups (the sql.IndexLookup implementation). Not appropriate to mix these two different use cases together.

If you want to do something like this, you need to have GetIndexes return disposable objects based on this object, so that the index definition isn't mutable.

zachmu (commented):
Think this is the wrong approach, see comments

PR #4068, opened 2022-08-11
Add storage format information to `dolt version` and add new sql function `dolt_storage_format()`. Add dolt init --new-format

druvv:
- Adds `--new-format` to `dolt init`. For example:
```
$ dolt init --new-format
```
- Adds storage format information to `dolt version`. For example:
```
$ dolt version
dolt version 0.40.24
database storage format: OLD ( __LD_1__ )
```

We considered making a separate command called `dolt storage_format`, but decided against it to promote discoverability of the current storage format that dolt is using.

- Adds a new sql function that returns the format used by the currently selected database. For example:
```
> SELECT dolt_storage_format();
+-----------------------+
| dolt_storage_format() |
+-----------------------+
| NEW ( __DOLT__ )      |
+-----------------------+
```
- `dolt_version()` is unchanged:
```
> SELECT dolt_version();
+----------------+
| dolt_version() |
+----------------+
| 0.40.24        |
+----------------+
```

In the SQL context, we decided to display the storage format in a new function. The reasoning is partly due to the fact that MySQL functions do not support multiple return columns and partly to make parsing either the dolt version or storage format by clients easy.

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -98,6 +98,7 @@ const (
 	RemoteParam      = "remote"
 	BranchParam      = "branch"
 	TrackFlag        = "track"
+	NewFormatFlag    = "new-format"
 )
 
 const (
--- a/go/libraries/doltcore/sqle/dfunctions/storage_format.go
+++ b/go/libraries/doltcore/sqle/dfunctions/storage_format.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Dolthub, Inc.
+// Copyright 2022 Dolthub, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/go/libraries/doltcore/sqle/dfunctions/storage_format.go
+++ b/go/libraries/doltcore/sqle/dfunctions/storage_format.go
@@ -18,9 +18,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/dolthub/go-mysql-server/sql"
+
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
 	"github.com/dolthub/dolt/go/store/types"
-	"github.com/dolthub/go-mysql-server/sql"
 )
 
 const StorageFormatFuncName = "dolt_storage_format"
--- a/go/libraries/doltcore/sqle/dfunctions/storage_format.go
+++ b/go/libraries/doltcore/sqle/dfunctions/storage_format.go
@@ -1,0 +1,91 @@
+// Copyright 2020 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dfunctions
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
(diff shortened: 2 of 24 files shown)


go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -98,6 +98,7 @@ const (
 	RemoteParam      = "remote"
 	BranchParam      = "branch"
 	TrackFlag        = "track"
+	NewFormatFlag    = "new-format"
```
fulghum, line 101: (minor) I'm wondering if there's a longer lasting name than "new-format" (and also if it matters). I'm just imagining a year or two out when we may still have this flag around and the format isn't new anymore. I know we'll have made it the default format by then, but just wondering if we should consider something like:

```
dolt init --dolt-format
```

This might be moot if you're already planning on removing that flag after we switch it to the default, but it still might be nice to have a more explicit name. 

Once we flip the default, will there be another option for customers to initialize with the old format? If so... maybe it makes more sense to have something like `--format=ld1` and `--format=dolt`? 

Not a huge deal either way, but worth thinking about the ergonomics and how it'll play out over time.

go/libraries/doltcore/sqle/dfunctions/storage_format.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2020 Dolthub, Inc.
```
fulghum, line 1:
```suggestion
// Copyright 2022 Dolthub, Inc.
```

go/libraries/doltcore/sqle/dfunctions/storage_format.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, errors.New("failed to load underlying dolt db")
+	}
+	f := GetStorageFormatDisplayString(ddb.Format())
+
+	return f, nil
+}
+
+func GetStorageFormatDisplayString(format *types.NomsBinFormat) string {
+	if types.IsFormat_DOLT(format) {
+		return fmt.Sprintf("NEW ( %s )", format.VersionString())
```
fulghum, line 60: `NEW` and `OLD` looked kinda odd in the output. Is the intent to make it really, really easy for any customer to know if they are using the new format or the old format? 

I think just returning `format.VersionString()` is probably what most users would expect. Could we do that and provide really clear docs/instructions on what the formats are?

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -1311,6 +1311,26 @@ func TestDoltVerifyConstraints(t *testing.T) {
...
+func TestDoltStorageFormat(t *testing.T) {
+	var expectedFormatString string
+	if types.IsFormat_DOLT(types.Format_Default) {
+		expectedFormatString = "NEW ( __DOLT__ )"
+	} else {
+		expectedFormatString = "OLD ( __LD_1__ )"
+	}
+	script := queries.ScriptTest{
+		Name:        "dolt storage format function works",
+		SetUpScript: []string{},
```
fulghum, line 1323: (super minor) You could completely leave this line out if you wanna save a line of code 😉

integration-tests/bats/init.bats
```diff
@@ -152,6 +152,55 @@ teardown() {
...
+
+    run dolt version
+    [ "$status" -eq 0 ]
+    [[ $output =~ "database storage format: OLD ( __LD_1__ )" ]] || false
+
+    run dolt sql -q "SELECT dolt_storage_format();"
+    [[ $output =~ "OLD ( __LD_1__ )" ]] || false
+}
+
+@test "init: empty database folder displays no version" {
```
fulghum, line 196: It's worth thinking about any testing for multi-repo environments at both the command line and through SQL, especially since we so frequently hit problems with multi-db envs. I know we have some checks in place now to ignore different formats, but I'm guessing there are still some sharp edges in there and some more testing might help root some of those out.

fulghum (commented):
Code looks good. 

I had some thoughts on tweaking the naming slightly, but I don't have a super strong opinion. `new`/`old` seem like they make a lot of sense right now, but may not age as well as explicitly specifying the format names. I understand there's probably a desire to make this super simple for customers though. 

My other thought was whether there's more testing around multi-db envs we should do. Seems prudent to write some tests for that, given how often we hit multi-db env issues, even though I know we have some safeguards in place already .

fulghum (approved):
Nice tests! Looks great. 👏

PR #4071, opened 2022-08-11
global auto increment tracking

zachmu:


go/libraries/doltcore/sqle/globalstate/auto_increment_tracker.go
```diff
@@ -110,20 +106,90 @@ func (a AutoIncrementTracker) Next(tbl string, insertVal interface{}) (uint64, e
...
+	if _, ok := a.sequences[tableName]; !ok {
+		a.sequences[tableName] = uint64(1)
+	}
 }
 
-func (a AutoIncrementTracker) DropTable(tableName string) {
+// DropTable drops the table with the name given.
+// To establish the new auto increment value, callers must also pass all other working sets in scope that may include
+// a table with the same name, omitting the working set that just deleted the table named.
+func (a AutoIncrementTracker) DropTable(ctx context.Context, tableName string, wses ...*doltdb.WorkingSet) error {
```
andrew-wm-arthur, line 159: this seems quite racy. state for concurrent transactions may be held only in this tracker. Consider transactions that  are concurrently doing a `SET AUTO_INCREMENT = x` and a `TRUNCATE`. 
Are we sure we need to update the tracker at all in the case of a `TRUNCATE` or drop table? MySQL compliance would dictate that we do for `TRUNCATE`, but we're already fudging that with the cross-branch sequence communication.

andrew-wm-arthur (commented):
One comment, generally looks good

go/libraries/doltcore/sqle/globalstate/global_state.go
```diff
@@ -18,42 +18,56 @@ import (
...
+		if err != nil {
+			return GlobalState{}, err
+		}
 
-func (g GlobalState) GetAutoIncrementTracker(ctx context.Context, ws *doltdb.WorkingSet) (AutoIncrementTracker, error) {
-	g.mu.Lock()
-	defer g.mu.Unlock()
+		ws, err := db.ResolveWorkingSet(ctx, wsRef)
+		if err == doltdb.ErrWorkingSetNotFound {
+			// skip, continue working on other branches
```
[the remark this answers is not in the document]
zachmu: Not necessarily, it could be from an older database format

go/libraries/doltcore/sqle/globalstate/auto_increment_tracker.go
zachmu: It's not really racy, but it's kinda busted in the general sense (not actually transactional). I'm not super worried about this though because the way this works in MySQL is already very strange, and MySQL doesn't respect transaction boundaries for DDL statements. If it's a problem we can just vary slightly farther from MySQL and not adjust auto increment values on drop table.

PR #4073, opened 2022-08-12
Preserve enum value case when emitting type metadata

fulghum:
Depends on: https://github.com/dolthub/go-mysql-server/pull/1179

### DIFF
--- a/integration-tests/bats/migration-integration.bats
+++ b/integration-tests/bats/migration-integration.bats
@@ -21,7 +21,7 @@ teardown() {
     run dolt tag -v
     [ "$status" -eq 0 ]
     [[ "$output" =~ "r9jv07tf9un3fm1fg72v7ad9er89oeo7" ]] || false
-    [[ ! "$output" =~ "tetumstdaa81t25h8p5tstsb1f94r4q1" ]] || false
+    [[ ! "$output" =~ "vkoodfe2im3pei1o0vurk66lna6jqhlc" ]] || false
 
     dolt migrate
     [[ $(cat ./.dolt/noms/manifest | cut -f 2 -d :) = "$TARGET_NBF" ]] || false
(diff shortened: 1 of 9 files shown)


integration-tests/bats/migration-integration.bats
```diff
@@ -21,15 +21,15 @@ teardown() {
     run dolt tag -v
     [ "$status" -eq 0 ]
     [[ "$output" =~ "r9jv07tf9un3fm1fg72v7ad9er89oeo7" ]] || false
-    [[ ! "$output" =~ "tetumstdaa81t25h8p5tstsb1f94r4q1" ]] || false
+    [[ ! "$output" =~ "vkoodfe2im3pei1o0vurk66lna6jqhlc" ]] || false
```
fulghum, line 24: I was surprised these tests broke. If there's a case-sensitive schema that's used as part of creating the commit hash, that would make sense about why this changed though.

zachmu: This is why it's happening:

Here's what we write to disk when serializing schema:

```
	for i := len(cols) - 1; i >= 0; i-- {
		col := cols[i]
...
		to := b.CreateString(sqlTypeString(col.TypeInfo))
...
		serial.ColumnAddSqlType(b, to)
	}
```

That method does this:

```
func sqlTypeString(t typeinfo.TypeInfo) string {
	typ := t.ToSqlType()
	if st, ok := typ.(sql.SpatialColumnType); ok {
		// for spatial types, we must append the SRID
		if srid, ok := st.GetSpatialTypeSRID(); ok {
			return fmt.Sprintf("%s SRID %d", typ.String(), srid)
		}
	}
	return typ.String()
}```

So when you changed the `String()` methods of all those types in GMS, it actually changed the commit hashes of anything using those databases.

Do we care about this enough to bump the feature version?

zachmu (approved):
I think this is fine but let's ask the team about the commit hash changing

integration-tests/bats/migration-integration.bats
fulghum: Just closing the loop here after the team discussion... I followed up with additional testing for diff and merge scenarios, particularly looking to make sure Dolt didn't interpret the difference in schema hashes (with the new casing) as a schema change. After testing with the old and new Dolt versions, everything was looking good for diff and merge so we're moving ahead.

PR #4087, opened 2022-08-15
go/store/{nbs,datas/pull}: Fix race condition in concurrent pulls which can truncate a table file.

reltuk:
In `NBSStore.WriteTableFile`, the code used to open the destination path with `O_TRUNC`, without checking in the file already exists or performing any synchronization with concurrent processes. It did this even before even calling `getRd()` to see if it would return an error.

As a result, a race condition exists, where a file gets written to the table file store and added to the manifest and then a later process calls `WriteTableFile` with that same filename and the file gets truncated. This is most likely to happen if there are two concurrent pulls for the same content, which will result in building up the same table files. One of the writes will win, and the later write will lose, but will still truncate the table file. In turn, this pull behavior is most likely to happen if you are a running a read replica sql-server, which might kick off two concurrent pulls for the same remote ref if two connections come in against that branch at the same time.

This PR changes the Puller path and some related paths to be less racey. In particular:

1) Change our use of `CmpChunkTableWriter` so that we don't flush to conflicting temporary file names. We can get a Reader without needing to generate a new temporary file name.
2) Change the write paths for NBSStore.WriteTableFile and gcCopier.copyTablesToDir to use the `tempfiles.MovableTempFileProvider` write-and-rename pattern. This is the same pattern that is used in file_table_persister, and it is safer than writing non-temp-tablefiles directly into the table file store.

### DIFF
--- a/go/store/datas/pull/puller.go
+++ b/go/store/datas/pull/puller.go
@@ -31,7 +31,6 @@ import (
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 
-	"github.com/dolthub/dolt/go/libraries/utils/file"
 	"github.com/dolthub/dolt/go/store/chunks"
 	"github.com/dolthub/dolt/go/store/hash"
 	"github.com/dolthub/dolt/go/store/nbs"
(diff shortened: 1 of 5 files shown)


andrew-wm-arthur (approved):
LGTM! mutli-process concurrency is hard

PR #4089, opened 2022-08-15
don't overwrite parent's `.doltcfg` if current is missing

JCOR11599:
fixes: https://github.com/dolthub/dolt/issues/4088

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -217,31 +217,33 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 		cfgDirPath = filepath.Join(dataDir, DefaultCfgDirName)
 	} else {
 		// Look in parent directory for doltcfg
-		path := filepath.Join("..", DefaultCfgDirName)
-		if exists, isDir := dEnv.FS.Exists(path); exists && isDir {
-			cfgDirPath = path
-		}
+		parentDirCfg := filepath.Join("..", DefaultCfgDirName)
+		parentExists, isDir := dEnv.FS.Exists(parentDirCfg)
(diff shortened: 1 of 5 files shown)


go/cmd/dolt/commands/sql.go
```diff
@@ -235,7 +235,11 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 			}
 			return HandleVErrAndExitCode(errhand.VerboseErrorFromError(ErrMultipleDoltCfgDirs.New(p1, p2)), usage)
 		}
-		cfgDirPath = path
+
+		// Only use current directory if there wasn't one in parent
```
zachmu, line 239: This isn't right, this is exactly opposite of the intended behavior

If you don't specify a doltcfg dir or a data dir, you should get the default, which is 1) ./.doltcfg if it exists, or 2) ../.doltcfg otherwise

zachmu (commented):
See comments

Always use the current directory's config if it exists and hasn't been overridden via flags

go/cmd/dolt/commands/sql.go
```diff
@@ -224,18 +224,22 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 
 		// Look in data directory (which is necessarily current directory) for doltcfg
 		path = filepath.Join(dataDir, DefaultCfgDirName)
```
zachmu, line 226: The way this is written is confusing. Use variable names to refer to actual directories, e.g. `currDirCfg`, `parentDirCfg`

Don't reuse path for these two different values, and don't assign anything to cfgDirPath (which exists outside this condition) until you decide what value it should have

zachmu (approved):
LGTM, just one comment

PR #4091, opened 2022-08-15
Lock new databases where source db is locked

max-hoffman:


go/libraries/doltcore/env/environment.go
```diff
@@ -1237,24 +1233,41 @@ func getProcessFromLockFile(fs filesys.Filesys, lockFile string) (int, error) {
...
+	lockFile, _ := fs.Abs(filepath.Join(dbfactory.DoltDir, ServerLockFile))
+	return fs.WriteFile(lockFile, []byte(fmt.Sprintf("%d", os.Getpid())))
 }
 
-// Unlock deletes this database's lockfile
-func (dEnv *DoltEnv) Unlock() error {
-	if dEnv.IgnoreLockFile {
-		return nil
+// FsIsLocked returns true if a lockFile exists with the same pid as
+// the calling process.
```
zachmu, line 1252: Not the calling process right? Any live process

zachmu (approved):
LGTM

PR #4093, opened 2022-08-15
allow optional empty value for string value for arg parsing

jennifersp:
`--socket` flag can be defined without value, which should get default value of `/tmp/mysql.sock`
To allow empty string as value, added a new arg option type.


fulghum (approved):
Great fix. Nice job knocking this one out so quickly! 🎉

PR #4100, opened 2022-08-16
Adding a skipped bats test for dolt_reset being disallowed in read-only mode

fulghum:
We had a skipped BATS test for `dolt_commit` being disallowed, but `dolt_reset` also needs one since it modifies the branch HEAD.


max-hoffman (commented):
Seems like we either need `*sql.Context` to expose read only, or a globally static system variable?

fulghum:
Yup, `*sql.Context` is what I was thinking. Seems like a pretty reasonable place to look for that info.

PR #4102, opened 2022-08-16
Various fixes for database revision specifiers

fulghum:
Changing revision databases to be instantiated on-demand, instead of trying to track them in DoltDatabaseProvider. This simplifies how we manage revision databases and avoids revision dbs hanging around in memory after customers are done using them. The biggest UI changes for customers is that `show databases` will now no longer show all revision dbs that have ever been used, but instead will only include the single revision database a customer is currently using. 

Fixes: https://github.com/dolthub/dolt/issues/4067
Fixes: https://github.com/dolthub/dolt/issues/4031


integration-tests/bats/deleted-branches.bats
```diff
@@ -73,34 +73,6 @@ make_it() {
     server_query "dolt_repo_$$/main" 1 "SELECT * FROM test" "id\n"
 }
 
-@test "deleted-branches: can DOLT_CHECKOUT on SQL connection with existing branch revision specifier when dolt_default_branch is invalid" {
```
fulghum, line 76: These two tests happened to work, because there was an existing session that `use`d the branch-qualified database revision, before the default branch was corrupted. Since this change stopped tracking database revisions as databases in the DatabaseProvider, these tests started failing. 

I spent some time digging through this, but ended up deciding this was not a super important edge case to continue supporting, so I removed the tests. I'm happy to give it some more time if people think this is important though.

zachmu: I think rather than deleting, the right approach is just to lower the bar for this test so the behavior is still defined and we know it doesn't panic. Basically anything can happen, we just want to make sure it's not a fatal error.

zachmu: Same for the one below obv

integration-tests/bats/sql-server.bats
```diff
@@ -1258,7 +1258,7 @@ END""")
     # can't drop a branch-qualified database name
     run server_query "" 1 "drop database \`test2/newbranch\`"
     [ "$status" -ne 0 ]
-    [[ "$output" =~ "database not found: test2/newbranch" ]] || false
+    [[ "$output" =~ "unable to drop revision database: test2/newbranch" ]] || false
 
     server_query "" 1 "drop database TEST2"
```
zachmu, line 1263: Would be cool to have some bats tests showing the output for `show databases` after something like `use mydb/HEAD~`

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -125,6 +125,146 @@ var DescribeTableAsOfScriptTest = queries.ScriptTest{
 	},
 }
 
+var DoltRevisionDbScripts = []queries.ScriptTest{
```
zachmu, line 128: Any tests for things like `HEAD~`? What about commit hashes?

zachmu (approved):
Overall looks really good, just some suggestions for tests.

Did `use mydb/HEAD~` not make it into this?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: No support for `HEAD~` refs yet, but I'd be more than happy to add that in a separate PR. Seems like a very natural and useful extension. 

I made BATS tests covering commit hashes in db rev specs, but didn't come up with a good way to get the commit hash and easily use it in the enginetests with the `ScriptTest` structures. I figured the BATS coverage was decent, but I'll try adding a non-ScriptTest and see if I can get some enginetest coverage for commits, too.

PR #4105, opened 2022-08-16
commit merge if no conflict

jennifersp:
This PR implements performing `dolt commit` if merge is not fast forward and there is no conflict and constraint violation
It adds `--commit`, `--no-commit` and `--no-edit` flag options for both `dolt merge` and `dolt pull`.
Same conditions apply for `dolt pull` merging a remote branch into current working set. 
Dolt merge does not log commit after committing successful merge.


go/libraries/doltcore/merge/action.go
```diff
@@ -98,26 +100,40 @@ func NewMergeSpec(ctx context.Context, rsr env.RepoStateReader, ddb *doltdb.Dolt
...
-		return nil, ExecuteFFMerge(ctx, dEnv, spec)
+		return false, nil, ExecuteFFMerge(ctx, dEnv, spec)
 	}
-	return ExecuteMerge(ctx, dEnv, spec)
+	tblStats, err := ExecuteMerge(ctx, dEnv, spec)
+	if err != nil {
+		return false, tblStats, err
+	}
+
+	// merge will commit if
```
fulghum, line 132: These seem like they'd be more useful in the function godocs– great explanation for the different behavior and how the different flags affect it.

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -213,6 +232,13 @@ func mergeIntoWorkingSet(ctx *sql.Context, sess *dsess.DoltSession, roots doltdb
 		return ws, noConflictsOrViolations, threeWayMerge, err
 	}
 
+	if !noCommit && ac {
```
fulghum, line 235: I'm not sure if we need to switch on the autocommit setting... it's a good question though... 

Looks like you were thinking if autocommit is disabled, we should skip the merge commit? (since dolt_commit creates a db/tx commit implicitly) 

My gut reaction is that we should not have autocommit come into play. My thinking is that it's one less dimension for customers to have to reason about, it's more consistent with the command line, and it seems like customers should expect dolt_merge to work like git_merge and create that merge commit by default. Plus, there is an available workaround where customers could specify `--no-commit` to still get that behavior. 

If you see another perspective, I'm always happy to reconsider and keep discussing if you feel like there's something important here.

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -293,7 +319,7 @@ func executeNoFFMerge(
 	dbName string,
 	ws *doltdb.WorkingSet,
 	dbData env.DbData,
-	//headCommit, mergeCommit *doltdb.Commit,
+	// headCommit, mergeCommit *doltdb.Commit,
```
fulghum, line 322: (minor) might as well remove this old comment; doesn't seem like it's doing anything helpful.

```suggestion
```

integration-tests/bats/merge.bats
```diff
@@ -851,3 +846,23 @@ SQL
...
+    dolt branch other
+    dolt sql -q "INSERT INTO test1 VALUES (1,2,3)"
+    dolt commit -am "add (1,2,3) to test1";
+
+    dolt checkout other
+    dolt sql -q "INSERT INTO test1 VALUES (2,3,4)"
+    dolt commit -am "add (2,3,4) to test1";
+
+    dolt checkout main
+    run dolt merge other -m "merge other"
```
fulghum, line 860: This would be a good place to test out `--no-commit`, too. You could run the merge first with `--no-commit`, test that the changes are staged, but not committed, then `reset --hard` and run the merge with a commit message.

integration-tests/bats/merge.bats
```diff
@@ -180,7 +180,7 @@ teardown() {
     dolt add test1
     dolt commit -m "add pk 0 = 2,2 to test1"
 
-    run dolt merge merge_branch
+    run dolt merge merge_branch --no-commit
```
fulghum, line 183: since this test is testing merge failing on a conflict, it might give us a little more coverage to specify a commit message here, instead of `--no-commit` since we expect the merge to fail and not commit anyway. That would just help test the case where the user requested to commit, but merge conflicts prevent it.

integration-tests/bats/merge.bats
```diff
@@ -456,7 +456,7 @@ SQL
     dolt commit -am "right commit"
```
fulghum, line 456: Could you check out the `skip` on line 434? Do we detect that constraint violation yet? Just wondering if we can remove that skip now.

integration-tests/bats/merge.bats
```diff
@@ -851,3 +846,23 @@ SQL
     log_status_eq 1
     [[ "$output" =~ "table with same name deleted and modified" ]] || false
 }
+
```
fulghum, line 849: line 817 has a good check with a merge that finds a constraint violation. Would be good to add a line there just to test that the commit didn't happen and that changes are in the working set.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -1183,7 +1183,7 @@ var MergeScripts = []queries.ScriptTest{
 		},
 		Assertions: []queries.ScriptTestAssertion{
 			{
-				Query:    "CALL DOLT_MERGE('feature-branch', '-m', 'this is a merge')",
+				Query:    "CALL DOLT_MERGE('feature-branch', '-m', 'this is a merge', '--no-commit')",
```
fulghum, line 1186: It would be good to have a little more coverage of dolt_merge in the enginetests. I'd cover the happy path with merge creating a commit and also fast-forward. Then I'd cover the error cases where a commit couldn't be created.

integration-tests/bats/sql-merge.bats
```diff
@@ -1669,9 +1669,9 @@ SQL
     dolt commit -am "changes to b2"
 
     dolt checkout main
-    run dolt merge b1
+    run dolt merge b1 -m "merge b1"
```
fulghum, line 1672: Would be good to throw in the `--commit` arg somewhere, like here. Just to make sure we have a test that uses it, even though it's the default. 

Speaking of... do we have code to error out if `--commit` and `--no-commit` are both specified? That would be a good error case to test, too.

fulghum (commented):
Nice work! Overall, this is looking very close. I left a bunch of comments, but a lot of them are just really simple wording suggestions that you can just one-click accept.

go/libraries/doltcore/merge/action.go
```diff
@@ -98,26 +103,39 @@ func NewMergeSpec(ctx context.Context, rsr env.RepoStateReader, ddb *doltdb.Dolt
...
 		Name:            name,
 		Date:            date,
-	}, true, nil
+	}, nil
 }
 
+// MergeCommitSpec returns whether to create new commit and MergeStats object after performing merge and commits
+// if applied. If merge can fast-forward, no commit is returned. ExecNoFFMerge function call commits after merge.
+// If merge is not fast-forward, 'no-commit' flag is not defined and if there are no conflicts and/or
+// constraint violations, commit(true) is returned.
```
fulghum, line 118: Good function docs. These will be a big help to future readers of the code! How about this small tweak to make it read better... let me know if this isn't still accurate... 

```suggestion
// MergeCommitSpec applies a merge spec, potentially fast-forwarding the current branch HEAD, and returns two 
// results: true if a new Dolt commit needs to be created to complete the merge, and a MergeStats object.
// If the merge can be applied as a fast-forward merge, no commit is needed, so false is returned for the first result.
// If the merge is a fast-forward merge, but --no-ff has been supplied, the ExecNoFFMerge function will call
// commit after merging. If the merge is not fast-forward, the 'no-commit' flag is not defined, and if there are 
// no conflicts and/or constraint violations, commit(true) is returned.
```

Actually... after rewriting this method doc comment... this seems like an odd interface. I think it's okay for this iteration, but take a step back and think about using this API as a caller. It's not intuitively obvious why this method behaves like it does – it commits in some cases, but not in others and the caller has to commit. This is a signal that we could simplify the code a bit more. Can we push the responsibility to create the commit into this method so that it's easier to explain how this method works? That could also let us remove the "doCommit" returned result, since the caller wouldn't need a signal to do a commit in some edge cases. 

Totally fine to do this as a separate PR if you prefer, so we can get this change landed in main first. Either way, I think it's a good experiment to try to see if we can simplify the interface a little bit and push more consistent responsibility into this method (which is sometimes responsible for making a commit, sometimes not).

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -133,11 +143,10 @@ func DoDoltMerge(ctx *sql.Context, args []string) (int, int, error) {
...
 // fast-forward, no fast-forward, merge commit, and merging into working set.
 // Returns a new WorkingSet, whether there were merge conflicts, and whether a
-// fast-forward was performed. This currently persists merge commits in the
-// database, but expects the caller to update the working set.
+// fast-forward was performed. This commits the working set if merge is successful and
+// -no-commit flag is not defined and autocommit is on
 // TODO FF merging commit with constraint violations requires `constraint verify`
-func mergeIntoWorkingSet(ctx *sql.Context, sess *dsess.DoltSession, roots doltdb.Roots, ws *doltdb.WorkingSet, dbName string, spec *merge.MergeSpec) (*doltdb.WorkingSet, int, int, error) {
-
+func mergeIntoWorkingSet(ctx *sql.Context, sess *dsess.DoltSession, roots doltdb.Roots, ws *doltdb.WorkingSet, dbName string, spec *merge.MergeSpec, noCommit bool, msg string) (*doltdb.WorkingSet, int, int, error) {
```
fulghum, line 149: This function is a good example of a nice encapsulations cope for a function – it handles all the merge logic and commits if it needs to be done. The caller has a really nice, clear, easy to understand interface to work with. This version is a good one to contrast with the earlier CLI/command version, from a caller perspective. This one takes care of everything and makes it really easy for the caller. 

However... we should change the name now 😉  `mergeIntoWorkingSet` is true... but it doesn't tell the full story. Most of the time this will create a commit, too, so workingset is a little misleading. `mergeAndCommit` or even just `performMerge` might be a little more intuitive of a name.

go/libraries/doltcore/sqle/dfunctions/dolt_merge.go
```diff
@@ -133,11 +143,10 @@ func DoDoltMerge(ctx *sql.Context, args []string) (int, int, error) {
 // mergeIntoWorkingSet encapsulates server merge logic, switching between
 // fast-forward, no fast-forward, merge commit, and merging into working set.
 // Returns a new WorkingSet, whether there were merge conflicts, and whether a
-// fast-forward was performed. This currently persists merge commits in the
-// database, but expects the caller to update the working set.
+// fast-forward was performed. This commits the working set if merge is successful and
+// -no-commit flag is not defined and autocommit is on
```
fulghum, line 147: Do we need the `autocommit` note here?

fulghum (commented):
This is looking really good now. Just a couple of comments. 

I just wanted to test the interactive tty code a little bit, before giving final approval.

fulghum (approved):
Everything is looking great! Thanks for talking through the `tty` detection changes to make sure we were doing the right thing. Good work! I'm very happy with these changes. 😄

PR #4110, opened 2022-08-16
sql-server host shouldn't affect default superuser

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/4079

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/server.go
+++ b/go/cmd/dolt/commands/sqlserver/server.go
@@ -126,11 +126,6 @@ func Serve(
 	} else if sErr != nil {
 		return sErr, nil
 	}
-	// TODO: move this to GMS
-	serverHost := serverConfig.Host()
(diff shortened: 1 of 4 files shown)


timsehn:
This is weird. The test you wrote failed with the old behavior.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -135,7 +135,7 @@ func Serve(
 		DoltCfgDirPath: serverConfig.CfgDir(),
 		ServerUser:     serverConfig.User(),
 		ServerPass:     serverConfig.Password(),
-		ServerHost:     serverConfig.Host(),
+		ServerHost:     "%",
```
zachmu, line 138: This isn't right. We don't want to give up on the -H option entirely for network routing, which this does. We just want to ignore the host name when creating the first user.

zachmu (commented):
See comments

zachmu (approved):
LGTM

PR #4121, opened 2022-08-17
Update README to reflect socket interface release

timsehn:
Socket interface supported but not on by default.


jennifersp (approved):
LGTM!

PR #4122, opened 2022-08-17
Burlier tests for multi-db replication

zachmu:


timsehn (commented):
Should we have a create database/bootsrap w/ clone test?

max-hoffman (approved):
LGTM

timsehn (commented):
I think we should have an explicit create database, set origin and then dolt clone, set replica test as well

zachmu:
> I think we should have an explicit create database, set origin and then dolt clone, set replica test as well

I'll add a test for that when I make it work, definitely broken right now.

PR #4156, opened 2022-08-19
Allows diff to be performed across primary key renames

druvv:


integration-tests/bats/diff.bats
```diff
@@ -869,3 +869,55 @@ SQL
...
+    [ $status -eq 0 ]
+
+    [[ "$output" =~ ' CREATE TABLE `t1` (' ]] || false
+    [[ "$output" =~ '-  `pk` int NOT NULL,' ]] || false
+    [[ "$output" =~ '+  `pk2` int NOT NULL,' ]] || false
+    [[ "$output" =~ '   `col1` int,' ]] || false
+    [[ "$output" =~ '-  PRIMARY KEY (`pk`)' ]] || false
+    [[ "$output" =~ '+  PRIMARY KEY (`pk2`)' ]] || false
+    [[ "$output" =~ ' ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin;' ]] || false
+    [[ "$output" =~ '+---+------+------+------+' ]] || false
```
zachmu, line 901: Maybe a little overspecified for the output checks but also fine to be lazy if you make it easy to change.

If you're gonna match large multiline text like this, use this method:

https://github.com/dolthub/dolt/blob/main/integration-tests/compatibility/test_files/bats/compatibility.bats#L135

zachmu (approved):
LGTM. You should think of ways we can make these column mathcing / mapping rules apply consistently throughout the code base, kind of piecemeal right now.

PR #4161, opened 2022-08-20
dolt commit -a ignores new tables

### LINKED ISSUE #4133: In the`dolt commit`, the `-a` option adds new tables, not just modified or deleted ones.
This deviates from the Git standard, and though I think most of us prefer how Dolt does it, we should match Git to avoid confusion in our user base.

tanyongzhi:
Fixes #4133.

This PR will fix the `-a` flag in `dolt commit` committing new tables. 

The bulk of this PR is fixing old integration tests, as many of them assumes the old behavior of `dolt commit -am`.

### DIFF
--- a/go/cmd/dolt/commands/commit.go
+++ b/go/cmd/dolt/commands/commit.go
@@ -103,7 +103,7 @@ func performCommit(ctx context.Context, commandStr string, args []string, dEnv *
 	}
 
 	if allFlag {
-		roots, err = actions.StageAllTables(ctx, roots)
+		roots, err = actions.StageModifiedAndDeletedTables(ctx, roots)
 		if err != nil {
 			return handleCommitErr(ctx, dEnv, err, help)
 		}
--- a/go/libraries/doltcore/env/actions/staged.go
+++ b/go/libraries/doltcore/env/actions/staged.go
@@ -37,7 +37,7 @@ func StageAllTables(ctx context.Context, roots doltdb.Roots) (doltdb.Roots, erro
 func StageModifiedAndDeletedTables(ctx context.Context, roots doltdb.Roots) (doltdb.Roots, error) {
 	_, unstaged, err := diff.GetStagedUnstagedTableDeltas(ctx, roots)
 	if err != nil {
-		return nil, err
+		return doltdb.Roots{}, err
 	}
 
 	tbls := []string{}
--- a/go/libraries/doltcore/env/actions/staged.go
+++ b/go/libraries/doltcore/env/actions/staged.go
@@ -17,6 +17,7 @@ package actions
 import (
 	"context"
 
+	"github.com/dolthub/dolt/go/libraries/doltcore/diff"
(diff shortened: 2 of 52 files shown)


fulghum (commented):
Nice job on this one @tanyongzhi! Thank you for jumping in and picking this one up! 🙏 

This code all looked good to me and takes care of the CLI code path. Would you mind also including [the change for the SQL code path](https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/sqle/dfunctions/dolt_commit.go#L69)? 

After that, the only thing missing would be making sure we have tests that assert for both the SQL path and the CLI path, that a newly created table isn't added when a user calls `dolt commit -am "..."` (or `call dolt_commit('-am', '...')` for the SQL path). 

Thank you for sending us this!

tanyongzhi:
My pleasure, and thank you for the quick review @fulghum.

I've updated the code for the SQL path. However many integration tests are now failing. These tests seem to belong to `go-mysql-server`, which to my knowledge belongs to another repository. How should we go about fixing these tests?

Lastly, I've also added a couple of bats tests to test the new intended behavior of `dolt commit -a` in `multiple-tables.bats`.

tanyongzhi:
I'm also struggling a little with a few failing bats tests. Would greatly appreciate it if someone could take a look!

fulghum:
This is looking great @tanyongzhi. Once we get the last few tests passing, this should be good to merge in to main. 🎉 

For the enginetests, those would use the `dolt_commit` stored procedure as part of creating test data, so makes sense that they started failing after the update to `dolt_commit`. Try grep'ing for `call dolt_commit(` inside the `go` directory of the dolt repo and you should find those references (e.g. `dolt_queries.go`). The BATS tests must be something similar – test data is calling `dolt_commit` but not adding some new tables. 

Many test cases do come from the `go-mysql-server` package, but since that layer isn't aware of anything about dolt or dolt's versioning, all the changes for this task will be inside the dolt repo. 

I know updating all the test references is a pain – definitely something we weren't excited about either! 😉 So, major thanks for chipping in with this. One idea zachmu and I were talking about yesterday was... in addition to this change to match Git's behavior with `dolt commit -a`, we could add a `dolt commit -A` that would also stage new tables, as an alternative to to explicitly calling `dolt add -A`. It's not something `git commit` supports, but feels like a nice ergonomic improvement and does match Git's use of `-A` with `git add`. If that makes updating the tests easier, then feel free to add that in. No worries if you want to keep this PR tightly scoped of course, just wanted to mention that idea in case it makes updating the tests easier. 

Let me know if you have any problems finding the other test references and we're happy to help more. 🙏

timsehn:
Thank you so much for all your hard work here!

fulghum (approved):
@tanyongzhi – this all looks great! Thank you so much for taking on this change and contributing back to Dolt! 🙏 This change needed to touch just about _every_ test file and we are _very_ grateful and appreciative that you put the time to contribute this! Thank you!!

I'll go ahead and get this merged in to main, see if our docs need any small updates for this behavior, and this will go out in our next release.

PR #4163, opened 2022-08-22
adding missing test cases for prepared tests

JCOR11599:
Companion PR:
 - https://github.com/dolthub/go-mysql-server/pull/1207

Adds skipped tests for Queries that don't yet work Prepared


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -822,29 +824,6 @@ var HistorySystemTableScriptTests = []queries.ScriptTest{
 					{3, 4},
 				},
 			},
-			{
```
max-hoffman, line 825: it might be easier to leave these here with `SkipPrepared=true`, but still append to a new list of tests titled `BrokenHistorySystemTableScriptTests `.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -1073,6 +1113,35 @@ func TestHistorySystemTable(t *testing.T) {
 			enginetest.TestScript(t, harness, test)
 		})
 	}
+	for _, test := range BrokenHistorySystemTableScriptTests {
+		harness.engine = nil
```
max-hoffman, line 1117: you verified that the tests that include this cannot reuse the engine?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
max-hoffman: in which case you don't need the comment above

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -66,21 +66,43 @@ func TestQueries(t *testing.T) {
 func TestSingleQuery(t *testing.T) {
 	t.Skip()
 
+	harness := newDoltHarness(t)
```
max-hoffman, line 69: in general it is courteous to limit unnecessary changes that make PRs more difficult to read

max-hoffman (approved):
LGTM, some small comments

PR #4193, opened 2022-08-24
create db with specific format depending on currentDB format

jennifersp:
`CREATE DATABASE` on new-format DB should create DB with new-format(same format as the currentDB)


andrew-wm-arthur (approved):
LGTM!

PR #4194, opened 2022-08-25
Support json output for `dolt diff`

zachmu:
Fixes https://github.com/dolthub/dolt/issues/2243


go/libraries/doltcore/table/table_reader.go
```diff
@@ -34,29 +34,29 @@ type TableReader interface {
...
 type SqlRowReader interface {
-	TableReadCloser
+	ReadCloser
 
 	ReadSqlRow(ctx context.Context) (sql.Row, error)
 }
 
-// SqlTableReader is a  TableReader that can read rows as sql.Row.
+// SqlTableReader is a Reader that can read rows as sql.Row.
 type SqlTableReader interface {
```
fulghum, line 56: (minor) Should `SqlTableReader` just embed an instance of `SqlRowReader` to pick up the `ReadSqlRow` method?

integration-tests/bats/json-diff.bats
```diff
@@ -0,0 +1,297 @@
...
+}
+
+@test "json-diff: new row" {
+    dolt add .
+    dolt commit -m table
+    dolt sql -q 'insert into test values (0,0,0,0,0,0)'
+    dolt add .
+    dolt commit -m row
+
+    dolt diff -r json
```
fulghum, line 32: (minor) is this line doing something? Seems like the exit status check for the next line would capture any failure.

integration-tests/bats/json-diff.bats
```diff
@@ -0,0 +1,297 @@
...
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ '{"tables":[{"name":"test","schema_diff":[],"data_diff":[{"from_row":{},"to_row":{"c1":0,"c2":0,"c3":0,"c4":0,"c5":0,"pk":0}}]}]}' ]] || false
+
+    dolt diff -r json head head^
+    run dolt diff -r json head head^
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ '{"tables":[{"name":"test","schema_diff":[],"data_diff":[{"from_row":{"c1":0,"c2":0,"c3":0,"c4":0,"c5":0,"pk":0},"to_row":{}}]}]}' ]] || false
+}
+
+@test "json-diff: data and schema changes" {
```
fulghum, line 57: I found an issue with schema+data changes while testing out a build...

I started with your BATS test table, then ran:
* ALTER TABLE `test` RENAME COLUMN `c2` TO `c2new`;
* update test set c2new = 1;

`dolt diff` shows:
```
+---+----+----+------+----+----+----+-------+
|   | pk | c1 | c2   | c3 | c4 | c5 | c2new |
+---+----+----+------+----+----+----+-------+
| < | 0  | 0  | 0    | 0  | 0  | 0  | NULL  |
| > | 0  | 0  | NULL | 0  | 0  | 0  | 1     |
+---+----+----+------+----+----+----+-------+
```

But `dolt diff -r json` shows:
```
 {
          "from_row": {
            "c1": 0,
            "c2new": 0,
            "c3": 0,
            "c4": 0,
            "c5": 0,
            "pk": 0
          },
          "to_row": {
            "c1": 0,
            "c3": 0,
            "c4": 0,
            "c5": 0,
            "pk": 0
          }
}
```

I would expect the c2new column to be listed in the to_row with the value `1`.

go/cmd/dolt/commands/diff.go
```diff
@@ -178,7 +164,7 @@ func (cmd DiffCmd) validateArgs(apr *argparser.ArgParseResults) errhand.VerboseE
 
 	f, _ := apr.GetValue(FormatFlag)
 	switch strings.ToLower(f) {
-	case "tabular", "sql", "":
+	case "tabular", "sql", "json", "":
```
fulghum, line 167: line 128 for the `-r` arg documentation could use an update to mention the new `json` param it accepts.

fulghum (approved):
Looks good overall. Biggest issue I noticed was output when renaming columns where it looks like the code isn't showing a correct result.

go/cmd/dolt/commands/diff.go
zachmu: Good catch!

integration-tests/bats/json-diff.bats
zachmu: It makes it so that you get the output the command you're testing when the test fails, kinda nice

go/libraries/doltcore/table/table_reader.go
zachmu: Gonna leave it for now

integration-tests/bats/json-diff.bats
zachmu: Good catch, fixed and tests added.

PR #4196, opened 2022-08-25
Add support for table-level collations

Hydrocharged:
This is the Dolt side of https://github.com/dolthub/go-mysql-server/pull/1215.


go/libraries/doltcore/sqle/database.go
```diff
@@ -824,7 +824,7 @@ func (db Database) removeTableFromAutoIncrementTracker(
 }
 
 // CreateTable creates a table with the name and schema given.
-func (db Database) CreateTable(ctx *sql.Context, tableName string, sch sql.PrimaryKeySchema) error {
+func (db Database) CreateTable(ctx *sql.Context, tableName string, sch sql.PrimaryKeySchema, collation sql.CollationID) error {
```
andrew-wm-arthur, line 827: is `collation sql.CollationID` not part of `sql.PrimaryKeySchema`?

go/serial/schema.fbs
```diff
@@ -22,6 +22,7 @@ table TableSchema {
   clustered_index:Index (required);
   secondary_indexes:[Index];
   checks:[CheckConstraint];
+  collation:uint16;
```
andrew-wm-arthur, line 25: In all other instances, SQL type information (including column collations) is stored as string data that can be interpreted by the SQL parser. We should either replicate that pattern here, or create a flatbuffers type alias for `uint16` and enumerate all possible values for `collation`. See `serial.Encoding` as an exampl

go/libraries/doltcore/sqle/database.go
Hydrocharged: It is not. I think it may belong outside of the schema from the GMS standpoint, along with the implementation engine (of which we only support InnoDB). In Dolt though, we pass along everything in the schema, so it’s far more straightforward to store it there.

go/serial/schema.fbs
Hydrocharged: I like the FlatBuffers type alias, I’ll implement that.

PR #4197, opened 2022-08-25
Add support for ancestor references in database revision specifiers

fulghum:
Enables use of `~` and `^` to reference ancestors in database revision specifiers. 

Fixes: https://github.com/dolthub/dolt/issues/2146

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -828,6 +828,42 @@ func (ddb *DoltDB) VisitRefsOfType(ctx context.Context, refTypeFilter map[ref.Re
 	})
 }
 
+// GetRefByNameInsensitive searches this Dolt database's branch, tag, and head refs for a case-insensitive
+// match of the specified ref name. If a matching DoltRef is found, it is returned; otherwise an error is returned.
+func (ddb *DoltDB) GetRefByNameInsensitive(ctx context.Context, refName string) (ref.DoltRef, error) {
+	branchRefs, err := ddb.GetBranches(ctx)
+	if err != nil {
+		return nil, err
+	}
+	for _, branchRef := range branchRefs {
+		if strings.ToLower(branchRef.GetPath()) == strings.ToLower(refName) {
+			return branchRef, nil
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -178,6 +178,81 @@ var DescribeTableAsOfScriptTest = queries.ScriptTest{
...
+			{
+				Query:    "select * from `mydb/tag1~1`.t01;",
+				Expected: []sql.Row{{1, 1}, {2, 2}},
+			},
+			{
+				Query:    "select * from `mydb/tag1~2`.t01;",
+				Expected: []sql.Row{},
+			},
+			{
+				Query:       "select * from `mydb/tag1~3`.t01;",
```
zachmu, line 235: Do you see a need for an error type for a refspec to a commit that doesn't exist? Like HEAD~3 in a database with 2 commits?

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -828,6 +828,42 @@ func (ddb *DoltDB) VisitRefsOfType(ctx context.Context, refTypeFilter map[ref.Re
 	})
 }
 
+// GetRefByNameInsensitive searches this Dolt database's branch, tag, and head refs for a case-insensitive
+// match of the specified ref name. If a matching DoltRef is found, it is returned; otherwise an error is returned.
+func (ddb *DoltDB) GetRefByNameInsensitive(ctx context.Context, refName string) (ref.DoltRef, error) {
+	branchRefs, err := ddb.GetBranches(ctx)
```
zachmu, line 834: This is fine for now but speaks to a need for constant time case insensitive lookups for these things, rather than this 3N approach

zachmu (approved):
Love it

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: Good question... the existing code does return a `doltdb.ErrInvalidAncestorSpec` when the ancestor spec can't be resolved to a commit, but I didn't have a test case for that. I just added a new assertion to this test to trigger that error.

PR #4201, opened 2022-08-25
proto,go/.../remotestorage: Add repo_token support.

reltuk:
doltremoteapi can return an opaque token which makes it cheaper to access the repository compared to just using the RepoId. This updates protos and remotestorage implementation to cache that token and use it for future accesses of the same repository.


druvv (approved):
LGTM

PR #4206, opened 2022-08-25
fix wrong error variable used

jennifersp:


timsehn (commented):
Should we have a bats test with a bad parquet file to catch things like this in the future?

jennifersp:
Ah yes, I'm not sure how to generate bad parquet file. What I was thinking is to either add skipped or error asserted test on importing Apache Arrow generated parquet file for now, and eventually migrate to its parquet go library.

timsehn:
I bet if you made a text file that had "This is a bad parquet file", it would catch most errors :-)

fulghum (approved):
LGTM!

PR #4208, opened 2022-08-25
go/store/nbs: byte_sink.go: Fix panic in some cases when we retry table file upload in push.

reltuk:
The buffered byte sink finish() implementation was expected to only be called,
but Reader() could be called multiple times for a retried upload.


druvv (approved):
thank you! LGTM

PR #4215, opened 2022-08-26
Secondary lookup cursor reuse

max-hoffman:
Locally on OSX `index_join_scan` goes from `3.4ms/q` -> `2.5ms/q`.


max-hoffman:
#benchmark

max-hoffman:
#benchmark

max-hoffman:
#benchmark

max-hoffman:
#benchmark

max-hoffman:
#benchmark

timsehn:
OHHHHHH BABY!!!!

go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -748,6 +748,12 @@ func pruneEmptyRanges(sqlRanges []sql.Range) (pruned []sql.Range, err error) {
 				break
 			}
 		}
+		for _, ce := range sr {
+			if lb, ok := ce.LowerBound.(sql.Below); ok && lb.Key == nil {
+				empty = true
+				break
+			}
+		}
```
reltuk, line 756: Where is a `LowerBound: sql.Below{Key: nil}` coming from? Shouldn't we be avoiding that?

go/libraries/doltcore/sqle/index/index_lookup.go
```diff
@@ -205,59 +220,214 @@ func NewLookupBuilder(part sql.Partition, idx DoltIndex, projections []uint64, p
...
+	}
+}
+
+var _ LookupBuilder = (*baseLookupBuilder)(nil)
+var _ LookupBuilder = (*nomsLookupBuilder)(nil)
+var _ LookupBuilder = (*coveringLookupBuilder)(nil)
+var _ LookupBuilder = (*keylessLookupBuilder)(nil)
+var _ LookupBuilder = (*nonCoveringLookupBuilder)(nil)
+
+// baseLookupBuilder is a common lookup builder for prolly covering and
```
reltuk, line 269: This is a little mind-bending where this `baseLookupBuilder` implements this interface for the prolly case, but it also still gets embedded in a nomsLookupBuilder. I think it would be more clear at this point if the common fields went on something else and this thing should be a prollyLookupBuilder or something.

reltuk (approved):
Looks great. One question, one comment...

go/libraries/doltcore/sqle/index/dolt_index.go
max-hoffman: Yeah this is a meaningless range that is produced as a side-effect of my range building refactors in specific edge cases with Left/Right joins. It was just ignored before, because of the way we were searching for cursor items, and panicked with the new direct cursor lookup. The correct fix is probably making this an empty range on the GMS side. I hope to clean this up in a follow-on PR.

go/libraries/doltcore/sqle/index/index_lookup.go
max-hoffman: `baseLookupBuilder.RowIter` is never called so I will panic there. I used `nonCoveringLookupBuilder` for `prollyIndexLookup` because it confused me less.

PR #4221, opened 2022-08-26
Bug fix for Tuple.Compare to return consistent results when comparing against a null value

fulghum:
`Tuple.Compare` / `Tuple.Less` returns inconsistent results when comparing different types against a null value. This caused non-null values for most types to be ordered before null values, but other types like time/date/datetime to have non-null values _after_ null values in the index. That resulted in incorrect query execution because the index range reader saw the null values and exited out before ever evaluating the non-null values. 

Fixes: https://github.com/dolthub/dolt/issues/4199

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -331,7 +331,7 @@ var DoltScripts = []queries.ScriptTest{
 	{
 		Name: "test null filtering in secondary indexes (https://github.com/dolthub/dolt/issues/4199)",
 		SetUpScript: []string{
-			"create table t (a int primary key auto_increment, d datetime, index index1 (d));",
+			"create table t (pk int primary key auto_increment, d datetime, index index1 (d));",
 			"insert into t (d) values (NOW()), (NOW());",
 			"insert into t (d) values (NULL), (NULL);",
 		},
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -328,6 +328,24 @@ var DoltRevisionDbScripts = []queries.ScriptTest{
...
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query:    "select count(*) from t where d is not null",
+				Expected: []sql.Row{{2}},
+			},
+			{
+				Query:    "select count(*) from t where d is null",
+				Expected: []sql.Row{{2}},
+			},
```
max-hoffman, line 346: The match all expression `<=>` is often good to test when we care about nulls. Index lookups also come in two flavors, static lookups like you have above, and dynamic lookups like in indexed joins.

go/store/types/tuple.go
```diff
@@ -690,6 +690,15 @@ func (t Tuple) TupleCompare(nbf *NomsBinFormat, otherTuple Tuple) (int, error) {
 		otherKind := otherDec.ReadKind()
 
 		if kind != otherKind {
+			// If we are comparing any type to a null type, always evaluate
```
max-hoffman, line 693: You might already know this, but the types package only applies to the old format. The old format also represents ranges differently and implements range scans differently. CI will fail if none of this works with the new format, so if it passes you're good.

max-hoffman (approved):
LGTM! just small comments

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: Thanks for the tips! I added assertions for the null-safe equals operator and for an indexed join query.

PR #4231, opened 2022-08-30
Benchmark release v0.40.31

max-hoffman:


max-hoffman:
#benchmark

timsehn:
Seems like we may have just gotten a bad dice roll on the release. Let's get these CTE things fixed and get another release to Threekit today and see if the metrics are better.

PR #4239, opened 2022-08-30
[no-release-notes] go/doltcore/sqle/writer: Fix `prollyFkIndexer` for primary keys not at the end of a schema

andrew-wm-arthur:


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -658,6 +658,49 @@ var DoltScripts = []queries.ScriptTest{
 			},
 		},
 	},
+	{
```
Hydrocharged, line 661: Add this to GMS as well, it's a good test to have there.

Hydrocharged (approved):
LGTM!

PR #4243, opened 2022-08-30
use commitItr instead of commitwalk for querying dolt_diff table

jennifersp:
This improves select query on unscoped `dolt_diff` table in old format for `commit_hash` column filter


fulghum (approved):
Nice job Jennifer! Just a couple of really minor comments from me. Good perf win for this table! 😎

PR #4255, opened 2022-09-01
allow init with empty working set

jennifersp:
In sql session, if the current head on CLI is deleted, the repostate head is empty branch ref. 
If the current head on CLI is renamed, the repostate head is updated with new branch ref.
`dolt_branches` table is read-only now.

### DIFF
--- a/go/libraries/doltcore/env/environment.go
+++ b/go/libraries/doltcore/env/environment.go
@@ -175,12 +175,7 @@ func loadWithFormat(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys
 	if rsErr == nil && dbLoadErr == nil && dbFormatErr == nil {
 		// If the working set isn't present in the DB, create it from the repo state. This step can be removed post 1.0.
 		_, err := dEnv.WorkingSet(ctx)
-		if err == doltdb.ErrWorkingSetNotFound {
-			err := dEnv.initWorkingSetFromRepoState(ctx)
-			if err != nil {
-				dEnv.RSLoadErr = err
-			}
-		} else if err != nil {
+		if err != nil && !errors.Is(err, doltdb.ErrWorkingSetNotFound) {
 			dEnv.RSLoadErr = err
 		}
 	}
--- a/go/libraries/doltcore/env/environment.go
+++ b/go/libraries/doltcore/env/environment.go
@@ -175,7 +175,9 @@ func loadWithFormat(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys
 	if rsErr == nil && dbLoadErr == nil && dbFormatErr == nil {
 		// If the working set isn't present in the DB, create it from the repo state. This step can be removed post 1.0.
 		_, err := dEnv.WorkingSet(ctx)
-		if err != nil && !errors.Is(err, doltdb.ErrWorkingSetNotFound) {
+		if errors.Is(err, doltdb.ErrWorkingSetNotFound) {
+			_ = dEnv.initWorkingSetFromRepoState(ctx)
+		} else if err != nil {
 			dEnv.RSLoadErr = err
 		}
 	}
--- a/integration-tests/bats/deleted-branches.bats
+++ b/integration-tests/bats/deleted-branches.bats
@@ -24,7 +24,11 @@ make_it() {
 @test "deleted-branches: can checkout existing branch after checked out branch is deleted" {
     make_it
 
-    dolt sql -q 'delete from dolt_branches where name = "main"'
+    run dolt sql -q 'delete from dolt_branches where name = "main"'
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "read-only" ]] || false
+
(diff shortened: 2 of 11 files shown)


integration-tests/bats/system-tables.bats
```diff
@@ -531,6 +531,7 @@ SQL
 @test "system-tables: cannot delete last branch in dolt_branches" {
```
timsehn, line 531: I would call this test "system-tables: dolt_branches is read-only".

Then I would add update and insert statement tests.

Then anywhere else there was a write to this table, just delete that part of the test and replace it with the equivalent `call dolt_branch()` call.

This makes the read-only test all in one place instead of spread out all over the code.

integration-tests/bats/deleted-branches.bats
```diff
@@ -24,7 +24,11 @@ make_it() {
 @test "deleted-branches: can checkout existing branch after checked out branch is deleted" {
     make_it
 
-    dolt sql -q 'delete from dolt_branches where name = "main"'
+    run dolt sql -q 'delete from dolt_branches where name = "main"'
```
timsehn, line 27: Delete these 3 lines.

timsehn (commented):
Some testing feedback.

integration-tests/bats/replication.bats
```diff
@@ -160,22 +160,6 @@ teardown() {
     [[ "$output" =~ "t1" ]] || false
 }
 
-@test "replication: push on branch table update" {
```
timsehn, line 163: I assume there's a test for creating a new branch with `call dolt_branch()`? That is why this is deleted?

timsehn (commented):
Another testing comment :-)

integration-tests/bats/replication.bats
jennifersp: `UPDATE dolt_branches ...` query can only be replicated with `CALL DOLT_COMMIT()` in this test. There is already a test below that tests this test with `CALL DOLT_COMMIT()`.

timsehn: Then tests look good to me :-)

fulghum (approved):
Fixes look good. Nice work! 🎸

We should add a test that shows customers can't get into a state where they delete all the branches and can't get back into their database. I *think* that'll mean preventing users from deleting the last branch. If these changes are flexible enough to prevent that now, it would be great to add in that test now, otherwise we can handle that as a separate PR. I just think it's important that we use this opportunity to think about other similar ways customers could get into a bad state and proactively address them. 

It would also be nice to add a test for SQL renaming a branch that the CLI has checked out, so we can get some test coverage over that code path.

go/libraries/doltcore/env/environment.go
```diff
@@ -175,11 +175,8 @@ func loadWithFormat(ctx context.Context, hdp HomeDirProvider, fs filesys.Filesys
 	if rsErr == nil && dbLoadErr == nil && dbFormatErr == nil {
 		// If the working set isn't present in the DB, create it from the repo state. This step can be removed post 1.0.
 		_, err := dEnv.WorkingSet(ctx)
-		if err == doltdb.ErrWorkingSetNotFound {
-			err := dEnv.initWorkingSetFromRepoState(ctx)
-			if err != nil {
```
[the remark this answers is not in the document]
jennifersp: This was the place were we were getting the error when we wanted to use any dolt command after deleting current branch on CLI. This code tries to init the working set from RepoState if the current working set is invalid/not found. This initially was for moving or changing the initial the `.dolt/` directory into different `./dolt` directory that would have different working set. In that case this code will init the working set correctly to new one, otherwise, the working set is not found which is the case of current head is deleted (I set the RepoState head to empty branch name, which is stored as `ref/heads/` ). This means the working set will be set if it finds any, if not it will be empty.

integration-tests/bats/deleted-branches.bats
```diff
@@ -24,7 +24,7 @@ make_it() {
 @test "deleted-branches: can checkout existing branch after checked out branch is deleted" {
     make_it
 
-    dolt sql -q 'delete from dolt_branches where name = "main"'
+    dolt sql -q 'call dolt_checkout("to_keep"); call dolt_branch("-D", "main");'
```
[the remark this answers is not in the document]
jennifersp: Ah it'll give error since it'd try to delete the current branch on SQL. I'll update the test title for CLI head specifically

integration-tests/bats/deleted-branches.bats
```diff
@@ -93,7 +93,8 @@ make_it() {
 
     start_sql_server "dolt_repo_$$"
 
-    server_query "dolt_repo_$$"  1 dolt "" 'delete from dolt_branches where name = "main"' ""
+    server_query "dolt_repo_$$"  1 dolt "" 'call dolt_checkout("to_keep"); call dolt_branch("-D", "main");' ""
+
```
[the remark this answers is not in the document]
jennifersp: They still throw an error, and they'll be resolved in separate PR :)

PR #4258, opened 2022-09-01
[store/prolly] Add DiffKeyRangeMaps and IterKeyRange

druvv:


go/store/prolly/map.go
```diff
@@ -124,10 +124,19 @@ func DiffMaps(ctx context.Context, from, to Map, cb DiffFn) error {
 	return diffOrderedTrees(ctx, from.tuples, to.tuples, cb)
 }
 
+// RangeDiffMaps returns diffs within a Range. See Range for which diffs are
+// returned.
 func RangeDiffMaps(ctx context.Context, from, to Map, rng Range, cb DiffFn) error {
 	return rangeDiffOrderedTrees(ctx, from.tuples, to.tuples, rng, cb)
 }
```
andrew-wm-arthur, line 131: Do we have a use case for this old function, given the new one?

go/store/prolly/map.go
```diff
@@ -124,10 +124,19 @@ func DiffMaps(ctx context.Context, from, to Map, cb DiffFn) error {
...
 
+// DiffKeyRangeMaps returns diffs within a physical key range. The key range is
+// specified by |startInclusive| and |stopExclusive|. If |startInclusive| or
+// |stopExclusive| is null, then the range is unbounded towards that end.
+func DiffKeyRangeMaps(ctx context.Context, from, to Map, startInclusive, stopExclusive val.Tuple, cb DiffFn) error {
```
andrew-wm-arthur, line 136:
```suggestion
func DiffMapKeyRanges(ctx context.Context, from, to Map, start, stop val.Tuple, cb DiffFn) error {
```

go/store/prolly/map.go
```diff
@@ -233,6 +242,13 @@ func (m Map) IterRange(ctx context.Context, rng Range) (MapIter, error) {
 	return filteredIter{iter: iter, rng: rng}, nil
 }
 
+// IterKeyRange iterates over a physical key range defined by |startInclusive|
+// and |stopExclusive|. If |startInclusive| and/or |stopExclusive| is nil, the
+// range will be open towards that end.
+func (m Map) IterKeyRange(ctx context.Context, startInclusive, stopExclusive val.Tuple) (MapIter, error) {
```
andrew-wm-arthur, line 248:
```suggestion
func (m Map) IterKeyRange(ctx context.Context, start, stop val.Tuple) (MapIter, error) {
```

andrew-wm-arthur (approved):
LGTM, couple comments on names

go/store/prolly/map.go
druvv: I believe we do. We could implement range diffs in the sql layer for our diff tables instead of applying a filter.

Issue #2261, opened 2021-10-14
Support VARCHAR(MAX)


noamross:
Currently the maximum size for a VARCHAR is 16383 and one needs to specify VARCHAR(16383) for it. It would be nice to support the VARCHAR(MAX) syntax so one doesn't have to remember the number.

timsehn:
This is a good first issue.

rafimuhammad01:
Hi @noamross @timsehn I just started to contribute in open source project and I choose this issue. 

I have created PR for this issue, my approach is to check manually if the query contains VARCHAR(MAX) then I replace it with the real value which is 16383. I also implementing this in new function called validateQuery so it will be easier if we want to add more validation to the query.

Please review my PR. Thank you so much!

rafimuhammad01:
Hi @noamross @timsehn I just look to the code and I found out that the error occur when the query is being parsed by go yacc (I don't really know about go yacc). Since you said that I need to make implementation on [go-mysql-server](https://github.com/dolthub/go-mysql-server) , I wonder about the flow of this program, Did it parse the query first in go yacc then go to the go-mysql-server or it is being process first in go-mysql-server then parse it with go yacc?

If the flow is parsing the query first with go yacc, I think I will not be able to make the implementation on go-mysql-server since the error is occur when the query being parsed with go yacc.

```go
// github.com/dolthub/vitess/go/vt/sqlparser/ast.go
func parseTokenizer(sql string, tokenizer *Tokenizer) (Statement, error) {
	if yyParsePooled(tokenizer) != 0 {
		if se, ok := tokenizer.LastError.(vterrors.SyntaxError); ok {
			return nil, vterrors.NewWithCause(vtrpcpb.Code_INVALID_ARGUMENT, tokenizer.LastError.Error(), se)
		} else {
			return nil, vterrors.New(vtrpcpb.Code_INVALID_ARGUMENT, tokenizer.LastError.Error())
		}
	}
	if tokenizer.ParseTree == nil {
		return nil, ErrEmpty
	}
	captureSelectExpressions(sql, tokenizer)
	adjustSubstatementPositions(sql, tokenizer)
	return tokenizer.ParseTree, nil
}
```
This is the code when the error occur, and I actually a bit confuse about the process, can you help me to explain it? 

Thank you!

rafimuhammad01:
Update:

I just found out the flow, that is:
1. receive client command (in this case `create table people (id int primary key, age int, first_name varchar(80), is_married bit)`)
[...]

PR #4261, opened 2022-09-02
Support VARCHAR(MAX)

rafimuhammad01:
Parse MAX in varchar(MAX) to its real value which is 16383. 

related issue #2261

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -401,6 +401,11 @@ func queryMode(
 
 	_, continueOnError := apr.GetValue(continueFlag)
 
+	query, verr := validateQuery(query)
+	if verr != nil {
+		return HandleVErrAndExitCode(verr, usage)
+	}
+
 	if saveName != "" {
 		verr := execQuery(ctx, mrEnv, query, format, config)
 		if verr != nil {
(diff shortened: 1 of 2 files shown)


timsehn (commented):
Thank you so much for the contribution. I actually think this needs to go in [go-mysql-server](https://github.com/dolthub/go-mysql-server) as it is functionality we would want to work across all integrators.

I'm not exactly certain where this code lives there but @zachmu or @fulghum can point you there.

Here's a guide for how to contribute/test go-mysql-server:

https://docs.dolthub.com/guides/contributing/go-mysql-server

fulghum:
Hey @rafimuhammad01, thank you for taking the time to send us this contribution! This will be a great feature to support. 🙏 
 I think the general approach of looking for `MAX` and replacing it with the max value is good, and I agree with Tim that if we push this lower, into go-mysql-server, then it'll work much more broadly. 

Take a look at this method in go-mysql-server that creates a string type: 
https://github.com/dolthub/go-mysql-server/blob/main/sql/stringtype.go#L97

That function takes in an integer length for the new string type, so we'd want to transform `MAX` right before we get into that code. Take a look at where that function is getting called and see if you can find where the parser calls it as it parses a column definition. Let us know if you get stuck and we're happy to dig in deeper and find a more exact spot for this code.

rafimuhammad01:
Hi @timsehn @fulghum thanks for the feedback.

I agree that it is better to implementing on go-mysql-server. Will check the code and submit my PR soon. Cheers! 🍻

timsehn:
Cool. I'll close this one :-)

PR #4265, opened 2022-09-02
unskip prepareds for dolt diff table function

JCOR11599:


JCOR11599:
changes are in https://github.com/dolthub/dolt/pull/4252

PR #4267, opened 2022-09-05
Add dolt diff --skinny flag

### LINKED ISSUE #790: `dolt diff --skinny`: Only show primary keys and columns with changes. Hide other columns.
During our talk for CMU, the host suggested a diff view where we hide columns that did not change except the primary key. Good idea. Let's do it.

tanyongzhi:
Closes #790 

Adds a `--skinny` flag to `dolt diff`, which only shows primary keys and columns with changes.

 Note that `writeDiffResults` has been split into 2 functions - `writeFilteredRows` and `getDiffRows`. The former is responsible for printing the rows, while the latter function collects the relevant columns, depending on if the `--skinny` flag is active.

Also note that in the event of row addition/deletions, all columns will be included when the `--skinny` flag is active.

### DIFF
--- a/go/cmd/dolt/commands/diff.go
+++ b/go/cmd/dolt/commands/diff.go
@@ -558,7 +558,6 @@ func diffRows(
 	if err != nil {
 		return errhand.VerboseErrorFromError(err)
 	}
-	defer rowWriter.Close(ctx)
 
 	// can't diff
 	if !diffable {
@@ -618,9 +617,11 @@ func diffRows(
 	}
 
 	defer rowIter.Close(sqlCtx)
+	defer rowWriter.Close(ctx)
 
+	var modifiedColNames map[string]bool
 	if dArgs.skinny {
-		modifiedColNames, err := getModifiedCols(sqlCtx, rowIter, unionSch, sch)
+		modifiedColNames, err = getModifiedCols(sqlCtx, rowIter, unionSch, sch)
 		if err != nil {
 			return errhand.BuildDError("Error running diff query:\n%s", query).AddCause(err).Build()
 		}
@@ -629,7 +630,7 @@ func diffRows(
 		var filteredUnionSch sql.Schema
 		for _, s := range unionSch {
 			for colName := range modifiedColNames {
-				if strings.Contains(s.Name, colName) {
+				if s.Name == colName {
 					filteredUnionSch = append(filteredUnionSch, s)
 				}
 			}
@@ -643,17 +644,20 @@ func diffRows(
 		defer rowWriter.Close(ctx)
 
 		// reset the row iterator
-		rowIter.Close(sqlCtx)
+		err = rowIter.Close(sqlCtx)
+		if err != nil {
+			return errhand.BuildDError("Error closing row iterator:\n%s", query).AddCause(err).Build()
+		}
 		_, rowIter, err = se.Query(sqlCtx, query)
+		defer rowIter.Close(sqlCtx)
 		if sql.ErrSyntaxError.Is(err) {
 			return errhand.BuildDError("Failed to parse diff query. Invalid where clause?\nDiff query: %s", query).AddCause(err).Build()
 		} else if err != nil {
 			return errhand.BuildDError("Error running diff query:\n%s", query).AddCause(err).Build()
 		}
-		defer rowIter.Close(sqlCtx)
 	}
 
-	err = writeDiffResults(sqlCtx, sch, unionSch, rowIter, rowWriter, dArgs.skinny)
+	err = writeDiffResults(sqlCtx, sch, unionSch, rowIter, rowWriter, modifiedColNames, dArgs.skinny)
 	if err != nil {
 		return errhand.BuildDError("Error running diff query:\n%s", query).AddCause(err).Build()
 	}
@@ -697,6 +701,7 @@ func writeDiffResults(
 	targetSch sql.Schema,
 	iter sql.RowIter,
 	writer diff.SqlRowDiffWriter,
+	modifiedColNames map[string]bool,
 	filterChangedCols bool,
 ) error {
 	ds, err := newDiffSplitter(diffQuerySch, targetSch)
@@ -719,15 +724,19 @@ func writeDiffResults(
 
 		if filterChangedCols {
 			var filteredOldRow, filteredNewRow rowDiff
-			for i, changeType := range oldRow.colDiffs {
-				if changeType != diff.None || targetSch[i].PrimaryKey {
-					filteredOldRow.row = append(filteredOldRow.row, oldRow.row[i])
-					filteredOldRow.colDiffs = append(filteredOldRow.colDiffs, oldRow.colDiffs[i])
-					filteredOldRow.rowDiff = oldRow.rowDiff
+			for i, changeType := range newRow.colDiffs {
+				if (changeType == diff.Added|diff.Removed) || modifiedColNames[targetSch[i].Name] {
+					if i < len(oldRow.row) {
+						filteredOldRow.row = append(filteredOldRow.row, oldRow.row[i])
+						filteredOldRow.colDiffs = append(filteredOldRow.colDiffs, oldRow.colDiffs[i])
+						filteredOldRow.rowDiff = oldRow.rowDiff
+					}
 
-					filteredNewRow.row = append(filteredNewRow.row, newRow.row[i])
-					filteredNewRow.colDiffs = append(filteredNewRow.colDiffs, newRow.colDiffs[i])
-					filteredNewRow.rowDiff = newRow.rowDiff
+					if i < len(newRow.row) {
+						filteredNewRow.row = append(filteredNewRow.row, newRow.row[i])
+						filteredNewRow.colDiffs = append(filteredNewRow.colDiffs, newRow.colDiffs[i])
+						filteredNewRow.rowDiff = newRow.rowDiff
+					}
 				}
 			}
 
@@ -751,6 +760,10 @@ func writeDiffResults(
 	}
 }
 
+// getModifiedCols returns a set of the names of columns that are modified, as well as the name of the primary key for a particular row iterator and schema.
+// In the case where rows are added or removed, all columns will be included
+// unionSch refers to a joint schema between the schema before and after any schema changes pertaining to the diff,
+// while diffQuerySch refers to the schema returned by the "dolt_diff" sql query.
 func getModifiedCols(
 	ctx *sql.Context,
 	iter sql.RowIter,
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/commands/diff.go
```diff
@@ -690,4 +766,6 @@ func writeDiffResults(
 			}
 		}
 	}
+
+	return writer.Close(ctx)
```
fulghum, line 770: (minor) might be just a little bit cleaner to have the `writer` closed in the function that creates it, instead of that being a side effect of `writeFilteredResults`. Still correct either way, but makes the lifetime of the `writer` resource a little easier to understand. I'd just use `defer writer.Close(ctx)` in the main function code that's calling this method.

go/cmd/dolt/commands/diff.go
```diff
@@ -651,30 +689,68 @@ func getColumnNamesString(fromSch, toSch schema.Schema) string {
...
+					filteredOldRow.colDiffs = append(filteredOldRow.colDiffs, oldRow.colDiffs[i])
+					filteredOldRow.rowDiff = oldRow.rowDiff
+
+					filteredNewRow.row = append(filteredNewRow.row, newRow.row[i])
+					filteredNewRow.colDiffs = append(filteredNewRow.colDiffs, newRow.colDiffs[i])
+					filteredNewRow.rowDiff = newRow.rowDiff
+				}
+			}
+
+			oldRows = append(oldRows, filteredOldRow)
```
fulghum, line 734: How hard would it be to keep this code working from an iterator and not having to load all diff results into memory at once? 

I think it would be ideal if the code could traverse the iterator once to figure out the changed columns and the right schema for the filtered view, then reset the iterator to traverse again and apply any filtering. The benefit is of course just that we don't have to load all the diff results into memory. That's fine for small diffs, but we do have to deal with very large diffs as well, and it seems like we could tweak this code just a little more to be more efficient there. Let me know what you think.

go/cmd/dolt/commands/diff.go
```diff
@@ -60,6 +60,7 @@ const (
 	limitParam  = "limit"
 	SQLFlag     = "sql"
 	CachedFlag  = "cached"
+	SkinnyFlag  = "skinny"
```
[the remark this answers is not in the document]
timsehn: Maybe both? I like skinny.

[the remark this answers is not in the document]
fulghum: Cool, cool; let's stick with `skinny` then 👍  always easy to add a synonym later if we want to.

go/cmd/dolt/commands/diff.go
tanyongzhi: That makes sense! I'll implement the changes in a follow-up commit.

go/cmd/dolt/commands/diff.go
```diff
@@ -613,7 +619,41 @@ func diffRows(
...
+		// reset the row iterator
+		rowIter.Close(sqlCtx)
+		_, rowIter, err = se.Query(sqlCtx, query)
+		if sql.ErrSyntaxError.Is(err) {
+			return errhand.BuildDError("Failed to parse diff query. Invalid where clause?\nDiff query: %s", query).AddCause(err).Build()
+		} else if err != nil {
+			return errhand.BuildDError("Error running diff query:\n%s", query).AddCause(err).Build()
+		}
+		defer rowIter.Close(sqlCtx)
+	}
```
tanyongzhi, line 654: Not sure if this is the correct way to reset the row iterator. Any feedback would be greatly appreciated!

go/cmd/dolt/commands/diff.go
```diff
@@ -691,3 +750,36 @@ func writeDiffResults(
 		}
 	}
 }
+
+func getModifiedCols(
```
fulghum, line 754: (minor) some simple godocs for this function would be nice, just to help future code readers understand the intent and any behavior quirks quickly. Explaining the `unionSch` and `diffQuerySch` params would probably be helpful for future readers, too.

go/cmd/dolt/commands/diff.go
```diff
@@ -613,7 +619,41 @@ func diffRows(
...
 
-	err = writeDiffResults(sqlCtx, sch, unionSch, rowIter, rowWriter)
+	if dArgs.skinny {
+		modifiedColNames, err := getModifiedCols(sqlCtx, rowIter, unionSch, sch)
+		if err != nil {
+			return errhand.BuildDError("Error running diff query:\n%s", query).AddCause(err).Build()
+		}
+
+		// instantiate a new schema that only contains the columns with changes
+		var filteredUnionSch sql.Schema
```
fulghum, line 629: (minor/suggestion) It looks like you could put the conversion from column names to a new schema in the `getModifiedCols` function. Might be worth pushing that logic into the function (and probably tweaking the name) to keep this main section of code a little tighter/shorter. Not a big deal either way though, just a suggestion to consider.

go/cmd/dolt/commands/diff.go
```diff
@@ -613,7 +619,41 @@ func diffRows(
...
+
+		// instantiate a new RowWriter with the new schema that only contains the columns with changes
+		rowWriter, err = dw.RowWriter(ctx, td, filteredUnionSch)
+		if err != nil {
+			return errhand.VerboseErrorFromError(err)
+		}
+		defer rowWriter.Close(ctx)
+
+		// reset the row iterator
+		rowIter.Close(sqlCtx)
```
fulghum, line 646: (minor) would be good to check any returned `err` here. I think `rowIter.Close` returns an error, but disregard if not.

go/cmd/dolt/commands/diff.go
fulghum: yup, this looks great! 

Would be good to move that `defer` up to line 648 to make sure that `Close` call always gets registered to be run. Currently, it won't run if any of that error code above it returns from the function. Not a huge deal for correctness, but definitely good practice to tidy up.

integration-tests/bats/diff.bats
```diff
@@ -720,6 +720,24 @@ SQL
     [[ "$output" = 'UPDATE `t` SET `val1`=30,`val3`=4 WHERE `pk`=1;' ]] || false
 }
 
+@test "diff: skinny flag only shows row changed" {
```
fulghum, line 723: It would be good to add tests for some other types of more advanced changes...

For example, when I tested altering the schema to add a column and then update some of the data in the table, I didn't see the results I expected.  When adding a column with all NULL values, `dolt diff` should show only the schema changes and not report any data changes. This worked with `dolt diff --skinny`, but if there were any real data changes, those were not shown. 

```sql
❯ dolt diff --data
diff --dolt a/t b/t
--- a/t @ nfgsd9f16fb4gt1ismhqjv74it1atri5
+++ b/t @ 8h3s3nvj1kjhe2lofpsq59knldv372m4
+---+----+----+--------+------+
|   | n1 | n2 | s      | s2   |
+---+----+----+--------+------+
| < | 2  | 2  | two    | NULL |
| > | 2  | 2  | two    | wha? |
| < | 3  | 3  | three  | NULL |
| > | 3  | 3  | purple | NULL |
+---+----+----+--------+------+

❯ dolt diff --data --skinny
diff --dolt a/t b/t
--- a/t @ nfgsd9f16fb4gt1ismhqjv74it1atri5
+++ b/t @ 8h3s3nvj1kjhe2lofpsq59knldv372m4
```

In the example above, I would expect the `--skinny` output to include the same output as `dolt diff --data` is showing above. 

I would add some tests for covering adding columns and removing columns, mixed in with updating data in the table to make sure we're testing some more non-basic cases. You might also consider other updates like inserting and deleting rows, just to round the tests out.

fulghum (commented):
Thanks for iterating on this @tanyongzhi! Code is looking good and I only had some minor comments. Nice work switching to an iterator approach so that we don't have to load everything into memory. That'll be really helpful for customers working with large diffs. 

I built the changes locally and played with it and noticed a couple non-basic diff cases didn't work as expected. I left some more notes with some concrete examples, but let me know if anything doesn't make sense. It should be straightforward to add some more BATS tests for those cases and hopefully that'll help debug through what's going on. Let us know if you need any help or hints of course. 

This is a cool feature and I'm excited to see it get released!

go/cmd/dolt/commands/diff.go
tanyongzhi: Hmm, I think I'll prefer to keep the code that converts the new schema in `diffRows` to keep `getModifiedCols` as simple as possible (ie. so that the function only does 1 thing). Let me know if you feel strongly about this!

go/cmd/dolt/commands/diff.go
fulghum: Beautiful docs! Thank you @tanyongzhi!

integration-tests/bats/diff.bats
```diff
@@ -720,6 +720,107 @@ SQL
...
+    dolt sql -q "UPDATE t SET val1=2 WHERE pk=1"
+    dolt commit -am "cm2"
+
+    dolt sql -q "UPDATE t SET val1=3 WHERE pk=1"
+    dolt commit -am "cm3"
+
+    run dolt diff --skinny HEAD~1
+    [ $status -eq 0 ]
+    [[ ! "$output" =~ 'val2' ]] || false
+    [[ "$output" =~ 'val1' ]] || false
```
fulghum, line 744: Probably good to assert that the primary key is also still included: 

```suggestion
    [[ "$output" =~ 'pk' ]] || false
    [[ "$output" =~ 'val1' ]] || false
```

integration-tests/bats/diff.bats
```diff
@@ -720,6 +720,107 @@ SQL
...
+    [[ "$output" =~ 's' ]] || false
+
+    dolt sql -q "ALTER TABLE t DROP COLUMN s"
+    dolt sql -q "UPDATE t SET val1=3 WHERE pk=1"
+    dolt sql -q "UPDATE t SET val1=4 WHERE pk=2"
+    dolt commit -am "cm2"
+
+    test=$(dolt diff --skinny --data HEAD~1)
+    echo $test
```
fulghum, line 792: These two testing lines can be pulled out. 

```suggestion
```

integration-tests/bats/diff.bats
```diff
@@ -720,6 +720,107 @@ SQL
...
+    run dolt diff --skinny --data HEAD~1
+    [ $status -eq 0 ]
+    [[ "$output" =~ 'pk' ]] || false
+    [[ "$output" =~ 'val1' ]] || false
+    [[ "$output" =~ 's' ]] || false
+    [[ ! "$output" =~ 'val3' ]] || false
```
fulghum, line 798: I'm not sure this line is needed, since `val3` isn't part of this test. 

```suggestion
```

fulghum (commented):
Hey @tanyongzhi, this looks really good. 🎉  Thank you for contributing this new feature! 🙏 

I just had a couple really minor comments about the tests. I'll go ahead and accept those suggestions and kick off tests to make sure they pass. After that, I'll give this one more look and approve/merge.

fulghum (approved):
Code looks good and all tests are passing! 🎉 

Thank you for this contribution @tanyongzhi! 🙏 I'll go ahead and get it merged in.

PR #4268, opened 2022-09-05
Feat/support aliyun oss store

xujihui1985:
support aliyun oss as backend store


xujihui1985:
@zachmu Hi, this PR is to add support for aliyun oss as backend store same as gcs and aws s3, dolt is an amazing project, and hope I can do some contribution for it.

go/libraries/doltcore/dbfactory/oss.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+)
+
+const (
+	ossEndpointEnvKey        = "OSS_ENDPOINT"
+	ossAccessKeyIDEnvKey     = "OSS_ACCESS_KEY_ID"
+	ossAccessKeySecretEnvKey = "OSS_ACCESS_KEY_SECRET"
+)
+
+// OSSFactory is a DBFactory implementation for creating GCS backed databases
```
bheni, line 24: Update comments on these methods so they don't say GCS

go/store/blobstore/oss.go
```diff
@@ -0,0 +1,127 @@
...
+	var options []oss.Option
+	if expectedVersion != "" {
+		options = append(options, oss.VersionId(expectedVersion))
+	}
+	var meta http.Header
+	options = append(options, oss.GetResponseHeader(&meta))
+	if err := ob.bucket.PutObject(ob.absKey(key), reader, options...); err != nil {
+		ossErr, ok := err.(oss.ServiceError)
+		if ok {
+			return "", CheckAndPutError{key, expectedVersion, fmt.Sprintf("unknown (OSS error code %d)", ossErr.StatusCode)}
```
bheni, line 94: Prefer struct initialization to use field names as opposed to positional arguments.

go/libraries/doltcore/dbfactory/oss.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/dolthub/dolt/go/store/prolly/tree"
+	"github.com/dolthub/dolt/go/store/types"
+	"net/url"
+	"os"
+)
+
+const (
+	ossEndpointEnvKey        = "OSS_ENDPOINT"
+	ossAccessKeyIDEnvKey     = "OSS_ACCESS_KEY_ID"
+	ossAccessKeySecretEnvKey = "OSS_ACCESS_KEY_SECRET"
```
bheni, line 21: If you use environment variables it means that dolt can only have a single oss remote.  It would be better if the user was able to setup multiple oss remotes.

bheni (commented):
Looks pretty good.  My only real issue here is that the environment variables limit the oss remotes that can be added.  All oss CreateDB calls will use the same endpoint, key_id and key_secret.

xujihui1985:
> Looks pretty good. My only real issue here is that the environment variables limit the oss remotes that can be added. All oss CreateDB calls will use the same endpoint, key_id and key_secret.

@bheni yes, thanks for reviewing, you are right, because aliyun oss doesn't support config file with different profile, I would like to go with aws strategy that connect to  oss with a config file with multiple profile configured, for example

```
[default]
endpoint=<endpoint>
accessKeyID=<id>
accessKeySecret=<secret>

[prod-profile]
endpoint=<endpoint>
accessKeyID=<id>
accessKeySecret=<secret>
```
the default credentials file will locate at $HOME/.oss/dolt_oss_credentials,  environment variable will still be supported for container environment. 

The priority order is
environment variable > user provide profile > default profile

dolt remote add --oss-creds-profile prod-profile --oss-creds-config <credentials filepath> origin oss://[bucket]/key

what do you think?

bheni:
@xujihui1985 I think that works. If customers show up and have issues with it we can look at other solutions.

xujihui1985:
ok, I'm working on it

xujihui1985:
> @xujihui1985 I think that works. If customers show up and have issues with it we can look at other solutions.
@bheni  It's busy last week, I implement oss store with oss cred file config, please have a look

xujihui1985:
@bheni Hi bheni, would you like to take a look at this PR and give some advice?

bheni (approved):
LGTM.  I can't really test out the oss specific code.  If oss users have issues I'll reach out to you.

xujihui1985:
> LGTM. I can't really test out the oss specific code. If oss users have issues I'll reach out to you.

of course, I'm glad to be a part of this project

fulghum:
Thank you for the contribution @xujihui1985!! 🙏 Very appreciated!

Test run looked good, so I went ahead and merged this in to main. This will go out with the next release of Dolt, later this week. 🎉

PR #4275, opened 2022-09-06
More point lookup optimizations

max-hoffman:
Use new `IndexLookup.IsPointLookup` to avoid duplicate work building range tuples.

companion PR: https://github.com/dolthub/go-mysql-server/pull/1239


max-hoffman:
#benchmark

max-hoffman:
#benchmark

go/libraries/doltcore/sqle/index/index_lookup.go
```diff
@@ -115,15 +116,63 @@ func NewRangePartitionIter(ctx *sql.Context, t DoltTableable, lookup sql.IndexLo
...
+		}
+		if err = PutField(ctx, ns, tb, j, v); err != nil {
+			return nil, err
+		}
+	}
+	tup := tb.BuildPermissive(sharePool)
+	return &pointPartitionIter{r: prolly.Range{Tup: tup, Desc: tb.Desc}}, nil
+}
+
+type pointPartitionIter struct {
```
andrew-wm-arthur, line 140: for every secondary lookup in a join we go from `sql.IndexLookup` -> `sql.PartitionIter` -> `sql.Partiiton` -> `sql.RowIter`. Is there any way to collapse that indirection, even if we keep the existing interfaces? Specifically I'm wondering if `pointPartitionIter` and `pointPartition` could be merged into one type

andrew-wm-arthur (approved):
LGTM!

go/libraries/doltcore/sqle/index/index_lookup.go
max-hoffman: Good call, that squeezed an extra 5% out of the join bench.

PR #4277, opened 2022-09-06
Adds `dolt_merge_status` system table

druvv:
Closes https://github.com/dolthub/dolt/issues/3504

The `dolt_merge_status` system table tells a user if a merge is active. It has the following schema:
```sql
CREATE TABLE `dolt_merge_status` (
-- Whether a merge is currently active or not
  `is_merging` tinyint NOT NULL, 
 -- The commit spec that was used to initiate the merge
  `source` text,
-- The commit that the commit spec resolved to at the time of merge
  `source_commit` text, 
-- The target destination working set
  `target` text, 
-- A list of tables that have conflicts or constraint violations
  `unmerged_tables` text 
)
```

For example, lets create a simple conflict:
```bash
dolt sql -q "CREATE TABLE t (a INT PRIMARY KEY, b INT);"
dolt add .
dolt commit -am "base"

dolt checkout -b right
dolt sql <<SQL
ALTER TABLE t ADD c INT;
INSERT INTO t VALUES (1, 2, 1);
SQL
dolt commit -am "right"

dolt checkout main
dolt sql -q "INSERT INTO t values (1, 3);"
dolt commit -am "left"

dolt merge right
```

Output of `SELECT * from dolt_merge_status;`:
```
+------------+--------+----------------------------------+-----------------+-----------------+
| is_merging | source | source_commit                    | target          | unmerged_tables |
+------------+--------+----------------------------------+-----------------+-----------------+
| true       | right  | fbghslue1k9cfgbi00ti4r8417frgbca | refs/heads/main | t               |
+------------+--------+----------------------------------+-----------------+-----------------+
```

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/merge_status_table.go
+++ b/go/libraries/doltcore/sqle/dtables/merge_status_table.go
@@ -1,0 +1,163 @@
+// Copyright 2022 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dtables
+
+import (
+	"context"
+	"io"
+	"strings"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
+	"github.com/dolthub/dolt/go/libraries/utils/set"
+	"github.com/dolthub/go-mysql-server/sql"
+)
+
+// MergeStatusTable is a sql.Table implementation that implements a system table
+// which shows information about an active merge.
+type MergeStatusTable struct {
+	dbName string
+}
+
+func (s MergeStatusTable) Name() string {
+	return doltdb.MergeStatusTableName
+}
+
+func (s MergeStatusTable) String() string {
+	return doltdb.MergeStatusTableName
+}
+
+func (s MergeStatusTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{Name: "is_merging", Type: sql.Boolean, Source: doltdb.MergeStatusTableName, PrimaryKey: false, Nullable: false},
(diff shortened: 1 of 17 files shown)


go/libraries/doltcore/sqle/dtables/merge_status_table.go
```diff
@@ -0,0 +1,164 @@
...
+func (s MergeStatusTable) Name() string {
+	return doltdb.MergeStatusTableName
+}
+
+func (s MergeStatusTable) String() string {
+	return doltdb.MergeStatusTableName
+}
+
+func (s MergeStatusTable) Schema() sql.Schema {
+	return []*sql.Column{
```
zachmu, line 45: Is this actually dolt_unmerged_tables, with one row per unmerged table?

No rows if no unmerged tables?

Get rid of `is_merging` and replace with a `merge_active()` function?

go/libraries/doltcore/sqle/dtables/merge_status_table.go
```diff
@@ -0,0 +1,164 @@
...
+		{Name: "is_merging", Type: sql.Boolean, Source: doltdb.MergeStatusTableName, PrimaryKey: false, Nullable: false},
+		{Name: "source", Type: sql.Text, Source: doltdb.MergeStatusTableName, PrimaryKey: false, Nullable: true},
+		{Name: "source_commit", Type: sql.Text, Source: doltdb.MergeStatusTableName, PrimaryKey: false, Nullable: true},
+		{Name: "target", Type: sql.Text, Source: doltdb.MergeStatusTableName, PrimaryKey: false, Nullable: true},
```
zachmu, line 49: Isn't this just `active_branch()`?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -1328,6 +1328,10 @@ var MergeScripts = []queries.ScriptTest{
 				Query:    "CALL DOLT_MERGE('feature-branch')",
 				Expected: []sql.Row{{1, 0}},
 			},
+			{
```
zachmu, line 1331: Need some tests with more than one table unmerged

zachmu (commented):
Some questions about this specific design. Seems like a row per unmerged table might be a bit more ergonomic?

go/libraries/doltcore/sqle/dtables/merge_status_table.go
druvv: A flat design could work here as well, but determining which tables are unmerged isn't the primary use case. The main reason we want a merge status function is to know if we need to commit a merge or if we need to abort the active one. 

The unmerged tables column was added for more similarity between `dolt status` and this new procedure.

go/libraries/doltcore/sqle/dtables/merge_status_table.go
druvv: Correct, target is just the active branch. Do you think we should name it differently than `target`?

PR #4283, opened 2022-09-07
fix enum panics in history table

JCOR11599:
Closes: https://github.com/dolthub/dolt/issues/4279

`sql.enumtype` contains a map, which in golang cannot be used in `==` operator


fulghum (approved):
Thanks for fixing this one James! ⚡

PR #4284, opened 2022-09-07
add nil check for string pointers during dolt table import

JCOR11599:
Closes: https://github.com/dolthub/dolt/issues/4209

### DIFF
--- a/integration-tests/bats/import-update-tables.bats
+++ b/integration-tests/bats/import-update-tables.bats
@@ -1223,3 +1223,20 @@ DELIM
     [[ "${lines[6]}" =~ "Lines skipped: 2" ]] || false
     [[ "${lines[7]}" =~ "Import completed successfully." ]] || false
 }
+
+@test "import-update-tables: test error when import bad csv with nulls" {
+    # Case where there are fewer values in a row than the number of columns in the schema
+    cat <<DELIM > bad-updates.csv
+i,j,k
+,,,
+DELIM
+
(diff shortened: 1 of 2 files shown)


integration-tests/bats/import-update-tables.bats
```diff
@@ -1223,3 +1223,20 @@ DELIM
...
+,,,
+DELIM
+
+    dolt sql -q "CREATE TABLE test(i int, j int, k int, l int)"
+
+    run dolt table import -u test bad-updates.csv
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "A bad row was encountered while moving data" ]] || false
+    [[ "$output" =~ "CSV reader expected 3 values, but saw 4" ]] || false
+    [[ "$output" =~ "row values:" ]]
```
fulghum, line 1240: do you need the `|| false` at the end of this one?

fulghum (approved):
Looks good! Thanks for fixing another panic! 🙏

integration-tests/bats/import-update-tables.bats
JCOR11599: thanks for catch, I definitely do

timsehn: As a bit of trivia. This is only necessary because of the ancient version of bash that ships with MacOS. Without it, that equality may show true when false non-deterministically.

fulghum: 🦇 🦇 🦇 @timsehn == BATSman 🦇 🦇 🦇

PR #4285, opened 2022-09-07
trim byte order mark present in UTF-8 encoded files

JCOR11599:
Closes: https://github.com/dolthub/dolt/issues/3790


druvv (approved):
lgtm if the file i put in the ticket works. That file is from Excel on Windows.

PR #4296, opened 2022-09-08
add `--author` flag to merge

jennifersp:


go/cmd/dolt/commands/merge.go
```diff
@@ -148,9 +148,14 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
 				return handleCommitErr(ctx, dEnv, err, usage)
 			}
 
-			name, email, err := env.GetNameAndEmail(dEnv.Config)
-			if err != nil {
-				return handleCommitErr(ctx, dEnv, err, usage)
+			var name, email string
+			if authorStr, ok := apr.GetValue(cli.AuthorParam); ok {
+				name, email, err = cli.ParseAuthor(authorStr)
```
fulghum, line 153: Looks like this `err` doesn't ever get checked? Could the err check block on line 156 get moved below this if/else statement so it checks the err for both branches?

fulghum (approved):
Nailed it! Nice work @jennifersp! ⚡️

PR #4301, opened 2022-09-09
have `TempTableFilesDir()` return err if `.dolt` is not found

jennifersp:
depends on https://github.com/dolthub/go-mysql-server/pull/1256
If `.dolt` dir is removed during runinng sql-server. Any connections will fail and it will cause the server to stop.


timsehn:
This was bigger than I expected :-)

go/cmd/dolt/commands/roots.go
```diff
@@ -85,7 +85,13 @@ func (cmd RootsCmd) Exec(ctx context.Context, commandStr string, args []string,
 	help, _ := cli.HelpAndUsagePrinters(cli.CommandDocsForCommandString(commandStr, cli.CommandDocumentationContent{}, ap))
 	apr := cli.ParseArgsOrDie(ap, args, help)
 
-	dir := filepath.Join(dEnv.GetDoltDir(), dbfactory.DataDir)
+	doltDir := dEnv.GetDoltDir()
+	if doltDir == "" {
+		cli.Println(color.YellowString("can no longer find a database on disk"))
```
max-hoffman, line 90: specifying the database name might be helpful

go/libraries/doltcore/env/environment.go
```diff
@@ -1115,8 +1129,12 @@ func (dEnv *DoltEnv) GetUserHomeDir() (string, error) {
...
 }
 
-func (dEnv *DoltEnv) TempTableFilesDir() string {
-	return mustAbs(dEnv, dEnv.GetDoltDir(), tempTablesDir)
+func (dEnv *DoltEnv) TempTableFilesDir() (string, error) {
+	doltDir := dEnv.GetDoltDir()
+	if doltDir == "" {
+		return "", ErrDoltRepositoryNotFound
+	}
+	return mustAbs(dEnv, doltDir, tempTablesDir), nil
```
max-hoffman, line 1137: it might make sense to just do the extra error check here, now that we're returning an error

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -905,7 +905,14 @@ func (d *DoltSession) AddDB(ctx *sql.Context, dbState InitialDbState) error {
 	// TODO: get rid of all repo state reader / writer stuff. Until we do, swap out the reader with one of our own, and
 	//  the writer with one that errors out
 	sessionState.dbData = dbState.DbData
-	sessionState.tmpFileDir = dbState.DbData.Rsw.TempTableFilesDir()
+	tmpDir, err := dbState.DbData.Rsw.TempTableFilesDir()
+	if err != nil {
+		if errors.Is(err, env.ErrDoltRepositoryNotFound) {
+			return env.ErrFailedToAccessDB
```
max-hoffman, line 911: wrap error maybe? seems like having the db name could be helpful

max-hoffman (approved):
LGTM, small comments

go/cmd/dolt/commands/roots.go
jennifersp: Should the name of the director the command is running be the database name?

go/libraries/doltcore/env/environment.go
jennifersp: I think it would be slower to check if the output of `mustAbs` method has substring of `.dolt` than to check if `doltDir == ""`

max-hoffman: i mean you can just put `return absPath, err := dEnv.FS.Abs(filepath.Join(path...))` there? no real reason to panic

jennifersp: Ah I see, makes sense.

PR #4310, opened 2022-09-12
Support Varchar(MAX)

rafimuhammad01:
related to #2261


fulghum (commented):
Thanks for the contribution @rafimuhammad01. If we can add a test to the enginetest test suite in go-mysql-server, that same test will run for go-mysql-server and for dolt (as well as any other implementors of go-mysql-server), so I'm comfortable relying on that coverage there and not adding it at this layer.

rafimuhammad01:
I see, will check the code and if it possible, I will close this PR. Thanks!

rafimuhammad01:
update: this PR is no needed anymore because as @fulghum suggested the test will be on go-mysql-server

PR #4315, opened 2022-09-12
create db when no current db is set should use the dbs available in session

jennifersp:
If there is no current database selected, use databases that are available within the session and use the common format for those databases.


andrew-wm-arthur (approved):
LGTM!

PR #4325, opened 2022-09-12
[no-release-notes] Refactoring: killed off last lingering uses of pipeline package

zachmu:


go/libraries/doltcore/table/untyped/tabular/formatter_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Dolthub, Inc.
+// Copyright 2022 Dolthub, Inc.
```
andrew-wm-arthur, line 1: don't rewrite these dates

andrew-wm-arthur (approved):
A thing of beauty

PR #4347, opened 2022-09-14
Add -A (--ALL) to `dolt commit` and `CALL DOLT_COMMIT`

### LINKED ISSUE #4309: Add `-A` option to `dolt commit` to add all new and modified tables
We recently changed `dolt commit -a` to match Git's behavior and *not* stage any new tables. It was convenient to have newly created tables added, but we thought it was more important to match Git's behavior exactly. As a compromise, we would like to add a new `-A` option to `dolt commit` that works in the same was as `dolt add -A` – it should stage all newly created and modified tables in the commit. 

This change should be made for the `dolt commit` command, as well as the `dolt_commit` SQL stored procedure.

druvv:
`-A` adds all tables (including new tables) to the staged set before committing.

Closes #4309

### DIFF
--- a/go/Godeps/LICENSES
+++ b/go/Godeps/LICENSES
@@ -4941,35 +4941,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ================================================================================
 
 ================================================================================
-= github.com/klauspost/cpuid/v2 licensed under: =
-
-The MIT License (MIT)
-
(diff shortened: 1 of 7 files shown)


fulghum (approved):
Nice work! We are all going to enjoy this one! 😁 

Looks like you need to increment `SkipPreparedsCount` to get that one, last check passing (none of the other dolt stored procedure tests add a prepared version). 

Doc update would be awesome, too. 😎

PR #4363, opened 2022-09-17
support dolt_diff_summary_table_function

jennifersp:
`dolt_diff_summary` table function is the same as CLI `dolt diff --summary` command. It takes fromCommit, toCommit and tableName, which is optional. 
If tableName is not defined, the table function will output all tables with data diff.
If the table does not have data diff, it will output empty result.
If the table is keyless, like CLI command, only row added and row deleted will have appropriate information, and the rest is nil.

Current issues:
- drop-column causes row values to be modified into NULL, it should be different from user-defined NULL or NULL added from add-column.
- for keyless table, update table causes delete old row and insert new row, which affects the row added and deleted result to be incorrect.


fulghum (commented):
Nice work! This is going to be a cool feature. I started off with some UX testing playing with the new function and seeing if I could break it 😉 I skimmed over the code quickly, too, and will do a deeper pass on it tomorrow, but wanted to share some initial feedback from testing it out... 

#### New table in working set 
When I had a new table in the working set, `dolt diff --summary` returned:
```
❯ dolt diff --summary
diff --dolt a/new b/new
added table
No data changes. See schema changes by using -s or --schema.
```
but `dolt_diff_summary` returned an error:
```
diffsummary> select * from dolt_diff_summary('WORKING', 'HEAD');
table not found: 
```

I got the same error after I committed that new table and ran dolt_diff_summary between HEAD and HEAD~1.  

btw... I eventually noticed that I had the from_commit and to_commit params backwards, but that shouldn't make a big difference since it should work both ways. It is an interesting UX issue that customers could make, too. We should be clear in the docs about which is the from_commit and which is the to_commit. I don't think we describe that well in the CLI docs for dolf diff today. 

I hit another weird issue with adding a new table in the working set after this, too... I had diff summarys showing up for multiple tables in the working set, then I created a new keyless table and dolt_diff_summary started returning an empty set. dolt diff --summary on the CLI still returned the right diff results though and when I added the third parameter for the other individual tables, they worked. 

If you get a chance to start on the docs for dolt_diff_summary, could you send me that, too? That would be awesome to review the code and the user docs at the same time.

go/libraries/doltcore/row/tagged_values.go
```diff
@@ -277,6 +277,8 @@ func CountCellDiffs(from, to types.Tuple) (uint64, error) {
 
 	for i, v := range f {
 		ov, ok := t[i]
+		// TODO : if the to has nil value, it can be drop column, so should not be cell change?
```
fulghum, line 280: Interesting! If this method had the to/from schemas passed in, would that be enough to differentiate between these two cases?

go/libraries/doltcore/schema/schema.go
```diff
@@ -233,6 +233,11 @@ func MapSchemaBasedOnTagAndName(inSch, outSch Schema) ([]int, []int, error) {
 	keyMapping := make([]int, inSch.GetPKCols().Size())
 	valMapping := make([]int, inSch.GetNonPKCols().Size())
 
+	// if inSch or outSch is empty schema. This can be from added or dropped table.
+	if (len(inSch.GetPKCols().cols) == 0 && len(inSch.GetNonPKCols().cols) == 0) || (len(outSch.GetPKCols().cols) == 0 && len(outSch.GetNonPKCols().cols) == 0) {
```
fulghum, line 237: Can this be simplified to...

```suggestion
	if len(inSch.GetAllCols().cols) == 0 || len(outSch.GetAllCols().cols) == 0 {
```

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+	&sql.Column{Name: "cells_added", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "cells_deleted", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "cells_modified", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "old_row_count", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "new_row_count", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "old_cell_count", Type: sql.Int64, Nullable: true},
+	&sql.Column{Name: "new_cell_count", Type: sql.Int64, Nullable: true},
+}
+
+// NewInstance implements the TableFunction interface
```
fulghum, line 57: (super minor) technically, not part of the TableFunction interface. Would be nice to put the standard "Create a new instance of ..."

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+
+// Children implements the sql.Node interface.
+func (ds *DiffSummaryTableFunction) Children() []sql.Node {
+	return nil
+}
+
+// WithChildren implements the sql.Node interface.
+func (ds *DiffSummaryTableFunction) WithChildren(children ...sql.Node) (sql.Node, error) {
+	if len(children) != 0 {
+		panic("unexpected children")
```
fulghum, line 117: It'd be safer to return an `error` here. That'll still stop query execution, but it won't make the whole go process stop, which is super important for sql-server. Generally, we should strongly avoid panics and prefer errors instead because of the danger of one query horking a running server.

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+	return ds, nil
+}
+
+// CheckPrivileges implements the interface sql.Node.
+func (ds *DiffSummaryTableFunction) CheckPrivileges(ctx *sql.Context, opChecker sql.PrivilegedOperationChecker) bool {
```
fulghum, line 123: Privs check for this one is another interesting area... I think you made a good choice by requiring select for the affected tables (which is potentially all user tables). 

This would be a really great point to mention in the function docs PR.

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+// RowIter implements the sql.Node interface
+func (ds *DiffSummaryTableFunction) RowIter(ctx *sql.Context, row sql.Row) (sql.RowIter, error) {
+	fromCommitVal, toCommitVal, tableName, err := ds.evaluateArguments()
+	if err != nil {
+		return nil, err
+	}
+
+	sqledb, ok := ds.database.(Database)
+	if !ok {
+		panic(fmt.Sprintf("unexpected database type: %T", ds.database))
```
fulghum, line 210: Turn this panic into an error, too. Even though it's something really unlikely (until we add a new type of database, which I'm sure we eventually will...) returning an error gets you the same effect you want (stopping execution and communicating the problem) without the danger of bringing down the multi-user sql-server.

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+			diffSummaries = append(diffSummaries, diffSum)
+		}
+	}
+
+	return NewDiffSummaryTableFunctionRowIter(diffSummaries), nil
+}
+
+// evaluateArguments returns fromCommitValStr, toCommitValStr and tableName.
+// It evaluates the argument expressions to turn them into values this DiffTableFunction
+// can use. Note that this method only evals the expressions, and doesn't validate the values.
```
fulghum, line 262: Wonderful function docs! Those string return params would be much harder for people to figure out without these nice docs. 🙌

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -0,0 +1,503 @@
...
+		acc.CellChanges += p.CellChanges
+		acc.NewRowSize += p.NewRowSize
+		acc.OldRowSize += p.OldRowSize
+		acc.NewCellSize += p.NewCellSize
+		acc.OldCellSize += p.OldCellSize
+
+		count++
+	}
+
+	pos = cli.DeleteAndPrint(pos, "")
```
fulghum, line 379: Seems like we don't want CLI print calls in the SQL path, right?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4851,6 +4886,532 @@ var DiffTableFunctionScriptTests = []queries.ScriptTest{
...
+
+			"select dolt_checkout('main');",
+			"insert into t values (2, 'two', 'three');",
+			"set @Commit6 = dolt_commit('-am', 'inserting row 2 in main');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			// TODO : applies to all skipped tests in this script
+			//       need to differentiate the NULL from drop column and user defined NULL/NULL from add column
+			{
+				SkipResultsCheck: true,
```
fulghum, line 5205: For these two `SkipResultsCheck`, I think we might actually want to still check the results so that we have coverage and protection from a regression, even though I know they aren't the results we *want* to return. My suggestion is that for now, we test against the results we have, but we leave a note (here or in the diff code or both if you want) about the enhancement.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4851,6 +4886,532 @@ var DiffTableFunctionScriptTests = []queries.ScriptTest{
...
+			"alter table t add column c2 varchar(20);",
+			"insert into t values (4, 'four', 'five');",
+			"update t set c2='foo' where pk=1;",
+			"set @Commit4 = dolt_commit('-am', 'adding column c2, inserting, and updating data');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			// TODO : applies to all skipped tests in this script
+			//       need to differentiate the NULL from drop column and user defined NULL/NULL from add column
+			{
+				SkipResultsCheck: true,
```
fulghum, line 5246: Same for these, too – overall, I think there's value in checking the results and making sure we get the results we expect to get (even if those aren't the ideal results we want to be able to return).

fulghum (approved):
Nice work @jennifersp! 😄 This is a really cool addition to Dolt's SQL diff functionality. 

Code looks great, just a few minor comments on some quick changes. I was particularly impressed with the thoroughness of the test scripts and how well-organized they are. Very nice work. 

Let me know if you want me to look at any other revisions on this one, but I think the follow ups are all pretty simple.

PR #4364, opened 2022-09-17
JSON read perf

max-hoffman:
Re: https://github.com/dolthub/go-mysql-server/pull/1270

`types_table_scan` from `425 ms/q` -> `134 ms/q`, ~70% faster. The nightly benchmark should be around 2.5 (from 7.9).


max-hoffman:
#benchmark

PR #4366, opened 2022-09-17
Fix bug in colDiffs

### LINKED ISSUE #4326: UI issue for `dolt diff` and dropped columns
When dropping a column, there is different behavior for `dolt diff` depending on the type of column being dropped.

In both examples, the columns `s` and `val2` are dropped for `varchar` and `int` respectively. Notice that there is a red font color for the `varchar` column, but not for the `int` column.

tanyongzhi:
Fixes #4326 

- Fix a bug in `splitDiffResultRow` where a key existence is not being checked in a map. Specifically, `ds.fromTo[i]]` defaulted to `0` (the zero value for `int`) in the case where `i` did not exist as a key in `ds.fromTo`.

image

### DIFF
--- a/go/cmd/dolt/commands/diffsplitter.go
+++ b/go/cmd/dolt/commands/diffsplitter.go
@@ -118,9 +118,14 @@ func (ds diffSplitter) splitDiffResultRow(row sql.Row) (rowDiff, rowDiff, error)
 			oldRow.row[ds.queryToTarget[i]] = row[i]
 
 			if diffTypeStr == "modified" {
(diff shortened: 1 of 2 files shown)


fulghum:
Hey @tanyongzhi, thanks for digging in and fixing this one! 🙏 The code change looks great. I'm just rerunning the latest CI tests to make sure they pass, too, then I'll approve and merge in.

fulghum (approved):
Thanks for contributing this fix @tanyongzhi! 🙏 Nice work knocking this one out so quickly! ⚡ 

I'll get this merged into main and it'll go out in the next release.

PR #4372, opened 2022-09-19
go/dotlcore/sqle/writer: Handle `UNIQUE KEY`s overlapping with primary keys

### LINKED ISSUE #4369: Dolt panic on duplicate insertion when a column is part of both a primary key and unique key
Reproduction case on dolt 0.41.2:

```
dolt> CREATE TABLE `test` (
  `col_a` varchar(100),
  `col_b` varchar(100),
  `col_c` varchar(100),
  PRIMARY KEY (`col_a`,`col_b`),
  UNIQUE KEY `test_col_b_col_c_unique` (`col_b`,`col_c`)
);
dolt> insert into `test` (`col_a`, `col_b`, `col_c`) VALUES ('0', '0', '0');
dolt> insert into `test` (`col_a`, `col_b`, `col_c`) VALUES ('1', '0', '0');
panic: runtime error: slice bounds out of range [-2:]

goroutine 1 [running]:
github.com/dolthub/dolt/go/store/val.Tuple.Count(...)
	/src/store/val/tuple.go:192
github.com/dolthub/dolt/go/store/val.Tuple.GetField({0x0?, 0xf999f6?, 0x227b9a8?}, 0xc000f76d20?)
	/src/store/val/tuple.go:155 +0x1a5
```

Changing the primary key to only col_a causes the panic to no longer occur, and the above test case then errors gracefully, referencing the duplicate key violation.

I tested this in mysql and it fails gracefully with a duplicate key error as expected.

Note: this was discovered when trying to perform a query of the form "INSERT ... VALUES ... ON DUPLICATE KEY ...". The expectation here would be no error and the ON DUPLICATE KEY clause would fire to update the existing data.

andrew-wm-arthur:
fixed #4369


druvv (approved):
lgtm!

PR #4387, opened 2022-09-21
remotestorage,remotesrv: Be less restrictive in repository paths for remotesapi repos that are not served through dolthub.

reltuk:
This uses the repo_path parameter to allow single and 3+ path component paths for repository names.

### DIFF
--- a/go/libraries/doltcore/remotesrv/http.go
+++ b/go/libraries/doltcore/remotesrv/http.go
@@ -101,6 +101,7 @@ func (fh filehandler) ServeHTTP(respWr http.ResponseWriter, req *http.Request) {
 		}
 
 		i := strings.LastIndex(path, "/")
+		// a table file name is currently 32 characters, plus the '/' is 33.
 		if i < 0 || len(path[i:]) != 33 {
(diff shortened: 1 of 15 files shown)


go/libraries/doltcore/remotesrv/http.go
```diff
@@ -100,16 +100,15 @@ func (fh filehandler) ServeHTTP(respWr http.ResponseWriter, req *http.Request) {
 			return
 		}
 
-		tokens := strings.Split(path, "/")
-		if len(tokens) != 3 {
+		i := strings.LastIndex(path, "/")
+		if i < 0 || len(path[i:]) != 33 {
```
andrew-wm-arthur, line 104: `len("https://doltremoteapi.dolthub.com")`?

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/remotesrv/http.go
reltuk: Good point, deserves a comment. `len("/" + tableFileID)`, which is also len of a serialized dolt commit hash.

PR #4397, opened 2022-09-22
print spatial types as hex in shells only

JCOR11599:
Related PR: https://github.com/dolthub/go-mysql-server/pull/1278

Since the `SQL` methods for `Point`, `LineString`, `Polygon`, and `Geometry` no longer return the hex formatted string, need to do a little extra work for prettier shell output.

### DIFF
--- a/go/libraries/doltcore/sqle/sqlutil/sql_row.go
+++ b/go/libraries/doltcore/sqle/sqlutil/sql_row.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"math"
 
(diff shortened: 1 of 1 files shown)


go/libraries/doltcore/sqle/sqlutil/sql_row.go
```diff
@@ -320,6 +321,13 @@ func SqlColToStr(sqlType sql.Type, col interface{}) (string, error) {
 			} else {
 				return "false", nil
 			}
+		case sql.Point, sql.LineString, sql.Polygon, sql.GeometryType:
```
max-hoffman, line 324: `sql.SpatialColumnType`?

max-hoffman (approved):
LGTM

PR #4400, opened 2022-09-23
allow drop current database directory and fix drop db with '-'

jennifersp:
This PR fixes two issues:
- if the server is running in the database itself, the database can be dropped regardless of being selected database.
- drop database with '-' in dir name and '_' in database name.

### DIFF
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -539,39 +539,16 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error 
 	dbKey := formatDbMapKeyName(name)
 	db := p.databases[dbKey]
 
-	// Get the DB's directory
-	exists, isDir := p.fs.Exists(db.Name())
-	if !exists {
-		// is directory itself the database being dropped?
-		doltDirExists, _ := p.fs.Exists(dbfactory.DoltDir)
-		if doltDirExists {
-			// guess the database name for this directory
-			absPath, err := p.fs.Abs("")
-			if err != nil {
-				return err
-			}
-			var possibleDBName string
-			if s := strings.Split(absPath, "/"); len(s) > 0 {
-				possibleDBName = strings.Replace(s[len(s)-1], "-", "_", -1)
-			}
-			if possibleDBName == db.Name() {
-				err = p.fs.Delete(dbfactory.DoltDir, true)
-				if err != nil {
-					return err
-				}
-			} else {
-				return sql.ErrDatabaseNotFound.New(db.Name())
-			}
-		} else {
-			return sql.ErrDatabaseNotFound.New(db.Name())
-		}
-	} else if !isDir {
-		return fmt.Errorf("unexpected error: %s exists but is not a directory", dbKey)
-	} else {
-		err = p.fs.Delete(db.Name(), true)
-		if err != nil {
+	// try with user-defined database name, if not found try alternative name with '-' instead of '_'
+	found, err := p.findAndDeleteDatabaseDir(db.Name(), dbKey)
+	if !found {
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -568,6 +570,46 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
 	return nil
 }
 
+// findAndDeleteNestedDatabaseDir tries to find matching database against nested database.
+// This function returns whether the database was found and deleted successfully.
+func (p DoltDatabaseProvider) findAndDeleteNestedDatabaseDir(dbName, dbKey string) (bool, error) {
```
fulghum, line 575: `DoltDatabaseProvider` has a `dbLocations` field with a map of database names to `FileSys`. It feels like this logic could be a bit simpler (and safer maybe?) if we used that as the authoritative source for database locations instead of redoing work to guess directories and checking to see if it's the right one.

fulghum (commented):
Left a suggestion on what might turn out to be a cleaner/simpler/safer way to find the database directory. Check that out and let me know if it helps or not.

go/libraries/doltcore/sqle/database_provider.go
jennifersp: Ah I see, that definitely makes it lot easier and safer.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -539,21 +539,44 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
...
-	} else if !isDir {
-		return fmt.Errorf("unexpected error: %s exists but is not a directory", dbKey)
 	}
-
-	err = p.fs.Delete(db.Name(), true)
+	dropDbLoc, err := dbLoc.Abs("")
+	if err != nil {
+		return err
+	}
+	currentDbLoc, err := p.fs.Abs("")
```
fulghum, line 551: `currentDbLoc` could be a little misleading of a name. It makes me think of the current database a user has selected, but I don't think that's what we mean here. 

How about something like `rootDbLoc` or something similar to convey this is the root directory where all dbs are stored?

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -539,21 +539,44 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
...
 	if err != nil {
 		return err
 	}
+	dirToDelete := ""
+	// if the database is in the directory itself, we remove '.dolt' directory rather than
+	// the whole directory itself because it can have other databases that are nested.
```
fulghum, line 557: Excellent comment! This is a very important point, so good use of a comment to make sure it's clear to all future readers. Deleting other databases a user did not request would be a very bad mistake. 

Speaking of... this would be a really good thing for us to test carefully, too... for example... have a test set up that has a database directly in the db root directory as well as nested db directories, delete the root db, and then do some sanity query like `show databases` or something to make sure the other databases are still present.

integration-tests/bats/sql-server.bats
```diff
@@ -1739,3 +1739,53 @@ s.close()
...
+
+    mkdir mydb
+    cd mydb
+    dolt init
+
+    start_sql_server >> server_log.txt 2>&1
+
+    server_query "" 1 dolt "" "DROP DATABASE mydb"
+
+    run grep "database not found: mydb" server_log.txt
```
fulghum, line 1754: Running a query like `show databases;` and/or using the database to run a sanity test query would probably be a little bit more direct way to test that the toBeDeleted database is gone and that the other nested databases are still present.

fulghum (approved):
Code looks good. Just a few small comments, most important of which is a little deeper testing to make sure the other databases are still present and working after deleting the root database.

PR #4401, opened 2022-09-23
go/commands: Add sql-server option for validating query results against a MySQL instance

andrew-wm-arthur:
depends on https://github.com/dolthub/go-mysql-server/pull/1279


zachmu (approved):
LGTM

PR #4402, opened 2022-09-26
[interactive-pr] Branch Control Demo

Hydrocharged:
Take a look over this for now, and I'll walk through it with you @zachmu later today (along with a live demo showing its use). This is only the _core_ Dolt changes so far.

As a big warning, this is all temporary code, and I expect very little of it to make it into the final PR. However, I will be building upon many of the ideas here, so if any of these are fundamentally wrong in some way, this is a great time to catch it. I've also littered comments throughout, going over how I plan to actually implement much of the functionality.

As a refresher, we have two tables `dolt_branch_control` (`dbc`) and `dolt_branch_control_protected` (`dbcp`). The first, `dbc`, handles the permissions that each branch and user combination has. These are implemented using the `LIKE` expression syntax, which will (eventually) match using the "longest match wins" rule. The second, `dbcp`, is a sort of blacklist that restricts branch creation. If a branch matches, then we return the result of the user and host fields matching. If no expressions match the branch name, then it is allowed. This means it **only** fails when we've matched a branch name, but the user and host do not match.

Included are three files: the core `temp_branch_control/branch_control.go` file, which is the hub for the test implementation. `dtables/branch_control.go`, which is mainly to show the schema. And `prolly_table_writer.go`, which represents how the checks are added throughout the codebase, and also shows the level at which they're implemented (i.e. above the `types` package, but not in GMS). This does present a problem regarding other engineers adding new functionality, but I have some ideas on the implementation of a PR check that may catch some of it.

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/branch_control.go
+++ b/go/libraries/doltcore/sqle/dtables/branch_control.go
@@ -1,0 +1,407 @@
+// Copyright 2022 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


Hydrocharged:
Also, here's the doc in case you want to look over it again. I'll go over anything relevant later though, so it's not necessary:
https://docs.google.com/document/d/1kZ-fOtjWUZT2DVeeW0pmHCgAhr8c7Wo26jwKv0AigMk

PR #4413, opened 2022-09-27
turn off opt-in feature for unix socket

jennifersp:
Initial implementation of unix socket has opt-in feature for unix socket support. The default of mysql unix socket when socket flag is not defined is when host is defined as 'localhost' or undefined.

### DIFF
--- a/integration-tests/bats/sql-server.bats
+++ b/integration-tests/bats/sql-server.bats
@@ -1542,6 +1542,30 @@ databases:
     [[ "$output" =~ "database is locked to writes" ]] || false
 }
 
+@test "sql-server: start server without socket flag should set default socket path" {
+    skiponwindows "unix socket is not available on Windows"
+    cd repo2
+    DEFAULT_DB="repo2"
+    let PORT="$$ % (65536-1024) + 1024"
+
+    dolt sql-server --port $PORT --user dolt >> log.txt 2>&1 &
+    SERVER_PID=$!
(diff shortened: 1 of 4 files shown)


integration-tests/bats/sql-server.bats
```diff
@@ -1542,6 +1542,30 @@ databases:
     [[ "$output" =~ "database is locked to writes" ]] || false
 }
 
+@test "sql-server: start server without socket flag should set default socket path" {
+    skiponwindows "unix socket is not available on Windows"
+    cd repo2
+    DEFAULT_DB="repo2"
+    let PORT="$$ % (65536-1024) + 1024"
```
fulghum, line 1549: (future-refactoring) Not necessary for this PR, but just an idea for future refactoring/cleanup... it would be cool if we had a standard helper function we could call that would select a good port and potentially even check that it's not in use yet. Looks like this port selection code is in this BATS file a couple dozen times, so good candidate for a helper function. Not needed for this review, just sharing a future refactoring thought.

fulghum (approved):
Looks good! Exciting to turn this feature on by default now!

integration-tests/bats/sql-server.bats
fulghum: Speaking of... this actually bit me this afternoon in a BATS run! one of the `sql-server` tests selected a port that was already in use and I had to restart the run. This would be a nice little cleanup task for the sql-server tests. 😄

PR #4429, opened 2022-09-29
go: sqle: cluster: Add a dolt_cluster database, exposing a dolt_cluster_status table which queries replication status.

reltuk:


andrew-wm-arthur (approved):
LGTM!

PR #4431, opened 2022-09-29
Make NomsBinFormat `__DOLT__` the default format

andrew-wm-arthur:


go/libraries/doltcore/sqle/alterschema_test.go
```diff
@@ -369,35 +363,32 @@ func TestDropColumnUsedByIndex(t *testing.T) {
...
 		expectedIndex  bool
 		expectedSchema schema.Schema
-		expectedRows   []row.Row
 	}{
 		{
 			name:           "remove int",
 			colName:        "age",
 			expectedIndex:  true,
 			expectedSchema: dtestutils.RemoveColumnFromSchema(dtestutils.TypedSchema, dtestutils.AgeTag),
-			expectedRows:   dtestutils.TypedRows,
```
zachmu, line 379: Would be better to keep these assertions

zachmu: Maybe transform to SQL rows and run a select *

go/libraries/doltcore/sqle/alterschema_test.go
```diff
@@ -802,7 +780,6 @@ func TestModifyColumn(t *testing.T) {
 		newColumn      schema.Column
 		order          *sql.ColumnOrder
 		expectedSchema schema.Schema
-		expectedRows   []row.Row
```
zachmu, line 805: Same comment here

go/libraries/doltcore/sqle/index/mergeable_indexes_test.go
```diff
@@ -30,13 +30,18 @@ import (
...
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/typed/noms"
+	"github.com/dolthub/dolt/go/store/types"
 )
 
 // This tests mergeable indexes by using the SQL engine and intercepting specific calls. This way, we can verify that
 // the engine is intersecting and combining the proper number of lookups, and we can also examine the ranges before
 // they're converted into a format that Noms understands to verify that they were handled correctly. Lastly, we ensure
 // that the final output is as expected.
 func TestMergeableIndexes(t *testing.T) {
+	if types.Format_Default != types.Format_LD_1 {
```
zachmu, line 41: Add a note about mergeable indexes only being used for testing?

go/libraries/doltcore/sqle/sqlselect_test.go
```diff
@@ -969,216 +975,6 @@ var AsOfTests = []SelectTest{
 	},
 }
 
-// SQL is supposed to be case insensitive. These are tests of that promise.
-// Many of these are currently broken in go-myqsl-server. The queries return the correct results in all cases, but the
-// column names in the result schemas often have the wrong case. They sometimes use the case from the table, rather
-// than the case of the expression in the query (the correct behavior). This is a minor issue, but we should fix it
-// eventually.
-var CaseSensitivityTests = []SelectTest{
```
zachmu, line 977: You think this functionality is well represented in existing engine tests?

go/libraries/doltcore/table/editor/index_editor_test.go
```diff
@@ -52,8 +52,9 @@ var id2, _ = uuid.NewRandom()
 var id3, _ = uuid.NewRandom()
 
 func TestIndexEditorConcurrency(t *testing.T) {
-	format := types.Format_Default
+	format := types.Format_LD_1
```
zachmu, line 55: Probably better to skip?

zachmu (approved):
Good enough for now, but who owns migrating all the tests that still only work with LD_1?

go/libraries/doltcore/sqle/sqlselect_test.go
andrew-wm-arthur: I do, and most of these tests were skipped anyway

go/libraries/doltcore/table/editor/index_editor_test.go
andrew-wm-arthur: these tests are specific to `__LD_1__` and I think they're useful to that end

go/libraries/doltcore/sqle/alterschema_test.go
andrew-wm-arthur: I looked into it, but decided all these DDL tests are over-specified and don't need to mess with row data

PR #4439, opened 2022-09-30
.github/workflows: Run full CI against both `__DOLT__` and `__LD_1__`

andrew-wm-arthur:
The default NomsBinFormat is now `__DOLT__`. This change updates CI to continue running golang tests and BATS against the old NBF `__LD_1__`


coffeegoddd (approved):
LGTM

PR #4443, opened 2022-09-30
go: sqle: cluster: When the server has role "standby", make all dolt databases read-only.

reltuk:


go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -372,38 +390,74 @@ func (c *Controller) RemoteSrvServerArgs(ctx *sql.Context, args remotesrv.Server
...
+// called with c.mu held
 func (c *Controller) gracefulTransitionToStandby() error {
+	c.setProviderIsStandby(true)
+	c.killRunningQueries()
+	// TODO: this can block with c.mu held, although we are not too
+	// interested in the server proceeding gracefully while this is
+	// happening.
+	if err := c.waitForHooksToReplicate(); err != nil {
+		c.setProviderIsStandby(false)
+		c.killRunningQueries()
```
andrew-wm-arthur, line 405: why do we call this again here?

andrew-wm-arthur (approved):

go/libraries/doltcore/sqle/cluster/controller.go
reltuk: Because connections could have come in while we were trying to true up the standby and they will have busted InitialDbStates and stuff in their session.

PR #4449, opened 2022-10-01
integration-tests/go-sql-server-driver: Proof of concept a declarative approach to integration tests for sql-server interactions.

reltuk:
Converts bats/sql-server-config.bats to use a new approach.


integration-tests/go-sql-server-driver/cmd.go
```diff
@@ -0,0 +1,241 @@
...
+		Repo:   r,
+	}, nil
+}
+
+func (s *SqlServer) GracefulStop() error {
+	err := s.Cmd.Process.Signal(syscall.SIGTERM)
+	if err != nil {
+		return err
+	}
+	<-s.Done
```
zachmu, line 196: Timeout?

integration-tests/go-sql-server-driver/testdef.go
```diff
@@ -0,0 +1,201 @@
...
+								require.Error(t, err)
+								require.Regexp(t, q.ErrorMatch, err.Error())
+							}
+							require.NoError(t, err)
+							defer rows.Close()
+							cols, err := rows.Columns()
+							require.NoError(t, err)
+							require.Equal(t, q.Result.Columns, cols)
+							for _, r := range q.Result.Rows {
+								require.True(t, rows.Next())
```
zachmu, line 156: Rather than this assertion inside the loop, it would be better to consume the entire row buffer, then do a simple require.Equals(expected, found) at the end, which will include a length check and have a better error message

zachmu (approved):
Not a fan of the yaml but agree this is going to be more reliable than the original

PR #4457, opened 2022-10-04
storing prefix lengths, and showing them in `SHOW CREATE TABLE`

JCOR11599:


go/libraries/doltcore/schema/schema.go
```diff
@@ -69,6 +69,13 @@ type Schema interface {
 	// SetPkOrdinals specifies a primary key column ordering. See GetPkOrdinals.
 	SetPkOrdinals([]int) error
 
+	// GetPkPrefixLengths returns a slice of pairs (tag, prefix length) containing the prefix lengths
+	// for any columns that use a text/blob index
+	GetPkPrefixLengths() []uint16
+
+	// SetPkPrefixLengths specifies the tags and prefix lengths for the columns that use text/blob indexes
+	SetPkPrefixLengths([]uint16) error
+
```
andrew-wm-arthur, line 78: revert these for now, we'll start with prefixes for secondary indexes only, per offline discussion

go/libraries/doltcore/schema/index_coll.go
```diff
@@ -64,6 +64,8 @@ type IndexCollection interface {
 	RenameIndex(oldName, newName string) (Index, error)
 	//SetPks changes the pks or pk ordinals
 	SetPks([]uint64) error
+	// SetIndexPrefixLength sets the prefix lengths for the specified index
+	SetIndexPrefixLength(indexName string, prefixLengths []uint16) error
```
andrew-wm-arthur, line 68: this should be a method on `Index`

andrew-wm-arthur (commented):
Generally looks good, as we discussed, we're going to start with secondary indexes only

PR #4460, opened 2022-10-04
Add --all flag to dolt push

druvv:
In git, the `--all` flag can be specified which pushes all local branches. In addition, the `-u` flag can be specified which also tracks those branches.

This PR adds the `--all` flag to `dolt push`. It will be used in the migration jobs to push a migrated repo to an empty remote.

### DIFF
--- a/go/cmd/dolt/commands/push.go
+++ b/go/cmd/dolt/commands/push.go
@@ -74,6 +74,7 @@ func (cmd PushCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParser()
 	ap.SupportsFlag(cli.SetUpstreamFlag, "u", "For every branch that is up to date or successfully pushed, add upstream (tracking) reference, used by argument-less {{.EmphasisLeft}}dolt pull{{.EmphasisRight}} and other commands.")
 	ap.SupportsFlag(cli.ForceFlag, "f", "Update the remote with local history, overwriting any conflicting history in the remote.")
+	ap.SupportsFlag(cli.AllFlag, "a", "Pushes all local branches. Equivalent to calling {{.EmphasisLeft}}dolt push <branch_name>{{.EmphasisRight}} for all branches. -u can also be specified to track the remote branches.")
 	return ap
 }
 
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/commands/push.go
```diff
@@ -88,6 +89,53 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		verr := errhand.BuildDError("failed to fetch branches").AddCause(err).Build()
+		return HandleVErrAndExitCode(verr, usage)
+	}
+
+	if len(apr.Args) > 1 {
+		verr := errhand.BuildDError("only <remote> may be specified with --all").SetPrintUsage().Build()
+		return HandleVErrAndExitCode(verr, usage)
+	}
+
+	remoteName := "origin"
```
zachmu, line 111: Is this what git does? Can't tell from the docs.

Normally `git push` without a remote named will push to the upstream tracking branch, but that can't be the case with `--all`

go/cmd/dolt/commands/push.go
```diff
@@ -88,6 +89,53 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
 
+	if apr.Contains(cli.AllFlag) {
+		return cmd.pushAllBranches(ctx, apr, dEnv, usage)
+	}
+
+	return cmd.doPush(ctx, apr, dEnv, usage)
+}
+
+func (cmd PushCmd) pushAllBranches(ctx context.Context, apr *argparser.ArgParseResults, dEnv *env.DoltEnv, usage cli.UsagePrinter) int {
+	branches, err := dEnv.DoltDB.GetBranches(ctx)
```
zachmu, line 100: This includes e.g. remote branches too, right? Or are those stored under a different ref namespace?

integration-tests/bats/remotes.bats
```diff
@@ -2188,3 +2188,87 @@ SQL
...
+
+    # Push branch1 and branch2 to origin
+    run dolt push --all
+    [ $status -eq 0 ]
+
+    # Verify that branch1 and branch2 were pushed by checking them out in separate repo
+    cd ../repo2
+    dolt fetch
+    run dolt checkout branch1
+    [ $status -eq 0 ]
```
zachmu, line 2230: If your only check is `status eq 0`, you can just omit `run`, same thing

zachmu (approved):
LGTM, just a couple questions about weird edge cases

PR #4461, opened 2022-10-04
wrap spatial type string in single quotes when exporting

jennifersp:


integration-tests/bats/sql-spatial-types.bats
```diff
@@ -265,3 +265,22 @@ SQL
     [ "$status" -eq 1 ]
     [[ "$output" =~ "error executing query on line 2: unsupported geospatial type: MultiPoint from value: 0x0" ]] || false
 }
+
+@test "sql-spatial-types: round trip dolt dump with spatial type data" {
```
fulghum, line 269: Do you think it's worth also updating your Sakila dump test to see if we can roundtrip? I don't have a strong opinion, but figured I'd mention it since this reminded me of those tests.

fulghum (approved):
Very nice! This is a great bug fix. 🐞

PR #4467, opened 2022-10-05
store/datas/pull: If the BPS is less than 1, report 0

druvv:
image
 ^ prevents the push / pull progress reporter from reporting very small BP/s units.


timsehn (approved):
LGTM

PR #4522, opened 2022-10-11
fix `insert ... on duplicate key update ...` for unique keys

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/4475

only happens in new format, solution was just to validate any key violations before inserting anything.

### DIFF
--- a/go/libraries/doltcore/sqle/writer/prolly_index_writer.go
+++ b/go/libraries/doltcore/sqle/writer/prolly_index_writer.go
@@ -71,6 +71,7 @@ func getPrimaryKeylessProllyWriter(ctx context.Context, t *doltdb.Table, sqlSch 
 type indexWriter interface {
 	Name() string
 	Map(ctx context.Context) (prolly.Map, error)
+	ValidateKeyViolations(ctx context.Context, sqlRow sql.Row) error
 	Insert(ctx context.Context, sqlRow sql.Row) error
 	Delete(ctx context.Context, sqlRow sql.Row) error
 	Update(ctx context.Context, oldRow sql.Row, newRow sql.Row) error
@@ -106,7 +107,7 @@ func (m prollyIndexWriter) Map(ctx context.Context) (prolly.Map, error) {
 	return m.mut.Map(ctx)
 }
 
-func (m prollyIndexWriter) Insert(ctx context.Context, sqlRow sql.Row) error {
+func (m prollyIndexWriter) ValidateKeyViolations(ctx context.Context, sqlRow sql.Row) error {
 	for to := range m.keyMap {
 		from := m.keyMap.MapOrdinal(to)
 		if err := index.PutField(ctx, m.mut.NodeStore(), m.keyBld, to, sqlRow[from]); err != nil {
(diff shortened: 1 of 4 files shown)


timsehn:
Maybe I should move the GMS bug back and reopen it? 

I thought this was a bug in both GMS and Dolt but if it's just a Dolt regression from new format, we should reopen the GMS bug because in-memory does not support unique indexes?

JCOR11599:
> Maybe I should move the GMS bug back and reopen it?
> 
> I thought this was a bug in both GMS and Dolt but if it's just a Dolt regression from new format, we should reopen the GMS bug because in-memory does not support unique indexes?



> Maybe I should move the GMS bug back and reopen it?
> 
> I thought this was a bug in both GMS and Dolt but if it's just a Dolt regression from new format, we should reopen the GMS bug because in-memory does not support unique indexes?

I think that makes sense, technically everything involving unique indexes wont work for anyone using solely GMS, since they're not supported

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -122,10 +123,21 @@ func (m prollyIndexWriter) Insert(ctx context.Context, sqlRow sql.Row) error {
...
+}
+
+func (m prollyIndexWriter) Insert(ctx context.Context, sqlRow sql.Row) error {
+	for to := range m.keyMap {
+		from := m.keyMap.MapOrdinal(to)
+		if err := index.PutField(ctx, m.mut.NodeStore(), m.keyBld, to, sqlRow[from]); err != nil {
+			return err
+		}
+	}
+	k := m.keyBld.Build(sharePool)
```
andrew-wm-arthur, line 136: maybe factor a `keyFromRow(sqlRow sql.Row) (key val.Tuple, error) {...}`

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -267,13 +279,16 @@ func (m prollySecondaryIndexWriter) Map(ctx context.Context) (prolly.Map, error)
 	return m.mut.Map(ctx)
 }
 
-func (m prollySecondaryIndexWriter) Insert(ctx context.Context, sqlRow sql.Row) error {
+func (m prollySecondaryIndexWriter) ValidateKeyViolations(ctx context.Context, sqlRow sql.Row) error {
 	if m.unique {
 		if err := m.checkForUniqueKeyErr(ctx, sqlRow); err != nil {
 			return err
 		}
```
andrew-wm-arthur, line 286: might as well inline this method if it's only called here

go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
```diff
@@ -45,6 +45,10 @@ func (k prollyKeylessWriter) Map(ctx context.Context) (prolly.Map, error) {
 	return k.mut.Map(ctx)
 }
 
+func (k prollyKeylessWriter) ValidateKeyViolations(ctx context.Context, sqlRow sql.Row) error {
+	return nil
```
andrew-wm-arthur, line 49: add a comment explaining noop implementation

andrew-wm-arthur (approved):
Looks Good, let's add one or two enginetests

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
JCOR11599: `checkForUniqueKeyErr` is also called in `prollySecondaryIndexWriter.Update()`

timsehn:
> > Maybe I should move the GMS bug back and reopen it?
> > I thought this was a bug in both GMS and Dolt but if it's just a Dolt regression from new format, we should reopen the GMS bug because in-memory does not support unique indexes?
> 
> I think that makes sense, technically everything involving unique indexes wont work for anyone using solely GMS, since they're not supported

Done. https://github.com/dolthub/go-mysql-server/issues/571

PR #4524, opened 2022-10-11
add file read progress info for `dolt sql --file=<file.sql>` command

jennifersp:
There will be percentage progress output printed for `--file` flag of `dolt sql` command. It can be used with or without `--batch` flag.


fulghum (approved):
Looks good!

PR #4538, opened 2022-10-12
add list of `dolt config` variables to command description

jennifersp:


fulghum (approved):
Thanks for adding these! This is super helpful to have documented in the help content. 🎉

PR #4539, opened 2022-10-12
Replace incorrect character.

PavelSafronov:
Replaces an incorrect ` ' ` character with the correct `` ` `` character.

The current code results in [broken docs html](https://docs.dolthub.com/cli-reference/cli#text-dolt-sql-server):
Screen Shot 2022-10-12 at 3 18 26 PM


zachmu (approved):
Nice, your first PR

Also what is this filth

image

Issue #4526, opened 2022-10-12
Duplicate tag error


evyalyy:
when creating table
```
CREATE TABLE hero_slot (
	hero_id INTEGER NOT NULL, 
	slot_id INTEGER NOT NULL, 
	equipment_type TEXT NOT NULL, 
	durability INTEGER NOT NULL, 
	condition_unlock TEXT NOT NULL, 
	visual TEXT NOT NULL, 
	CONSTRAINT hero_slot_pkey PRIMARY KEY (hero_id, slot_id), 
	CONSTRAINT hero_slot_fk1 FOREIGN KEY(hero_id) REFERENCES hero (hero_id)
)
```
Got an error `Cannot create column durability, the tag 14653 was already used in table seasonal_missions_slots`
Mentioned table is created as follows
```
CREATE TABLE seasonal_missions_slots (
	type_id INTEGER NOT NULL, 
	id INTEGER NOT NULL, 
	slot_id INTEGER NOT NULL, 
	quests_pool JSON NOT NULL, 
	enabled BOOL NOT NULL, 
	CONSTRAINT seasonal_missions_slots_pkey PRIMARY KEY (type_id, id, slot_id), 
	CHECK (enabled IN (0, 1))
)
```
Started to get this error only when table `seasonal_missions_slots` has been added.
Also it's reproduces on both old and new storage formats

Issue #4406, opened 2022-09-26
JSON default column normalization issue causing merge conflict


druvv:
Customer ran into a schema conflict on a JSON column that defines a default value. The repro is still being investigated.

druvv:
The specific error message reported by the customer is:
```
ERROR 1105 (HY000): schema conflict found, merge aborted. Please alter schema to prevent schema conflicts before merging.
schema conflicts for table $table:
	different column definitions for our column $col and their column $col
```

This error is thrown when $col has changed in both branches with differring schema. The type of $col must not have been changed otherwise a new tag would have been generated. The factors that affect column equality are tag, type, primary key membership, column constraints, and the column's default value.

I focused on trying to produce a situation where the column's default value looks identical in the information schema, but is not equivalent in dolt merge.

The first thing I checked was to see if information schema normalized the json value when being printed. But that is not the case. I created a json column with the default of `{"hi": 1}` on one branch and `{"hi":  1}` (an extra space) on another branch. A conflict was thrown when merged, but the information schema also showed the extra space.

I also investigated how the column defaults are being pulled from postgres, but in the end couldn't find anything obviously wrong.

druvv:
Here is the minimal repro:

```bash
dolt init
dolt sql -q "create table t (pk int PRIMARY KEY);"
dolt commit -Am "add table t"
dolt branch right
dolt sql <<SQL
alter table t add column col1 json default ('{"test":1}');
SQL
dolt commit -am "add default"

dolt checkout right
dolt sql <<SQL
alter table t add column col1 json default ("{\"test\": 1}");
SQL
dolt commit -am "right"

dolt checkout main
dolt merge right
```

Returns the following error:
```
Updating c5d347orbrdk8gpo5l1fmlhvoss2nl3p..9mdpreapepser1gv5tvf85hfhvob3lmc
Unable to stage changes: add and commit to finish merge
error: Failed to commit changes.
cause: schema conflict found, merge aborted. Please alter schema to prevent schema conflicts before merging.
       schema conflicts for table t:
       	different column definitions for our column col1 and their column col1
```

If we inspect the schema on branch main, and right we see that they are identical:
```
> dolt checkout right
> dolt sql -q "show create table t;"
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                         |
+-------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #4540, opened 2022-10-12
[no-release-notes] Spec for Table, Column matching

andrew-wm-arthur:
During diff and merge operations, it is necessary to resolve table and column identity create an entity mapping between Dolt root values (logical databases). Currently, table and column identity is determined by [column tags](https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/schema/column.go#L65). During DDL operations, we make some effort to instantiate column tags such that common operations "just work". However, when these heuristics fails, the results can be hard to debug and the errors are often opaque to users. (See #4406 and #4526).

This PR includes (unimplemented) tests specifying new matching behavior. The proposed change would match tables and columns using only the SQL schema and table data.

### DIFF
--- a/go/libraries/doltcore/diff/column_identity_test.go
+++ b/go/libraries/doltcore/diff/column_identity_test.go
@@ -1,0 +1,328 @@
+// Copyright 2021 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


timsehn:
Would this obviate the need for tags?

go/libraries/doltcore/diff/column_identity_test.go
```diff
@@ -0,0 +1,328 @@
...
+						{name: "b", enc: val.Int32Enc, pk: true},
+						{name: "c0", enc: val.Int32Enc},
+					},
+				},
+			},
+			matches: []match{
+				{
+					leftTbl: "t", rightTbl: "t",
+					columnMatches: [][2]string{
+						{"pk", "a"},
```
zachmu, line 120: "a", "b" ?

go/libraries/doltcore/diff/column_identity_test.go
```diff
@@ -0,0 +1,328 @@
+// Copyright 2021 Dolthub, Inc.
```
zachmu, line 1: Forgot to wind your watch

PR #4548, opened 2022-10-13
prevent creating views with same name as existing table

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/4549


zachmu (commented):
This should go in GMS, not in dolt.

PR #4550, opened 2022-10-13
get next port number if random port is in use

jennifersp:
fixed in bats testing:
- if PORT in use, get the next unused PORT number when starting a sql-server
- kill a process only if it's still running when stopping a sql-server


fulghum (approved):
Woohoo! This is such a nice improvement for the sql-server tests! Thanks for taking this on. 😁👏

PR #4559, opened 2022-10-17
 Branch Control Pt. 6

Hydrocharged:
This fixes a few issues found through testing, and also adds persistence for the tables. Also re-added the `binlog_table.go` file. I originally removed it, but misread the feedback that said to not expose it rather than to remove it.

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -85,24 +85,26 @@ By default this command uses the dolt database in the current working directory,
 var ErrMultipleDoltCfgDirs = errors.NewKind("multiple .doltcfg directories detected: '%s' and '%s'; pass one of the directories using option --doltcfg-dir")
 
 const (
-	QueryFlag         = "query"
-	FormatFlag        = "result-format"
-	saveFlag          = "save"
-	executeFlag       = "execute"
-	listSavedFlag     = "list-saved"
-	messageFlag       = "message"
-	BatchFlag         = "batch"
-	DataDirFlag       = "data-dir"
-	MultiDBDirFlag    = "multi-db-dir"
-	CfgDirFlag        = "doltcfg-dir"
-	DefaultCfgDirName = ".doltcfg"
-	PrivsFilePathFlag = "privilege-file"
-	DefaultPrivsName  = "privileges.db"
-	continueFlag      = "continue"
-	fileInputFlag     = "file"
-	UserFlag          = "user"
-	DefaultUser       = "root"
-	DefaultHost       = "localhost"
+	QueryFlag             = "query"
+	FormatFlag            = "result-format"
+	saveFlag              = "save"
+	executeFlag           = "execute"
+	listSavedFlag         = "list-saved"
+	messageFlag           = "message"
+	BatchFlag             = "batch"
+	DataDirFlag           = "data-dir"
+	MultiDBDirFlag        = "multi-db-dir"
+	CfgDirFlag            = "doltcfg-dir"
+	DefaultCfgDirName     = ".doltcfg"
+	PrivsFilePathFlag     = "privilege-file"
+	BranchCtrlPathFlag    = "branch-control-file"
+	DefaultPrivsName      = "privileges.db"
+	DefaultBranchCtrlName = "branch_control.db"
+	continueFlag          = "continue"
+	fileInputFlag         = "file"
+	UserFlag              = "user"
+	DefaultUser           = "root"
+	DefaultHost           = "localhost"
 
 	welcomeMsg = `# Welcome to the DoltSQL shell.
 # Statements must be terminated with ';'.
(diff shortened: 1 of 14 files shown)


go/cmd/dolt/commands/sql.go
```diff
@@ -256,6 +259,15 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 		}
 	}
 
+	// If no privilege filepath specified, default to doltcfg directory
```
zachmu, line 262: copypasta error

go/libraries/doltcore/branch_control/access.go
```diff
@@ -95,6 +107,64 @@ func (tbl *Access) GetIndex(branchExpr string, userExpr string, hostExpr string)
...
+	for _, val := range tbl.Values {
+		val.Serialize(buffer)
+	}
+	// Write the binlog
+	_ = tbl.binlog.Serialize(buffer)
+}
+
+// Deserialize populates the table with the given data. Returns an error if the data cannot be deserialized, or if the
+// table has already been written to. Deserialize must be called on an empty table.
+func (tbl *Access) Deserialize(data []byte, position *uint64) error {
```
zachmu, line 140: Typically this kind of function returns `(bytes read, error)` rather than mutating an offset pointer

go/libraries/doltcore/branch_control/branch_control.go
```diff
@@ -56,22 +66,86 @@ func init() {
...
+	if !enabled {
+		return nil
+	}
+
+	// Do not attempt to load from an empty file path
+	if len(branchControlFilePath) == 0 {
+		return nil
+	}
+
+	StaticController.branchControlFilePath = branchControlFilePath
```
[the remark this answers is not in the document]
zachmu: Eventually we probably want a way to bake this down to a single file that has references to multiple DBs and their permissions, so that it's possible to e.g. push it to another host during replication

go/libraries/doltcore/branch_control/binlog.go
```diff
@@ -176,36 +221,36 @@ func (row *BinlogRow) Serialize(buffer *bytes.Buffer) {
 	writeUint64(buffer, row.Permissions)
 }
 
-// deserializeBinlogRow returns a BinlogRow from the data at the given position. Also returns the new position. Assumes
-// that the given data's encoded integers are big-endian.
-func deserializeBinlogRow(data []byte, position uint64) (BinlogRow, uint64) {
+// deserializeBinlogRow returns a BinlogRow from the data at the given position. Assumes that the given data's encoded
+// integers are big-endian.
+func deserializeBinlogRow(data []byte, position *uint64) BinlogRow {
```
zachmu, line 226: Something to consider: this is a lot of code to maintain in the future when flatbuffers is already widely deployed in this codebase. Did you try a prototype with that?

go/libraries/doltcore/branch_control/branch_control.go
```diff
@@ -56,22 +66,86 @@ func init() {
...
+		return err
+	}
+	if err = StaticController.Namespace.Deserialize(data, &position); err != nil {
+		return err
+	}
+	return nil
+}
+
+// SaveData saves the data from the context's controller to the location pointed by it.
+func SaveData(ctx context.Context) error {
```
zachmu, line 116: Probably makes more sense as `Serialize(ctx, filepath)`, but maybe that can wait until it's a method on the controller object?

go/libraries/doltcore/sqle/dtables/binlog_table.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2022 Dolthub, Inc.
```
zachmu, line 1: What's the point of these? Just have them around for debugging?

go/libraries/doltcore/sqle/dtables/branch_control_table.go
```diff
@@ -234,17 +245,27 @@ func (tbl BranchControlTable) Update(ctx *sql.Context, old sql.Row, new sql.Row)
...
-			return fmt.Errorf("`%s`@`%s` cannot update the row [%q, %q, %q] to the new branch expression %q",
-				insertUser, insertHost, oldBranch, oldUser, oldHost, newBranch)
+			return branch_control.ErrUpdatingToRow.New(insertUser, insertHost, oldBranch, oldUser, oldHost, newBranch)
 		}
 	}
 
+	// We check if we're updating to a subset of an already-existing row. If we are, we deny the update as the existing
+	// row will already match against ALL possible values for this updated row.
+	_, modPerms := tbl.Match(newBranch, newUser, newHost)
+	if modPerms&branch_control.Permissions_Admin == branch_control.Permissions_Admin {
```
zachmu, line 260: Test for this use case?

zachmu (approved):
Looks good, just a few comments.

Have you tried what this looks like with flatbuffers?

PR #4561, opened 2022-10-17
Support two dot log for CLI

tbantle22:


go/cmd/dolt/commands/log.go
```diff
@@ -126,44 +146,153 @@ func (cmd LogCmd) logWithLoggerFunc(ctx context.Context, commandStr string, args
...
+		notCs, err := getCommitSpec(commit)
+		if err != nil {
+			return nil, nil, "", err
 		}
-	} else { // dolt log ref table
-		cs, err := doltdb.NewCommitSpec(apr.Arg(0))
+		return nil, notCs, "", err
+	}
+
+	argIsRef := actions.ValidateIsRef(ctx, str, dEnv.DoltDB, dEnv.RepoStateReader())
```
fulghum, line 272: There's probably some interesting edge cases here where we can't disambiguate between a branch named "branch1" used as a ref and a table named "branch1" used as a table. Probably worth thinking about and trying a test or two and even adding a test and documenting the behavior that when there is ambiguity between a ref/branch name and a table name that we'd assume a branch ref.

go/cmd/dolt/commands/log.go
```diff
@@ -99,6 +104,7 @@ func (cmd LogCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsFlag(parentsParam, "", "Shows all parents of each commit in the log.")
 	ap.SupportsString(decorateParam, "", "decorate_fmt", "Shows refs next to commits. Valid options are short, full, no, and auto")
 	ap.SupportsFlag(oneLineParam, "", "Shows logs in a compact format.")
+	ap.SupportsString(notParam, "", "revision", "Reverses the meaning of the ^ prefix (or lack thereof) for all following revision specifiers, up to the next --not.")
```
fulghum, line 107: The `--not` param feels a little confusing. I see that this is 100% just coming from how Git uses it; it just feels kinda harder to think about for me than the '^' sign. It seems weird that it reverses all following params until the next `--not` is seen, but I get that we're just following Git. It doesn't look like we can support multiple `--not`s though? So maybe we could simplify the docs a little bit knowing that? Might also want to add a BATS test for `--not` reversing multiple params just so we've got that case covered, too. Or, if you're also feeling weird about `--not` we could omit it for now.

go/cmd/dolt/commands/log.go
```diff
@@ -117,6 +123,20 @@ func (cmd LogCmd) logWithLoggerFunc(ctx context.Context, commandStr string, args
...
 
+	opts, err := parseLogArgs(ctx, dEnv, apr)
+	if err != nil {
+		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+	}
+	if opts.commitSpec == nil {
+		opts.commitSpec = dEnv.RepoStateReader().CWBHeadSpec()
+	}
+	if len(opts.tableName) > 0 {
+		return handleErrAndExit(logTableCommits(ctx, dEnv, opts))
```
fulghum, line 134: Am I reading this wrong, or does this mean the tableName param is not accepted anymore?

fulghum (approved):
Cool new feature for `dolt log`! Just a few comments/suggestions, nothing major.

go/cmd/dolt/commands/log.go
tbantle22: `tableName` is part `opts` now so using `opts.tableName` [here](https://github.com/dolthub/dolt/pull/4561/files#diff-c39bf5c66faf0dc0c5de474674ff85852e5d2043a181b9fcc4bffddeb69da453R438) instead of explicitly passing the tableName as an argument. Can change that back if you think it was clearer

go/cmd/dolt/commands/log.go
tbantle22: Added some bats for this [here](https://github.com/dolthub/dolt/pull/4561/files#diff-90a77217a84d199ff6b5f127e1b52cf4dde4df409835e09e925c51a8023a2a25R125). If there's a branch and table named `myname`, dolt log will have these defaults:
```
dolt log myname[branch]
dolt log myname[branch] myname[table]
```
That means `dolt log myname[table]` won't work unless you provide a revision

go/cmd/dolt/commands/log.go
```diff
@@ -65,7 +70,7 @@ var logDocs = cli.CommandDocumentationContent{
 
 The command takes options to control what is shown and how.`,
 	Synopsis: []string{
-		`[-n {{.LessThan}}num_commits{{.GreaterThan}}] [{{.LessThan}}commit{{.GreaterThan}}] [[--] {{.LessThan}}table{{.GreaterThan}}]`,
+		`[-n {{.LessThan}}num_commits{{.GreaterThan}}] [{{.LessThan}}revision-range{{.GreaterThan}}] [[--] {{.LessThan}}table{{.GreaterThan}}]`,
```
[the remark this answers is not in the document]
tbantle22: How's this description? https://github.com/dolthub/dolt/pull/4561/files#diff-c39bf5c66faf0dc0c5de474674ff85852e5d2043a181b9fcc4bffddeb69da453R71-R85

go/cmd/dolt/commands/log.go
tbantle22: Changed the copy here/more bats so that it's clearer we only support one `--not`

go/cmd/dolt/commands/log.go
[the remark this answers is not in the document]
fulghum: Very nice!

go/cmd/dolt/commands/log.go
fulghum: No worries; this makes sense. 👍

PR #4564, opened 2022-10-17
Add basic `dolt_log` table function

tbantle22:
Supports `select * from dolt_log(<revision>)` (`<revision>` is optional). Adding more options in a follow up PR

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -749,7 +749,7 @@ func makeLargeInsert(sz int) string {
 // DoltUserPrivTests are tests for Dolt-specific functionality that includes privilege checking logic.
 var DoltUserPrivTests = []queries.UserPrivilegeTest{
 	{
-		Name: "dolt_diff table function privilege checking",
+		Name: "table function privilege checking",
 		SetUpScript: []string{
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4919,6 +4940,92 @@ var DiffTableFunctionScriptTests = []queries.ScriptTest{
...
+
+			"insert into t values(1, 'one', 'two'), (2, 'two', 'three');",
+			"set @Commit2 = dolt_commit('-am', 'inserting into t');",
+
+			"call dolt_checkout('-b', 'new-branch')",
+			"insert into t values (3, 'three', 'four');",
+			"set @Commit3 = dolt_commit('-am', 'inserting into t again');",
+			"call dolt_checkout('main')",
+		},
+		Assertions: []queries.ScriptTestAssertion{
```
fulghum, line 5000: It would be nice to test some of the output coming from `dolt_log`, too, in addition to asserting the correct number of rows.

fulghum (approved):
Nice work! Code looks great. Just one suggestion for slightly deeper testing of the content coming from `dolt_log`. 

Thanks for breaking this out into a PR for the foundation for the new function and a separate for the two-dot support. That definitely helped make this one easy to review.

PR #4571, opened 2022-10-18
Branch Control Pt. 8

Hydrocharged:
Per the suggestion, this is a rewrite of branch control persistence using flatbuffers. Using the rest of the codebase as an example, it looks like we write vectors and such first before writing them to the object that they belong to, so I'm following that style.


zachmu (approved):
LGTM!

@reltuk might want to take a look to comment on flatbuffers best practice

reltuk:
LGTM on the flatbuffers side :).

Hydrocharged:
Failing tests were flakes on the server port

PR #4572, opened 2022-10-18
Support two dot log in dolt_log table function

tbantle22:


go/libraries/doltcore/env/actions/commitwalk/commitwalk.go
```diff
@@ -295,3 +295,78 @@ func GetTopNTopoOrderedCommitsMatching(ctx context.Context, ddb *doltdb.DoltDB,
...
 	return commitList, nil
 }
+
+// GetDotDotRevisionsIterator returns an iterator for commits generated with the same semantics as
+// GetDotDotRevisions
+func GetDotDotRevisionsIterator(ctx context.Context, ddb *doltdb.DoltDB, startCommitHash, excludingCommitHash hash.Hash) (doltdb.CommitItr, error) {
+	return newDotDotCommiterator(ctx, ddb, startCommitHash, excludingCommitHash)
+}
+
+type dotDotCommiterator struct {
```
fulghum, line 305: "Commiterator" is a pretty funny portmanteau 😄

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -5024,6 +5064,65 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
+			"insert into t values (3, 'three', 'four');",
+			"set @Commit3 = dolt_commit('-am', 'inserting into t 3');",
+			"insert into t values (4, 'four', 'five');",
+			"set @Commit4 = dolt_commit('-am', 'inserting into t 4');",
+
+			"call dolt_checkout('main');",
+			"insert into t values (5, 'five', 'six');",
+			"set @Commit5 = dolt_commit('-am', 'inserting into t 5');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
```
fulghum, line 5087: Would be nice to spot check some of the output from dolt_log, in addition to asserting the row count is what we expect.

fulghum (approved):
Looks good! Just a few small questions/suggestions. Biggest feedback was that the input values for this function are a little more special that other functions, so really good, easy to understand error messages for input validation errors would be particularly helpful for customers here.

PR #4576, opened 2022-10-19
Add Docker Image support for `linux/arm64`

jennifersp:
Added Docker Image for `linux/arm64` addition to `linux/amd64` architecture. Each will be pushed to DockerHub as separate tags, but both image will be available for `latest` or `<release_version>` tags.


coffeegoddd (approved):
LGTM!

PR #4577, opened 2022-10-19
Add `--not`, `--min-parents`, and `--merges` to dolt_log table function

tbantle22:


go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -142,24 +163,78 @@ func (ltf *LogTableFunction) Expressions() []sql.Expression {
...
+	}
+	ltf.minParents = minParents
+
+	// Gets revisions, excluding any flag-related expression
+	var filteredExpressions []sql.Expression
+	for i, ex := range expression {
+		if !strings.Contains(ex.String(), "--") && !(i > 0 && strings.Contains(expression[i-1].String(), "--")) {
+			filteredExpressions = append(filteredExpressions, ex)
+		}
+	}
```
tbantle22, line 226: Not sure if this is the best way to handle the expressions/flags. We use the arg parser to handle arguments for procedures, but that converts all the arguments to primitive types instead of expressions, which we seem to rely on more heavily in these table functions. Right now this is converting flags to a primitive types and keeps the revisions as sql expressions

fulghum (approved):
Very nice @tbantle22! 👏

PR #4583, opened 2022-10-20
avoid hidden secondary indexes when creating foreign keys that prefix of pk

JCOR11599:
TODO: maybe use empty string instead of PRIMARY for name of index used by foreign key.

When creating foreign keys, we typically create a new secondary index so violations and cascades are handled quickly. 
However, there is no need to create a secondary index when the foreign key references a prefix to an existing primary key.
This PR makes it so that the Primary Key shows up as an Index with the name "PRIMARY".


andrew-wm-arthur (approved):
Generally looks good, couple of comments. 
@Hydrocharged should look as well

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -203,7 +203,7 @@ func (fk ForeignKey) ValidateTableSchema(sch schema.Schema) error {
 			return fmt.Errorf("foreign key `%s` has entered an invalid state, table `%s` has unexpected schema", fk.Name, fk.TableName)
 		}
 	}
-	if !sch.Indexes().Contains(fk.TableIndex) {
+	if !sch.Indexes().Contains(fk.TableIndex) && fk.TableIndex == "PRIMARY" && sch.GetPKCols().Size() == 0 {
```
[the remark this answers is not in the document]
JCOR11599: you're right, it should be
```go
if !sch.Indexes().Contains(fk.TableIndex) || (fk.TableIndex == "PRIMARY" && sch.GetPKCols().Size() == 0) {}
```

go/libraries/doltcore/doltdb/table.go
```diff
@@ -561,6 +561,10 @@ func (t *Table) GetIndexRowData(ctx context.Context, indexName string) (durable.
 		return nil, err
 	}
 
+	if indexName == "PRIMARY" {
```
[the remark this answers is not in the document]
JCOR11599: A couple paths use this, but I think the important one was `newConstraintViolationsLoadedTable()` from `violations_fk.go`.
However, I might be able to remove this, since I added: https://github.com/dolthub/dolt/pull/4583/files#diff-665d6a79567c44e029b188e4cd624a25af54288cb1dbf621adb559597f5564adR506

Maybe using empty string instead of `PRIMARY` for the name of this index would be a good idea? It should be impossible for users to create an index with empty string name, since the name by default is the columns that make up that index.

go/libraries/doltcore/doltdb/foreign_key_coll.go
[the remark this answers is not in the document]
JCOR11599: jk, it should be 
```go
if (fk.TableIndex != "PRIMARY" && !sch.Indexes().Contains(fk.TableIndex)) || (fk.TableIndex == "PRIMARY" && sch.GetPKCols().Size() == 0) {}
```

go/libraries/doltcore/doltdb/foreign_key_coll.go
```diff
@@ -203,7 +203,7 @@ func (fk ForeignKey) ValidateTableSchema(sch schema.Schema) error {
 			return fmt.Errorf("foreign key `%s` has entered an invalid state, table `%s` has unexpected schema", fk.Name, fk.TableName)
 		}
 	}
-	if !sch.Indexes().Contains(fk.TableIndex) {
+	if (fk.TableIndex != "" && !sch.Indexes().Contains(fk.TableIndex)) || (fk.TableIndex == "" && sch.GetPKCols().Size() == 0) {
```
Hydrocharged, line 206: Maybe want to verify that the primary key length is `>=` the column count, rather than just checking that there are primary keys. Something like `sch.GetPKCols().Size() >= len(fk.TableColumns)`

go/libraries/doltcore/schema/index.go
```diff
@@ -68,12 +68,17 @@ type indexImpl struct {
 	comment       string
 }
 
-func NewIndex(name string, tags, allTags []uint64, indexColl *indexCollectionImpl, props IndexProperties) Index {
+func NewIndex(name string, tags, allTags []uint64, indexColl IndexCollection, props IndexProperties) Index {
+	var idxColl *indexCollectionImpl
```
Hydrocharged, line 72: Need a better name. `indexColl` and `idxColl` are the same name. Something like `indexCollImpl`.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1971,76 +1984,80 @@ func (t *AlterableDoltTable) AddForeignKey(ctx *sql.Context, sqlFk sql.ForeignKe
...
-			})
-			if err != nil {
-				return err
-			}
-			tableIndex = idxReturn.NewIndex
-			tbl = idxReturn.NewTable
-			root, err = root.PutTable(ctx, t.tableName, idxReturn.NewTable)
-			if sqlFk.IsSelfReferential() {
-				refTbl = idxReturn.NewTable
+			if !isPkPrefix(sqlFk.Columns, t.sch.GetPKCols().GetColumns()) {
```
Hydrocharged, line 1994: This seems incorrect. This should _only_ get to this state when the engine matched a primary key, as that's the only time that the engine found an index but Dolt does not have the index. So I think this block should assume that the primary key IS the target index. With this check, it seems like there's an opportunity for some other additional index to be created, which would be a bug. We should also keep the comment stating as such (of course updated).

General guidance regarding comments when you're editing code, if you see a comment that no longer applies (and isn't just a "TODO"), it's better to update the comment than to remove it. Generally they're placed in sections that are worth documenting, so removing the comment may make it much harder to reason about some section of code in the future. This is a good example, as we know that this block should be "primary key only" _because_ of the old comment.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1971,76 +1984,80 @@ func (t *AlterableDoltTable) AddForeignKey(ctx *sql.Context, sqlFk sql.ForeignKe
...
 		if !ok {
-			// The engine matched on a primary key, and Dolt does not yet support using the primary key within the
-			// schema.Index interface (which is used internally to represent indexes across the codebase). In the
-			// meantime, we must generate a duplicate key over the primary key.
-			//TODO: use the primary key as-is
-			var refPkTags []uint64
-			for _, i := range refSch.GetPkOrdinals() {
-				refPkTags = append(refPkTags, refSch.GetAllCols().GetByIndex(i).Tag)
-			}
+			if !isPkPrefix(sqlFk.ParentColumns, refSch.GetPKCols().GetColumns()) {
```
Hydrocharged, line 2021: Same remark here as in the comment above.

Hydrocharged (approved):
Overall looks good! Just a few remarks. Especially look at the comment one, it's not a hard rule, but it's one that will save you a _ton_ of time in the future, trust me on that. In general, adding comments is great, but I probably comment more than most people, so I'm sure someone feels I comment _too much_ lol. It's an overall good habit to get into though.

Hydrocharged:
This PR fixes https://github.com/dolthub/dolt/issues/2082

go/libraries/doltcore/sqle/tables.go
JCOR11599: I think the `!ok` from `findIndexWithPrefix` is just when there isn't a `SECONDARY INDEX` that matches; it is still possible that the columns used in the `FOREIGN KEY` are not a prefix of the `PRIMARY KEY`, which is why I added the `isPkPrefix` check, and a new `SECONDARY INDEX` should be made.

I can update the comment to reflect this.

JCOR11599: Nevermind, MySQL only creates indexes for the table, and not the referenced table.

PR #4585, opened 2022-10-20
Branch Control Pt. 9

Hydrocharged:
This PR:
* Moves the branch controller into the context
* Adds the remaining stored procedures
* Handles declaring the super user via CLI arguments or YAML
* Fixes bugs found through additional testing
* Enables branch control globally

Things still missing:
* Bats tests (will mainly copy the engine tests, but makes use of saving/loading from disk)
* Ways to interact with the binlog
* Support for roles

Role support is actually a bit more involved than originally anticipated. It involves coordination with the privilege tables, which were not built to be interacted with outside of their specific GMS context. I'm thinking it should get pushed back to a v2 implementation, unless it's high priority.

### DIFF
--- a/go/cmd/dolt/commands/engine/sqlengine.go
+++ b/go/cmd/dolt/commands/engine/sqlengine.go
@@ -126,9 +126,12 @@ func NewSqlEngine(
 	}
 
 	// Load the branch control permissions, if they exist
-	if err = branch_control.LoadData(sql.NewEmptyContext(), config.BranchCtrlFilePath, config.DoltCfgDirPath); err != nil {
+	var bcController *branch_control.Controller
+	if bcController, err = branch_control.LoadData(config.BranchCtrlFilePath, config.DoltCfgDirPath); err != nil {
 		return nil, err
 	}
+	// Set the server's super user
+	branch_control.SetSuperUser(config.ServerUser, config.ServerHost)
 
 	// Set up engine
 	engine := gms.New(analyzer.NewBuilder(pro).WithParallelism(parallelism).Build(), &gms.Config{
@@ -166,7 +169,7 @@ func NewSqlEngine(
 		dbStates = append(dbStates, dbState)
 	}
 
-	sess, err := dsess.NewDoltSession(sql.NewEmptyContext(), sql.NewBaseSession(), pro, mrEnv.Config(), dbStates...)
+	sess, err := dsess.NewDoltSession(sql.NewEmptyContext(), sql.NewBaseSession(), pro, mrEnv.Config(), bcController, dbStates...)
 	if err != nil {
 		return nil, err
 	}
@@ -185,7 +188,7 @@ func NewSqlEngine(
 	return &SqlEngine{
 		dbs:            nameToDB,
 		contextFactory: newSqlContext(sess, config.InitialDb),
-		dsessFactory:   newDoltSession(pro, mrEnv.Config(), config.Autocommit),
+		dsessFactory:   newDoltSession(pro, mrEnv.Config(), config.Autocommit, bcController),
 		engine:         engine,
 		resultFormat:   format,
 	}, nil
@@ -318,7 +321,6 @@ func newSqlContext(sess *dsess.DoltSession, initialDb string) func(ctx context.C
 		if sessionDB := sess.GetCurrentDatabase(); sessionDB != "" {
 			sqlCtx.SetCurrentDatabase(sessionDB)
 		} else {
-
 			sqlCtx.SetCurrentDatabase(initialDb)
 		}
 
@@ -326,7 +328,8 @@ func newSqlContext(sess *dsess.DoltSession, initialDb string) func(ctx context.C
 	}
 }
 
-func newDoltSession(pro dsqle.DoltDatabaseProvider, config config.ReadWriteConfig, autocommit bool) func(ctx context.Context, mysqlSess *sql.BaseSession, dbs []sql.Database) (*dsess.DoltSession, error) {
+func newDoltSession(pro dsqle.DoltDatabaseProvider, config config.ReadWriteConfig,
+	autocommit bool, bc *branch_control.Controller) func(ctx context.Context, mysqlSess *sql.BaseSession, dbs []sql.Database) (*dsess.DoltSession, error) {
 	return func(ctx context.Context, mysqlSess *sql.BaseSession, dbs []sql.Database) (*dsess.DoltSession, error) {
 		ddbs := dsqle.DbsAsDSQLDBs(dbs)
 		states, err := getDbStates(ctx, ddbs)
@@ -334,7 +337,7 @@ func newDoltSession(pro dsqle.DoltDatabaseProvider, config config.ReadWriteConfi
 			return nil, err
 		}
 
-		dsess, err := dsess.NewDoltSession(sql.NewEmptyContext(), mysqlSess, pro, config, states...)
+		dsess, err := dsess.NewDoltSession(sql.NewEmptyContext(), mysqlSess, pro, config, bc, states...)
 		if err != nil {
 			return nil, err
 		}
--- a/go/cmd/dolt/commands/sqlserver/server.go
+++ b/go/cmd/dolt/commands/sqlserver/server.go
@@ -138,16 +138,17 @@ func Serve(
 
 	// Create SQL Engine with users
 	config := &engine.SqlEngineConfig{
-		InitialDb:         "",
-		IsReadOnly:        serverConfig.ReadOnly(),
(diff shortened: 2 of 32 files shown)


go/libraries/doltcore/branch_control/branch_control.go
```diff
@@ -228,28 +256,58 @@ func CanCreateBranch(ctx context.Context, branchName string) error {
...
+	}
+	controller := branchAwareSession.GetController()
+	if controller == nil {
+		return nil
+	}
+
+	user := branchAwareSession.GetUser()
+	host := branchAwareSession.GetHost()
+	// Check if we already have admin permissions for the given branch, as there's no need to do another insertion if so
+	controller.Access.RWMutex.RLock()
```
zachmu, line 299: This is a bad pattern. The access object should manage its own concurrency needs without requiring locking on the outside like this.

go/libraries/doltcore/branch_control/branch_control.go
```diff
@@ -228,28 +256,58 @@ func CanCreateBranch(ctx context.Context, branchName string) error {
...
 	if (perms&Permissions_Write == Permissions_Write) || (perms&Permissions_Admin == Permissions_Admin) {
 		return nil
 	}
 	return ErrCannotDeleteBranch.New(user, host, branchName)
 }
 
+// AddAccessEntryForContext adds an entry in the access table for the user represented by the given context. If the
+// context is missing some functionality that is needed to perform the addition, such as a user or the Controller, then
+// this simply returns.
+func AddAccessEntryForContext(ctx context.Context, branchName string) error {
```
zachmu, line 286: This adds Admin permissions if not present, so the name should reflect that

Maybe `GrantBranchAdmin`

go/libraries/doltcore/branch_control/branch_control.go
```diff
@@ -55,95 +57,122 @@ type Controller struct {
...
 	}
-	StaticController = CreateControllerWithSuperUser(context.Background(), "root", "localhost")
 }
 
-// CreateController returns an empty *Controller.
-func CreateController(ctx context.Context) *Controller {
-	return CreateControllerWithSuperUser(ctx, "", "")
+// IsSuperUser returns whether the given user and host combination is the super user.
+func IsSuperUser(user string, host string) bool {
+	if user == superUser && ((host == superHost) || (superHostIsLoopback && net.ParseIP(host).IsLoopback())) {
```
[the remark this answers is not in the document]
zachmu: We're doing this check on every insert, so this matters for benchmarks

go/libraries/doltcore/sqle/dfunctions/verify_constraints.go
```diff
@@ -55,6 +56,9 @@ func (vc *ConstraintsVerifyFunc) Eval(ctx *sql.Context, row sql.Row) (interface{
 }
 
 func DoDoltConstraintsVerify(ctx *sql.Context, args []string) (int, error) {
+	if err := branch_control.CheckAccess(ctx, branch_control.Permissions_Write); err != nil {
```
zachmu, line 59: Does this need write?

go/libraries/doltcore/sqle/dprocedures/dolt_remote.go
```diff
@@ -46,6 +47,9 @@ func doDoltRemote(ctx *sql.Context, args []string) (int, error) {
 	if len(dbName) == 0 {
 		return 1, fmt.Errorf("Empty database name.")
 	}
+	if err := branch_control.CheckAccess(ctx, branch_control.Permissions_Write); err != nil {
```
zachmu, line 50: This is overly broad, the remote command doesn't actually change the branch head in most cases

go/libraries/doltcore/sqle/dprocedures/dolt_tag.go
```diff
@@ -40,6 +41,9 @@ func doDoltTag(ctx *sql.Context, args []string) (int, error) {
 	if len(dbName) == 0 {
 		return 1, fmt.Errorf("Empty database name.")
 	}
+	if err := branch_control.CheckAccess(ctx, branch_control.Permissions_Write); err != nil {
```
zachmu, line 44: Tags are separate from branches, no reason to protect afaict

zachmu (approved):
LGTM

go/cmd/dolt/commands/engine/sqlengine.go
```diff
@@ -318,23 +321,23 @@ func newSqlContext(sess *dsess.DoltSession, initialDb string) func(ctx context.C
...
-
 			sqlCtx.SetCurrentDatabase(initialDb)
 		}
 
 		return sqlCtx, nil
 	}
 }
 
-func newDoltSession(pro dsqle.DoltDatabaseProvider, config config.ReadWriteConfig, autocommit bool) func(ctx context.Context, mysqlSess *sql.BaseSession, dbs []sql.Database) (*dsess.DoltSession, error) {
+func newDoltSession(bc *branch_control.Controller, pro dsqle.DoltDatabaseProvider, config config.ReadWriteConfig,
```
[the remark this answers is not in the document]
Hydrocharged: Sure can!

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -138,16 +138,17 @@ func Serve(
...
-		ServerUser:        serverConfig.User(),
-		ServerPass:        serverConfig.Password(),
-		ServerHost:        serverConfig.Host(),
-		Autocommit:        serverConfig.AutoCommit(),
-		JwksConfig:        serverConfig.JwksConfig(),
-		ClusterController: clusterController,
+		InitialDb:          "",
+		IsReadOnly:         serverConfig.ReadOnly(),
+		PrivFilePath:       serverConfig.PrivilegeFilePath(),
+		BranchCtrlFilePath: serverConfig.BranchControlFilePath(),
```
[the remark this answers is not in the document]
Hydrocharged: I'm only working on the server implementation for now, multi-db support comes later.

go/libraries/doltcore/branch_control/branch_control.go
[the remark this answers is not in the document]
Hydrocharged: Checks `user == superUser && host == superHost` first. For benchmarks these should match so we're not doing the IP parsing. For other cases we're not going to match on the user at all. Should be rare to actually require an IP check.

go/libraries/doltcore/branch_control/branch_control.go
Hydrocharged: I can take another pass at it, but it definitely complicated the model. For example, when inserting a row, we also read the table. Right now the insertion function on the table just gets a `Lock`. Changing it so that each function acquires a lock means you can get into a state where we read a row to update and it's changed by another user before we can do anything with it. So the way around that is to basically move all of the dolt_table logic to the `Access` struct, and same for `Namespace`.

Like if no `Access` functions acquire a lock (except `Serialize/Deserialize` which won't be nested in anything), then we can easily manage it from the layers above (`branch_control.go` and the dolt_tables). I really think this keeps everything a bit more straightforward.

go/libraries/doltcore/sqle/dfunctions/verify_constraints.go
Hydrocharged: I think so, verifying constraints can modify the system tables for that branch, so it _can_ do a write operation on the branch.

go/libraries/doltcore/sqle/dprocedures/dolt_remote.go
Hydrocharged: But it does modify branch information, hence the guard. You don't want someone to just add or remove remotes without permission.

go/libraries/doltcore/sqle/dprocedures/dolt_tag.go
Hydrocharged: That's fair. Since tags are writing _something_ I decided to gate it, but I can remove the check for them.

go/libraries/doltcore/sqle/dprocedures/dolt_remote.go
zachmu: Remotes are separate from branches. What am I missing?

Hydrocharged: Oh I’m thinking of tracking remotes, not the remotes themselves. My bad.

go/cmd/dolt/commands/engine/sqlengine.go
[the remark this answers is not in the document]
Hydrocharged: Removing it caused failing tests, so we can do that at another time.

PR #4590, opened 2022-10-20
go/doltcore/schema: Added doc comments to `schema.Schema`

andrew-wm-arthur:


go/libraries/doltcore/schema/schema.go
```diff
@@ -26,16 +26,34 @@ import (
...
 type Schema interface {
-	// GetPKCols gets the collection of columns which make the primary key. They
-	// are always returned in ordinal order.
+	// GetPKCols gets the collection of columns which make the primary key.
+	// Columns in this collection are ordered by storage order, which is
+	// defined in the 'PRIMARY KEY(...)' clause of a CREATE TABLE statement.
 	GetPKCols() *ColCollection
 
 	// GetNonPKCols gets the collection of columns which are not part of the primary key.
+	// Columns in this collection are ordered by schema order (display order), which is
```
max-hoffman, line 50: are there instances where nonPkCols schema and storage order is not the same?

max-hoffman (approved):
LGTM, one small comment

go/libraries/doltcore/schema/schema.go
andrew-wm-arthur: currently, we always store them in schema order. In the future we'd like to reorder them to place fixed-width columns first to enable fast-access optimizaitons

PR #4593, opened 2022-10-20
tests for adding pk to keyless table

JCOR11599:
companion PR; https://github.com/dolthub/go-mysql-server/pull/1343


go/libraries/doltcore/sqle/enginetest/ddl_queries.go
```diff
@@ -707,3 +707,61 @@ var AddIndexScripts = []queries.ScriptTest{
 		},
 	},
 }
+
+var AddPrimaryKeyScripts = []queries.ScriptTest{
```
zachmu, line 711: These aren't specific to dolt, they should go in GMS

PR #4594, opened 2022-10-20
Make dolt log work with `^` ancestor spec

tbantle22:


integration-tests/bats/log.bats
```diff
@@ -89,6 +89,22 @@ teardown() {
     [[ ! "$output" =~ "BRANCH1" ]] || false
     run dolt log main..main
     [ $status -eq 0 ]
+    
+    run dolt log main^..branch1
+    [ $status -eq 0 ]
+    [[ ! "$output" =~ "MAIN" ]] || false
+    [[ ! "$output" =~ "AFTER" ]] || false
```
fulghum, line 96: I was expecting `AFTER` would be in the results... `main` looks like it's currently pointing at the commit with the `AFTER` commit message, so `main^` should move the exclusion commit to the parent of `main`, and cause `main`'s current commit to be included in the log. I might just be misreading something though...

tbantle22: So if the commit graph looks like this:
```
A - B - F (main)
     \
       C - D - E (branch1)
```
`dolt log main..branch` is essentially `dolt log F..E` (all `branch1`, excluding commits from `main` - `E`, `D`, `C`)

 And then `dolt log main^..branch` is `dolt log B..E` (all `branch1`, excluding commits reachable from `B`, which is still  `E`, `D`, `C` because `branch1` does not reach `F`)

fulghum: ahhh... yes! Thank you for drawing it out and double checking that. Figured I must just be missing something if everything else was working. Looks great!

fulghum (approved):
Looks great! ✨

PR #4596, opened 2022-10-21
Made `server_query` error on bad queries where we expect no results. 

timsehn:
Then fixed all the broken bats tests.


zachmu (approved):
LGTM

PR #4613, opened 2022-10-25
Fixing variable name in `getPORT` helper function

fulghum:
I noticed some errors with grep syntax in a failed BATS test. Looks like we just had an empty env var so there wasn't anything passed to `grep`, which caused errors and prevented us from fully checking if a port is in use.

### DIFF
--- a/integration-tests/bats/helper/query-server-common.bash
+++ b/integration-tests/bats/helper/query-server-common.bash
@@ -256,7 +256,7 @@ definePORT() {
   for i in {0..9}
   do
     let getPORT="($$ + $i) % (65536-1024) + 1024"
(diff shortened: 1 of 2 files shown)


jennifersp (approved):
Ah looks good! Thank you for catching this! 😄

timsehn:
I saw this too and couldn't figure out what was going on. Jason wins a neckbeard award! 🧔🏻‍♂️

jennifersp:
Sorry, I just found out an issue I missed that we start a server with `definePORT` and when we run query on the server we also use `definePORT`, which will get the next available port number instead of the one being in use, so we should not define PORT in `server_query()` method. https://github.com/dolthub/dolt/pull/4613/files#diff-d086554194f4def391e5db4be455ce037199863749f8e3c0b13e011a9fe88964R240

PR #4619, opened 2022-10-25
[dolt migrate]: don't prefetch leaf nodes when validating migrated data

druvv:
Bounds memory usage to size of table file indices and number of cpu cores

Migrating https://www.dolthub.com/repositories/dolthub/museum-collections pre-pr with 8 cores and 64gb of ram takes 5,049 seconds. 

Post-pr it takes  6,017 seconds on the same machine.

Only a `(5049 - 6017) / 6017 = -16.09%` performance hit. 

Previously memory usage was spiking to 42gb, now it remains under 2gb.


go/libraries/doltcore/sqle/rows.go
```diff
@@ -247,15 +245,79 @@ func DoltTablePartitionToRowIter(ctx *sql.Context, name string, table *doltdb.Ta
...
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
-	iter, err := newRowIterator(ctx, table, pkSch.Schema, nil, p)
+	return pkSch.Schema, rowIter, nil
+}
+
+type limitingMapIterator struct {
```
andrew-wm-arthur, line 292: this makes more sense in pkg `types`

andrew-wm-arthur (approved):
LGTM!

max-hoffman:
#benchmark

PR #4626, opened 2022-10-26
more permissive constraints and types when importing

JCOR11599:
There was a change added that [samples the rows](https://github.com/dolthub/dolt/pull/4471/files) when doing large imports, but this can incorrectly give columns a `NOT NULL` constraint. We can either not sample tables or just not apply the `NOT NULL` constraintl. I chose to not apply the `NOT NULL` constraint. Additionally, the sampling change makes it possible to incorrectly make certain columns `UNISIGNED INT`, when they contain negative numbers. I changed it so infer will always treat columns as signed, so integers that are too large/small for signed int will become strings.

fix for: https://github.com/dolthub/dolt/issues/4620

### DIFF
--- a/go/libraries/doltcore/env/actions/infer_schema.go
+++ b/go/libraries/doltcore/env/actions/infer_schema.go
@@ -61,7 +61,6 @@ func InferColumnTypesFromTableReader(ctx context.Context, rd table.ReadCloser, a
 
 	var curr, prev row.Row
 	i := newInferrer(rd.GetSchema(), args)
-
 OUTER:
 	for j := 0; true; j++ {
 		var err error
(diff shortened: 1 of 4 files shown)


zachmu (approved):
LGTM, generally supportive of being more permissive on import or when inferring a schema

liuliu-dev:
does this also fix this failing import? https://dolthub.awsdev.ld-corp.com/repositories/liuliu/merge_test/jobs/39b25658-d820-47db-91f5-0885638867ad
it's mentioned in https://github.com/dolthub/dolthub-issues/issues/353 and being linked in https://github.com/dolthub/dolt/issues/4620, didn't paste a dolt repro tho.

JCOR11599:
@liuliu-dev Good catch, this probably does not fix that one. I'll look into it. Thanks!

zachmu (approved):
LGTM

PR #4630, opened 2022-10-26
[dolt migrate]: Validate ws migration, skip table validation if table did not change

druvv:


go/libraries/doltcore/migrate/validation.go
```diff
@@ -67,6 +67,25 @@ func validateRootValue(ctx context.Context, old, new *doltdb.RootValue) error {
...
+				return err
+			}
+			oldParentHash, err := op.HashOf()
+			if err != nil {
+				return err
+			}
+			if oldHash.Equal(oldParentHash) {
+				continue
+			}
+		}
```
druvv, line 87: Maybe I should check that the new parent commit and the new commit's table has the same hash?

andrew-wm-arthur: this is fine

andrew-wm-arthur (approved):

PR #4631, opened 2022-10-26
go/doltcore/schema: Set primary key ordinals on schema migration

andrew-wm-arthur:


go/libraries/doltcore/migrate/transform.go
```diff
@@ -481,10 +481,19 @@ func migrateSchema(ctx context.Context, tableName string, existing schema.Schema
 		}
 	}
 
-	if patched {
-		return schema.SchemaFromCols(schema.NewColCollection(cols...))
+	if !patched {
+		return existing, nil
+	}
+
+	sch, err := schema.SchemaFromCols(schema.NewColCollection(cols...))
```
druvv, line 488: Maybe we shouldn't export SchemaFromCols if it's so important to also set the PK ordinals. What if we un-exported the function, and created a new function that accepted the column collection and the ordinals?

druvv (approved):
lgtm!

go/libraries/doltcore/migrate/transform.go
andrew-wm-arthur: yeah, that would be significant improvement

PR #4636, opened 2022-10-26
Support --amend in dolt_commit

PavelSafronov:
Fixes https://github.com/dolthub/dolt/issues/4595: dolt_commit procedure does not support --amend


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -468,14 +468,39 @@ func (d *DoltSession) NewPendingCommit(ctx *sql.Context, dbName string, roots do
...
 	var mergeParentCommits []*doltdb.Commit
 	if sessionState.WorkingSet.MergeActive() {
 		mergeParentCommits = []*doltdb.Commit{sessionState.WorkingSet.MergeState().Commit()}
+	} else if props.Amend {
+		var parentsHeadForAmend []*doltdb.Commit
+		numParentsHeadForAmend := headCommit.NumParents()
+		for i := 0; i < numParentsHeadForAmend; i++ {
+			parentCommit, err := headCommit.GetParent(ctx, i)
+			if err == nil {
+				parentsHeadForAmend = append(parentsHeadForAmend, parentCommit)
```
fulghum, line 483: It doesn't look like we ever read the contents of `parentsHeadForAmend`? Seems like we should be passing it to the commit staging action. 

We should add a test for merge commit handling to make sure we preserve both parents in the new commit; I believe we had a bug with preserving both parents for a merge commit in the initial contribution for the CLI, so would be good to make sure we test that for the SQL path, too.

fulghum (commented):
Nice work! Looks very good. 

Just a few minor comments and some other test suggestions. We should also add a quick sanity test for `--amend` in `sql-commit.bats` just so that we've got a little bit of coverage there, too.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -7401,9 +7401,24 @@ var DoltCommitTests = []queries.ScriptTest{
 				Expected: []sql.Row{{sql.NewOkResult(1)}},
 			},
 			{
-				Query:            "CALL DOLT_COMMIT('-ALL', '-m', 'update table t');",
+				Query:            "CALL DOLT_COMMIT('-ALL', '-m', 'update table terminator');",
 				SkipResultsCheck: true,
 			},
+			// check initial commit
```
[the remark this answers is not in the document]
PavelSafronov: Updating.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -7401,9 +7401,24 @@ var DoltCommitTests = []queries.ScriptTest{
...
+				Query:    "select message from dolt_log limit 1",
+				Expected: []sql.Row{{"update table terminator"}},
+			},
+			// amend the commit
+			{
+				Query:            "CALL DOLT_COMMIT('-amend', '-m', 'update table t');",
+				SkipResultsCheck: true,
+			},
+			// check amended commit
+			{
```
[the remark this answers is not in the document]
PavelSafronov: Good idea, will add a few more commit/amend combinations.

Regarding the `dolt_status`, `dolt_log`, and `dolt_diff_<tablename> tables: I've been able to run the same steps through CLI and SQL (`dolt commit --amend ...` vs `call dolt_commit('--amend'...)`) and I can see that both are behaving identically. But that's just some manual testing. Should I try to add tests that verify the data in these various tables?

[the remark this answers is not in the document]
fulghum: Thank you! Yeah... my thinking is that since `--amend` allows you to change several things (e.g. commit message, adding included files, removing included files) we should test a little deeper to make sure it's working as expected. Some of those system tables are likely helpful to writing those tests to verify, but let me know if you get stuck and can't find the data you need.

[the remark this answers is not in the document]
PavelSafronov: Added more tests, including updating the commit message, adding an extra change to a commit, and updating a message in a merge commit. Let me know if these new tests were added correctly/in the right spot.

I'm not sure how to test "removing included files", can you share what that operation might look like?

go/libraries/doltcore/sqle/dsess/session.go
PavelSafronov: Right you are, this was a bug. Adding the extra tests (specifically, amending a merge commit) surfaced this issue. It's fixed now.

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -468,14 +468,38 @@ func (d *DoltSession) NewPendingCommit(ctx *sql.Context, dbName string, roots do
...
+	headCommit := sessionState.headCommit
+	headHash, _ := headCommit.HashOf()
+
 	var mergeParentCommits []*doltdb.Commit
 	if sessionState.WorkingSet.MergeActive() {
 		mergeParentCommits = []*doltdb.Commit{sessionState.WorkingSet.MergeState().Commit()}
+	} else if props.Amend {
+		numParentsHeadForAmend := headCommit.NumParents()
+		for i := 0; i < numParentsHeadForAmend; i++ {
+			parentCommit, err := headCommit.GetParent(ctx, i)
```
fulghum, line 480: We should add handling for when this `err` is not nil. Swallowed errors are a pain to debug!

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Query:            "INSERT INTO test (id) VALUES (2)",
+				SkipResultsCheck: true,
+			},
+			{
+				Query:    "SELECT COUNT(*) FROM dolt_status;",
+				Expected: []sql.Row{{1}},
+			},
+			{
+				Query:            "CALL DOLT_ADD('.');",
+				SkipResultsCheck: true,
```
fulghum, line 94: Try to only use `SkipResultsCheck` when absolutely necessary (e.g. when a commit hash is being returned and we can't reasonably know ahead of time what value to expect). 

For these `CALL` stored procedure calls, since they should return 0 on success, you should be able to expect: 
`[]sql.Row{{0}}` 

For the `Insert into` statements, you should be able to expect something like:
`[]sql.Row{{sql.NewOkResult(1)}},`

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
```diff
@@ -0,0 +1,189 @@
...
+					{"Initialize data repository"},
+				},
+			},
+		},
+	})
+}
+func TestCommitChangesAmend(t *testing.T) {
+	harness := newDoltHarness(t)
+	enginetest.TestScript(t, harness, queries.ScriptTest{
+		Name: "Sysbench Transactions Shouldn't Cause Constraint Violations",
```
fulghum, line 63: Looks like this test name is copy/pasted and still needs to be filled in.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Expected: []sql.Row{
+					{"modified commit message"},
+					{"checkpoint enginetest database mydb"},
+					{"Initialize data repository"},
+				},
+			},
+		},
+	})
+}
+func TestCommitChangesAmend(t *testing.T) {
```
fulghum, line 60: These tests look great. I would just recommend moving all of these over to: `DoltCommitTests`

That var lives in the `dolt_queries.go` file. We generally try to keep the engine tests to be all contained in `dolt_engine_tests.go`, which has top level methods for `TestDoltCommits` and `TestDoltMerges` already, so hooking into those will just help keep these new tests organized as part of those existing test collections.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
```diff
@@ -0,0 +1,189 @@
...
+				SkipResultsCheck: true,
+			},
+			{
+				Query:            "CALL DOLT_COMMIT('--amend', '-m', 'updated commit message');",
+				SkipResultsCheck: true,
+			},
+			{
+				Query:    "SELECT COUNT(*) FROM dolt_status;",
+				Expected: []sql.Row{{0}},
+			},
```
fulghum, line 127: This is checking that the table changes aren't staged anymore, which is good, but it would be better to test that the commit itself contains the data change you added. I think the easiest way to do that would be to look into the `dolt_diff_test` system table, which will show you the exact data changes made in each commit. 

This is a pretty cool system table to know about. Here's some docs on it: https://docs.dolthub.com/sql-reference/version-control/dolt-system-tables#dolt_diff_usdtablename

Maybe something like `select to_id, from_id from dolt_diff_test limit 1` and then assert that `to_id` is 3 and `from_id` is NULL.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
```diff
@@ -0,0 +1,189 @@
...
+					{"checkpoint enginetest database mydb"},
+					{"Initialize data repository"}},
+			},
+			{
+				Query:            "SET @hash=(SELECT commit_hash FROM dolt_log LIMIT 1);",
+				SkipResultsCheck: true,
+			},
+			{
+				Query:    "SELECT COUNT(parent_hash) FROM dolt_commit_ancestors WHERE commit_hash= @hash;",
+				Expected: []sql.Row{{2}},
```
fulghum, line 185: Nice! Great use of capturing a SQL var to test the data. 😄

fulghum (commented):
Looking very good! Just a few really minor comments on the tests. 

It would be good to add something in the BATS tests to cover this, too. There's an existing `sql-commit.bats` file where you could add in some tests for this. I don't think you need to be as exhaustive in BATS, but it's definitely useful to have some level of testing there, since it goes end to end using the CLI and fairly regularly finds issues for us. I would recommend at least adding a test there that covers amending the commit message and amending the included table data changes.

go/libraries/doltcore/sqle/dsess/session.go
PavelSafronov: Good point, adding handling code.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
PavelSafronov: Excellent point and thanks for the suggestions, removing all the possible `SkipResultsCheck`.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
PavelSafronov: Yup! :(
Fixing.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
PavelSafronov: Perfect, moving the tests into that collection.

go/libraries/doltcore/sqle/enginetest/commit_amend_test.go
PavelSafronov: Excellent advice, adding dolt_diff_BLAH checks for these tests.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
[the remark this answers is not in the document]
PavelSafronov: Discussed offline, latest update has the updated tests.

fulghum (approved):
Looks great! Nice work. 

I think you can remove the `commit_amend_test.go` file, since those tests have been moved into the existing `DoltCommitScripts` tests.

PR #4638, opened 2022-10-26
go/libraries/doltcore/sqle/cluster: Add support for configured tls_{cert,key,ca} on the cluster.remotesapi.

reltuk:


go/cmd/dolt/commands/sqlserver/yaml_config.go
```diff
@@ -517,9 +517,24 @@ func (c *ClusterYAMLConfig) RemotesAPIConfig() cluster.RemotesAPIConfig {
 }
 
 type clusterRemotesAPIYAMLConfig struct {
-	P int `yaml:"port"`
+	Port_    int    `yaml:"port"`
```
zachmu, line 520: What's with the underscores? Prevents conflict between field and method name?

zachmu: Cursing golang for not giving us a yaml parser that can see private fields

go/libraries/doltcore/dbfactory/grpc.go
```diff
@@ -36,7 +36,7 @@ var GRPCDialProviderParam = "__DOLT__grpc_dial_provider"
 
 // GRPCDialProvider is an interface for getting a *grpc.ClientConn.
 type GRPCDialProvider interface {
-	GetGRPCDialParams(grpcendpoint.Config) (string, []grpc.DialOption, error)
+	GetGRPCDialParams(grpcendpoint.Config) (string, []grpc.DialOption, grpcendpoint.HTTPFetcher, error)
```
zachmu, line 39: My main comment here is that 3 is generally the limit for return parameters, any more than that and you're straying into really hard to understand territory.

What about returning a struct / new interface that has needed methods on it?

zachmu (approved):
Mostly fine, but you were right that the interface you modified is getting a bit narsty

PR #4639, opened 2022-10-27
Add sleeps to a couple bats tests that were starting a SQL Server

timsehn:
Give the server a chance to start up. Was failing weirdly on my Mac without this.


fulghum (approved):
Nice! I'm surprised the other spots where we start a sql-server didn't also need the same treatment.

PR #4641, opened 2022-10-27
integration-tests/bats/sql-server.bats: Move the temporary table session specific test to go-sql-server-driver.

reltuk:


timsehn:
Ship it.

timsehn (approved):

PR #4644, opened 2022-10-27
Support two and three dot syntax in `dolt_diff` table function

tbantle22:
Two dot: `dolt_diff('main..feature', 'table')` (equivalent to `dolt_diff('main', 'feature', 'table')`)
Three dot: `dolt_diff('main...feature', 'table')`

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_diff_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_diff_table_function.go
@@ -157,10 +157,6 @@ func (dtf *DiffTableFunction) RowIter(ctx *sql.Context, _ sql.Row) (sql.RowIter,
 		return nil, err
 	}
 
-	if len(fromCommitStr) == 0 || len(toCommitStr) == 0 {
-		return nil, fmt.Errorf("expected strings for from and to revisions, got: %v, %v", fromCommitStr, toCommitStr)
-	}
-
 	ddb := sqledb.GetDoltDB()
 	dp := dtables.NewDiffPartition(dtf.tableDelta.ToTable, dtf.tableDelta.FromTable, toCommitStr, fromCommitStr, dtf.toDate, dtf.fromDate, dtf.tableDelta.ToSch, dtf.tableDelta.FromSch)
 
@@ -201,10 +197,6 @@ func loadDetailsForRefs(ctx *sql.Context, fromRef, toRef, dotRef interface{}, db
 		return nil, nil, err
 	}
 
-	if len(fromCommitStr) == 0 || len(toCommitStr) == 0 {
-		return nil, nil, fmt.Errorf("must have valid from and to refs: %s, %s", fromCommitStr, toCommitStr)
-	}
-
 	sess := dsess.DSessFromSess(ctx.Session)
 
 	fromDetails, err := resolveRoot(ctx, sess, db.Name(), fromCommitStr)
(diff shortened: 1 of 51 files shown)


go/libraries/doltcore/sqle/dolt_diff_table_function.go
```diff
@@ -99,16 +103,27 @@ func (dtf *DiffTableFunction) WithExpressions(expression ...sql.Expression) (sql
 		}
 	}
 
-	dtf.fromCommitExpr = expression[0]
-	dtf.toCommitExpr = expression[1]
-	dtf.tableNameExpr = expression[2]
+	if strings.Contains(expression[0].String(), "..") {
```
fulghum, line 106: Seems like this will throw an error without the `len(expression)` check if no expressions are passed in? Would be good to add a quick test assertion for `select * from dolt_diff()` returning an invalid argument error.

go/libraries/doltcore/sqle/dolt_diff_table_function.go
```diff
@@ -127,29 +142,33 @@ func (dtf *DiffTableFunction) RowIter(ctx *sql.Context, _ sql.Row) (sql.RowIter,
...
 
-	dp := dtables.NewDiffPartition(dtf.tableDelta.ToTable, dtf.tableDelta.FromTable, toHash, fromHash, dtf.toDate, dtf.fromDate, dtf.tableDelta.ToSch, dtf.tableDelta.FromSch)
+	fromCommitStr, toCommitStr, err := loadCommitStrings(ctx, fromCommitVal, toCommitVal, dotCommitVal, sqledb)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(fromCommitStr) == 0 || len(toCommitStr) == 0 {
+		return nil, fmt.Errorf("expected strings for from and to revisions, got: %v, %v", fromCommitStr, toCommitStr)
+	}
```
fulghum, line 162: Should `loadCommitStrings` throw this error? It looks like we do this check both times `loadCommitStrings` is called, although I do see you are using different error messages. Just wondering if it's possible to fold them into the same error message and not have to check this in two places.

go/libraries/doltcore/sqle/dolt_diff_table_function.go
```diff
@@ -127,29 +142,33 @@ func (dtf *DiffTableFunction) RowIter(ctx *sql.Context, _ sql.Row) (sql.RowIter,
...
-	fromHash, fromOk := fromCommit.(string)
-	toHash, toOk := toCommit.(string)
-	if !fromOk || !toOk {
-		return nil, fmt.Errorf("expected strings for from and to revisions, got: %v, %v", fromHash, toHash)
-	}
-
 	sqledb, ok := dtf.database.(Database)
 	if !ok {
-		panic("unable to get dolt database")
+		return nil, fmt.Errorf("unable to get dolt database")
```
fulghum, line 152: Nice!! Thanks for removing that `panic`!!

fulghum (approved):
Looks great! Awesome to see this syntax supported in `dolt_diff()` now! 😺

go/libraries/doltcore/sqle/dolt_diff_table_function.go
tbantle22: I added that check/test in my `dolt_diff_summary` PR [here](https://github.com/dolthub/dolt/pull/4646/files#diff-51ab7d0a893924f00c2f31c46b1b61a64af3f1ea9a7a9ddda7cdf885d61094e1R4506). Will merge that into this one merge I merge to main

PR #4646, opened 2022-10-27
Support two and three dot syntax in `dolt_diff_summary` table function

tbantle22:
Two dot: `dolt_diff_summary('main..feature', 'table')` (equivalent to `dolt_diff_summary('main', 'feature', 'table')`)
Three dot: `dolt_diff_summary('main...feature', 'table')`

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
@@ -299,8 +299,8 @@ func (ds *DiffSummaryTableFunction) RowIter(ctx *sql.Context, row sql.Row) (sql.
 	return NewDiffSummaryTableFunctionRowIter(diffSummaries), nil
 }
 
-// evaluateArguments returns fromCommitValStr, toCommitValStr and tableName.
-// It evaluates the argument expressions to turn them into values this DiffTableFunction
+// evaluateArguments returns fromCommitVal, toCommitVal, dotCommitVal, and tableName.
+// It evaluates the argument expressions to turn them into values this DiffSummaryTableFunction
 // can use. Note that this method only evals the expressions, and doesn't validate the values.
 func (ds *DiffSummaryTableFunction) evaluateArguments() (interface{}, interface{}, interface{}, string, error) {
 	var tableName string
--- a/go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"strings"
 
 	"github.com/dolthub/go-mysql-server/sql"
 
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -84,16 +86,29 @@ func (ds *DiffSummaryTableFunction) FunctionName() string {
...
 // String implements the Stringer interface
 func (ds *DiffSummaryTableFunction) String() string {
+	if ds.dotCommitExpr != nil {
+		if ds.tableNameExpr != nil {
+			return fmt.Sprintf("DOLT_DIFF_SUMMARY(%s, %s)", ds.dotCommitExpr.String(), ds.tableNameExpr.String())
+		}
+		return fmt.Sprintf("DOLT_DIFF_SUMMARY(%s)", ds.dotCommitExpr.String())
+	}
 	if ds.tableNameExpr != nil {
 		return fmt.Sprintf("DOLT_DIFF_SUMMARY(%s, %s, %s)", ds.fromCommitExpr.String(), ds.toCommitExpr.String(), ds.tableNameExpr.String())
```
fulghum, line 113: (minor) I like that you checked for `ds.tableNameExpr` above to avoid printing the tablename when it's not specified. Would be nice to do the same treatment here.

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
```diff
@@ -259,39 +302,40 @@ func (ds *DiffSummaryTableFunction) RowIter(ctx *sql.Context, row sql.Row) (sql.
 // evaluateArguments returns fromCommitValStr, toCommitValStr and tableName.
```
fulghum, line 302: Would be good to update this doc text, especially since the return types aren't very self-descriptive (`interface{}`, `string`).

go/libraries/doltcore/sqle/dolt_diff_table_function.go
```diff
@@ -94,6 +94,10 @@ func (dtf *DiffTableFunction) Expressions() []sql.Expression {
 
 // WithExpressions implements the sql.Expressioner interface
 func (dtf *DiffTableFunction) WithExpressions(expression ...sql.Expression) (sql.Node, error) {
+	if len(expression) < 2 {
+		return nil, sql.ErrInvalidArgumentNumber.New(dtf.FunctionName(), "2 to 3", len(expression))
+	}
+
```
fulghum, line 100: Nice catch!

fulghum (approved):
Looks great! 🚀

go/libraries/doltcore/sqle/dolt_diff_summary_table_function.go
fulghum: oops! nevermind... the last line of this function was cut off and I didn't see where it handled the two arg form and thought it was here. my bad!

PR #4656, opened 2022-10-28
Improve dump-docs error and add bats, fix diff docs

tbantle22:


fulghum (approved):
Looks great! Awesome to have that BATS test to catch these problems asap!

andrew-wm-arthur (approved):
LGTM!

PR #4686, opened 2022-11-02
Prisma compatibility smoke testing 

fulghum:
Prisma is a popular ORM library for TypeScript/JavaScript applications, with around 25k GitHub stars. 

This PR extends our nascent ORM testing image to run a simple smoke test for Prisma. 

Next, we need to get these tests running as part of CI and would be great to do deeper testing with Prisma and other ORM libraries, ideally running their test suites against a Dolt sql-server.


jennifersp (approved):
Looks good!

PR #4691, opened 2022-11-02
Support three dot log

tbantle22:
CLI:
`dolt log main...branchA`
`dolt log main branchA --not $(dolt merge-base main branchA)`

SQL:
`dolt_log('main...branchA')`

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_log_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_log_table_function.go
@@ -536,7 +536,8 @@ func expressionToString(ctx *sql.Context, expr sql.Expression) (string, error) {
 	return valStr, nil
 }
 
-// Gets revisionName and/or secondRevisionName from sql expression
+// getRevisionsFromExpr returns the revisionName and/or secondRevisionName, as
+// well as a threeDot boolean from sql expression
 func getRevisionsFromExpr(ctx *sql.Context, expr sql.Expression, canDot bool) (string, string, bool, error) {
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -512,23 +536,27 @@ func expressionToString(ctx *sql.Context, expr sql.Expression) (string, error) {
 	return valStr, nil
 }
 
-// Gets revisionName and/or excludingRevisionName from sql expression
-func getRevisionsFromExpr(ctx *sql.Context, expr sql.Expression, canDot bool) (string, string, error) {
+// Gets revisionName and/or secondRevisionName from sql expression
+func getRevisionsFromExpr(ctx *sql.Context, expr sql.Expression, canDot bool) (string, string, bool, error) {
```
fulghum, line 540: (minor) For Go to associate the comment with the function, it needs the function name as the first word. 

```suggestion
// getRevisionsFromExpr returns the revisionName and/or secondRevisionName from sql expression
func getRevisionsFromExpr(ctx *sql.Context, expr sql.Expression, canDot bool) (string, string, bool, error) {
```

fulghum: Also, would be helpful for future code-readers to document what that boolean return parameter is. That would be difficult for someone to figure out without having to jump in and read the function code.

fulghum (approved):
Looks good! 🎉

PR #4692, opened 2022-11-02
Got rid of testing libraries from dolt binary

zachmu:
This decreases the size of the linux `dolt` binary by 8MB, from 75 to 67MB.


go/store/prolly/write_amplification_test.go
```diff
@@ -1,213 +0,0 @@
-// Copyright 2021 Dolthub, Inc.
```
andrew-wm-arthur, line 1: this needs some work to revive it, but I don't think it's worth deleting

andrew-wm-arthur (approved):
LGTM!

PR #4697, opened 2022-11-03
warning instead of error for unix socket address already in use when starting sql server

jennifersp:
depends on https://github.com/dolthub/go-mysql-server/pull/1372


go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -202,12 +203,13 @@ func Serve(
 		)
 	}
 
-	if startError != nil {
+	if errors.Is(startError, server.UnixSocketInUseError) {
+		lgr.Warn("unix socket set up failed: address already in use: listen unix at ", serverConf.Socket)
```
fulghum, line 207: (minor) just shortening up the error message a little bit... 

```suggestion
		lgr.Warn("unix socket set up failed: file already in use: ", serverConf.Socket)
```

integration-tests/bats/sql-server.bats
```diff
@@ -1504,9 +1504,9 @@ databases:
     dolt sql-server --port=$PORT --socket="$REPO_NAME/mysql.sock" --user dolt > log.txt 2>&1 &
     SERVER_PID=$!
     run wait_for_connection $PORT 5000
-    [ "$status" -eq 1 ]
+    [ "$status" -eq 0 ]
 
-    run grep 'address already in use' log.txt
+    run grep 'unix socket set up failed: address already in use: listen unix at' log.txt
```
fulghum, line 1509: Nice test for this! 

It might be nice to also test that when we shut down the second sql-server, it doesn't delete the `mysql.sock` file that it tried to create, but wasn't able to. I think this will happen correctly from the code changes, but probably worth a quick line or two to assert it behaves that way, too.

fulghum (approved):
Very nice! 🧦

PR #4717, opened 2022-11-07
[TESTING] Benchmark privilege cache changes

Hydrocharged:


Hydrocharged:
#benchmark

Hydrocharged:
#benchmark

Hydrocharged:
#benchmark

timsehn:
Beautiful!

timsehn:
On second look. Not beautiful.

Hydrocharged:
There shouldn’t really be a difference between the two. I’m trading a map lookup for a mutex lock as far as time complexity goes, but I don’t know if the lock is even necessary. I’ll try removing it and seeing if it makes a difference.

Hydrocharged:
#benchmark

Hydrocharged:
There is a lot of run-to-run variance, so it looks like any improvements or regressions are within the margin of error.

PR #4722, opened 2022-11-07
implementing prefix indexes for string types

JCOR11599:
This PR implements prefix indexes for inline string data type (`CHAR`, `VARCHAR`, `BINARY`, and `VARBINARY`) columns.
The changes seem to work for `BLOB` and `TEXT` though.

Changes:
1. `doltIndex` now has a `prefixLengths` field which is a `[]uint16`
   - it contains the prefixes for each of the columns in the index definition
   - it is in order of the prefix definition
   - a nil or empty list means that the columns in the index should not be trimmed
     - none of the columns are string type
     - the columns are inline string, but the entire string should be used for comparison
2. The flatbuffer files, serialization, and deserialization code deal with the prefix index lengths
3. Currently, dolt will indicate that it won't support this index, because...it can't; at least not completely right now (see analyzer changes at bottom)

TODO:  analyzer changes
- primary key prefixes are futher away, so we still throw an unsupported error for those.
- there needs to be an analyzer change to actually take advantage of this index.
Example:
```sql
> create table t (v varchar(2), index (v(1)));
> insert into t values ("aa"), ("ba"), ("bb"), ("bc"), ...;
> select v from t where v = "aa";
```
The query here should be able to take advantage of the secondary index; it should do a range prefix lookup over the table for just the rows that start with `"a"`, then apply an equality filter over the resulting table. Currently, it will just do a full table scan.


go/libraries/doltcore/schema/index_coll.go
```diff
@@ -64,6 +64,8 @@ type IndexCollection interface {
 	RenameIndex(oldName, newName string) (Index, error)
 	//SetPks changes the pks or pk ordinals
 	SetPks([]uint64) error
+	// SetIndexPrefixLength sets the prefix lengths for the specified index
+	SetIndexPrefixLength(indexName string, prefixLengths []uint16) error
```
Hydrocharged, line 68: This is the wrong abstraction here. `SetPks` applies to the index collection, which is modifiable by nature.  This is not the case for indexes, especially having a function on the collection to modify indexes. This needs to be moved to index creation.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -108,39 +108,58 @@ func TestSingleQuery(t *testing.T) {
...
-						{1, 3},
-						{11, 20},
-					},
-				},
-				{
-					Query: "select z from b",
-					Expected: []sql.Row{
-						{100},
-					},
+	script := queries.ScriptTest{
```
Hydrocharged, line 111: Remember to add the skip back here

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -108,39 +108,58 @@ func TestSingleQuery(t *testing.T) {
...
 		},
 	}
 
 	harness := newDoltHarness(t)
-	for _, test := range scripts {
-		enginetest.TestScript(t, harness, test)
-	}
+	enginetest.TestScript(t, harness, script)
+
+	//t.Skip()
```
Hydrocharged, line 130: Or just uncomment all of this and undo your changes lol, we definitely need to not have this massive comment block though

go/libraries/doltcore/sqle/enginetest/validation.go
```diff
@@ -215,6 +220,7 @@ func validatePkIndex(ctx context.Context, sch schema.Schema, def schema.Index, p
 			if j < pkSize {
 				builder.PutRaw(i, key.GetField(j))
 			} else {
+
```
Hydrocharged, line 223: Random new line?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1778,7 +1796,8 @@ func (t *AlterableDoltTable) CreateIndex(ctx *sql.Context, idx sql.IndexDef) err
 	for _, idxCol := range idx.Columns {
 		col := t.DoltTable.sqlSch.Schema[t.DoltTable.sqlSch.IndexOfColName(idxCol.Name)]
 		if idxCol.Length > 0 && sql.IsText(col.Type) {
-			return sql.ErrUnsupportedIndexPrefix.New(col.Name)
+			// TODO: delete me
+			//return sql.ErrUnsupportedIndexPrefix.New(col.Name)
```
Hydrocharged, line 1800: Is this supposed to have been deleted? If so the `if` statement can also be removed.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -2246,6 +2284,22 @@ func (t *AlterableDoltTable) CreateIndexForForeignKey(ctx *sql.Context, idx sql.
 		columns[i] = indexCol.Name
 	}
 
+	var hasNonZeroPrefixLength bool
```
Hydrocharged, line 2287: I'd add another TODO for that helper method just to make sure you don't forget

go/libraries/doltcore/sqle/temp_table.go
```diff
@@ -267,11 +267,28 @@ func (t *TempTable) CreateIndex(ctx *sql.Context, idx sql.IndexDef) error {
 		cols[i] = c.Name
 	}
 
+	var hasNonZeroPrefixLength bool
```
Hydrocharged, line 270: Another TODO here as well

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -413,3 +413,192 @@ func FormatKeyForUniqKeyErr(key val.Tuple, d val.TupleDesc) string {
 	sb.WriteString("]")
 	return sb.String()
 }
+
+type prollySecondaryIndexPrefixWriter struct {
```
Hydrocharged, line 417: Seems like this could just be a branch in the regular index writer. I'm not seeing much difference here for the additional 200 lines of maintenance when some refactoring would work

go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
```diff
@@ -331,3 +331,173 @@ func (writer prollyKeylessSecondaryWriter) HasEdits(ctx context.Context) bool {
 func (writer prollyKeylessSecondaryWriter) IterRange(ctx context.Context, rng prolly.Range) (prolly.MapIter, error) {
 	return writer.mut.IterRange(ctx, rng)
 }
+
+type prollyKeylessSecondaryPrefixWriter struct {
```
Hydrocharged, line 335: Same comment as the non-keyless writer, this should be a branch on the original

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -78,15 +79,30 @@ func getSecondaryProllyIndexWriters(ctx context.Context, t *doltdb.Table, sqlSch
...
-		writers[defName] = prollySecondaryIndexWriter{
-			name:    defName,
-			mut:     idxMap.Mutate(),
-			unique:  def.IsUnique(),
-			idxCols: def.Count(),
-			keyMap:  keyMap,
-			keyBld:  val.NewTupleBuilder(keyDesc),
-			pkMap:   pkMap,
-			pkBld:   val.NewTupleBuilder(pkDesc),
+		prefixLengths := def.GetPrefixLengths()
```
Hydrocharged, line 82: I think we always pass the prefix lengths and just branch depending on whether there is or isn't a prefix

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -114,15 +130,30 @@ func getSecondaryKeylessProllyWriters(ctx context.Context, t *doltdb.Table, sqlS
...
-		writers[defName] = prollyKeylessSecondaryWriter{
-			name:      defName,
-			mut:       m.Mutate(),
-			primary:   primary,
-			unique:    def.IsUnique(),
-			keyBld:    val.NewTupleBuilder(keyDesc),
-			prefixBld: val.NewTupleBuilder(keyDesc.PrefixDesc(def.Count())),
-			hashBld:   val.NewTupleBuilder(val.NewTupleDescriptor(val.Type{Enc: val.Hash128Enc})),
-			keyMap:    keyMap,
+		prefixLengths := def.GetPrefixLengths()
```
Hydrocharged, line 133: Same comment here regarding always passing the lengths

go/libraries/doltcore/table/editor/creation/index.go
```diff
@@ -109,6 +110,11 @@ func CreateIndex(
 		return nil, err
 	}
 
+	err = sch.Indexes().SetIndexPrefixLength(indexName, prefixLengths)
```
Hydrocharged, line 113: Makes far more sense to have another parameter in `AddIndexByColNames` rather than this new function on the interface

go/libraries/doltcore/schema/index.go
```diff
@@ -54,6 +54,10 @@ type Index interface {
 	// ToTableTuple returns a tuple that may be used to retrieve the original row from the indexed table when given
 	// a full index key (and not a partial index key).
 	ToTableTuple(ctx context.Context, fullKey types.Tuple, format *types.NomsBinFormat) (types.Tuple, error)
+	// GetPrefixLengths returns the prefix lengths for the index
+	GetPrefixLengths() []uint16
+	// SetPrefixLengths applies to prefix lengths
+	SetPrefixLengths(prefixLengths []uint16)
```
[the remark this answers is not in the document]
JCOR11599: I did think of this, but I saw that the index creation code had a lot of uses and wanted to avoid changing a bunch of functions 😅.

Regardless, should `PrefixLengths` be in another argument or should it go in `IndexProperties`?
For example, should 
```go
func NewIndex(name string, tags, allTags []uint64, indexColl IndexCollection, props IndexProperties) Index
```
change to 
```go
func NewIndex(name string, tags, allTags []uint64, prefixLengths []uint16, indexColl IndexCollection, props IndexProperties) Index
```
or should `IndexProperties` just have another field?

I'm in favor of adding it as a new parameter.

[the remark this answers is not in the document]
Hydrocharged: We should just change all of those functions lol. I’m in favor of a new parameter rather than another field, but if there are too many parameters, we could move everything to index properties.

go/libraries/doltcore/sqle/writer/prolly_table_writer.go
```diff
@@ -54,6 +54,7 @@ type prollyTableWriter struct {
 
 var _ TableWriter = &prollyTableWriter{}
 
+// TODO: return a prefix index writer if there is a prefix defined on this table
```
Hydrocharged, line 57: Comment no longer applies

integration-tests/bats/migration-integration.bats
```diff
@@ -29,7 +29,7 @@ teardown() {
     dolt tag -v
     run dolt tag -v
     [ "$status" -eq 0 ]
-    [[ "$output" =~ "tdkt7s7805k1ml4hu37pm688g5i0b8ie" ]] || false
+    [[ "$output" =~ "7jrvg1ajcdq6t9sevcejv4e9o0fgrmle" ]] || false
```
Hydrocharged, line 32: So for this test (and the following one), I'm assuming you ran these commands locally and got the new hashes, so it's not `tdkt7s7805k1ml4hu37pm688g5i0b8ie`, but `7jrvg1ajcdq6t9sevcejv4e9o0fgrmle` now. That's fine. But you're not replacing all instances of `tdkt7s7805k1ml4hu37pm688g5i0b8ie`. It's probably easier to reverse this file, and just do a find->replace on that exact hash. That will fix your test failure.

Hydrocharged (approved):
LGTM, much better after the improvements!

PR #4728, opened 2022-11-08
Added validation test to check for test libraries in binary

zachmu:


timsehn (approved):
Thought the function was a bit overkill but then I remembered if you want to use `|` in bats you have to use `bash -c`. So, if you don't like the function, do that.

Here's an example:

https://github.com/dolthub/dolt/blob/a26bae31d5e63c114d0171c811ff045ccc2988af/integration-tests/bats/sql-shell.bats#L67

So your test would be:
```
run bash -c "strings `which dolt` | grep testify"
[ $status -ne 0 ]
[ "$output" = "" ]
```

PR #4744, opened 2022-11-10
Added basic smoke test for TypeORM

fulghum:


integration-tests/orm-tests/orm-tests.bats
```diff
@@ -58,6 +62,14 @@ teardown() {
   # https://github.com/prisma/prisma/tree/main/packages/integration-tests/src/__tests__/integration/mysql
 }
 
+@test "TypeORM smoke test" {
```
jennifersp, line 65: A small description what TypeORM is for like what Peewee and Prisma smoke test have on top of them would be nice.

jennifersp (approved):
Looks good!

integration-tests/orm-tests/orm-tests.bats
fulghum: Good suggestion! I'll add those in before I merge.

PR #4749, opened 2022-11-10
go/libraries/doltcore/sqle/cluster: Add JWT authentication for peer communication to cluster replicas.

reltuk:


reltuk:
Yeah, exactly. All this new functionality is enabled all the time, so the existing tests for tls and non-tls standby replication test the new auth flow.

PR #4753, opened 2022-11-10
go/store/val: allow year encoding to store value 0

druvv:
Previously, the year encoding did not support a value of 0. It now uses uint8 255 to represent 0 since that value is free in the encoding space.

Companion GMS PR:
https://github.com/dolthub/go-mysql-server/pull/1395

### DIFF
--- a/go/store/val/codec.go
+++ b/go/store/val/codec.go
@@ -430,14 +430,29 @@ func compareDecimal(l, r decimal.Decimal) int {
 }
 
 const minYear int16 = 1901
+const maxYear int16 = 2155
+const zeroToken uint8 = 255
 
 func readYear(val []byte) int16 {
 	expectSize(val, yearSize)
-	return int16(readUint8(val)) + minYear
+	v := readUint8(val)
+	if v == zeroToken {
+		return int16(0)
+	}
+	return int16(v) + minYear
 }
 
+// 1901 - 2155 encodes to uint8 range 0 - 254 for legacy reasons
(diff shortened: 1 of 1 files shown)


go/store/val/codec.go
```diff
@@ -430,14 +430,29 @@ func compareDecimal(l, r decimal.Decimal) int {
 }
 
 const minYear int16 = 1901
+const maxYear int16 = 2155
+const zeroToken uint8 = 255
```
max-hoffman, line 434: how would you feel about:
```go
type yearOffset uint8
const (
        maxOffset yearOffset = 254
        nilOffset = 255
)
```
Not sure it's strictly better, but it seems like we are encoding offsets? And this might give `zeroToken` a more sensible name in that context.

go/store/val/codec.go
```diff
@@ -430,14 +430,29 @@ func compareDecimal(l, r decimal.Decimal) int {
...
-	return int16(readUint8(val)) + minYear
+	v := readUint8(val)
+	if v == zeroToken {
+		return int16(0)
+	}
+	return int16(v) + minYear
 }
 
+// 1901 - 2155 encodes to uint8 range 0 - 254 for legacy reasons
+// 0 encodes to uint8 255
```
max-hoffman, line 446: maybe move expanded description to the base/offset definitions, have the docstring indicate that we are writing an offset relative to minYear

max-hoffman (approved):
LGTM, naming suggestion

PR #4760, opened 2022-11-11
wrap db read error

max-hoffman:


coffeegoddd (approved):
LGTM! should we do this on fetch and the other place this err is used?

max-hoffman:
> LGTM! should we do this on fetch and the other place this err is used?

If error wrapping is a frequent problem for doltlab, i'd suggest looking into automated solutions. There are maybe thousands of the missing wraps, think there might be more efficient ways of doing this.

PR #4767, opened 2022-11-11
fix DOLT_ROOT_PATH bug

bheni:
Currently there is an issue where if you run something like:

```DOLT_ROOT_PATH=. dolt backup restore <remote> subfolder/reponame```

This fixes that by returning the absolute path.


druvv (approved):
LGTM

PR #4781, opened 2022-11-14
fix merge --no-ff --no-commit

jennifersp:
If the editor process is killed or the editor was saved with empty message, it gives error.
Fixes `dolt merge --no-ff --no-commit`, now it does not fast-forward and commit. It leaves the working set with changes.


fulghum (approved):
Nice work! 👏  Code looks great and I manually tested the behavior when the editor process is killed. Thanks for closing this gap!

PR #4795, opened 2022-11-15
Table func join tests

max-hoffman:
sibling PR here: https://github.com/dolthub/go-mysql-server/pull/1405


go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -186,7 +186,7 @@ func (ltf *LogTableFunction) Expressions() []sql.Expression {
 
 // getDoltArgs builds an argument string from sql expressions so that we can
 // later parse the arguments with the same util as the CLI
-func getDoltArgs(ctx *sql.Context, expressions []sql.Expression, functionName string) ([]string, error) {
+func getDoltArgs(ctx *sql.Context, expressions []sql.Expression, Name string) ([]string, error) {
```
fulghum, line 189: (minor) lower case "name"?

```suggestion
func getDoltArgs(ctx *sql.Context, expressions []sql.Expression, name string) ([]string, error) {
```

go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -196,7 +196,7 @@ func getDoltArgs(ctx *sql.Context, expressions []sql.Expression, functionName st
 		}
 
 		if !sql.IsText(expr.Type()) {
-			return args, sql.ErrInvalidArgumentDetails.New(functionName, expr.String())
+			return args, sql.ErrInvalidArgumentDetails.New(Name, expr.String())
```
fulghum, line 199:
```suggestion
			return args, sql.ErrInvalidArgumentDetails.New(name, expr.String())
```

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -5817,6 +5828,13 @@ var DiffSummaryTableFunctionScriptTests = []queries.ScriptTest{
 				Query:       "SELECT * from dolt_diff_summary(@Commit1, @Commit5, 't');",
 				ExpectedErr: sql.ErrTableNotFound,
 			},
+			{
+				Query: `
+SELECT *
+from dolt_diff_summary(@Commit3, @Commit4, 't') 
+inner join t as of @Commit3 on rows_unmodified = t.pk;`,
```
fulghum, line 5835: Are there any other join types worth testing?

fulghum (approved):
Looks good! Fast work on this one. ⚡

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
max-hoffman: The diff table function isn't going to join against itself, but the fix is a bit involved, not sure this is the right place. I will make an issue. Otherwise I made every table function a generic join node leaf, I am not super worried about the execution logic having issues. Can you think of any table functions (other than json_table) whose join order can't be rearranged?

max-hoffman: https://github.com/dolthub/dolt/issues/4798

fulghum: Nice! I agree joining to itself doesn't seem high priority. It should be fine to rearrange the join order for any of the (non json_table) table functions we have.

PR #4797, opened 2022-11-15
allow `not` and `is null` clauses with dolt diff table

jennifersp:


fulghum (approved):
Looks good! 🚢  Thanks for fixing this one.

PR #4799, opened 2022-11-15
Fix missing projection on indexed table

max-hoffman:
This error only applies to `DoltIndexedTable`, which as far as I can tell is limited to the history tables.


zachmu (approved):
LGTM

PR #4800, opened 2022-11-15
Skip empty diffs caused by non-canonicalized tuples.

druvv:
At one point, we started to trim null suffixes from tuples. In versions prior to this change, tuples could have a null field at the end of the tuple. Now we always truncate the tuple.

If old versions insert rows with null values as the last column, it produces a tuple that is longer than the tuple that would be inserted by the latest version. This leads to empty diffs.


andrew-wm-arthur (approved):
LGTM

PR #4803, opened 2022-11-15
go/store/nbs: first pass at journaling chunk store

andrew-wm-arthur:
Adds `chunkJournal` a log-based `ChunkStore` that integrates into current `NomsBlockStore` machinery. `chunkJournal` implements `tablePersister` and `manifest`. When `Memtable`s are passed to `Persist()`, their contents appended to the journal file at current offset and a  `journalChunkSource` (implementing `chunkSource`) is returned. `journalChunkSource` is an in-memory object containing offset metadata for reading back chunks from the journal file. The `chunkJournal` also persists root hash updates to the log via the `manifest` interface. 

This implementation is currently behind a feature flag. Lots of work remains to fully support current NBS functionality. 
- Push/Pull/Clone paths are largely unsupported. 
- `chunkJournal` is limited to a fixed size, doesn't have support for finalizing the current journal as a table file
- `chunkJournal` needs to flush current in-memory manifest contents to manifest file
- Entire journal file must be processed at startup (needs out-of-band index)
- Need to shore-up NBS instantiation. The implementation here uses a global singleton to handle multiple in-process instantiations, however this breaks multi-db support. 
- Current chunk record format is temporary and needs support for format extension

### DIFF
--- a/go/performance/scripts/local_sysbench.sh
+++ b/go/performance/scripts/local_sysbench.sh
@@ -35,6 +35,9 @@ do
         --row2) export ENABLE_ROW_ITER_2=true
             ;;
 
+        --journal) export DOLT_ENABLE_CHUNK_JOURNAL=true
+            ;;
+
         # specify sysbench benchmark
         *) SYSBENCH_TEST="$1"
             ;;
@@ -123,6 +126,7 @@ sysbench \
   --db-ps-mode=disable \
   "$SYSBENCH_TEST" run
 
+unset DOLT_ENABLE_CHUNK_JOURNAL
 unset DOLT_DEFAULT_BIN_FORMAT
 unset ENABLE_ROW_ITER_2
 unset SINGLE_THREAD_FEATURE_FLAG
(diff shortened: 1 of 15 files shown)


andrew-wm-arthur:
I found a bug in the manifest update logic that I will address (either here or in a future PR). 

The intention of this design is to integrate mutable log-style persistence into the current machinery the works with immutable table files. To accomplish this, `journalChunkSources` act like "in-memory table files". Their contents are immutable, but their backing persistent structure (the journal file) is _mutable_. Each append to the journal file creates a new immutable in-memory `chunkSource` that contains the necessary offset metadata to read chunks from the journal file. 

This design is complicated by the need to update the manifest with the current table file set. "In-memory table files" (`journalChunkSource`s) must be able to be identified by `tableSpec` references in order to correctly flow through the `tableSet` machinery. *However* we do not want to write `tableSpec`s corresponding to `journalChunkSource`s into the manifest as these tableSpecs do not identify a persisted table file. Special care must be taken to add and subtract in-memory `tableSpec`s to the `manifestContents` when reading/writing it.

PR #4804, opened 2022-11-15
A smoke test for MikroORM

jennifersp:


fulghum (approved):
Very nice! Looks great! 🚢

PR #4807, opened 2022-11-16
[no-release-notes] moving tests to gms

JCOR11599:
tests that used to just work for dolt, now work for in-memory tables, so moving the tests there.
companion PR: https://github.com/dolthub/go-mysql-server/pull/1407


fulghum (approved):
Looks good! 🚢

PR #4815, opened 2022-11-17
dolt init fails with invalid arguments

jennifersp:


fulghum (approved):
This is a good fix for the `init` command. Would be good to think about how we could do this more generally, for both CLI commands and SQL builtin procedures.

Issue #4818, opened 2022-11-17
SELECT * from dolt_diff_summary('commitA', 'commitB') should not abort if a primary key change is detected.


druvv:
**Current behavior:**
SELECT * from dolt_diff_summary('commitA', 'commitB') returns `diff summary will not compute due to primary key set change` if any table has a primary key change.

**Expected behavior:**
SELECT * from dolt_diff_summary('commitA', 'commitB') returns a diff summary for all tables except those that include primary key changes.

We could also include the errored table in the list. All diffs would be recorded as 0 and a different column would indicate a pk key change.

PR #4821, opened 2022-11-17
Change dolt_diff_summary table function to return empty diffs for tab…

druvv:
…les with pk set changes.

- The 3 argument version returns an error if the table specified has a pk set change.
- The 2 argument verison returns empty diffs and returns a warning.

I considered approximating the diff for the pk set change, but it feels like an overreach for the customers need.

Fixes: #4818


fulghum (approved):
Awesome change! Looks great. 🚢

PR #4825, opened 2022-11-18
Blob chunker rewrite

max-hoffman:
Rewrite blob chunking to use a a fixed amount of memory.

image

(name) (iterations) (nanoseconds/iteration) (bytes/iteration) (allocations/iteration)
```
-- after
BenchmarkBlobBuilder/datasize:_3
BenchmarkBlobBuilder/datasize:_3-12         	  234128	      5673 ns/op	    2608 B/op	       8 allocs/op
BenchmarkBlobBuilder/datasize:_4
BenchmarkBlobBuilder/datasize:_4-12         	   26074	     41468 ns/op	   17560 B/op	      24 allocs/op
BenchmarkBlobBuilder/datasize:_5
BenchmarkBlobBuilder/datasize:_5-12         	    2737	    378950 ns/op	  133004 B/op	     134 allocs/op
BenchmarkBlobBuilder/datasize:_6
BenchmarkBlobBuilder/datasize:_6-12         	     282	   3974650 ns/op	 1288281 B/op	    1269 allocs/op

-- before
BenchmarkBlobBuilder
BenchmarkBlobBuilder/datasize:_3
BenchmarkBlobBuilder/datasize:_3-12         	   15231	     75445 ns/op	  460464 B/op	      17 allocs/op
BenchmarkBlobBuilder/datasize:_4
BenchmarkBlobBuilder/datasize:_4-12         	    6844	    171266 ns/op	  923121 B/op	      50 allocs/op
BenchmarkBlobBuilder/datasize:_5
BenchmarkBlobBuilder/datasize:_5-12         	    1875	    549793 ns/op	 1040770 B/op	     270 allocs/op
BenchmarkBlobBuilder/datasize:_6
BenchmarkBlobBuilder/datasize:_6-12         	     273	   4389403 ns/op	 2668899 B/op	    2543 allocs/op
```


max-hoffman:
#import-benchmark

max-hoffman:
> @max-hoffman **DOLT**
> 
> read_tests	from_latency_median	to_latency_median	is_faster
> covering_index_scan	2.71	2.71	0
> groupby_scan	16.71	17.01	0
> index_join	4.57	4.57	0
> index_join_scan	3.89	3.89	0
> index_scan	52.89	53.85	0
> oltp_point_select	0.48	0.48	0
> oltp_read_only	8.74	8.74	0
> select_random_points	0.77	0.77	0
> select_random_ranges	1.14	1.14	0
> table_scan	62.19	62.19	0
> types_table_scan	186.54	189.93	0
> write_tests	from_latency_median	to_latency_median	is_faster
> bulk_insert	0.001	0.001	0
> oltp_delete_insert	11.87	10.84	0
> oltp_insert	2.71	2.71	0
> oltp_read_write	17.01	17.32	0
> oltp_update_index	5.37	4.82	1
> oltp_update_non_index	5.57	5.18	0
> oltp_write_only	8.13	8.13	0
> types_delete_insert	16.12	12.08	1

The `types_delete_insert` change is partially substantial, I see a 10% improvement locally. I think the other differences are noise.

go/store/prolly/tree/immutable_tree.go
```diff
@@ -29,137 +31,276 @@ import (
...
+		lChunkSize: chunkSize,
+		iChunkSize: chunkSize / hash.ByteLen,
+		buf:        make([]byte, chunkSize),
+		keys:       keys,
+		na:         &nodeArena{},
+		subtrees:   make([]uint64, chunkSize/hash.ByteLen),
+	}, nil
+}
+
+type BlobBuilder struct {
```
reltuk, line 71: Is the important thing here to write all the leaf chunks before we write any internal nodes and to write the internal nodes for each level contiguously? Or is the mostly allocation and duffcopy optimization?

This seems simpler to structure and maybe achieves less allocations if we build a chain of writers, one for each level and just call Write(reader) on it? Leaf writer just writes a leaf blob after reading into a buffer its constructed with. Internal writer just calls its child repeatedly, accumulating addrs into the buffer it's constructed with and flushing once it's full, returning to the addr to its caller.

A possible downside is this approach only writes runs of leaf nodes which fit in the first level internal node and doesn't write runs of internal nodes at all...

I have a small proof of concept which seems easy to follow and ~90 lines shorter but, not sure if it achieves the intent here...

reltuk (approved):
Generally looks good to me. Not sure BlobBuilder is structured as well as it could be, but depends on the use case. A few comments on some of our attempts to optimize allocation here which I think maybe aren't working as intended.

go/store/prolly/tree/node.go
```diff
@@ -113,6 +113,33 @@ func walkOpaqueNodes(ctx context.Context, nd Node, ns NodeStore, cb NodeCb) erro
...
+	}
+	n := a.Get()
+	n.keys = keys
+	n.values = values
+	n.count = count
+	n.level = level
+	n.msg = msg
+	return n, nil
+}
+
```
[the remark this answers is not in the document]
max-hoffman: hmm, yeah makes sense

go/store/prolly/tree/immutable_tree.go
```diff
@@ -29,137 +31,276 @@ import (
 
 const DefaultFixedChunkLength = 4000
 
-var ErrInvalidChunkSize = errors.New("invalid chunkSize; value must be > 1")
+var ErrInvalidChunkSize = errors.New("invalid chunkSize; value must be a multiple of 20")
 
-// buildImmutableTree writes the contents of |reader| as an append-only
+var chunkBufPool = sync.Pool{
```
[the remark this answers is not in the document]
max-hoffman: adding the missed `Put` in reset

go/store/prolly/tree/immutable_tree.go
max-hoffman: can you share your POC? it sounds like the regular chunker/old blob chunker algo

max-hoffman: Do you mean i should have kept the old DFS chunk building order, but have been smarter dividing a common backing slice for the level buffers? My main intent was to use the same two slices for all blobs, a byte buffer for leaf chunks, and an internal `[]hash.Hash` buffer for node addresses, though I wasn't even doing the address buffer properly. I need to write more go bench tests with increasing blob sizes, compare mallocs more closely.

go/store/prolly/tree/immutable_tree.go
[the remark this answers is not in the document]
reltuk: Does Reset() get called after Chunk()? Should the Put() just be at the end of Chunk?

go/store/prolly/tree/immutable_tree.go
reltuk: https://github.com/dolthub/dolt/pull/4830. Didn't optimize away the subtree counts here...

I just don't know which optimizations are the most important here...if we have a pool with the []byte, [][]byte, []uint64...I don't know if the DFS is important.

I guess another approach in #4830 would be to forgo the one-big-[]byte for the chain and just put every internal builder into the Pool and take out the number of you need for your height...

max-hoffman: thanks for highlighting the implementation forks, i also do not have a good sense of which are most important. i will benchmark the cpu/mem tradeoffs for the options on monday.

PR #4826, opened 2022-11-18
[no-release-notes] Fixed db issue with branch permissions test

Hydrocharged:
I was unable to fix the root cause, so this just works around the issue. I was eventually able to get new contexts to recognize created databases, but then the created branches weren't being recognized, so I'm leaving the issue alone for now.

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/branch_control_test.go
+++ b/go/libraries/doltcore/sqle/enginetest/branch_control_test.go
@@ -941,10 +941,6 @@ func TestBranchControl(t *testing.T) {
 	for _, test := range BranchControlTests {
 		harness := newDoltHarness(t)
 		t.Run(test.Name, func(t *testing.T) {
-			//TODO: fix whatever is broken with test db handling
-			if test.Name == "Proper database scoping" {
-				return
-			}
 			engine, err := harness.NewEngine(t)
 			require.NoError(t, err)
 			defer engine.Close()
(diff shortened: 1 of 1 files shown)


max-hoffman (approved):
LGTM

go/libraries/doltcore/sqle/enginetest/branch_control_test.go
```diff
@@ -971,15 +966,10 @@ func TestBranchControl(t *testing.T) {
...
-				var ctx *sql.Context
-				var ok bool
-				if ctx, ok = ctxMap[user+"@"+host]; !ok {
-					ctx = enginetest.NewContextWithClient(harness, sql.Client{
-						User:    user,
-						Address: host,
-					})
-					ctxMap[user+"@"+host] = ctx
-				}
+				ctx = ctx.NewCtxWithClient(sql.Client{
```
max-hoffman, line 969: i dont think contexts are reusable, or at least we make that hard assumption across enginetests. not sure i would have expected the old way to work

Hydrocharged: We definitely don't expect them to be reusable, but this at least gets the tests to pass. The old way (reusing contexts) got some other tests to pass that simply instantiating a new context every assertion failed to do. I wouldn't expect this workaround to work for other engine tests.

PR #4829, opened 2022-11-19
add Hibernate smoke test

jennifersp:


integration-tests/orm-tests/hibernate/DoltHibernateSmokeTest/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <!-- Hibernate -->
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
+      <version>5.6.14.Final</version>
```
fulghum, line 20: Declaring a dependency range will help keep our tests running with the latest third-party releases and give us an early signal if something breaks in our integration. For Maven, that looks something like:  
`<version>[5.0.0,6.0.0)</version>`

That syntax should select the latest version in the 5.x series, but not select anything from the 6.x series. This StackOverflow post looks like a good explanation of how those ranges work. Would be good to do that for mysql-connector-j, too. 

Speaking of versions... it looks like Hibernate has a stable 6.x series out, too. Could you try seeing if this test works with 6.x without any changes? If so we should probably start off with 6.x.

fulghum (approved):
Very nice work! 🙌  Just some minor notes.

integration-tests/orm-tests/hibernate/DoltHibernateSmokeTest/pom.xml
jennifersp: Initially, I tried like this `<version>[5.0.0)</version>`, which is anything later than `version 5.0.0`. It was working fine locally on macos, but in Docker container it was failing. I found out that the linux container was getting the latest version regardless of alpha or beta build, which beta builds do not work, but on my mac, it was getting slightly earlier alpha build than linux one which works. I'll check if there is any alpha build only option i can set.

PR #4831, opened 2022-11-21
Fix issues with docker entrypoint

nkonev:
1. Fix unability to process .sql in initdb
If user has some .sql files in `/docker-entrypoint-initdb.d` they will have an error:
```
2022-11-21 11:02:13+00:00 [Note] [Entrypoint]: /usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/init.sql
/usr/local/bin/docker-entrypoint.sh: line 98: docker_process_sql: command not found
```

This PR fixes it

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -19,8 +19,12 @@ mysql_error() {
 	mysql_log ERROR "$@" >&2
 	exit 1
 }
+docker_process_sql() {
+  dolt sql
+}
 
 CONTAINER_DATA_DIR="/var/lib/dolt"
+INIT_COMPLETED="$CONTAINER_DATA_DIR/.init_completed"
 DOLT_CONFIG_DIR="/etc/dolt/doltcfg.d"
 SERVER_CONFIG_DIR="/etc/dolt/servercfg.d"
 DOLT_ROOT_PATH="/.dolt"
@@ -80,7 +84,7 @@ get_config_file_path_if_exists() {
 #    ie: docker_process_init_files /always-initdb.d/*
 # process initializer files, based on file extensions
 docker_process_init_files() {
-	echo
+	mysql_note "Running init scripts"
 	local f
 	for f; do
 		case "$f" in
@@ -106,12 +110,6 @@ docker_process_init_files() {
 	done
 }
 
-start_server() {
-    # start the server in fixed data directory at /var/lib/dolt
-    cd $CONTAINER_DATA_DIR
-    "$@"
-}
-
 # if there is config file provided through /etc/dolt/doltcfg.d,
 # we overwrite $HOME/.dolt/config_global.json file with this file.
 set_dolt_config_if_defined() {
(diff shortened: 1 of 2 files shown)


nkonev:
Marking as draft because of I found some other issues with entrypoint:
1. scripts from initdb are invoked every time, although in MySQL theya are invoken exactly one time, they have a check https://github.com/docker-library/mysql/blob/master/8.0/docker-entrypoint.sh#L402
2. Defunct processes in `ps` output
```
[nkonev@ip-10-64-8-18 companies]$ docker exec -it companies_doltdb_1 bash
root@195511b88240:/# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 11:10 ?        00:00:00 dolt sql-server --host=0.0.0.0 --port=3306
root          25       1  0 11:10 ?        00:00:00 [dolt] <defunct>
root          55       0  0 11:10 pts/0    00:00:00 bash
root          92       1  0 11:12 pts/0    00:00:00 [dolt] <defunct>
root         122       1  0 11:15 pts/0    00:00:00 [dolt] <defunct>
root         153       1  0 11:15 pts/0    00:00:00 [dolt] <defunct>
root         183       1  0 11:32 pts/0    00:00:00 [dolt] <defunct>
root         211       1  0 11:32 pts/0    00:00:00 [dolt] <defunct>
root         225       0  0 13:17 pts/1    00:00:00 bash
root         233     225  0 13:17 pts/1    00:00:00 ps -ef
```
Most likely, the reason of it here - it seems we try to run server again https://github.com/dolthub/dolt/blob/9e33b03c777a3ef66bd4727e13197e3df8329ef6/docker/docker-entrypoint.sh#L158
although we already ran in earlier https://github.com/dolthub/dolt/blob/9e33b03c777a3ef66bd4727e13197e3df8329ef6/docker/docker-entrypoint.sh#L151

I want to investingate more and after this will mark it as ready

docker/docker-entrypoint.sh
```diff
@@ -132,6 +130,8 @@ _main() {
         set -- dolt sql-server --host=0.0.0.0 --port=3306 "$@"
     fi
 
+    cd $CONTAINER_DATA_DIR
```
nkonev, line 133: `dolt sql` and `dolt sql-server` are need to use the same directory, so I set it here

docker/docker-entrypoint.sh
```diff
@@ -106,12 +110,6 @@ docker_process_init_files() {
 	done
 }
 
-start_server() {
```
nkonev, line 109: It actually started nothing because "$@" means empty string because we didn't pass arguments to start_server()

docker/docker-entrypoint.sh
```diff
@@ -19,8 +19,12 @@ mysql_error() {
 	mysql_log ERROR "$@" >&2
 	exit 1
 }
+docker_process_sql() {
+  dolt sql
```
nkonev, line 23: I think it's ok to just start dolt as a commandline for sake initialization. Pros of this approach is that the container isn't listen to TCP during the initialization, that is good for e. g. readiness probes

nkonev:
Speaking about `<defunct>`, I noticed, that any invocation of dolt inside a container, even `dolt version` leads us to `<defunct>`. It seems, it cannot be solved by changing entrypoint bash, I think it should be solved on the Go side.

docker/serverDockerfile
```diff
@@ -20,4 +20,5 @@ RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 ENTRYPOINT ["docker-entrypoint.sh"]
 
 EXPOSE 3306 33060
+WORKDIR /var/lib/dolt
```
nkonev, line 23: dolt sql and dolt sql-server are need to use the same directory, so I set it here

jennifersp (approved):
Hi @nkonev, thank you for this contribution! The code changes look good! I'll get this PR merged into main.

nkonev:
Thanks, team!

PR #4832, opened 2022-11-21
Online shallow gc

tbantle22:
Adds a `dolt_gc` procedure that runs shallow online garbage collection (prunes unused table files). A more complete online garbage collection is coming soon


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -973,6 +973,15 @@ func TestDoltReset(t *testing.T) {
...
 }
 
+func TestDoltGC(t *testing.T) {
+	// TODO: This does not work because `db.chunkStore().(nbs.TableFileStore)`
+	// returns not ok in PruneTableFiles
+	t.Skip()
+	for _, script := range DoltGC {
+		enginetest.TestScript(t, newDoltHarness(t), script)
+	}
+}
```
tbantle22, line 983: Is there a way to test the `dolt_gc` procedure here? Seems like the enginetests aren't set up to support a `TableFileStore` chunkstore?

PR #4845, opened 2022-11-22
history system table missing projections

max-hoffman:
history table was returning size-0 rows to the filter in the test case; the filter was erroring for the out of bounds index


fulghum (approved):
Nice fix! 🚢

PR #4846, opened 2022-11-22
Schemas with different PK ordinals should not be equal.

druvv:
Fixes https://github.com/dolthub/dolt/issues/4844

### DIFF
--- a/go/libraries/doltcore/schema/schema.go
+++ b/go/libraries/doltcore/schema/schema.go
@@ -148,6 +148,16 @@ func SchemasAreEqual(sch1, sch2 Schema) bool {
 	if !colCollIsEqual {
 		return false
 	}
+
+	// Pks and Non-pks are in the same order as the key tuple and value tuple fields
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/schema/schema.go
```diff
@@ -148,6 +148,16 @@ func SchemasAreEqual(sch1, sch2 Schema) bool {
 	if !colCollIsEqual {
 		return false
 	}
+
```
andrew-wm-arthur, line 151: do we still need the check on 147 if we're doing these both below?

andrew-wm-arthur (approved):
LGTM

go/libraries/doltcore/schema/schema.go
druvv: Yes. You can have two different orderings for the schema definition that have the same key and value orderings.

```sql
CREATE table A (
    pk int PRIMARY KEY,
    col1 int,
    col2 int
);

CREATE table B (
    col1 int,
    pk int PRIMARY KEY,
    col2 int,
);

-- Should the schema for table A equal table B?
```

PR #4847, opened 2022-11-22
pick new viable underlying index for foreign keys when current one is dropped

JCOR11599:
When dropping underlying secondary indexes used by foreign key, another existing viable one should be used in its place, unless there isn't one.

TODO:
 - we should show duplicate indexes to primary key because MySQL does


go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1186,6 +1186,90 @@ INSERT INTO child_non_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5'
 	}
 }
 
+func TestDropIndex(t *testing.T) {
```
zachmu, line 1189: Probably better expressed as a Sql script test in ddl_queries.go

zachmu (commented):
Good start, needs a little work

go/libraries/doltcore/sqle/sqlddl_test.go
JCOR11599: I don't think there are any queries that show what underlying secondary index a foreign key is using, which is why I wrote the tests here.

go/libraries/doltcore/sqle/tables.go
```diff
@@ -2408,7 +2408,43 @@ func parseFkReferentialAction(refOp sql.ForeignKeyReferentialAction) (doltdb.For
...
+		newIdx, ok, err := findIndexWithPrefix(t.sch, oldIdx.ColumnNames())
+		if err != nil {
+			return nil, nil, err
+		}
+		newFk := fk
+		if ok {
+			newFk.ReferencedTableIndex = newIdx.Name()
+		} else {
+			// if a replacement index wasn't found; it matched on primary key, so use empty string
+			newFk.ReferencedTableIndex = ""
```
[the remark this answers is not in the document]
JCOR11599: I think we can make this assumption, because we do in other places like [here](https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/sqle/tables.go#L2207) and [here](https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/sqle/tables.go#L2241).

This is because GMS also checks if a valid prefix index exists, and errors when there isn't (I updated the tests to confirm this behavior). In the case of switching to a PK as the underlying index, GMS actually considers PKs as indexes and will correctly not error. However, on dolt side, we don't treat PK as an index, so `findIndexWithPrefix` wont be able to find it. A side effect of this is that if there is a viable PK and multiple viable secondary indexes, the PK will only be chosen if we drop all other viable secondary indexes first. 

GMS Code for reference:
 - [drop index] (https://github.com/dolthub/go-mysql-server/blob/main/sql/plan/alter_index.go#L251)
 - [find prefix] (https://github.com/dolthub/go-mysql-server/blob/main/sql/plan/alter_foreign_key.go#L525)

go/libraries/doltcore/sqle/sqlddl_test.go
zachmu: I missed the explicit checks into foreign keys here, my fault.

There are some decent functional tests in GMS for this, so I think you're covered there.

But you don't have coverage for a bunch of scenarios you should be testing. Specifically, there aren't multiple indexes to choose from when you drop the existing one here.

A unit test for findIndexWithPrefix could be more effective for that, your call

zachmu (commented):
See comments

go/cmd/dolt/commands/indexcmds/ls.go
```diff
@@ -113,6 +113,16 @@ func (cmd LsCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
 		}
 	}
 
+	// TODO: delete this
```
zachmu, line 116: ?

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1186,6 +1186,110 @@ INSERT INTO child_non_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5'
...
+	require.True(t, ok)
+	require.Equal(t, "xy", fkChild.TableIndex)
+	require.Equal(t, "ij", fkChild.ReferencedTableIndex)
+
+	// add primary key
+	root, err = ExecuteSql(dEnv, root, "alter table parent add primary key (i, j);")
+	require.NoError(t, err)
+
+	// dropping secondary index ij, should choose ijk
+	root, err = ExecuteSql(dEnv, root, "alter table parent drop index ij;")
```
zachmu, line 1215: Is this deterministic? Odd that it doesn't choose the primary key if suitable

zachmu: Add a note that this isn't by design

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1186,6 +1186,110 @@ INSERT INTO child_non_unq VALUES ('1', 1), ('2', NULL), ('3', 3), ('4', 3), ('5'
...
+	require.Error(t, err)
+
+	// drop tables
+	root, err = ExecuteSql(dEnv, root, "drop table child;")
+	require.NoError(t, err)
+	root, err = ExecuteSql(dEnv, root, "drop table parent;")
+	require.NoError(t, err)
+
+	// setup
+	root, err = ExecuteSql(dEnv, root, "create table parent (i int);")
```
zachmu, line 1249: You have two separate test cases in this method. Name them and run them in t.Run() blocks

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1851,6 +1851,7 @@ func (t *AlterableDoltTable) CreateIndex(ctx *sql.Context, idx sql.IndexDef) err
 	if err != nil {
 		return err
 	}
+	// TODO: why do we need to do this?
```
zachmu, line 1854: ?

zachmu, line 1854: ?

zachmu (approved):
LGTM, just a couple small comments and some code that probably should be removed before checkin

PR #4851, opened 2022-11-23
use tini init system in docker container

jennifersp:
It's helpful for avoiding zombie processes and provides signal handling such as `SIGTERM` from `docker stop`


fulghum (approved):
Nice change! 🚀

PR #4858, opened 2022-11-23
add ORM integration tests job to github nightly jobs

jennifersp:
If the ORM tests fail, the github email notification will likely to go to the github actor of the scheduled job.
It can be set to fixed emails through `dawidd6/action-send-mail@v2` action


coffeegoddd (approved):
LGTM

PR #4871, opened 2022-11-28
fixes for dolt Docker images and ORM integration tests

jennifersp:
- Added WORKDIR for client Dockerfile
- Fix path issue in dolt server docker entrypoint
- Fix ORM integration tests


fulghum (approved):
Looks good! Thanks for jumping on these so fast. ⚡

PR #4872, opened 2022-11-28
Changing `loglevel` param to be case insensitive. 

fulghum:
This allows the `loglevel` CLI param and `log_level` YAML config key to be case insensitive.

Fixes: https://github.com/dolthub/dolt/issues/4864


zachmu (approved):
Nice change

PR #4873, opened 2022-11-28
use range instead of point lookups for non-unique indexes

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/4861
tests here: https://github.com/dolthub/go-mysql-server/pull/1425


fulghum (approved):
Nice fix! ⚡

Issue #4862, opened 2022-11-23
dolt migrate drops secondary indexes, checks, and collations when there's a string column


JCOR11599:
repro:
``` shell
$ dolt init --old-format
Successfully initialized dolt data repository.

$ dolt sql -q "create table t (i int primary key, j int, v varchar(100) collate utf8mb4_0900_ai_ci, index (j), check (j = 0));"
$ dolt sql -q "show create table t"
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                          |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `i` int NOT NULL,
  `j` int,
  `v` varchar(100) COLLATE utf8mb4_0900_ai_ci,
  PRIMARY KEY (`i`),
  KEY `j` (`j`),
  CONSTRAINT `chk_l3afs7p3` CHECK ((`j` = 0))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

$ dolt migrate
migrating database at tmp dir:  C:\Users\james\AppData\Local\Temp\dolt_migration_1669243824915824300
2022-11-23 22:50:24.9564803 +0000 UTC Wrote commit mapping!! [commit_mapping_ref: i5qvor3vop8kj8t4han1t7dvjb1ck9q5]

$ dolt sql -q "show create table t"
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                               |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------+
| t     | CREATE TABLE `t` (
  `i` int NOT NULL,
  `j` int,
  `v` varchar(100),
  PRIMARY KEY (`i`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin |
+-------+------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

This only happens because there's a `varchar` column in the table.

druvv:
[...]

PR #4875, opened 2022-11-28
Changes to schema instantiation

druvv:
Fix for #4862 

Changes the way we instantiate schema.Schema in Dolt. In the past, we've had a couple bugs that have been caused by codepaths that build new schema objects. We've accidentally dropped primary key ordinals twice before and in #4862 we dropped indexes, checks, and collations.

### DIFF
--- a/go/libraries/doltcore/schema/check_coll.go
+++ b/go/libraries/doltcore/schema/check_coll.go
@@ -33,6 +33,8 @@ type CheckCollection interface {
 	DropCheck(name string) error
 	// AllChecks returns all the checks in the collection
 	AllChecks() []Check
+	// Equals returns whether the provided check collection is equal or not.
+	Equals(other CheckCollection) bool
 	// Count returns the size of the collection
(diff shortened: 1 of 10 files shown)


go/libraries/doltcore/schema/check_coll.go
```diff
@@ -33,6 +33,8 @@ type CheckCollection interface {
 	DropCheck(name string) error
 	// AllChecks returns all the checks in the collection
 	AllChecks() []Check
+	// Equals returns whether the provided check collection is equal or not.
+	Equals(other CheckCollection) bool
```
andrew-wm-arthur, line 37: there's really no reason for `CheckCollection` to be an interface, but that's for another day

andrew-wm-arthur (approved):
LGTM, a definite improvement over what we have today. Need to follow this up and remove the deprecated functions

zachmu (approved):
LGTM

PR #4879, opened 2022-11-29
Fix value being read in param suffix

### LINKED ISSUE #4616: Dolt argument parser gets confused by suffixes
Repro bash script:
```
echo "creating db"
rm -rf db
mkdir db
cd db
dolt init

echo "running sql"
dolt sql << EOF
use db;
create table test (id int primary key);
insert into test (id) values (1);
call dolt_add('.');
call dolt_commit('--messageWithExtraRandomSuffixYay111', 'first commit message');
EOF

dolt sql << EOF
use db;
insert into test (id) values (2);
EOF

dolt add .
dolt commit --messageWithAnotherWeirdArg "second commit message"

echo "---------------------"
dolt log --oneline
```

Result output:
```
...
8jdtjuvsrksndcacftesh70oc4dta9rc (HEAD -> main) WithAnotherWeirdArg
uqa8lhjspia0d8shr21f0sgnk8m24vh0 WithExtraRandomSuffixYay111
9plqeeg6se4kfd7kbna7kc6m7m2lt696 Initialize data repository
```

Note that the commit messages are not the strings we passed in, but the suffixes of the `--message` argument: `WithExtraRandomSuffixYay111` and `WithAnotherWeirdArg`.

tanyongzhi:
This commit will:
- Fix an issue with the arg parser, where params suffixes were being recognized as param values (eg. `--messageTest` where `Test` is taken as the value for the `--message` param).

Note that with this new change, values being passed as suffixes to params is only allowed for long for params (ie `--` flags).

Fixes #4616


go/libraries/utils/argparser/args_test.go
```diff
@@ -112,23 +112,23 @@ func TestParsing(t *testing.T) {
 			expectedArgs: []string{"value"},
 		},
 		{
-			name:         "-mvalue",
+			name:         "-m value",
 			options:      []*Option{forceOpt, messageOpt},
-			args:         []string{"b", "-mvalue", "c"},
+			args:         []string{"b", "-m value", "c"},
```
fulghum, line 117: Still being able to support `-mvalue` would be a nice feature to have, and would be consistent with Git's arg parsing behavior (e.g. `git init -bfoo` and `git init -b foo` are both valid).

go/libraries/utils/argparser/args_test.go
```diff
@@ -112,23 +112,23 @@ func TestParsing(t *testing.T) {
...
 			expectedOpts: map[string]string{"message": "value"},
 			expectedArgs: []string{"b", "c"},
 		},
 		{
 			name:         "--messagevalue",
 			options:      []*Option{forceOpt, messageOpt},
-			args:         []string{"b", "-messagevalue", "c"},
+			args:         []string{"b", "--messagevalue", "c"},
```
fulghum, line 124: Disallowing long form args to have value suffixes seems like a good change. This would also be consistent with Git's arg parsing behavior (e.g. `git init --initial-branchfoo` does **not** work, but `git init --initial-branch foo` does).

fulghum (commented):
Hey @tanyongzhi, thanks for sending this contribution over! 🙏 

Overall, this looks pretty good. In the associated issue, you mentioned being hesitant to remove the suffix behavior for the short form args; taking a closer look at this... I think you're correct that we want to keep that ability to specify multiple short form args together, as well as ideally keeping the ability to specify the value for short form args (e.g. `-udolt`). I left two comments with concrete examples of how I'd like to see Dolt match Git's arg parsing behavior. Apologies if I did not do a good job explaining what behavior we were looking for earlier! 😊 I could imagine one day having even smarter arg parsing that would only allow a suffix value for short form args that accept a value and erroring out when a value isn't allowed, but I think that's probably best to tackle as a separate step, down the road. 

If you're willing and able to update this PR, we'd love to accept it with that tweak to the behavior.

go/libraries/utils/argparser/args_test.go
tanyongzhi: Ah I see, I think I misunderstood what you said earlier. I'll try and make the appropriate changes.

tanyongzhi: Ah my bad, I seem to have reversed the behavior for short and long form flags.

fulghum: No worries at all! Apologies if I didn't describe it clearly enough.

fulghum (approved):
Looks great @tanyongzhi! Thank you for contributing this nice fix! 🙏 

I only had one or two minor formatting comments, so I'll go ahead and commit those suggestions, then get this merged to main for you. 

Thank you yet again for another great contribution to Dolt! 🙏 🎉

PR #4894, opened 2022-11-30
Truncate table should not drop conflicts or CVs

druvv:


go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -220,11 +220,6 @@ func (tx *DoltTransaction) doCommit(
 
 			// otherwise (not a ff), merge the working sets together
 			start := time.Now()
-			// TODO: this loses track of merge conflicts in the working set, clearing them out and replacing them with any
```
druvv, line 223: Not true for the new format. Conflicts get merged.

zachmu (approved):
LGTM, thx

PR #4900, opened 2022-12-01
fixes dolt table export appending success message to csv when using redirect syntax

stephkyou:
`dolt table export` was previously appending "Successfully exported data." to the end of the csv file when using the redirect syntax (e.g. `dolt table export hospitals > hospitals.csv`). The success message didn't get appended when redirect syntax wasn't used (e.g. `dolt table export hospitals hospitals.csv`). This fix makes it so that both formats work as expected.

issue: https://github.com/dolthub/dolt/issues/4889


fulghum (approved):
Looks great! Nice work @stephkyou! 🙌

PR #4909, opened 2022-12-02
go/store/nbs: cleanup NBS interfaces

andy-wm-arthur:
- refactored `conjoiner` to `conjoinStrategy` which is now only responsible for deciding when to conjoin and which `tableSpecs` to conjoin. A common implementation of `conjoin()` is used in all cases
- added `getRecordRanges(requests []getRecord)` to the `chunkSource` interface, rather than asserting past the interface in `NomsBlockStore.GetChunkLocations`


andy-wm-arthur:
#benchmark

andy-wm-arthur:
#benchmark

andy-wm-arthur:
#benchmark

tbantle22 (approved):

PR #4913, opened 2022-12-02
Remove `GetGCKeepers` 

tbantle22:


go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1165,6 +1165,12 @@ func (ddb *DoltDB) GC(ctx context.Context, uncommittedVals ...hash.Hash) error {
 	if err != nil {
 		return err
 	}
+
+	uncommittedVals, err := ddb.getGCKeepers(ctx, datasets)
```
andy-wm-arthur, line 1169: I think this is unnecessary now, because the roots of each dataset are reachable from the root of datasets map. it doesn't hurt to add them as keeps tho

andy-wm-arthur (approved):
LGTM!

andy-wm-arthur (approved):

PR #4923, opened 2022-12-05
found minor bug in validate.go

JCPrice0024:
In validate.go I found a small bug where there is a comparison on the same object (len(new) != len(new)). I assume this was meant to compare len(new) with len(old) so I fixed that here.


fulghum:
Hey @JCPrice0024, thank you for taking the time to send us this fix!! 🙏 

Code change looks good to me. I just kicked off the CI tests and we'll make sure they pass before we merge (`Format PR` is expected to fail for community PRs, so don't worry about that one).

fulghum (approved):

JCPrice0024:
@fulghum glad I could help let me know if there is anything else to be done!

PR #4924, opened 2022-12-05
found another minor bug in results.go

JCPrice0024:
Found the same kind of bug as before this time in results.go. I changed the comparison to be correct as before.

Question about line 36. We are requiring exact order and not just content, is that correct or should we sort the args before comparing them?

### DIFF
--- a/go/libraries/utils/argparser/results.go
+++ b/go/libraries/utils/argparser/results.go
@@ -28,8 +28,10 @@ type ArgParseResults struct {
 	parser  *ArgParser
 }
 
+// Equals res and other are only considered equal if the order and contents of their arguments
+// are the same.
 func (res *ArgParseResults) Equals(other *ArgParseResults) bool {
-	if len(res.Args) != len(other.Args) || len(res.options) != len(res.options) {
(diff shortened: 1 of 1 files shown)


fulghum:
Hey @JCPrice0024, another good catch! Thanks for sending this fix over, too. 🙏 

> Question about line 36. We are requiring exact order and not just content, is that correct or should we sort the args before comparing them?

Yup, you're correct that the code currently requires the same arg order in each instance. Sorting the args isn't a bad idea at all, although... I don't actually see this function being used currently. Because of that, I'd say it's great to fix the correctness bug and it might be nice to just document the current behavior with requiring the exact same arg order as part of a function comment. I'm not sure that it's worth sorting the args since I couldn't find any referencing using this function.

JCPrice0024:
Got you, I commented the behavior and squashed it.

fulghum (approved):
Great catch on this one!

fulghum:
Merged to main 🎉  

Thanks a ton for sending these PRs over. 🙏  Let us know if there's anything we can help you with!

PR #4929, opened 2022-12-06
Adds skipped bats for silent merge failures

druvv:
5 skipped bats showing how merge silently fails today.

After this PR, next steps is to throw an error if we are three-way merging these kinds of schema changes. Still thinking about how to detect these issues and fail-fast.


andy-wm-arthur (approved):
LGTM, good start on finding edge cases here. 

My opinion is that we need to leverage the SQL DDL path to address these issues comprehensively

PR #4958, opened 2022-12-09
go/commands: added verbose mode to filter-branch

andy-wm-arthur:


druvv (approved):
LGTM. Once we verify that the commit isn't changing, I wonder what the next steps are in debugging Tome's issue.

PR #4959, opened 2022-12-09
First pass at online GC

tbantle22:


go/store/nbs/store.go
```diff
@@ -1412,7 +1432,17 @@ func (nbs *NomsBlockStore) PruneTableFiles(ctx context.Context) (err error) {
...
 
+func (nbs *NomsBlockStore) setGCInProgress(inProgress bool) {
+	nbs.cond.L.Lock()
+	defer nbs.cond.L.Unlock()
+	nbs.gcInProgress = inProgress
+	nbs.cond.Broadcast()
+}
+
 func (nbs *NomsBlockStore) MarkAndSweepChunks(ctx context.Context, last hash.Hash, keepChunks <-chan []hash.Hash, dest chunks.ChunkStore) error {
+	nbs.setGCInProgress(true)
```
andy-wm-arthur, line 1443: we need some guard here to ensure that only one GC operation is running at a time. Either an `atomic.Bool` with CAS or maybe a reader-writer lock. I think a boolean would fit better with where we want to take this next

tbantle22:
#benchmark

andy-wm-arthur (approved):
LGTM!

PR #4968, opened 2022-12-12
[no-release-notes] GMS bump

JCOR11599:
We don't support UnresolvedFunctions as dolt table arguments, but those functions are fuly resolved when running them prepared, so additional logic is needed to skip them.


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -111,27 +110,19 @@ func TestSingleScript(t *testing.T) {
 	t.Skip()
 	var scripts = []queries.ScriptTest{
 		{
-			Name: "trigger before update, with indexed update",
```
max-hoffman, line 114: nit `git checkout main -- go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go`

max-hoffman (approved):
mostly LGTM one comment

PR #4969, opened 2022-12-12
Removed deprecated dolt functions

Hydrocharged:


go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -15,16 +15,314 @@
...
+	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqlserver"
+	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+	"github.com/dolthub/dolt/go/libraries/utils/filesys"
+)
+
+var (
+	EmptyBranchNameErr = errors.New("error: cannot branch empty string")
```
zachmu, line 36: "invalid branch name %s" is better

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -15,16 +15,314 @@
...
+
+		return false, nil
+	})
+}
+
+func loadConfig(ctx *sql.Context) *env.DoltCliConfig {
+	// When executing branch actions from SQL, we don't have access to a DoltEnv like we do from
+	// within the CLI. We can fake it here enough to get a DoltCliConfig, but we can't rely on the
+	// DoltEnv because tests and production will run with different settings (e.g. in-mem versus file).
+	dEnv := env.Load(ctx, env.GetCurrentUserHomeDir, filesys.LocalFS, doltdb.LocalDirDoltDB, "")
```
zachmu, line 253: So I traced this down and the only reason we need this is to do this:

```go
		ms, err := doltdb.NewCommitSpec(env.GetDefaultInitBranch(config))
```

Super lame use of resources. Can you add a TODO to fix this with the note that we super don't need an entire client config here? (It should be available in the server context in some other way)

go/libraries/doltcore/sqle/dprocedures/dolt_pull.go
```diff
@@ -15,16 +15,218 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-statsCh:
+		default:
+		}
+	}
+}
+
+func progFunc(ctx context.Context, progChan <-chan pull.PullProgress) {
```
zachmu, line 193: Can you actually get rid of all these prog updater funcs while you're in here? You can just pass nil to the pull methods instead. These don't do anything and actually take a bunch of compute for no reason

zachmu (approved):
LGTM

timsehn:
@Hydrocharged Can we get this one in please?

Hydrocharged:
> @Hydrocharged Can we get this one in please?

Yeah the bats test runner was broken and Dustin just fixed today. I’m doing a release after I can get this in.

PR #4971, opened 2022-12-12
Reset source branch after checkout

PavelSafronov:
When checking out a branch, clear the source branch's working set only after the destination checkout succeeds without errors.

### DIFF
--- a/go/libraries/doltcore/env/actions/reset.go
+++ b/go/libraries/doltcore/env/actions/reset.go
@@ -136,7 +136,17 @@ func ResetHardTables(ctx context.Context, dbData env.DbData, cSpecStr string, ro
 	return resetHardTables(ctx, dbData, cSpecStr, roots)
 }
 
-func ResetHard(ctx context.Context, dEnv *env.DoltEnv, cSpecStr string, roots doltdb.Roots, headRef *ref.DoltRef, ws *doltdb.WorkingSet) error {
+// ResetHard resets the working, staged, and head based on the given parameters.
+// The reset can be performed on a non-current branch and working set.
+// Returns an error if the reset fails.
+func ResetHard(
+	ctx context.Context,
+	dEnv *env.DoltEnv,
+	cSpecStr string,
+	roots doltdb.Roots,
+	headRef *ref.DoltRef,
+	ws *doltdb.WorkingSet,
+) error {
 	dbData := dEnv.DbData()
 
 	newHead, roots, err := resetHardTables(ctx, dbData, cSpecStr, roots)
--- a/go/libraries/doltcore/env/actions/reset.go
+++ b/go/libraries/doltcore/env/actions/reset.go
@@ -17,6 +17,7 @@ package actions
 import (
 	"context"
 	"fmt"
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/env/actions/reset.go
```diff
@@ -136,26 +136,21 @@ func ResetHardTables(ctx context.Context, dbData env.DbData, cSpecStr string, ro
 	return resetHardTables(ctx, dbData, cSpecStr, roots)
 }
 
-func ResetHard(ctx context.Context, dEnv *env.DoltEnv, cSpecStr string, roots doltdb.Roots) error {
+func ResetHard(ctx context.Context, dEnv *env.DoltEnv, cSpecStr string, roots doltdb.Roots, headRef *ref.DoltRef, ws *doltdb.WorkingSet) error {
```
fulghum, line 139: (minor) Would be nice to add some godocs for this function, especially since the args list is getting longer and it's a public/exposed function from this package.

fulghum (commented):
Code changes make sense; is there a BATS test we can add to trigger this case so the fix has some test coverage?

fulghum (approved):
Just chatted with Pavel in person. We couldn't come up with a good way to trigger this condition from a test yet; we'll see if zachmu has any ideas, otherwise the changes look good and definitely seem safer in this order.

go/libraries/doltcore/env/actions/reset.go
```diff
@@ -136,26 +136,31 @@ func ResetHardTables(ctx context.Context, dbData env.DbData, cSpecStr string, ro
...
-func ResetHard(ctx context.Context, dEnv *env.DoltEnv, cSpecStr string, roots doltdb.Roots) error {
+// ResetHard resets the working, staged, and head based on the given parameters.
+// The reset can be performed on a non-current branch and working set.
+// Returns an error if the reset fails.
+func ResetHard(
+	ctx context.Context,
+	dEnv *env.DoltEnv,
+	cSpecStr string,
+	roots doltdb.Roots,
+	headRef *ref.DoltRef,
```
zachmu, line 147: Why is this a pointer?

go/libraries/doltcore/env/actions/reset.go
```diff
@@ -136,26 +136,31 @@ func ResetHardTables(ctx context.Context, dbData env.DbData, cSpecStr string, ro
 	return resetHardTables(ctx, dbData, cSpecStr, roots)
 }
 
-func ResetHard(ctx context.Context, dEnv *env.DoltEnv, cSpecStr string, roots doltdb.Roots) error {
+// ResetHard resets the working, staged, and head based on the given parameters.
```
zachmu, line 139: and head to the ones in the Roots provided.

zachmu (approved):
LGTM!

go/libraries/doltcore/env/actions/reset.go
PavelSafronov: Good question, not sure. No longer a pointer.

In general, which objects should/shouldn't be passed as pointers?
You mentioned `doltdb.Roots` as an example of a struct that we always pass by value. How are we deciding this?

go/libraries/doltcore/env/actions/reset.go
PavelSafronov: Docs added.

go/libraries/doltcore/env/actions/reset.go
PavelSafronov: Updated documentation as suggested.

go/libraries/doltcore/env/actions/reset.go
zachmu: I'll come talk about it

PR #4973, opened 2022-12-12
Fix working set not found error for dolt migrate

druvv:
In dolt versions <= 0.26.10, a working set was not created when a branch was created. In newer dolt versions it is created. Migrate expected a working set to exist for every branch ref.

Now, if a working set is missing we pretend it exists.


andy-wm-arthur (approved):
LGTM

PR #4976, opened 2022-12-13
Fixed DELIMITER

Hydrocharged:
Previously we required delimiters to have the previous delimiter at the end, but that requirement no longer exists, which matches MySQL's behavior.


integration-tests/bats/sql.bats
```diff
@@ -2586,6 +2586,7 @@ SQL
 }
 
 @test "sql: vertical query format in sql shell" {
+    skip "Not sure how to fix the error, nor what exactly is being tested"
```
zachmu, line 2589: Query ending in \G should give vertical result output

go/cmd/dolt/commands/sql.go
```diff
@@ -1002,30 +983,18 @@ func runShell(ctx context.Context, se *engine.SqlEngine, mrEnv *env.MultiRepoEnv
 		query = strings.TrimSuffix(query, shell.LineTerminator())
 		resultFormat := se.GetResultFormat()
 
-		// TODO: it would be better to build this into the statement parser rahter than special case it here
+		// TODO: it would be better to build this into the statement parser rather than special case it here
 		for _, terminator := range verticalOutputLineTerminators {
 			if strings.HasSuffix(query, "\\G") {
```
zachmu, line 988: This is a bug: it should use `terminator` instead of `"\\G"` here

As a result, `\G` works but `\g` does not

zachmu (approved):
LGTM, I played with it locally and it seems to work fine

go/cmd/dolt/commands/sql.go
Hydrocharged: Ohhh I see, that makes sense! I’ll fix this

PR #5005, opened 2022-12-15
Perf improvements for remote read replicas

zachmu:
This significantly speeds up the process of syncing a remote-based replica by 1) pulling chunks once, rather than once per branch in the case of replicas that want everything, and 2) by eliminating a bunch of duplicated and wasted work. Latency for up-to-date case is now <100 ms.


go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -131,132 +151,189 @@ func (rrd ReadReplicaDatabase) PullFromRemote(ctx *sql.Context) error {
...
+		}
 
-		srcDBCommitA, err := rrd.limiter.Run(ctx, branch.String(), func() (any, error) {
-			// fetch remote tracking ref only
-			var srcDBCommit *doltdb.Commit
-			{
-				srcDBCommit, err = actions.FetchRemoteBranch(fetchCtx, rrd.tmpDir, rrd.remote, rrd.srcDB, rrd.ddb, branch, actions.NoopRunProgFuncs, actions.NoopStopProgFuncs)
-				if err != nil {
-					return nil, err
+		err = rrd.ddb.PullChunks(ctx, rrd.tmpDir, rrd.srcDB, srcRoot, nil, nil)
```
reltuk, line 257: Nit: It seems more correct, clearer and very marginally more performance to make an interface that will pull more than one addr at once and then to pass all the addrs in `remoteRefs` to it.

reltuk (approved):
LGTM! Thanks for making that change.

PR #5012, opened 2022-12-16
Move TableFileStore from nbs to chunks

tbantle22:


go/libraries/doltcore/env/actions/clone.go
```diff
@@ -32,9 +32,9 @@ import (
 	"github.com/dolthub/dolt/go/libraries/utils/filesys"
 	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
 	"github.com/dolthub/dolt/go/libraries/utils/strhelp"
+	stchunks "github.com/dolthub/dolt/go/store/chunks"
```
andy-wm-arthur, line 35: why the package alias here?

andy-wm-arthur (approved):
LGTM!

go/libraries/doltcore/env/actions/clone.go
tbantle22: There's a `chunks` variable on line 94 that conflicts. Can change the name of that if that's preferred?

PR #5018, opened 2022-12-16
go/store/{blobstore,nbs}: Added conjoin for GCS remotes

andy-wm-arthur:


go/store/blobstore/blobstore.go
```diff
@@ -22,10 +22,20 @@ import (
...
+	Get(ctx context.Context, key string, br BlobRange) (rc io.ReadCloser, version string, err error)
+
+	// Put creates a new blob from |reader| keyed by |key|, it returns the latest store version.
+	Put(ctx context.Context, key string, reader io.Reader) (version string, err error)
+
+	// CheckAndPut updates the blob keyed by |key| using a check-and-set on |expectedVersion|.
+	CheckAndPut(ctx context.Context, expectedVersion, key string, reader io.Reader) (version string, err error)
+
+	// Contatenate creates a new blob named |key| by concatenating |sources|.
+	Contatenate(ctx context.Context, key string, sources []string) (version string, err error)
```
reltuk, line 38: Concatenate...

go/store/blobstore/gcs.go
```diff
@@ -155,3 +162,61 @@ func (bs *GCSBlobstore) CheckAndPut(ctx context.Context, expectedVersion, key st
...
+	// GCS compose has a batch size limit,
+	// recursively compose sources
+	for len(sources) > composeBatch {
+		// compose subsets of |sources| in batches,
+		// store tmp composite objects in |next|
+		var next []string
+		for len(sources) > 0 {
+			tmp := uuid.New().String()
+			batch := min(composeBatch, len(sources))
+			_, err := bs.composeObjects(ctx, tmp, sources[:batch])
```
reltuk, line 176: We expect to call this with ~250 sources in the current code paths and common configurations?

It would be nice, but maybe not necessary, to parallelize the calls to composeObjects at each level...

```go
var next []string, batches [][]string
for len(sources > 0) {
    batchSz := min(composeBatch, len(sources))
    batches = append(batches, sources[:batchSz])
    next = append(next, uuid.New().String())
    sources = sources[batchSz:]
}
eg, ctx := errgroup.WithContext(ctx)
for i := 0; i < len(batches); i++ {
    i := i
    eg.Go(func() error {
        _, err := bs.composeObjects(ctx, next[i], batches[i])
        return err
    })
}
err := eg.Wait()
if err != nil {
    return "", err
}
sources = next

File: go/store/blobstore/gcs.go [hunk 3]
```diff
@@ -155,3 +162,61 @@ func (bs *GCSBlobstore) CheckAndPut(ctx context.Context, expectedVersion, key st
...
+
+func (bs *GCSBlobstore) composeObjects(ctx context.Context, composite string, sources []string) (gen string, err error) {
+	if len(sources) > composeBatch {
+		return "", fmt.Errorf("too many objects to compose (%d > %d)", len(sources), composeBatch)
+	}
+
+	var a *storage.ObjectAttrs
+	objects := make([]*storage.ObjectHandle, len(sources))
+	for i := range objects {
+		oh := bs.bucket.Object(path.Join(bs.prefix, sources[i]))
```
reltuk, line 196: Again a source for concurrency...maybe less important than doing the top-level composes in parallel.

go/store/blobstore/local.go
```diff
@@ -232,3 +211,22 @@ func (bs *LocalBlobstore) Exists(ctx context.Context, key string) (bool, error)
...
+		path := filepath.Join(bs.RootDir, sources[i]) + bsExt
+		if readers[i], err = os.Open(path); err != nil {
+			return "", err
+		}
+	}
+
+	ver, err = bs.Put(ctx, key, io.MultiReader(readers...))
+
+	for i := range readers {
+		if cerr := readers[i].(io.Closer).Close(); err != nil {
```
reltuk, line 227: Nit: Should Reader above be ReadCloser?

go/store/blobstore/local.go
```diff
@@ -232,3 +211,22 @@ func (bs *LocalBlobstore) Exists(ctx context.Context, key string) (bool, error)
 
 	return err == nil, err
 }
+
+func (bs *LocalBlobstore) Contatenate(ctx context.Context, key string, sources []string) (ver string, err error) {
+	readers := make([]io.Reader, len(sources))
+	for i := range readers {
+		path := filepath.Join(bs.RootDir, sources[i]) + bsExt
+		if readers[i], err = os.Open(path); err != nil {
+			return "", err
```
reltuk, line 220: Nit: Is there a way to restructure this slightly so we get to Close() the opened readers in the error case without relying on in gc finalization?

reltuk (approved):
A few small comments. LGTM!

go/store/blobstore/local.go
andy-wm-arthur: initially it was, but `io.MultiReader()` can only take a `[]io.Reader` not an `[]io.ReadCloser` (furp)

PR #5022, opened 2022-12-19
[no-release-notes] go/store/datas/pull: Update Pull interfaces to take a slice of hashes to pull, not just one.

reltuk:
This allows client code to be more efficient when it needs to pull multiple hashes.


go/store/chunks/tablefile.go
```diff
@@ -0,0 +1,58 @@
+package chunks
```
zachmu, line 1: Seems like this got included by accident?

zachmu (approved):
LGTM!

PR #5037, opened 2022-12-21
makes batching default for dolt dump

stephkyou:
Currently `dolt dump` only batches `INSERT` statements when the `--batch` flag is included. This change flips the behavior to default to batching `INSERT` statements with the `--batch` flag now turning off this behavior.

fixes: https://github.com/dolthub/dolt/issues/4954


go/cmd/dolt/commands/dump.go
```diff
@@ -90,7 +90,7 @@ func (cmd DumpCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(filenameFlag, "fn", "file_name", "Define file name for dump file. Defaults to `doltdump.sql`.")
 	ap.SupportsString(directoryFlag, "d", "directory_name", "Define directory name to dump the files in. Defaults to `doltdump/`.")
 	ap.SupportsFlag(forceParam, "f", "If data already exists in the destination, the force flag will allow the target to be overwritten.")
-	ap.SupportsFlag(batchFlag, "", "Returns batch insert statements wherever possible.")
+	ap.SupportsFlag(batchFlag, "", "Turns off batching insert statements wherever possible.")
```
zachmu, line 93: If we're going to make this change, then change the flag to --no-batch and make the existing --batch a no-op

go/cmd/dolt/commands/dump.go
```diff
@@ -91,6 +92,7 @@ func (cmd DumpCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(directoryFlag, "d", "directory_name", "Define directory name to dump the files in. Defaults to `doltdump/`.")
 	ap.SupportsFlag(forceParam, "f", "If data already exists in the destination, the force flag will allow the target to be overwritten.")
 	ap.SupportsFlag(batchFlag, "", "Returns batch insert statements wherever possible.")
```
zachmu, line 94: Mention this is enabled by default

go/cmd/dolt/commands/dump.go
```diff
@@ -403,18 +410,19 @@ func getDumpOptions(fileName string, rf string) *dumpOptions {
 
 // newTableArgs returns tableOptions of table name and src table location and dest file location
 // corresponding to the input parameters
-func newTableArgs(tblName string, destination mvdata.DataLocation, batched bool, autocommitOff bool) *tableOptions {
+func newTableArgs(tblName string, destination mvdata.DataLocation, batched bool, batchedOff bool, autocommitOff bool) *tableOptions {
 	return &tableOptions{
 		tableName:     tblName,
 		dest:          destination,
 		batched:       batched,
```
zachmu, line 417: Definitely don't want both batched and !batched here

zachmu: And this new param isn't being used by anything anyway

go/cmd/dolt/commands/dump.go
```diff
@@ -403,18 +410,19 @@ func getDumpOptions(fileName string, rf string) *dumpOptions {
...
+		batchedOff:    batchedOff,
 		autocommitOff: autocommitOff,
 	}
 }
 
 // dumpTables returns nil if all tables is dumped successfully, and it returns err if there is one.
 // It handles only csv and json file types(rf).
-func dumpTables(ctx context.Context, root *doltdb.RootValue, dEnv *env.DoltEnv, force bool, tblNames []string, rf string, dirName string, batched bool) errhand.VerboseError {
+func dumpTables(ctx context.Context, root *doltdb.RootValue, dEnv *env.DoltEnv, force bool, tblNames []string, rf string, dirName string, batched bool, batchedOff bool) errhand.VerboseError {
```
zachmu, line 425: Can remove this new param and its use below

go/cmd/dolt/commands/dump.go
```diff
@@ -91,6 +92,7 @@ func (cmd DumpCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(directoryFlag, "d", "directory_name", "Define directory name to dump the files in. Defaults to `doltdump/`.")
 	ap.SupportsFlag(forceParam, "f", "If data already exists in the destination, the force flag will allow the target to be overwritten.")
 	ap.SupportsFlag(batchFlag, "", "Returns batch insert statements wherever possible.")
+	ap.SupportsFlag(noBatchFlag, "", "Turns off batching insert statements wherever possible.")
```
zachmu, line 95: "Always emit one row per statement, rather than batching multiple rows into each statement."

go/cmd/dolt/commands/dump.go
zachmu: Also just return, not returns

zachmu (commented):
New param looks good, see other comments

integration-tests/bats/dump.bats
```diff
@@ -177,7 +201,7 @@ teardown() {
     dolt add .
     dolt commit -m "create tables"
 
-    run dolt dump
```
zachmu, line 180: Any reason to use --no-batch here?

zachmu (approved):
LGTM

integration-tests/bats/dump.bats
stephkyou: With batching we get a `CREATE TABLE 'dolt_procedures'` line that will error out upon import.

PR #5038, opened 2022-12-21
Block merges that corrupt indexes

druvv:
Today, many merges in the new format can leave indexes in a corrupted state. This PR blocks those merges as a short-term fix. In the long term we will make these merges automatic.


andy-wm-arthur (approved):
LGTM

PR #5044, opened 2022-12-22
prevent ddl to database through branch syntax

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/4352

companion pr: https://github.com/dolthub/go-mysql-server/pull/1500


zachmu (commented):
Let's break this up into its separable components. You're fixing two different related issues here, let's consider them separately so we think about them more clearly.

The issues:

1) Referencing a non-checked-out branch via the db/branch syntax should be read-only
2) USE db/branch behaves differently from call dcheckout(branch)

Let's first concentrate on fixing 1) so we can talk about it in isolation and discuss the consequences (it also requires no GMS changes). Trim this PR down to just that and move the changes for 2) to another branch / PR for the time being.

timsehn:
How we looking here?

timsehn:
Are we ever going to merge this?

PR #5049, opened 2022-12-22
fix panic for `select active_branch()` on revision db

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/5043

it errors now


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -786,6 +786,10 @@ func TestDoltRevisionDbScripts(t *testing.T) {
 				Query:    "use mydb/" + commithash,
 				Expected: []sql.Row{},
 			},
+			{
+				Query:          "select active_branch();",
+				ExpectedErrStr: "working set not found",
```
zachmu, line 791: Not a great error message. This should say "error: no branch selected"

zachmu: Tests for this function after `use mydb/mybranch`?

JCOR11599: They are here: https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/sqle/enginetest/dolt_server_test.go#L122

zachmu (approved):
LGTM

PR #5064, opened 2022-12-29
adding --set as valid flag for dolt config

stephkyou:
Adds `--set` as a valid flag for `dolt config`. `--set` does the same as `--add`.

fixes: https://github.com/dolthub/dolt/issues/5050


zachmu (approved):
LGTM, nice work!

PR #5074, opened 2022-12-30
fixes for changes on info_schema tables

jennifersp:
Depends on https://github.com/dolthub/go-mysql-server/pull/1501


go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
```diff
@@ -291,10 +291,11 @@ func SqlRowAsCreateFragStmt(r sql.Row) (string, error) {
...
 	}
 	defStr := sqlparser.String(defStmt)
-	if typeStr == "TRIGGER" { // triggers need the create trigger <trig_name> to be cut off
+	// triggers and views need 'CREATE VIEW <view_name>' and 'CREATE TRIGGER <trig_name>' to be cut off
+	if typeStr == "TRIGGER" {
 		defStr = defStr[len("CREATE TRIGGER ")+len(nameStr)+1:]
-	} else { // views need the prefixed with "AS"
-		defStr = "AS " + defStr
+	} else {
+		defStr = defStr[len("CREATE VIEW ")+len(nameStr)+1:]
```
fulghum, line 298: I know this pattern was already here, but thought it was worth pointing out that this seems like it's going to break when options are used and this prefix string doesn't match (e.g. "CREATE TRIGGER IF NOT EXISTS myTrigger..."). Do those extra options already show up here, too, and break this prefix matching? 

I think this is okay, since it's the same behavior as today, but it's worth recognizing that this could break in the future.

fulghum (approved):
Looks good! 🚢  Just one or two small requests to check out.

PR #5089, opened 2023-01-04
counting milliseconds wrong

JCOR11599:


zachmu (approved):
LGTM, but write a test as long as you're in there, just a couple simple cases

(unit test in the same package)

zachmu (commented):
This test is not great:

1) It sleeps. It should construct time objects manually
2) It doesn't test fractional seconds (which is where the bug was)
3) Its float diff tolerance is way too loose

zachmu (approved):
LGTM, thanks for the changes

PR #5091, opened 2023-01-04
go/store/nbs: fixed manifest flush check

andy-wm-arthur:


go/store/nbs/journal.go
```diff
@@ -247,8 +247,7 @@ func (j *chunkJournal) Update(ctx context.Context, lastLock addr, next manifestC
 	}
 
 	// if |next| has a different table file set, flush to |j.backing|
-	if j.contents.lock != next.lock || j.contents.gcGen != next.gcGen {
-		// todo: why is this necessary?
+	if !equalSpecs(j.contents.specs, next.specs) {
```
andy-wm-arthur, line 250: The manifest lock is generated using both the table spec hashes _and_ the root value hash, so it naturally changes with each `Commit`/`Update`. We only want to flush to the backing manifest here if the set of table specs changes

PR #5093, opened 2023-01-04
[no-release-notes] spatial types refactor

JCOR11599:


JCOR11599:
moving changes to auto-bump

PR #5096, opened 2023-01-05
[no-release-notes] Use in mem blobstore in MemFactory

tbantle22:
Removed datas/pull and noms_sync. Can ignore those changes

### DIFF
--- a/go/store/nbs/table_persister.go
+++ b/go/store/nbs/table_persister.go
@@ -64,7 +64,7 @@ type tableFilePersister interface {
 	tablePersister
 
 	// CopyTableFile copies the table file with the given fileId from the reader to the TableFileStore.
-	CopyTableFile(ctx context.Context, r io.ReadCloser, fileId string) error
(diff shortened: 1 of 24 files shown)


go/store/nbs/table_persister.go
```diff
@@ -63,6 +63,9 @@ type tablePersister interface {
 type tableFilePersister interface {
 	tablePersister
 
+	// CopyTableFile copies the table file with the given fileId from the reader to the TableFileStore.
+	CopyTableFile(ctx context.Context, r io.ReadCloser, fileId string, chunkCount uint32) error
+
 	// Path returns the file system path.
 	Path() string
```
andy-wm-arthur, line 70: i think we want to remove this right?

andy-wm-arthur (approved):
LGTM

go/store/nbs/table_persister.go
tbantle22: We talked about this and I think decided not to remove it because it's used in remotesrv? https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/remotesrv/grpc.go#L127

PR #5104, opened 2023-01-05
Change current README edit to be more out of band

timsehn:
I was not a fan of the wording of the previous edit but thought the intent was good.


fulghum (approved):
Looks good! Makes this section a little easier to digest.

PR #5105, opened 2023-01-06
add --schema-only flag for `dolt dump`

jennifersp:
--schema-only is used for dump table schemas with no data into sql file.

### DIFF
--- a/go/cmd/dolt/commands/dump.go
+++ b/go/cmd/dolt/commands/dump.go
@@ -17,7 +17,6 @@ package commands
 import (
 	"context"
 	"fmt"
-	"github.com/dolthub/dolt/go/libraries/doltcore/table/untyped/sqlexport"
 	"os"
 	"path/filepath"
 	"strings"
@@ -33,6 +32,7 @@ import (
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table"
 	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
+	"github.com/dolthub/dolt/go/libraries/doltcore/table/untyped/sqlexport"
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
 	"github.com/dolthub/dolt/go/libraries/utils/filesys"
 	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/dump.go
```diff
@@ -94,6 +96,7 @@ func (cmd DumpCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsFlag(batchFlag, "", "Return batch insert statements wherever possible, enabled by default.")
 	ap.SupportsFlag(noBatchFlag, "", "Always emit one row per statement, rather than batching multiple rows into each statement.")
 	ap.SupportsFlag(noAutocommitFlag, "na", "Turns off autocommit for each dumped table. Used to speed up loading of outputted sql file")
+	ap.SupportsFlag(schemaOnlyFlag, "", "Dumping a table schema with no data to sql file")
```
fulghum, line 99: I was looking at our argument doc strings for the CLI the other day and we are mostly consistent around an imperative verb form to start them off (e.g. "Define the type",  "Return batch"). It would be good to try and standardize around that more so the docs look more consistent. I also added consistent punctuation at the end of the last two doc strings to match the others. 

```suggestion
	ap.SupportsFlag(noBatchFlag, "", "Emit one row per statement, instead of batching multiple rows into each statement.")
	ap.SupportsFlag(noAutocommitFlag, "na", "Turn off autocommit for each dumped table. Useful for speeding up loading of output SQL file.")
	ap.SupportsFlag(schemaOnlyFlag, "", "Dump a table's schema, without including any data, to the output SQL file.")
```

go/cmd/dolt/commands/dump.go
```diff
@@ -173,24 +184,24 @@ func (cmd DumpCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		err = dumpNonSqlTables(ctx, root, dEnv, force, tblNames, jsonFileExt, outputFileOrDirName, false)
 		if err != nil {
 			return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 		}
 	case parquetFileExt:
-		err = dumpTables(ctx, root, dEnv, force, tblNames, parquetFileExt, name, false)
+		err = dumpNonSqlTables(ctx, root, dEnv, force, tblNames, parquetFileExt, outputFileOrDirName, false)
 		if err != nil {
 			return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 		}
```
fulghum, line 207: (minor) Could these three cases also be combined? Seems like they are the same, except for one param, which should be `resFormat`.

fulghum (approved):
Nice work @jennifersp! 🙌

PR #5114, opened 2023-01-06
add "create database" header to dolt dump sql files

stephkyou:
automatically adds `CREATE DATABASE IF NOT EXISTS ...; USE db;` to all dolt dump sql files, preventing errors from importing to dolt in non database directories

fixes: https://github.com/dolthub/dolt/issues/4837

### DIFF
--- a/go/cmd/dolt/commands/dump.go
+++ b/go/cmd/dolt/commands/dump.go
@@ -46,6 +46,7 @@ const (
 	noBatchFlag      = "no-batch"
 	noAutocommitFlag = "no-autocommit"
 	schemaOnlyFlag   = "schema-only"
+	noCreateDbFlag   = "no-create-db"
 
 	sqlFileExt     = "sql"
 	csvFileExt     = "csv"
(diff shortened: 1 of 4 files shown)


go/cmd/dolt/commands/dump.go
```diff
@@ -467,3 +476,36 @@ func addBulkLoadingParadigms(dEnv *env.DoltEnv, fPath string) error {
...
 }
+
+// addCreateDatabaseHeader adds a CREATE DATABASE header to prevent `no database selected` errors on dump file ingestion.
+func addCreateDatabaseHeader(dEnv *env.DoltEnv, fPath, dbName string) error {
+	writer, err := dEnv.FS.OpenForWriteAppend(fPath, os.ModePerm)
+	if err != nil {
+		return err
+	}
+
+	_, err = writer.Write([]byte("CREATE DATABASE IF NOT EXISTS " + dbName + "; USE " + dbName + "; \n"))
```
fulghum, line 487: (minor) It might be worth wrapping `dbName` in backtick/identifier quotes. That way we can support databases with MySQL reserved words that are allowed to be used, but must be backtick/identifier quoted. For example, `interval` is a reserved MySQL word that needs to be backtick quoted. Would be good to add a test for that case, too. 

MySQL docs on keywords and reserved words: https://dev.mysql.com/doc/refman/8.0/en/keywords.html

go/cmd/dolt/commands/dump.go
```diff
@@ -167,7 +167,16 @@ func (cmd DumpCmd) Exec(ctx context.Context, commandStr string, args []string, d
 			return HandleVErrAndExitCode(err, usage)
 		}
 
-		err2 := addBulkLoadingParadigms(dEnv, fPath)
+		dbName, err2 := getActiveDatabaseName(ctx, dEnv)
```
fulghum, line 170: It would be a little more idiomatic to reuse `err` instead of creating `err2`. If you change this var name to `err` it'll "shadow" the previous `err` variable – that's a pretty common pattern you'll see in the codebase.

fulghum (approved):
Looks good! Just two small suggestions.

fulghum (approved):
Nice work digging into those data dump docker test failures and getting them fixed up!

PR #5116, opened 2023-01-07
go/libraries/doltcore/remotestorage: chunk_store: Fix semantics of Rebase and Root. Root stays fixed until a Rebase call.

reltuk:
When remotestorage.DoltChunkStore was originally written, it forwarded all ChunkStore calls to the remotesapi. In reality, a client should expect a snapshot of the database on open, and the root should only change when Rebase() or Commit() is called. This fixes up that initial flaw. Rebase() becomes a refresh of the Root from the gRPC service, and Root() itself is just a read of the cached root hash.

This reduces unnecessary traffic to the remotestorage service and improves performance of things like sql-server read replicas.

### DIFF
--- a/go/libraries/doltcore/remotesrv/grpc.go
+++ b/go/libraries/doltcore/remotesrv/grpc.go
@@ -383,15 +383,6 @@ func (rs *RemoteChunkStore) Rebase(ctx context.Context, req *remotesapi.RebaseRe
 		return nil, status.Error(codes.Internal, "Could not get chunkstore")
 	}
 
-	logger.Printf("found %s", repoPath)
-
-	err = cs.Rebase(ctx)
-
-	if err != nil {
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/remotesrv/grpc.go
```diff
@@ -383,15 +383,6 @@ func (rs *RemoteChunkStore) Rebase(ctx context.Context, req *remotesapi.RebaseRe
 		return nil, status.Error(codes.Internal, "Could not get chunkstore")
 	}
 
-	logger.Printf("found %s", repoPath)
-
-	err = cs.Rebase(ctx)
```
andy-wm-arthur, line 388: ChunkStore state is centrally coordinated through the server process so we don't need to rebase `cs` here?
(also clients are using a getRoot call anyway)

andy-wm-arthur (approved):
LGTM

PR #5131, opened 2023-01-11
Add merge.GetForeignKeyViolatedTables

druvv:
Refactors `merge.AddForeignKeyViolations` by creating `merge.GetForeignKeyViolations` and the `FKViolationReceiver` interface.

Previously, our mechanic to check if edits to a root value created constraint violations was to write the constraint violations to the artifacts table and check if it was empty for all tables. 

Instead, we can now do a read-only operation to get constraint violations created by edits to a root value.


druvv:
cc: @tbantle22

andy-wm-arthur (approved):
LGTM

PR #5132, opened 2023-01-12
connect to remote branch not existent locally

jennifersp:
If the branch in the connection string is not available as local branch, try to find remote tracking branch that matches the defined branch name. If there is only one match, it creates local branch from that remote tracking branch and sets its upstream to it. If there are multiple remotes with the same branch names, it errors. For now, we assume there will not be remote name provided in the connection string, only `<db_name>/<branch_name>` is allowed.


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -1007,20 +1007,103 @@ func isBranch(ctx context.Context, db SqlDatabase, branchName string, dialer dbf
...
+			for _, rf := range remoteRefs {
+				if remRef, ok := rf.(ref.RemoteRef); ok && remRef.GetRemote() == remName && remRef.GetBranch() == brName {
+					err = createBranchOnDBAndSetUpstream(ctx, srcDB, ddb, brName, remRef.GetPath(), remRef)
+					if err != nil {
+						return "", false, err
+					}
+					return brName, true, nil
+				}
+			}
+		}
```
fulghum, line 1069: (minor) It's probably worth adding a `HasRemoteTrackingBranch` function to`DoltDB` to mirror the existing `HasBranch` function. That would probably make the `isRemoteBranch` function look really similar to the `isLocalBranch` function, and it `HasRemoteTrackingBranch` seems like a useful method to have on `DoltDB`.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -1007,20 +1007,103 @@ func isBranch(ctx context.Context, db SqlDatabase, branchName string, dialer dbf
...
+			return brName, true, nil
 		}
 	}
 
 	return "", false, nil
 }
 
+// isRemoteBranch is called only when the branch in connection string does not exist locally. It searches for a remote branch.
+// If there is a match, it creates a new local branch from the remote branch and sets its upstream to the remote branch.
+func isRemoteBranch(ctx context.Context, srcDB SqlDatabase, ddbs []*doltdb.DoltDB, branchName string) (string, bool, error) {
```
fulghum, line 1046: "only when the branch in connection string does not exist locally" probably isn't totally accurate, since I think this behavior would also kick in when someone does `use mydb/a-remote-tracking-branch`. 

Would be good to use the term "remote tracking branch" to disambiguate between a "remote branch" and "remote tracking branch". It's a small distinction, but I think an important one – "remote tracking branch" implies that we have a local tracking branch in our db metadata that tells us about the remote branch. 

From just the name of this function, I wasn't expecting it to create the local branch from the remote tracking branch (although, I do see you called that out in the function docs). I think this would be a little more intuitive to future readers if you either separated the existence check from the creation of a local branch from the remote tracking branch. Alternatively, you could also consider changing the name of this function to something like `createLocalBranchFromRemoteTrackingBranch` or something else nice and self-descriptive.

fulghum (commented):
Code and tests looks good! Just a couple small suggestions to take a look at. Let me know if anything doesn't make sense or if you wanna talk through anything in more detail.

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -715,6 +715,34 @@ func (ddb *DoltDB) HasBranch(ctx context.Context, branchName string) (string, bo
...
 
+// HasRemoteTrackingBranch returns whether the DB has a remote tracking branch with the name given, case-insensitive.
+// Returns the case-sensitive matching branch if found, as well as a bool indicating if there was a case-insensitive match,
+// remote tracking branchRef that is the only match for the branchName and any error.
+func (ddb *DoltDB) HasRemoteTrackingBranch(ctx context.Context, branchName string) (string, bool, ref.RemoteRef, error) {
+	remoteRefFound := false
+	var remoteRef ref.RemoteRef
+
+	remoteRefs, err := ddb.GetRemoteRefs(ctx)
+	if err == nil {
```
fulghum, line 726: Would be better to check if `err != nil` and return that `err` up, instead of swallowing it here.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -262,6 +262,9 @@ func (p DoltDatabaseProvider) attemptCloneReplica(ctx *sql.Context, dbName strin
 
 func (p DoltDatabaseProvider) HasDatabase(ctx *sql.Context, name string) bool {
 	_, err := p.Database(ctx, name)
+	if err != nil {
+		ctx.GetLogger().Errorf(err.Error())
+	}
```
fulghum, line 267: Very smart to log this error instead of letting it get swallowed! Good catch! 👏

fulghum (approved):
Nice work! 🙌  I was very happy to see the BATS test covering the case when there are multiple remote tracking branches with the same name.

PR #5146, opened 2023-01-13
go/libraries/doltcore/sqle: read_replica_database: Retry optimistic lock failures on read replica databases when updating our working set.

reltuk:


andy-wm-arthur (approved):
LGTM!

PR #5152, opened 2023-01-17
go/libraries/doltcore/sqle/read_replica_database.go: Optimize the sql-server read replica case when syncing a lot of new branches.

reltuk:
We already fetch all the new commits we need at once. Some lingering code had us refetching the commit values themselves from the upstream instead of directly setting the new branch to the commit hash which we just fetched.


zachmu (approved):
LGTM!

PR #5157, opened 2023-01-18
Fix `call DOLT_VERIFY_CONSTRAINTS()` false-positive

druvv:
Fix for bug reported by Fritter in Discord.

Here's what happened:
- Parent table has a simple pk
- Child table has a compound pk of length 3
- The foreign key is a single int column. It is the first column in the child table's pk, and the only column of parent's pk.
- No additional indexes are defined because we can reuse the pk index.

If we reuse the pk index, we have some special case code that builds a fake index definition. The issue is that we happened to pass sorted tags of the pks, instead the column-order tags. We then use the index definition to prefix iter secondary indexes.

This issue would not appear if you got lucky with the tags of your columns. Specifically, the sorted tag order had to match the tag order. For tables with one pk, this is always the case, and for the test we have for two pks, we also happen to get lucky. Unfortunately, Fritter has three column compound pk and he did not get lucky.

### DIFF
--- a/go/libraries/doltcore/merge/violations_fk.go
+++ b/go/libraries/doltcore/merge/violations_fk.go
@@ -661,7 +661,7 @@ func newConstraintViolationsLoadedTable(ctx context.Context, tblName, idxName st
 			IsUserDefined: false,
 			Comment:       "",
 		}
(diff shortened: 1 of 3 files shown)


Hydrocharged (approved):
LGTM! Great catch

PR #5165, opened 2023-01-19
Fixed branch -d behavior on unmerged branches to match git

### LINKED ISSUE #5139: Dolt incorrectly checks "fully merged" state when trying to safely delete branches
`dolt branch --delete` (without the `--force` option) first tries to check whether a branch is "fully merged" before deleting it, however the logic for defining "fully merged" does not match Git. From Git's [documentation](https://git-scm.com/docs/git-branch#Documentation/git-branch.txt--d):

```
-d
--delete
Delete a branch. The branch must be fully merged in its upstream branch, or in HEAD if no upstream was set with --track or --set-upstream-to.
```

In contrast, Dolt checks whether the branch deletion candidate is in the history of the default branch (typically `main`) by checking if the deletion candidate can be fast-forwarded to the default branch. Further, the error message reported:
`"attempted to delete a branch that is not fully merged into its parent; use `-f` to force"` gives the impression that the deletion candidate has diverged from some "parent" branch that does not exist.

We should match Git's behavior here and check the upstream branch, or the head of the currently checked-out branch if no upstream remotes exists.

zachmu:
Fixes #5139

### DIFF
--- a/go/libraries/doltcore/env/actions/branch.go
+++ b/go/libraries/doltcore/env/actions/branch.go
@@ -211,15 +211,15 @@ func validateBranchMergedIntoCurrentWorkingBranch(ctx context.Context, dbdata en
 		// TODO: no common ancestor is not an error
 		return err
 	}
-	
+
 	if !isMerged {
 		return fmt.Errorf(ErrUnmergedBranchDelete.Error(), branch.GetPath(), branch.GetPath())
 	}
-	
+
 	return nil
 }
 
-// validateBranchMergedIntoUpstream returns an error if the branch provided is not fully merged into its upstream	
+// validateBranchMergedIntoUpstream returns an error if the branch provided is not fully merged into its upstream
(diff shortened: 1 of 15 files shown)


go/libraries/doltcore/env/actions/branch.go
```diff
@@ -142,48 +142,132 @@ func DeleteBranchOnDB(ctx context.Context, dbData env.DbData, config *env.DoltCl
...
+		if errors.Is(err, doltdb.ErrUpToDate) {
+			return nil
 		}
-		if !isMerged {
-			return ErrUnmergedBranchDelete
+		if errors.Is(err, doltdb.ErrIsAhead) {
+			return ErrUnmergedBranch
 		}
+
+		// TODO: no common ancestor is not an error
```
andy-wm-arthur, line 211: I think we're safe here due to the init commit

andy-wm-arthur (approved):
LGTM

PR #5167, opened 2023-01-19
Removed branch control database file size limitation

Hydrocharged:
The resulting database file is fully compatible with older dolt clients since the read path skips the size and noms prefix. Switching the strings to `CreateShared` also gave a decent space savings as well.

To see if there was a noticeable performance difference, I tried removing everything except for the binlog, and reconstructing the branch control tables from the binlog, and the differences were negligible. The resulting file was definitely smaller, but performance-wise there wasn't much difference when reading or writing from the file. We spend so much time in the engine that it doesn't really matter. So I decided to go this route since we still retain full compatibility.


reltuk (approved):
LGTM!

nkonev:
Thank you!

PR #5169, opened 2023-01-19
Fix log table RowCount panic

max-hoffman:
`CommitClosure` methods are not nil-safe, check before calling `CommitClosure.Count()`.


fulghum (approved):
Looks great! 🚢

PR #5173, opened 2023-01-20
Fix 3 merge bugs related to secondary indexes

druvv:
- The schema of the dolt_constraint_violations table was incorrect.
- Unique key violations were being thrown by merge even if the index key included a null value.
- The primary key was being built incorrectly for the unique key violations generated by merge. Previously it was assumed that the primary index's key was a suffix of the secondary index's key. This is not the case when the index key includes a column in the primary key.


reltuk (approved):
LGTM!

PR #5176, opened 2023-01-20
float lexing and z encoding

JCOR11599:


go/libraries/doltcore/sqle/index/z_encoding.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/dolthub/go-mysql-server/sql/types"
+)
+
+// LexFloat maps the float64 into an uint64 representation in lexicographical order
+// For negative floats, we flip all the bits
+// For non-negative floats, we flip the signed bit
+func LexFloat(f float64) uint64 {
+	b := math.Float64bits(f)
+	if b>>63 == 1 {
```
andy-wm-arthur, line 30: maybe just this?
```suggestion
	if b < 0 {
```

andy-wm-arthur (approved):
LGTM!

go/libraries/doltcore/sqle/index/z_encoding.go
JCOR11599: `b` is a `uint64`, so it's never less than `0`.
It could be `f < 0.0`, but I wanted to check the sign bit.

PR #5180, opened 2023-01-20
go/libraries/doltcore/sqle/dsess: transactions.go: When merging working sets on call dolt_commit, ensure new commit also gets the merged commit root value.

reltuk:


zachmu (approved):
LGTM!

PR #5201, opened 2023-01-24
dolt pull without configs doesnt return error when up to date

stephkyou:
`dolt pull` should not error out when there is no user name or email set. However, when `dolt pull` is run in a workspace that is already up to date with remote, the CLI returns an error `up to date` instead of printing `Everything up-to-date.`. This change fixes `dolt pull` to not return an error when the workspace is already up to date.

fixes: https://github.com/dolthub/dolt/issues/5079

### DIFF
--- a/go/cmd/dolt/commands/pull.go
+++ b/go/cmd/dolt/commands/pull.go
@@ -17,11 +17,11 @@ package commands
 import (
 	"context"
 	"fmt"
-	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
 	eventsapi "github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi/v1alpha1"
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/commands/pull.go
```diff
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
```
fulghum, line 21: (minor) you can move this down to line 25 or so to keep all the `dolthub/dolt/go/libraries` import lines together.

integration-tests/bats/remotes.bats
```diff
@@ -1295,6 +1295,10 @@ SQL
     [ "$status" -eq 0 ]
     [[ "$output" =~ "test commit" ]] || false
 
+    # test pull with workspace up to date
+    run dolt pull
+    [ "$status" -eq 0 ]
```
fulghum, line 1300: probably worth a quick check out `$output` here, too, since the bug prevented us from showing the right error message.

fulghum (approved):
Fix looks good! Just two really minor suggestions. 🚢

PR #5205, opened 2023-01-24
go/libraries/doltcore/sqle/dsess: When commiting a transaction with a  DOLT_COMMIT, merge against the current HEAD if necessary.

reltuk:
Update the transaction logic to land a merge of current HEAD and the staged root value when we are creating a commit and the current HEAD has moved since the start of the transaction.

Dolt allows two concurrent transactions to both land commits on a branch HEAD. In the past, Dolt would simply commit the staged value on each transaction as the value of the database for the commit, and would merge the working sets so that the branch working set was updated with the latest data. This would result in histories with commits which seemed to delete previously committed data, only for it to still be present in the working set and to reappear in later commits. This was particularly problematic under remote replication, where the dolt commits themselves are replicated.

A change landed recently attempted to fix this by writing the merged working set value as the committed value. That works in the case of @@dolt_transaction_commit, assuming there are no manual calls to DOLT_COMMIT with manually staged changes, for example, but it has the wrong semantics when a user is manually staging and commiting changes.

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/session.go
+++ b/go/libraries/doltcore/sqle/dsess/session.go
@@ -570,16 +570,17 @@ func (d *DoltSession) NewPendingCommit(ctx *sql.Context, dbName string, roots do
 			mergeParentCommits = append(mergeParentCommits, parentCommit)
 		}
 
-		err = actions.ResetSoftToRef(ctx, sessionState.dbData, "HEAD~1")
+		root, err := actions.ResetSoftToRef(ctx, sessionState.dbData, "HEAD~1")
 		if err != nil {
 			return nil, err
 		}
+		roots.Head = root
 	}
 
 	pendingCommit, err := actions.GetCommitStaged(ctx, roots, sessionState.WorkingSet.MergeActive(), mergeParentCommits, sessionState.dbData.Ddb, props)
 	if err != nil {
 		if props.Amend {
-			err = actions.ResetSoftToRef(ctx, sessionState.dbData, headHash.String())
+			_, err = actions.ResetSoftToRef(ctx, sessionState.dbData, headHash.String())
 			if err != nil {
 				return nil, err
 			}
--- a/go/libraries/doltcore/sqle/dsess/session.go
+++ b/go/libraries/doltcore/sqle/dsess/session.go
@@ -570,6 +570,8 @@ func (d *DoltSession) NewPendingCommit(ctx *sql.Context, dbName string, roots do
 			mergeParentCommits = append(mergeParentCommits, parentCommit)
 		}
 
+		// TODO: This is not the correct way to write this commit as an amend. While this commit is running
+		// the branch head moves backwards and concurrency control here is not principled.
 		root, err := actions.ResetSoftToRef(ctx, sessionState.dbData, "HEAD~1")
 		if err != nil {
 			return nil, err
--- a/go/libraries/doltcore/sqle/dsess/transactions.go
+++ b/go/libraries/doltcore/sqle/dsess/transactions.go
@@ -156,6 +156,7 @@ func doltCommit(ctx *sql.Context,
 		return nil, nil, err
 	}
 
+	// We check if the branch HEAD has changed since our transaction started.
 	if curHead != nil {
 		curRootVal, err := curHead.ResolveRootValue(ctx)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


reltuk:
#benchmark

reltuk:
#benchmark

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -570,16 +570,17 @@ func (d *DoltSession) NewPendingCommit(ctx *sql.Context, dbName string, roots do
 			mergeParentCommits = append(mergeParentCommits, parentCommit)
 		}
 
-		err = actions.ResetSoftToRef(ctx, sessionState.dbData, "HEAD~1")
+		root, err := actions.ResetSoftToRef(ctx, sessionState.dbData, "HEAD~1")
```
zachmu, line 573: It would be better if we didn't do this... the problem is that the action calls repoStateWriter.UpdateStagedRoot, which in turn calls session.UpdateWorkingSet, which 1) has its own update semantics / lifecycle that interact with the work being done here, and 2) are inappropriate in the context of what *should* be a purely functional constructor

A better solution would be for the action to return a doiltdb.Roots struct, which the caller can decide how to apply to update state as necessary. Since this is just on the --amend path this isn't critical, maybe add a TODO

go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -143,13 +143,55 @@ func doltCommit(ctx *sql.Context,
...
+			return nil, nil, err
+		}
+		headRootValHash, err := pending.Roots.Head.HashOf()
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if curRootValHash != headRootValHash {
+			start := time.Now()
+			mo := merge.MergeOpts{IsCherryPick: false}
```
zachmu, line 175: Surely this is `merge.MergeOpts{}` (or `merge.DefaultOpts` if that exists)?

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -200,6 +200,22 @@ func TestDoltTransactionCommitTwoClients(t *testing.T) {
 				Query:    "/* client b */ START TRANSACTION;",
 				Expected: []sql.Row{},
 			},
+			// Concurrent with the two transactions which are going to (dolt_)commit changes, we
```
zachmu, line 203: This could use at least one equivalent test with manual `call dolt_commit` in `dolt_transaction_queries.go`

go/libraries/doltcore/sqle/enginetest/dolt_transaction_commit_test.go
```diff
@@ -262,22 +278,26 @@ func TestDoltTransactionCommitTwoClients(t *testing.T) {
...
 			},
 			{
 				Query:    "/* client c */ SELECT * FROM x ORDER BY y;",
+				Expected: []sql.Row{{1, 1}, {2, 2}, {3, 3}, {4, 4}},
+			},
+			{
+				Query:    "/* client c */ SELECT * FROM x AS OF 'HEAD' ORDER BY y;",
 				Expected: []sql.Row{{1, 1}, {2, 2}, {3, 3}},
 			},
 			// After we commit both of these transactions, our working set should not have any pending changes.
```
zachmu, line 295: Comment no longer accurate

go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -143,13 +143,55 @@ func doltCommit(ctx *sql.Context,
...
+			pending.Roots.Staged, _, err = merge.MergeRoots(
```
zachmu, line 176: It bothers me that we're doing two merge operations here, one for the working set and one for the HEAD. It seems like it should be possible to unify the logic here and do a single merge to get our new root value, then set the root values in the pending commit as necessary.  Two merges is harder to reason about and is also going to be slower.

Am I missing something obvious about this? Isn't it the case that the working set and the HEAD will contain the same value after a dolt_commit operation? If so, why do the work to calculate that single value more than once?

zachmu (approved):
LGTM

I have reservations about this approach, but as far as I can reason about, it should be correct which is better than what's in our current release

PR #5206, opened 2023-01-25
adding new codec for spatial indexes

JCOR11599:


andy-wm-arthur (approved):

go/libraries/doltcore/sqle/index/prolly_fields.go
```diff
@@ -118,6 +118,8 @@ func GetField(ctx context.Context, td val.TupleDesc, i int, tup val.Tuple, ns tr
 		}
 	case val.CommitAddrEnc:
 		v, ok = td.GetCommitAddr(i, tup)
+	case val.CellEnc:
+		v, ok = td.GetCell(i, tup)
```
andy-wm-arthur, line 122: We should throw an error here. We never want to be returning `Cell`s up to the SQL layer, they are only meaning full at the storage/indexing layer

andy-wm-arthur (approved):
Looks Great! thanks for iterating on these interfaces

PR #5207, opened 2023-01-25
go/libraries/doltcore/remotesrv: Response with HTTP status code 206 when writing partial responses.

reltuk:
Also add Accept-Ranges headers in all HTTP responses for portions of table files and Content-Range headers for partial file responses.

This makes the HTTP server more correct and can improve compatibility with various HTTP middleware.


coffeegoddd (approved):
LGTM!

PR #5208, opened 2023-01-25
go/store/nbs: add concurrency to blobstore persister operations

andy-wm-arthur:


reltuk (approved):
LGTM!

Might want to use https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit on the checkAllTablesExist implementation, but maybe expected concurrency there isn't too outrageous and downstream the concurrency will already be limited on the contested resources by things like connection limits or outstanding request windows...

PR #5223, opened 2023-01-26
Added `dolt_show_branch_databases` system variable

zachmu:
When set to 1, will cause a revision database for each branch to appear in `show databases`, information schema tables, etc.

### DIFF
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -290,14 +290,14 @@ func (p DoltDatabaseProvider) AllDatabases(ctx *sql.Context) (all []sql.Database
 				continue
 			}
 			all = append(all, revisionDbs...)
-			
-			// if one of the revisions we just expanded matches the curr db, mark it so we don't double-include that 
+
+			// if one of the revisions we just expanded matches the curr db, mark it so we don't double-include that
 			// revision db
 			if !foundDatabase && currDb != "" {
 				for _, revisionDb := range revisionDbs {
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -279,6 +281,15 @@ func (p DoltDatabaseProvider) AllDatabases(ctx *sql.Context) (all []sql.Database
 			foundDatabase = true
 		}
 		all = append(all, db)
+
+		if showBranches {
+			revisionDbs, err := p.allRevisionDbs(ctx, db)
+			if err != nil {
+				// TODO: this interface is wrong, needs to return errors
```
fulghum, line 288: Probably still worth logging a warning or error even if we can't return it here.

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -279,6 +281,15 @@ func (p DoltDatabaseProvider) AllDatabases(ctx *sql.Context) (all []sql.Database
 			foundDatabase = true
```
fulghum, line 281: You might want to set `foundDatabase` in the `if showBranches` loop, too; otherwise, I think you get the revision db showing up in the list twice if you have this var set and you have `use`d a revision db in the current session.

fulghum (approved):
Looks great!

go/libraries/doltcore/sqle/database_provider.go
zachmu: Very good catch, this was double-included revision dbs

PR #5225, opened 2023-01-26
add bats tests for information schema tables

jennifersp:
Added bats tests for some supported information_schema tables for specific column values that must match to MySQL information_schema result. Some tables and columns are not covered here because some has constant information such as `keywords` table, which is tested in `GMS` engine tests, and some columns are `created_time` or `sql_mode` which always varies.


fulghum (approved):
Awesome to have the `information_schema` tables tested as part of the BATS tests. Looks great! 🚢

PR #5228, opened 2023-01-26
Quick fix for unnecessary error being logged

fulghum:
We started logging an error instead of swallowing it, which is good, but we need to check if it's the `ErrDatabaseNotFound` error we know we might see, and not log an error if it's that expected case.


jennifersp (approved):
Ah right! Looks good! Thank you for catching this 😄

PR #5233, opened 2023-01-27
add `dolt_cherry_pick` procedure

jennifersp:
Add `DOLT_CHERRY_PICK()` and `DCHERRY_PICK()` procedures for SQL interface
Removed author arg parsing from CLI cherry-pick since this arg's is not available for current implementation (can be supported).


integration-tests/bats/sql-cherry-pick.bats
```diff
@@ -0,0 +1,423 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
+
+setup() {
+    setup_common
+
+dolt sql -q "CREATE TABLE test(pk BIGINT PRIMARY KEY, v varchar(10))"
```
fulghum, line 7: (minor/whitespace/indenting)

```suggestion
    dolt sql -q "CREATE TABLE test(pk BIGINT PRIMARY KEY, v varchar(10))"
```

go/libraries/doltcore/sqle/dprocedures/dolt_cherry_pick.go
```diff
@@ -0,0 +1,208 @@
...
+	dbData, ok := dSess.GetDbData(ctx, dbName)
+	if !ok {
+		return nil, "", fmt.Errorf("failed to get dbData")
+	}
+
+	cherryCommitSpec, err := doltdb.NewCommitSpec(cherryStr)
+	if err != nil {
+		return nil, "", err
+	}
+	cherryCommit, err := doltDB.Resolve(context.TODO(), cherryCommitSpec, dbData.Rsr.CWBHeadRef())
```
fulghum, line 140: Is there a reason to not pass `ctx` in as the context to this call?

go/libraries/doltcore/sqle/dprocedures/dolt_cherry_pick.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil, "", fmt.Errorf("conflicts in table {'%s'}", tblNames)
+	}
+
+	workingRootHash, err = mergedRoot.HashOf()
+	if err != nil {
+		return nil, "", err
+	}
+
+	if headRootHash.Equal(workingRootHash) {
+		// TODO: should it be warning? or nothing?
```
fulghum, line 198: What does Git do in this case? Specifically... does it return a 0 exit code to indicate success? Does this only happen if the cherry-picked commit is an empty commit? 

The user requested to cherry-pick, so if it was applied successfully (even if nothing changed), then it seems like it's okay to return a success/0 exit code. Let me know if I'm just missing something about when this block would trigger.

fulghum (approved):
Code looks good and tests looked super thorough; nice work on those! Just some really minor suggestions from me. 

It would be interesting to see if we can combine more of the CLI cherry-pick and SQL cherry-pick code, but we can follow up with that later.

go/libraries/doltcore/sqle/dprocedures/dolt_cherry_pick.go
jennifersp: We return exit code 0 as nothing failed but just a message is printed saying nothing to commit in the CLI command. I was not sure whether to return 0 for success and return empty hash or the current head hash?

fulghum: Actually... checking out Git's behavior... if I try to cherrypick an empty commit or the current HEAD commit, I get a `1` exit code indicating the commit tree was not changed and a message about `nothing to commit, working tree clean`. What do you think about matching that behavior here?

jennifersp: sounds good!

PR #5234, opened 2023-01-27
add `--author` arg to dolt tag command on both CLI and SQL

jennifersp:


fulghum (approved):
Looks good! 🚢 

Nice refactoring to share the command parser between the CLI and SQL code, too.

PR #5240, opened 2023-01-27
go/store/nbs: Update table-file-exists sanity check to not check for files which are already in our view of upstream.

reltuk:


andy-wm-arthur (approved):
LGTM!

PR #5242, opened 2023-01-27
output correct result for empty JSON

jennifersp:


fulghum (approved):
Nice work! Thanks for knocking this one out so quickly! 🙌

PR #5244, opened 2023-01-27
go/store/nbs: upload table parts on push for GCS backed NBS

andy-wm-arthur:


go/store/nbs/bs_persister.go
```diff
@@ -176,22 +177,55 @@ func (bsp *blobstorePersister) Path() string {
...
 }
 
-func (bsp *blobstorePersister) CopyTableFile(ctx context.Context, r io.ReadCloser, fileId string, chunkCount uint32) error {
-	var err error
-
+func (bsp *blobstorePersister) CopyTableFile(ctx context.Context, r io.ReadCloser, name string, fileSz uint64, chunkCount uint32) (err error) {
 	defer func() {
-		cerr := r.Close()
-		if err == nil {
+		if cerr := r.Close(); cerr != nil {
```
zachmu, line 182: I think this was correct before, you don't want to overwrite a previous error with the close error even if one occurs

zachmu (approved):
LGTM!

PR #5251, opened 2023-01-28
go/store/nbs: blobstore skip concatenate

andy-wm-arthur:


go/store/datas/pull/clone.go
```diff
@@ -209,7 +209,9 @@ func clone(ctx context.Context, srcTS, sinkTS nbs.TableFileStore, eventCh chan<-
 		}
 	}
 
-	sinkTS.AddTableFilesToManifest(ctx, fileIDToNumChunks)
+	if err = sinkTS.AddTableFilesToManifest(ctx, fileIDToNumChunks); err != nil {
```
zachmu, line 212: Love that this was assumed to not return an error before this change, we are great at software

zachmu (approved):
LGTM

PR #5252, opened 2023-01-28
Fix int overflow on 32-bit architectures

Arnie97:
```py
$  GOARCH=386 GOOS=linux go build ./...
# github.com/dolthub/dolt/go/store/skip
store/skip/list.go:152:21: maxCount (untyped int constant 4294967294) overflows int
```
See also https://github.com/dolthub/vitess/pull/216

### DIFF
--- a/go/store/skip/list.go
+++ b/go/store/skip/list.go
@@ -21,7 +21,7 @@ import (
 
 const (
 	maxHeight  = 9
-	maxCount   = math.MaxUint32 - 1
+	maxCount   = math.MaxInt - 1
 	sentinelId = nodeId(0)
 	initSize   = 8
 )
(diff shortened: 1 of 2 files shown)


timsehn:
Thanks @reltuk will look at this one.

go/store/skip/list.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	maxHeight  = 9
-	maxCount   = math.MaxUint32 - 1
+	maxCount   = math.MaxInt - 1
```
reltuk, line 24: I think this might be better as:

```suggestion
	maxCount   = math.MaxInt32 - 1
```

for now. The structure will not work with a number of entries > `MaxUint32`, so we don't want to set it to `MaxInt - 1` on a 64-bit platform.

For now we can give up the two billion possible extra entries on 64-bit in order to get this working on 32-bit.

reltuk (commented):
Thank you for the report! Generally looks great to me. One suggested change.

PR #5255, opened 2023-01-30
interface changes for spatial indexes

JCOR11599:
`di.spatial` is never assigned anywhere, so it should remain false.
This is correct for now, since we don't support Spatial Index on dolt yet, just in GMS.

also add a feature flag if you want the query to succeed


JCOR11599:
These changes are in: https://github.com/dolthub/dolt/pull/5255

PR #5256, opened 2023-01-30
go/store/blobstore: gcs.go: Make Get() make one round trip instead of reading the metadata in a separate request.

reltuk:


zachmu (approved):
LGTM

PR #5257, opened 2023-01-30
add bats test for release all locks on session end

jennifersp:
Depends on https://github.com/dolthub/go-mysql-server/pull/1569


fulghum (approved):
Fast work on this one! Good job! ⚡

PR #5267, opened 2023-01-31
go/libraries/doltcore/sqle/dprocedures: Do not register dolt_gc.

reltuk:


zachmu (approved):
LGTM

nkonev:
Why ? Do you have plans to restore `dolt_gc` procedure after fixing if any ?

timsehn:
Yes. We will restore it. The current version failed under concurrency so it was unsafe to have in Prod until the bugs are fixed. In the meantime, stop the server and run `dolt gc` if you need it.

See:

https://github.com/dolthub/dolt/pull/5274
https://github.com/dolthub/dolt/pull/5273
https://github.com/dolthub/dolt/pull/5270
https://github.com/dolthub/dolt/pull/5269

PR #5269, opened 2023-01-31
go/store/nbs: store.go: Fix some lock ordering issues with GC and mm.LockByName. Change swapTables to not drop tables.novel or nbs.mt.

reltuk:


andy-wm-arthur (approved):
LGTM

PR #5270, opened 2023-01-31
go/store/types: value_store.go: Change GC implementation to call TableFileStore.PruneTableFiles after the copy is complete.

reltuk:


andy-wm-arthur (approved):
LGTM

PR #5273, opened 2023-02-01
go/store/types: ValueStore: Fix deadlocks for concurrent writes, bufferMu and NBSStore.

reltuk:
At times ValueStore calls into ChunkStore.PutChunk, for example, with bufferMu held. With online-GC, that call can block until a GC process is finished. So a calling goroutine would block with bufferMu held. However, in order for the GC process to finish, ValueStore.GC needs to grab bufferMu in order to finalize its bookkeeping state after ChunkStore GC is successful.

This change allows reads to proceed against ValueStore while it waits for GC to complete by adding a condition variable which we toggle before we start GC and broadcast on after we're done.

### DIFF
--- a/go/store/types/value_store.go
+++ b/go/store/types/value_store.go
@@ -85,6 +85,9 @@ type ValueStore struct {
 	decodedChunks        *sizecache.SizeCache
 	nbf                  *NomsBinFormat
 
+	gcCond  *sync.Cond
+	doingGC bool
+
 	versOnce sync.Once
 }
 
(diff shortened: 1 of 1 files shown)


andy-wm-arthur (approved):
Looks Good 

This makes me want to simplify the GC interaction between ValueStore and NBS. The `ChunkStoreGarbageCollector` interface was changed to take a `WalkAddrs` callback, we might we able to move all this state management down into NBS. Old vs new gen refs adds a wrinkle tho

PR #5274, opened 2023-02-01
go/store/nbs: store.go: Change memtable reference sanity check to happen at table_set.append() instead of trying to check novel chunk sources.

reltuk:
Removes enforce completeness functionality from ValueStore, which becomes superfluous.


go/store/nbs/store.go
```diff
@@ -978,68 +973,55 @@ func (nbs *NomsBlockStore) commit(ctx context.Context, current, last hash.Hash,
...
 	}
 
+	defer nbs.mu.Unlock()
+
 	// check for dangling references in |nbs.mt|
-	nbs.mu.Lock()
-	if err = nbs.errorIfDangling(checker); err != nil {
+	if err = nbs.errorIfDangling(current, checker); err != nil {
+		if errors.Is(err, ErrDanglingRef) {
+			nbs.mt = nil
```
andy-wm-arthur, line 995: should we clear the memtable in `errorIfDangling`? is there ever a path where we don't want to clear it?

andy-wm-arthur (approved):
LGTM

PR #5275, opened 2023-02-01
verify fk with secondary index even when empty

JCOR11599:
The optimization for verifying foreign keys using secondary indexes isn't used when the tables are completely different (so only additions to the merge) or if the index potentially didn't exist before merging. Rather than defaulting to using the primary key to verify the foreign keys, it's possible to just create an empty index and use the faster secondary index to verify.

This includes code from https://github.com/dolthub/dolt/pull/5253

### DIFF
--- a/go/libraries/doltcore/merge/violations_fk.go
+++ b/go/libraries/doltcore/merge/violations_fk.go
@@ -373,26 +373,29 @@ func childFkConstraintViolations(
 	preChildRowData durable.Index,
 	receiver FKViolationReceiver,
 ) error {
-	if preChildRowData.Format() == types.Format_DOLT {
-		if empty, err := preChildRowData.Empty(); err != nil {
+	if preChildRowData.Format() != types.Format_DOLT {
+		m := durable.NomsMapFromIndex(preChildRowData)
+		return nomsChildFkConstraintViolations(ctx, foreignKey, postParent, postChild, preChild.Schema, m, receiver)
+	}
+	if preChild.IndexData != nil && postChild.Schema.GetPKCols().Size() > 0 && preChild.Schema.GetPKCols().Size() > 0 {
+		empty, err := preChildRowData.Empty()
+		if err != nil {
 			return err
-		} else if empty {
-			emptyIdx, err := durable.NewEmptyIndex(ctx, postChild.Table.ValueReadWriter(), postChild.Table.NodeStore(), postChild.Schema)
+		}
+		var idx durable.Index
+		if empty {
+			idx, err = durable.NewEmptyIndex(ctx, postChild.Table.ValueReadWriter(), postChild.Table.NodeStore(), postChild.Schema)
 			if err != nil {
 				return err
 			}
-			m := durable.ProllyMapFromIndex(emptyIdx)
-			return prollyChildSecDiffFkConstraintViolations(ctx, foreignKey, postParent, postChild, m, receiver)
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/merge/violations_fk.go
```diff
@@ -369,17 +369,30 @@ func parentFkConstraintViolations(
...
+	postParent, postChild, preChild *constraintViolationsLoadedTable,
 	preChildRowData durable.Index,
-	receiver FKViolationReceiver) error {
+	receiver FKViolationReceiver,
+) error {
 	if preChildRowData.Format() == types.Format_DOLT {
+		if empty, err := preChildRowData.Empty(); err != nil {
+			return err
+		} else if empty {
+			emptyIdx, err := durable.NewEmptyIndex(ctx, postChild.Table.ValueReadWriter(), postChild.Table.NodeStore(), postChild.Schema)
```
max-hoffman, line 380: it'd be a bit nicer organizationally if it is possible to condense the empty setting in the above call next to the empty primary index check , something like preChild.IndexData = NewEmptyIndex. then you could just go into the secondary diff here

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -109,16 +109,25 @@ func prollyChildFkConstraintViolations(
 	partialDesc := idxDesc.PrefixDesc(len(foreignKey.TableColumns))
 	partialKB := val.NewTupleBuilder(partialDesc)
 
+	//panic(fmt.Sprintf("%s, %d, " ,foreignKey.Name, postChild.Schema.GetPKCols().Size()))
```
max-hoffman, line 112: hanging comment

max-hoffman (approved):
lgtm small organizational QOL things, tests pass and these are all improvements so we should merge soon

go/libraries/doltcore/merge/violations_fk.go
```diff
@@ -133,13 +133,13 @@ func GetForeignKeyViolations(ctx context.Context, newRoot, baseRoot *doltdb.Root
 			if err != nil {
 				return err
 			}
-			err = childFkConstraintViolations(ctx, foreignKey, postParent, postChild, postChild.Schema, emptyIdx, receiver)
+
+			err = childFkConstraintViolations(ctx, foreignKey, postParent, postChild, postChild, emptyIdx, receiver)
```
max-hoffman, line 137:
```suggestion
			preChild.IndexData, err = durable.NewEmptyIndex(ctx, postChild.Table.ValueReadWriter(), postChild.Table.NodeStore(), postChild.Schema)
			if err != nil {
				return err
			}
			
			err = childFkConstraintViolations(ctx, foreignKey, postParent, postChild, postChild, emptyIdx, receiver)
```

go/libraries/doltcore/merge/violations_fk.go
```diff
@@ -369,17 +369,30 @@ func parentFkConstraintViolations(
...
+		empty, err := preChildRowData.Empty()
+		if err != nil {
+			return err
+		}
+		if empty {
+			preChild.IndexData, err = durable.NewEmptyIndex(ctx, postChild.Table.ValueReadWriter(), postChild.Table.NodeStore(), postChild.Schema)
+			if err != nil {
+				return err
+			}
+		}
```
max-hoffman, line 390: can delete if pushed into parent call

max-hoffman (approved):
LGTM ship once we get one with passing tests. I appreciate the help on this!

PR #5279, opened 2023-02-01
reuse cursor when verifying fks using secondary indexes

JCOR11599:


go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -146,39 +146,55 @@ func prollyChildSecDiffFkConstraintViolations(
...
-				postChild.IndexSchema,
-				k,
-				v,
-				preChildScndryIdx.Pool())
-			if hasNulls {
-				return nil
+			k := val.Tuple(diff.Key)
+			// TODO: possible to skip this if there are not null constraints over entire index
+			for i := 0; i < k.Count(); i++ {
+				if k.FieldIsNull(i) {
```
max-hoffman, line 167: this one confuses me a bit. if a child fk has `(nil,1)` and the parent only has `(1,1)` there is no a constraint violation?

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -254,6 +257,7 @@ func createCVIfNoPartialKeyMatchesSec(
 	}
 
 	return receiver.ProllyFKViolationFound(ctx, primaryIdxKey, value)
+
```
max-hoffman, line 260: nit
```suggestion
```

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -146,39 +146,55 @@ func prollyChildSecDiffFkConstraintViolations(
...
 			if err != nil {
 				return err
 			}
+			if !parentSecIdxCur.Valid() {
+				return createNewCVForSecIdx(ctx, k, childPriKD, childPriKB, postChildRowData, postChildRowData.Pool(), receiver)
+			}
+
+			rng := prolly.PrefixRange(k, partialDesc)
+			key := val.Tuple(parentSecIdxCur.CurrentKey())
+			if !rng.Matches(key) {
```
max-hoffman, line 193: this one is over my head. maybe add a comment here to indicate which cases require the extra range match, otherwise we miss constraint violations?

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -329,7 +333,7 @@ func createCVsForPartialKeyMatches(
 }
 
 func makePartialKey(kb *val.TupleBuilder, tags []uint64, idxSch schema.Index, tblSch schema.Schema, k, v val.Tuple, pool pool.BuffPool) (val.Tuple, bool) {
-	if idxSch.Name() != "" {
+	if idxSch.Name() != "" && len(idxSch.IndexedColumnTags()) <= len(tags) {
```
max-hoffman, line 336: this one could use a comment too, an example of how we'd hit this condition.

max-hoffman (approved):
LGTM, comments that explain some of the edge cases and/or tests that call out the edge cases would be helpful to make this code more understandable

go/libraries/doltcore/merge/violations_fk_prolly.go
JCOR11599: Yep, here's a BATS test with this behavior:
https://github.com/dolthub/dolt/blob/main/integration-tests/bats/verify-constraints.bats#L100

I also tried this in MySQL and it looks like this is what they expect
```sql
mysql> create table parent (i int primary key, x int, y int, index (x,y));
Query OK, 0 rows affected (0.0246 sec)
mysql> create table child (j int primary key, x int, y int, foreign key (x,y) references parent(x,y));
Query OK, 0 rows affected (0.0207 sec)
mysql> insert into parent values (0,1,1);
Query OK, 1 row affected (0.0047 sec)
mysql> insert into child values (0,1,1);
Query OK, 1 row affected (0.0098 sec)
mysql> insert into child values (1,999,null);
Query OK, 1 row affected (0.0020 sec)
```


Some docs fwiw: https://dev.mysql.com/doc/mysql-reslimits-excerpt/8.0/en/ansi-diff-foreign-keys.html

max-hoffman: 👍 nice research. could we still put that in an enginetest so it surfaces quickly in testing?

go/libraries/doltcore/merge/violations_fk_prolly.go
JCOR11599: The range check is still necessary here because it's possible that the key in question `k` is smaller than every key in `parentSecIdxCur`, so the cursor is still valid.

go/libraries/doltcore/merge/violations_fk_prolly.go
JCOR11599: We hit this condition when the parent's index `idxSch` is longer than the partial key we have `k` with length `len(tags)`, so we panic trying to build a key that's too long.

go/libraries/doltcore/merge/violations_fk_prolly.go
max-hoffman: would it be equivalent to do something like `paretnSecIdx.Tuples().Order.Compare(k, parentSecIdxCur.CurrentKey())`? might be more self-documenting

PR #5286, opened 2023-02-02
allow different format options for dolt branch with track flag

jennifersp:
Git allows:
`git branch feature --track origin/feature`
`git branch --track request origin/feature`
`git branch feature --track=direct origin/feature`
`git branch --track=direct feature origin/feature`
for creating new branch that tracks given remote tracking branch

does not allow:
`git branch feature --track direct origin/feature`, but Dolt allows it as `--track direct` and `--track=direct` are considered the same

### DIFF
--- a/integration-tests/bats/remotes.bats
+++ b/integration-tests/bats/remotes.bats
@@ -2075,7 +2075,8 @@ SQL
     [ "$status" -eq 0 ]
     [[ "$output" =~ "Everything up-to-date." ]] || false
 
-    run dolt branch --track direct feature origin/other
+    # TODO: this command fails on GitHub requiring `--track=direct`, when both branch name and starting point name are defined, but Dolt allows both formats.
+    run dolt branch feature --track direct origin/other
     [ "$status" -eq 0 ]
     [[ "$output" =~ "branch 'feature' set up to track 'origin/other'" ]] || false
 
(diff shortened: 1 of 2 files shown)


integration-tests/bats/remotes.bats
```diff
@@ -2075,7 +2075,8 @@ SQL
     [ "$status" -eq 0 ]
     [[ "$output" =~ "Everything up-to-date." ]] || false
 
-    run dolt branch --track direct feature origin/other
+    # TODO: this command fails on GitHub requiring `--track=direct`, when both branch name and starting point name are defined, but Dolt allows both formats.
```
fulghum, line 2078: (minor) You might change this to `NOTE:` instead of `TODO:` – I think this is such a small difference that there's no action we need to take, so it didn't seem like a `TODO` to me. 

When I see "GitHub", I think about the GitHub website, but I think here you actually just mean "with git"?

integration-tests/bats/remotes.bats
```diff
@@ -2075,7 +2075,8 @@ SQL
...
+    run dolt branch feature --track direct origin/other
```
fulghum, line 2079: It would be good to make sure we have test coverage over all the forms you listed in the PR description. 

I saw these two: 
- git branch --track request origin/feature
- git branch feature --track=direct origin/feature

But I don't think I saw these covered: 
- git branch feature --track origin/feature
- git branch --track=direct feature origin/feature

fulghum (commented):
Code change looks fine, just a couple of notes to make sure we have enough test coverage. 

Could you also take a look at the SQL path for `dolt_branch` and see if we need to update anything there to have the same functionality and test coverage?

fulghum (approved):
Nice! Assuming all those CI tests pass, this looks good to go! 🙌

PR #5291, opened 2023-02-02
update query timing to be accurate when using dolt sql-client

stephkyou:
Query timings have not been accurate in `dolt sql-client`, printing out the time to print the result instead of the time to execute the query. This change fixes the timing to be accurate to the time to execute the query.

related: https://github.com/dolthub/go-mysql-server/pull/1581
fixes: https://github.com/dolthub/dolt/issues/5241


integration-tests/bats/sql-client.bats
```diff
@@ -147,3 +147,15 @@ teardown() {
     [ $status -eq 0 ]
     [[ $output =~ " test_dashes " ]] || false
 }
+
+@test "sql-client: prints accurate query timing" {
+    skiponwindows "Missing dependencies"
```
fulghum, line 152: Are you sure this one won't run on windows? It looks pretty similar to other tests in this file that don't seem to be skipped on windows. Worth trying out if you haven't tested it yet.

fulghum (approved):
Great job on this one Steph! 🙌

PR #5294, opened 2023-02-02
add bats test for binary string encoded to utf8

jennifersp:
Depends on https://github.com/dolthub/go-mysql-server/pull/1583
DataGrip sets the `character_set_results` system variable to `utf8`, which raw binary strings cannot be encoded into. MySQL returns binary type strings in hex format.

### DIFF
--- a/integration-tests/bats/sql-server.bats
+++ b/integration-tests/bats/sql-server.bats
@@ -1873,3 +1873,13 @@ s.close()
     [ $status -eq 0 ]
     [[ "$output" =~ "$EXPECTED" ]] || false
 }
+
+@test "sql-server: binary literal is printed as hex string for utf8 charset result set" {
+    cd repo1
+    start_sql_server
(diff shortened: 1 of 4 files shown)


integration-tests/bats/sql-server.bats
```diff
@@ -1873,3 +1873,13 @@ s.close()
     [ $status -eq 0 ]
     [[ "$output" =~ "$EXPECTED" ]] || false
 }
+
+@test "sql-server: binary literal is printed as hex string for utf8 charset result set" {
+    cd repo1
+    start_sql_server
```
timsehn, line 1879: Does this need to be in server mode? Could the same thing be accomplished via `dolt sql`?

timsehn (commented):
Not sure this needs to be in server mode.

fulghum (approved):
Test looks good. Tim has a good comment about whether we could just use `dolt sql` instead of going through a sql-server. From what we just discussed, it seems like it's true that `dolt sql` should be enough to trigger the issue? Could you test that out and confirm?

integration-tests/bats/sql-server.bats
jennifersp: The error cannot be produced via `dolt sql`, so keeping it as sql-server test.

PR #5295, opened 2023-02-02
Trim compare string prefix

max-hoffman:
this improves a low-selectivity index scan for a customer query.


Hydrocharged (approved):
I like the change! This looks to be based on Aaron's suggestion for how we could improve comp speeds. The implementation is simpler than I expected.

PR #5302, opened 2023-02-03
fix panic for unexpected json file format

jennifersp:
returns error message instead of panic

### DIFF
--- a/go/cmd/dolt/commands/schcmds/import.go
+++ b/go/cmd/dolt/commands/schcmds/import.go
@@ -58,9 +58,13 @@ const (
 
 var MappingFileHelp = "A mapping file is json in the format:" + `
 
-	{
-		"source_field_name":"dest_field_name"
-		...
+	{ "rows": 
+		[
+			{
+				"source_field_name":"dest_field_name"
+				...
+			}, ...
+		]
 	}
 
 where source_field_name is the name of a field in the file being imported and dest_field_name is the name of a field in the table being imported to.
--- a/go/cmd/dolt/commands/schcmds/import.go
+++ b/go/cmd/dolt/commands/schcmds/import.go
@@ -56,15 +56,11 @@ const (
 	delimParam          = "delim"
 )
 
-var MappingFileHelp = "A mapping file is json in the format:" + `
+var mappingFileHelp = "A mapping file is json in the format:" + `
 
-	{ "rows": 
-		[
-			{
-				"source_field_name":"dest_field_name"
-				...
-			}, ...
-		]
+	{
+		"source_field_name":"dest_field_name"
+		...
 	}
 
 where source_field_name is the name of a field in the file being imported and dest_field_name is the name of a field in the table being imported to.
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/table/typed/json/reader.go
```diff
@@ -111,7 +111,12 @@ func (r *JSONReader) ReadSqlRow(ctx context.Context) (sql.Row, error) {
 		return nil, io.EOF
 	}
 
-	return r.convToSqlRow(metaRow.Value.(map[string]interface{}))
+	mapVal, ok := metaRow.Value.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("unexpected JSON format received, expected format: { \"rows\": [ json_row_objects... ] } ")
```
fulghum, line 116: Good error message! 👏 This should be helpful for customers to understand the input format. 

Could you also add some more docs to the `dolt table import` command so that this expected input format is described there, too? I didn't see anything in the docs yet describing what we expect for a JSON input file. Hopefully putting a note about that format in the docs will help customers more easily understand how to use it.

fulghum (commented):
Fix looks great! I just had one note to update the `dolt table import` docs, too, so that the JSON format we require is documented there, too, in addition to in this new error message.

go/cmd/dolt/commands/schcmds/import.go
```diff
@@ -58,9 +58,13 @@ const (
...
-	{
-		"source_field_name":"dest_field_name"
-		...
+	{ "rows": 
+		[
+			{
+				"source_field_name":"dest_field_name"
+				...
+			}, ...
+		]
```
fulghum, line 67: This is good content, but I think it's in the wrong section... the mapping file maps names between the input file and the destination schema to rename items, but we want to add some description of the input file.

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -68,6 +68,20 @@ const (
 	disableFkChecks   = "disable-fk-checks"
 )
 
+var jsonInputFileHelp = "A mapping file is json in the format:" + `
```
fulghum, line 71: Close, but this text should be something like "when using a JSON input file, the expected format is:"

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -84,7 +98,7 @@ If the schema for the existing table does not match the schema for the new file,
 
 A mapping file can be used to map fields between the file being imported and the table being written to. This can be used when creating a new table, or updating or replacing an existing table.
 
-` + schcmds.MappingFileHelp +
+` + jsonInputFileHelp +
```
fulghum, line 101: Probably still helpful to include the mapping file info here, since you can use a mapping file and a JSON input file together (from what I understand at least).

fulghum (approved):
Looks great! Thanks for iterating on the details.

PR #5306, opened 2023-02-03
dolt_remote_branches table

zachmu:
New system table shows remote branches with same schema as existing `dolt_branches` table

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/branches_table.go
+++ b/go/libraries/doltcore/sqle/dtables/branches_table.go
@@ -18,11 +18,12 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/types"
+
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
-	"github.com/dolthub/go-mysql-server/sql"
-	"github.com/dolthub/go-mysql-server/sql/types"
 )
 
 var _ sql.Table = (*BranchesTable)(nil)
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/sqle/dtables/branches_table.go
```diff
@@ -34,34 +35,51 @@ var _ sql.ReplaceableTable = (*BranchesTable)(nil)
...
 // BranchesTableName
 func (bt *BranchesTable) String() string {
+	if bt.all {
+		return doltdb.AllBranchesTableName
+	}
 	return doltdb.BranchesTableName
 }
 
 // Schema is a sql.Table interface function that gets the sql.Schema of the branches system table
 func (bt *BranchesTable) Schema() sql.Schema {
```
fulghum, line 70: Seems like it would be nice to have a column that indicated if a branch was a local branch or remote branch so you could quickly filter to just remote branches without having to do string prefix matching on the name.

go/libraries/doltcore/sqle/dtables/branches_table.go
```diff
@@ -88,23 +106,37 @@ type BranchItr struct {
...
+	for i, branch := range branchRefs {
+		commit, err := ddb.ResolveCommitRef(ctx, branch)
 
 		if err != nil {
 			return nil, err
 		}
 
-		branchNames[i] = branch.GetPath()
+		if branch.GetType() == ref.RemoteRefType {
+			branchNames[i] = "  " + "remotes/" + branch.GetPath()
```
fulghum, line 135: Why the leading whitespace? Seems like it would just need to be trimmed off for the value to be used.

fulghum (approved):
Nice! My only note that the whitespace prefix in front of the name to indicate that it's a remote tracking branch seemed odd and a column indicating that it's a remote tracking branch would probably be easier to use (no need to trim leading whitespace, no need to string prefix match). 

Alternatively... having `dolt_branches` and `dolt_remote_branches` might be an even cleaner approach and wouldn't need the extra qualification on local versus remote. Would be easy to `union` them together to get all branches.

go/libraries/doltcore/sqle/dtables/branches_table.go
zachmu: Copypasta error

zachmu:
Great suggestion on `dolt_remote_branches` table, that's much better.

PR #5307, opened 2023-02-03
go/store/nbs: Make a chunkSource able to return a real io.Reader.

reltuk:
With TableFileStore.WriteTableFile() being used on the clone path, we want to turn chunk sources into an io.ReadCloser. In the past, store/nbs always did this by using the ReaderAt to emulate an io.Reader. For files this maybe has OK performance, but for AWS/Blobstore remotes it causes a serial sequence of lots of small reads against blob storage as the io.Copy() implementation copies a few dozen KBs at a time.

This change makes it so a chunkSource can return an actual Reader. file_table_reader leaves the reader at emulation in place for now. One reason to potentially get rid of even that layer is to enable something like .WriteTo() -> sendfile() translations. Those are not currently directly in place because we currently instrument the returned Reader with byte counting, for example, so we can report on Clone progress. One reason to leave the ReaderAt implementation in place, on the other hand, is interactions with file descriptors and the file descriptor cache. This is an investigation for later; it seemed most prudent to leave it for now since it's not currently causing pain.

### DIFF
--- a/go/store/nbs/bs_persister.go
+++ b/go/store/nbs/bs_persister.go
@@ -233,6 +233,11 @@ type bsTableReaderAt struct {
 	bs  blobstore.Blobstore
 }
 
+func (bsTRA *bsTableReaderAt) Reader(ctx context.Context) (io.ReadCloser, error) {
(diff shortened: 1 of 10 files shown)


andy-wm-arthur (approved):
LGTM!

PR #5308, opened 2023-02-03
Allow the pprof server to handle connections from non-localhost

PavelSafronov:
Allow the pprof server to handle connections from non-localhost machines.

This change is necessary to reach a Docker pprof server from host machine.


bheni (approved):
lgtm

PR #5311, opened 2023-02-04
makes -f default behavior for dolt fetch

stephkyou:
This change makes force fetch the default behavior for `dolt fetch` and removes the `-f` force flag to prevent errors when fetching after a force push.

fixes: https://github.com/dolthub/dolt/issues/4943


fulghum (commented):
This looks great! While you're in here... could you also please take a quick look at `dolt_fetch.go` and apply the same change so that the SQL `dolt_fetch()` stored procedure has the same behavior as `dolt fetch` on the command line?

fulghum (approved):
Nice job! Looks great! ✨

PR #5315, opened 2023-02-06
slightly better prefix compare

JCOR11599:
A slight performance improvement for `findPrefix` when looking through table indexes that avoids a `make()` call and `bytes.Compare`

### DIFF
--- a/go/store/nbs/table_index.go
+++ b/go/store/nbs/table_index.go
@@ -333,8 +333,6 @@ func (ti onHeapTableIndex) lookupOrdinal(h *addr) (uint32, error) {
 // findPrefix returns the first position in |tr.prefixes| whose value == |prefix|.
 // Returns |tr.chunkCount| if absent
 func (ti onHeapTableIndex) findPrefix(prefix uint64) (idx uint32) {
-	query := make([]byte, addrPrefixSize)
-	binary.BigEndian.PutUint64(query, prefix)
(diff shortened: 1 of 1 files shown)


go/store/nbs/table_index.go
```diff
@@ -333,16 +333,15 @@ func (ti onHeapTableIndex) lookupOrdinal(h *addr) (uint32, error) {
...
-	binary.BigEndian.PutUint64(query, prefix)
 	// NOTE: The golang impl of sort.Search is basically inlined here. This method can be called in
 	// an extremely tight loop and inlining the code was a significant perf improvement.
 	idx, j := 0, ti.count
 	for idx < j {
 		h := idx + (j-idx)/2 // avoid overflow when computing h
 		// i ≤ h < j
 		o := int64(prefixTupleSize * h)
-		if bytes.Compare(ti.prefixTuples[o:o+addrPrefixSize], query) < 0 {
+		tmp := binary.BigEndian.Uint64(ti.prefixTuples[o : o+addrPrefixSize])
```
max-hoffman, line 343: aaron suggested maybe keeping the original prefix as `[8]bytes` in the parent caller, eliminate either conversion.

max-hoffman (approved):
LGTM, one possible improvement

go/store/nbs/table_index.go
JCOR11599: After some investigation, it turns out that `bytes.Compare` is slower than just casting and comparing `uint64`s

PR #5326, opened 2023-02-07
fixes timing for multi statement queries

stephkyou:
This change fixes timing for multi statement queries to print the time to execute the query instead of the time to print the results or the time since the first query started.


fulghum (approved):
Nice work! Looks great! 🚢

PR #5327, opened 2023-02-07
print something for FK diffs

JCOR11599:
It is possible to commit foreign keys that haven't been resolved yet.
If there are differences in the column tags referenced in a foreign key, but not in the tables' schemas, our diffs are blank (even the hashes are the same). 

This just prints that something was modified for the affected foreign keys.


go/cmd/dolt/commands/diff_output.go
```diff
@@ -247,6 +247,12 @@ func (t tabularDiffWriter) WriteSchemaDiff(ctx context.Context, toRoot *doltdb.R
 		cli.Println(textdiff.LineDiff(fromCreateStmt, toCreateStmt))
 	}
 
+	if len(td.FromFks) > 0 || len(td.ToFks) > 0 {
```
Hydrocharged, line 250: This should be checking if the foreign keys have been resolved. AFAICT, this is going to print `resolved foreign key` whenever _any_ change is made to a foreign key. In fact, if a foreign key was deleted, we'll still get `resolved foreign key`. Need some tests for all of the scenarios:

- New foreign key is added and resolved
- New foreign key is added and not resolved
- Existing foreign key that was resolved is deleted
- Existing foreign key that was not resolved is deleted
- Existing foreign key that was resolved is "_modified_" (dropped then readded with a change)
- Existing foreign key that was not resolved is "_modified_"
- Existing foreign key is resolved

Hydrocharged (changes requested):
Needs quite a bit more work

integration-tests/bats/diff.bats
```diff
@@ -449,6 +449,106 @@ EOF
...
+create table child (j int primary key, constraint fk foreign key (j) references parent (i));
+SQL
+    run dolt diff
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ 'resolved foreign key' ]] || false
+
+    dolt add -A
+    dolt commit -m "init commit"
+    dolt sql -q "alter table child drop foreign key fk"
+    dolt sql -q "alter table child add constraint fk foreign key (j) references parent (i)"
```
Hydrocharged, line 526: This actually doesn't register as a diff at all, so it doesn't show `resolved foreign key` because there was no diff in the first place. You need to modify the calling column or something to actually create a change in the foreign key.

go/cmd/dolt/commands/diff_output.go
```diff
@@ -247,6 +247,12 @@ func (t tabularDiffWriter) WriteSchemaDiff(ctx context.Context, toRoot *doltdb.R
 		cli.Println(textdiff.LineDiff(fromCreateStmt, toCreateStmt))
 	}
 
+	for _, fk := range td.ToFks {
+		if len(fk.ReferencedTableColumns) > 0 {
+			cli.Println(fmt.Sprintf("resolved foreign key `%s` on table `%s`", fk.Name, fk.TableName))
+		}
+	}
+
```
Hydrocharged, line 255: My previous feedback was:
> This should be checking if the foreign keys have been resolved. AFAICT, this is going to print `resolved foreign key` whenever _any_ change is made to a foreign key. In fact, if a foreign key was deleted, we'll still get `resolved foreign key`.

This looks like you've just traded it in for a different issue. Instead of printing something on any change, this looks like it'll print something on any diff involving a resolved foreign key. We're not checking if we went from unresolved to resolved, we're just checking if the new key is resolved.

Hydrocharged (changes requested):
Still needs work

integration-tests/bats/diff.bats
```diff
@@ -449,6 +449,107 @@ EOF
...
+    dolt add -A
+    dolt commit -m "init commit"
+    dolt sql -q "alter table child drop foreign key fk"
+
+    run dolt diff
+    [ "$status" -eq 0 ]
+    [[ ! "$output" =~ 'resolved foreign key' ]] || false
+}
+
+@test "diff: existing foreign key that was resolved is modified" {
```
Hydrocharged, line 514: Add a variation of this test where a foreign key that was not resolved is modified

Hydrocharged (approved):
LGTM! Just need to add one more test

PR #5330, opened 2023-02-07
Fixed race when multiple clients attempt to connect to the same branc…

zachmu:
…h the first time a replica fetches it

### DIFF
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -264,7 +264,7 @@ func (p DoltDatabaseProvider) attemptCloneReplica(ctx *sql.Context, dbName strin
 func (p DoltDatabaseProvider) HasDatabase(ctx *sql.Context, name string) bool {
 	_, err := p.Database(ctx, name)
 	if err != nil && !sql.ErrDatabaseNotFound.Is(err) {
-		ctx.GetLogger().Errorf(err.Error())
+		ctx.GetLogger().Warnf("Error getting database %s: %s", name, err.Error())
 	}
 	return err == nil
 }
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -193,7 +193,7 @@ func (p DoltDatabaseProvider) Database(ctx *sql.Context, name string) (db sql.Da
 
 	if !ok {
 		db, err = p.databaseForClone(ctx, name)
-		
+
 		if err != nil {
 			return nil, err
 		}
@@ -733,7 +733,7 @@ func (p DoltDatabaseProvider) databaseForRevision(ctx *sql.Context, revDB string
 	if isBranch {
 		// fetch the upstream head if this is a replicated db
 		if replicaDb, ok := srcDb.(ReadReplicaDatabase); ok {
-			// TODO move this out of analysis phase, should only happen at read time, when the transaction begins (like is 
+			// TODO move this out of analysis phase, should only happen at read time, when the transaction begins (like is
 			//  the case with a branch that already exists locally)
 			err := p.ensureReplicaHeadExists(ctx, resolvedRevSpec, replicaDb)
 			if err != nil {
@@ -760,7 +760,7 @@ func (p DoltDatabaseProvider) databaseForRevision(ctx *sql.Context, revDB string
 	if isTag {
 		// TODO: this should be an interface, not a struct
 		replicaDb, ok := srcDb.(ReadReplicaDatabase)
-		
+
 		if ok {
 			srcDb = replicaDb.Database
 		}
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -215,8 +216,59 @@ func (rrd ReadReplicaDatabase) PullFromRemote(ctx *sql.Context) error {
 	return nil
 }
 
-func (rrd ReadReplicaDatabase) RebaseSourceDb(ctx *sql.Context) error {
-	return rrd.srcDB.Rebase(ctx)
+// CreateLocalBranchFromRemote pulls the given branch from the remote database and creates a local tracking branch for it.
+func (rrd ReadReplicaDatabase) CreateLocalBranchFromRemote(ctx *sql.Context, branchRef ref.BranchRef) error {
+	_, err := rrd.limiter.Run(ctx, "pullNewBranch", func() (any, error) {
```
max-hoffman, line 221: this is maybe subject to a correctness error where two clients try to connect to two different branches, and they will both end up connecting to the branch for whichever calls this func first. I.e. the name needs to be unique to the branch?

zachmu: This is substantially rewritten, plz take another look

zachmu:
This breaks a couple of the async push bats tests. I didn't change anything directly related to that, and adding sleeps makes it work again. Which makes me think that clean shutdown logic has been broken for some time, and we just passed some threshold where we are now losing the race.

zachmu:
It looks like graceful server shutdown on kill is just broken... there's no signal handling to invoke the shutdown logic, it just relies on a defer:

```go
	sqlEngine, err := engine.NewSqlEngine(
		ctx,
		mrEnv,
		engine.FormatTabular,
		config,
	)
	if err != nil {
		return err, nil
	}
	defer sqlEngine.Close()
```

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -1088,58 +1036,24 @@ func isLocalBranch(ctx context.Context, ddbs []*doltdb.DoltDB, branchName string
...
+func isRemoteBranch(ctx context.Context, ddbs []*doltdb.DoltDB, branchName string) (string, bool, error) {
 	for _, ddb := range ddbs {
-		bn, branchExists, remoteRef, err := ddb.HasRemoteTrackingBranch(ctx, branchName)
+		bn, branchExists, _, err := ddb.HasRemoteTrackingBranch(ctx, branchName)
 		if err != nil {
 			return "", false, err
 		}
 
 		if branchExists {
-			err = createLocalBranchFromRemoteTrackingBranch(ctx, srcDB.DbData(), ddb, branchName, remoteRef)
```
max-hoffman, line 1102: oof

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -361,6 +350,142 @@ func pullBranches(
...
+						}
+					}
+
+					continue REFS
+				} else {
+					switch remoteRef.Ref.GetType() {
+					case ref.BranchRefType:
+						err := rrd.createNewBranchFromRemote(ctx, remoteRef, trackingRef)
+						if errors.Is(err, datas.ErrOptimisticLockFailed) {
+							continue OPTIMISTIC_RETRY
```
max-hoffman, line 410: is there a different rrd for every client? i'm confused how two clients could trigger this create retry, since the limiter is supposed to collapse the parallel calls. same for `pullLocalBranch` but that one seems more reasonable. What are the concurrent events that trigger retries? a write overlapping a pull?

max-hoffman (approved):
I think the changes are correct, but I don't understand how they are triggered, and i'm not seeing tests that try to trigger optimistic lock failures.

go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -361,6 +350,140 @@ func pullBranches(
...
+		pullErr := rrd.ddb.PullChunks(ctx, rrd.tmpDir, rrd.srcDB, remoteHashes, nil)
+
+	REFS: // every successful pass through the loop below must end with CONTINUE REFS to get out of the retry loop
+		for _, remoteRef := range remoteRefs {
+			trackingRef := ref.NewRemoteRef(rrd.remote.Name, remoteRef.Ref.GetPath())
+			localRef, localRefExists := localRefsByPath[remoteRef.Ref.GetPath()]
+
+			// loop on optimistic lock failures
+		OPTIMISTIC_RETRY:
+			for {
```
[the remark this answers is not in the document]
zachmu: The problem is that `break` breaks out of the switch, not the loop. I'll see if I can eliminate the switch and make this easier to read.

go/libraries/doltcore/sqle/read_replica_database.go
zachmu: Any manifest update can cause an optimistic lock write failure. There are writes that take place outside a limiter run (specifically, during pullBranchesAndUpdateWorkingSet)

Actually the work in CreateLocalBranch needs similar protection for this reason, but I'll follow up for that.

PR #5331, opened 2023-02-08
Add child refs to memtable only when a chunk was added.

PavelSafronov:
Fixes https://github.com/dolthub/dolt/issues/5301

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -639,7 +639,7 @@ func (nbs *NomsBlockStore) addChunk(ctx context.Context, ch chunks.Chunk, addrs 
 	if addChunkRes == chunkAdded {
 		nbs.mt.addChildRefs(addrs)
 	}
-	return true, nil
+	return addChunkRes == chunkAdded || addChunkRes == chunkExists, nil
 }
 
 // refCheck checks that no dangling references are being committed.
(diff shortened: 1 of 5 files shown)


go/store/nbs/store.go
```diff
@@ -634,12 +634,12 @@ func (nbs *NomsBlockStore) addChunk(ctx context.Context, ch chunks.Chunk, addrs
...
 		nbs.mt = newMemTable(nbs.mtSize)
-		ok = nbs.mt.addChunk(a, ch.Data())
+		addChunkRes = nbs.mt.addChunk(a, ch.Data())
 	}
-	if ok {
+	if addChunkRes == chunkAdded {
 		nbs.mt.addChildRefs(addrs)
 	}
-	return ok, nil
+	return true, nil
```
reltuk, line 642: To maintain the previous semantics, I think:

```suggestion
	return addChunkRes != chunkNotAdded, nil
```

or

```suggestion
	return addChunkRes == chunkAdded || addChunkRes == chunkExists, nil
```

For now I think the only time that would be hit is if a chunk exceeded maxMemTableSize, which would be a whole set of different problems in itself...

reltuk (approved):
LGTM!

go/store/nbs/store.go
PavelSafronov: Going with the second approach.

PR #5335, opened 2023-02-08
Added special diffs for stored procedures

Hydrocharged:
This adds special git-style diffs for stored procedures. There are two modes. The default displays line changes with the familiar `+` and `-` prepended to each line. The second requires the flag `--proc-in-place`, and shows the changes in-place with the color signifying the change.


Hydrocharged:
Screenshots from the bats tests.

This is the new default view for stored procedure diffs
image

This is with the `--proc-in-place` argument
image

timsehn:
Should we use the same format for views as well?

timsehn:
And maybe LICENSE/README?

Hydrocharged:
@timsehn  I’ll make a separate PR for LICENSE/README. I should also probably rename the argument since `—proc-in-place` is fairly specific. What are some good ideas for a more general argument name?

Hydrocharged:
Regarding views, those would be a little trickier to implement since they share a system table. I’ll gather some opinions regarding whether we should apply it there too.

timsehn:
`--use-text-diff`
`--text-diff`
`--diff-mode=tabular, text` #default tabular

Those are my ideas.

timsehn:
Now I understand...I think you have to go with `text-diff-mode=normal, off, in-place` default being normal. Off just shows you a normal table diff. in-palce does your special one.

Hydrocharged:
I like that, I’ll make the changes

go/libraries/doltcore/diff/diff.go
```diff
@@ -40,6 +41,14 @@ const (
 	ModifiedNew
 )
 
+var ColDiffColors = map[ChangeType]*color.Color{
```
zachmu, line 44: You didn't create this problem, but this is the wrong layer in the stack to encode color information for display. Should live in the CLI package, and get passed into diff writers at construction time so they know what colors to write.

zachmu (approved):
LGTM, nice new functionality.

My only real comment is that this shouldn't be limited in scope to just stored procedures, it should be generalized for all textual diffs in any table. It's easy to add for docs and dolt_schema, but it would be really cool to figure out how to adapt it for all text columns. That would require some clever algorithms to give reasonable looking output when a table has multiple text columns, or a mix of text and non-text columns. But I think treating text based diff differently is compelling enough it's worth spending some time playing with.

Hydrocharged:
This is how it looks as of the current iteration
image

go/cmd/dolt/commands/diff.go
```diff
@@ -201,6 +211,14 @@ func parseDiffArgs(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
 	switch strings.ToLower(f) {
 	case "tabular":
 		dArgs.diffOutput = TabularDiffOutput
+		switch strings.ToLower(apr.GetValueOrDefault(DiffMode, "line")) {
```
zachmu, line 214: Not sure about changing this default. Maybe let's keep it the same for now (row based) and give the team a little time to play with it.

go/libraries/doltcore/diff/diff.go
```diff
@@ -63,6 +63,9 @@ type SqlRowDiffWriter interface {
 	// the input row.
 	WriteRow(ctx context.Context, row sql.Row, diffType ChangeType, colDiffTypes []ChangeType) error
 
+	// WriteCombinedRow writes the diff of the rows given as a single, combined row.
+	WriteCombinedRow(ctx context.Context, oldRow, newRow sql.Row, inPlace bool) error
```
zachmu, line 67: Rather have an enum than a bool for the last param

go/libraries/doltcore/table/untyped/tabular/fixedwidth_tablewriter.go
```diff
@@ -142,28 +164,30 @@ func (w *FixedWidthTableWriter) WriteColoredRow(ctx context.Context, r sql.Row,
...
 	}
 
-	for i := range r {
-		str, err := w.stringValue(i, r[i])
-		if err != nil {
-			return row, err
+	for i, str := range r {
+		var stringWidth FixedWidthString
+		if len(widths[i].Lines) == 0 {
+			stringWidth = NewFixedWidthString(str)
```
zachmu, line 181: var name could use an update

go/libraries/doltcore/table/untyped/tabular/fixedwidth_tablewriter.go
```diff
@@ -142,28 +164,30 @@ func (w *FixedWidthTableWriter) WriteColoredRow(ctx context.Context, r sql.Row,
...
 }
 
-func (w *FixedWidthTableWriter) sampleRow(r sql.Row, colors []*color.Color) (tableRow, error) {
+func (w *FixedWidthTableWriter) sampleRow(r []string, widths []FixedWidthString, colors []*color.Color) (tableRow, error) {
+	if len(widths) == 0 {
+		widths = make([]FixedWidthString, len(r))
+	}
 	row := tableRow{
-		columns: make([]string, len(r)),
+		columns: r,
```
zachmu, line 172: Kinda weird that we have both `columns` and `widths` here... seems like this separation is a holdover from the previous implementation that use integer widths. FixedWidthString is the processed version of the string result and should be self-contained, usable for printing the table by itself. No need for a separate set of unprocessed column strings, right?

go/cmd/dolt/commands/diff.go
zachmu: On consideration, I think I'm fine with this. It will only differ from row based in the case that one of the columns being diffed is a text column with a newline in it, correct?

zachmu (approved):
LGTM, very cool to see this improved diff output

(I remember briefly trying to get strings with newlines to print better in like 2019 and giving up because the mysql shell also does a terrible job with newlines)

PR #5340, opened 2023-02-09
go/store/nbs: Optimize memory and disk utilization during GC.

reltuk:


reltuk:
#benchmark

go/store/nbs/gc_copier.go
```diff
@@ -92,13 +93,28 @@ func (gcc *gcCopier) copyTablesToDir(ctx context.Context, tfp tableFilePersister
 		}, nil
 	}
 
+	// Attempt to rename the file to the destination if we are working with a fsTablePersister...
+	if fstp, ok := tfp.(*fsTablePersister); ok {
```
zachmu, line 97: Probably better to have an interface for this set of requirements / capabilities

go/store/nbs/cmp_chunk_table_writer.go
```diff
@@ -80,11 +75,6 @@ func (tw *CmpChunkTableWriter) AddCmpChunk(c CompressedChunk) error {
 		panic("NBS blocks cannot be zero length")
 	}
 
-	if tw.chunkHashes.Has(c.H) {
```
zachmu, line 83: Removing this check changes the contract for the interface. Now the caller has to ensure no duplicate chunks or you end up with duplicates in the table files produced.

Is that expectation documented anywhere? Not super familiar with this set of files, but it worries me a bit that we aren't consistent with how these interfaces are expected to be used

zachmu (approved):
LGTM, small misgivings to think about on future passes

go/store/nbs/cmp_chunk_table_writer.go
reltuk: It does change the contract, but in the past, you would get an error --- not produce a table file that had duplicate chunks.

It's perfectly valid for a table file to have duplicate chunks, and we create them that way by design when we conjoin table files. I don't know why this check exists, but I'm guessing it was here to check prevent regressions where upstream code changed in a way that created a bunch of duplicate chunks in a written table file.

If we wanted to retain close-to-them-same semantics, we could preform the uniqueness check when we write the index. We could also do that, but optionally enable it and use that when running bats and unit tests but not in the shipped product.

Definitely this hash map isn't carrying its weight, but I'm not even sure the check itself is carrying its weight...

zachmu (approved):
LGTM

PR #5345, opened 2023-02-10
Update README.md

bpf120:
added new links/copy in the what is Dolt for section


timsehn:
This is a bit aggressive. Let me think on it.

PR #5348, opened 2023-02-13
fix --track for dolt_checkout

stephkyou:
`--track` is not working with `call dolt_checkout` as expected. This change updates `dolt_checkout` to mirror the before of `dolt checkout` in the CLI.

### DIFF
--- a/integration-tests/bats/remotes.bats
+++ b/integration-tests/bats/remotes.bats
@@ -2020,6 +2020,44 @@ SQL
     [[ "$output" =~ "Everything up-to-date." ]] || false
 }
 
+@test "remotes: call dolt_checkout track flag sets upstream" {
+    mkdir remote
+    mkdir repo1
+
+    cd repo1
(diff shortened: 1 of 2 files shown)


integration-tests/bats/remotes.bats
```diff
@@ -2020,6 +2020,44 @@ SQL
...
+
+    cd ..
+    dolt clone file://./remote repo2
+
+    cd repo2
+    dolt branch
+    [[ ! "$output" =~ "other" ]] || false
+
+    run dolt sql << SQL
+    call dolt_checkout('--track', 'origin/other');
```
fulghum, line 2044: Would be nice to test another form of using `--track`, such as if a local branch name is supplied or to test the error case where no arg supplied to `--track` should return an error message.

fulghum (approved):
Code looks good. Just one suggestion for a couple more test cases to cover in BATS.

PR #5358, opened 2023-02-14
[no-release-notes] Removing an unneeded skipped test

fulghum:
We had a skipped test for remote branch refs, but it is no longer needed now that we added BATS tests for the new `dolt_remote_branches` system table. 

Also fixes `system-tables: query dolt_remote_branches system table` test being flakey – it was testing for non-existence of branch `b1`, but occasionally `b1` will show up in a commit hash, so I changed the test the not select all columns.


zachmu (approved):
LGTM

PR #5365, opened 2023-02-15
Handle multiple lines in tabular diffs

Hydrocharged:
This changes our tabular output so that multiline strings display in a consistent manner. This will allow us to properly display text diffs as previewed in https://github.com/dolthub/dolt/pull/5335. Modifications to that PR will follow this one.

This is how diffs will look like with this PR:
image

Performance-wise, I tested a `SELECT` over 5000 rows with two text columns containing all multiline strings for every row, and I averaged around 10% **faster** than the old implementation. We now sample every 100 rows rather than only the first 100, and this makes each local chunk of 100 rows visually consistent. In addition, I've cut down on the amount of formatting that is done to the columns, which is what I believe is the source of the speed increase.

As a side benefit, `dolt sql-client` uses the same tabular output logic, so now our SQL client will handle multiline strings from _any_ compatible database, not just Dolt. The default MySQL client behaves like our old implementation, so I think this is another cool win.


go/libraries/doltcore/table/untyped/tabular/stringwidth.go
```diff
@@ -19,19 +19,63 @@ import (
...
+type DisplayLine struct {
+	Width     int // Width is the display width of this specific line.
+	ByteStart int // ByteStart is the beginning offset of the original string that this line represents.
+	ByteEnd   int // ByteEnd is the ending offset of the original string that this line represents (excludes newline).
+}
+
 // StringWidth returns the number of horizontal cells needed to print the given text. It splits the text into its
 // grapheme clusters, calculates each cluster's width, and adds them up to a total.
-func StringWidth(text string) (width int) {
+func StringWidth(text string) DisplayString {
```
zachmu, line 38: This is now a constructor, call it NewFixedWidthString

go/libraries/doltcore/table/untyped/tabular/stringwidth.go
```diff
@@ -19,19 +19,63 @@ import (
 	"github.com/rivo/uniseg"
 )
 
+// DisplayString contains all of the information needed to properly display a multiline string in tabular mode.
+type DisplayString struct {
```
zachmu, line 23: FixedWidthString?

go/libraries/doltcore/table/untyped/tabular/stringwidth.go
```diff
@@ -19,19 +19,63 @@ import (
...
+	TotalWidth   int
```
zachmu, line 24: TotalWidth and DisplayWidth could use some disambiguation

go/libraries/doltcore/table/untyped/tabular/fixedwidth_tablewriter_test.go
```diff
@@ -129,7 +129,7 @@ func TestFixedWidthWriter(t *testing.T) {
 		assert.Equal(t, expectedTableString, stringWr.String())
```
zachmu, line 129: Could use some tests of the new multiline logic here

zachmu (approved):
Change looks good overall, but surprisingly lacking tests for a daylon PR

Hydrocharged:
I'll add the feedback to my other PR and get this merged in

Hydrocharged:
It's lacking tests as the tests will primarily be in the other PR

PR #5374, opened 2023-02-17
go/store/nbs: Fix some quota leaks in conjoin, GC.

reltuk:
Adds a paranoid mode where we noisely detect unclosed table files. The mode can be enabled by setting an environment variable.

Fixes some unit tests, including all of go/store/... to run cleanly under the paranoid mode.

Changes the quota interface to:
* Release |sz int| bytes instead of requiring a []byte with the correct length to show up.
* Work with |int| instead of |uint64|, since MaxUint64 is never allocatable and MaxInt32+z is only allocatable on 64-bit platforms.
* Not return an error on Release(). Implementations should not fail to release quota.

### DIFF
--- a/go/store/nbs/conjoiner.go
+++ b/go/store/nbs/conjoiner.go
@@ -197,6 +197,13 @@ func conjoinTables(ctx context.Context, conjoinees []tableSpec, p tablePersister
 			return
 		})
 	}
+	defer func() {
+		for _, cs := range toConjoin {
+			if cs != nil {
+				cs.close()
+			}
+		}
+	}()
 	if err = eg.Wait(); err != nil {
 		return tableSpec{}, err
 	}
@@ -207,6 +214,7 @@ func conjoinTables(ctx context.Context, conjoinees []tableSpec, p tablePersister
 	if err != nil {
 		return tableSpec{}, err
 	}
+	defer conjoinedSrc.close()
 
 	stats.ConjoinLatency.SampleTimeSince(t1)
 	stats.TablesPerConjoin.SampleLen(len(toConjoin))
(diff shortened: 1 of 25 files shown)


go/store/nbs/conjoiner.go
```diff
@@ -197,6 +197,13 @@ func conjoinTables(ctx context.Context, conjoinees []tableSpec, p tablePersister
 			return
 		})
 	}
+	defer func() {
```
reltuk, line 200: This was the primary leak, AFAICT. For a conjoin, all table files are newly opened and read, instead of using our existing table file indexes. The newly conjoined file is also opened as a chunk source down below. All of these chunk sources were previously leaked.

go/store/nbs/store.go
```diff
@@ -1625,9 +1625,9 @@ func (nbs *NomsBlockStore) swapTables(ctx context.Context, specs []tableSpec) (e
 	if err != nil {
 		return err
 	}
+	oldTables := nbs.tables
 	nbs.tables, nbs.upstream = ts, upstream
-
-	return nil
+	return oldTables.close()
```
reltuk, line 1630: This was a leak on GC. Not relevant in the doltremoteapi context, but needs to be fixed for paranoid mode.

go/store/nbs/table_index.go
```diff
@@ -506,19 +539,11 @@ func (ti onHeapTableIndex) Close() error {
...
-	if err := ti.q.ReleaseQuotaBytes(ti.offsets2); err != nil {
-		return err
-	}
-	if err := ti.q.ReleaseQuotaBytes(ti.suffixes); err != nil {
-		return err
+	if TableIndexAssertClosedWithGCFinalizer {
+		runtime.SetFinalizer(ti.refCnt, nil)
 	}
-	return ti.q.ReleaseQuotaBytes(ti.footer)
+	ti.q.ReleaseQuotaBytes(len(ti.prefixTuples) + len(ti.offsets1) + len(ti.offsets2) + len(ti.suffixes) + len(ti.footer))
```
reltuk, line 545: Assumes no overflow here...

andy-wm-arthur (approved):
LGTM!

PR #5380, opened 2023-02-17
`dolt_diff` respects qualified database

stephkyou:
This change corrects the `dolt_diff` system table to use the qualified database if provided instead of always defaulting to the session's current database.

fixes: https://github.com/dolthub/dolt/issues/5226

### DIFF
--- a/go/libraries/doltcore/sqle/database.go
+++ b/go/libraries/doltcore/sqle/database.go
@@ -425,7 +425,7 @@ func (db Database) getTableInsensitive(ctx *sql.Context, head *doltdb.Commit, ds
 			}
 		}
 
-		dt, found = dtables.NewUnscopedDiffTable(ctx, db.ddb, db.name, head), true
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/database.go
```diff
@@ -425,7 +425,7 @@ func (db Database) getTableInsensitive(ctx *sql.Context, head *doltdb.Commit, ds
 			}
 		}
 
-		dt, found = dtables.NewUnscopedDiffTable(ctx, db.ddb, head), true
+		dt, found = dtables.NewUnscopedDiffTable(ctx, db.ddb, db.name, head), true
```
fulghum, line 428: (minor) it would be nice to pass `db.name` before `db.ddb`, just to match the other `New...` function calls in this method (i.e. lines 430 and 450).

fulghum (approved):
Looks good! 🚢  Just one small suggestion on the param order.

PR #5385, opened 2023-02-18
go/store/nbs: Make the panic in the finalizer for an unclosed table file index always on.

reltuk:
Add a flag to make it not take the stack trace at the time of index creation.
Disable stack trace taking by default.

Make all unit tests pass the sanity checking.


andy-wm-arthur (commented):
I'm uncertain if we want to capture the stack trace by default. I think I prefer not to do this for Dolt releases in particular, and to only leave it on for dev builds and services

reltuk:
> I'm uncertain if we want to capture the stack trace by default. I think I prefer not to do this for Dolt releases in particular, and to only leave it on for dev builds and services

Yeah, the way it is set up here, it is enabled by default in services and unit tests. Every dolt release has it off. Technically remotesrv in this PR also has it on…

timsehn:
We ever gonna merge this?

reltuk:
@timsehn I'll sync up with Andy about whether we should and whether the defaults are currently right.

andy-wm-arthur (approved):
LGTM!

PR #5386, opened 2023-02-18
go/store/nbs: aws_table_persister.go: Fix conjoin code to not error on a single short Read() by using io.ReadFull.

reltuk:
In #5307 we started returning a real io.Reader for chunkSource.reader(ctx). That means the Read() calls can come back short if they are directly attached to a net.TCPConn, for example. It is not an error for a Read() to come back short. Use io.ReadFull to correctly express the intent in the conjoin code for AWS remotes.

Fixes a bug where doltremoteapi fails commits sometimes.

### DIFF
--- a/go/store/nbs/aws_table_persister.go
+++ b/go/store/nbs/aws_table_persister.go
@@ -25,6 +25,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


andy-wm-arthur (approved):
LGTM

PR #5387, opened 2023-02-20
go/store/nbs: table_set.go: Fix potential to leak cloned table index without closing it in rebase().

reltuk:


reltuk:
Superseded by #5451

PR #5389, opened 2023-02-20
split ZRanges to improve Spatial Index Lookup

JCOR11599:
This PR implements `BIGMIN/LITMAX` algorithm to split up large Z-Ranges into multiple more concise Z-Ranges.
Up to a certain extent, this improves query times.


go/libraries/doltcore/sqle/index/z_encoding.go
```diff
@@ -184,28 +179,118 @@ func ZCell(v types.GeometryValue) val.Cell {
...
+}
+
+// isContinuous checks if the provided zRange is entirely within the bounding box
+func isContinuous(zl, zh uint64, prefixLength int) bool {
+	mask := uint64(math.MaxUint64 >> prefixLength)
+	return (zl&mask) == 0 && (zh&mask) == mask
+}
+
+// splitZRanges is a helper function to SplitZRanges
+func splitZRanges(zRange ZRange, zSize float64, zShamt, depth int, acc []ZRange) []ZRange {
```
andy-wm-arthur, line 236: This function is pretty impossible to read. I think most all devs will find it opaque if they need to modify it. At minimum you should replace all the literals with constants and try to describe their purpose

andy-wm-arthur (approved):
The perf wins here are awesome, the code is not ideal. I think we should land this for now and come back an refactor later

PR #5393, opened 2023-02-21
go/store/nbs: use chunk journal index to speedup bootstrapping

andy-wm-arthur:


go/store/nbs/journal_writer.go
```diff
@@ -142,14 +149,72 @@ var _ io.Closer = &journalWriter{}
...
+				}
+				// populate range hashmap
+				for _, l := range deserializeLookups(r.payload) {
+					wr.ranges.put(l.a, l.r)
+				}
+				// record a high-water-mark for the indexed portion of the journal
+				wr.indexed = int64(r.end)
+				// todo: uncompressed size
+			default:
+				return fmt.Errorf("unknown index record kind (%d)", r.kind)
```
andy-wm-arthur, line 196: reading an unknown record kind will throw an error, causing the index to be deleted in `corruptIndexRecovery`... probably not the behavior we want

go/store/nbs/journal_writer.go
```diff
@@ -142,14 +149,72 @@ var _ io.Closer = &journalWriter{}
...
+
+	if ok {
+		var info os.FileInfo
+		if info, err = wr.index.Stat(); err != nil {
+			return hash.Hash{}, err
+		}
+		err = processIndexRecords(ctx, wr.index, info.Size(), func(o int64, r indexRec) (err error) {
+			switch r.kind {
+			case tableIndexRecKind:
+				// |r.stop| is expected to point to a root hash record in |wr.journal|
```
reltuk, line 180:
```suggestion
				// |r.end| is expected to point to a root hash record in |wr.journal|
```

reltuk (approved):
Generally looks totally reasonable. A few comments.

go/store/nbs/journal_index_record.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		l := readUint32(buf)
+		if buf, err = rdr.Peek(int(l)); err != nil {
+			break
+		}
+
+		// we do not zero-fill the journal index and expect
+		// only complete records that will checksum
+		if !validateIndexRecord(buf) {
```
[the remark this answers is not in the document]
andy-wm-arthur: for now we're overly restrictive on the well-formedness of journal indexes. as you suggested, we could be more permissive or incrementally clean up the index, but for now we'll punt

PR #5399, opened 2023-02-21
level pruning by short circuiting search

JCOR11599:


JCOR11599:
punting on this optimization for now

PR #5404, opened 2023-02-22
[no-release-notes] Refactored session initialization

zachmu:


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -259,6 +258,10 @@ func (d *DoltSession) StartTransaction(ctx *sql.Context, tCharacteristic sql.Tra
 	// TODO: rather than a single database, we need to take a snapshot of all available databases that we use for the
 	//  duration of the transaction
 
+	// for db, state := range d.dbStates {
+	// 	state.
+	// }
+
```
fulghum, line 264: (minor) some dead code hanging around

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -900,6 +919,7 @@ func (d *DoltSession) SwitchWorkingSet(
 		return err
 	}
 
+	// TODO: should this be an error if any database in the transaction is dirty, or just this one?
```
fulghum, line 922: seems like if any database has been marked dirty from the current transaction, we'd want to return an error, since we know we won't be able to commit across all the dirty dbs. If someone doesn't care about the data changes to another db, that could be annoying to block them, but still seems like a reasonable behavior.

go/libraries/doltcore/env/multi_repo_env.go
```diff
@@ -52,9 +57,147 @@ type MultiRepoEnv struct {
...
+	} else {
+		path, err := fs.Abs("")
+		if err != nil {
+			return nil, err
+		}
+		envName := getRepoRootDir(path, string(os.PathSeparator))
+		dbName = dirToDBName(envName)
+		dEnv = oldDEnv
+		// TODO: idk how or why, but this breaks docs.bats
+		//dEnv = Load(ctx, GetCurrentUserHomeDir, fs, doltdb.LocalDirDoltDB, version)
```
fulghum, line 92: (minor) Do we still need this note?

fulghum (approved):
Looks good! 🚢

go/libraries/doltcore/env/multi_repo_env.go
zachmu: Good call, cleaned this logic up

go/libraries/doltcore/sqle/dsess/session.go
zachmu: Yeah, I'll make this concrete in my next PR

PR #5413, opened 2023-02-23
add dolt_patch() stored procedure

jennifersp:
The result of `dolt_patch()` stored procedure is equivalent to `dolt diff -r sql`.
The input to `dolt_patch()` stored procedure is `dolt_patch(<from_revision>, <to_revision>, [<tablename>])`, which are optional 2 revisions and optional table names. 
It does not support two and three dot diffs.
Only `--cached` flag is supported for now.
It returns both schema and data diff result.
Error cases of not being able to get data diff go into warning.


go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -958,3 +958,74 @@ func CalculateDiffSchema(fromSch, toSch schema.Schema) (schema.Schema, error) {
 
 	return schema.UnkeyedSchemaFromCols(schema.NewColCollection(cols...)), nil
 }
+
+//------------------------------------
+// diffTableFunctionRowIter
+//------------------------------------
+
```
fulghum, line 965: It's a little odd to have the diffTableFunctionRowIter in the `diff_table.go` file, since it isn't used by diff_table. Putting this into `diff_iter.go` would probably be better organized. 

Also, might make more sense to rename it to something like `DiffPartitionRowIter` or something, since otherwise the `DiffTableFunction` name is a little misleading if we're using it for other purposes besides the `dolt_diff()` table function.

fulghum (commented):
Looking good. I had lots of small comments, but they're mostly about small nits, naming, or package structuring. Overall, the logic and tests for `dolt_patch()` looked pretty solid.

go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
```diff
@@ -76,10 +78,14 @@ func FmtColPrimaryKey(indent int, colStr string, newline bool) string {
...
 	sb := strings.Builder{}
+	sb.WriteString("  ")
 	if index.IsUnique() {
 		sb.WriteString("UNIQUE ")
 	}
-	sb.WriteString("INDEX ")
+	if index.IsSpatial() {
+		sb.WriteString("SPATIAL ")
+	}
+	sb.WriteString("KEY ")
```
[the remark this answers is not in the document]
jennifersp: Just checked that `INDEX` is used for `CREATE INDEX` statements, and `KEY` is used for `CREATE TABLE` statements. This function was not used anywhere else, but I'll create separate function for using in `CREATE TABLE` statements.

go/libraries/doltcore/sqle/dprocedures/dolt_patch.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (p *patchRowIter) incrementIndexes() {
+	p.idx++
+	if p.idx >= p.stmtLen {
+		p.idx = 0
+		p.stmts = nil
+	}
+}
+
+func maybeResolve(ctx *sql.Context, dbData env.DbData, doltDB *doltdb.DoltDB, spec string) (*doltdb.RootValue, bool, error) {
```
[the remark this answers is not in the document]
jennifersp: Oh this is to be deleted, the new `MaybeResolve` function is now used for both `commands` and `dprocedures` packages.

go/cmd/dolt/commands/diff.go
```diff
@@ -148,7 +145,7 @@ func (cmd DiffCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(FormatFlag, "r", "result output format", "How to format diff output. Valid values are tabular, sql, json. Defaults to tabular.")
 	ap.SupportsString(whereParam, "", "column", "filters columns based on values in the diff.  See {{.EmphasisLeft}}dolt diff --help{{.EmphasisRight}} for details.")
 	ap.SupportsInt(limitParam, "", "record_count", "limits to the first N diffs.")
-	ap.SupportsFlag(CachedFlag, "c", "Show only the unstaged data changes.")
+	ap.SupportsFlag(cli.CachedFlag, "c", "Show only the unstaged data changes.")
```
fulghum, line 148: Looks like we had this doc string backwards! In git, `--cached` is a synonym for `--staged`, so supplying `--cached` means to do a diff using the staged content (and not the unstaged working set content). `dolt diff` is already working correctly, we just need to fix this backwards doc string. 

```suggestion
	ap.SupportsFlag(cli.CachedFlag, "c", "Show only the staged data changes.")
```

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -310,6 +311,12 @@ func CreateLogArgParser() *argparser.ArgParser {
 	return ap
 }
 
+func CreatePatchArgParser() *argparser.ArgParser {
+	ap := argparser.NewArgParser()
+	ap.SupportsFlag(CachedFlag, "c", "Show only the unstaged data changes.")
```
fulghum, line 316: There's another comment below about the `--cached` flag being documented backwards. Similar change needed here, too: 
```suggestion
	ap.SupportsFlag(CachedFlag, "c", "Show only the staged data changes.")
```

go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -517,3 +522,233 @@ func fkSlicesAreEqual(from, to []doltdb.ForeignKey) bool {
...
+		return "", nil
+	default:
+		return "", fmt.Errorf("unexpected row diff type: %v", rowDiffType)
+	}
+}
+
+func getCreateTableStmt(tblName string, sch schema.Schema, pkSchema sql.PrimaryKeySchema, fks []doltdb.ForeignKey, fksParentSch map[string]schema.Schema) (string, error) {
+	sqlSch := pkSchema.Schema
+	colStmts := make([]string, len(sqlSch))
+
```
fulghum, line 664: Thanks for refactoring out as much as we could to share this logic from one place in GMS. Could you just add a note here in the code to explain to future readers why we can't share more? The goal is to encode some of the context and intent so that future readers understand the direction we want to move in and the constraints we hit that made us stop short of having a single function we could call to get the full create table statement generated. 

I'd probably mention something about us wanting to keep that `create table` statement generation code in one place and mention that because Dolt uses it's own Schema and Column types that aren't directly compatible with GMS' versions, we refactored out common generation parts, but ideally we could find a way to smooth that out even more in the future.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -777,6 +777,53 @@ var DoltScripts = []queries.ScriptTest{
...
+			"CALL DOLT_COMMIT('-Am','added tables')",
+
+			"ALTER TABLE child ADD CONSTRAINT fk_named FOREIGN KEY (v1) REFERENCES parent(v1);",
+			"insert into parent values (0, 1, 2, NULL);",
+			"ALTER TABLE parent DROP PRIMARY KEY;",
+			"ALTER TABLE parent ADD PRIMARY KEY(id, id_ext);",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query: "CALL DOLT_PATCH()",
```
fulghum, line 794: This is such a cool feature! 😄

go/libraries/doltcore/sqle/dtables/diff_iter.go
```diff
@@ -484,3 +484,66 @@ func maybeTime(t *time.Time) interface{} {
...
+
+type diffPartitionRowIter struct {
+	diffPartitions   *DiffPartitions
+	ddb              *doltdb.DoltDB
+	joiner           *rowconv.Joiner
+	currentPartition *sql.Partition
+	currentRowIter   *sql.RowIter
+}
+
+func NewDiffTableFunctionRowIterForSinglePartition(partition sql.Partition, ddb *doltdb.DoltDB, joiner *rowconv.Joiner) *diffPartitionRowIter {
```
fulghum, line 502: Don't forget to fix up this constructor function name, too. 

Since we're passing in a single partition and there's no longer another constructor function, I'd probably just name this one: `NewDiffPartitionRowIter`

integration-tests/bats/sql-patch.bats
```diff
@@ -0,0 +1,391 @@
...
+    dolt sql -q 'INSERT INTO test (pk, c1, c2, c3, c4, c5) VALUES (2, 11, 0, 0, 0, 0)'
+    dolt add test
+    dolt commit -m "Added a third row"
+
+    run dolt sql -q "CALL DOLT_PATCH('firstbranch','newbranch')"
+    [ "$status" -eq 0 ]
+    # 3 lines for tabular lines and 1 for header and 1 for output statement
+    [ "${#lines[@]}" -eq 5 ]
+
+    match_diff_and_patch_results firstbranch newbranch
```
fulghum, line 40: Oh nice! This is much cleaner! Good job. 😄

integration-tests/bats/sql-patch.bats
```diff
@@ -0,0 +1,391 @@
...
+    diff_array=( "${lines[@]}" )
+
+    if [ -z "$1" ] && [ -z "$2" ]; then
+        run dolt sql -q "CALL DOLT_PATCH()"
+    else
+        run dolt sql -q "CALL DOLT_PATCH('$1','$2')"
+    fi
+    [ "$status" -eq 0 ]
+    [[ "${lines[1]}" =~ "statement" ]] || false
+    patch_array=( "${lines[@]:3}" )
```
fulghum, line 383: Would be nice to do a quick assertion that the diff_array doesn't contain any extra elements (otherwise we could compare all the ones in patch_array and miss that diff_array had more elements). 

ChatGPT told me this should work 😉 (you'll just need to account for the two extra elements in patch_array)
`[ ${#array1[@]} -eq ${#array2[@]} ];`

integration-tests/bats/sql-patch.bats
```diff
@@ -0,0 +1,391 @@
...
+setup() {
+    setup_common
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+@test "sql-patch: output reconciles INSERT query" {
```
fulghum, line 13: I didn't notice a test for the `--cached` flag. I wanted to make sure we were executing it correctly since the doc string was switched around before.

fulghum (approved):
Good work! Thanks for iterating on this one and following up with the extra refactorings. This turned out nicely. 

I'll hop on the related GMS PR next.

integration-tests/bats/sql-patch.bats
jennifersp: I think this might not work for cases where `dolt patch` prints out error message for a table that has schema diff but cannot get the data diff error message. I made it so that this error message comes out as warning for `DOLT_PATCH()`. I added length check for each test for `CALL DOLT_PATCH()` before matching the array results. This would be helpful to check expected number of rows.

PR #5421, opened 2023-02-23
Add new version of `dolt_diff_summary` table function 

tbantle22:
The old version of the `dolt_diff_summary` table function was renamed to `dolt_diff_stat`
This new version of `dolt_diff_summary` returns a summary of changed tables between two refs
```sql
us_jails> select * from dolt_diff_summary("WORKING", "taylor/test");
+-----------------------------+-----------+-------------+---------------+
| table_name                  | diff_type | data_change | schema_change |
+-----------------------------+-----------+-------------+---------------+
| jails                       | modified  | true        | false         |
| test                        | dropped   | false       | true          |
| inmate_population_snapshots | added     | true        | true          |
+-----------------------------+-----------+-------------+---------------+
3 rows in set (0.01 sec)
```

### DIFF
--- a/go/libraries/doltcore/diff/table_deltas.go
+++ b/go/libraries/doltcore/diff/table_deltas.go
@@ -464,8 +464,8 @@ func (td TableDelta) GetSummary(ctx context.Context) (*TableDeltaSummary, error)
 		}, nil
 	}
 
-	// TODO: Modified columns (rename, add) without a data change are not
-	// accounted for here, `dataChanged` is true when it should be false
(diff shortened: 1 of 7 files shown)


fulghum (approved):
Beautiful code (as usual) @tbantle22. I couldn't find much, but had a couple suggestions that might be able to simplify a little bit. 

Would be nice to throw in a quick privilege checking test into `DoltUserPrivTests` in `dolt_queries.go`.

fulghum:
Nice work on the `TableDelta.GetSummary(ctx)` function. Very nice interface now!

PR #5422, opened 2023-02-23
add `dolt stash`

jennifersp:
First pass on `dolt stash` commands include:
- `dolt stash`
- `dolt stash pop [<stash>]`
- `dolt stash drop [<stash>]`
- `dolt stash clear`
- `dolt stash list`

No optional flags are supported for above commands.
Staged changes are not handled in this version.
Available only as CLI commands


go/store/datas/database.go
```diff
@@ -83,6 +83,18 @@ type Database interface {
...
+	// Otherwise, it updates the stashes Dataset as there can only be one stashes Dataset.
+	Stash(ctx context.Context, ds Dataset, stashRootRef types.Ref, headCommitAddr hash.Hash, meta *StashMeta) (Dataset, error)
+
+	// DropStash removes a Stash message from stashes Dataset, which cannot be performed by Delete function.
+	// This function removes a single stash only and does not get removed if there are no entries left.
+	DropStash(ctx context.Context, ds Dataset, idx int) (Dataset, error)
+
+	// ClearStashes removes all Stash messages from stashes Dataset.
+	ClearStashes(ctx context.Context, ds Dataset) (Dataset, error)
+
```
andy-wm-arthur, line 97: Ideally we'd reduce these methods down to something like
`UpdateStashLIst(ctx context.Context, ds Dataset, v types.Value) (Dataset, error)`
that way `Database` is just responsible for reading/writing the stash list atomically, and we can defer manipulation of the stash list to the `doltdb` layer

andy-wm-arthur (commented):
Generally looks great, my only significant feedback is to streamline the changes to `datas.Database`.
Nice work!

andy-wm-arthur (approved):
Thanks for the update! LGTM

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1392,3 +1392,146 @@ func (ddb *DoltDB) GetBranchesByRootHash(ctx context.Context, rootHash hash.Hash
...
+	if !ds.HasHead() {
+		return hash.Hash{}, errors.New("No stash entries found.")
+	}
+
+	return getStashHashAtIdx(ctx, ds, ddb.NodeStore(), idx)
+}
+
+// GetStashAtIdx returns root value of stash working set and head commit of the branch that the stash was made on
+// of the stash at given index.
+func (ddb *DoltDB) GetStashAtIdx(ctx context.Context, idx int) (*RootValue, *Commit, error) {
```
fulghum, line 1526: (minor) Maybe `GetStashRootAtIdx`, since we aren't returning a `Stash` struct, but a `RootValue` (and `Commit`)?

integration-tests/bats/stash.bats
```diff
@@ -0,0 +1,295 @@
...
+    run dolt sql -q "SELECT * FROM test" -r csv
+    [[ "$output" =~ "1,b" ]] || false
+    [[ ! "$output" =~ "1,a" ]] || false
+
+    run dolt stash list
+    [ "$status" -eq 0 ]
+    [ "${#lines[@]}" -eq 1 ]
+}
+
+skiponoldformat() {
```
fulghum, line 290: You could reuse the `skip_nbf_not_dolt()` function from `common.bash` and avoid duplicating this.

jennifersp: I did not use it because it also skips `__DOLT_DEV__`, which we can still test since it's I think using new format. I moved this function into `common.bash` so it's usable in other bats tests as well. 🙂

go/cmd/dolt/commands/stashcmds/stash.go
```diff
@@ -0,0 +1,174 @@
...
+	stagedHash, err := stagedRoot.HashOf()
+	if err != nil {
+		return err
+	}
+
+	if headHash.Equal(workingHash) && headHash.Equal(stagedHash) {
+		return fmt.Errorf("No local changes to save")
+	}
+
+	// TODO: handle cases with staged changes?
```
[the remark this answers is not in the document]
jennifersp: Agreed to get a separate PR out for handling staged changes. For now, having staged set of changes will not allow user to stash their changes. It will be fixed soon.

PR #5423, opened 2023-02-24
updates reset script to disable foreign_key_checks temporarily

stephkyou:
Currently, when reset scripts run between Prepared Script Tests errors will occur as tables with foreign keys get dropped and re-added. This change disables foreign key checks while tables are being dropped and re-added as to avoid these foreign key errors.


max-hoffman (approved):
LGTM ship it

fulghum (approved):
Looks great! 🚢

PR #5426, opened 2023-02-24
[no-release-notes] SIGINT gracefully shuts down merge

max-hoffman:


reltuk:
I think this pattern is generally clearer as:

```go
if ctx.Err() != nil {
    return ctx.Err()
}
```

or

```go
if err := ctx.Err(); err != nil {
    return err
}
```

I'm not sure we should add cancelation checks adhoc throughout the stack. It's probably better to add them at principled places where we potentially block and don't currently have a natural cancelation check in something like a channel send/recv. Or, if you know you have a lot of CPU work, you can add them at some fixed interval on that work...but that doesn't seem to be the case here. It seems like for this use case we might be missing cancelation checks in `nbs.table_set.get()` and `nbs.NomsBlockStore.addChunk()`? Might also need some in `chunks.MemoryStore` for testing?

Do we understand why the process exits with SIGINT when we believe the signal handler is installed? I still don't understand how this currently breaks down, instead of just fails to exit in a timely manner on receiving a SIGINT...

max-hoffman:
@reltuk I picked these couple places because they are expensive, but checking in the blockstore does makes more sense. 

And I've just been going off the signal documentation, I'm less familiar with the OS API, but I thought the custom handler intentionally prevents SIGINT from stopping the process, forcing you to manually check ctx to exit? If we never handle the signal, the process just keeps running.

PR #5428, opened 2023-02-24
Add support for using full length commit ids in hashof() procedure

macneale4:


go/libraries/doltcore/sqle/dfunctions/hashof.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Dolthub, Inc.
+// Copyright 2021 Dolthub, Inc.
```
fulghum, line 1: You can go ahead and extend the copyright year range: 
```suggestion
// Copyright 2020-2023 Dolthub, Inc.
```

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -777,6 +777,38 @@ var DoltScripts = []queries.ScriptTest{
 			},
 		},
 	},
+	{
+		Name: "hashof matches commit",
+		SetUpScript: []string{
+			"SELECT commit_hash from dolt_commits",
```
fulghum, line 783: This line isn't doing anything, right? 
```suggestion
```

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -777,6 +777,38 @@ var DoltScripts = []queries.ScriptTest{
...
+			"CREATE TABLE hashof_test (pk int primary key, c1 int)",
+			"INSERT INTO hashof_test values (1,1), (2,2), (3,3)",
+			"CALL DOLT_ADD('hashof_test')",
+			"CALL DOLT_COMMIT('-a', '-m', 'first commit')",
+			"SET @Commit1 = (SELECT commit_hash FROM DOLT_LOG() LIMIT 1)",
+			"INSERT INTO hashof_test values (4,4), (5,5), (6,6)",
+			"CALL DOLT_COMMIT('-a', '-m', 'second commit')",
+			"SET @Commit2 = (SELECT commit_hash from DOLT_LOG() LIMIT 1)",
```
fulghum, line 791: It would be nice to create a branch with a conflicting hash name and assert that `hashof()` will always select the branch. 

Actually... since we don't seem to have a lot of tests that **directly** test `hashof()`, it might also be nice to throw in an invalid commit hash and cover the error case, too, while you're in here.

fulghum (approved):
Looks good! Just a couple minor suggestions.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
macneale4: I attempted creating a branch that has a 32 char name with only the base32 characters, and it was rejected as a bad branch name. I guess we planned for that. I can create one that looks a lot like a full commit id, but I'm not sure what that will add in confidence.

I'l create tests which takes bogus commit id and branch names and ensures they behave.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
macneale4: nope.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
fulghum: oh interesting! no worries if we can't create the branch with a commit hash. Good call to test a bogus branch/commit input.

fulghum (approved):
Looks great! 🚢

PR #5440, opened 2023-02-28
go/store/nbs: file_table_persister: Rework PruneTableFiles so that writes can proceed while we cleanup the old files.

reltuk:


reltuk:
#benchmark

reltuk:
#benchmark

PR #5448, opened 2023-02-28
Add `from_table_name` and `to_table_name` columns to `dolt_diff_summary` table function

tbantle22:
- Also fixes `dolt_status` system table output for renamed tables
- Table name in `dolt diff --summary` CLI command for renamed table uses arrow syntax (i.e. `from_table -> to_table`)


fulghum (approved):
Looks good! 🚢  Nice job catching that rename bug in `dolt_status`.

PR #5450, opened 2023-03-01
MySQL replication fixes for `START/STOP REPLICA`

fulghum:
I found a few places where `START REPLICA` and `STOP REPLICA` weren't working correctly. There were two major cases where the binlog replication applier routine was blocking and not processing the `stopReplication` signal:
- during connection retry and backoff
- while blocking and waiting to read more binlog events over the MySQL connection

I restructured how those cases work, including pulling out a separate goroutine that makes the blocking `conn.ReadBinlogEvent()` call and then feeds the results back to the applier routine through channels.


go/libraries/doltcore/sqle/binlogreplication/binlog_replica_event_producer.go
```diff
@@ -0,0 +1,85 @@
...
+		errorChan: errorChan,
+	}
+	producer.running.Store(true)
+	return producer
+}
+
+// Go starts this binlogEventProducer in a new goroutine
+func (p *binlogEventProducer) Go(_ *sql.Context) {
+	go func() {
+		for p.running.Load() {
```
zachmu, line 51: `for p.IsRunning()` here would be more consistent with below

go/libraries/doltcore/sqle/binlogreplication/binlog_replica_applier.go
```diff
@@ -254,65 +258,68 @@ func (a *binlogReplicaApplier) replicaBinlogEventHandler(ctx *sql.Context) error
 	}
 	engine := server.Engine
 
-	conn, err := a.connectAndStartReplicationEventStream(ctx)
-	if err != nil {
-		return err
-	}
+	var conn *mysql.Conn
+	var eventProducer *binlogEventProducer
+	var binlogEventChan = make(chan mysql.BinlogEvent)
```
zachmu, line 263: We never close these channels, which could leak resources over time (not sure if open channels get cleaned up when they go out of scope or not)

Best practice is for the send-side of a channel to manage its lifecycle including closing it. One reasonable way to do this would be for the Go method of eventProducer to return both these methods, and to close them when it is stopped.

zachmu (approved):
Looks pretty good, just a few suggestions

go/libraries/doltcore/sqle/binlogreplication/binlog_replica_applier.go
```diff
@@ -324,6 +331,20 @@ func (a *binlogReplicaApplier) processBinlogEvent(ctx *sql.Context, engine *gms.
 	createCommit := false
 	commitToAllDatabases := false
 
+	// We don't support checksum validation, so we must strip off any checksum data if present, otherwise
+	// it could get interpreted as part of the data fields and corrupt the fields we pull out. There is not
+	// a future-proof guarantee on the checksum size, so we can't strip a checksum until we've seen the
+	// Format binlog event that definitively tells us if checksums are enabled and what algorithm they use.
+	if a.format.IsZero() == false {
```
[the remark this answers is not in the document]
fulghum: yeah... I tried explaining this in the comment, but I don't think I was direct enough about what `IsZero()` means (just that it's uninitialized and is the zero value for that type. I'll update the comment to be more direct about explaining that.

go/libraries/doltcore/sqle/binlogreplication/binlog_replica_event_producer.go
fulghum: good catch!

go/libraries/doltcore/sqle/binlogreplication/binlog_replica_applier.go
```diff
@@ -254,65 +258,68 @@ func (a *binlogReplicaApplier) replicaBinlogEventHandler(ctx *sql.Context) error
...
+				} else if strings.HasPrefix(sqlError.Message, io.ErrUnexpectedEOF.Error()) {
+					DoltBinlogReplicaController.updateStatus(func(status *binlogreplication.ReplicaStatus) {
+						status.LastIoError = io.ErrUnexpectedEOF.Error()
+						status.LastIoErrNumber = ERNetReadError
+						currentTime := time.Now()
+						status.LastIoErrorTimestamp = &currentTime
+					})
+					eventProducer.Stop()
+					eventProducer = nil
+					conn.Close()
```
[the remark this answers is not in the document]
fulghum: Good question, but none of these calls expose errors (even though it seems like `mysql.conn.Close()` would)

PR #5458, opened 2023-03-01
[no-release-notes] go/store/nbs: Close files in more places to tests can run faster on windows and cleanup after themselves.

reltuk:


andy-wm-arthur (approved):
LGTM

PR #5460, opened 2023-03-01
Include all diff columns when a table is created or dropped

macneale4:
https://github.com/dolthub/dolt/issues/5439


go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -902,26 +902,31 @@ func GetDiffTableSchemaAndJoiner(format *types.NomsBinFormat, fromSch, toSch sch
...
-// Either may be nil, in which case it will be missing from the resulting schema.
+// Either may be nil, in which case the nil argument will use the schema of the non-nil argument
 func CalculateDiffSchema(fromSch, toSch schema.Schema) (schema.Schema, error) {
-
-	colCollection := schema.NewColCollection()
-	if fromSch != nil {
-		colCollection = fromSch.GetAllCols()
+	var toClmCol, fromClmCol *schema.ColCollection
+	if fromSch == nil && toSch == nil {
+		panic("non-nil argument required to CalculateDiffSchema")
```
fulghum, line 909: Avoid using `panic` and prefer just returning an error. Because `panic` will kill the entire go process, it should only be used if there is an extremely serious problem (e.g. data integrity risk) that would warrant stopping everything immediately. We want to protect a sql-server process used by multiple users from being killed abruptly just because one user hit a bug we weren't expecting. 

It this case, it does seem like this shouldn't ever happen, but returning an error instead will still have the effect of erroring out the query that caused the problem, without taking down the whole server.

fulghum (commented):
Approach looks good!

go/libraries/doltcore/sqle/dtables/diff_table.go
macneale4: yep. facepalm

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -890,44 +876,62 @@ func GetDiffTableSchemaAndJoiner(format *types.NomsBinFormat, fromSch, toSch sch
...
 }
 
-// CalculateDiffSchema returns the schema for the dolt_diff table based on the schemas from the from and to tables.
-// Either may be nil, in which case it will be missing from the resulting schema.
-func CalculateDiffSchema(fromSch, toSch schema.Schema) (schema.Schema, error) {
-
-	colCollection := schema.NewColCollection()
-	if fromSch != nil {
-		colCollection = fromSch.GetAllCols()
+// Convert schemas to ColCollections. One argument must be non-nil. If one is null, the result will be the columns
```
fulghum, line 891: (minor) Godoc style for function or member docs requires that the first word of the comment string be the function or member name. 

```suggestion
// ResolveSchemaCollections converts schemas to ColCollections. One argument must be non-nil. If one is null, the result will be the columns
```

fulghum: You can probably also lowercase this function – that'll keep it from being exported out of the package it lives in.

fulghum: "resolve" also has a pretty specific meaning when used (e.g. "resolving unqualified tables or columns"), so it might be better to pick a less loaded term, like `generateDiffSchemas` or something.

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -890,44 +876,62 @@ func GetDiffTableSchemaAndJoiner(format *types.NomsBinFormat, fromSch, toSch sch
...
-	toSch = schema.MustSchemaFromCols(colCollection)
+	newToSch = schema.MustSchemaFromCols(toClmCol)
+
+	return
+}
+
+// CalculateDiffSchema returns the schema for the dolt_diff table based on the schemas from the from and to tables.
+// Either may be nil, in which case the nil argument will use the schema of the non-nil argument
+func CalculateDiffSchema(fromSch, toSch schema.Schema) (schema.Schema, error) {
+	//
```
fulghum, line 924: Might as well tidy up this empty comment 

```suggestion
```

fulghum (approved):
Logic and tests look good. Just a couple minor Go style comments.

PR #5464, opened 2023-03-01
Three way differ and merge optimizations

max-hoffman:
Refactor merge in a way that makes add-heavy merges faster. Expose interface for computing a three way diff without performing a merge (for statistic/PR info gathering). The compute overhead is mixed for updates, but parentFkConstraint checking is still the main bottleneck there.

There is a general tradeoff executing merges quite larger than memory. Under that threshold, sharing one diff and batching lookups into secondary indexes is mostly faster. Above a certain table size it is faster for every data merge and correctness check to perform its own map diff. In the future I hope we find a smarter way to overlap diffs, merges, and lookups for all of the indexes and checks (primary, secondary, uniqueness checks, foreign key child, secondary child). Hopefully the current organizational pattern pushes us in that direction.


max-hoffman:
#merge-benchmark

max-hoffman:
#merge-benchmark

max-hoffman:
#merge-benchmark

max-hoffman:
#merge-benchmark

max-hoffman:
The update test looks like it spends ~30% of its time checking for parent constraint violations. We should optimize that the same way we did child fk's in a separate PR, use a secondary to iterate.
image

integration-tests/bats/sql-merge.bats
```diff
@@ -1509,77 +1509,6 @@ SQL
     [[ "$output" =~ "our check 'c1' and their check 'c0' both reference the same column(s)" ]] || false
 }
 
-# TODO: what happens when the new data conflicts with modified check?
-@test "sql-merge: non-conflicting constraint modification is preserved" {
```
max-hoffman, line 1513: converted into enginetests

max-hoffman:
#benchmark

PR #5465, opened 2023-03-01
[no-release-notes] go/store/nbs: NomsBlockStore: Have Size() return the size of hte current table set chunk sources, instead of reading the manifest.

reltuk:


coffeegoddd (approved):
LGTM!

PR #5469, opened 2023-03-02
bug fix: binlog multi-db support

fulghum:
We were missing the call to set the session's current database for binlog `Query` events. This was causing queries to execute against the wrong database when using binlog replication with multiple databases. Our binlog tests didn't catch this because the multi-db tests used fully qualified table names, so it didn't matter what database was currently selected. 

This change adds the call to set the current database and updates the multi-db tests to use a mix of `use` statements and fully qualified table names (and also cleans up some of our query error logging to make log output more helpful for debugging).


timsehn (approved):
LGTM.

PR #5489, opened 2023-03-03
Test for for dropping primary keys when foreign keys are present

zachmu:
Also adds a skipped test for a merge that now hangs indefinitely.


go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -648,8 +648,8 @@ var MergeScripts = []queries.ScriptTest{
 		},
 		Assertions: []queries.ScriptTestAssertion{
 			{
-				Query:    "call dolt_merge('b1')",
-				Expected: []sql.Row{{0, 0}},
+				Query:          "call dolt_merge('b1')",
+				ExpectedErrStr: "error: cannot merge table t1 because its primary keys differ",
```
zachmu, line 652: @andy-wm-arthur 

This is now an error, which agrees with the changes I made: can't merge where there are schema changes on both sides of the merge w.r.t. the ancestor, even if left and right have the same schema. In general this isn't safe to do as it can cause the diff algorithm to loop infinitely (in the case where key length has changed only so far as I know).

But in principle we say this is an error until schema merge matures, right? Seem right to you?

PR #5501, opened 2023-03-06
Minor bug fixes for MySQL binlog replication

fulghum:
Fixes a nil pointer dereference when `filters` is set to `nil` after `RESET REPLICA ALL`. Adds a test for that case and a test for handling very long hostnames.


max-hoffman (approved):
lgtm! sorry for the delay here, got caught up with a couple things

PR #5502, opened 2023-03-07
support staged set of changes for `dolt stash`

jennifersp:
- added support for handling staged set of changes for `dolt stash`.
- added support for `--include-untracked` flag for `dolt stash`.
- fixed untracked files being dropped when stashing working set of changes.


integration-tests/bats/stash.bats
```diff
@@ -352,3 +349,306 @@ teardown() {
...
+
+    run dolt stash
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Saved working directory and index state" ]] || false
+
+    run dolt sql -q "SELECT * FROM test"
+    [ "$status" -eq 0 ]
+    [ "$output" = "" ]
+
+    dolt gc
```
fulghum, line 550: oh! very nice! was GC already smart enough to follow all the datasets in the database and see that the stashed rootvalues are still in use? If you create a dolt commit after the stash, does that change anything? (without knowing anything about our gc internals... I'm just wondering if we're getting lucky with that rootvalue being the most recent one created or something)

fulghum (approved):
Very cool to see support for staged changes, too! 🙌  Code looks good; just some minor notes and a question or two. 

Two more really small requests while we're in here that came up while I was doing some manual testing: 
- could you add a new line between the output of `dolt status` and the `Dropped refs/stash@{0} (pop8lc59hqjegi7p1r2v8gd51ujaijii)`? Since status has newlines in it between the sections, I think the CLI output will read a little better that way. 
- For this error message, can we change this to mention the conflicting table names? I think that would help people more easily figure out how to resolve the issue: `table with same name added in 2 commits can't be merged`

Great work on this! 😁

andy-wm-arthur (approved):
LGTM!

PR #5503, opened 2023-03-07
Display DDL for views and triggers in dolt diff

macneale4:
Second attempt of botched PR: https://github.com/dolthub/dolt/pull/5476


fulghum (approved):
Looks good! 🚢

PR #5512, opened 2023-03-07
[no-release-notes] Ensure all dolt sql-server child processes get cleaned up

fulghum:
`dolt sql-server` processes were not getting properly cleaned up by simply killing the top level process. This change switches to kill all the child processes in the process group. I'm unable to repro the lingering `dolt sql-server` processes with this change.


reltuk (approved):
LGTM!

PR #5514, opened 2023-03-08
fix keyless duplicate update

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/5442 and https://github.com/dolthub/dolt/issues/5433

tests: https://github.com/dolthub/go-mysql-server/pull/1642

### DIFF
--- a/go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
+++ b/go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
@@ -160,14 +160,16 @@ func (k prollyKeylessWriter) uniqueKeyError(ctx context.Context, keyStr string, 
 		vd := k.valBld.Desc
 		for from := range k.valMap {
 			to := k.valMap.MapOrdinal(from)
-			if existing[to], err = index.GetField(ctx, vd, from, value, k.mut.NodeStore()); err != nil {
-				return err
-			}
+				// always skip first field of keyless, as it's the count?
+				if existing[to], err = index.GetField(ctx, vd, from+1, value, k.mut.NodeStore()); err != nil {
+					return err
+				}
 		}
 		return
 	})
 
 	return sql.NewUniqueKeyErr(keyStr, isPk, existing)
+	// TODO: skip over the ordinals that are the unique column
 }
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
```diff
@@ -160,7 +160,8 @@ func (k prollyKeylessWriter) uniqueKeyError(ctx context.Context, keyStr string,
 		vd := k.valBld.Desc
 		for from := range k.valMap {
 			to := k.valMap.MapOrdinal(from)
-			if existing[to], err = index.GetField(ctx, vd, from, value, k.mut.NodeStore()); err != nil {
+			// offset from index for keyless rows, as first field is the count
+			if existing[to], err = index.GetField(ctx, vd, from+1, value, k.mut.NodeStore()); err != nil {
```
max-hoffman, line 164: it might be better for valMap to just have the right mappings, embed the cardinality offset so that other methods don't have to add 1 everywhere

max-hoffman (approved):
lgtm, small suggestion so that we don't have to add a cardinality offset every time we use the ordinality mapping

go/libraries/doltcore/sqle/writer/prolly_index_writer_keyless.go
JCOR11599: That was my initial approach, but I don't think the `valMap` or the tuple descriptors can tell if the first field is encoding the count or a column of type uint64. Maybe it's worth modifying them? or changing `MapOrdinal` to take a keyless bool or something like that?

max-hoffman: something like this?
```go
func ordinalMappingsFromSchema(from sql.Schema, to schema.Schema) (km, vm val.OrdinalMapping) {
	keyless := schema.IsKeyless(to)
	km = makeOrdinalMapping(from, to.GetPKCols(), keyless)
	vm = makeOrdinalMapping(from, to.GetNonPKCols(), keyless)
	return
}
```

PR #5527, opened 2023-03-08
make dolt_patch() stored procedure into dolt_patch() table function

jennifersp:
This PR removes `dolt_patch()` stored procedure and adds `dolt_patch()` table function.
The `dolt_patch()` table function supports:
- `WORKING`, `STAGED`, `HEAD` and any valid commit hash revisions.
- three dot diff for sql patch statements.
- `schema` or `data` only statements can be queried with `WHERE` clause of `diff_type` column.
- has `statement_order` auto_increment integer value as primary key.
- includes hashes for from and to revisions

Added few bats tests for `dolt stash` and updated its synopsis and long descriptions.


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -757,55 +757,60 @@ func (d *DoltSession) GetRoots(ctx *sql.Context, dbName string) (doltdb.Roots, b
 // ResolveRootForRef returns the root value for the ref given, which refers to either a commit spec or is one of the
 // special identifiers |WORKING| or |STAGED|
 // Returns the root value associated with the identifier given and its commit time
-func (d *DoltSession) ResolveRootForRef(ctx *sql.Context, dbName, hashStr string) (*doltdb.RootValue, *types.Timestamp, error) {
+func (d *DoltSession) ResolveRootForRef(ctx *sql.Context, dbName, hashStr string) (*doltdb.RootValue, *types.Timestamp, string, error) {
```
fulghum, line 760: Is the returned hash string value always going to be the same as the `hashStr` input param? From reading this code, it seems like `hashStr` always needs to be a hash string (or the special `WORKING` or `STAGED` values). I'm guessing that this is actually being used to take in branch names or tags, too? 

If that's the case... then I think it's worth updating the `hashStr` variable name to `ref` or something so that the code is more self-describing (otherwise a reader would assume that `hashStr` has to always be a hash string).

fulghum (approved):
Code looks really great! ✨ Just minor notes from me. Great work turning this around so fast!

PR #5533, opened 2023-03-09
go/doltcore/sqle/index: use Map.Get() for point lookups

andy-wm-arthur:


andy-wm-arthur:
#benchmark

andy-wm-arthur:
#benchmark

andy-wm-arthur:
#benchmark

max-hoffman:
The merge benchmarks are ~30% slower with this change, presumably because this reverts one of the optimizations james made to speedup child fk lookups.

max-hoffman:
#systab-benchmark

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -169,31 +168,14 @@ func prollyChildSecDiffFkConstraintViolations(
 				}
 			}
 
-			if parentSecIdxCur == nil {
```
max-hoffman, line 172: Hit to merge perf is kind of painful. Can we hide this logic in a function in prolly to maintain the private cursor and get the rest of changes through?

max-hoffman (approved):
LGTM, my only concern is the merge perf regression on child fk lookups

andy-wm-arthur:
#merge-benchmark

go/libraries/doltcore/merge/violations_fk_prolly.go
andy-wm-arthur: addressed this [098d40a](https://github.com/dolthub/dolt/pull/5533/commits/098d40a44f669dee5508b7a9309d30c686934da4) with `prolly.Map.HasPrefix()`

PR #5542, opened 2023-03-09
Use auth host for `RPCCreds` audience, use dolthub.com for issuer

tbantle22:


go/libraries/doltcore/env/grpc_dial_provider.go
```diff
@@ -116,7 +116,9 @@ func (p GRPCDialProvider) getRPCCreds() (credentials.PerRPCCredentials, error) {
 	if !valid {
 		return nil, nil
 	}
-	return dCreds.RPCCreds(), nil
+
```
reltuk, line 119: I think this is a private method that's being used just about (`config.WithEnvCreds` branch...), and I think we should be passing the `endpoint` we've determined from the grpcendpoint.Config which was provided to us...

go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -138,17 +138,18 @@ func NewController(lgr *logrus.Logger, cfg Config, pCfg config.ReadWriteConfig)
 
 	keyID := creds.PubKeyToKID(ret.pub)
 	keyIDStr := creds.B32CredsEncoding.EncodeToString(keyID)
+	audience := "dolthub-remote-api.liquidata.co" // TODO
```
reltuk, line 141: This is going to have to change to be a factory like `grpcCreds func(aud string) *creds.RPCCreds` or something --- we need to sign outbound requests differently depending on which replica we connect to.

go/cmd/dolt/commands/credcmds/check.go
```diff
@@ -82,27 +82,27 @@ func (cmd CheckCmd) Exec(ctx context.Context, commandStr string, args []string,
...
 
 	return commands.HandleVErrAndExitCode(verr, usage)
 }
 
-func loadEndpoint(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) string {
+func loadEndpoint(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) (string, string) {
 	earg, ok := apr.GetValue("endpoint")
 	if ok {
-		return earg
+		return "", earg // TODO
```
tbantle22, line 100: Should I parse the endpoint here to get the host?

go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -97,6 +97,8 @@ type procedurestore interface {
 const (
 	DoltClusterRoleVariable      = "dolt_cluster_role"
 	DoltClusterRoleEpochVariable = "dolt_cluster_role_epoch"
+	// Since we fetch the keys from the other replicas we’re going to use a fixed string here.
```
tbantle22, line 100: Tried to regurgitate when you said here but it might not be right lol

go/cmd/dolt/commands/credcmds/check.go
reltuk: Yeah, I think it's OK to do that in this case. We can also drop any error from the Parse and use a default audience if we want to...

go/libraries/doltcore/sqle/cluster/controller.go
reltuk: Think this is fine...

reltuk (approved):
Generally looks great to me.

PR #5546, opened 2023-03-09
go/doltcore/diff: refactor diff splitter to handle columns unmatched across revisions

andy-wm-arthur:


go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -25,13 +25,29 @@ import (
...
-	toFrom        map[int]int
-	fromLen       int
+	// results schema of dolt_diff(...)
+	// sql table function
+	querySch sql.Schema
+	// output schema for CLI diff
+	targetSch sql.Schema
+	// maps querySch fields to targetSch
+	queryToTarget []int
+	// maps "to_..." and "from_..." columns
```
zachmu, line 45: Might comment a bit more about the contents of this slice

go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -40,128 +56,172 @@ type RowDiff struct {
...
 
-		resultToTarget[i] = targetIdx
+	qtt, err := mapQuerySchemaToTargetSchema(querySch, targetSch)
+	if err != nil {
+		return nil, err
 	}
 
-	if fromLen == -1 {
-		fromLen = len(diffQuerySch) - 1
+	// not all "from..." columns will have a "to..." column and vice-versa
```
zachmu, line 76: This comment probably belongs on either the struct or on the method

go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -40,128 +56,172 @@ type RowDiff struct {
...
-		oldRow.Row = make(sql.Row, len(ds.targetSch))
-		if diffTypeStr == "modified" {
-			oldRow.RowDiff = ModifiedOld
-		} else {
-			oldRow.RowDiff = Removed
+	query = query[:len(query)-1]
+
+	mapping = make([]int, len(query))
+	for i, col := range query {
+		switch {
```
zachmu, line 124: Nit: this kind of conditional logic is better expressed as `if ... else if .. else`

go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -40,128 +56,172 @@ type RowDiff struct {
...
+func mapToAndFromColumns(query sql.Schema) (mapping []int, err error) {
+	last := query[len(query)-1]
+	if last.Name != "diff_type" {
+		return nil, errors.New("expected last diff column to be 'diff_type'")
+	}
+	query = query[:len(query)-1]
+
+	mapping = make([]int, len(query))
+	for i, col := range query {
+		switch {
```
zachmu, line 150: Same comment here

Much easier to understand mapping algorithm tho

go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -25,13 +25,29 @@ import (
...
+	// sql table function
+	querySch sql.Schema
+	// output schema for CLI diff
+	targetSch sql.Schema
+	// maps querySch fields to targetSch
+	queryToTarget []int
+	// maps "to_..." and "from_..." columns
+	// to each other and vice-versa.
+	// NB: not all columns have matches
+	queryToQuery []int
```
zachmu, line 48: This doesn't seem to be used at all in fact

go/libraries/doltcore/diff/diffsplitter.go
```diff
@@ -40,128 +56,172 @@ type RowDiff struct {
...
+		for i, col := range ds.targetSch {
+			cmp, err = col.Type.Compare(from.Row[i], to.Row[i])
+			if err != nil {
+				return RowDiff{}, RowDiff{}, err
+			} else if cmp != 0 {
+				from.ColDiffs[i] = ModifiedOld
+				to.ColDiffs[i] = ModifiedNew
 			} else {
-				newRow.ColDiffs[ds.queryToTarget[i]] = Added
+				from.ColDiffs[i] = None
```
zachmu, line 216: Isn't this the zero value?

zachmu (approved):
LGTM!

go/libraries/doltcore/diff/diffsplitter.go
andy-wm-arthur: yes, but I thought it better to be explicit

PR #5547, opened 2023-03-10
dolt checkout table restores foreign key collection

jennifersp:


go/libraries/doltcore/env/actions/checkout.go
```diff
@@ -49,14 +49,21 @@ func MoveTablesFromHeadToWorking(ctx context.Context, roots doltdb.Roots, tbls [
...
 			return doltdb.Roots{}, err
 		}
 
 		if !ok {
 			tbl, ok, err = roots.Head.GetTable(ctx, tblName)
+			if err != nil {
+				return doltdb.Roots{}, err
+			}
 
+			fkc, err = roots.Staged.GetForeignKeyCollection(ctx)
```
fulghum, line 66: Can you reuse the `fkc` data from line 55 and avoid calling `GetForeignKeyCollection` again?

integration-tests/bats/checkout.bats
```diff
@@ -104,6 +104,41 @@ SQL
...
+    [[ "$output" =~ "Changes not staged for commit:" ]] || false
+    [[ "$output" =~ "modified:       z" ]] || false
+
+    run dolt checkout z
+    [ "$status" -eq 0 ]
+    [ "$output" = "" ]
+
+    run dolt status
+    [[ "$output" =~ "On branch main" ]] || false
+    [[ "$output" =~ "nothing to commit, working tree clean" ]] || false
```
fulghum, line 123: (minor) might be worth an explicit call to `show create table` or something and grep that there is or is not the `foreign_key1` constraint name – it would just be a very direct test of the foreign key being there (or not). I'm guessing that before this change the `dolt checkout z` call either failed, or left the change in the working set, so this test makes sense, but just might be more obvious with the direct foreign key check for future readers.

fulghum (approved):
Nice fix!

PR #5548, opened 2023-03-10
make ordinal mapping better

JCOR11599:


timsehn:
@JCOR11599 should we close this? Having PRs hang out for months is not a good look for us.

PR #5552, opened 2023-03-11
Sec index differ for parent fk constraint checking

max-hoffman:


max-hoffman:
#merge-benchmark

andy-wm-arthur (approved):
LGTM! nice perf win here

PR #5553, opened 2023-03-11
Fix incorrect recorded dolt commit times for MySQL replication

fulghum:
A few small code clean ups, and a bug fix for not setting unique times in dolt commits. 

Fixes: https://github.com/dolthub/dolt/issues/5563


zachmu (approved):
LGTM!

fulghum:
> Why does this have no release notes? Feels like we should advertise the binlog replication commit time fix?

I originally had a couple of non-customer facing changes in this PR, and just added the commit time fix without removing the `[no-release-notes]` hint. You are right that the last change makes this customer-facing though, so we should totally include these in the release notes. I'll fix it now.

PR #5555, opened 2023-03-11
use revision database name from current database for accessing DbData for `dolt_checkout()`

jennifersp:


integration-tests/bats/sql-checkout.bats
```diff
@@ -584,6 +584,19 @@ SQL
   [[ "$output" =~ "0" ]] || false
 }
 
+@test "sql-checkout: checkout new branch should use revision database name to access DbData on current database" {
+    skiponwindows "Need to install expect and make this script work on windows."
+    mkdir doltsql
+    cd doltsql
+
+    run expect $BATS_TEST_DIRNAME/sql-checkout.expect
```
fulghum, line 592: I'm wondering if there's a good way to avoid using `expect` here... It seems like it might actually be less lines (and one less file) if the commands were just here in the BATS `@test` method. (and maybe this test would also run on windows that way and not need a skip) It also lets us check the output of the tests more deeply, which we don't do with `expect`.

I know you mentioned having some problems getting all the commands to run in the same session with `dolt sql`, which makes sense. You'd have to shove them into the same `dolt sql -q '...'` line to get them to use the same session, which doesn't sound very usable. :-) 

Alternatively... how about testing this one with an engine ScriptTest as part of the Dolt ScriptTests? It seems like that would ensure that you are using the same session for the commands.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -43,13 +43,13 @@ func doltCheckout(ctx *sql.Context, args ...string) (sql.RowIter, error) {
 }
 
 func doDoltCheckout(ctx *sql.Context, args []string) (int, error) {
-	dbName := ctx.GetCurrentDatabase()
-
-	if len(dbName) == 0 {
-		return 1, fmt.Errorf("Empty database name.")
+	revDbName := ctx.GetCurrentDatabase()
+	if len(revDbName) == 0 {
+		return 1, fmt.Errorf("Empty database branchName.")
```
fulghum, line 48: `revDbName` seems a little misleading here, since the current database doesn't have to be a revision database. Similar comments for the error message string mentioning `database branchName`.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -64,13 +64,14 @@ func doDoltCheckout(ctx *sql.Context, args []string) (int, error) {
 		return 1, errors.New("Improper usage.")
 	}
 
-	// Checking out new branch.
 	dSess := dsess.DSessFromSess(ctx.Session)
-	dbData, ok := dSess.GetDbData(ctx, dbName)
+	// dbData should use the current database data, which can be at revision database.
+	dbData, ok := dSess.GetDbData(ctx, revDbName)
```
fulghum, line 69: Ahh!! I see the bug now. Even for when we were using a revision db, we always only grabbed the DbData for the main database. Makes sense. 

I think the `revDbName` could be a little clearer though... instead of calling it `revDbName` (which sometimes it will be a revDb, sometimes it won't be) maybe there's something better like `currentDatabase` that would be a little clearer?

fulghum (commented):
Nice job finding the bug for this one! The fix looks correct to me, I just had a couple comments on how we test it and some of the variable naming.

fulghum (approved):
Beautiful fix! 💎

PR #5556, opened 2023-03-11
Invalidate all sessions' db state for `drop database`

fulghum:
Fix for: https://github.com/dolthub/dolt/issues/5349

When a database is deleted, the cached dbState held by every session needs to be invalidated. Otherwise sessions think the deleted database still exists and are can still reference the deleted db's data, which does not match MySQL's behavior. If a new database is then created with the same name, then sessions will still use the old db info and aren't able to see the new database state.

This change means that a session's `dbState` member now needs to be accessed concurrently, so I've also added synchronization using the exiting session mutex. 

Related GMS PR: https://github.com/dolthub/go-mysql-server/pull/1654


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -698,6 +706,45 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
...
+	return p.invalidateDbStateInAllSessions(ctx, name)
+}
+
+// invalidateDbStateInAllSessions removes the db state for this database from every session. This is necessary when a
+// database is dropped, so that other sessions don't use stale db state.
+func (p DoltDatabaseProvider) invalidateDbStateInAllSessions(ctx *sql.Context, name string) error {
+	// TODO: this might be a better fit as a function on SessionManager
+	// TODO: Instead of deleting the db state, could we just mark it as invalid and continue using that same instance
+	//       in memory, even when we create a new db with the same name? reusing for creation of a new db might be tricky.
+	runningServer := sqlserver.GetRunningServer()
```
reltuk, line 719: This global is super spooky! Can we thread the server through when we construct the database provider for the engine? Definitely not related to this PR per se, but seeing this global makes me want to immediately go try to remove it...

fulghum: Well... this turned out to be a little messier than I was expecting... It looks like we're always using `DoltDatabaseProvider` as a struct for some reason, and not as a pointer to a struct. The `DoltDatabaseProvider` struct gets created when we create the engine and then passed (as a copy) into GMS. After that, the `sqlEngine` is instantiated and we can start the sql-server at which point we get the server reference that we want to set in the provider, but we can't change the provider registered in GMS, because a copy was passed in, and we don't have a reference to that memory. 

I'm happy to change `DoltDatabaseProvider` to be used as a pointer (seems like a good change to make for several reasons), but it feels like that might be cleaner in a separate PR since it'll be more than just a couple real quick changes like I thought this was going to be.

reltuk: Thanks for the explanation and investigating this. I agree this doesn't need to be addressed here :).

go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -698,6 +706,35 @@ func (p DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
...
+		sessionManager := runningServer.SessionManager()
+		err := sessionManager.Iter(func(session sql.Session) (bool, error) {
+			dsess, ok := session.(*dsess.DoltSession)
+			if !ok {
+				return false, fmt.Errorf("unexpected session type: %T", session)
+			}
+
+			// We need to invalidate this database state for EVERY session, even if other sessions aren't actively
+			// using this database, since they could still reference it with a db-qualified table name.
+			err := dsess.RemoveDbState(ctx, name)
```
zachmu, line 727: This is all a bit crunchy. Probably fine to check in for now, but it seems like this is a sign of missing abstractions on the GMS side.

Historically the `Session` interface was really underdefined for how much it was responsible for, and this kind of thing is a similar situation. We really want to engine to orchestrate dropping the database, including things like 1) preventing new connections to that DB once the drop operation begins, 2) cleanly invalidating any sessions / transactions that have that DB in scope. Might need to define a SessionProvider / SessionManager interface for this to work cleanly.

fulghum: Totally agree. I briefly thought about pushing some of this down to SessionManager (seems appropriate), but the state is at the Dolt layer of course, so it needed a callback hook and felt like the scope was increasing. You're right that it definitely smells like the abstraction is not ideal here.

PR #5562, opened 2023-03-13
[no-release-notes] skip on duplicate key update tests on old format

JCOR11599:


max-hoffman (approved):
lgtm, w/ prepareds is preferrable

PR #5564, opened 2023-03-13
`dolt show` command

solipsis-project:
Am initial implementation of the `dolt show` command.

There's room for further work, but this is a working first iteration.

Some limitations:
- Merge commits can't be shown.
- show.go calls methods defined in log.go and diff.go. We probably want to move those methods into a common file/package for commands that display info.
- There's some minor code duplication between show.go and log.go and diff.go.
- There may be more efficient ways to implement this, but I went for what would be easiest to implement as a first pass.
- It can't show a diff for the initial commit (but `dolt init` makes an empty initial commit, so unless someone is editing their history this shouldn't matter.)

This likely warrants integration tests, but I'm still learning BATS. Writing those tests can be my next step.


fulghum:
Very cool contribution @solipsis-project! I'll get started reviewing this one later this afternoon for ya.

proto/dolt/services/eventsapi/v1alpha1/event_constants.proto
```diff
@@ -90,6 +90,7 @@ enum ClientEventType {
     STASH_DROP = 58;
     STASH_LIST = 59;
     STASH_POP = 60;
+	SHOW = 61;
```
fulghum, line 93: (minor) looks like a little formatting weirdness. Can you check this out and see if it's just an errant tab messing up the indent?

go/cmd/dolt/commands/show.go
```diff
@@ -0,0 +1,265 @@
...
+	ShortDesc: `Show information about a specific commit`,
+	LongDesc:  `Show information about a specific commit`,
+	Synopsis: []string{
+		`[{{.LessThan}}revision{{.GreaterThan}}]`,
+	},
+}
+
+type ShowCmd struct{}
+
+// Name is returns the name of the Dolt cli command. This is what is used on the command line to invoke the command
```
fulghum, line 51:
```suggestion
// Name returns the name of the Dolt cli command. This is what is used on the command line to invoke the command
```

go/cmd/dolt/commands/show.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+
+	return nil
+}
+
+func showCommit(ctx context.Context, dEnv *env.DoltEnv, opts *showOpts, commitSpec *doltdb.CommitSpec) error {
+
+	comm, err := dEnv.DoltDB.Resolve(ctx, commitSpec, dEnv.RepoStateReader().CWBHeadRef())
+
```
fulghum, line 168: (minor/nit) Could you remove the newlines between the statement returning an `err` and the `err` checking if block? (lines 168, 175, and 198) Most of the codebase follows that style and it just makes the error handling feel a little more connected to the code that generated the error.

fulghum (commented):
This is looking great @solipsis-project. Thank you for contributing this! 🙏 

I only had a couple very minor comments. I think the only thing missing before merging is getting some BATS tests in place. I'm happy to chat about that over Discord if you wanna chat about anything there.

proto/dolt/services/eventsapi/v1alpha1/event_constants.proto
solipsis-project: Not sure what happened that, but it explains why the protobuf CI check is failing for this PR. It wasn't obvious to me because of the tab width in my editor.

solipsis-project:
I added some bats tests to cover the basic functionality, but these don't cover the flags that are inherited from the diff and log commands.

Unrelated, show is still calling methods defined in log and diff. I should move those into their own unit.

integration-tests/bats/show.bats
```diff
@@ -0,0 +1,86 @@
...
+    
+    dolt sql -q 'insert into testtable values (4)'
+    dolt add .
+    dolt commit -m "commit: add values"
+    
+	run dolt show
+	[ $status -eq 0 ]
+	[[ "$output" =~ "commit: add values" ]] || false
+	[[ "$output" =~ "|   | pk |" ]] || false
+	[[ "$output" =~ "| + | 4  |" ]] || false
```
fulghum, line 85: (minor/whitespace)

```suggestion
    run dolt show
    [ $status -eq 0 ]
    [[ "$output" =~ "commit: add values" ]] || false
    [[ "$output" =~ "|   | pk |" ]] || false
    [[ "$output" =~ "| + | 4  |" ]] || false
```

fulghum (approved):
Thanks for iterating on the feedback and adding some BATS tests! This looks good! 🚢

PR #5567, opened 2023-03-13
adds the dolt_column_diff system table

stephkyou:
This change adds a new system table, `dolt_column_diff` which lists the tables and columns changed in every commit reachable from the active branch’s HEAD.

fixes: https://github.com/dolthub/dolt/issues/5254


go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
...
+	defer d.incrementColIndex(tableDelta, staged)
+
+	change, err := tableDelta.GetSummary(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var colName string
+	if tableDelta.FromSch != nil {
+		colName = tableDelta.FromSch.GetAllCols().GetColumnNames()[d.colIndex]
```
jennifersp, line 266: I think getting modified column names here is not correct. I'm getting some of the unmodified column names as well, in this case, the column was dropped. Here is a test case on it:
```
dolt sql -q "create table test (id int primary key, c1 int, c2 int, c3 int);"
dolt commit -Am "add table"
dolt sql -q "alter table test drop column c2;"
dolt sql -q "select commit_hash, table_name, column_name, data_change, schema_change from dolt_column_diff;"
+----------------------------------+------------+-------------+-------------+---------------+
| commit_hash                      | table_name | column_name | data_change | schema_change |
+----------------------------------+------------+-------------+-------------+---------------+
| WORKING                          | test       | id          | false       | true          |
| WORKING                          | test       | c1          | false       | true          |
| WORKING                          | test       | c2          | false       | true          |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | id          | false       | true          |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c1          | false       | true          |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c2          | false       | true          |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c3          | false       | true          |
+----------------------------------+------------+-------------+-------------+---------------+
```

This gave me an idea that it might be helpful to have information about schema changes on columns such as `add`, `drop` and `change/modify` column. A way to add this information can be making `schema_change` column to have enum(`added`,`dropped`,`modified`) values instead of bool value option. Let me know what you think 🙂

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
+// Copyright 2022 Dolthub, Inc.
```
jennifersp, line 1:
```suggestion
// Copyright 2023 Dolthub, Inc.
```

go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
```diff
@@ -3677,6 +3677,96 @@ var UnscopedDiffSystemTableScriptTests = []queries.ScriptTest{
 	},
 }
 
+var ColumnDiffSystemTableScriptTests = []queries.ScriptTest{
```
jennifersp, line 3680: Would be great to add more tests on cases such as 'add column', 'drop column', 'modify column', etc. I like to use the existing tests on similar features as template to track different test cases I am not aware of and some edge cases as well 😁

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		return &tableColChange{
+			tableName:    delta.FromName,
+			colNames:     delta.FromSch.GetAllCols().GetColumnNames(),
+			dataChange:   !isEmpty,
+			schemaChange: true,
+		}, nil
```
jennifersp, line 489: small suggestion: drop, rename and add cases share very similar code with some repeated code, so this might be good part to be extracted out into its own function that all these cases can call.

jennifersp (changes requested):
Good start! 😄
I'm wondering if we can reuse or combine some of the functions of this feature with dolt_diff's as they are using a lot of common codes.

I have one suggestion on this feature: to add information about whether the column was `added` or `dropped` or `modified` on `schema_changed` column of this system table. 
An example result would be:
```
dolt sql -q "create table test (id int primary key, c1 int, c2 int, c3 int);"
dolt commit -Am "add table"
dolt sql -q "alter table test drop column c2;"
dolt sql -q "select commit_hash, table_name, column_name, data_change, schema_change from dolt_column_diff;"
+----------------------------------+------------+-------------+-------------+---------------+
| commit_hash                      | table_name | column_name | data_change | schema_change |
+----------------------------------+------------+-------------+-------------+---------------+
| WORKING                          | test       | c2          | false       | dropped       |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | id          | false       | added         |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c1          | false       | added         |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c2          | false       | added         |
| 2geog1nqcirna32l93nvnpf4vm078bbe | test       | c3          | false       | added         |
+----------------------------------+------------+-------------+-------------+---------------+
```

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
...
+
+	return &tableColChange{
+		tableName:    delta.ToName,
+		colNames:     colNames,
+		dataChange:   dataChanged,
+		schemaChange: schemaChanged,
+	}, nil
+}
+
+func calculateColDelta(ctx *sql.Context, delta diff.TableDelta, iter sql.RowIter, diffTableSchema schema.Schema) ([]string, error) {
```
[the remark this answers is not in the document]
stephkyou: The point of getting the projections is to get the column names for the rows the iterator is returning. I think we can you're right that we might be able to get the column names from somewhere else, but I'm not confident that they'll be in the same order as the cells in the row. I'll experiment to see if it works.

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
...
+
+	for _, filter := range dt.partitionFilters {
+		ri = plan.NewFilterIter(filter, ri)
+	}
+
+	return ri, nil
+}
+
+// incrementColIndex increments the column index and table changes index.  When the end of the column names array is
+// reached, moves to the next table changes delta.
```
fulghum, line 220: Good use of comments and godoc!

go/libraries/doltcore/sqle/dtables/column_diff_table.go
fulghum: Along the same line... Taylor recently added some nice correctness fixes to `TableDelta`. You might be able to simplify this a little bit if you can reuse functions in `TableDelta`, such as `HasDataChange()` and `HasSchemaChange()`, but I haven't looked through the details enough to see if that'll fully work out.

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,649 @@
...
+
+		return &tableColChange{
+			tableName:    delta.ToName,
+			colNames:     delta.ToSch.GetAllCols().GetColumnNames(),
+			dataChange:   !isEmpty,
+			schemaChange: true,
+		}, nil
+	}
+
+	dataChanged, err := delta.HasHashChanged()
```
fulghum, line 522: For this one, it would be good to use `delta.HasDataChanged()`. Taylor found a correctness bug here the other day... the hash for the table includes the schema, so if only the schema changed, then this should also show as true. The `HasDataChanged()` function she added should give you the correct result here.

go/libraries/doltcore/sqle/dtables/diff_table.go
```diff
@@ -944,7 +944,7 @@ func CalculateDiffSchema(fromSch, toSch schema.Schema) (schema.Schema, error) {
 
 	j := toSch.GetAllCols().Size()
 	err = fromSch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
-		fromCol, err := schema.NewColumnWithTypeInfo(diff.FromColNamer(col.Name), uint64(i), col.TypeInfo, false, col.Default, false, col.Comment)
+		fromCol, err := schema.NewColumnWithTypeInfo(diff.FromColNamer(col.Name), uint64(j), col.TypeInfo, false, col.Default, false, col.Comment)
```
jennifersp, line 947: Great catch! 🙌

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,646 @@
...
+			}
+			return resultColNames, resultDiffTypes, nil
+		}
+	}
+}
+
+// calculateColSchemaDiff calculates which columns were modified, added, or dropped between to and from schemas and
+// returns a list of column names for each type of change, the total list of column names, and a corresponding list of
+// diff_types for each column
+func calculateColSchemaDiff(toCols *schema.ColCollection, fromCols *schema.ColCollection) ([]string, []string, []string, []string, []string) {
```
jennifersp, line 598: Great godocs! It's an important one to inform about which returned value corresponds to with column array especially if there are many returned values of the same type. (small suggestion, but up to your preference 🙂) you can also make small struct with added, dropped, modified and all columns array of strings to return, then it can be easily accessed with each members.

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,656 @@
+// Copyright 2022 Dolthub, Inc.
```
jennifersp, line 1: I think this change got reset.

```suggestion
// Copyright 2023 Dolthub, Inc.
```

go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
```diff
@@ -3701,6 +3701,740 @@ var UnscopedDiffSystemTableScriptTests = []queries.ScriptTest{
...
+				Expected: []sql.Row{{2}},
+			},
+			{
+				Query:    "select table_name, column_name, diff_type from dolt_column_diff where commit_hash = @Commit2;",
+				Expected: []sql.Row{{"t", "d", "added"}},
+			},
+		},
+	},
+	{
+		Name: "add column - working set",
```
jennifersp, line 3832: These tests are great! 🙌

jennifersp (approved):
These changes looks great! 😄

go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -0,0 +1,656 @@
...
+			}
+			return resultColNames, resultDiffTypes, nil
+		}
+	}
+}
+
+// ColSchemaDiff is a collection of column names that hold the results of doing a schema diff between to/from schemas,
+// i.e. a list of column names for each type of change, the total list of column names, and a corresponding list of
+// diff_types for each column
+type ColSchemaDiff struct {
```
fulghum, line 592: (minor) Try to avoid exporting (i.e. capitalizing) types unless you have to. It's nice to keep them package protected by default. In this case, `ColSchemaDiff` looks like it's only used in this file, so you could rename it to `colSchemaDiff` to keep it visible only in the current package.

fulghum (approved):
Nice work on this Steph! 🙌

PR #5580, opened 2023-03-14
Fun commits

solipsis-project:
Fixes https://github.com/dolthub/dolt/issues/4015

I discussed in the issue a potential means of separating out the "fun" logic from the business logic. Let me know if you think that should be done before merging.

### DIFF
--- a/go/cmd/dolt/commands/init.go
+++ b/go/cmd/dolt/commands/init.go
@@ -34,6 +34,7 @@ const (
 	initBranchParamName = "initial-branch"
 	newFormatFlag       = "new-format"
 	oldFormatFlag       = "old-format"
+	requiresFunHashName = "fun"
 )
 
 var initDocs = cli.CommandDocumentationContent{
@@ -79,6 +80,7 @@ func (cmd InitCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(initBranchParamName, "b", "branch", fmt.Sprintf("The branch name used to initialize this database. If not provided will be taken from {{.EmphasisLeft}}%s{{.EmphasisRight}} in the global config. If unset, the default initialized branch will be named '%s'.", env.InitBranchName, env.DefaultInitBranch))
 	ap.SupportsFlag(newFormatFlag, "", fmt.Sprintf("Specify this flag to use the new storage format (%s).", types.Format_DOLT.VersionString()))
 	ap.SupportsFlag(oldFormatFlag, "", fmt.Sprintf("Specify this flag to use the old storage format (%s).", types.Format_LD_1.VersionString()))
+	ap.SupportsFlag(requiresFunHashName, "", "")
 	return ap
 }
 
--- a/go/cmd/dolt/commands/init.go
+++ b/go/cmd/dolt/commands/init.go
@@ -153,7 +153,8 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		}
 	}
 
-	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t)
+	requiresFunHash := apr.Contains(requiresFunHashName)
+	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t, requiresFunHash)
 	if err != nil {
 		cli.PrintErrln(color.RedString("Failed to initialize directory as a data repo. %s", err.Error()))
 		return 1
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -1527,3 +1527,10 @@ func (ddb *DoltDB) GetStashRootAndHeadCommitAtIdx(ctx context.Context, idx int) 
 
 	return getStashAtIdx(ctx, ds, ddb.vrw, ddb.NodeStore(), idx)
 }
+
+var funRegExp = regexp.MustCompile("^d[o0][1l]t")
+
+func isFunHash(hashVal hash.Hash) bool {
+	hashString := hashVal.String()
+	return funRegExp.MatchString(hashString)
+}
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -187,18 +188,41 @@ func (ddb *DoltDB) WriteEmptyRepoWithCommitTimeAndDefaultBranch(
 		return err
 	}
 
-	cm, _ := datas.NewCommitMetaWithUserTS(name, email, "Initialize data repository", t)
+	var firstCommit *datas.Commit
+	for {
+		cm, _ := datas.NewCommitMetaWithUserTS(name, email, "Initialize data repository", t)
 
-	commitOpts := datas.CommitOptions{Meta: cm}
+		commitOpts := datas.CommitOptions{Meta: cm}
 
-	cb := ref.NewInternalRef(CreationBranch)
-	ds, err = ddb.db.GetDataset(ctx, cb.String())
+		cb := ref.NewInternalRef(CreationBranch)
+		ds, err = ddb.db.GetDataset(ctx, cb.String())
 
-	if err != nil {
-		return err
+		if err != nil {
+			return err
+		}
+
+		firstCommit, err = ddb.db.BuildNewCommit(ctx, ds, rv.nomsValue(), commitOpts)
+
+		if err != nil {
+			return err
+		}
+
+		// --fun is a hidden command line option that rerolls the initial commit until
+		// it starts with "d0lt" or similar. We achieve this by modifying the timestamp,
+		// one millisecond at a time, until it produces a fun hash.
+		if !requireFunHash {
+			break
+		}
+
+		if isFunHash(firstCommit.Addr()) {
+			break
+		}
+
+		// The Time type uses nanosecond precision. Subtract one millino nanoseconds (one ms)
+		t = t.Add(-1_000_000)
 	}
 
-	firstCommit, err := ddb.db.Commit(ctx, ds, rv.nomsValue(), commitOpts)
+	firstCommitDs, err := ddb.db.WriteCommit(ctx, ds, firstCommit)
 
 	if err != nil {
 		return err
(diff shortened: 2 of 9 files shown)


integration-tests/bats/init.bats
```diff
@@ -297,6 +297,15 @@ teardown() {
     [[ "$output" =~ "__DOLT__" ]] || false
 }
 
+@test "init: fun flag produces an initial commit with the right hash" {
```
timsehn, line 300: Will this add 3-5 minutes to every bats run?

timsehn (commented):
Just one ask on bats

integration-tests/bats/init.bats
timsehn: Testing myself right now.

timsehn: That was very fast.

timsehn: Truly fun!
```
$ time dolt init --fun
dolt init --fun  0.15s user 0.07s system 14% cpu 1.512 total
$ dolt log
commit do1tnhqr6llq4o2kpoeb6n3idbm8j3du (HEAD -> main) 
Author: timsehn <tim@dolthub.com>
Date:  Tue Mar 14 15:48:16 -0700 2023

        Іnіtіаlіzе dаta repositоry

$ 
```

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -186,18 +188,72 @@ func (ddb *DoltDB) WriteEmptyRepoWithCommitTimeAndDefaultBranch(
...
+		{'i', '\u0456'},
+		{'o', '\u043e'},
+	}
+	attempt := 0
+	var firstCommit *datas.Commit
+	for {
+		generateInitialCommitDesc := func(attempt int) string {
+			// "Initialize data repository", with characters that could be Cyrillic replaced.
+			descFmt := "%cn%ct%c%cl%cz%c d%ct%c r%cp%cs%ct%cry"
+			choices := []any{}
```
fulghum, line 215: (minor go tip) Mostly just sharing this Go convention we use... the perf impact is very very small here, but it's still a good pattern to know. If we know the length of the slice (len of descriptionReplacementCandidates here) then we prefer to preallocate the capacity we need and not have to worry about the slice auto resizing while we're adding to it. Note that the middle param (`0`) is really important if you're using append to build the slice (but not what you want if you're going to walk through the index and set positions manually). 

```suggestion
			choices := make([]rune, 0, len(descriptionReplacementCandidates))
```

fulghum (commented):
Clever approach here with the alternate characters @solipsis-project! I took a quick scan through this and things are looking generally good. I left a couple really minor comments and I'll go back and take a deeper look tomorrow.

go/cmd/dolt/commands/init.go
```diff
@@ -151,7 +153,8 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		}
 	}
 
-	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t)
+	requiresFunHash := apr.Contains(requiresFunHashName)
+	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t, requiresFunHash)
```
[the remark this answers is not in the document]
solipsis-project: Trouble is, it's not just `InitRepoWithTime`.  `InitRepoWithTime` calls `InitDBAndRepoState` which calls `InitDBWithTime` which calls `WriteEmptyRepoWithCommitTime` which calls `WriteEmptyRepoWithCommitTimeAndDefaultBranch`. Each of these methods is visible outside the package and many of them have multiple callers that needed to be updated, and some of them contain logic that, if we make new functions, will either need to be duplicated or outlined.

I don't think there's going to be a tidy solution to this sadly (at least in a language without optional parameters). Having the extra argument and updating the callsites is not great, but I think trying to make new functions for all of them is worse because of how it would complicate/duplicate the logic in them.

Fortunately, it doesn't look like any of these functions are called outside of the doltcore library, so other libraries don't need to worry about this in any case.

go/libraries/doltcore/doltdb/doltdb.go
solipsis-project: I refactored it. Note that the array has to be of type []any, not []rune, because go arrays are invariant and Sprintf expects an []any.

go/libraries/doltcore/env/environment.go
```diff
@@ -431,8 +431,8 @@ func (dEnv *DoltEnv) configureRepo(doltDir string) error {
 
 // Inits the dolt DB of this environment with an empty commit at the time given and writes default docs to disk.
 // Writes new repo state with a main branch and current root hash.
```
fulghum, line 433: (minor) would be nice to mention the `|requiresFunHash|` param, especially since this is part of the public API for this type. We don't always do a great job with godocs for functions and with describing params, but we're trying to be more thorough there.

go/cmd/dolt/commands/init.go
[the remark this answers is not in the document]
fulghum: Ahh... yes... I followed this a little closer and see what you mean. It would require forking a bunch of calls and I think you're right that it wouldn't necessarily end up making things simpler. Thank you for taking the time to think about the idea!

fulghum (commented):
This is looking great @solipsis-project! Thanks for all the changes from the feedback. 

I just had a couple more really small requests, and then I'm ready to approve and merge this in! 😁

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -186,18 +188,72 @@ func (ddb *DoltDB) WriteEmptyRepoWithCommitTimeAndDefaultBranch(
...
+	for {
+		generateInitialCommitDesc := func(attempt int) string {
+			// "Initialize data repository", with characters that could be Cyrillic replaced.
+			descFmt := "%cn%ct%c%cl%cz%c d%ct%c r%cp%cs%ct%cry"
+			choices := make([]any, 0, len(descriptionReplacementCandidates))
+			for i := 0; i < len(descriptionReplacementCandidates); i++ {
+				choices = append(choices, descriptionReplacementCandidates[i][(attempt>>i)%2])
+			}
+			return fmt.Sprintf(descFmt, choices...)
+		}
```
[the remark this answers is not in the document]
solipsis-project: I'll do you one better.

After reflecting on the problem, I found a way to pull ALL the "fun hash" code into its own file, and also not change any of the public APIs. What we have now is significantly more readable.

go/libraries/doltcore/env/environment.go
solipsis-project: `requiresFunHash` param no longer exists.

solipsis-project:
I made a significant refactor that was able to remove ALL the "fun hash" logic out of doltdb.go, and also avoid changing any of the existing APIs (at the cost of a few new methods.)

I feel like this is much more readable. Let me know what you think.

go/store/datas/fun_commit_hash.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	description := fmt.Sprintf(descFmt, choices...)
+
+	g.attempt += 1
+
+	return NewCommitMetaWithUserTS(g.name, g.email, description, g.timestamp)
+}
+
+func (g *funHashCommitMetaGenerator) IsGoodCommit(commit *Commit) bool {
+	var funRegExp = regexp.MustCompile("^d[o0][1l]t")
```
fulghum, line 72: (minor) Recompiling this regex many thousands of times is going to be wasteful. Generally it's good practice to have regex's compiled once and reused. In this case, you could compile it as part of the `MakeFunCommitMetaGenerator` construction function and then reuse it. That way, it would also not be compiled if it's going to be used (as opposed to using an `init` function for example).

go/cmd/dolt/commands/init.go
```diff
@@ -154,7 +155,15 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
 
 	requiresFunHash := apr.Contains(funHashFlag)
-	err := dEnv.InitRepoWithTime(context.Background(), types.Format_Default, name, email, initBranch, t, requiresFunHash)
+	commitMeta := func() datas.CommitMetaGenerator {
+		if requiresFunHash {
+			return datas.MakeFunCommitMetaGenerator(name, email, t)
+		} else {
+			return datas.MakeCommitMetaGenerator(name, email, t)
+		}
+	}()
```
fulghum, line 164: This could be simplified a bit... no need to create the anonymous function just to call it immediately...

```suggestion
	generator := datas.MakeCommitMetaGenerator(name, email, t)
	if requiresFunHash {
		generator = datas.MakeFunCommitMetaGenerator(name, email, t)
	}
```

go/cmd/dolt/commands/init.go
```diff
@@ -154,7 +155,15 @@ func (cmd InitCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+
+	err := dEnv.InitRepoWithCommitMeta(context.Background(), types.Format_Default, initBranch, commitMeta)
```
fulghum, line 166: Ohhh... I like this new approach a lot! 

Just one suggestion... how about renaming the function to: `InitRepoWithCommitMetaGenerator`. I think that's a better, more self-descriptive name because there is another type called `CommitMeta`, so I initially assumed this function was taking in a `CommitMeta` instance (plus the var was named `commitMeta`). Using `CommitMetaGenerator` would make it a little clearer and easier to understand quickly.

fulghum (commented):
Nice work on this @solipsis-project! Thank you for taking the time to think through the function signatures and interfaces and figure out a way to make this change less invasive to the existing APIs! I really like the new interface you created – it's a clever and clean way to solve this. 👏 

I just had a couple very small comments. I think this one is good to merge after that!

go/cmd/dolt/commands/init.go
solipsis-project: I did the anonymous function because it avoids calling `MakeCommitMetaGenerator` if it doesn't have to. In practice, this code will only run once and the performance impact is so minor that it really doesn't matter. If you feel like your suggestion is more readable, then lets go with that.

go/store/datas/fun_commit_hash.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	description := fmt.Sprintf(descFmt, choices...)
+
+	g.attempt += 1
+
+	return NewCommitMetaWithUserTS(g.name, g.email, description, g.timestamp)
+}
+
+func (g *funHashCommitMetaGenerator) IsGoodCommit(commit *Commit) bool {
+
```
fulghum, line 79:
```suggestion
```

fulghum (approved):
Very cool contribution @solipsis-project! 🙌  Thank you for all the time, effort, and thought you put in. 🙏  It was fun working with you on this one.

PR #5583, opened 2023-03-15
Bug fix for view and trigger diffs in SQL and JSON diff outputs

zachmu:


go/cmd/dolt/commands/diff.go
```diff
@@ -603,17 +615,15 @@ func diffUserTable(
...
 	dArgs *diffArgs,
 	dw diffWriter,
 ) errhand.VerboseError {
-	err := dw.BeginTable(sqlCtx, td)
-	if err != nil {
-		return errhand.VerboseErrorFromError(err)
-	}
-
-	query := fmt.Sprintf("select from_fragment,to_fragment from dolt_diff('%s','%s','%s')", dArgs.fromRef, dArgs.toRef, doltdb.SchemasTableName)
+	// TODO: does this work when the table has been deleted in the current revision?
```
macneale4, line 622: More context on this comment. Why would the current revision have any impact on this. fromRef and toRef should work regardless of what HEAD is, yes? Also, If you deleted the table in the toRef (HEAD or otherwise) it works. There is a test for the dolt_diff method which shows this.

go/cmd/dolt/commands/diff_output.go
```diff
@@ -266,6 +271,18 @@ func (t tabularDiffWriter) WriteSchemaDiff(ctx context.Context, toRoot *doltdb.R
 	return nil
 }
 
+func (t tabularDiffWriter) WriteTriggerDiff(ctx context.Context, triggerName, oldDefn, newDefn string) error {
+	diffString := textdiff.LineDiff(oldDefn, newDefn)
+	cli.Println(diffString)
+	return nil
+}
+
+func (t tabularDiffWriter) WriteViewDiff(ctx context.Context, viewName, oldDefn, newDefn string) error {
```
macneale4, line 280: Identical impl to WriteTriggerDiff. Perhaps just call it

go/cmd/dolt/commands/diff_output.go
```diff
@@ -390,9 +450,138 @@ func (j *jsonDiffWriter) RowWriter(ctx context.Context, td diff.TableDelta, unio
...
+	}
+
+	j.triggersWritten++
+	return nil
+}
+
+func (j *jsonDiffWriter) WriteViewDiff(ctx context.Context, viewName, oldDefn, newDefn string) error {
+	if j.viewsWritten == 0 {
+		// begin the document if necessary
+		if j.tablesWritten == 0 && j.triggersWritten == 0 {
```
macneale4, line 511: This seems frail. I'd add a method which adds the "{" char once and only once. Then you can call it whenever you think you may need to start the document.

go/cmd/dolt/commands/diff_output.go
```diff
@@ -390,9 +450,138 @@ func (j *jsonDiffWriter) RowWriter(ctx context.Context, td diff.TableDelta, unio
...
+			_, err := j.wr.Write([]byte("{"))
+			if err != nil {
+				return err
+			}
+		}
+
+		// end the previous block if necessary
+		if j.tablesWritten > 0 && j.triggersWritten == 0 {
+			_, err := j.wr.Write([]byte(jsonDataDiffFooter + ","))
```
macneale4, line 520: Again, this seems frail. Maybe we shouldn't be building this up as a stream, but as objects in a hierarchy. Then you can wrap junks in brackets and join them all together with ',' and the separator.

macneale4 (approved):
Left some comments, but nothing worth blocking for

go/cmd/dolt/commands/diff.go
zachmu: I meant the `to` revision, but it as a just a note to myself, I'll delete

go/cmd/dolt/commands/diff_output.go
zachmu: Introduced a method for this logic.

go/cmd/dolt/commands/diff_output.go
zachmu: It's a little better with the beginDocument method, but still pretty frail. Not going to restructure it that much right now though.

PR #5584, opened 2023-03-15
libraries/doltcore/doltdb: fix dropped errors

alrs:
This fixes two dropped `err` variables in the `libraries/doltcore/doltdb`.


fulghum (approved):
Thanks for the contribution @alrs! 😄

PR #5597, opened 2023-03-16
Bug fix: binlog replication no longer assumes there is a `root` user

fulghum:
The binlog replication applier needs to execute SQL statements that modify any database or table being replicated from the primary. We were previously assuming the `root` user existed and had the right permissions for this, but that is not always true (e.g. Hosted Dolt does not create a `root` user on Dolt DBs). Since we can't rely on an existing account with the right privileges for executing those replicated SQL statements, we now create a custom `dolt-binlog-applier` user that has super user privs and is locked, to disable logins. 

This change fixes the correctness issue in https://github.com/dolthub/dolt/issues/5592 – if a `root` user doesn't exist, binlog replication will still work correctly. 

If this direction looks good, I'll add some deeper testing around this.


Hydrocharged (commented):
I think this is the correct direction to go with this. MySQL has a few locked users with special permissions, and so we're adding our own locked user for binlog rep in this case.

PR #5601, opened 2023-03-17
go/store/nbs: make chunk journal lock NBS manifest while open

andy-wm-arthur:


go/store/nbs/file_manifest_test.go
```diff
@@ -77,6 +79,7 @@ func TestFileManifestLoadIfExists(t *testing.T) {
 }
 
 func TestFileManifestLoadIfExistsHoldsLock(t *testing.T) {
+	t.Skip("TODO")
```
andy-wm-arthur, line 82: this test fails now that we don't hold the lock while reading

reltuk: Remove it? Its name seems to imply it's testing for behavior we never intend to implement.

go/store/nbs/journal.go
```diff
@@ -407,6 +419,117 @@ func (c journalConjoiner) chooseConjoinees(upstream []tableSpec) (conjoinees, ke
...
+		lock, err = nil, nil // read only
+	} else if err != nil {
+		return nil, err
+	}
+	m = &journalManifest{dir: dir, lock: lock}
+
+	var f *os.File
+	f, err = openIfExists(filepath.Join(dir, manifestFileName))
+	if err != nil {
+		return nil, err
```
reltuk, line 438: I think we want to unlock m.lock here and in other error paths. Thoughts?

integration-tests/bats/sql-client.bats
```diff
@@ -37,7 +37,7 @@ setup() {
 }
 
 teardown() {
-    stop_sql_server
+    stop_sql_server 1 && sleep 0.5
```
reltuk, line 40: Nit: Can we `wait` the pid we killed, or is this send metrics nonsense? A comment about why?

reltuk (approved):
LGTM! A few suggestions.

integration-tests/bats/sql-client.bats
andy-wm-arthur: I think this was `send-metrics` nonsense, and is now fixed

PR #5602, opened 2023-03-17
Added test for avoiding multiple parens issue.

macneale4:
https://github.com/dolthub/dolt/issues/5478

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -170,7 +170,8 @@ var ShowCreateTableScriptTests = []queries.ScriptTest{
                                    b int not null default 42,
                                    c int not null default (24),
                                    d int not null default '-108',
-                                   e int not null default ((((7+11)))));`,
+                                   e int not null default ((((7+11)))),
+                                   f int default (now()))`,
 			`call dolt_commit('-Am', 'new table');`,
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -90,76 +90,111 @@ var ViewsWithAsOfScriptTest = queries.ScriptTest{
...
+						"  UNIQUE KEY `unique_c2` (`c2`)\n" +
+						") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin",
+					},
 				},
 			},
 		},
 	},
+	{
+		// "https://github.com/dolthub/dolt/issues/5478"
+		Name: "show table for default types with unique indexes",
```
max-hoffman, line 167: Keyless and unique keyless indexes have distinct implementations. We can ship even if those fail, but it would be useful to have skipped tests in that case.

max-hoffman: here are the function defaults tested in `TestColumnDefaults`
https://github.com/dolthub/go-mysql-server/blob/6ed5b50a67bec31be01a2263d43c1eff67b35321/enginetest/enginetests.go#L5954

max-hoffman (approved):
lgtm, a couple suggests for additional tests

PR #5604, opened 2023-03-20
[no-release-notes] Refactored db / session initialization logic 

zachmu:
This is preparatory to catching multi-db transaction errors


fulghum (approved):
Pretty squeaky clean. I couldn't find much. 🚢

PR #5606, opened 2023-03-20
go/store/nbs: store.go: Conjoin if we need to when we add new table files to a manifest.

reltuk:
This allows us to avoid huge manifests with lots of table files when we run multiple full GCs over a period of time.


andy-wm-arthur (approved):
LGTM!

PR #5610, opened 2023-03-21
User/password authentication for dolt clone

tbantle22:


go/cmd/dolt/commands/clone.go
```diff
@@ -228,3 +232,14 @@ func validateAndParseDolthubUrl(urlStr string) (string, bool) {
 
 	return "", false
 }
+
+func getUserAndPassConfig(apr *argparser.ArgParseResults) *creds.DoltCredsForPass {
```
reltuk, line 236: I see we will support this for `clone` for now, with `pull/fetch` to come later probably?

I think I would make this error --- if `apr.Contains(cli.UserParam)` but `os.Getenv("DOLT_REMOTE_PASSWORD") == nil`, the user's environment is probably setup incorrectly and we could give them a more useful error than just an auth failure against the remote.

go/libraries/doltcore/creds/creds.go
```diff
@@ -177,3 +194,14 @@ func (dc DoltCreds) RPCCreds(audience string) *RPCCreds {
 		RequireTLS: false,
 	}
 }
+
+func (dcp DoltCredsForPass) ToBase64Str() string {
+	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", dcp.Username, dcp.Password)))
+}
+
+func (dc DoltCredsForPass) RPCCreds() *RPCCreds {
```
reltuk, line 202: How deep does the upstream dependency on `*RPCCreds` go? I might prefer to just make `DoltCredsForPass` implement `GetRequestMetadata(ctx context.Context, uri ...string)`, for example. I'm not sure the modal behavior on `*RPCCreds` as implemented now is the best, if we can just separate the two functions.

go/libraries/doltcore/remotesrv/interceptors.go
```diff
@@ -0,0 +1,98 @@
...
+			si.Lgr.Info("incoming request had malformed authentication header")
+			return status.Error(codes.Unauthenticated, "unauthenticated")
+		}
+		authTrim := strings.TrimPrefix(auth, "Basic ")
+		uDec, err := base64.URLEncoding.DecodeString(authTrim)
+		if err != nil {
+			si.Lgr.Infof("incoming request authorization header failed to decode: %v", err)
+			return status.Error(codes.Unauthenticated, "unauthenticated")
+		}
+		compare := subtle.ConstantTimeCompare(uDec, []byte(fmt.Sprintf("%s:%s", si.ExpectedUserAuth.User, si.ExpectedUserAuth.Password)))
```
reltuk, line 89: Nit: I would encourage us to pre-render this comparison byte array.

go/libraries/doltcore/sqle/remotesrv.go
```diff
@@ -60,9 +60,14 @@ func (s remotesrvStore) Get(path, nbfVerStr string) (remotesrv.RemoteSrvStore, e
 	return rss, nil
 }
 
-func RemoteSrvServerArgs(ctx *sql.Context, args remotesrv.ServerArgs) remotesrv.ServerArgs {
+func RemoteSrvServerArgs(ctx *sql.Context, args remotesrv.ServerArgs, userAuth *remotesrv.UserAuth) remotesrv.ServerArgs {
```
reltuk, line 63: Nit: I think I would encourage something like:

```
func WithUserPasswordAuth(args remotesrv.ServerArgs, userAuth *remotesrv.UserAuth) remotesrv.ServerArgs
```

and leaving this function alone, since it doesn't touch `args.Options` as is.

I think with that structure, `ServerArgs.ServerInterceptor` could be left off. The server interceptor struct could be public and the `WithUserPasswordAuth` function could just construct an interceptor with an appropriate value based on the passed in `UserAuth` pointer.

Feel free to push back if this makes it noticeably worse, but it seems right to me.

integration-tests/bats/sql-server-remotesrv.bats
```diff
@@ -156,3 +156,59 @@ select count(*) from vals;
...
+    export DOLT_REMOTE_PASSWORD="pass0"
+
+    dolt sql-server -u $DOLT_REMOTE_USER  -p $DOLT_REMOTE_PASSWORD --remotesapi-port 50051 &
+    srv_pid=$!
+
+    cd ../
+    run dolt clone http://localhost:50051/remote repo1
+    [[ "$status" != 0 ]] || false
+    [[ "$output" =~ "Unauthenticated" ]] || false
+}
```
reltuk, line 214: Nit: Can we add one more test where we pass the correct user with bad password and assert Unauthenticated as well.

reltuk (commented):
Looks really good. Just a few comments which shouldn't take long to address.

PR #5611, opened 2023-03-21
Bug fixes for branch deletion edge cases

fulghum:
While debugging a customer issue, I found a few edge cases around deleting branches to fix:
- the `sql-server.lock` file wasn't getting created when creating the first database for a sql-server (i.e. when running the sql-server from a bare, non-init'ed directory). 
- `dolt checkout` on the CLI was exiting too early if it couldn't find its working set and not actually checking anything out, but returning a success error code. This meant customers couldn't use the CLI to recover from a deleted default branch. We did have a test for this, but it wasn't testing thoroughly enough and only looked for a success return code from `dolt checkout`. 
- `dolt branch -d` (from SQL or CLI) should not allow you to delete the repo's default branch. This is how the customer got their database into a bad state with out a default branch. 
- a couple tests in `deleted-branches.bats` were skipped; I removed the skips and confirmed they pass.

These fixes will help prevent customers from getting into this state and ensure they can fix a deleted default branch using the dolt CLI. It would be ideal to continue on this thread and also make `dolt_checkout()` work in sql-server with the default branch is invalid, but I think that's best handled as a follow up PR.


integration-tests/bats/sql-branch.bats
```diff
@@ -162,3 +162,13 @@ SQL
     run dolt sql -q "CALL DOLT_BRANCH('-D', 'branch_with_unpushed_commit');"
     [ $status -eq 0 ]
 }
+
+@test "sql-branch: CALL DOLT_BRANCH -d error cases" {
```
zachmu, line 166: Do we need an equivalent test for the sql command? What should the behavior be there?

zachmu (approved):
LGTM, just one comment

integration-tests/bats/sql-branch.bats
fulghum: Good catch. I added a new BATS test for the `dolt branch -D` to assert that you cannot delete the checked out branch. _(I know you said for the sql command, but I generalized to make sure we have all the branch delete codepaths covered.)_

Now we've got the following cases covered:
- dolt sql shell ([sql-branch.bats](https://github.com/dolthub/dolt/blob/ae23057590ce6a27e615e2d51a6a7669a5541b9a/integration-tests/bats/sql-branch.bats#L166))
- dolt sql-server connection ([sql-server.bats](https://github.com/dolthub/dolt/blob/ae23057590ce6a27e615e2d51a6a7669a5541b9a/integration-tests/bats/sql-server.bats#L575))
- dolt CLI ([branch.bats](https://github.com/dolthub/dolt/blob/40965c6d7a7544354f6bf5f312cad491ea6661ca/integration-tests/bats/branch.bats#L142))

PR #5615, opened 2023-03-22
Allow logging base64-encoded queries, instead of replacing bad chars.

PavelSafronov:


PavelSafronov:
Matching GMS PR: https://github.com/dolthub/go-mysql-server/pull/1667

zachmu (approved):
LG

PR #5617, opened 2023-03-22
go/libraries/doltcore/sqle/cluster: Improve the robustness of graceful transitions to standby role.

reltuk:
Some subtle races could cause graceful transitions to fail when they could have succeeded or could cause replication to block for a long time while the attempt to transition was ongoing.


andy-wm-arthur (approved):
LGTM!

PR #5620, opened 2023-03-23
[no-release-notes] go/doltcore/merge: spec tests for schema merge

andy-wm-arthur:


go/libraries/doltcore/merge/schema_merge_test.go
```diff
@@ -15,68 +15,569 @@
...
+	t.Run("column reordering tests", func(t *testing.T) {
+		testSchemaMerge(t, columnReorderingTests)
 	})
 	t.Run("primary key change tests", func(t *testing.T) {
-		testSchemaMerge(t, colMappingTests)
+		testSchemaMerge(t, keyChangeTests)
+	})
+	t.Run("secondary index tests", func(t *testing.T) {
+		testSchemaMerge(t, secondaryIndexTests)
+	})
```
fulghum, line 75: (minor) seems like we want to run all the test suite with multiple secondary indexes, to make sure all those cases work and that we exercise the multiple secondary index codepaths. I tried out the `validateIndexes` function you pointed me at last Friday and it triggered as expected, so I'm wondering if we can just bake multiple secondary indexes into all the tests instead of having a separate section of test cases for them.

fulghum (approved):
Nice test suite! Just one thought about how we should test secondary indexes. I'll pull this in and switch my testing over to this.

go/libraries/doltcore/merge/schema_merge_test.go
andy-wm-arthur: I think this makes a lot of sense. We can run multiple iterations of the test to validate the merge works with various secodary indexes. I'll add this in a follow up PR

PR #5624, opened 2023-03-24
Fix dolt_gc for repositories newly using chunk journal.

reltuk:
When a repository was created with < 0.75.0, or if it was created and used with `DOLT_DISABLE_CHUNK_JOURNAL=true`, it does not have a chunk journal. If it then runs dolt in a mode where it uses the chunk journal, the journal will not be bootstrapped until a write is done against the database. In the state where `dolt` is using the chunk journal but the chunk journal is not bootstrapped, `dolt gc` and `call dolt_gc()` were broken. They would fail with the message: `concurrent manifest edit during GC, before swapTables. GC failed.`.

This fixes that by making the chunk journal forward UpdateGCGen manifest update calls in its non-bootstrapped state, the same way it forwards Update manifest calls.

### DIFF
--- a/go/libraries/doltcore/dtestutils/sql_server_driver/cmd.go
+++ b/go/libraries/doltcore/dtestutils/sql_server_driver/cmd.go
@@ -178,6 +178,12 @@ func WithArgs(args ...string) SqlServerOpt {
 	}
 }
 
+func WithEnvs(envs ...string) SqlServerOpt {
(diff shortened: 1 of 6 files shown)


max-hoffman (approved):
LGTM, we talked offline about how safeguards around the journaling and GC lifecycle for messier edge cases might be helpful

PR #5626, opened 2023-03-24
Schema merge support for adding/dropping/renaming/reordering columns

fulghum:
First pass at getting schema merge working for column adds/drops/renames/reorders. Includes support for:
- migrating existing data in a primary index to the merged schema
- mapping changes from the right side, to the merged schema and merging them into primary and secondary indexes, as well as conflict and constraint violation detection.


go/libraries/doltcore/merge/merge_noms_rows.go
```diff
@@ -40,6 +42,76 @@ type rowMerger func(ctx context.Context, nbf *types.NomsBinFormat, sch schema.Sc
 
 type applicator func(ctx context.Context, sch schema.Schema, tableEditor editor.TableEditor, rowData types.Map, stats *MergeStats, change types.ValueChanged) error
 
+// mergeNomsTable merges a Noms table, which includes updating row data, secondary index data, and applying the specified |mergedSch|.
+func mergeNomsTable(ctx context.Context, tm *TableMerger, mergedSch schema.Schema, vrw types.ValueReadWriter, opts editor.Options) (*doltdb.Table, *MergeStats, error) {
```
fulghum, line 46: No code changes in this function; it was just extracted from `merge_rows.go` to keep the noms-specific code and prolly-specific code separate.

go/libraries/doltcore/merge/merge_rows.go
```diff
@@ -363,24 +273,6 @@ func (rm *RootMerger) maybeShortCircuit(ctx context.Context, tm TableMerger, opt
 	return nil, nil, nil
 }
 
-func (rm *RootMerger) maybeAbortDueToUnmergeableIndexes(tableName string, leftSchema, rightSchema, targetSchema schema.Schema) error {
```
fulghum, line 366: Moved to prolly specific code files

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -99,22 +164,26 @@ func mergeTableData(ctx context.Context, tm TableMerger, finalSch schema.Schema,
...
+			err = conflicts.merge(ctx, diff, nil)
 			if err != nil {
 				return nil, nil, err
 			}
+			// TODO: Why would we merge the secondary index data if we can't merge the primary index?
+			//       This doesn't seem correct... was this needed for unique index validation or something?
+			//err = sec.merge(ctx, diff, nil)
+			//if err != nil {
+			//	return nil, nil, err
+			//}
```
fulghum, line 179: @max-hoffman – I wanted to ping you about this code before I pulled it out. It seems like we shouldn't be merging into the secondary indexes here if we don't merge data into the primary index, but I wanted to double check that I'm not just missing something.

max-hoffman: Yeah, secMerger doesn't seem to do anything with the diff. It looks like i was a little more careful with erroring in the default case for most of the handlers

fulghum: sweet; thanks for the sanity check! just wanted to make sure I wasn't misunderstanding some details about the 3way differ.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -740,11 +868,113 @@ func newValueMerger(merged, leftSch, rightSch, baseSch schema.Schema, syncPool p
...
+
 // tryMerge performs a cell-wise merge given left, right, and base cell value
 // tuples. It returns the merged cell value tuple and a bool indicating if a
 // conflict occurred. tryMerge should only be called if left and right produce
 // non-identical diffs against base.
 func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
+	// We can't merge two divergent keyless rows
 	if m.keyless {
 		return nil, false
 	}
```
andy-wm-arthur, line 980: This is... not great. The cell-wise merge callback in `ThreeWayDiffer` is called when equal keys with unequal values are encountered. For keyless tables, this only occurs when the "cardinality" field of the value tuple has been changed on both sides. For example:
```
key:      val.Tuple(xxHash3({c1, c2, c3}))
ancestor: val.Tuple({cardinality=6, c1, c2, c3})
left:     val.Tuple({cardinality=4, c1, c2, c3})
right:    val.Tuple({cardinality=10, c1, c2, c3})
```
In this case, `left` deleted two copies of the keyless row and `right` added four copies. So the correct merged value tuple should be `val.Tuple({cardinality=8, c1, c2, c3})`

andy-wm-arthur (approved):
Generally looks great! One bug leftover from the last merge refactor

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -36,7 +36,72 @@ import (
...
+	stats.Conflicts = int(n)
+
+	mergeTbl, err = mergeAutoIncrementValues(ctx, tm.leftTbl, tm.rightTbl, mergeTbl)
+	if err != nil {
+		return nil, nil, err
+	}
+	return mergeTbl, stats, nil
+}
+
+// mergeProllyTableData three-way merges rows and indexes for a given table. First,
```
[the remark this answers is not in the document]
fulghum: good call! I just pushed commit 187db749db709236c7730f41f92509db1f6a5087 to update that function comment.

go/libraries/doltcore/merge/merge_prolly_rows.go
fulghum: ooh, good catch. I just added a new test case to trigger this and pushed a fix in a615e3aeef1f3587db10c5d10b119184d0e25640. Thanks for finding this one!

fulghum: Actually... I was playing with this one some more, and I think returning `nil, false` from `tryMerge` is actually correct here... 

In the case you described, it seems like this should still be a conflict, since we can't decide whether we should delete two rows, add four rows, or something in between, so it needs user involvement to resolve. I tested this with an LD format repo and a DOLT format repo and got that behavior. The [TestKeylessMergeConflicts](https://github.com/dolthub/dolt/blob/4998109a960b541f6037f39772da72c18f202311/go/libraries/doltcore/merge/keyless_integration_test.go#L133) also seem to encode this behavior. 

By returning `nil, false` from the `tryMerge` callback function (which should probably be renamed to `tryResolveMergeConflict` actually...) is called when the conflict is detected, so by not resolving it there, it gets tracked correctly as a conflict for a user to resolve manually. 

Let me know if I just completely misunderstood or misinterpreted the feedback.

PR #5649, opened 2023-03-29
updates sql-client to print query timing for more statement types

stephkyou:
fixes: https://github.com/dolthub/dolt/issues/5518

### DIFF
--- a/integration-tests/bats/sql-client.bats
+++ b/integration-tests/bats/sql-client.bats
@@ -58,11 +58,10 @@ teardown() {
     create_test_table
     run show_tables
     [ "$status" -eq 0 ]
-    echo $output
     [ "${lines[0]}" = '# Welcome to the Dolt MySQL client.' ]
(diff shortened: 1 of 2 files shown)


integration-tests/bats/sql-client.bats
```diff
@@ -58,14 +58,16 @@ teardown() {
     create_test_table
     run show_tables
     [ "$status" -eq 0 ]
+    echo $output
```
fulghum, line 61: (minor) you can pull out this debugging line. btw... I just recently learned (from Neil) about the `--verbose-run` flag when running bats tests locally – it automatically outputs `$output` for any `run` call. Pretty handy for debugging bats failures (locally at least), although I sometimes still use `echo $output` calls, especially if I can't repro something locally and need to see the output in the CI job. 

```suggestion
```

fulghum (approved):
Changes look great! Nice work on this one. ✨

PR #5653, opened 2023-03-29
Base64-encode binary data when creating SQL.

PavelSafronov:
Fixes https://github.com/dolthub/dolt/issues/5650


PavelSafronov:
It looks like MySQL's mysqldump outputs a similar binary format when it's outputting SQL. But MySQL is able to process a MySQL dump of binary data, whereas we are not able to process MySQL's binary output or our own.

So while this approach will allow us to generate printable SQL statements, and will allows us to ingest our own `dolt_patch`/`dolt diff` output, we may still want to update dolt to handle both MySQL's and our own binary dumps.

PR #5655, opened 2023-03-30
Add BATS test that dolt log correctly tags HEAD commits.

solipsis-project:
Regression test for https://github.com/dolthub/dolt/issues/5558


integration-tests/bats/log.bats
```diff
@@ -632,3 +632,15 @@ teardown() {
...
 }
+
+@test "log: identify HEAD" {
+    dolt commit --allow-empty -m "commit 1"
+    dolt tag commit1
+    dolt commit --allow-empty -m "commit 2"
+    dolt tag commit2
+    run dolt log commit1
+    [[ !("$output" =~ "HEAD") ]] || false
+    dolt log commit2 | cat
```
fulghum, line 643: Seems like you don't need this line and the following line that calls the same command with `run`

```suggestion
```

integration-tests/bats/log.bats
```diff
@@ -632,3 +632,15 @@ teardown() {
...
+    run dolt log commit2 | cat
```
fulghum, line 644: Do you need the `| cat`? `run` doesn't work well with pipes, so I'm guessing this isn't doing what it looks like:
https://bats-core.readthedocs.io/en/stable/writing-tests.html#run-and-pipes

fulghum (commented):
Thanks for sending this new test over! 🙏 Test looks mostly good; just a couple really minor comments. 

It looks like the CI test is still failing though:
https://github.com/dolthub/dolt/actions/runs/4559395128/jobs/8043303962?pr=5655#step:18:1243

From looking at the test failure... my first guess is that this branch doesn't have the log/head fix from your earlier PR. I can repro the behavior with an old dolt version, but can't repro it with the latest behavior, so seems like it's just a branch that's missing commits. This could have been easy to miss running the bats tests locally, if your `dolt` version was a newer build with the fix.

integration-tests/bats/log.bats
solipsis-project: Whoops. That was leftover from when I was debugging the test. Removed.

integration-tests/bats/log.bats
solipsis-project: That was leftover from when I was trying to get the test to suppress ANSI color codes, but I don't think we've implemented that yet. I just changed the test to not need it. Removed.

solipsis-project:
> From looking at the test failure... my first guess is that this branch doesn't have the log/head fix from your earlier PR. 

Thanks. I pulled from main and pushed to this branch, so it should have the fix in now.

PR #5667, opened 2023-03-31
go/libraries/doltcore/sqle/cluster: commithook.go: Fix initialization of commithook so it does not report isCaughtUp right after construction.

reltuk:
Misreporting isCaughtUp would race to allow a primary to gracefully transition to standby without true-ing up the replicas immediately after creating a new database, for example.


zachmu (approved):
LGTM!

PR #5670, opened 2023-03-31
Fix point lookup bug w/ prefix matching secondary

max-hoffman:
re: https://github.com/dolthub/dolt/issues/5648, test here: https://github.com/dolthub/go-mysql-server/pull/1685

### DIFF
--- a/go/store/prolly/tuple_map.go
+++ b/go/store/prolly/tuple_map.go
@@ -246,8 +246,8 @@ func (m Map) Get(ctx context.Context, key val.Tuple, cb tree.KeyValueFn[val.Tupl
 	return m.tuples.Get(ctx, key, cb)
 }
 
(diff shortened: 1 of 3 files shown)


go/store/prolly/tuple_map.go
```diff
@@ -246,6 +246,12 @@ func (m Map) Get(ctx context.Context, key val.Tuple, cb tree.KeyValueFn[val.Tupl
 	return m.tuples.Get(ctx, key, cb)
 }
 
+// GetPrefix searches for the key-value pair keyed by |key| and passes the results to the callback.
```
andy-wm-arthur, line 249: We should mention that this method is only valid for indexes where `prefDesc` corresponds to a set of unique columns

andy-wm-arthur (approved):
LGTM!

PR #5672, opened 2023-03-31
go/libraries/doltcore/doltdb: durable/table: Always write the table value we are refing. Ensures it always makes it across the GC safepoint.

reltuk:


reltuk:
#benchmark

andy-wm-arthur (approved):
LGTM!

PR #5697, opened 2023-04-04
[auto-bump] [no-release-notes] dependency by JCOR11599

coffeegoddd:
:coffee: *An Automated Dependency Version Bump PR* :crown:

### Initial Changes

The initial changes contained in this PR were produced by \`go get\`ing the dependency.

\`\`\`
\$ cd ./go
\$ go get github.com/dolthub/<dependency>/go@<commit>
\`\`\`

### Before Merging

This PR must have passing CI and a review before merging.

### DIFF
--- a/integration-tests/MySQLDockerfile
+++ b/integration-tests/MySQLDockerfile
@@ -121,7 +121,7 @@ RUN service postgresql start
 
 # install mysql_fdw
 WORKDIR /mysql-client-tests/mysql_fdw
-RUN git clone https://github.com/EnterpriseDB/mysql_fdw --branch REL-2_7_0
+RUN git clone https://github.com/EnterpriseDB/mysql_fdw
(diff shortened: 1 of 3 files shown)


integration-tests/MySQLDockerfile
```diff
@@ -121,7 +121,7 @@ RUN service postgresql start
 
 # install mysql_fdw
 WORKDIR /mysql-client-tests/mysql_fdw
-RUN git clone https://github.com/EnterpriseDB/mysql_fdw --branch REL-2_7_0
+RUN git clone https://github.com/EnterpriseDB/mysql_fdw
```
reltuk, line 124: I think we should still pick a release, or else the tests are non-deterministic and non-reproducible. I can see the argument for wanting to know if tip breaks, but it's better to pin dependencies in CI at least.

PR #5702, opened 2023-04-06
go/libraries/doltcore/env/actions: fix dropped error

alrs:
This fixes a dropped `err` variable in `go/libraries/doltcore/env/actions`.


fulghum:
It looks like the code is relying on **not** throwing that error in this case. I didn't dig in too deep, but noticed that CI tests were failing consistently and from looking at the code, it makes sense that we would be okay with ignoring errors there. 

If you'd like to change this PR to explicitly ignore the error response, with a comment explaining why, we could accept this one, but as is now, it looks like this will cause a regression in behavior.

PR #5705, opened 2023-04-06
sqle: track column renames on alter/modify column

andy-wm-arthur:
companion to: https://github.com/dolthub/go-mysql-server/pull/1702


PavelSafronov (approved):
Changes look good, approving.

PR #5708, opened 2023-04-06
[no-release-notes] got rid of the last uses of UpdateWorkingRoot, UpdateStagedRoot

zachmu:


go/libraries/doltcore/env/actions/reset.go
```diff
@@ -154,7 +154,17 @@ func ResetHard(
 		return err
 	}
```
andy-wm-arthur, line 156: in instances like this, it would be nice to pass less repo state in to the library function. We could resolve the current working branch outside of the `actions` package and avoid passing `dEnv` altogether

integration-tests/bats/sql-merge.bats
andy-wm-arthur, line 1: what's going on with these deleted tests?

andy-wm-arthur (approved):
seems reasonable, a couple comments

integration-tests/bats/sql-merge.bats
zachmu: They're redundant. We had tests of both `SELECT DOLT_MERGE()` and `CALL DOLT_MERGE()` when we supported both, but then at some point somebody converted all the tests to use CALL so we ended up with duplicates.

go/libraries/doltcore/env/actions/reset.go
zachmu: Yeah the actions package in general is kind of a mess, there's a ton of stuff there that's only called from CLI path. I'll do another pass to remove some of it.

PR #5717, opened 2023-04-08
support `EVENT` storage handling

jennifersp:
Depends on GMS PR: https://github.com/dolthub/go-mysql-server/pull/1701
The description of the GMS PR applies here as well that this PR implements events storage only (non-versioned metadata such as `last_executed` is not implemented yet). The event execution is not implemented yet.

The events are versioned and are stored on all branches. This means that checking out a new branch will have the copy of existing events of the starting point branch. The events versioned metadata is stored in `dolt_events` system table.


zachmu (commented):
This seems fine, but why the new table? Why not `dolt_schemas`?

go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
```diff
@@ -1343,6 +1343,49 @@ inner join t on to_pk = t.pk;`,
...
+			"CREATE TABLE messages (id INT PRIMARY KEY AUTO_INCREMENT, message VARCHAR(255) NOT NULL, created_at DATETIME NOT NULL);",
+			"CREATE EVENT IF NOT EXISTS msg_event ON SCHEDULE AT CURRENT_TIMESTAMP + INTERVAL 1 YEAR DISABLE DO INSERT INTO messages(message,created_at) VALUES('Test Dolt Event 1',NOW());",
+			"CREATE EVENT my_commit ON SCHEDULE EVERY 1 DAY DISABLE DO CALL DOLT_COMMIT('--allow-empty','-am','my daily commit');",
+			"CALL DOLT_ADD('.')",
+			"SET @Commit1 = '';",
+			"CALL DOLT_COMMIT_HASH_OUT(@Commit1, '-am', 'Creating table and events')",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query: "SELECT name, definer, execute_every, preserve, status, comment, definition FROM dolt_events;",
```
fulghum, line 1358: (minor) Looks like this needs the update to pull from `dolt_schemas`? Just double-checking that I understood where everything was stored.

fulghum (approved):
Looks good! 🚢  Just a couple really minor notes.

PR #5742, opened 2023-04-13
Update uses of go-mysql-server error objects to point to new package.

nicktobey:
The dolt side of https://github.com/dolthub/dolt/issues/5656.

This is needed for dolt to build against go-mysql-server at HEAD.


max-hoffman:
We aren't in the habit of doing reviews for automatic or small bumps, if tests pass go ahead and merge.

PR #5753, opened 2023-04-14
Add error when commands have extra arguments.

nicktobey:
Multiple commands would previously silently ignore extra positional arguments. This happened most frequently with commands that aren't supposed to take any positional arguments, but also occurred with some others, most notably merge, which caused https://github.com/dolthub/dolt/issues/5729.

This fixes that by requiring users of ArgParser to specify how many positional arguments they expect by calling either `NewArgParserWithMaxArguments(int)` or `NewArgParserWithVariableArguments()`. The error message can be customized by assigning to `ArgParser.TooManyArgsError`. The goal is that by requiring developers to be explicit about this, it will help avoid similar issues in the future.

This adds BATS tests for `dolt merge` in order to test that the new functionality works. I'm not going to add tests for every other command because there's like 80, but new commands should probably have a test for too many positional args.


nicktobey:
Note that this does change the observed behavior slightly for some commands by standardizing what previously had to be implemented separately in each command.

For instance, the decision of whether to print the full help string or just the usage string when a command has too many parameters. Previously some commands were printing the full help string for invalid flags and the usage string for too many flags. As written now, it's not possible to do this, and the behavior for both is the same (usually the full help string.)

PR #5762, opened 2023-04-18
Add error when commands have extra arguments. 

nicktobey:
Rolling this "Add error when commands have extra arguments." again.

`dolt sql-client` inherits its arg parser from `dolt sql-server`, which didn't play nice with the original PR's restriction on sql-server's positional arguments. I relaxed the arg parser and added an extra verification check into sql-server itself.


timsehn:
I'm testing this right now in Bats.

Please change the title to something w/o `Nicktobey/` as that shows up in the release notes.

timsehn:
This fails these two bats tests:

```
 ✗ sql-shell: inline query with missing -q flag should error
   (in test file integration-tests/bats/sql-shell.bats, line 823)
     `[[ "$output" =~ "Invalid Argument:" ]] || false' failed

@test "sql-shell: inline query with missing -q flag should error" {
    run dolt sql "SELECT * FROM test;"
    [ $status -eq 1 ]
    [[ "$output" =~ "Invalid Argument:" ]] || false
}
```

```
 ✗ remotes: dolt_clone procedure
   (in test file integration-tests/bats/remotes.bats, line 2247)
     `[[ "$output" =~ "error: invalid number of arguments" ]] || false' failed
   remotesrv log available here /var/folders/54/v92yp9tj6ll0j10w7rjg_3jc0000gn/T/remotes-88765/remotesrv.log
   Uploaded 382 B of 382 B @ 0 B/s.
   
   Switched to branch 'other'
   
   /Users/timsehn/dolthub/git/dolt/integration-tests/bats/remotes.bats: line 18: 88795 Terminated: 15          remotesrv --http-port 1234 --dir ./remotes-$$ >&./remotes-$$/remotesrv.log 3>&-  (wd: /var/folders/54/v92yp9tj6ll0j10w7rjg_3jc0000gn/T)

  run dolt sql -q "call dolt_clone('file://$tempDir/remote', 'foo', 'bar')"
  [ "$status" -eq 1 ]
  [[ "$output" =~ "error: invalid number of arguments" ]] || false
```

We need to update that error messages. I'm not sure what it is but it should be easy.

I also pushed a change where the old form of `dolt sql-client` caught some badly specified tests. I think I actually like that form better even though it will require fixing some badly specified bats test. I think we can merge this if it passes bats and then maybe go back and make `dolt sql-client` more strict like it was before you changed it in this version of this PR.

fulghum:
Looks like you need to merge main into your dev branch to resolve the conflicts GitHub lists. That will allow the CI checks to run on this PR.

nicktobey:
> I think I actually like that form better even though it will require fixing some badly specified bats test.

Agreed, but given that the plan is to deprecate `dolt sql-client` soon-ish, I decided it was better not to block this PR digging into the exact semantics of chaining multiple queries.

It also occurs to me that `run dolt sql -q "call dolt_clone('file://$tempDir/remote', 'foo', 'bar')"` has two different reasons that it might error here: either `dolt sql` or `dolt_clone` could have too many arguments, and we should make sure that the error message is able to distinguish between them.

go/libraries/utils/argparser/parser.go
```diff
@@ -49,15 +49,31 @@ func ValidatorFromStrList(paramName string, validStrList []string) ValidationFun
...
+	Name              string
+	MaxArgs           int
+	TooManyArgsError  func(receivedArgs []string) error
 	Supported         []*Option
 	nameOrAbbrevToOpt map[string]*Option
 	ArgListHelp       [][2]string
 }
 
-func NewArgParser() *ArgParser {
+func NewArgParserWithVariableArgs(name string) *ArgParser {
```
fulghum, line 60: (minor) Give public/exported functions godocs (as well as any non-simple private functions). This makes a big difference to other people reading your code. For example, from just looking at the call site, it's not obvious why `name` is required here. Add some godocs for these new public functions that explains what they do, their params, return vals, etc. Doesn't need to be anything fancy, but think about what will be most helpful to future code readers who don't know these implementations.

go/libraries/utils/argparser/parser.go
```diff
@@ -49,15 +49,31 @@ func ValidatorFromStrList(paramName string, validStrList []string) ValidationFun
...
+	return NewArgParserWithMaxArgs(name, -1)
+}
+
+func NewArgParserWithMaxArgs(name string, maxArgs int) *ArgParser {
+	tooManyArgsErrorGenerator := func(receivedArgs []string) error {
+		args := strings.Join(receivedArgs, ", ")
+		if maxArgs == 0 {
+			return fmt.Errorf("error: %s does not take positional arguments, but found %d: %s", name, len(receivedArgs), args)
+		}
+		return fmt.Errorf("error: %s has too many positional arguments. Expected at most %d, found %d: %s", name, maxArgs, len(receivedArgs), args)
```
fulghum, line 70: Good error messages! Specific and exact! 🙌

go/libraries/utils/argparser/parser.go
```diff
@@ -49,15 +49,31 @@ func ValidatorFromStrList(paramName string, validStrList []string) ValidationFun
 }
 
 type ArgParser struct {
+	Name              string
+	MaxArgs           int
+	TooManyArgsError  func(receivedArgs []string) error
```
fulghum, line 54: (minor/nit) I'd suggest naming this something like `TooManyArgsErrorFunc` – people are going to see `FooError` and naturally think "This thing is an error", so adding `Func` at the end seems a little more self-documenting.

integration-tests/bats/config.bats
```diff
@@ -208,7 +208,7 @@ teardown() {
     start_sql_server
     
     dolt sql-client --use-db '' -u dolt -P $PORT -q "create database testdb"
-    run dolt sql-client --use-db '' -u dolt -P $PORT -r csv -q "show databases"
+    run dolt sql-client --use-db '' -u dolt -P $PORT --result-format csv -q "show databases"
```
fulghum, line 211: Does `-r` not work anymore?

fulghum (commented):
Code changes look good, but I don't see any test coverage added that asserts the new positional arg parsing limits work. (Let me know if I just missed it.) It would be ideal to see the unit tests for ArgParser cover this, as well as a spot check or two in the BATS tests that assert the expected error messages are being returned.

integration-tests/bats/config.bats
timsehn: `-r` is overloaded now. I think because sql-client inherits args from sql-server as well? Not sure how that happened.

It now means `read-only` or `result-format`.

https://docs.dolthub.com/cli-reference/cli#dolt-sql-client

go/libraries/utils/argparser/parser.go
nicktobey: Fixed with https://github.com/dolthub/dolt/pull/5771

go/libraries/utils/argparser/parser.go
nicktobey: Fixed with https://github.com/dolthub/dolt/pull/5771

nicktobey:
I added a unit test in https://github.com/dolthub/dolt/pull/5771. Spot checks already existed in the original PR for several commands, here's an example of one of them being updated to use the final error wording: https://github.com/dolthub/dolt/pull/5762/commits/779de4804d2a4903685bb0d50b027aff5205ac87

PR #5767, opened 2023-04-18
go/doltcore/{doltdb, sqle}: dolt_schema_conflicts system table

andy-wm-arthur:
…m table


go/cmd/dolt/commands/cnfcmds/auto_resolve.go
```diff
@@ -63,26 +62,77 @@ func AutoResolveAll(ctx context.Context, dEnv *env.DoltEnv, strategy AutoResolve
...
+		err = ResolveDataConflicts(ctx, dEnv, ws, tblName, strategy)
+		if err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
-// ResolveTable resolves all conflicts in the given table according to the given
+func ResolveSchemaConflicts(ctx context.Context, ddb *doltdb.DoltDB, ws *doltdb.WorkingSet, tables []string, strategy AutoResolveStrategy) (*doltdb.WorkingSet, error) {
```
fulghum, line 91: (minor) might be worth putting this code into a shared package, if there's an easy one that won't cause an import cycle. It looked pretty compatible with the version in `dprocedures/dolt_conflicts_resolve.go`, but I didn't check every detail.

fulghum (commented):
Changes are looking good! Mostly just minor comments from me. The only big question I had was whether this is doing the right thing to migrate secondary index data. It seems like there's more work Dolt needs to do for that when an auto-resolution strategy is chosen. Let me know if I'm just missing something!

go/cmd/dolt/commands/cnfcmds/auto_resolve.go
```diff
@@ -63,26 +62,77 @@ func AutoResolveAll(ctx context.Context, dEnv *env.DoltEnv, strategy AutoResolve
...
+	if err != nil {
+		return nil, err
+	}
+
+	root := ws.WorkingRoot()
+	for name, tbl := range updates {
+		if root, err = root.PutTable(ctx, name, tbl); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
fulghum: Thanks for talking through this one with me. I was imagining a case where we had been able to merge some of the tables and update their data, but those roots would never be written back if we hit schema conflicts on other tables, so I think you're right that the auto resolution really is as simple as just taking the left or right side, without anything else needed.

fulghum (approved):
Thanks for talking through the partially merged data case with me... I poked back through the code and you're right... since the modified roots would never be written back if we hit any schema conflicts on later tables in the merge, there's no additional data to migrate or unmigrate. 🚢

go/cmd/dolt/commands/cnfcmds/auto_resolve.go
andy-wm-arthur: yeah I think we should, will follow up with this

PR #5768, opened 2023-04-18
Add Server Runtime details to the sql-server.lock file

macneale4:
Add runtime information for the server to the sql-server.lock file. We track three pieces of information:

1) The PID. As before.
2) The port of the server, if there is one. You can grab the lock as a cli process, which has no port.
3) A secret. Currently implemented as a UUID. This will be used to ensure that the connecting through the server is only performed by user's and processes which have read access to the server.

In addition, this change modifies the sql-server.lock file to be only readable by the owner.

Step (1) of:  https://github.com/dolthub/dolt/issues/3922.

### DIFF
--- a/go/gen/proto/dolt/services/eventsapi/v1alpha1/client_event.pb.go
+++ b/go/gen/proto/dolt/services/eventsapi/v1alpha1/client_event.pb.go
@@ -23,12 +23,13 @@
 package eventsapi
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	durationpb "google.golang.org/protobuf/types/known/durationpb"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
-	reflect "reflect"
-	sync "sync"
 )
 
 const (
--- a/go/libraries/doltcore/sqlserver/dolt_server.go
+++ b/go/libraries/doltcore/sqlserver/dolt_server.go
@@ -15,31 +15,55 @@
 package sqlserver
 
 import (
+	"fmt"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env"
 	"sync"
 
 	"github.com/dolthub/go-mysql-server/server"
 )
 
-var mySQLServer *server.Server
-var mySQLServerMutex sync.Mutex
+var lockedDetails *serverAndLockfile
+var mutex sync.Mutex
+
+// Struct for holding a *server.Server and a *env.DBLock
+type serverAndLockfile struct {
+	Server   *server.Server
+	Lockfile *env.DBLock
+}
 
 // RunningInServerMode returns true if the current process is running a SQL server.
 func RunningInServerMode() bool {
-	mySQLServerMutex.Lock()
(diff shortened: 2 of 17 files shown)


go/gen/proto/dolt/services/eventsapi/v1alpha1/client_event.pb.go
```diff
@@ -23,12 +23,13 @@
 package eventsapi
 
 import (
+	reflect "reflect"
+	sync "sync"
+
```
fulghum, line 28: Looks like the CI formatting checks are failing on this change, which is weird because it sure seems like a GH Action made this edit for you! I took a quick look at the GH Action logs, but didn't see anything obvious... I know Dustin was making some fixes there, so might just been that. Try reverting these formatting changes to the generated files and if the GH Actions apply this again, we can dig in and figure out why it's doing this (my first guess would be maybe it's formatting with an older go version; I think I saw something similar locally before).

macneale4: Rebasing history looks like it generated a new format change at the tip. Good to know that's a thing

go/libraries/doltcore/sqlserver/dolt_server.go
```diff
@@ -15,31 +15,56 @@
...
+
 	"github.com/dolthub/go-mysql-server/server"
 )
 
-var mySQLServer *server.Server
-var mySQLServerMutex sync.Mutex
+var lockedDetails *serverAndLockfile
+var mutex sync.Mutex
+
+// Struct for holding a *server.Server and a *env.DBLock
```
fulghum, line 29: (minor) godocs should start with the identifier (`serverAndLockfile` here) 

```suggestion
// serverAndLockfile holds a *server.Server and a *env.DBLock for a running server 
```

integration-tests/bats/sql-server.bats
```diff
@@ -1446,6 +1446,22 @@ databases:
     [ "$status" -eq 1 ]
 }
 
+@test "sql-server: sql-server sets permissions on sql-server.lock" {
```
fulghum, line 1449: (minor) Might be nice to have a test somewhere that asserts the contents of the lock file; we probably have a bats test that checks for port that could maybe be extended easily. 

Might also be worth having a smoke test for the legacy lock file format, just to ensure we don't blow up if we see one. Seems like we'll need to keep that functionality around until 1.0 probably?

fulghum (approved):
Code looks great! 🚢  Thank you for breaking this project down into such nicely sized chunks!

Before you merge... it looks like there is still some GitHub Actions weirdness... the latest commit on this PR doesn't seem to have had CI checks run for it, and I'm not sure why. I'd try pushing a trivial commit to the PR and see if that kicks it enough to get them to run and make sure everything is still passing.

integration-tests/bats/sql-server.bats
macneale4: This is actually pretty difficult to test legitimately. Since we don't have a previous version of the binary in the test, we can't produce a lock file in the old format (we could of course write one to disk manually).

We have several tests which validate that if we load a lock file which isn't understood that the binary blows it away and takes over. That's really only going to happen for us if they roll back their binary to a previous version. This all seems overboard for the amount of risk here.

PR #5772, opened 2023-04-19
Schema conflict detection edge cases

fulghum:
https://github.com/dolthub/dolt/issues/2973 pointed out an edge case where a schema conflict wasn't getting reported. That uncovered some other edge cases, so I refactored our column merge logic a bit to try and better organize it. 

Fixes: https://github.com/dolthub/dolt/issues/2973


go/libraries/doltcore/merge/schema_integration_test.go
```diff
@@ -290,13 +290,13 @@ var mergeSchemaConflictTests = []mergeSchemaConflictTest{
 			ColConflicts: []merge.ColConflict{
 				{
 					Kind:   merge.NameCollision,
-					Ours:   newColTypeInfo("c4", uint64(4696), typeinfo.Int32Type, false),
-					Theirs: newColTypeInfo("c4", uint64(8539), typeinfo.Int32Type, false),
+					Ours:   newColTypeInfo("C6", uint64(13258), typeinfo.Int32Type, false),
```
fulghum, line 293: New implementation returns these same two column conflicts, but in a different order, so that's why this test data is reordered.

andy-wm-arthur (approved):
LGTM!

PR #5773, opened 2023-04-19
Fixed a bug where foreign key changes in the working set weren't carried forward during `dolt checkout`

### LINKED ISSUE #1310: Foreign Keys don't travel with working set on `dolt checkout`
https://github.com/dolthub/dolt/blob/9c8ae0de518f2de0c44dcc697cc82d1c226d3cc1/bats/foreign-keys.bats#L1217

zachmu:
Fixes #1310


go/libraries/doltcore/env/actions/checkout.go
```diff
@@ -92,3 +97,539 @@ func MoveTablesFromHeadToWorking(ctx context.Context, roots doltdb.Roots, tbls [
...
+	if err != nil {
+		return nil, err
+	}
+
+	if oldHash == changedHash {
+		return newFks, nil
+	} else if oldHash == newHash {
+		return changedFks, nil
+	} else if newHash == changedHash {
+		return oldFks, nil
```
fulghum, line 467: This case didn't click with me... if the new root we're switching to has the same FK content as the working/staged root that was passed in, then we return the old FKs? From the earlier if case, we know that changedHash != oldHash. Seems odd to return the different `oldFks`.

go/libraries/doltcore/env/actions/checkout.go
```diff
@@ -92,3 +97,539 @@ func MoveTablesFromHeadToWorking(ctx context.Context, roots doltdb.Roots, tbls [
...
+		oldHash := doltdb.CombinedHash(oldFksForTable)
+		newHash := doltdb.CombinedHash(newFksForTable)
+		changedHash := doltdb.CombinedHash(changedFksForTable)
+
+		if oldHash == changedHash {
+			fksByTable[tblName] = append(fksByTable[tblName], newFksForTable...)
+		} else if oldHash == newHash {
+			fksByTable[tblName] = append(fksByTable[tblName], changedFksForTable...)
+		} else if newHash == changedHash {
+			fksByTable[tblName] = append(fksByTable[tblName], oldFksForTable...)
```
fulghum, line 509: Similar comment here... seems like if the working/staged set has the same FKs as the new root we're moving to (and they differ from the current HEAD root), then we'd want to return new/changed FKs.

I don't think the BATS tests would trigger this case where you move to a branch that has the same FKs as were added in the working set, so maybe adding that will confirm if this is the right code or not.

fulghum (approved):
Change makes sense. Just a couple minor comments and one larger question about some logic I didn't understand that I think may be missing a BATS test case.

integration-tests/bats/foreign-keys.bats
```diff
@@ -1358,12 +1358,114 @@ SQL
...
+SQL
+
+    dolt add . && dolt commit -m "new tables"
+    dolt sql -q "ALTER TABLE child2 ADD CONSTRAINT fk_v2 FOREIGN KEY (v1) REFERENCES parent2(v1);"
+
+    run dolt schema show child2
+    [ "$status" -eq "0" ]
+    [[ "$output" =~ "fk_v2" ]] || false
+
+    dolt checkout main
```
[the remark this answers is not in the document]
zachmu: This was an error, good catch. Already fixed in a later revision

go/libraries/doltcore/env/actions/checkout.go
zachmu: It's unreachable, good catch. Doesn't make sense, and would be a bug if it were reachable.

go/libraries/doltcore/env/actions/checkout.go
zachmu: Same deal.

PR #5787, opened 2023-04-21
Run schema merge tests in both directions

fulghum:
Running the `ThreeWayMergeWithSchemaChangeTestScripts` tests in both directions has been pretty helpful to catch a few bugs, so I took a stab at making `testSchemaMerge` run merge tests in both directions. 

There shouldn't be any changes to the test cases, but it does help remove duplication in the test cases so that they are a little easier to maintain.


andy-wm-arthur (approved):
LGTM!

PR #5788, opened 2023-04-21
Adding basic support for conflict validation in new schema merge test framework

fulghum:
This also pointed out an issue in the framework where any secondary index and check constraint definitions are silently swallowed.

### DIFF
--- a/go/libraries/doltcore/merge/schema_merge_test.go
+++ b/go/libraries/doltcore/merge/schema_merge_test.go
@@ -504,26 +504,45 @@ var simpleConflictTests = []schemaMergeTest{
 		conflict: true,
 	},
 	{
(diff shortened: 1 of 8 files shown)


go/libraries/doltcore/merge/schema_merge_test.go
```diff
@@ -611,6 +633,7 @@ func sch(definition string) namedSchema {
 	ns := denv.DoltDB.NodeStore()
 	ctx := context.Background()
 	root, _ := doltdb.EmptyRootValue(ctx, vrw, ns)
+	// TODO: ParseCreateTableStatement silently drops any indexes or check constraints in the definition
```
andy-wm-arthur, line 636: oof, nice catch. seems problematic!

andy-wm-arthur (approved):
LGTM

PR #5790, opened 2023-04-21
Bug fix and test for index merge bug that didn't detect incompatible index types

fulghum:
The equality functions on `Index` weren't considering `prefixLengths` or `isSpatial`, which resulted in a merge bug where dolt would get confused and match two indexes as identical when they really weren't.

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
@@ -4151,6 +4151,28 @@ var ThreeWayMergeWithSchemaChangeTestScripts = []MergeScriptTest{
 
 	// Schema conflict test cases
 	{
+		Name: "index conflicts: both sides add an index with the same name, same columns, but different type",
+		AncSetUpScript: []string{
+			"CREATE table t (pk int primary key, col1 int, col2 varchar(100));",
+		},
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4150,6 +4150,28 @@ var ThreeWayMergeWithSchemaChangeTestScripts = []MergeScriptTest{
...
+			"INSERT into t values (1, 10, '100');",
+		},
+		LeftSetUpScript: []string{
+			"alter table t add index idx1 (col2);",
+			"INSERT into t values (2, 20, '200');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query:       "call dolt_merge('right');",
+				ExpectedErr: merge.ErrSchemaConflict,
```
andy-wm-arthur, line 4169: I think this will be a "successful" merge with schema conflicts now

andy-wm-arthur (approved):
LGTM!

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
fulghum: Nice! 🙌

PR #5795, opened 2023-04-24
Correct busted interface change for CliContext

macneale4:
TIL that interface pointers in golang are not useful. This changes that, and add in a couple calls in command.go which were neglected in the first refactor.

Step 3 of N: https://github.com/dolthub/dolt/issues/3922


fulghum (approved):
Looks good! 🚢  Sorry I didn't catch that interface pointer in the first pass!

PR #5797, opened 2023-04-24
Rewriting index merging logic 

fulghum:
_Rough first pass at tidying up Dolt's logic for merging indexes together..._

Our current merge logic uses `IndexCollection.GetIndexByTags()` to find a matching index on other sides of the merge, but unfortunately this is not enough to uniquely identify an index. When multiple indexes are defined over the same column set, that code will often choose the wrong index and report an incorrect schema conflict that errors out the merge. 

This PR:
- adds a new `GetIndexesByTags()` function to `IndexCollection` that returns all indexes covering those tags
- marks the old `GetIndexByTag` as deprecated with a warning message about the race condition when multiple indexes exist over the same column set exist
- rewrites the `mergeIndexes` logic into a two step process to find the index mappings and then to merge them together and collect any index conflicts


fulghum:
Closing this in favor of a different approach in PR https://github.com/dolthub/dolt/pull/5815

PR #5798, opened 2023-04-25
--data-dir flag enabled before the sql subcommand

macneale4:
The complexity of creating global arguments is being broken into multiple pieces. This step includes the breaking up of the top level dolt command set to enable global argument parsing and initialization of the CliContext. At present, the sql subcommand is the only one getting this treatment, but it will be extended with additional commands in the future.

This change doesn't change much that is visible to the user, with the exception of the --data-dir flag being enabled before the sql subcommand.

Step 4 of N for: https://github.com/dolthub/dolt/issues/3922


go/cmd/dolt/cli/cli_context.go
```diff
@@ -14,6 +14,9 @@
 
 package cli
 
+import "github.com/dolthub/dolt/go/libraries/utils/argparser"
+
 // CliContexct is used to pass top level command information down to subcommands.
 type CliContext interface {
+	GlobalArgs() *argparser.ArgParseResults
```
fulghum, line 21: Would be nice to provide godocs for all interface functions

fulghum (commented):
Looking good! Happy to take another quick last look when you finish getting all the tests wrestled into submission.

fulghum (approved):
Looks good! 🚢  Good luck with that last failing BATS test!

PR #5809, opened 2023-04-26
Add `dolt_ignore` system table.

nicktobey:
Some notes:

1. The implementation of `dolt_ignore` is slightly more complicated than the implementation of existing writable persisted system tables (like `dolt_procedures` and `dolt_docs`), but I think it's better this way.

For context, tables like `dolt_procedures` exist as DoltTables stored in the user's db alongside user created tables. The user can read and write from these tables, but cannot create, drop, or alter them. CLI commands reference these tables and create them if they don't already exist. This works, but it means that the logic that governs these tables is spread out across the different commands that interact with them, and the user can't interact with them until running a command that creates them.

`dolt_ignore` works slightly differently: it's _backed_ by a DoltTable, but it also has its own sql.Table implementation. This allows us to define behavior for when the backing table hasn't been created yet: reads return an empty index, and writes create the table. As far as the mysql engine is concerned, this table _always_ exists. So for the commands that care about `dolt_table`, the logic can be simpler.

2. I'm not entirely happy with the implementation: I end up defining the schema twice, once as an sql.Schema, and once as a schema.Schema. I couldn't find an easy way to convert between them. If there's discrepancies between the two definitions, it could cause issues.

3. `dolt_ignore` only works with FORMAT_DOLT at the moment. If using FORMAT_LD, the table will be ignored by `dolt add` and `dolt_stash.` There's no reason it can't work with FORMAT_LD, but the logic for getting a RowIter from an Index is slightly different and I didn't invest the time to figure it out. This is currently only documented in a comment, but can be documented more widely if needed.

4. `dolt_ignore` only affects the staging and stashing of untracked files. This is consistent with git's behavior. If a table that has already been added, `dolt_ignore` will not prevent further changes to it. It also won't prevent a staged table from being renamed, either to or from an ignored table name.


nicktobey:
More tests to come, including tests for `dolt stash` with `dolt_ignore`

go/libraries/doltcore/sqle/dtables/ignore_table.go
```diff
@@ -0,0 +1,275 @@
...
+func (iw *ignoreWriter) Delete(ctx *sql.Context, r sql.Row) error {
+	if err := iw.errDuringStatementBegin; err != nil {
+		return err
+	}
+	return iw.tableWriter.Delete(ctx, r)
+}
+
+// StatementBegin is called before the first operation of a statement. Integrators should mark the state of the data
+// in some way that it may be returned to in the case of an error.
+func (iw *ignoreWriter) StatementBegin(ctx *sql.Context) {
```
fulghum, line 159: It feels like we should be able to make this code simpler... I like the experience and design you're going after, just wondering if there's an easier way to implement it...  I'll dig around a little more and see if I can come up with any helpful ideas.

fulghum (commented):
Looking very good! I'll take another look after the next iteration and I also wanted to see if we could find you a way to simplify the table implementation. 

Looks like some BATS tests are failing for the old `LD_1` format; you should be able to find existing helper bash functions that will skip those tests on LD, but let me know if you can't find 'em. 

It's probably worth spending a few minutes playing around with other dolt use cases to look for edge cases. We don't need to cover everything up front, but would be nice to know if there are any major issues. For example, `dolt dump` will not be aware of `dolt_ignore` yet, but that's probably fine to start off with. You might try some simple branching/merging and just make sure there aren't any critical issues.

go/cmd/dolt/commands/stashcmds/stash.go
```diff
@@ -137,17 +139,22 @@ func stashChanges(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPars
...
 		} else if !apr.Contains(IncludeUntrackedFlag) && allUntracked {
 			// if all changes in working set are untracked files, then no local changes to save
 			cli.Println("No local changes to save")
 			return nil
+		} else if allIgnored, err := workingSetContainsOnlyIgnoredTables(ctx, roots); err != nil {
+			return err
+		} else if !apr.Contains(AllFlag) && allIgnored {
+			// if all changes in working set are ignored files, then no local changes to save
+			cli.Println("No local changes to save")
+			return nil
```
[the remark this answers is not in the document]
nicktobey: Reworked this function because it wasn't even correct.

Only untracked tables can be ignored (this is consistent with git), and whatever documentation we release for dolt_ignore should mention this. So in your example case, the ignored table will be stashed only when --all is passed, and the untracked table will be stashed when either -u or --all is passed. Tests have been added for this.

go/libraries/doltcore/doltdb/errors.go
```diff
@@ -170,3 +170,48 @@ func GetUnreachableRootCause(err error) error {
...
+func (dc DoltIgnoreConflict) Error() string {
+	return fmt.Sprintf("dolt_ignore has multiple conflicting rules for %s", dc.Table)
+}
+
+func IsDoltIgnoreInConflict(err error) bool {
+	_, ok := err.(DoltIgnoreConflict)
+	return ok
+}
+
+func DoltIgnoreConflictTableName(err error) string {
```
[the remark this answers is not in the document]
nicktobey: I was trying to mimic the style of the other error classes in this file, but what you're pitching is just better. Changed.

go/libraries/doltcore/sqle/dtables/ignore_table.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/writer"
+	"github.com/dolthub/dolt/go/store/hash"
+	"github.com/dolthub/dolt/go/store/types"
+)
+
+var _ sql.Table = (*BranchesTable)(nil)
+var _ sql.UpdatableTable = (*BranchesTable)(nil)
+var _ sql.DeletableTable = (*BranchesTable)(nil)
+var _ sql.InsertableTable = (*BranchesTable)(nil)
+var _ sql.ReplaceableTable = (*BranchesTable)(nil)
```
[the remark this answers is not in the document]
nicktobey: Whoops! Good catch!

integration-tests/bats/ignore.bats
```diff
@@ -0,0 +1,107 @@
...
+@test "ignore: simple matches" {
+
+    dolt sql <<SQL
+CREATE TABLE ignoreme (pk int);
+CREATE TABLE dontignore (pk int);
+SQL
+
+    dolt add -A
+
+    working=$(get_working_tables)
```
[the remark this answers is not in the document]
nicktobey: Nope, this was an oversight. I'll add a "--show-ignored" flag to `dolt status`

go/cmd/dolt/commands/stashcmds/stash.go
```diff
@@ -236,6 +251,34 @@ func allAreUntrackedFilesInWorkingSet(ctx context.Context, roots doltdb.Roots) (
...
+		return false, err
+	}
+
+	ignorePatterns, err := doltdb.GetIgnoredTablePatterns(ctx, roots)
+	if err != nil {
+		return false, err
+	}
+
+	for _, tableDelta := range unstaged {
+		if !(tableDelta.IsAdd()) {
```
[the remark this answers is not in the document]
nicktobey: Whoops, that's supposed to be the reverse: only newly created tables are considered, because those are the only tables affected by dolt_ignore. (That's how git does it.)

go/libraries/doltcore/env/actions/staged.go
```diff
@@ -22,6 +22,11 @@ import (
 )
 
 func StageTables(ctx context.Context, roots doltdb.Roots, tbls []string) (doltdb.Roots, error) {
+	tbls, err := doltdb.FilterIgnoredTables(ctx, tbls, roots)
```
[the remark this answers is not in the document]
nicktobey: Noted, leaving that for a follow-up.

integration-tests/bats/ignore.bats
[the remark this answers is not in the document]
nicktobey: Fixed.

go/libraries/doltcore/sqle/dtables/ignore_table.go
nicktobey: I still think there's probably a way to simplify this but I haven't come up with any good ideas yet, are you okay with submitting it as is so we can get the feature out?

go/cmd/dolt/commands/status.go
```diff
@@ -57,14 +57,15 @@ func (cmd StatusCmd) Docs() *cli.CommandDocumentation {
 
 func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 0)
+	ap.SupportsFlag(cli.ShowIgnoredFlag, "i", "Show tables that are ignored (according to dolt_ignore)")
```
fulghum, line 60: We should stick to git's syntax wherever possible, unless there's a really compelling reason not to. For `git status`, the flag it uses to show ignored files is `--ignored`, with no shorthand `-i` allowed. (technically `--ignored` can take values as well, but they aren't required, so I'd keep it simple and not support any values yet)

go/cmd/dolt/commands/status.go
```diff
@@ -86,15 +87,15 @@ func (cmd StatusCmd) Exec(ctx context.Context, commandStr string, args []string,
...
 
-	err = PrintStatus(ctx, dEnv, staged, notStaged, as)
+	err = PrintStatus(ctx, dEnv, staged, notStaged, apr.Contains(cli.ShowIgnoredFlag), as)
 	if err != nil {
 		return handleStatusVErr(err)
 	}
 	return 0
 }
 
 // TODO: working docs in conflict param not used here
```
fulghum, line 97: I know you didn't do this one, but could you pull out this line while you're in here? It doesn't seem to make any sense any more. 🤷‍♂️ 

```suggestion
```

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -1055,6 +1055,36 @@ func UnionTableNames(ctx context.Context, roots ...*RootValue) ([]string, error)
 	return tblNames, nil
 }
 
+type IgnoredTables struct {
+	Ignore     []string
+	DontIgnore []string
+	Conflicts  []DoltIgnoreConflictError
+}
+
+// FilterIgnoredTables takes a slice of table names and returns two new slices: one with any table names that are specified by dolt_ignore, and the other with only those tables.
```
fulghum, line 1064: (minor) Looks like this godoc comment is out of date

go/libraries/doltcore/doltdb/root_val.go
```diff
@@ -1055,6 +1055,36 @@ func UnionTableNames(ctx context.Context, roots ...*RootValue) ([]string, error)
 	return tblNames, nil
 }
 
+type IgnoredTables struct {
```
fulghum, line 1058: The additions to this file seem more related to the code in `ignore_patterns.go` than to the code in `root_val.go`. I think it would probably fit better in there, but let me know if I'm just missing something of course. You could even rename that file to `ignore.go` or something, if you wanted the filename to carve out a bigger scope to hold more of the ignore implementation.

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -117,6 +117,8 @@ const (
 	ListFlag         = "list"
 	UserParam        = "user"
 	NoPrettyFlag     = "no-pretty"
+	AddIgnoredFlag   = "add-ignored"
+	ShowIgnoredFlag  = "show-ignored"
```
fulghum, line 121: Strive to always match Git's command arguments unless there's a really compelling reason not to. It's frustrating to users who know git (or who are just searching for git help online to try and figure out dolt) if we only partially follow git's cli syntax. 

For `git add` I would expect this to be `dolt add --force` or `dolt add -f` to match git's syntax. 

For `git status` I would expect this to be `dolt status --ignored` (with no shorthand option).

go/libraries/doltcore/sqle/dtables/ignore_table.go
fulghum: Yeah, I don't have a better suggestion for ya, and we can always come back and clean up these internals, so just rolling forward and getting this out sounds great.

integration-tests/bats/ignore.bats
```diff
@@ -0,0 +1,234 @@
...
+
+    dolt stash pop
+
+    conflicts=$(get_conflict_tables)
+
+    [[ ! -z $(echo "$conflicts" | grep "commit_ignore") ]] || false
+
+}
+
+@test "ignore: allow staging ingnored files if 'add --force' is supplied" {
```
fulghum, line 196: typo 😉 
```suggestion
@test "ignore: allow staging ignored files if 'add --force' is supplied" {
```

fulghum (approved):
Looks good! Just some really minor comments. The biggest notes were just making sure we stick to git's command line syntax as much as possible. Once you get those updated, this looks good to ship! 🚢 🎉

go/cmd/dolt/cli/arg_parser_helpers.go
nicktobey: Don't worry, for `dolt add`, the command is `dolt add --force`

Somehow, I had convinced myself that `git commit --force` was a command for staging ignored files during commit, but because we already created a `dolt commit --force` I had to make a new flag. Turns out, there's no flag in git for staging ignored files and I'm not sure why I was so convinced that there was.

`add-ignored` has been removed.

For `git status` I had gone with `show-ignored` because I thought that `git status ignored` required a value and didn't want to implement the same flag name with different semantics. I've switched it to be just `--ignored`.

go/libraries/doltcore/doltdb/root_val.go
nicktobey: Done, and renamed the file.

go/libraries/doltcore/doltdb/root_val.go
nicktobey: Fixed.

go/cmd/dolt/commands/status.go
nicktobey: Fixed.

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -142,6 +144,7 @@ func CreateCommitArgParser() *argparser.ArgParser {
 	ap.SupportsFlag(AllowEmptyFlag, "", "Allow recording a commit that has the exact same data as its sole parent. This is usually a mistake, so it is disabled by default. This option bypasses that safety.")
 	ap.SupportsString(DateParam, "", "date", "Specify the date used in the commit. If not specified the current system time is used.")
 	ap.SupportsFlag(ForceFlag, "f", "Ignores any foreign key warnings and proceeds with the commit.")
+	ap.SupportsFlag(AddIgnoredFlag, "i", "Allow adding otherwise ignored files.")
```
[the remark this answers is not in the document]
nicktobey: Removed.

PR #5815, opened 2023-04-26
Prevent merges when either side has multiple indexes covering the same set of columns

fulghum:
When merging multiple indexes that cover the same set of columns (even if they are different types, such as a unique index and a prefix index), there is a race condition because of how our index support assumes the column tag set will be unique for every index. This can cause the wrong index to be matched and the wrong merge decisions to be made. 

I tried to come up with another way to opportunistically match indexes across the sides of a merge in PR https://github.com/dolthub/dolt/pull/5797, but I still found more edge cases and the extra complexity didn't seem worth it. It seemed better to have Dolt be vocal about the issue and let users resolve it, instead of trying to be clever when we can't merge that situation correctly every time.


andy-wm-arthur (approved):
LGTM!
We're getting into the territory for some rather subtle merge bugs now.

PR #5818, opened 2023-04-27
Transaction isolation for branch-qualified databases

zachmu:
As well as substantially address the issues about concurrent access to `dolt_branches` called out in https://github.com/dolthub/dolt/issues/5784

As a consequence, the `dolt_branch` procedure now implicitly commits any current transaction.


zachmu:
This is in kind of an intermediate state to get it in one customer's hands a bit faster, which is why there's a lot of substantially duplicated methods in doltdb.DoltDb. Going to follow up with a new reader interface for DoltDB as of a particular noms root hash. That change will (I expect) break a lot of other things and be much larger.

go/libraries/doltcore/sqle/dtables/branches_table.go
```diff
@@ -106,17 +107,24 @@ type BranchItr struct {
...
+
 	if remote {
-		branchRefs, err = ddb.GetRefsOfType(ctx, map[ref.RefType]struct{}{ref.RemoteRefType: {}})
+		branchRefs, err = ddb.GetRefsOfTypeByNomsRoot(ctx, map[ref.RefType]struct{}{ref.RemoteRefType: {}}, txRoot)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		branchRefs, err = ddb.GetBranches(ctx)
+		branchRefs, err = ddb.GetBranchesByNomsRoot(ctx, txRoot)
```
reltuk, line 127: TODO: This should be getting Refs + Addrs and not re-resolving the refs down below.

reltuk (approved):
Generally LGTM. As discussed offline, one comment about the better approach for branches table to actually take.

go/libraries/doltcore/sqle/dtables/branches_table.go
zachmu: Will follow up with this

PR #5828, opened 2023-04-27
go/store/nbs: reduce startup latency for chunk journal

andy-wm-arthur:
https://github.com/dolthub/dolt/pull/5393 added a "journal index" to accelerate chunk journal bootstrapping. Unfortunately it never worked! Bugs in the journal index bootstrapping process caused sanity checks to fail and the journal index was subsequently dropped. The chunk journal itself was still valid and the bugs had no effect on correctness. However, without a valid journal index we were forced to always take the slow-path on startup.

### DIFF
--- a/go/store/nbs/journal_index_record.go
+++ b/go/store/nbs/journal_index_record.go
@@ -167,12 +167,16 @@ func readJournalIndexRecord(buf []byte) (rec indexRec, err error) {
 	return
 }
 
-func validateIndexRecord(buf []byte) (ok bool) {
-	if len(buf) > (indexRecLenSz + indexRecChecksumSz) {
-		off := len(buf) - indexRecChecksumSz
-		ok = crc(buf[:off]) == readUint32(buf[off:])
+func validateIndexRecord(buf []byte) bool {
+	if len(buf) < (indexRecLenSz + indexRecChecksumSz) {
+		return false
 	}
-	return
+	off := readUint32(buf)
+	if int(off) > len(buf) {
+		return false
+	}
+	off -= indexRecChecksumSz
+	return crc(buf[:off]) == readUint32(buf[off:])
 }
 
 // processIndexRecords reads a sequence of index records from |r| and passes them to the callback. While reading records
--- a/go/store/nbs/journal_writer.go
+++ b/go/store/nbs/journal_writer.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 
 	"github.com/dolthub/swiss"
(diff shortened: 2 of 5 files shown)


go/store/nbs/journal_index_record.go
```diff
@@ -167,12 +167,16 @@ func readJournalIndexRecord(buf []byte) (rec indexRec, err error) {
 	return
 }
 
-func validateIndexRecord(buf []byte) (ok bool) {
-	if len(buf) > (indexRecLenSz + indexRecChecksumSz) {
-		off := len(buf) - indexRecChecksumSz
```
andy-wm-arthur, line 172: previously, we assumed that `buf` was pre-sliced to the correct record length (it wasn't)

go/store/nbs/journal_writer.go
```diff
@@ -187,36 +188,67 @@ func (wr *journalWriter) bootstrapJournal(ctx context.Context) (last hash.Hash,
...
+		ch := make(chan []lookup, 4)
+
 		// process the indexed portion of the journal
-		err = processIndexRecords(ctx, wr.index, info.Size(), func(o int64, r indexRec) (err error) {
-			switch r.kind {
-			case tableIndexRecKind:
-				// |r.end| is expected to point to a root hash record in |wr.journal|
-				// containing a hash equal to |r.lastRoot|, validate this here
-				var h hash.Hash
-				if h, err = peekRootHashAt(wr.journal, int64(r.end)); err != nil {
```
andy-wm-arthur, line 197: `peekRootHash` uses a fixed size buffer of 1024 bytes and does not slice the buffer before passing it to `validateIndexRecord`, so the checksum step always failed.
https://github.com/dolthub/dolt/blob/29e0f200122d801f78f4c29a53d80ffb28a82142/go/store/nbs/journal_record.go#L253

go/store/nbs/journal_writer.go
```diff
@@ -187,36 +188,67 @@ func (wr *journalWriter) bootstrapJournal(ctx context.Context) (last hash.Hash,
...
-				// todo: uncompressed size
-			default:
-				return fmt.Errorf("unknown index record kind (%d)", r.kind)
 			}
-			return nil
 		})
+
+		err = eg.Wait()
 		if err != nil {
-			// todo: issue warning on corrupt index recovery
```
andy-wm-arthur, line 216: failing to process `journal.idx` was considered a recoverable error, so we were (always) silently dropping the journal index

reltuk (approved):
LGTM!

PR #5830, opened 2023-04-27
Automatically resolve additive enum/set changes in merges

fulghum:
This PR adds a new type compatibility checker for resolving compatible type changes during a merge. The intent is to be able to automatically resolve merges where a column's type has changed in a compatible way, starting with adding a new value at the end of an enum or a set. 

I considered a few places to put this logic. The TypeInfo interface seemed like a logical place, but I wasn't completely comfortable with every type having to know the implementation details for each storage layer's implementation. There are also only two types that we're implementing this logic for initially (sets and enums), so editing all types seemed like overkill right now. I figured it would be a good start to get the initial logic together here, then iterate on expanding it or moving it around.


andy-wm-arthur (approved):
LGTM!

PR #5831, opened 2023-04-27
prevent panics on chain alters

JCOR11599:
Partial fix for: https://github.com/dolthub/dolt/issues/5786
Companion PR: https://github.com/dolthub/go-mysql-server/pull/1736


max-hoffman (approved):
LGTM, only extra check would be scanning `IndexOfColName` to see if there are any other unsafe index accesses like this

PR #5838, opened 2023-04-28
Adding a mysql connector integ test for the Ruby mysql2 library

fulghum:
Issue https://github.com/dolthub/dolt/issues/5834 uncovered that we weren't testing what seems to be the [recommended Ruby MySQL connector library (`mysql2`)](https://dev.mysql.com/doc/refman/8.0/en/apis-ruby.html), so this PR adds a connector integration test for it.


timsehn (approved):
LGTM. Thanks for at least doing a basic assert test. I wrote this so blame me :-)

PR #5842, opened 2023-04-28
Output for `dolt status` should say "tables", not "files"

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/5840


timsehn (approved):
Ship it.

PR #5843, opened 2023-04-29
BATS Executor Exemplar

macneale4:
A utility command which runs bats test in parallel.

```
$ batsee --help
NAME
        batsee - Run the Bats Tests concurrently

SYNOPSIS
        batsee [-t <threads>] [--skip-slow] [--max-time <time>] [--retries <retries>] [--only test1,test2,...]

DESCRIPTION
        From within the integration-test/bats directory, run the bats tests concurrently.
        Output for each test is written to a file in the batsee_output directory.
        Example:  batsee -t 42 --max-time 1h15m -r 2 --only types.bats,foreign-keys.bats

OPTIONS
        -t <threads>, --threads=<threads>
          Number of tests to execute in parallel. Defaults to 12

        -s, --skip-slow
          Skip slow tests. This is a static list of test we know are slow, may grow stale.

        --max-time=<duration>
          Maximum time to run tests. Defaults to 30m

        --only
          Only run the specified test, or tests (comma separated)

        -r <retries>, --retries=<retries>
          Number of times to retry a failed test. Defaults to 1
```

### DIFF
--- a/go/cmd/dolt/commands/batsee.go
+++ b/go/cmd/dolt/commands/batsee.go
@@ -61,6 +61,7 @@ func (b BatseeCmd) Hidden() bool {
 func (b BatseeCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParserWithVariableArgs(b.Name())
 	ap.SupportsUint("threads", "t", "processes", "Number of tests to execute in parallel. Defaults to 12")
+	ap.SupportsFlag("skip-slow", "", "Skip slow tests")
 	return ap
 }
 
@@ -72,6 +73,15 @@ type batsResult struct {
 	runtime time.Duration
 	path    string
 	err     error
+	skipped bool
+}
+
+var skipCommands = map[string]bool{
+	"export-tables.bats":      true,
+	"garbage_collection.bats": true,
+	"remotes.bats":            true,
+	"remotesrv.bats":          true,
+	"schema-changes.bats":     true,
 }
 
 // list of slow commands. These tend to run more than 5-7 min, so we want to run them first.
@@ -120,6 +130,9 @@ func (b BatseeCmd) Exec(ctx context.Context, commandStr string, args []string, d
 	// Insert the slow tests first
 	for key, _ := range slowCommands {
 		workQueue = append(workQueue, key)
+		if _, ok := skipCommands[key]; !ok {
+			skipCommands[key] = true
+		}
 	}
 	// Then insert the rest of the tests
 	for _, file := range files {
@@ -152,19 +165,30 @@ func (b BatseeCmd) Exec(ctx context.Context, commandStr string, args []string, d
 	wg.Wait()
 	close(results)
 
-	passStr := color.GreenString("PASS")
-	failStr := color.RedString("FAIL")
+	passStr := color.GreenString("PASSED")
+	failStr := color.RedString("FAILED")
+	skippedStr := color.YellowString("SKIPPED")
 	failedQ := []batsResult{}
+	skippedQ := []batsResult{}
 	for result := range results {
+		if result.skipped {
+			skippedQ = append(skippedQ, result)
+			continue
+		}
+
 		if result.err != nil {
 			failedQ = append(failedQ, result)
 		} else {
-			cli.Println(fmt.Sprintf("%s completed in %s with status %s", result.path, durationStr(result.runtime), passStr))
+			cli.Println(fmt.Sprintf("%10s %-40s (time: %s)", passStr, result.path, durationStr(result.runtime)))
 		}
 	}
+	for _, result := range skippedQ {
+		cli.Println(fmt.Sprintf("%10s %-40s (time:NA)", skippedStr, result.path))
+	}
+
 	exitStatus := 0
 	for _, result := range failedQ {
-		cli.Println(fmt.Sprintf("%s completed in %s with status %s", result.path, durationStr(result.runtime), failStr))
+		cli.Println(fmt.Sprintf("%10s %-40s (time:%s)", failStr, result.path, durationStr(result.runtime)))
 		exitStatus = 1
 	}
 
@@ -188,6 +212,12 @@ func runBats(path string, resultChan chan<- batsResult) {
 	cmd := exec.Command("bats", path)
 
 	result := batsResult{path: path}
+
+	if _, ok := skipCommands[path]; ok {
+		result.skipped = true
+		resultChan <- result
+		return
+	}
 
 	startTime := time.Now()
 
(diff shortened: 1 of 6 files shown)


nicktobey:
Exciting! I'm going to check it out.

What was the motivation behind making this a dolt sub-command instead of a separate dev tool?

go/cmd/dolt/commands/batsee.go
```diff
@@ -0,0 +1,377 @@
...
+	"github.com/dolthub/dolt/go/cmd/dolt/cli"
+	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
+	"github.com/dolthub/dolt/go/libraries/doltcore/env"
+	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+)
+
+var batseeDoc = cli.CommandDocumentationContent{
+	ShortDesc: `Run the Bats Tests concurrently`,
+	LongDesc:  `From within the integration-test/bats directory, run the bats tests concurrently. Output for each test is written to a file in the batsee_output directory.`,
+	Synopsis: []string{
```
nicktobey, line 41: Our other commands use a template as the synopsis, not examples. So it might
```suggestion
	Synopsis: []string{
	    "[-t {{.LessThan}}threads{{.GreaterThan}}] [--skip-slow] [--max-time {{.LessThan}}time{{.GreaterThan}}] [--retries {{.LessThan}}retries{{.GreaterThan}}] [--only test1,test2,...]
```

If you want to include an example you can do that in the LongDesc.

go/cmd/dolt/commands/batsee.go
```diff
@@ -0,0 +1,377 @@
...
+	"index-on-writes-2.bats":     true,
+	"sql.bats":                   true,
+	"remotes.bats":               true,
+}
+
+func (b BatseeCmd) Exec(ctx context.Context, commandStr string, args []string, dEnv *env.DoltEnv, cliCtx cli.CliContext) int {
+	ap := b.ArgParser()
+	help, usage := cli.HelpAndUsagePrinters(cli.CommandDocsForCommandString(commandStr, batseeDoc, ap))
+
+	apr, err := ap.Parse(args)
```
nicktobey, line 110: You can simplify this with

```suggestion
	ap := cmd.ArgParser()
	help, usage := cli.HelpAndUsagePrinters(cli.CommandDocsForCommandString(commandStr, batseeDoc, ap))
	apr := cli.ParseArgsOrDie(ap, args, help)
```

go/cmd/dolt/commands/batsee.go
```diff
@@ -0,0 +1,377 @@
...
+		durationInput = "30m"
+	}
+	duration, err := time.ParseDuration(durationInput)
+	if err != nil {
+		cli.Println("Error parsing duration:", err)
+		return 1
+	}
+
+	skipSlow := false
+	_, hasVal := apr.GetValue("skip-slow")
```
nicktobey, line 136: For flags you can use apr.Contains("skip-slow")

go/cmd/dolt/commands/batsee.go
```diff
@@ -0,0 +1,377 @@
...
+	return cli.NewCommandDocumentation(batseeDoc, b.ArgParser())
+}
+
+func (b BatseeCmd) Hidden() bool {
+	return true
+}
+
+func (b BatseeCmd) ArgParser() *argparser.ArgParser {
+	ap := argparser.NewArgParserWithMaxArgs(b.Name(), 0)
+	ap.SupportsUint("threads", "t", "threads", "Number of tests to execute in parallel. Defaults to 12")
```
nicktobey, line 72: The flag names should be defined as constants at the top of the file, like

```
const (
    ThreadsFlag = "threads"
)

ap.SupportsUint(ThreadsFlag, ...)

threads, hasThreads : = apr.GetValue(ThreadsFlag)
```

nicktobey:
Looks good to me.

Can we get some tests for the command line parsing / results formatting?

If we want to avoid the obvious infinite loop, we could have a hidden flag that makes `dolt batsee` not actually run any tests and assume that they all pass/fail/timeout etc, so that we can inspect that the results are formatted correctly.

macneale4:
> Looks good to me.
> 
> Can we get some tests for the command line parsing / results formatting?
> 
> If we want to avoid the obvious infinite loop, we could have a hidden flag that makes `dolt batsee` not actually run any tests and assume that they all pass/fail/timeout etc, so that we can inspect that the results are formatted correctly.

Testing this tool seems a little overkill at this point. The payoff on value is pretty low until such time that we depend on it as release infrastructure. Going to punt to we need it.

Regarding making is a subcommand - No strong reason other than having everything in one executable. I thought the signal handling would come in handy, but it's really only a few lines of code to replicate that. More than anything else it was easiest.

go/cmd/dolt/commands/batsee.go
macneale4: There is a more general problem which is the differentiation between `help` and `usage` - I'm not convinced this is required, but the error handling here was trying to preserve a difference. I'll do it the way you suggest, but there is room for general clean up across the board here.

go/cmd/dolt/commands/batsee.go
macneale4: I'm soooooo tepid to this suggestion :-)

go/cmd/dolt/commands/batsee.go
macneale4: Done

go/cmd/dolt/commands/batsee.go
```diff
@@ -67,13 +66,21 @@ func (b BatseeCmd) Hidden() bool {
 	return true
 }
 
+const (
+	threadsF  = "threads"
```
nicktobey, line 70: I think we should just say "threadsFlag" instead of "threadsF". There's no reason to abbreviate it.

macneale4: > Regarding making is a subcommand - No strong reason other than having everything in one executable. I thought the signal handling would come in handy, but it's really only a few lines of code to replicate that. More than anything else it was easiest.

I have a strong reason no to do this - Windows. The signal handling won't work. I'll make it a separate command.

nicktobey:
Approved, pending my last comment.

I definitely think that we're going to want to incorporate this into the CI infrastructure, and when we do, we want to test it. But I'm willing to punt on that until we actually add it.

macneale4:
> Regarding making is a subcommand - No strong reason other than having everything in one executable. I thought the signal handling would come in handy, but it's really only a few lines of code to replicate that. More than anything else it was easiest.

I have a good reason not to do this - Windows. signal handling won't work and is breaking the build on that platform. I'll move it to a separate command.

go/cmd/dolt/commands/batsee.go
macneale4: I disagree, but don't care enough to argue.

Issue #5850, opened 2023-05-01
`CALL dolt_clone()` does not configure the cloned database for replication.


reltuk:
If a server is configured for cluster or write replication, `CREATE DATABASE ...` configures the newly create database for replication.

However, getting the database with `CALL dolt_clone()` does not configure the newly cloned database for replication. It should be setup with replication hooks, expected remotes should be created with the desired url templates, etc.

PR #5852, opened 2023-05-01
go/libraries/doltcore/sqle: database_provider.go: When cloning a remote into a sql-server, always run the InitDatabaseHook on it.

reltuk:
Failing to run the hook means that the cloned database is not appropriately configured for replication, for example.

Fixes: #5850


zachmu (approved):
Yikes, good catch

PR #5853, opened 2023-05-01
`dolt diff` shouldn't print ignored tables.

### LINKED ISSUE #5841: Newly created tables show up in diff even if they are ignored.
Repro:
```
$ dolt init --fun                                                             
$ dolt sql -q "create table this_table_sux_anyways(pk int primary key, c1 int)"
$ dolt sql -q "create table greatest_table_ever(pk int primary key, c1 int)"   
$ dolt sql -q "insert into dolt_ignore values ('this_table_sux_anyways', 1)"      
$ dolt status
On branch main
Untracked files:
  (use "dolt add <table>" to include in what will be committed)
	new table:        dolt_ignore
	new table:        greatest_table_ever
$ dolt diff
diff --dolt a/dolt_ignore b/dolt_ignore
added table
+CREATE TABLE `dolt_ignore` (
+  `pattern` varchar(16383) NOT NULL,
+  `ignored` tinyint,
+  PRIMARY KEY (`pattern`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin;
+---+------------------------+---------+
|   | pattern                | ignored |
+---+------------------------+---------+
| + | this_table_sux_anyways | 1       |
+---+------------------------+---------+
diff --dolt a/greatest_table_ever b/greatest_table_ever
added table
+CREATE TABLE `greatest_table_ever` (
+  `pk` int NOT NULL,
+  `c1` int,
+  PRIMARY KEY (`pk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin;
diff --dolt a/this_table_sux_anyways b/this_table_sux_anyways
added table
+CREATE TABLE `this_table_sux_anyways` (
+  `pk` int NOT NULL,
+  `c1` int,
+  PRIMARY KEY (`pk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin;
```

The `this_table_sux_anyways` table should not show in dolt diff.

nicktobey:
Fixes #5841

### DIFF
--- a/integration-tests/bats/ignore.bats
+++ b/integration-tests/bats/ignore.bats
@@ -277,3 +277,19 @@ SQL
     [[ "$output" =~ "ignoreme" ]] || false
 
 }
+
+@test "ignore: don't display ignored tables in dolt diff" {
+    skip_nbf_ld_1
+
+    dolt sql <<SQL
+CREATE TABLE ignoreme (pk int);
+CREATE TABLE nomatch (pk int);
+SQL
+
+    run dolt diff
+
+    [ "$status" -eq 0 ]
+
+    [[ "$output" =~ "nomatch" ]] || false
+    [[ ! ["$output" =~ "ignoreme"] ]] || false
+}
--- a/integration-tests/bats/ignore.bats
+++ b/integration-tests/bats/ignore.bats
@@ -238,7 +238,7 @@ SQL
 
     [ "$status" -eq 0 ]
 
-    [[ ! ["$output" =~ "diff --dolt a/ignoreme b/ignoreme"] ]] || false
+    ! [["$output" =~ "diff --dolt a/ignoreme b/ignoreme"]] || false
 
 }
 
@@ -255,8 +255,8 @@ SQL
     [ "$status" -eq 0 ]
 
     [[ "$output" =~ "nomatch" ]] || false
-    [[ ! ["$output" =~ "Ignored tables"] ]] || false
-    [[ ! ["$output" =~ "ignoreme"] ]] || false
+    ! [["$output" =~ "Ignored tables"]] || false
+    ! [["$output" =~ "ignoreme"]] || false
 
 }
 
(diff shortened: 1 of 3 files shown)


integration-tests/bats/ignore.bats
```diff
@@ -276,4 +276,20 @@ SQL
...
+CREATE TABLE ignoreme (pk int);
+CREATE TABLE nomatch (pk int);
+SQL
+
+    run dolt diff
+
+    [ "$status" -eq 0 ]
+
+    [[ "$output" =~ "nomatch" ]] || false
+    [[ ! ["$output" =~ "ignoreme"] ]] || false
```
timsehn, line 294:
```suggestion
    ! [["$output" =~ "ignoreme" ]] || false
```

timsehn: I think this is the way to do this

timsehn (commented):
Bats suggestion

integration-tests/bats/ignore.bats
nicktobey: Updated all tests in ignore.bats

max-hoffman (approved):
LGTM, only thing i can think of is doing the reverse diff to test the working set on the other side of the diff

nicktobey:
Is it possible to reverse the diff? We don't support `diff -R`, and `diff` doesn't recognize WORKING as a refspec, so I'm not sure what command currently supported could produce the reverse diff.

nicktobey:
I added `dolt diff -R` just because it was really simple to add.

max-hoffman (approved):
LGTM again, unrelated but it doesn't look like the sql diff function is getting the same hidden table treatment? Is that something we are interested in making an issue for?

nicktobey:
Oh shoot, you're right. This also affects the `dolt_status` system table (We can't really pass flags to a table, but we could create a new table name like `dolt_status_ignored` if we want to pollute the namespace of table names)

I can make follow up issues for these.

PR #5856, opened 2023-05-02
fix empty string import for ENUM and SET types

jennifersp:


fulghum (approved):
I went back and dug into this ticket a little more and realized our table import/export code is already distinguishing between `,,` and `,"",`, even though that's not fully standard with CSVs. That clears up the concern I had before and also make sure that we can roundtrip data from `dolt table export` back to `dolt table import` again without losing data. 

I tweaked the test just a little bit – I changed it to assert the full contents of the output csv file exactly matched the input to test the roundtripping explicitly.

PR #5860, opened 2023-05-03
go: env/actions: remotes.go: In SyncRoots, if we are syncing to an empty destination repository, use pull.Clone instead of pull.Pull.

reltuk:
pull.Clone uses the TableFileStore interface to transit whole table files without needing to follow chunk references or do any reconciliation with the destination database regarding what it already has. It is much faster against every time of remote.

### DIFF
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -455,6 +455,35 @@ func FetchRefSpecs(ctx context.Context, dbData env.DbData, srcDB *doltdb.DoltDB,
 	return nil
 }
 
+// SyncRoots is going to copy the root hash of the database from srcDb to destDb.
+// We can do this |Clone| if (1) destDb is empty, (2) destDb and srcDb are both
+// |TableFileStore|s, and (3) srcDb does *not* have a journal file. The most
+// common scenario where this occurs is when we are restoring a backup.
+//
+// The journal's interaction with TableFileStore is not great currently ---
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -481,6 +484,69 @@ func SyncRoots(ctx context.Context, srcDb, destDb *doltdb.DoltDB, tempTableDir s
...
+
+						for _, v := range stats {
+							if v.Percent > 0.001 {
+								totalsentbytes += v.Read
+								totalbytes += uint64(float64(v.Read) / v.Percent)
+							}
+						}
+
+						// We fake some of these values.
+						toemit := pull.Stats{
```
zachmu, line 515: toemit 🦶🧤

`toEmit`

go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -481,6 +484,69 @@ func SyncRoots(ctx context.Context, srcDb, destDb *doltdb.DoltDB, tempTableDir s
...
+			for {
+				select {
+				case tfe, ok := <-tfCh:
+					if !ok {
+						return
+					}
+					if tfe.EventType == pull.DownloadStats {
+						stats[tfe.TableFiles[0]] = tfe.Stats[0]
+
+						totalsentbytes := uint64(0)
```
zachmu, line 504: Bro did your shift key break

zachmu (approved):
LGTM!

PR #5865, opened 2023-05-03
Drop the formatting flag from the engine

macneale4:
Small step to untangling shell and query engine. https://github.com/dolthub/dolt/issues/3922

Existing tests cover behavior. This is a fairly straight forward refactor.


fulghum (approved):
Looks good! 🚢

PR #5867, opened 2023-05-03
Support adding columns to keyless tables

fulghum:
Fixes: https://github.com/dolthub/dolt/issues/5846

New tests in GMS: https://github.com/dolthub/go-mysql-server/pull/1744


zachmu (approved):
Too easy!

PR #5868, opened 2023-05-03
go through primary keys backwards

JCOR11599:
This PR implements an iterator to go over Primary Keys backwards, allowing us to order by descending quickly.

Companion PR: https://github.com/dolthub/go-mysql-server/pull/1742
Fix for: https://github.com/dolthub/dolt/issues/5812


go/libraries/doltcore/sqle/index/index_lookup.go
```diff
@@ -460,6 +469,7 @@ type nomsLookupBuilder struct {
 // NewRowIter implements IndexLookup
 func (lb *nomsLookupBuilder) NewRowIter(ctx *sql.Context, part sql.Partition) (sql.RowIter, error) {
 	p := part.(rangePartition)
+	p.nomsRange.Reverse = p.isReverse
```
max-hoffman, line 472: should this be an error if true?

max-hoffman (approved):
LGTM, more enginetests with edge conditions would be nice. E.g. empty tables, single row tables, etc

go/libraries/doltcore/sqle/index/index_lookup.go
JCOR11599: My mistake, this line should just be removed entirely, since we're not going to support reverse `IndexTableAccess` on old format.

PR #5870, opened 2023-05-03
Improve dolt_ignore pattern rules

nicktobey:
This PR does two things:

- introduces % as an alias for *, since % is the SQL wildcard
- adds an additional check for dolt_ignore rules that are equivalent but have different `ignored` values.


fulghum (approved):
Looks good! 🚢

Issue #5531, opened 2023-03-09
Foreign Keys can be dropped in a merge of branches that both resolve foreign keys


timsehn:
Repro here:

https://github.com/dolthub/dolt_dropped_fk_repro

timsehn:
Related to #5529 5531

fulghum:
Fixed in https://github.com/dolthub/dolt/pull/5874

PR #5874, opened 2023-05-03
go/doltcore/merge: use unresolved-aware fk matching method in Foreign…

andy-wm-arthur:
…KeysMerge 

fix for #5531


fulghum (approved):
Looks good! 🚢

PR #5878, opened 2023-05-04
Branch Control fixes and improvements

Hydrocharged:
There was a bug where, under very specific conditions, we could get the branch control table to have a nil map when there shouldn't be one, which would throw a panic. This situation also seemed to cause a bug where you could "infinitely" delete branches without them actually being removed, and our duplicate key detection failed to work in these cases. The root cause of both of those issues have been fixed.

In addition, if a customer wanted to reject a user from using a particular branch, but allow others, then it was very, very verbose to implement this functionality. I've instead modified our branch control system to take the longest match (which will be the most specific one due to our folding of expressions), and also added an explicit `"read"` permission, to force some (or all) users to only have `read` permissions even if there is a global allowance. This also matches the namespace table, which already operated on a longest-match principle.

### DIFF
--- a/go/libraries/doltcore/branch_control/access.go
+++ b/go/libraries/doltcore/branch_control/access.go
@@ -31,8 +31,9 @@ type Permissions uint64
 const (
 	Permissions_Admin Permissions = 1 << iota // Permissions_Admin grants unrestricted control over a branch, including modification of table entries
 	Permissions_Write                         // Permissions_Write allows for all modifying operations on a branch, but does not allow modification of table entries
+	Permissions_Read                          // Permissions_Read allows for reading from a branch, which is equivalent to having no permissions
 
-	Permissions_None Permissions = 0 // Permissions_None represents a lack of permissions
+	Permissions_None Permissions = 0 // Permissions_None represents a lack of permissions, which defaults to allowing reading
 )
 
 // Access contains all of the expressions that comprise the "dolt_branch_control" table, which handles write Access to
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/branch_control/access.go
```diff
@@ -77,9 +78,14 @@ func newAccess() *Access {
 // Requires external synchronization handling, therefore manually manage the RWMutex.
 func (tbl *Access) Match(database string, branch string, user string, host string) (bool, Permissions) {
 	results := tbl.Root.Match(database, branch, user, host)
+	length := uint32(0)
```
zachmu, line 81: What's the point of this variable def? Seems like it would clearer to inline

integration-tests/bats/branch-control.bats
```diff
@@ -229,3 +229,64 @@ branch', 'test', '%', 'admin')"
...
+
+  start_sql_server
+
+  run dolt sql-client -P $PORT --use-db "dolt_repo_$$" -u test -q "call dolt_checkout('test-branch'); create table t (c1 int)"
+  [ $status -ne 0 ]
+  [[ $output =~ "does not have the correct permissions" ]] || false
+
+  dolt sql-client -P $PORT --use-db "dolt_repo_$$" -u root -q "delete from dolt_branch_control where branch = 'test-branch'"
+
+  dolt sql-client -P $PORT --use-db "dolt_repo_$$" -u test -q "call dolt_checkout('test-branch'); create table t (c1 int)"
```
zachmu, line 248: You need a `run` here or the checks below are stale

zachmu (approved):
LGTM

PR #5881, opened 2023-05-04
Remove overzealous validation check.

nicktobey:
This check appears to be redundant, in order to fail early and return a more helpful error message. But it's not helpful if it's wrong... which it is. It errors fast on inputs that should be correct (such as auto-increment columns), and it not only fails on the example from https://github.com/dolthub/dolt/issues/5855, but it would prevent us from implementing a permanent fix.

Removing the check, all tests still pass, because invalid imports are still detected by the remaining checks.


fulghum (approved):
Looks good 🚢

PR #5882, opened 2023-05-04
Apply column defaults when merging in new columns

fulghum:
Currently, when merging in a new column and updating existing rows for the new column, any column default value for the new column is ignored. This can result in inconsistent data that customers do not expect (e.g. https://github.com/dolthub/dolt/issues/4456). 

This PR adds initial support for applying column defaults when merging in a new column. Only literal default values are supported in this PR. Supporting column references and function references will require more work to properly resolve those references through the analyzer's code. If an unsupported column default value is present, this change detects that we cannot apply it and errors out the merge, instead of dropping the default value and leaving the new column null. 

Fixes https://github.com/dolthub/dolt/issues/4456 (which specifically only needs support for literal default values)

### DIFF
--- a/go/libraries/doltcore/merge/merge_prolly_rows.go
+++ b/go/libraries/doltcore/merge/merge_prolly_rows.go
@@ -20,20 +20,27 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/parse"
+	errorkinds "gopkg.in/src-d/go-errors.v1"
 	"io"
-
-	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -712,6 +720,62 @@ func remapTuple(tuple val.Tuple, desc val.TupleDesc, mapping val.OrdinalMapping)
...
+		var value interface{}
+		if from == -1 {
+			// If the column is a new column, then look up any default value
+			col := mergedSch.GetNonPKCols().GetByIndex(to)
+			if col.Default != "" {
+				// We need a sql.Context to call parse.StringToColumnDefaultValue, so see if we have one already;
+				// if not, create one, since this code also gets called from the CLI merge code path.
+				sqlCtx, ok := ctx.(*sql.Context)
+				if !ok {
+					sqlCtx = sql.NewContext(ctx)
```
andy-wm-arthur, line 741: creating a new sql context is surprisingly expensive due to the cost of creating the embedded sql session, we should create this once per table/root if possible.

andy-wm-arthur (approved):
This looks great! I think punting on non-literal values is a good call for now

go/libraries/doltcore/merge/merge_prolly_rows.go
fulghum: Oh, thanks for pointing that out! I just pushed a new commit that moves this out of the loop and reuses a single instance for the entire table merge.

PR #5885, opened 2023-05-05
go/libraries/doltcore/env/actions: remotes.go: Have fetch pull all the branch heads at once.

reltuk:
Also avoids round-tripping to the remote to load a commit whose value we do not need yet.


zachmu (approved):
LGTM!

PR #5890, opened 2023-05-05
normalize windows file paths better

JCOR11599:
Depending on if the windows file paths start with `file:///C:...` or `file://C:...`, the url package would parse these slightly differently. As a result, the environment would be stored in the singleton under one path, and retrieved with a different one, causing errors.

fix for: https://github.com/dolthub/dolt/issues/5780


fulghum (approved):
Looks good! 🚢

PR #5892, opened 2023-05-05
Fix binlog connection re-establishment when connection closes between messages

fulghum:
This PR changes our handling of `io.EOF` to trigger our connection re-establishment logic. Previously, only `io.UnexpectedEOF` was triggering a connection re-establishment. 

We found this by digging into some flakiness in the `TestBinlogReplicationAutoReconnect` test:
- https://github.com/dolthub/dolt/actions/runs/4888594518/jobs/8726401205
- https://github.com/dolthub/dolt/actions/runs/4888594518/jobs/8726401205
- https://github.com/dolthub/dolt/actions/runs/4888566463/jobs/8726344999

This test is a little more complex than the other binlog replication tests – it uses Toxiproxy to simulate a dying connection to test the automatic reconnection feature. 

In each of the failures above, the test fails after timing out waiting for the replica to report it has processed the same GTID set that the primary reports. Each time, the primary reports GTIDS 1-1004, but the replica only has GTIDS 1-3. 

Comparing the failed test run log output to a successful test run... the successful run shows the log message `no binlog connection to source, attempting to establish one` many more times. While the Toxiproxy has a data_limit toxic in place, we keep re-establishing a connection and reading the data before Toxiproxy starts causing chaos. This is as intended, and when the data_limit toxic is removed, the binlog replica re-establishes a connection again, and this time is able to keep pulling off events. However, in the failed tests above, I'm only seeing this connection re-establishment happen once, and then the replica just keeps reporting `No more binlog messages; retrying in 1s...`, which indicates it has a connection, but isn't receiving anything. 

This seems to happen when Toxiproxy happens to close the connection cleanly between binlog messages. If the connection close happens while a binlog message is in-flight, we receive an `io.UnexpectedEOF` error and correctly re-establish a connection. If the timing is just right, then we will receive an `io.EOF`. I previously thought that we would receive an `EOF` if there was no binlog event to read over the connection after a period of time, but after looking at the vitess connection handling code more, this doesn't seem to be the case, and `EOF` should be treated the same as `UnexpectedEOF`. 

There are also a few other minor test tweaks in here, like ensuring we check for any returned error when managing the toxics loaded on Toxiproxy.

### DIFF
--- a/go/libraries/doltcore/sqle/binlogreplication/binlog_replication_reconnect_test.go
+++ b/go/libraries/doltcore/sqle/binlogreplication/binlog_replication_reconnect_test.go
@@ -54,7 +54,7 @@ func TestBinlogReplicationAutoReconnect(t *testing.T) {
 	// Remove the limit_data toxic so that a connection can be reestablished
 	err := mysqlProxy.RemoveToxic("limit_data")
 	require.NoError(t, err)
-	fmt.Printf("Toxiproxy proxy limit_data toxic removed")
+	t.Logf("Toxiproxy proxy limit_data toxic removed")
 
 	// Assert that all records get written to the table
 	waitForReplicaToCatchUp(t)
@@ -157,7 +157,7 @@ func configureToxiProxy(t *testing.T) {
 		toxiproxyServer.Listen("localhost", strconv.Itoa(toxiproxyPort))
 	}()
 	time.Sleep(500 * time.Millisecond)
-	fmt.Printf("Toxiproxy control plane running on port %d \n", toxiproxyPort)
+	t.Logf("Toxiproxy control plane running on port %d", toxiproxyPort)
 
 	toxiClient = toxiproxyclient.NewClient(fmt.Sprintf("localhost:%d", toxiproxyPort))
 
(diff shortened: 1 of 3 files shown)


go/libraries/doltcore/sqle/binlogreplication/binlog_replication_reconnect_test.go
```diff
@@ -40,15 +41,20 @@ func TestBinlogReplicationAutoReconnect(t *testing.T) {
...
+
 	for i := 0; i < 1000; i++ {
 		value := "foobarbazbashfoobarbazbashfoobarbazbashfoobarbazbashfoobarbazbash"
 		primaryDatabase.MustExec(fmt.Sprintf("insert into reconnect_test values (%v, %q)", i, value))
 	}
 	// Remove the limit_data toxic so that a connection can be reestablished
-	mysqlProxy.RemoveToxic("limit_data")
+	err := mysqlProxy.RemoveToxic("limit_data")
+	require.NoError(t, err)
+	fmt.Printf("Toxiproxy proxy limit_data toxic removed")
```
reltuk, line 57: Nit: `t.Logf(...)` will ensure this prints if `-test.v` is supplied or anytime the test fails, which can be nice.

reltuk (approved):
LGTM! Thanks for digging in.

go/libraries/doltcore/sqle/binlogreplication/binlog_replication_reconnect_test.go
fulghum: ooh... thanks for pointing that out – I hadn't come across that yet. I went ahead and switched this file over to `t.Logf` and will prefer that func in the future, too.

PR #5899, opened 2023-05-09
go: sqle: dolt_gc.go: Disable CALL dolt_gc() on a standby replica, where it is not currently safe. Shallow GC is still available.

reltuk:


zachmu (approved):
LGTM!

PR #5902, opened 2023-05-09
go/store/datas/pull/clone.go: Fix to clone to interact with sql-server remotes better.

reltuk:
Before this fix, |dolt clone| against a sql-server remote can fail with a confusing error message if the sql-server has a chunk journal. Adding the chunk journal to our destination ChunkStore causes us to update the root hash, which the Clone code was not expecting.

This change updates the Clone code to look for the possible update to the destination's root hash and to not try to set the root hash on the ChunkStore if it has already been done by the code handling the chunk journal.

### DIFF
--- a/go/store/datas/pull/clone.go
+++ b/go/store/datas/pull/clone.go
@@ -55,7 +55,7 @@ func Clone(ctx context.Context, srcCS, sinkCS chunks.ChunkStore, eventCh chan<- 
 		return fmt.Errorf("%w: sink db is not a Table File Store", ErrCloneUnsupported)
 	}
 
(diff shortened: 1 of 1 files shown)


andy-wm-arthur (approved):
LGTM!

For posterity: the presumed error case was a clone from a running SQL server that was actively writing the the chunk journal. The cloning client asked for repo metadata describing the current chunk store root. By the time the client asked for or downloaded the chunk journal, new root records had been written to it. Therefore the setting the chunk store root to the value originally read by the get repo metadata call fails with an optimistic lock failure.

Issue #3963, opened 2022-07-29
Eliminate column tags, use column name for identity in all cases


druvv:
Today, Dolt has an internal tag system that is used to assign identities to columns. The tag system helps Dolt diff and merge columns even when they have been renamed or modified. 

For example, let's say I have a users table with a name as the primary key and a phone number stored as an integer. If I rename the phone number column, the internal tag of that column remains the same.
```
$ dolt sql -q "CREATE table users (name VARCHAR(100) PRIMARY KEY, phone BIGINT);"
$ dolt sql -q "INSERT INTO users VALUES ('druvv', 4224224242);"
$ dolt sql -q "SELECT * from users;"
+-------+------------+
| name  | phone      |
+-------+------------+
| druvv | 4224224242 |
+-------+------------+
$ dolt schema tags
+-------+--------+-------+
| table | column | tag   |
+-------+--------+-------+
| users | name   | 9815  |
| users | phone  | 15960 |
+-------+--------+-------+
$ dolt sql -q "ALTER TABLE users RENAME column phone to phone_number"
$ dolt schema tags
+-------+--------------+-------+
| table | column       | tag   |
+-------+--------------+-------+
| users | name         | 9815  |
| users | phone_number | 15960 |
+-------+--------------+-------+
```

When I merge the users table with the renamed column against another users table with the original columns names, the columns containing the phone numbers are matched by their tags. Their row values are merged.
```
$ dolt commit -am "initial data"
commit rqnmsovvbk8nokh2nphbaf4d2pslfasg (HEAD -> main)
Author: druvv <dhruv@dolthub.com>
Date:  Thu Jul 28 16:41:22 -0700 2022

        initial data
$ dolt checkout -b other
Switched to branch 'other'
$ dolt sql -q "INSERT INTO users VALUES ('vader', 9999999999)";
$ dolt commit -am "add vader"
commit dmfftn1quviend96n8ml1qpi7u77b8ul (HEAD -> other)
Author: druvv <dhruv@dolthub.com>
Date:  Thu Jul 28 16:42:23 -0700 2022

        add vader
$ dolt checkout main
$ dolt sql -q "ALTER TABLE users RENAME column phone to phone_number"
commit rinom96u0ilhutuh84gs812lqcmh88ak (HEAD -> main)
Author: druvv <dhruv@dolthub.com>
Date:  Thu Jul 28 16:45:16 -0700 2022

        rename column phone to phone_number
        
$ dolt merge other
[...]

PR #5922, opened 2023-05-10
Changing schema equality to not consider tag when comparing columns

fulghum:
Two tables can have identical schemas, but different column tags if the columns were created in different orders (e.g. a different sequence of alters to get to the same schema). When we try to merge those two tables, we see that there is not a common ancestor, so we require the schemas to be exactly identical in order to merge the data. This results in a confusing merge error for customers because we tell them the schemas are different, but they look exactly the same (except for the different column tags). 

This change removes column tags from equality checks when testing equality of two schemas and enables the case above to merge cleanly.

### DIFF
--- a/integration-tests/bats/column_tags.bats
+++ b/integration-tests/bats/column_tags.bats
@@ -344,15 +344,7 @@ DELIM
     dolt sql -q "ALTER TABLE target DROP COLUMN badCol;"
     dolt commit -Am "fixup"
 
-    run dolt schema tags
-    [[ $output =~ "| target | col1   | 14690 |" ]] || false
-
     dolt checkout main
-
-    run dolt schema tags
-    [ $status -eq 0 ]
-    [[ $output =~ "| target | col1   | 14649 |" ]] || false
-
     run dolt merge other -m "merge other into main"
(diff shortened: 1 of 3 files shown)


timsehn:
This should allow us to resolve #3963. I think that would make it so Dolt column tags are a table storage detail not a customer visible thing.

integration-tests/bats/column_tags.bats
```diff
@@ -353,14 +353,6 @@ DELIM
     [ $status -eq 0 ]
     [[ $output =~ "| target | col1   | 14649 |" ]] || false
```
zachmu, line 354: Are these `tags` calls still relevant?

zachmu (approved):
LGTM!

integration-tests/bats/column_tags.bats
fulghum: oooh... good point. they aren't relevant to this test anymore. I'll yank 'em out!

timsehn: Maybe make a new test of `dolt schema tags` if it's not tested anywhere else?

PR #5923, opened 2023-05-10
Prepend table name to auto-generated constraint check names.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/5915


jennifersp (approved):
Looks great!

PR #5940, opened 2023-05-11
go: sqle: cluster: Implement dolt_cluster_ack_writes_timeout_secs, a way to block COMMIT ack until the commit is replicated to cluster standbys.

reltuk:
Setting this system variable to a non-zero value on a primary replica in a sql-server cluster will cause dolt to block a SQL client performing a commit until that client's commit is fully replicated to the replicas.

If there is a timeout, a session warning is placed on the SQL connection.

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -1079,7 +1079,7 @@ func (ddb *DoltDB) GetRefsOfTypeByNomsRoot(ctx context.Context, refTypeFilter ma
 
 // NewBranchAtCommit creates a new branch with HEAD at the commit given. Branch names must pass IsValidUserBranchName.
 // Silently overwrites any existing branch with the same name given, if one exists.
-func (ddb *DoltDB) NewBranchAtCommit(ctx context.Context, branchRef ref.DoltRef, commit *Commit) error {
+func (ddb *DoltDB) NewBranchAtCommit(ctx context.Context, branchRef ref.DoltRef, commit *Commit, replicationStatus *ReplicationStatusController) error {
 	if !IsValidBranchRef(branchRef) {
 		panic(fmt.Sprintf("invalid branch name %s, use IsValidUserBranchName check", branchRef.String()))
 	}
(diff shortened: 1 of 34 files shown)


go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -1184,7 +1184,7 @@ func (ddb *DoltDB) deleteRef(ctx context.Context, dref ref.DoltRef) error {
 		}
 	}
 
-	_, err = ddb.db.Delete(ctx, ds)
+	_, err = ddb.db.withReplicationStatusController(replicationStatus).Delete(ctx, ds)
```
zachmu, line 1187: Occurs to me that this is the better overall solution to use everywhere (rather than params in affected methods), but can be punted on for now I guess

go/libraries/doltcore/doltdb/hooksdatabase.go
```diff
@@ -54,27 +55,63 @@ func (db hooksDatabase) SetCommitHookLogger(ctx context.Context, wr io.Writer) h
...
 			go func() {
 				defer wg.Done()
-				err = hook.Execute(ctx, ds, db)
+				f, err := hook.Execute(ctx, ds, db)
 				if err != nil {
 					hook.HandleError(ctx, err)
 				}
+				if rsc != nil {
+					rsc.Wait[i+ioff] = f
+					if nf, ok := hook.(interface {
```
zachmu, line 91: Probably worth defining this interface formally rather than inline here

go/libraries/doltcore/sqle/cluster/commithook.go
```diff
@@ -53,6 +53,20 @@ type commithook struct {
 	// commithooks are caught up with replicating to the standby.
 	waitNotify func()
 
+	// This is a slice of notification channels maintained by the
+	// commithook. The semantics are:
+	// 1. All accesses to |successChs| must happen with |mu| held.
+	// 2. There maybe be |0| or more channels in the slice.
```
zachmu, line 59: Typo

go/libraries/doltcore/sqle/cluster/commithook.go
```diff
@@ -53,6 +53,20 @@ type commithook struct {
...
+	// 3. As a reader, if |successChs| is non-empty, you should just read a value, for example, |successChs[0]| and use it. All entries will be closed at the same time. If |successChs| is empty when you need a channel, you should add one to it.
+	// 4. If you read a channel out of |successChs|, that channel will be closed on the next successful replication attempt. It will not be closed before then.
+	successChs []chan struct{}
+
+	// If this is true, the waitF returned by Execute() will fast fail if
+	// we are not already caught up, instead of blocking on a successCh
+	// actually indicated we are caught up. This is set to by a call to
+	// NotifyWaitFailed(), an optional interface on CommitHook.
+	circuitBreakerOpen bool
```
zachmu, line 68: Would make this name more explicit, like `failFastOnStale`

zachmu: `failFastWhenBehind`?

zachmu (approved):
LGTM

My one major comment is that this change introduces a bunch of param clutter and potential errors at the DoltDB layer. This is kind of an ongoing theme with that interface, maybe add a TODO for this.

PR #5943, opened 2023-05-11
Don't panic when in detached head state.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/5871

Many commands currently panic when in a detached head state, such as commit (https://github.com/dolthub/dolt/issues/5744), checkout (https://github.com/dolthub/dolt/issues/5839), and many others. This turns those panics into errors.


go/libraries/doltcore/sqle/database.go
```diff
@@ -654,6 +661,9 @@ func (db Database) GetWorkingSet(ctx *sql.Context) (*doltdb.WorkingSet, error) {
 	if !ok {
 		return nil, fmt.Errorf("no root value found in session")
 	}
+	if dbState.WorkingSet == nil {
```
zachmu, line 664: Don't know if we actually want this behavior. In general an accessor like this shouldn't second-guess what a client is going to do with the things it returns. It's a valid state for a session to not have a working set, and returning a nil value here is fine, it doesn't need to be an error. Generally speaking we don't want the presence of an error to signal needed changes in business logic, only actual errors that mean something went wrong, which isn't the case here.

go/libraries/doltcore/sqle/dsess/session_state_adapter.go
```diff
@@ -61,29 +61,30 @@ func (s SessionStateAdapter) GetRoots(ctx context.Context) (doltdb.Roots, error)
 	return state.GetRoots(), nil
 }
 
-func (s SessionStateAdapter) CWBHeadRef() ref.DoltRef {
+func (s SessionStateAdapter) CWBHeadRef() (ref.DoltRef, error) {
 	workingSet, err := s.session.WorkingSet(sql.NewContext(context.Background()), s.dbName)
 	if err != nil {
-		// TODO: fix this interface
-		panic(err)
+		return nil, err
```
zachmu, line 67: This is where the error should originate, not session.WorkingSet

zachmu (approved):
LGTM, one major comment and a few nits

go/libraries/doltcore/sqle/dsess/session_state_adapter.go
nicktobey: In principle, I agree with your comments. It should be up to the client to decide whether the lack of a working set is valid, and up the api to return results in a form that require clients to consider this.

The trouble is that a lot of places call session.Workingset without calling CWBHeadRef, and don't do any checking on the result, and adding checks to all of them would complicate all those commands while doing nothing to prevent this same mistake from happening if we add another command that calls session.Workingset.

The long term solution is to change the interface to one where instead of clients requesting the working set, they request what they want the working set for, and the and session returns the best result. But until we have that, the priority is to prevent panics and make it harder to introduce panics in the future. This does that.

go/cmd/dolt/commands/checkout.go
```diff
@@ -195,7 +202,11 @@ func checkoutNewBranch(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.Ar
 		if !remoteOk {
 			return nil
 		}
-		verr = SetRemoteUpstreamForBranchRef(dEnv, remoteName, remoteBranchName, dEnv.RepoStateReader().CWBHeadRef())
+		headRef, err := dEnv.RepoStateReader().CWBHeadRef()
```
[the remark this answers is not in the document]
nicktobey: Fixed.

go/cmd/dolt/commands/push.go
```diff
@@ -96,16 +96,20 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		var verr errhand.VerboseError
 		switch err {
 		case env.ErrNoUpstreamForBranch:
-			currentBranch := dEnv.RepoStateReader().CWBHeadRef()
-			remoteName := "<remote>"
-			if defRemote, verr := env.GetDefaultRemote(dEnv.RepoStateReader()); verr == nil {
-				remoteName = defRemote.Name
+			currentBranch, err := dEnv.RepoStateReader().CWBHeadRef()
+			if err != nil {
+				verr = errhand.BuildDError("fatal: The current branch could not be identified").Build()
```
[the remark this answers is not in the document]
nicktobey: Fixed.

go/cmd/dolt/commands/branch.go
```diff
@@ -172,7 +175,11 @@ func printBranches(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
 		}
 
 		if verbose {
-			cm, err := dEnv.DoltDB.Resolve(ctx, cs, dEnv.RepoStateReader().CWBHeadRef())
+			headRef, err := dEnv.RepoStateReader().CWBHeadRef()
```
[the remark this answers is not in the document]
nicktobey: Fixed.

go/libraries/doltcore/sqle/dsess/session_state_adapter.go
zachmu: I'm fine with this for now, please add a TODO to this effect in the method doc for Database.GetWorkingSet()

nicktobey: Done

timsehn:
Where we at with this one?

PR #5949, opened 2023-05-15
go: sqle: cluster: commithook: Periodically heartbeat to a standby when we are primary. This allows replication_lag on the standby to more accurately reflect the possible drift locally.

reltuk:


go/libraries/doltcore/sqle/cluster/commithook.go
```diff
@@ -175,6 +182,37 @@ func (h *commithook) primaryNeedsInit() bool {
...
+	defer func() {
+		if h.cancelReplicate != nil {
+			h.cancelReplicate()
+		}
+		h.cancelReplicate = nil
+	}()
+	h.mu.Unlock()
+	datasDB := doltdb.HackDatasDatabaseFromDoltDB(destDB)
+	cs := datas.ChunkStoreFromDatabase(datasDB)
+	cs.Commit(ctx, head, head)
```
zachmu, line 212: Kinda surprised this works tbh, seems a bit fragile

Can we get the underlying commit hook and ping it instead?

zachmu (approved):
This seems fine tactically but the interface is a bit wonky and there are not tests. Follow up maybe?

Or a TODO in any case

reltuk:
> This seems fine tactically but the interface is a bit wonky and there are not tests. Follow up maybe?
> 
> Or a TODO in any case

Took a pass at an integration test here: https://github.com/dolthub/dolt/pull/5949/files#diff-1425872ab3c293f81d61ed0acbd7df134e10c96c2a38a23ef8aa60557ef5cee2. Let me know if you have any thoughts :).

PR #5951, opened 2023-05-15
Handle unresolved/resolved foreign keys when determining equality

fulghum:
When foreign keys are unresolved, the data on which columns they point to in the child table and the parent table is stored differently, and requires extra work to "resolve" the columns before we can test for equality. We currently do not always calculate equality correctly, which can result in confusing diff output (e.g. https://github.com/dolthub/dolt/issues/5934).

This PR adds a new `Equals` method on `ForeignKey` that takes in enough data to resolve FK column references, if needed, and test equality for any two foreign keys.


Hydrocharged (approved):
LGTM!

PR #5957, opened 2023-05-16
Add support for `dolt table import -a`

nicktobey:
I ended up going with @zachmu's suggestion of making this a new operation type like `-u`, `-c`, and `-r` instead of a separate flag after I realized that `-c` and `-r` already disallow overwriting fields and it didn't really make sense to add a flag that could only be used with `-u`.

This is less ambitious than what was outlined in https://github.com/dolthub/dolt/issues/5936. Notably, I made a big deal in the issue about how you can import the same file multiple times or files that contain previously imported rows without issue. This was a mostly a pet feature that I erroneously assumed would be simple to implement. Given that it wasn't a requirement, after some exploration I ultimately decided to drop it.

This should be enough to unblock the customer; we talked about implementing a `--strict` flag too but I don't think it's necessary for their use case.

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/import.go
+++ b/go/cmd/dolt/commands/tblcmds/import.go
@@ -529,10 +529,13 @@ func move(ctx context.Context, rd table.SqlRowReader, wr *mvdata.SqlEngineTableW
 	badRowCB := func(row sql.Row, err error) (quit bool) {
 		// record the first error encountered unless asked to ignore it
 		if row != nil && rowErr == nil && !options.contOnErr {
-			if ue, ok := err.(sql.WrappedInsertError).Cause.(*errors.Error).Cause().(sql.UniqueKeyError); ok {
-				rowErr = fmt.Errorf("row %s would be overwritten by %s: %w", sql.FormatRow(ue.Existing), sql.FormatRow(row), err)
-			} else {
-				rowErr = fmt.Errorf("A bad row was encountered: %s: %w", sql.FormatRow(row), err)
+			rowErr = fmt.Errorf("A bad row was encountered: %s: %w", sql.FormatRow(row), err)
+			if wie, ok := err.(sql.WrappedInsertError); ok {
+				if e, ok := wie.Cause.(*errors.Error); ok {
+					if ue, ok := e.Cause().(sql.UniqueKeyError); ok {
+						rowErr = fmt.Errorf("row %s would be overwritten by %s: %w", sql.FormatRow(ue.Existing), sql.FormatRow(row), err)
+					}
+				}
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -520,7 +529,11 @@ func move(ctx context.Context, rd table.SqlRowReader, wr *mvdata.SqlEngineTableW
 	badRowCB := func(row sql.Row, err error) (quit bool) {
 		// record the first error encountered unless asked to ignore it
 		if row != nil && rowErr == nil && !options.contOnErr {
-			rowErr = fmt.Errorf("A bad row was encountered: %s: %w", sql.FormatRow(row), err)
+			if ue, ok := err.(sql.WrappedInsertError).Cause.(*errors.Error).Cause().(sql.UniqueKeyError); ok {
```
nicktobey, line 532: I'm not thrilled about this line, but I think the improved error messaging is worth it, and this way doesn't touch GMS.

integration-tests/bats/import-append-tables.bats
```diff
@@ -0,0 +1,78 @@
...
+setup() {
+    setup_common
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+@test "import-append-tables: disallow overwriting row during append" {
```
zachmu, line 13: These tests should verify the contents of the table after the import, not just the duplicate error handling

zachmu (approved):
LGTM, see comments

go/cmd/dolt/commands/tblcmds/import.go
nicktobey: I assumed that ok would be true if all the type conversions were successful, and ue would be nil if any of them failed. Turns out it doesn't work that way. Good to know.

integration-tests/bats/import-append-tables.bats
nicktobey: Fixed.

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -280,8 +284,12 @@ func validateImportArgs(apr *argparser.ArgParseResults) errhand.VerboseError {
 		return errhand.BuildDError("parameters %s and %s are mutually exclusive", schemaParam, primaryKeyParam).Build()
 	}
 
-	if !apr.Contains(createParam) && !apr.Contains(updateParam) && !apr.Contains(replaceParam) {
-		return errhand.BuildDError("Must include '-c' for initial table import or -u to update existing table or -r to replace existing table.").Build()
+	if !apr.ContainsAny(createParam, updateParam, replaceParam, appendParam) {
+		return errhand.BuildDError("Must include '-c' for initial table import or -u to update existing table or -a to append to existing table or -r to replace existing table.").Build()
```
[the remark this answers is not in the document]
nicktobey: Fixed.

go/cmd/dolt/commands/tblcmds/import.go
nicktobey: Fixed.

PR #5961, opened 2023-05-16
Use UTC timestamp in date comparison test.

macneale4:
Allows tests to run on laptops which aren't in UTC.


fulghum (approved):
Thanks for catching this!

PR #5966, opened 2023-05-16
Smart routing for sql command execution against running local servers

macneale4:
When `dolt sql` commands are executed against a database which is in used by a sql server, connect to it through the sql port. Currently this does not support authenticated transport. If the root user is still active, we'll use that be default. Otherwise, the `--user` flag can be used to specify a user, but no authentication token is supported. That will come in the next round of changes.

A step towards: https://github.com/dolthub/dolt/issues/3922

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -295,7 +295,12 @@ func sqlHandleVErrAndExitCode(queryist cli.Queryist, verr errhand.VerboseError, 
 			if _, ok := queryist.(*engine.SqlEngine); !ok {
 				// We are in a context where we are attempting to connect to a remote database. These errors
 				// are unstructured, so we add some additional context around them.
-				msg = fmt.Sprintf("Error connecting to remote database: \"%s\".", msg)
+				tmpMsg := `You've encountered a new behavior in dolt which is not fully documented yet.
+A local dolt server is using your dolt data directory, and in an attempt to service your request, we are attempting to 
+connect to it. That has failed. You should stop the server, or reach out to @macneale on https://discord.gg/gqr7K4VNKe
+for help.`
+				cli.PrintErrln(tmpMsg)
+				msg = fmt.Sprintf("A local server is running, and dolt is failing to connect. Error connecting to remote database: \"%s\".\n", msg)
 			}
 			cli.PrintErrln(msg)
 		}
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -234,7 +234,7 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 		fi, err := os.Stdin.Stat()
 		if err != nil {
 			if !osutil.IsWindows {
-				return HandleVErrAndExitCode(errhand.BuildDError("Couldn't stat STDIN. This is a bug.").Build(), usage)
+				return sqlHandleVErrAndExitCode(queryist, errhand.BuildDError("Couldn't stat STDIN. This is a bug.").Build(), usage)
 			}
 		} else {
 			isTty = fi.Mode()&os.ModeCharDevice != 0
@@ -247,11 +247,11 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 			isTty = false
 			input, err = os.OpenFile(fileInput, os.O_RDONLY, os.ModePerm)
 			if err != nil {
-				return HandleVErrAndExitCode(errhand.BuildDError("couldn't open file %s", fileInput).Build(), usage)
+				return sqlHandleVErrAndExitCode(queryist, errhand.BuildDError("couldn't open file %s", fileInput).Build(), usage)
 			}
 			info, err := os.Stat(fileInput)
 			if err != nil {
-				return HandleVErrAndExitCode(errhand.BuildDError("couldn't get file size %s", fileInput).Build(), usage)
+				return sqlHandleVErrAndExitCode(queryist, errhand.BuildDError("couldn't get file size %s", fileInput).Build(), usage)
 			}
 
 			// initialize fileReadProg global variable if there is a file to process queries from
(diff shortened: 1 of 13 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -568,12 +628,12 @@ func interceptSendMetrics(ctx context.Context, args []string) (bool, int) {
...
 	ap.SupportsString(commands.DataDirFlag, "", "directory", "Defines a directory whose subdirectories should all be dolt data repositories accessible as independent databases within. Defaults to the current directory.")
 	ap.SupportsString(commands.CfgDirFlag, "", "directory", "Defines a directory that contains configuration files for dolt. Defaults to `$data-dir/.doltcfg`. Will only be created if there is a change to configuration settings.")
 	ap.SupportsString(commands.PrivsFilePathFlag, "", "privilege file", "Path to a file to load and store users and grants. Defaults to `$doltcfg-dir/privileges.db`. Will only be created if there is a change to privileges.")
 	ap.SupportsString(commands.BranchCtrlPathFlag, "", "branch control file", "Path to a file to load and store branch control permissions. Defaults to `$doltcfg-dir/branch_control.db`. Will only be created if there is a change to branch control permissions.")
+	ap.SupportsString(commands.UseDbFlag, "", "database", "The name of the database to use when executing SQL queries. Defaults to the first database alphabetically.")
```
fulghum, line 636: Just wondering... not sure on my opinion yet... if there is a db rooted at `datadir`, should we default to that one? That matches current behavior, but more importantly also seems like it may be more likely to be what a customer wants/expects (i.e. if they have explicitly provided `datadir` or if they have moved their CWD into a database dir). Picking a different DB by alphabetic sorting and choosing the first when I'm sitting in a db dir seems a little unexpected.

fulghum (commented):
Generally looking good; the command startup feels like it's getting complex, but I guess that's unavoidable to some extent since we're including more logic. I held off from approving just to make sure I understood a couple questions/thoughts first. 

Looks like there's just one failing BATS test left, too! 🎉

go/libraries/doltcore/env/multi_repo_env.go
```diff
@@ -58,41 +58,50 @@ type MultiRepoEnv struct {
...
+}
+
 // MultiEnvForDirectory returns a MultiRepoEnv for the directory rooted at the file system given. The doltEnv from the
 // invoking context is included. If it's non-nil and valid, it will be included in the returned MultiRepoEnv, and will
 // be the first database in all iterations.
 func MultiEnvForDirectory(
 	ctx context.Context,
 	config config.ReadWriteConfig,
-	fs filesys.Filesys,
+	fs filesys.Filesys, // fs which is the root of the multi-env. This is essentially the --data-dir flag or cwd.
```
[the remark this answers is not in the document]
macneale4: I think renaming it to `dataDirFS` is more appropriate.

go/cmd/dolt/commands/utils.go
```diff
@@ -78,17 +78,22 @@ func MaybeGetCommitWithVErr(dEnv *env.DoltEnv, maybeCommit string) (*doltdb.Comm
...
+		return nil, verr
 	}
 	return cli.NewCliContext(argparser.NewEmptyResults(), lateBind)
 }
 
 // BuildSqlEngineQueryist Utility function to build a local SQLEngine for use interacting with data on disk using
 // SQL queries. ctx and dEnv must be non-nil. apr can be nil.
-func BuildSqlEngineQueryist(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgParseResults) (cli.LateBindQueryist, errhand.VerboseError) {
-	if ctx == nil || dEnv == nil || apr == nil {
+func BuildSqlEngineQueryist(ctx context.Context, dEnv *env.DoltEnv, mrEnv *env.MultiRepoEnv, apr *argparser.ArgParseResults) (cli.LateBindQueryist, errhand.VerboseError) {
```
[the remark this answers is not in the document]
macneale4: Agreed. I was able simplify a little.

go/cmd/dolt/dolt.go
macneale4: That's actually the behavior. Updating the incorrect documentation.

go/cmd/dolt/commands/sql.go
```diff
@@ -261,30 +262,54 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
+// sqlHandleVErrAndExitCode is a helper function to print errors to the user. Currently, the Queryist interface is used to
+// determine if this is a local or remote execution. This is hacky, and too simplistic. We should possibly add an error
+// messaging interface to the CliContext.
+func sqlHandleVErrAndExitCode(queryist cli.Queryist, verr errhand.VerboseError, usage cli.UsagePrinter) int {
+	if verr != nil {
+		if msg := verr.Verbose(); strings.TrimSpace(msg) != "" {
+			if _, ok := queryist.(*engine.SqlEngine); !ok {
+				// We are in a context where we are attempting to connect to a remote database. These errors
+				// are unstructured, so we add some additional context around them.
+				msg = fmt.Sprintf("Error connecting to remote database: \"%s\".", msg)
```
fulghum, line 298: Is this only for the case when we are connecting to a local database via a TCP connection? Just wondering if we can avoid saying "remote" and be more specific in this error message. I think there's a good chance that customers trying to use `dolt sql` will get confused by this error message and need to ask for support since they aren't trying to connect off box.

macneale4: I updated the message to acknowledge that this is undocumented, and to reach out to me for help:

```
You've encountered a new behavior in dolt which is not fully documented yet.
A local dolt server is using your dolt data directory, and in an attempt to service your request, we are attempting to
connect to it. That has failed. You should stop the server, or reach out to @macneale on https://discord.gg/gqr7K4VNKe
for help.
A local server is running, and dolt is failing to connect. Error connecting to remote database: "Error 1045: User not found 'root'".
```

PR #5968, opened 2023-05-16
Transaction and DB name resolution changes

zachmu:
This change introduces two major behavior changes to how dolt manages database names and session state in transactions:

1) It's now possible to edit different branch heads and databases without explicitly switching to a new working set via `dolt_checkout()`. E.g., this session updates the branch `b1`:

```sql
use mydb;
insert into `mydb/b1`.t1 values (1);
commit;
```

These database-qualified `INSERT` / `UPDATE` etc. statements used to silently drop data if the database named wasn't the current database, in some cases. They now work correctly in all cases. We still fail transactions that attempt to make changes to more than one working set for now, but this restriction will be removed in a future revision.

2) The behavior of `dolt_checkout()` has changed. It is no longer an error to call `dolt_checkout()` with a dirty working set. `USE mydb/branch` is now almost exactly equivalent to `dolt_checkout('branch')`. `dolt_checkout()` now has the side-effect of setting the session's current DB name to the base name of the database. E.g. after `use mydb/branch`, `select database(), active_branch()` returns `mydb/branch, branch`. If you then `call dolt_checkout('branch2')`, then the same query returns `mydb, branch2`. `dolt_checkout` is different from `USE` statements in that it changes which branch the unqualified database name (`mydb`) resolves to for this session.


zachmu:
Tests not fully passing yet, mostly due to the expectations of tests no longer being true. But the semantics in this PR are what I intend to merge.

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
```diff
@@ -2370,4 +2380,322 @@ var MultiDbTransactionTests = []queries.ScriptTest{
...
+			"call dolt_commit('-am', 'new table')",
+			"call dolt_branch('b1')",
+			"set autocommit = on", // unnecessary but make it explicit
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query: "insert into `mydb/b1`.t1 values (1)",
+				Expected: []sql.Row{
+					{types.OkResult{RowsAffected: 1}},
+				},
```
fulghum, line 2397: If `@@dolt_transaction_commit` is enabled, does this commit on `mydb/b1` also create a dolt commit on that branch? Might be worth testing somewhere in here.

fulghum (commented):
Wow, lots of awesome changes and fixes and improvements in here! Changes looked very clean overall; I just had a few minor comments. The biggest feedback I had was that the new behavior you described where you can commit changes from a non-checked out branch head seemed pretty surprising. I wasn't sure if that also means you can `dolt_commit()` to create commits on non-checked out branches, which seems weird. 

I just have two more files to look at (session and some tests), but figured I'd send these comments over in case you were waiting on them. I'll finish up those last two files tomorrow morning when my brain is fresh again.

go/libraries/doltcore/sqle/enginetest/dolt_transaction_queries.go
```diff
@@ -2370,4 +2380,322 @@ var MultiDbTransactionTests = []queries.ScriptTest{
...
+				Expected: []sql.Row{},
+			},
+			{
+				Query:    "select * from `db1/b1`.t1 order by a",
+				Expected: []sql.Row{{1}, {2}},
+			},
+		},
+	},
+	{
+		Name: "committing to more than one branch at a time",
```
fulghum, line 2613: It seems like a sharp edge to allow people to modify multiple branches but not let them SQL commit to multiple branches. If someone got into this situation without knowing the rules, it might be hard for them to unwind their DB and undo the change, unless they checked every branch. Might be painful for customers with many many branches. 

This is the cross branch commit you mentioned as a fast-follow? (I think I originally got confused in the PR description between Dolt commits and SQL commits)

fulghum (approved):
Changes look great, but I did find a couple of small bugs while I was playing with it. 

I initially got a little confused on the described behavior change, specifically on Dolt commits versus SQL commits. I see now that we are continuing to restrict Dolt commits to the currently checked out branch (so disregard that earlier question) and it is only SQL commits we wanna to make work across branches/dbs as the next step here, which sounds great.

zachmu:
#benchmark

go/libraries/doltcore/sqle/database.go
```diff
@@ -836,7 +858,7 @@ func (db Database) CreateIndexedTable(ctx *sql.Context, tableName string, sch sq
 	return db.createIndexedSqlTable(ctx, tableName, sch, idxDef, collation)
 }
 
-// Unlike the exported version CreateTable, createSqlTable doesn't enforce any table name checks.
+// Unlike the exported version CreateTable, createSqlTable doesn't enforce any table baseName checks.
```
fulghum, line 861: (minor) Might as well make this follow godoc format if you're going to put it in front of the method. Same comment for line 903.

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -223,149 +268,89 @@ func (d *DoltSession) ValidateSession(ctx *sql.Context, dbName string) error {
...
-		// if the current head is not found, the branch was force deleted, so use nil working set.
-		if errors.Is(err, doltdb.ErrBranchNotFound) {
-			return ErrCurrentBranchDeleted
-		} else if err != nil {
-			return err
-		}
-	} else if err != nil {
-		return err
-	}
 	return nil
```
fulghum, line 271: (super minor) You could just return `d.validateErr` directly instead of checking for `nil`

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -223,149 +268,89 @@ func (d *DoltSession) ValidateSession(ctx *sql.Context, dbName string) error {
...
-	} else if err != nil {
+	tx, err := NewDoltTransaction(ctx, txDbs, tCharacteristic)
+	if err != nil {
 		return nil, err
 	}
 
-	// logrus.Tracef("starting transaction with working root %s", ws.WorkingRoot().DebugString(ctx, true))
+	// The engine sets the transaction after this call as well, but since we begin accessing data below, we need to set
+	// this now to avoid seeding the session state with stale data in some cases. The duplication is harmless since the
+	// code below cannot error, but it would to get rid of
```
fulghum, line 323: (minor) looks like the end of this comment go cut off? Should this be  "it would be hard to get rid of"?

fulghum (approved):
Changes look good. 👍 

While reviewing... I was wondering if `use <db>/<branch>` *should* actually checkout the branch so that the next time `use <db>` is called, it goes to that branch. I haven't thought through it deeply, but it seems like that could be more intuitive to customers, instead of trying to teach all customers that `use <db>/<branch>` is subtly different from dolt_checkout in that one respect. The MySQL `use` docs state: "The USE statement tells MySQL to use the named database as the default (current) database for subsequent statements". It seems like that extends nicely to branches so that the branch `use`d is the default branch for subsequent statements with that database.

timsehn:
Latency benchmarks do not look good. I think I prefer correctness over speed. Can you spend a day trying to figure out if it's slower for an easy reason? If not, ship it and we'll try to win the performance back.

zachmu:
Yes, I'll see if there are any easy perf wins before checking this in.

zachmu:
#benchmark

go/libraries/doltcore/sqle/enginetest/dolt_server_test.go
```diff
@@ -235,11 +235,11 @@ var DoltBranchMultiSessionScriptTests = []queries.ScriptTest{
 			},
 			{
 				Query:          "/* client a */ select name from dolt_branches;",
-				ExpectedErrStr: "Error 1105: current branch has been force deleted. run 'USE <database>/<branch>' to checkout a different branch, or reconnect to the server",
+				ExpectedErrStr: "Error 1105: database not found: dolt/branch1",
```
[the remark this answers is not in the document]
zachmu: Yeah, this error was coming from a very specific code path that got altered beyond recognition, agree it would have been nice to save but it's such an edge case

go/libraries/doltcore/sqle/dsess/session.go
zachmu: I think it was supposed to be "but it would be nice to get rid of" but I rephrased to make it clearer

go/libraries/doltcore/sqle/dprocedures/dolt_merge.go
```diff
@@ -106,6 +106,9 @@ func doDoltMerge(ctx *sql.Context, args []string) (int, int, error) {
 			return noConflictsOrViolations, threeWayMerge, err
 		}
 
+		// Because this commit happens outside the engine's control, we need to manually clear the transaction
+		ctx.SetTransaction(nil)
+
```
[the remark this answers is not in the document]
zachmu: Good suggestion, now the case

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -285,16 +251,28 @@ func checkoutNewBranch(ctx *sql.Context, dbName string, dbData env.DbData, apr *
 	} else if autoSetupMerge, err := loadConfig(ctx).GetString("branch.autosetupmerge"); err != nil || autoSetupMerge != "false" {
 		remoteName, remoteBranchName = actions.ParseRemoteBranchName(startPt)
 		refSpec, err = ref.ParseRefSpecForRemote(remoteName, remoteBranchName)
-		if err != nil {
-			return nil
-		}
-		err = env.SetRemoteUpstreamForRefSpec(dbData.Rsw, refSpec, remoteName, ref.NewBranchRef(remoteBranchName))
-		if err != nil {
-			return err
+		if err == nil {
```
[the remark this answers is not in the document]
zachmu: Intended in this case.

PR #5981, opened 2023-05-17
[no-release-notes] Remove the direct package dependency from `dsess` to `merge`

fulghum:
The `dsess` package currently has a direct reference to the `merge` package, which prevents us from using `dsess` directly from within the `merge` package, which is needed as we start to analyze/evaluate check expressions and column default value expressions as part of merges. This PR introduces a new interface to decouple `dsess` from needing to directly reference `merge`.


zachmu (commented):
This isn't a terrible solution but it's also not great.

Architecturally, anything in the `doltcore` namespace should generally be accessible by the sql packages, including `dsess`.  The reverse I don't expect to be true. What are you trying to do in the `merge` package that requires the `dsess` package? Is possible to do this change in reverse, i.e. define interfaces in `doltcore/*` that are satisfied by `dsess.Session` and use those in the `merge` package?

If you can show me a draft of what you're trying to do it might make more sense why this is the best path forward.

fulghum:
Closing this one... found another way to access the functionality I need from the analyzer for this, without having to access `dsess`.

Issue #5983, opened 2023-05-18
LOAD Data into a table with column defaults panics


timsehn:
Schema and data repro is private. 

Will try for a minimal repro.

```
2023-05-17T18:52:49-07:00 DEBUG [conn 2] Starting query {connectTime=2023-05-17T18:51:04-07:00, connectionDb=test_subra, query=LOAD DATA local infile 'foo.csv' into table foo FIELDS TERMINATED BY ',' LINES TERMINATED BY '\n' IGNORE 1 LINES}
panic: UnresolvedColumnDefault is a placeholder node, but Eval() was called

goroutine 23 [running]:
github.com/dolthub/go-mysql-server/sql.UnresolvedColumnDefault.Eval(...)
	/Users/timsehn/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.15.1-0.20230517042856-2405a8d89854/sql/columndefault.go:252
github.com/dolthub/go-mysql-server/sql.(*ColumnDefaultValue).Eval(0xc00012c690, 0xc000274280?, {0xc0004a9b20?, 0x31d78b8?, 0xc000f8fab8?})
	/Users/timsehn/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.15.1-0.20230517042856-2405a8d89854/sql/columndefault.go:73 +0x3e
github.com/dolthub/go-mysql-server/sql/rowexec.loadDataIter.Next({0xc00023e000, {0x31f0c50, 0xc00063bb80}, {0x31d78b8, 0xc000f8fab8}, 0x0, {0xc0001feb60, 0xe, 0xe}, {0x437fa60, ...}, ...}, ...)
	/Users/timsehn/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.15.1-0.20230517042856-2405a8d89854/sql/rowexec/ddl_iters.go:89 +0x288
github.com/dolthub/go-mysql-server/sql/rowexec.(*projectIter).Next(0xc000276180, 0x0?)
	/Users/timsehn/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.15.1-0.20230517042856-2405a8d89854/sql/rowexec/rel_iters.go:329 +0x2f
github.com/dolthub/go-mysql-server/sql/rowexec.(*insertIter).Next(0xc000572820, 0xc00057a780)
[...]

Issue #5982, opened 2023-05-18
LOAD DATA does not support replace or ignore options


timsehn:
```
$ mysql -h"timsehn-clone-baby-clone.dbs.hosted.doltdb.com" -u"f394um477mw7cjrw" -p"<redacted>" -e "set global local_infile=1; LOAD DATA local infile 'out.csv' replace into table clone_baby_clone.t"
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1105 (HY000) at line 1: syntax error at position 41 near 'replace'
$ mysql -h"timsehn-clone-baby-clone.dbs.hosted.doltdb.com" -u"f394um477mw7cjrw" -p"<redacted>" -e "set global local_infile=1; LOAD DATA local infile 'out.csv' ignore into table clone_baby_clone.t"
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1105 (HY000) at line 1: syntax error at position 40 near 'ignore'
```

timsehn:
Another customer ran into this this morning.

timsehn:
GMS PR: https://github.com/dolthub/go-mysql-server/pull/1812

timsehn:
@tbantle22 is working on this now.

tbantle22:
This is in the 1.3.0 dolt release https://github.com/dolthub/dolt/releases/tag/v1.3.0

PR #5985, opened 2023-05-18
`LOAD DATA` bats tests improvements

timsehn:
A customer ran into some issues using `LOAD DATA` with Hosted Dolt. Added some additional testing in response. Pleasantly surprised with our compatibility. Adds skipped tests for #5982 and #5983.


fulghum (approved):
Thanks for the help testing! 🙏

PR #6003, opened 2023-05-22
Add new implementation of ParseGlobalArgs

nicktobey:
The main benefit of this new implementation is that it won't incorrectly parse situations where a global arg value (such as username) matches a dolt subcommand. (This would have been especially problematic when we got around to converting `dolt admin`.)

This also adds a nifty little burndown chart of commands that still need to be converted from using dEnv to using CliContext, in order to determine whether to generate the CliContext for a command. It'll be extra satisfying to remove elements from that list as we convert them.


go/libraries/utils/argparser/parser.go
```diff
@@ -281,90 +308,103 @@ func (ap *ArgParser) Parse(args []string) (*ArgParseResults, error) {
...
+		return nil, ap.TooManyArgsErrorFunc(list)
+	}
 
-			results[opt.Name] = ""
-		}
+	return &ArgParseResults{results, list, ap}, nil
+}
 
-		opt, value := ap.matchValueOption(rest, isLongFormFlag)
+func (ap *ArgParser) parseToken(args []string, i int, list []string, results map[string]string) (int, []string, map[string]string, error) {
```
macneale4, line 336: Can you add names to the output variables.  It's not clear what is returned, and I suspect all you need it to add descriptive names - i -> index, list -> ???, results -> foundArguments.

This is one of these cases where gDoc would be reasonable, but probably not required if you had more descriptive names.

go/libraries/utils/argparser/parser.go
```diff
@@ -271,6 +271,33 @@ func (ap *ArgParser) matchValueOption(arg string, isLongFormFlag bool) (match *O
...
+
+	i := 0
+	for ; i < len(args); i++ {
+		arg := args[i]
+
+		if len(arg) == 0 || arg == "--" {
+			continue
+		}
+
+		if arg[0] != '-' {
```
macneale4, line 286: Probably worth a one line comment here that no subcommand starts with '-', and that we assume this is a valid subcommand.

I don't see anyplace were we reject totally bogus commands - or having no command at all. Could remaining list be len(0) and line 531 in dolt.go result in a out of range error?

integration-tests/bats/no-repo.bats
```diff
@@ -363,3 +363,10 @@ NOT_VALID_REPO_ERROR="The current directory is not a valid dolt repository."
     [ "$output" == "" ]
     run kill -9 $PID # Kill process if it doesn't pass
 }
+
+@test "no-repo: check that we correctly parse args when connecting with a username that matches a subcommand" {
```
macneale4, line 367: please add a test (go or bats) which balks at bogus subcommands.

go/cmd/dolt/dolt.go
```diff
@@ -120,6 +118,67 @@ var doltSubCommands = []cli.Command{
 	stashcmds.StashCommands,
 	&commands.Assist{},
 }
+
+var subCommandsUsingDEnv = []cli.Command{
```
macneale4, line 122: Should these be flags on a function or field in the cli.Command struct? Seems a little heavy weight.

Is your plan here to remove commands until there are no more?

go/libraries/utils/argparser/parser.go
nicktobey: Done

go/cmd/dolt/dolt.go
PavelSafronov: A flag or an empty interface would be a good approach here. Having a source of truth so far from the actual code can be problematic.

go/libraries/utils/argparser/parser.go
nicktobey: If there's no subcommand we return an error on line 299.

`remaining` always contains at least the subcommand, so out of range is not possible there.

I added a comment.

integration-tests/bats/no-repo.bats
nicktobey: Done.

go/cmd/dolt/dolt.go
```diff
@@ -470,7 +528,7 @@ The sql subcommand is currently the only command that uses these flags. All othe
 	}
 
 	var cliCtx cli.CliContext = nil
-	if initCliContext {
+	if initCliContext(remainingArgs[0]) {
```
PavelSafronov, line 531: Assign `remainingArgs[0]` to a named variable, so it's clear that this is the command name.

go/libraries/utils/argparser/parser.go
macneale4: Huge improvement! Bravo!

macneale4 (approved):

go/cmd/dolt/dolt.go
nicktobey: Done

PR #6011, opened 2023-05-23
Rewrite `dolt status` subcommand to use SQL queries to generate results.

PavelSafronov:
Rewrite `dolt status` subcommand to use SQL queries to generate results.

This change adds `remote` column to the `dolt_branches` table. This column identifies the remote for the branch.
This change adds `dolt_count_commits` function, which returns the number of commits between two commits.
Added more BATS tests for `dolt status`.

### DIFF
--- a/go/cmd/dolt/commands/status.go
+++ b/go/cmd/dolt/commands/status.go
@@ -19,11 +19,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/dolthub/go-mysql-server/sql/types"
 	"strconv"
 	"strings"
 
 	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/types"
 	"github.com/fatih/color"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
--- a/go/cmd/dolt/commands/status.go
+++ b/go/cmd/dolt/commands/status.go
@@ -33,6 +33,8 @@ import (
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
 )
 
+// printData structure is used to pass parsed status data to the print function
+// This ensures that the business logic is separate from the printing logic: first we parse the data, then we print it.
 type printData struct {
 	branchName,
 	remoteName,
@@ -112,12 +114,11 @@ func (cmd StatusCmd) Exec(ctx context.Context, commandStr string, args []string,
 	}
 
 	// get status information from the database
-	pd, err := cmd.createPrintData(err, queryist, sqlCtx, showIgnoredTables)
+	pd, err := createPrintData(err, queryist, sqlCtx, showIgnoredTables)
 	if err != nil {
 		return handleStatusVErr(err)
 	}
 
-	// print everything
 	err = printEverything(pd)
 	if err != nil {
 		return handleStatusVErr(err)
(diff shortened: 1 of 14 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -561,7 +561,7 @@ func splitArgsOnSubCommand(args []string) (globalArgs, subArgs []string, initCli
 		if _, ok := commandSet[arg]; ok {
 			// SQL is the first subcommand to support the CLIContext. We'll need a more general solution when we add more.
 			// blame, table rm, and table mv commands also depend on the sql command, so they are also included here.
-			initCliContext := "sql" == arg || "blame" == arg || "table" == arg
+			initCliContext := "sql" == arg || "blame" == arg || "table" == arg || "status" == arg
```
macneale4, line 564: Talk to nick about this. I think he has something better on the way

go/cmd/dolt/commands/status.go
```diff
@@ -63,215 +66,287 @@ func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
...
 		} else {
-			cli.Println(allMergedHeader)
+			workingTableNames[tableName] = true
 		}
 	}
 
-	n := printStagedDiffs(cli.CliOut, stagedTbls, true)
-	n, err = PrintDiffsNotStaged(ctx, dEnv, cli.CliOut, notStagedTbls, true, showIgnoredTables, n, as)
+	// get merge status
+	mergeRows, err := getRowsForSql(queryist, sqlCtx, "select * from dolt_merge_status;")
```
macneale4, line 132: There are a lot of separate calls to the server for this. Ideas on how we can reduce the number of round trips?

macneale4 (changes requested):
Not sure if all commands are going to require this many round trips. Let's discuss before shipping. Asked for one additional bats test too

go/cmd/dolt/dolt.go
PavelSafronov: Nick's PR has been approved. Once that's merged, I'll rebase my changes and this should no longer be necessary.

go/cmd/dolt/commands/status.go
PavelSafronov: Removed one call that was unnecessary.
From offline convo, it was decided that there can be improvements in this area, but those can wait until after we ship these features.

integration-tests/bats/status.bats
```diff
@@ -453,3 +456,36 @@ SQL
     mv .dolt/repo_state.backup .dolt/repo_state.json
     [ "$status" -eq 0 ]
 }
+
+@test "status: ignored tables show up" {
```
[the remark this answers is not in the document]
PavelSafronov: Adding.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +122,72 @@ teardown() {
     [[ "$output" =~ "defaultDB does not exist" ]] || false
 }
 
+@test "sql-local-remote: test 'status' and switch between server/no server" {
+    start_sql_server defaultDB
+
+    run dolt --verbose-engine-setup --user dolt status
```
macneale4, line 128: You should drop the --verbose-engine-setup flag since you aren't looking for the output from it (and I don't think you need to). Update three over uses below.

macneale4 (approved):

macneale4:
Looks like there are failures in `bats/constraint-violations.bats` immediately after calling `dolt status` - something isn't right.

go/cmd/dolt/dolt.go
macneale4: Rebase onto main. You'll get a conflict, but Nick's new ParseGlobalArgs method is better than my hack

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +121,72 @@ teardown() {
     [[ "$output" =~ "defaultDB does not exist" ]] || false
 }
 
+@test "sql-local-remote: test 'status' and switch between server/no server" {
+    start_sql_server defaultDB
+
+    run dolt --user dolt status
+    [ "$status" -eq 0 ] || false
+    [[ "$output" =~ "On branch main" ]] || false
```
macneale4, line 129: I think a test which would be easier to read and more correct would be to take all of the output from the remote invocation, and compare it with all the output of the local invocation. The point of the test is to verify that the two modes produce the same output, not to make sure it reads like status

integration-tests/bats/status.bats
```diff
@@ -453,3 +456,35 @@ SQL
...
+    dolt sql <<SQL
+CREATE TABLE t (pk int PRIMARY KEY);
```
macneale4, line 462: can you use identifiers which are 3 chars or more? Easier for future maintainers to search/replace. I know we do single char vars all over the place - just trying to raise the bar one change at a time :-)

go/cmd/dolt/commands/status.go
```diff
@@ -63,215 +96,574 @@ func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
...
 
-	staged, notStaged, err := diff.GetStagedUnstagedTableDeltas(ctx, roots)
+	// get status information from the database
+	pd, err := cmd.createPrintData(err, queryist, sqlCtx, showIgnoredTables)
 	if err != nil {
 		return handleStatusVErr(err)
 	}
 
-	ws, err := dEnv.WorkingSet(ctx)
+	// print everything
```
macneale4, line 120: This is not a useful comment. It's clear that's what happening by the function name

go/cmd/dolt/commands/status.go
```diff
@@ -16,22 +16,49 @@ package commands
...
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
-	"github.com/dolthub/dolt/go/libraries/doltcore/diff"
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
-	"github.com/dolthub/dolt/go/libraries/doltcore/merge"
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
 )
 
+type printData struct {
```
macneale4, line 36: Brief godoc here would help the reader know the purpose of this structure. Not essential, but when combined with the undocuments createPrintData function, I'm forced to infer that there is a stage where you read all the essential data, then follow that up with printing all the data.

I know about this staging because of comments in the Exec method. My style is to flip this on it's head, and put most comments outside of the code on structure and function definitions. This seems to be the pattern followed on the dolt code I've been exposed to

go/cmd/dolt/commands/status.go
```diff
@@ -63,215 +96,574 @@ func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
...
+		mergeActive:               mergeActive,
+		stagedTables:              stagedTables,
+		unstagedTables:            unstagedTables,
+		untrackedTables:           untrackedTables,
+		unmergedTables:            unmergedTables,
+		ignoredTables:             ignoredTables,
+		constraintViolationTables: constraintViolationTables,
+		schemaConflictTables:      schemaConflictTables,
+		ignorePatterns:            ignorePatterns,
+		dataConflictTables:        dataConflictTables,
```
macneale4, line 408: This function is 200 lines long. Possible to break into bite size chunks?

go/libraries/doltcore/sqle/dprocedures/init.go
```diff
@@ -30,6 +30,7 @@ var DoltProcedures = []sql.ExternalStoredProcedureDetails{
 	{Name: "dolt_commit", Schema: stringSchema("hash"), Function: doltCommit},
 	{Name: "dolt_commit_hash_out", Schema: stringSchema("hash"), Function: doltCommitHashOut},
 	{Name: "dolt_conflicts_resolve", Schema: int64Schema("status"), Function: doltConflictsResolve},
+	{Name: "dolt_count_commits", Schema: int64Schema("ahead", "behind"), Function: doltCountCommits},
```
macneale4, line 33: Is this a public procedure? Do we need to document this somewhere for user consumption?

macneale4 (changes requested):
This was a lot of work! Broad feedback that I want to give you you should factor your code into smaller functions. There is a pattern of long functions with comments  to indicate processing steps. My feedback is to drop the comments, and convert into single use functions with descriptive names. Other than that, I'm happy with how this is coming. Not a trivial command at all.

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Adding the comparison between local and remote executions.

integration-tests/bats/status.bats
PavelSafronov: Using (slightly) more descriptive identifiers.

go/cmd/dolt/commands/status.go
PavelSafronov: Splitting out most of the "gathering" logic into smaller functions.

go/libraries/doltcore/sqle/dprocedures/init.go
PavelSafronov: Yup, this is a public procedure. We'll need to add documentation for it to https://github.com/dolthub/docs
I'll be opening a PR for docs after this PR is merged.

integration-tests/bats/sql-local-remote.bats
macneale4: You can go further and drop all of the line - line validation. Up to you.

PavelSafronov: I think there's value to doing at least one check of the contents, so I'll remove the repeat checks.

integration-tests/bats/status.bats
macneale4: Looks like you changed them in sql-local-remote.bats, but not in status.bats. I'm not going to block on this. Do as you see fit.

macneale4:
Related to: https://github.com/dolthub/dolt/issues/3922

PR #6015, opened 2023-05-23
Migrate `dolt add` to use Sql backend.

nicktobey:
This PR migrates `dolt add` to invoke SQL commands instead of manipulating the database directly. This allows it to work even on remote connections.


go/cmd/dolt/commands/add.go
```diff
@@ -59,40 +61,74 @@ func (cmd AddCmd) ArgParser() *argparser.ArgParser {
 	return cli.CreateAddArgParser()
 }
 
+// Generate the query that will call the `DOLT_ADD` stored proceudre.
+// This is safe because the only inputs are flag names and validated table names.
```
macneale4, line 65: Comments on functions should conform to godocs.

// generateSql returns ... This function assumes validated/safe input, and does not error.

go/cmd/dolt/commands/add.go
```diff
@@ -59,40 +61,74 @@ func (cmd AddCmd) ArgParser() *argparser.ArgParser {
...
+	}
+	buffer.WriteString(")")
+	return buffer.String()
+}
+
 // Exec executes the command
 func (cmd AddCmd) Exec(ctx context.Context, commandStr string, args []string, dEnv *env.DoltEnv, cliCtx cli.CliContext) int {
+	queryist, sqlCtx, closeFunc, err := cliCtx.QueryEngine(ctx)
+	if err != nil {
+
```
macneale4, line 99: Ignored Error

go/cmd/dolt/commands/add.go
```diff
@@ -59,40 +61,74 @@ func (cmd AddCmd) ArgParser() *argparser.ArgParser {
...
-		}
-	} else {
-		roots, err = actions.StageTables(ctx, roots, apr.Args, !apr.Contains(cli.ForceFlag))
-		if err != nil {
-			return HandleStageError(err)
-		}
+	schema, rowIter, err := queryist.Query(sqlCtx, generateSql(apr))
+	if err != nil {
+		cli.PrintErrln(errhand.VerboseErrorFromError(err))
+		return 1
```
macneale4, line 124: Looks like you have a duplicate call for no reason. You are checking the error and then doing the same thing again, no?

macneale4 (changes requested):

go/cmd/dolt/commands/add.go
```diff
@@ -119,15 +118,17 @@ func (cmd AddCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
-			cli.PrintErrln(errhand.VerboseErrorFromError(err))
-			return 1
-		}
+	schema, rowIter, err := queryist.Query(sqlCtx, generateSql(apr))
+	if err != nil {
+		cli.PrintErrln(errhand.VerboseErrorFromError(err))
+		return 1
+	}
+
+	_, err = sql.RowIterToRows(sqlCtx, schema, rowIter)
```
macneale4: Is there a better abstraction to ensure we always finish the transaction? Seems like it should be in the deferred close. Ok to ship what works, but I think you'll get tired of this behavior

go/cmd/dolt/commands/add.go
macneale4: Not clear why it's showing those line.... I'm talking about the two calls to `queryist.Query(sqlCtx, generateSql(apr))`

go/cmd/dolt/commands/add.go
```diff
@@ -59,40 +61,74 @@ func (cmd AddCmd) ArgParser() *argparser.ArgParser {
 	return cli.CreateAddArgParser()
 }
 
+// Generate the query that will call the `DOLT_ADD` stored proceudre.
```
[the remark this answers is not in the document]
nicktobey: Done

go/cmd/dolt/commands/add.go
nicktobey: Done

go/cmd/dolt/commands/add.go
nicktobey: Whoops, good catch.

go/cmd/dolt/commands/add.go
nicktobey: I'm not sure what happened here. I must have done something wrong during the rebase.

Fixed.

go/cmd/dolt/commands/add.go
nicktobey: I think I will too. I don't think it's unreasonable that row iterators can have side effects and you have to iterate over all of them to complete the transaction, but it's going to be a pain to have to remember this.

In another language like rust we might be able to make forgetting to close the row iterator a compiler error. But I can't think of anything in go that would make our lives easier here short of making Query return a closefunc (which would be a big refactor but we could do.)

macneale4: I bet there is a way to do this, I just don't know what it is. Maybe a dolt/go pro has a suggestion?

go/cmd/dolt/commands/cherry-pick.go
```diff
@@ -97,12 +97,12 @@ func (cmd CherryPickCmd) Exec(ctx context.Context, commandStr string, args []str
 		return HandleVErrAndExitCode(verr, usage)
 	}
 
-	verr := cherryPick(ctx, dEnv, cherryStr)
+	verr := cherryPick(ctx, dEnv, cliCtx, cherryStr)
```
macneale4, line 100: How is cherryPick needed for this change?

macneale4 (changes requested):
Command code change is simple and readable. More testing is required though. Is there an existing test somewhere which validates `dolt add AAA` doesn't add `BBB`? What about error messages? If so, where are those tests? I left almost all of my comments on your new test.

go/cmd/dolt/commands/add.go
nicktobey: So I don't think we want to simply defer the close because closing the rowIter commits the transaction, which we don't want to do if something goes wrong while iterating.

I think the way to do it with a callback, like this:

```
QueryWithCallback(sqlCtx SqlContext, q Queryist, query string, callback func(r Row) error) error {
    schema, rowIter, err := q.Query(sqlCtx, query)
    for _, row := range rowIter {
        err := callback(row)
        if err != nil {
            return err
        }
    }
    rowIter.close()
    return nil
}
```

The trouble is that if our motivation is to make sure that commands don't forget to close the RowIter, then we need to make it impossible for commands to call Queryist.Query directly. We can do this if we make the method private and make all the implementations wrapper classes in the same module as the interface, but that's a bit beyond the scope of this PR.

Thoughts?

go/cmd/dolt/commands/cherry-pick.go
nicktobey: CherryPick calls Add internally. If we migrate Add to use `cliCtx` but we don't migrate CherryPick to pass in `cliCtx`, then CherryPick will break.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +116,19 @@ teardown() {
...
+    mkdir -p someplace_new/fun
+    cd someplace_new/fun
+
+    dolt --verbose-engine-setup --data-dir="$ROOT_DIR/altDB" --user dolt sql -q "create table testtable (pk int PRIMARY KEY)"
+    dolt --verbose-engine-setup --data-dir="$ROOT_DIR/altDB" --user dolt add .
+
+    stop_sql_server 1
+
+    staged=$(get_staged_tables)
+
```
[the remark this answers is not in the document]
nicktobey: Shouldn't the local context already be covered by `add.bats`? Only the remote context needs to be tested here.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +116,19 @@ teardown() {
...
 }
 
+@test "sql-local-remote: verify simple dolt add behavior." {
+    start_sql_server altDb
+    ROOT_DIR=$(pwd)
+
+    mkdir -p someplace_new/fun
+    cd someplace_new/fun
+
+    dolt --verbose-engine-setup --data-dir="$ROOT_DIR/altDB" --user dolt sql -q "create table testtable (pk int PRIMARY KEY)"
```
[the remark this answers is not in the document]
nicktobey: Done

integration-tests/bats/sql-local-remote.bats
```diff
@@ -21,6 +21,14 @@ teardown() {
     teardown_common
 }
 
+get_staged_tables() {
+    dolt status | awk '
+        match($0, /new table:\ */) { print substr($0, RSTART+RLENGTH) }
+        /Untracked tables:/ { exit }
+        /Tables with conflicting dolt_ignore patterns:/ { exit }
+    '
```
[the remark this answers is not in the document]
nicktobey: This was designed to mirror the output of git status, which doesn't support structural output. I agree it would be nice. And trivial to implement with `git status` migrated. I created an issue for it.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +116,19 @@ teardown() {
...
 
+@test "sql-local-remote: verify simple dolt add behavior." {
+    start_sql_server altDb
+    ROOT_DIR=$(pwd)
+
+    mkdir -p someplace_new/fun
+    cd someplace_new/fun
+
+    dolt --verbose-engine-setup --data-dir="$ROOT_DIR/altDB" --user dolt sql -q "create table testtable (pk int PRIMARY KEY)"
+    dolt --verbose-engine-setup --data-dir="$ROOT_DIR/altDB" --user dolt add .
```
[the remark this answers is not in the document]
nicktobey: Done

integration-tests/bats/sql-local-remote.bats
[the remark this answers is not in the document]
nicktobey: Turns out we didn't have a bats test for just `dolt add`. We do now. It's a pretty simple command.

This new bats file will automatically benefit from https://github.com/dolthub/dolt/pull/6060 too.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +116,19 @@ teardown() {
...
+
+    stop_sql_server 1
+
+    staged=$(get_staged_tables)
+
+    [[ ! -z $(echo "$staged" | grep "testtable") ]] || false
```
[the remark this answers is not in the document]
nicktobey: I added `dolt add` tests.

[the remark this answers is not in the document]
macneale4: Given that, I think you should just drop the tests from sql-local-remote. These don't add more coverage

integration-tests/bats/add.bats
```diff
@@ -0,0 +1,70 @@
...
+    dolt add addedTable
+
+    staged=$(get_staged_tables)
+    working=$(get_working_tables)
+
+    [[ ! -z $(echo "$staged" | grep "addedTable") ]] || false
+    [[ -z $(echo "$working" | grep "addedTable") ]] || false
+    [[ ! -z $(echo "$working" | grep "notAddedTable") ]] || false
+    [[ -z $(echo "$staged" | grep "notAddedTable") ]] || false
+}
```
macneale4, line 70: Are there existing tests to validate ignore behavior works correctly?

macneale4 (approved):
This seems good to ship. I have a couple questions about test coverage and clean up, but you can use your judgment and ship when ready

integration-tests/bats/sql-local-remote.bats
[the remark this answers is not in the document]
nicktobey: As discussed privately, the plan going forward is to have a single test in `sql-local-remote.bats` for each command that runs in both local and remote mode and inspects the `--verbose-engine-setup` output like we see here. This gives us increased confidence that even in the absence of the new CI tests, the command is able to connect to a remote server. The new CI tests then prove that the command's behavior is correct.

go/cmd/dolt/commands/add.go
macneale4: I think a callback is probably necessary here. Given what you are seeing with the error handling, we may need to pin down the full workflow and all of the call backs used to execute a command end to end. That's on me, but I think we can outline as a group pretty quickly.

nicktobey: Yeah, I think this is needed but it's outside the scope of this PR.

go/cmd/dolt/dolt.go
```diff
@@ -581,6 +580,17 @@ func buildLateBinder(ctx context.Context, cwdFS filesys.Filesys, mrEnv *env.Mult
 		targetEnv = mrEnv.GetEnv(useDb)
 	}
 
+	// There is no target environment detected. This is allowed for a small number of command.
```
PavelSafronov, line 583: Nit: `commands.`

PR #6018, opened 2023-05-24
Optimize COMMIT performance for large databases.

reltuk:
A bottleneck in sanity checks that we run when we add new chunks to the database could make commit performance slow down as the database increased in size. On some benchmarks we ran, a typical slowdown might be ~100ms per 100GBs of database size. This optimizes the code paths involved to use more efficient algorithms and to use a scan resistant LRU cache.

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -34,12 +34,12 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/dustin/go-humanize"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
-	lru "github.com/hashicorp/golang-lru/v2"
 
 	"github.com/dolthub/dolt/go/store/blobstore"
 	"github.com/dolthub/dolt/go/store/chunks"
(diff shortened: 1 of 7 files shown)


go/store/nbs/store.go
```diff
@@ -1655,6 +1688,8 @@ func (nbs *NomsBlockStore) swapTables(ctx context.Context, specs []tableSpec) (e
 		return errors.New("concurrent manifest edit during GC, before swapTables. GC failed.")
 	}
 
+	nbs.hasCache.Purge()
```
zachmu, line 1691: Not clear from casual read why this is necessary here and only here

zachmu (approved):
LGTM

go/store/nbs/table_set.go
```diff
@@ -286,7 +287,13 @@ func (ts tableSet) Size() int {
 
 // append adds a memTable to an existing tableSet, compacting |mt| and
 // returning a new tableSet with newly compacted table added.
-func (ts tableSet) append(ctx context.Context, mt *memTable, checker refCheck, stats *Stats) (tableSet, error) {
+func (ts tableSet) append(ctx context.Context, mt *memTable, checker refCheck, hasCache *lru.TwoQueueCache[addr, struct{}], stats *Stats) (tableSet, error) {
```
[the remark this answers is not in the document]
reltuk: tableSet is currently immutable (I think?) and we use the hasCache outside of it in the errorIfPending...I don't think I want the same hasCache hanging on multiple table sets because it would be nice if we could eventually remove some unnecessary locking overhead if it turns out to be a win...

go/store/nbs/store.go
reltuk: Good point. Will add a comment.

PR #6025, opened 2023-05-24
Use the SQL user as the Dolt commit author when operating in a SQL context

fulghum:
When operating in SQL mode (e.g. `dolt sql` or connected to a `dolt sql-server`), the Dolt commit author is now recorded as the SQL user who created the Dolt commit, instead of using the user info configured through `dolt config`. Committing from the CLI continues to use the `dolt config` configured user info. 

Fixes https://github.com/dolthub/dolt/issues/5607

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_commit.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_commit.go
@@ -87,8 +87,26 @@ func doDoltCommit(ctx *sql.Context, args []string) (string, error) {
 			return "", err
 		}
 	} else {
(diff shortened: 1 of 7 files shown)


zachmu (approved):
LGTM

timsehn:
As discussed, it makes sense to just make this the default with no config option to change.

timsehn (changes requested):
We should make this the default with no option to change.

fulghum:
> We should make this the default with no option to change.

I agree. I got the tests cleaned up and merged those changes back into this PR. Once they pass, this should be good to go.

PR #6026, opened 2023-05-24
Remove __LD_1__ from CI

nicktobey:
This is an initial pass at removing integration tests that test the old data format or the migration to the new format.

The most important thing is the changes to the GitHub action yaml files. We no longer create actions where the default format is __LD_1__.

This PR also removes some individual tests that test the old format, but not all of them, only the really low hanging fruit to start. Removing all of these tests will likely be a gradual approach as we edit bats tests.


integration-tests/bats/json-new-fmt.bats
```diff
@@ -6,8 +6,7 @@ load $BATS_TEST_DIRNAME/helper/common.bash
 
 setup() {
     setup_common
-    skip_nbf_not_dolt
-}
+doltdb.IgnoreTableName}
```
macneale4, line 9: Why delete the newline?

integration-tests/bats/sql-merge.bats
```diff
@@ -974,8 +973,7 @@ SQL
     [[ "$output" =~ "CONSTRAINT \`c\` CHECK ((\`i\` < 10))" ]] || false
     dolt commit -am "changes to main"
 
-    skip_nbf_not_dolt
-    dolt merge other -m "merge other"
+doltdb.IgnoreTableName    dolt merge other -m "merge other"
```
macneale4, line 976: Are you mixing newline char types or something? Seems like all of your newlines are rendering poorly in the GitHub diff view.

integration-tests/bats/json-new-fmt.bats
macneale4: newlines?

macneale4 (changes requested):

timsehn:
Partially addresses: https://github.com/dolthub/dolt/issues/6007

macneale4 (approved):
Latest change LGTM. Ship when tests complete

PR #6027, opened 2023-05-24
Change the Dolt commit author in SQL mode to be the current SQL user

fulghum:
Trying this is a separate PR to catch test failures from this change

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_commit.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_commit.go
@@ -87,8 +87,26 @@ func doDoltCommit(ctx *sql.Context, args []string) (string, error) {
 			return "", err
 		}
 	} else {
-		name = dSess.Username()
-		email = dSess.Email()
+		value, err := ctx.GetSessionVariable(ctx, dsess.DoltSqlUserIsCommitter)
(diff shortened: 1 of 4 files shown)


timsehn (approved):
LGTM.

fulghum:
Making this behavior the default is easy, but it's going to break a bunch of tests. I'll dig through the test failures and make sure it's safe to just update those tests. The code change is already isolated to the SQL path, but I want to double check that we aren't somehow causing weird results for any of the CLI use cases.

timsehn:
> Making this behavior the default is easy, but it's going to break a bunch of tests. I'll dig through the test failures and make sure it's safe to just update those tests. The code change is already isolated to the SQL path, but I want to double check that we aren't somehow causing weird results for any of the CLI use cases.

I'm surprised we have a lot of tests that check the committer on the SQL path.

fulghum:
> I'm surprised we have a lot of tests that check the committer on the SQL path.

You're right... it wasn't as bad as I thought it was going to be at first. Hopefully this next test run is clean and then we can get this merged in.

fulghum:
Closing this test PR now that CI checks are passing cleanly; see https://github.com/dolthub/dolt/pull/6025 for the final change

PR #6033, opened 2023-05-25
Changed replication configuration errors to log warnings, rather than stop the server from starting

zachmu:
Fixes https://github.com/dolthub/dolt/issues/6019


timsehn (approved):
Much better interface I think.

PR #6039, opened 2023-05-26
Storage optimization: Optimize GetMany operations when table files are very large.

reltuk:
Previously we would perform a linear scan of the table file index, matching up queried chunks with their location in the table file. This changes it to use a binary search, which is the right tradeoff for large table files and small(-ish) query sizes.


zachmu (approved):
LGTM!

PR #6040, opened 2023-05-26
Only bats on Dolt defaults during CI (`__DOLT__` format with chunk journal on)

timsehn:


timsehn:
We may be able to just remove the option but I'm not smart enough to figure out how to do that.

coffeegoddd (approved):
LGTM! (assuming chunk journal works on macos)

timsehn:
Turns out, I may be smart enough.

PR #6041, opened 2023-05-26
Fix for bug when evaluating check constraints during merge 

fulghum:
Fixes check constraint evaluation during merge when a table has a primary key that does not appear as the first columns in the table schema.


max-hoffman (approved):
LGTM!

PR #6050, opened 2023-05-26
Synchronize the two queryist implementations

PavelSafronov:
Synchronize the two queryist implementations so they produce the same results.

- keep the current MySQL iterator that returns everything as a string
- add SqlEngineQueryist, which wraps calls to SqlEngine and returns all data as a string
- SqlEngineQueryist is only going to be used by dolt subcommands, so no one is being broken by this change

## More info about this change

The current Queryist logic flips between 2 implementations:
1. SQLEngine querying - here, we access Dolt directly, and get back properly typed responses. We're also receiving `bool` objects in some cases, like with [the `staged` column  on `dolt_status` table](https://github.com/dolthub/dolt/blob/main/go/libraries/doltcore/sqle/dtables/status_table.go#L49). This mode is used when we are running `dolt foo` from a dolt directory.
2. MySQL querying - we are currently using MysqlRowWrapper, an existing wrapper that considers every column a string and returns strings. This mode is used when we are running `dolt foo` commands against a remote server.

So with the current Queryist logic, if we execute the query `select * from dolt_status`, then the 2nd column will either be `bool` (with SQLEngine) or `string` (with MySQL).
That means that currently Queryist consumers would have to consider and handle both cases.

This PR synchronizes the two flavors of Queryist implementations by always returning strings from dolt. This is a smaller change than the previous version of this PR and does not impact any dolt consumers.

### DIFF
--- a/go/cmd/dolt/commands/sql_engine_queryist.go
+++ b/go/cmd/dolt/commands/sql_engine_queryist.go
@@ -85,7 +85,7 @@ func (s SqlEngineRowIter) valToString(ctx *sql.Context, index int, val interface
 	case uint64:
 		switch ct.(type) {
 		case types.BitType_:
-			newValue = fmt.Sprintf("%v", string(t))
+			newValue = string(rune(t))
 		default:
 			newValue = fmt.Sprintf("%d", t)
 		}
--- a/go/cmd/dolt/commands/sql_engine_queryist.go
+++ b/go/cmd/dolt/commands/sql_engine_queryist.go
@@ -1,0 +1,133 @@
+package commands
+
+import (
+	"fmt"
+	"github.com/dolthub/dolt/go/cmd/dolt/cli"
+	"github.com/dolthub/dolt/go/cmd/dolt/commands/engine"
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/types"
+	"github.com/shopspring/decimal"
+	"time"
+)
+
+type SqlEngineQueryist struct {
+	se *engine.SqlEngine
+}
+
+func (s SqlEngineQueryist) Query(ctx *sql.Context, query string) (sql.Schema, sql.RowIter, error) {
+	schema, iter, err := s.se.Query(ctx, query)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newIter := NewSqlEngineRowIter(iter, schema)
+	return newIter.newSchema, newIter, nil
+}
+
+var _ cli.Queryist = SqlEngineQueryist{}
+
+func NewSqlEngineQueryist(se *engine.SqlEngine) cli.Queryist {
+	return SqlEngineQueryist{se}
+}
+
+type SqlEngineRowIter struct {
+	iter      sql.RowIter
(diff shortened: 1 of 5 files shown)


timsehn:
This is cool work coming out of merging these two paths

go/cmd/dolt/commands/mysql_rows_iter.go
```diff
@@ -0,0 +1,278 @@
...
+	"strings"
+)
+
+// MySqlRowsIter is an iterator over rows returned by MySQL
+type MySqlRowsIter struct {
+	rows        *dsql.Rows
+	types       []sql.Type
+	columnTypes []*dsql.ColumnType
+	schema      sql.Schema
+}
```
macneale4, line 19: Where is 'types' ever use once the object is constructed? It looks like you never use it, so I wouldn't bother putting it in the struct.

go/cmd/dolt/commands/mysql_rows_iter.go
```diff
@@ -0,0 +1,278 @@
...
+func (b MySqlRowsIter) Close(_ *sql.Context) error {
+	return b.rows.Close()
+}
+
+// Schema returns the schema for the rows returned by this iterator.
+func (b MySqlRowsIter) Schema() sql.Schema {
+	return b.schema
+}
+
+func scanResultRow(results *dsql.Rows, cols []*dsql.ColumnType) (sql.Row, error) {
```
macneale4, line 74: You're going to need a lot of tests for this to convince me it's right. This feels like a lot of code what we could get wrong.

go/cmd/dolt/commands/mysql_rows_iter.go
```diff
@@ -0,0 +1,278 @@
...
+		scanRow[i] = t
+	}
+}
+
+var typeDefaults = map[string]string{
+	"char":      "char(255)",
+	"binary":    "binary(255)",
+	"varchar":   "varchar(65535)",
+	"varbinary": "varbinary(65535)",
+}
```
macneale4, line 216: Why so few? What's the thinking on this specific set.

macneale4 (changes requested):
Thanks for going deep on this! This is pretty high risk. Need lots of tests, or I need to be convinced that we have existing tests which cover all of the crazy ways schema's can be different. Given all the logic in these translation laters, we could get this very wrong.

go/cmd/dolt/commands/sql_engine_queryist.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+var _ sql.RowIter = (*SqlEngineRowIter)(nil)
+
+func NewSqlEngineRowIter(iter sql.RowIter, schema sql.Schema) *SqlEngineRowIter {
+	newSchema := make(sql.Schema, len(schema))
+	for i, col := range schema {
+		newSchema[i] = &sql.Column{
+			Name:     col.Name,
+			Type:     types.LongText,
```
macneale4, line 94: I don't understand this. Seems like reduced behavior over what we already have.

macneale4 (changes requested):

nicktobey (approved):

PR #6060, opened 2023-05-30
Test CI with automatic server connection for CLI commands.

nicktobey:
This adds additional CI workflows that start a Dolt SQL server before running the bats tests. The dolt commands in the tests should connect to the running server instead of running their own engine.

The file `local-remote.bats` contains a list of bats files that are not yet confirmed to work against a remote server because they contain dolt commands that haven't been migrated yet. This is a burndown list that will shrink as we migrate more commands.


integration-tests/bats/helper/local-remote.bash
```diff
@@ -0,0 +1,163 @@
+load helper/query-server-common
+
+SKIP_SERVER_TESTS=$(cat <<-EOM
```
macneale4, line 3: Are we going to treat this as a burndown? I'm assuming this is all files with the exception of a few. Can you add a comment which describes what's in this test please.

.github/workflows/ci-bats-unix.yaml
macneale4, line 1: Does Windows get any love from this change?

integration-tests/bats/helper/query-server-common.bash
```diff
@@ -21,12 +21,12 @@ import os
...
 args = sys.argv[sys.argv.index('--') + 1:]
-working_dir, database, port_str, timeout_ms = args
+working_dir, database, port_str, timeout_ms, user = args
 os.chdir(working_dir)
 
 from pytest import wait_for_connection
-wait_for_connection(port=int(port_str), timeout_ms=int(timeout_ms), database=database, user='dolt')
-" -- "$PYTEST_DIR" "$DEFAULT_DB" "$1" "$2"
+wait_for_connection(port=int(port_str), timeout_ms=int(timeout_ms), database=database, user=user)
+" -- "$PYTEST_DIR" "$DEFAULT_DB" "$1" "$2" "${SQL_USER:-dolt}"
```
macneale4, line 29: Not clear we need the additional user argument. Is this because there are many tests which don't use the `dolt` user?

macneale4 (changes requested):
A couple questions, but this will be really cool to have.

Please update the title before shipping :)

integration-tests/bats/helper/query-server-common.bash
nicktobey: It's because currently commands that don't supply a user but end up using the remote queryist assume a default user of "root", whereas the `start_sql_server` helper function previously assumed a user of "dolt". `local-remote.bash` will set this env variable to `root` when running against a server process.

.github/workflows/ci-bats-unix.yaml
nicktobey: If I gave Windows some love, would we be able to test it? I don't see the windows tests in our CI results.

integration-tests/bats/helper/local-remote.bash
nicktobey: Yes. This is a burndown. Once every dolt command in a bats file is migrated, we can remove that file from this list and all the tests in that file will automatically be run in both modes.

In practice, most of these files test a single command and then use a couple of other commands (sql, status, diff) to either setup the test environment or examine the results. So once these "helper" commands are migrated it should be one-for-one with the subsequent commands.

macneale4 (approved):
Update the description and title, then ship it!

PR #6065, opened 2023-05-31
update dolt blame to use sql backend and accept specific revision

stephkyou:
This change updates `dolt blame` to use the appropriate sql engine to generate results. This change also allows users to specify a specific revision to annotate from.

### DIFF
--- a/go/cmd/dolt/commands/blame.go
+++ b/go/cmd/dolt/commands/blame.go
@@ -23,10 +23,12 @@ import (
 	eventsapi "github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi/v1alpha1"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+	"github.com/dolthub/go-mysql-server/sql"
 )
 
 const (
-	blameQueryTemplate = "SELECT * FROM dolt_blame_%s"
+	blameQueryTemplate     = "SELECT * FROM dolt_blame_%s"
+	blameAsOfQueryTemplate = "SELECT * FROM dolt_blame_%s AS OF %s"
 )
 
 var blameDocs = cli.CommandDocumentationContent{
@@ -55,7 +57,7 @@ func (cmd BlameCmd) Docs() *cli.CommandDocumentation {
 }
 
 func (cmd BlameCmd) ArgParser() *argparser.ArgParser {
-	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 1)
+	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 2)
 	return ap
 }
 
@@ -85,7 +87,7 @@ func (cmd BlameCmd) Exec(ctx context.Context, commandStr string, args []string, 
 	help, usage := cli.HelpAndUsagePrinters(cli.CommandDocsForCommandString(commandStr, blameDocs, ap))
 	apr := cli.ParseArgsOrDie(ap, args, help)
 
-	if apr.NArg() != 1 {
+	if apr.NArg() > 2 || apr.NArg() == 0 {
 		usage()
 		return 1
 	}
@@ -98,10 +100,17 @@ func (cmd BlameCmd) Exec(ctx context.Context, commandStr string, args []string, 
 		defer closeFunc()
 	}
 
-	schema, ri, err := queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(0)))
+	var schema sql.Schema
+	var ri sql.RowIter
+	if apr.NArg() == 1 {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(0)))
+	} else {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameAsOfQueryTemplate, apr.Arg(1), apr.Arg(0)))
+	}
 	if err != nil {
 		return 1
 	}
+
 	err = engine.PrettyPrintResults(sqlCtx, engine.FormatTabular, schema, ri)
 	if err != nil {
 		return 1
--- a/go/cmd/dolt/commands/blame.go
+++ b/go/cmd/dolt/commands/blame.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/blame.go
```diff
@@ -84,11 +87,34 @@ func (cmd BlameCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+	if closeFunc != nil {
+		defer closeFunc()
+	}
+
+	var schema sql.Schema
+	var ri sql.RowIter
+	if apr.NArg() == 1 {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(0)))
+	} else {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameAsOfQueryTemplate, apr.Arg(1), apr.Arg(0)))
```
macneale4, line 108: I'd do one query template, and specify HEAD when no option is given. It's easier to follow that way.

integration-tests/bats/blame.bats
```diff
@@ -50,55 +50,53 @@ SQL
...
 
 @test "blame: works with HEAD as the commit ref" {
-    skip "SQL views do no support AS OF queries"
-
     run dolt blame HEAD blame_test
     [ "$status" -eq 0 ]
-    [[ "$output" =~ "Thomas Foolery" ]] || false
-    [[ "$output" =~ "create blame_test table" ]] || false
+
+    [[ "$output" =~ "1".*"Thomas Foolery".*"create blame_test table" ]] || false
```
macneale4, line 88: Not clear what this does and why. Can you educate me?

stephkyou: `*` is a wildcard that matches any number of any character. My intention was to check that the correct phrases are on the correct lines associated with their `pk`. I see a later test (L147) that uses `lines` to do a similar thing in a cleaner way so I may switch over to that.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +108,29 @@ teardown() {
     [[ "$output" =~ "defaultDB does not exist" ]] || false
 }
 
+@test "sql-local-remote: verify dolt blame behavior is identical in switch between server/no server" {
+    cd defaultDB
```
macneale4, line 112: There is a bug here where we are out of sync with other command behavior. You shouldn't need to `cd defaultDB` - and in fact I tested this and you get an error stating that you need to dolt init first. This is incorrect behavior.

integration-tests/bats/blame.bats
macneale4: Got it. I had to take a look at blame's output to understand what you were doing.

The unfortunate thing here is that there are a lot of columns of output we are just ignoring. So long as other tests cover those I think this makes sense. Can you verify?

macneale4: Got it. I had to take a look at blame's output to understand what you were doing.

The unfortunate thing here is that there are a lot of columns of output we are just ignoring. So long as other tests cover those I think this makes sense. Can you verify?

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +108,29 @@ teardown() {
...
+    [[ "$output" =~  "1".*"insert initial value into test" ]] || false
+    [[ "$output" =~  "2".*"insert more values into test" ]] || false
+    [[ "$output" =~  "3".*"insert more values into test" ]] || false
+    stop_sql_server 1
+
+    run dolt blame test
+    [ "$status" -eq 0 ]
+    [[ "$output" =~  "1".*"insert initial value into test" ]] || false
+    [[ "$output" =~  "2".*"insert more values into test" ]] || false
+    [[ "$output" =~  "3".*"insert more values into test" ]] || false
```
macneale4, line 134: I think a simpler check would be to compare the total output of both blame calls. The line by line information will be the same if that checks out, and the tests as it stands now would allow for one implementation to drop the commit date and we'd never know it.

go/cmd/dolt/commands/blame.go
```diff
@@ -84,11 +87,34 @@ func (cmd BlameCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+	if closeFunc != nil {
+		defer closeFunc()
+	}
+
+	var schema sql.Schema
+	var ri sql.RowIter
+	if apr.NArg() == 1 {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(0), "HEAD"))
+	} else {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(1), apr.Arg(0)))
```
macneale4, line 108: We need to validate the input or we'll have a SQL injection bug here. There is hopefully code we can find around here which validates a revision value. Example of exploit:

`$ dolt blame 'HEAD;insert into tbl values (....);' defaultDB_tbl`

At the moment, this doesn't have any real impact because blame doesn't have a write transaction

macneale4 (changes requested):

integration-tests/bats/blame.bats
stephkyou: I can add the email column in these checks as well,, the commit date and commit hash might be harder to check since those will change every time the test is run I think. Do you think that suffices to check for accuracy?

integration-tests/bats/sql-local-remote.bats
stephkyou: I'm not sure I understand why I don't need to do `cd defaultDB` first. I would need to be in a dolt repo to do `dolt add` and `dolt commit` right?

integration-tests/bats/blame.bats
stephkyou: Just read your comment below. I'll add a check for the headers if the table and see if I can test that the commit date and commit hash columns aren't empty at least.

go/cmd/dolt/commands/blame.go
```diff
@@ -84,11 +94,44 @@ func (cmd BlameCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+		defer closeFunc()
+	}
+
+	var schema sql.Schema
+	var ri sql.RowIter
+	if apr.NArg() == 1 {
+		schema, ri, err = queryist.Query(sqlCtx, fmt.Sprintf(blameQueryTemplate, apr.Arg(0), "HEAD"))
+	} else {
+		// validate input
+		_, err = ResolveCommitWithVErr(dEnv, apr.Arg(0))
```
macneale4, line 117: See chat discussion. This is a very reliable way to validate the input - but it used dEnv which is off limits.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -108,3 +108,25 @@ teardown() {
...
+
+    start_sql_server altDB
+    run dolt --user dolt blame test
+    [ "$status" -eq 0 ]
+    export out="$output"
+    stop_sql_server 1
+
+    run dolt blame test
+    [ "$status" -eq 0 ]
+    [[ "$output" =  $out ]] || false
```
macneale4, line 130: Noice. I like this test a lot more now :-)

macneale4 (approved):
Ship with whatever you come up to get around the use of dEnv.

PR #6068, opened 2023-05-31
Add support for parquet as an output format for `dolt sql`

fulghum:
Adds support for outputting the results of queries as parquet data. For example, `dolt sql -r parquet -q "select * from mytable;" > out.parquet` will select all rows from `mytable`, output them to stdout as parquet data, and redirect that output to the `out.parquet` file.

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -147,7 +147,7 @@ func (cmd SqlCmd) Docs() *cli.CommandDocumentation {
 func (cmd SqlCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 0)
 	ap.SupportsString(QueryFlag, "q", "SQL query to run", "Runs a single query and exits.")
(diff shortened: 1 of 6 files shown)


timsehn (approved):
Bats tests look good!

go/cmd/dolt/commands/sql.go
```diff
@@ -584,6 +584,8 @@ func GetResultFormat(format string) (engine.PrintResultFormat, errhand.VerboseEr
 		return engine.FormatNull, nil
 	case "vertical":
 		return engine.FormatVertical, nil
+	case "pq":
```
jennifersp, line 587: I think we should use the format naming as `parquet` instead of `pq`, since some of formats we support here such as `csv` and `json` are also used as file extensions.

jennifersp (approved):
Looks great! One small suggestion. 🙂

go/cmd/dolt/commands/sql.go
fulghum: Good feedback! I thought I had seen another spot where we had used `pq`, but now that I look again, I see `parquet` used for `dolt dump` and `dolt table export` uses the file extension to detect the output format. Thanks for catching this; I'll push a new commit with the format named changed to `parquet`.

PR #6077, opened 2023-06-01
Evaluate column default expressions during merge

fulghum:
Previously, we only supported evaluating column default expressions during merge when the expression only contained literals. This PR expands that support to evaluate column default expressions that contain functions and column references.


zachmu (approved):
Very cool!

PR #6083, opened 2023-06-02
Print useful error message when running CliContext-using-commands outisde of a repo

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6082

Currently, commands that use the new `CliContext` return unhelpful error messages when run outside a repo. This change standardizes these commands to print `The current directory is not a valid dolt repository.`, the same message currently returned by non-`CliContext` commands.

This is a change to `dolt sql`'s error messages, and additionally changes `dolt sql -q "show databases"` from a success to a failure. I'm not sure if we're okay with this, looking for feedback.

### DIFF
--- a/go/cmd/dolt/dolt.go
+++ b/go/cmd/dolt/dolt.go
@@ -19,6 +19,7 @@ import (
 	crand "crypto/rand"
 	"encoding/binary"
 	"fmt"
+	"github.com/dolthub/go-mysql-server/sql"
 	"math/rand"
 	"net/http"
 	_ "net/http/pprof"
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -581,6 +582,14 @@ func buildLateBinder(ctx context.Context, cwdFS filesys.Filesys, mrEnv *env.Mult
...
 
+	// There is no target environment detected. Return a cli.LateBindQueryist that returns an error.
+	// This stops commands that need to run queries, while allowing other commands to run.
+	if targetEnv == nil {
+		return func(ctx context.Context) (cli.Queryist, *sql.Context, func(), error) {
+			return nil, nil, nil, fmt.Errorf("The current directory is not a valid dolt repository.")
+		}, nil
+	}
+
 	// nil targetEnv will happen if the user ran a command in an empty directory - which we support in some cases.
```
macneale4, line 593: This line seems to indicate that nil targets are sometimes expected. I'm pretty sure this is what happens when you are in a directory that `dolt sql-server` has started on, but no databases exist yet, so I don't think that the fix above is what you want in all cases.

macneale4 (commented):
We need to determine what the right behavior is for running a server on an empty database. You are suggesting that all commands (dolt sql included) need to error out. How do you create a database then? There is a chicken and egg issue here.

nicktobey:
This is covered by #6015

PR #6084, opened 2023-06-02
Pavel/dolt status

macneale4:


integration-tests/bats/sql-local-remote.bats
```diff
@@ -7,6 +7,19 @@ make_repo() {
   cd "$1"
   dolt init
   dolt sql -q "create table $1_tbl (id int)"
+  dolt sql <<SQL
+CREATE TABLE t (pk int PRIMARY KEY);
```
macneale4, line 11: can you use identifiers which are 3 chars or more? Easier for future maintainers to search/replace. I know we do single char vars all over the place - just trying to raise the bar one change at a time :-)

PR #6085, opened 2023-06-02
release: simplify homebrew bump job

chenrui333:
not sure why, but currently the build does not quite work, https://github.com/dolthub/dolt/actions/runs/5158143131/jobs/9291442093

this PR is just to simplify the job (using the defaults by removing the redundant specifications), also the workflow_dispatch job can be totally replaced with `brew bump --open-pr dolt` (this is what i am doing with [this PR](https://github.com/Homebrew/homebrew-core/pull/132677)). In other words, if needed, I think we can just apply `brew bump --open-pr` for the workflow_dispatch event (no need to specify the version, it would bump to the latest release automatically)

### DIFF
--- a/.github/workflows/cd-bump-homebrew.yaml
+++ b/.github/workflows/cd-bump-homebrew.yaml
@@ -1,11 +1,6 @@
 name: Bump Dolt on Homebrew
 
 on:
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'SemVer format release tag, i.e. 0.24.5'
-        required: true
   repository_dispatch:
     types: [ bump-homebrew ]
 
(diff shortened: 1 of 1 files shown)


timsehn:
Thanks for the contribution. I forget who is our Homebrew expert. I'll find him/her and have him/her review.

timsehn:
Looks like @coffeegoddd is all over the blame for this file. Assigning review to him.

.github/workflows/cd-bump-homebrew.yaml
```diff
@@ -1,11 +1,6 @@
 name: Bump Dolt on Homebrew
 
 on:
-  workflow_dispatch:
-    inputs:
-      version:
-        description: 'SemVer format release tag, i.e. 0.24.5'
-        required: true
```
coffeegoddd, line 8: we want to keep this, as sometimes we need to run the workflow manual after a release has already run.

coffeegoddd: you can dedupe the code by doing stuff like:
```
tag-name: ${{format('refs/tags/v{0}', github.event.client_payload.version || github.event.inputs.version)}}
```

where `github.event.client_payload.version` is from the `workflow_dispatch` event and `github.event.inputs.version` is from the `repository_dispatch` event.

coffeegoddd (changes requested):
Thanks for the submission, a couple comments to address before we merge this.

timsehn:
Closing this one because contributor did not follow up.

PR #6086, opened 2023-06-02
Add support for `--skip-empty` to `dolt commit` and `dolt_commit()`

fulghum:
The new `--skip-empty` flag can be passed to `dolt commit ...` or `call dolt_commit(...)` and have the commit operation be a no-op, instead of an error, if there are no changes staged to commit. It is an error to use `--skip-empty` together with `--allow-empty`.

Fixes: https://github.com/dolthub/dolt/issues/5678


JCOR11599 (approved):
LGTM

PR #6087, opened 2023-06-02
[no-release-notes] add more info to constraint violation error

JCOR11599:
Changed `ErrUnresolvedConstraintViolationsCommit` to print all the information we have on the violation for easier debugging.

### DIFF
--- a/go/libraries/doltcore/merge/violations_unique_prolly.go
+++ b/go/libraries/doltcore/merge/violations_unique_prolly.go
@@ -163,7 +163,7 @@ func (m NullViolationMeta) ToString(ctx *sql.Context) (string, error) {
 // CheckCVMeta holds metadata describing a check constraint violation.
 type CheckCVMeta struct {
 	Name       string `json:"Name"`
-	Expression string `jason:"Expression"`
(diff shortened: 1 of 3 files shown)


fulghum (approved):
I don't think it hurts to add more debugging info, but I'm not convinced just the table names is going to be enough to debug the merge error we've been seeing.

coffeegoddd:
@fulghum what information is needed to solve this?

fulghum:
> @fulghum what information is needed to solve this?

Good question. I haven't been debugging this issue today; I just looked over this PR. So, I think you guys probably have more context on the issue than I do so are probably in a better place to say what info is needed. It seems like if we don't know what constraint violation is triggering the error though, then we'd need as much info on the violation as we can get. That's just my take from looking at this PR though. Do you guys think you can debug once you know the table name?

coffeegoddd:
Ideally it would be nice to surface exactly what the violation is, and like the named keys and indexes, and i guess the required order of the keys/indexes? im not sure exactly what's different about the dolthub context and other contexts. this violation happens after insert statements, which is supposed be impossible

PR #6090, opened 2023-06-05
Actually run bats tests in a configuration where all commands connect to a server process.

nicktobey:
Due to a typo in the .yaml script, we weren't actually running bats tests with the `$SQL_ENGINE` environment variable set, which meant that we weren't actually testing the ability for commands to seamlessly connect to running server processes. This PR fixes this.

### DIFF
--- a/integration-tests/bats/sql-local-remote.bats
+++ b/integration-tests/bats/sql-local-remote.bats
@@ -132,23 +132,3 @@ teardown() {
     [ "$status" -eq 0 ]
     [[ "$output" =  $out ]] || false
 }
-
-@test "sql-local-remote: verify simple dolt add behavior." {
-    start_sql_server altDB
-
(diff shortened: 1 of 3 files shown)


integration-tests/bats/sql-local-remote.bats
```diff
@@ -11,6 +11,9 @@ make_repo() {
 }
 
 setup() {
+    if [ "$SQL_ENGINE" = "remote-engine" ]; then
+      skip "This test tests remote connections directly, SQL_ENGINE is not needed."
```
macneale4, line 15: Not clear why you are skipping it. Shouldn't you unset the env var and allow the test to continue?

macneale4 (changes requested):

macneale4 (approved):

PR #6104, opened 2023-06-07
[no-release-notes] Removing collation error handling tests

fulghum:
These test assertions will get in the way as we fill out our collation support, and our error reporting code is tested in other tests, so this block of code isn't really adding anything.


Hydrocharged (approved):
🚀

Issue #6100, opened 2023-06-06
Incorrect behavior when attempting to rename active branches with `dolt_branch` stored procedure.


nicktobey:
There are three possibly related issues, depending on whether the user is connecting to a server or a local db, and whether the active branch is "session active branch" within the SQL session or the "default active branch" set by the db. I'm grouping them here because decisions about the desired behavior will affect all three.

## Bug 1: Renaming the session active branch on a local db results in action to be only partially successful.

To reproduce:
```
dolt init
dolt branch other
echo "call dolt_checkout('other'); call dolt_branch('-m', 'other', 'newOther')" | dolt sql
dolt branch
```

expected output:
```
newOther
other
```

actual output:

`dolt sql` call fails with `Cannot set cwb head ref with a SessionStateAdapter`
```
main
newOther
other
```

The new branch name has been created but the old name still exists.

## Bug 2: Renaming the session active branch on a remote db results in misleading error message.

To reproduce:
```
mkdir repo
cd repo
dolt init
dolt branch other
dolt sql-server
echo "use repo;\\n call dolt_checkout('other');\\n call dolt_branch('-m', 'other', 'newOther')" | dolt sql-client -u root
dolt branch
```

Error:

```
Error 1105: unsafe to delete or rename branches in use in other sessions; use --force to force the change
```

The branch isn't in use in other sessions, but it *is* in use in this session. Calling `dolt_branch` with `--force` results in the same behavior as Bug 1.

I don't know if we actually want to forbid this rename or not, but if so we might want to improve the error message.

## Bug 3: Renaming the default active branch on a remote db succeeds, but does not update the head in `repo_state.json`

To reproduce:

```
mkdir repo
cd repo
dolt init
dolt branch other
dolt sql-server
echo "use repo;\\n call dolt_checkout('other');\\n call dolt_branch('-m', 'main', 'newMain')" | dolt sql-client -u root
dolt branch
```

Output:

```
error: failed to read local branches from db
cause: Error 1105: branch not found
```

Furthermore:

```
echo "use dbname;" | dolt sql-client -u root
```

results in 

```
Error 1105: branch not found
```

A user can escape this by running `dolt branch <other_branch>`, but this requires them to know the name of an existing branch, and they can't use `dolt branch` to find one.

This happens because `repo_state.json` was never updated and continues to point to the now nonexistent branch name.

I'm not sure if we want to allow or forbid renaming the default active branch while a server is running. But this shouldn't happen.

PR #6112, opened 2023-06-07
Allow `call dolt_branch` to rename the active session branch.

nicktobey:
Fixes the first two bugs outlined in #6100

Basically, `SessionStateAdapter.SetCWBHeadRef` was previously unimplemented, but has fairly straightforward behavior.

This also fixes logic where all branches are considered to be "conflicting changes with other sessions" because they're being compared to themselves.

### DIFF
--- a/integration-tests/bats/sql-server.bats
+++ b/integration-tests/bats/sql-server.bats
@@ -1934,7 +1934,7 @@ behavior:
     cd repo1
     dolt branch other
     start_sql_server
-    dolt sql-client -P $PORT -u dolt --use-db '' -q "call dolt_checkout('other'); call dolt_branch('-m', 'main', 'newMain')"
+    dolt sql-client -P $PORT -u dolt --use-db repo1 -q "call dolt_checkout('other'); call dolt_branch('-m', 'other', 'newOther')"
     run dolt branch
     [ $status -eq 0 ]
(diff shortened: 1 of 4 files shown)


integration-tests/bats/sql-server.bats
```diff
@@ -1937,3 +1937,15 @@ s.close()
     [ $status -eq 0 ]
     [[ "$output" =~ "0x1C4C4E338AD7442184509D5182816AC3" ]] || false
 }
+
+@test "sql-branch: CALL DOLT_BRANCH -m on session active branch (dolt sql-server)" {
+    cd repo1
+    dolt branch other
+    start_sql_server
+    dolt sql-client -P $PORT -u dolt --use-db '' -q "call dolt_checkout('other'); call dolt_branch('-m', 'main', 'newMain')"
+    run dolt branch
```
macneale4, line 1946: Does this test pass? you are checking for newOther below.

Also - what it the --use-db with an empty argument supposed to do? This seems odd to me

nicktobey: Whoops. Yeah, this isn't right.

Pushed the fixed version.

nicktobey:
This will be covered by https://github.com/dolthub/dolt/pull/6128

PR #6114, opened 2023-06-07
Better docs for sql-server command

zachmu:
Provides a full example of a YAML file with all currently supported fields and their defaults (several were missing).

Also removes the `databases` YAML config element.


reltuk (approved):
LGTM!

PR #6121, opened 2023-06-08
[no-release-notes] Remove Python from `wait_for_connection` function in BATS

fulghum:
We've been seeing Python errors in the code that detects if a sql-server has fully started up yet. This PR removes the use of Python from `wait_for_connection` and instead just polls with bash. There is still a little bit of Python left in BATS that could be cleaned up/simplified further.

Fixes: https://github.com/dolthub/dolt/issues/6109

### DIFF
--- a/integration-tests/bats/helper/query-server-common.bash
+++ b/integration-tests/bats/helper/query-server-common.bash
@@ -17,7 +17,7 @@ wait_for_connection() {
   # exit code, so we need to temporarily enable 'set +e', but be sure to turn 'set -e' back on before we exit.
   set +e
   while [ $SECONDS -lt $end_time ]; do
-    mysql -u $user -h localhost --port $port --protocol TCP --connect-timeout 1 -e "SELECT 1;"
+    dolt sql-client -u $user --host localhost --port $port --timeout 1 -q "SELECT 1;"
     if [ $? -eq 0 ]; then
       echo "Connected successfully!"
       set -e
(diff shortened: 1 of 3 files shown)


timsehn (approved):
LGTM.

integration-tests/bats/helper/query-server-common.bash
```diff
@@ -2,31 +2,33 @@ SERVER_REQS_INSTALLED="FALSE"
...
-wait_for_connection(port=int(port_str), timeout_ms=int(timeout_ms), database=database, user=user)
-" -- "$PYTEST_DIR" "$DEFAULT_DB" "$1" "$2" "${SQL_USER:-dolt}"
+  port=$1
+  timeout=$2
+  user=${SQL_USER:-dolt}
+  end_time=$((SECONDS+($timeout/1000)))
+
+  # BATS has 'set -e' enabled, which causes the script to fail immediately if any subcommand returns a non-zero
+  # exit code, so we need to temporarily enable 'set +e', but be sure to turn 'set -e' back on before we exit.
+  set +e
```
zachmu, line 18: Could you just use `run` instead?

zachmu (approved):
LGTM

integration-tests/bats/helper/query-server-common.bash
fulghum: That's a good idea. Seems a little safer than setting/unsetting modes. I'll try this out in a follow up PR.

PR #6128, opened 2023-06-08
Migrate `dolt branch` to new CLI framework.

nicktobey:
This PR migrates `dolt branch` to invoke SQL commands instead of manipulating the database directly. This allows it to work even on remote connections.

The only thing that hasn't been migrated yet is `dolt branch --datasets`. It's not a documented flag. We're currently using it for an internal test and can migrate it if we can figure out a way to rewrite the test.

### DIFF
--- a/go/cmd/dolt/commands/branch.go
+++ b/go/cmd/dolt/commands/branch.go
@@ -192,30 +192,6 @@ func getBranches(sqlCtx *sql.Context, queryEngine cli.Queryist, remote bool) ([]
 	}
 }
 
-func getActiveBranchName(sqlCtx *sql.Context, queryEngine cli.Queryist) (string, error) {
-	command := "SELECT active_branch()"
-	schema, rowIter, err := queryEngine.Query(sqlCtx, command)
-	if err != nil {
-		return "", err
-	}
-	rows, err := sql.RowIterToRows(sqlCtx, schema, rowIter)
-	if err != nil {
-		return "", err
-	}
-	if len(rows) != 1 {
-		return "", fmt.Errorf("unexpectedly received multiple rows in '%s': %s", command, rows)
-	}
-	row := rows[0]
-	if len(row) != 1 {
-		return "", fmt.Errorf("unexpectedly received multiple columns in '%s': %s", command, row)
-	}
-	branchName, ok := row[0].(string)
-	if !ok {
-		return "", fmt.Errorf("unexpectedly received non-string column in '%s': %s", command, row[0])
-	}
-	return branchName, nil
-}
-
 func printBranches(sqlCtx *sql.Context, queryEngine cli.Queryist, apr *argparser.ArgParseResults, _ cli.UsagePrinter) int {
 	branchSet := set.NewStrSet(apr.Args)
 
--- a/go/cmd/dolt/commands/branch.go
+++ b/go/cmd/dolt/commands/branch.go
@@ -106,7 +106,7 @@ func (cmd BranchCmd) Exec(ctx context.Context, commandStr string, args []string,
 
 	queryEngine, sqlCtx, closeFunc, err := cliCtx.QueryEngine(ctx)
 	if err != nil {
-		return HandleVErrAndExitCode(errhand.BuildDError("error: failed create query engine").AddCause(err).Build(), nil)
+		return HandleVErrAndExitCode(errhand.BuildDError("error: failed to create query engine").AddCause(err).Build(), nil)
 	}
 
 	if closeFunc != nil {
@@ -125,7 +125,7 @@ func (cmd BranchCmd) Exec(ctx context.Context, commandStr string, args []string,
 	case apr.Contains(cli.ListFlag):
 		return printBranches(sqlCtx, queryEngine, apr, usage)
 	case apr.Contains(showCurrentFlag):
-		return printCurrentBranch(dEnv)
+		return printCurrentBranch(sqlCtx, queryEngine)
 	case apr.Contains(datasetsFlag):
 		return printAllDatasets(ctx, dEnv)
 	case apr.NArg() > 0:
@@ -178,7 +178,8 @@ func getBranches(sqlCtx *sql.Context, queryEngine cli.Queryist, remote bool) ([]
 }
 
 func getActiveBranchName(sqlCtx *sql.Context, queryEngine cli.Queryist) (string, error) {
-	schema, rowIter, err := queryEngine.Query(sqlCtx, "SELECT active_branch()")
+	command := "SELECT active_branch()"
+	schema, rowIter, err := queryEngine.Query(sqlCtx, command)
 	if err != nil {
 		return "", err
 	}
@@ -187,15 +188,15 @@ func getActiveBranchName(sqlCtx *sql.Context, queryEngine cli.Queryist) (string,
 		return "", err
 	}
 	if len(rows) != 1 {
-		return "", fmt.Errorf("unexpectedly received multiple rows in 'SELECT active_branch': %s", rows)
+		return "", fmt.Errorf("unexpectedly received multiple rows in '%s': %s", command, rows)
 	}
 	row := rows[0]
 	if len(row) != 1 {
-		return "", fmt.Errorf("unexpectedly received multiple columns in 'SELECT active_branch': %s", row)
+		return "", fmt.Errorf("unexpectedly received multiple columns in '%s': %s", command, row)
 	}
 	branchName, ok := row[0].(string)
 	if !ok {
-		return "", fmt.Errorf("unexpectedly received non-string column in 'SELECT active_branch': %s", row[0])
+		return "", fmt.Errorf("unexpectedly received non-string column in '%s': %s", command, row[0])
 	}
 	return branchName, nil
 }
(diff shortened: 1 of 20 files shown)


go/cmd/dolt/commands/branch.go
```diff
@@ -103,89 +106,163 @@ func (cmd BranchCmd) Exec(ctx context.Context, commandStr string, args []string,
...
-				h, err := cm.HashOf()
-
-				if err != nil {
-					return HandleVErrAndExitCode(errhand.BuildDError("error: failed to hash commit").AddCause(err).Build(), nil)
-				}
-
-				commitStr = h.String()
-			}
+			commitStr = branch.hash
 		}
```
macneale4, line 267: Can you add a comment to future readers that this silliness is the result of the color characters in the branchName strings. This makes me very sad.

go/cmd/dolt/commands/branch.go
```diff
@@ -243,159 +320,131 @@ func printAllDatasets(ctx context.Context, dEnv *env.DoltEnv) int {
...
+	if apr.Contains(cli.VerboseFlag) {
+		cli.PrintErrln("--verbose/-v can only be supplied when listing branches, not when deleting branches")
+		return 1
 	}
 
-	err := actions.CreateBranchWithStartPt(ctx, dEnv.DbData(), newBranch, startPt, apr.Contains(cli.ForceFlag), nil)
+	return runQuery(sqlCtx, queryEngine, args)
+}
+
+func runQuery(sqlCtx *sql.Context, queryEngine cli.Queryist, args []string) int {
```
macneale4, line 439: Rename this function, and add a comment about this. copyBranch, createBranch, deleteBranches, and moveBranch all come here - after they validate arguments. But other commands such as printBranches do not.

`runQuery` would imply to me that you are passing is a query - but you're actually generating the query in this function. And generateBranchSql is only called from here. So make that function be immediately follow this one to indicate they are strongly tied together.

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -53,7 +53,9 @@ func doDoltBranch(ctx *sql.Context, args []string) (int, error) {
 		return 1, fmt.Errorf("Empty database name.")
 	}
 
-	apr, err := cli.CreateBranchArgParser().Parse(args)
+	ap := cli.CreateBranchArgParser()
+	ap.SupportsFlag(cli.RemoteParam, "r", "Delete a remote tracking branch.")
```
macneale4, line 57: Looks like this is using a different ArgParser than the one in branch.go. Is there a reason for that? Should we consolidate? At the very lest, add the -r flag to the original definition. Its weird to add it here.

integration-tests/bats/branch.bats
```diff
@@ -143,4 +143,84 @@ teardown() {
...
+    [ "$status" -ne 0 ]
+    [[ "$output" =~ "--all/-a can only be supplied when listing branches, not when moving branches" ]] || false
+
+    run dolt branch -a new
+    [ "$status" -ne 0 ]
+    [[ "$output" =~ "--all/-a can only be supplied when listing branches, not when creating branches" ]] || false
+}
+
+@test "branch: -v can only be supplied when listing branches" {
+    run dolt branch -v
```
macneale4, line 180: If you want to run a command and you expect it to return 0, it's better to just run it bare. There is more debugging information printed when it fails.

`dolt branch -v`

(and then can skip checking the status).

macneale4 (changes requested):
Looking good so far! There are some structural changes that could help future readers of this code. Should be quick changes

go/cmd/dolt/commands/branch.go
macneale4: f'ing github. I don't know why sometimes i comments on line that are different from what I want to comment on. I'm talking about this line:

`		fmtStr := fmt.Sprintf("%%s%%%ds\t%%s", 48-branchLen)`

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
nicktobey: The reason is that `dolt_branch.go` is only for actions that modify branches, not ones that only list branches.

The common flags are defined in CreateBranchArgParser.

Both versions have a `-r` flag, but their descriptions are different.

I added comments to explain.

integration-tests/bats/branch.bats
nicktobey: Oh right, I knew that.

Fixed.

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -53,7 +53,11 @@ func doDoltBranch(ctx *sql.Context, args []string) (int, error) {
 		return 1, fmt.Errorf("Empty database name.")
 	}
 
-	apr, err := cli.CreateBranchArgParser().Parse(args)
+	// CreateBranchArgParser has the common flags for the command line and the stored procedure.
+	// The stored procedure doesn't support all actions, so we have a shorter description for -r.
```
macneale4, line 57: Is there a reason for the limitation? Should the stored procedure be 1-1 with the CLI, especially since you are leveraging it?

go/cmd/dolt/commands/branch.go
```diff
@@ -103,91 +109,166 @@ func (cmd BranchCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+	}
+}
+
+type branchMeta struct {
+	name   string
+	hash   string
+	remote bool
+}
+
+func getBranches(sqlCtx *sql.Context, queryEngine cli.Queryist, remote bool) ([]branchMeta, error) {
```
macneale4, line 160: Are there other commands which will need this? Possibly put it in utils.go, but ok to punt until we need it.

go/cmd/dolt/commands/branch.go
```diff
@@ -243,159 +324,186 @@ func printAllDatasets(ctx context.Context, dEnv *env.DoltEnv) int {
 	return 0
 }
 
-func moveBranch(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgParseResults, usage cli.UsagePrinter) int {
-	if apr.NArg() != 2 {
-		usage()
-		return 1
+// generateBranchSql returns the query that will call the `DOLT_BRANCH` stored procedure.
+// This function assumes that the inputs are validated branch names or flags, neither of which can contain quotes.
```
macneale4, line 328: Tracking back up the stack, it's not immediately obvious that all validation for these args is done at line 126. Rather than say this "function assumes" I'd say it required prior validation - and callout where that is happening above. I'm not wild about this though; it's frail and will get broken by someone in the future.

Ultimately the interpolate query solution that Steph is working on is a better solution all around. Can you take a look at that - should be safer and more future proof.

macneale4 (changes requested):
A few small follow ups. Getting very close!

go/cmd/dolt/commands/branch.go
```diff
@@ -103,91 +109,166 @@ func (cmd BranchCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+		rowStrings, err := sqlfmt.SqlRowAsStrings(row, schema)
+		if err != nil {
+			return nil, err
+		}
+
+		branches = append(branches, branchMeta{name: rowStrings[0], hash: rowStrings[1], remote: remote})
+	}
+}
+
+func getActiveBranchName(sqlCtx *sql.Context, queryEngine cli.Queryist) (string, error) {
```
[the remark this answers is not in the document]
nicktobey: Moved this into utils.go and combined it with Pavel's implementation.

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
nicktobey: The stored procedure is only for modifying branches. We use system tables for reading them.

The CLI does both. Therefore, it has additional flags.

go/cmd/dolt/commands/branch.go
nicktobey: Maybe, but `dolt branch` is the canonical way to list branches, so until we need to, it makes sense to put it here.

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -854,6 +858,14 @@ func (d *DoltSession) SetRoots(ctx *sql.Context, dbName string, roots doltdb.Roo
 	return d.SetWorkingSet(ctx, dbName, workingSet)
 }
 
+func (d *DoltSession) SetFileSystem(fs filesys.Filesys) {
```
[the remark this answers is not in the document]
nicktobey: As discussed offline, `DoltSession` only exists when the process has a dolt session. For clients connecting to a server remotely, there's no DoltSession object on the the client (they get a BaseSession instead). So this shouldn't be a problem.

go/cmd/dolt/commands/branch.go
nicktobey: Done.

macneale4 (approved):
Once CI completes, free to ship. Woot!

PR #6138, opened 2023-06-12
update dolt commit to use sql queries

stephkyou:
This change updates `dolt commit` to use the appropriate sql engine to generate results. 


macneale4 (changes requested):

go/cmd/dolt/commands/commit.go
```diff
@@ -103,176 +113,150 @@ func performCommit(ctx context.Context, commandStr string, args []string, dEnv *
...
 		}
 	}
 
-	t := datas.CommitNowFunc()
-	if commitTimeStr, ok := apr.GetValue(cli.DateParam); ok {
-		var err error
-		t, err = cli.ParseDate(commitTimeStr)
+	// process query through prepared statement to prevent sql injection
+	query, params := callDoltCommitStoredProc(msg, apr)
+	q, err := dbr.InterpolateForDialect(query, params, dialect.MySQL)
```
macneale4, line 155: Let's avoid single char variables. Name 'q' specifically what it is. I suggest: interpolatedQuery

go/cmd/dolt/commands/commit.go
```diff
@@ -103,176 +113,150 @@ func performCommit(ctx context.Context, commandStr string, args []string, dEnv *
...
-		}
+	return 0, false
+}
 
-		err = dEnv.UpdateStagedRoot(ctx, newRoots.Staged)
-		if err != nil {
-			return handleResetError(err, usage), false
+// callDoltCommitStoredProc generates the sql query necessary to call the DOLT_COMMIT() stored procedure with placeholders
+// for arg input. Also returns a list of the inputs in the order in which they appear in the query.
+func callDoltCommitStoredProc(msg string, apr *argparser.ArgParseResults) (string, []interface{}) {
```
macneale4, line 184: This function needs a different name. maybe: constructParameterizedQuery

go/cmd/dolt/commands/commit.go
```diff
@@ -103,176 +113,150 @@ func performCommit(ctx context.Context, commandStr string, args []string, dEnv *
...
+	q, err := dbr.InterpolateForDialect(query, params, dialect.MySQL)
+	if err != nil {
+		cli.Println(err.Error())
+		return 1, false
+	}
 
-		if err != nil {
-			return HandleVErrAndExitCode(errhand.BuildDError("error: invalid date").AddCause(err).Build(), usage), false
-		}
+	_, _, err = queryist.Query(sqlCtx, q)
```
macneale4, line 161: Rather than drop the results and follow up with a log call, let's grab the hash here in the non-empty case. I _think_ you can skip the log call most of the time. I'm not sure, I'll leave that investigation up to you.

go/cmd/dolt/commands/commit.go
```diff
@@ -376,38 +364,46 @@ func checkIsTerminal() bool {
...
-	n, err = PrintDiffsNotStaged(ctx, dEnv, buf, notStagedTblDiffs, true, false, n, as)
+	n := printStagedDiffs(buf, stagedRows, true)
+	n, err = PrintDiffsNotStaged(sqlCtx, queryist, buf, notStagedRows, true, false, n)
 	if err != nil {
 		return "", err
 	}
 
-	currBranch, err := dEnv.RepoStateReader().CWBHeadRef()
+	// get current branch
+	schema, ri, err = queryist.Query(sqlCtx, "select active_branch()")
```
macneale4, line 397: Seems like this would be useful elsewhere. Can you create a getCurrentBranch() function so other cli commands can get at it? Nick and Paval may already have one somewhere.

integration-tests/bats/commit.bats
```diff
@@ -42,6 +42,46 @@ teardown() {
     [[ "$output" =~ "t2" ]] || false
 }
 
+@test "commit: --all (-a) adds existing, changed tables" {
```
macneale4, line 45: Update `dolt/integration-tests/bats/helper/local-remote.bash` and remove the commit.bats line. This will ensure your changes run in both modes.

macneale4 (changes requested):
This is getting really close!

go/cmd/dolt/commands/commit.go
stephkyou: I'm not sure we need the hash here, once `log` is migrated, we can switch back to a call of `dolt log -n-1` which wouldn't need the hash anyways. With the log call we can print the message as well which some of our bats tests check for.

go/cmd/dolt/commands/commit.go
```diff
@@ -394,15 +401,10 @@ func buildInitalCommitMsg(ctx context.Context, sqlCtx *sql.Context, queryist cli
 	}
 
 	// get current branch
-	schema, ri, err = queryist.Query(sqlCtx, "select active_branch()")
-	if err != nil {
-		return "", err
-	}
-	row, err := sql.RowIterToRows(sqlCtx, schema, ri)
+	currBranch, err := getBranchName(queryist, sqlCtx)
```
macneale4, line 404: Can you move that function our of status.go and into utils.go file.

macneale4 (approved):
Looks like there are still a fair number of bats failures. If there is nothing substantial needed to fix those, go ahead and ship it!

PR #6142, opened 2023-06-12
New interface method to return underlying dolt dbs from a SqlDatabase…

zachmu:
… implementation (rather than type switch)


max-hoffman (approved):
lgtm

macneale4 (approved):
LGTM

PR #6158, opened 2023-06-15
improve valid identifer logic

JCOR11599:
Changed our regex for identifiers to better match MySQL.
Now, we allow all characters within the unicode range `[\u0001-\uFFFF]`.
However, to utilize characters outside of ASCII range (most of unicode) they must be back-quoted
```sql
create table `ドルト` (i int primary key);
```
This works in MySQL without the backquotes, but that is a separate fix likely stemming from vitess.

This is now valid in dolt and MySQL, which might be bad for our branching capabilities.
```sql
create table `branch/table` (i int primary key);
```

Fix for: https://github.com/dolthub/dolt/issues/6156
Docs: https://dev.mysql.com/doc/refman/8.0/en/identifiers.html


go/libraries/doltcore/doltdb/table.go
```diff
@@ -34,43 +32,30 @@ import (
...
+	}
+	return IsValidIdentifier(name)
 }
 
-// IsValidIndexName returns true if the name matches the regular expression IndexNameRegexStr.
-func IsValidIndexName(name string) bool {
-	return indexNameRegex.MatchString(name)
+// IsValidIdentifier returns true according to MySQL's quoted identifier rules.
+// Docs here: https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
+func IsValidIdentifier(name string) bool {
```
JCOR11599, line 48: I initially wanted to implement this using a regex, but it appears that the specific syntax needed is not supported
Ref: https://github.com/google/re2/wiki/Syntax

I tried
```
^[\u0001-\uFFFF]+   // doesn't compile
^[\%u0001-\%uFFFF]+ // doesn't compile
^[\%u0001-FFFF]+    // does compile but wrong
```

go/libraries/doltcore/doltdb/table.go
```diff
@@ -34,43 +32,30 @@ import (
 	"github.com/dolthub/dolt/go/store/types"
 )
 
-const (
```
JCOR11599, line 37: These regex were following MySQL restrictions for unquoted identifiers, but any questionable unquoted identifers shouldn't make it this far. 
For example
```sql
create table 123(i int); -- syntax error in vitess
create table a-a(i int); -- syntax error in vitess
create table []<>{}(i int); -- syntax error in vitess
```

go/libraries/doltcore/doltdb/table.go
```diff
@@ -34,43 +32,30 @@ import (
...
-var (
-	tableNameRegex      = regexp.MustCompile(TableNameRegexStr)
-	foreignKeyNameRegex = regexp.MustCompile(ForeignKeyNameRegexStr)
-	indexNameRegex      = regexp.MustCompile(IndexNameRegexStr)
-
-	ErrNoConflictsResolved = errors.New("no conflicts resolved")
-)
+var ErrNoConflictsResolved = errors.New("no conflicts resolved")
 
 // IsValidTableName returns true if the name matches the regular expression TableNameRegexStr.
```
zachmu, line 37: Comment is out of date

zachmu (approved):
LGTM

PR #6162, opened 2023-06-15
Savepoint operations now support multiple databases in a transaction

zachmu:


go/libraries/doltcore/sqle/dsess/transactions.go
```diff
@@ -722,26 +722,24 @@ func (tx *DoltTransaction) findSavepoint(name string) int {
 
 // RollbackToSavepoint returns the root value associated with the savepoint name given, or nil if no such savepoint can
```
fulghum, line 723:
```suggestion
// RollbackToSavepoint returns the root values for all applicable databases associated with the savepoint name given, or nil if no such savepoint can
```

fulghum (approved):
Looks good! 🚢

PR #6163, opened 2023-06-15
add hash column to result set for `dolt_merge(..., '--no-ff')`

JCOR11599:
fix for: https://github.com/dolthub/dolt/issues/6159


go/libraries/doltcore/sqle/dprocedures/init.go
```diff
@@ -36,7 +36,7 @@ var DoltProcedures = []sql.ExternalStoredProcedureDetails{
 	// dolt_gc is enabled behind a feature flag for now, see dolt_gc.go
 	{Name: "dolt_gc", Schema: int64Schema("success"), Function: doltGC},
 
-	{Name: "dolt_merge", Schema: int64Schema("fast_forward", "conflicts"), Function: doltMerge},
+	{Name: "dolt_merge", Schema: stringSchema("hash", "fast_forward", "conflicts"), Function: doltMerge},
```
fulghum, line 39: It might be safer to keep the fast_forward and conflicts fields as ints – someone could be calling merge and loading the results into variables and they might get a type conflict if they come back as strings depending on the language and library they're using. For dolt_cherry_pick, I just created a quick var in that procedure's implementation file and defined a sql.Schema that I reused in init.

fulghum (approved):
Looks good!

PR #6164, opened 2023-06-15
Bug fix: Enable `dolt_cherry_pick` stored procedure to work with `@@autocommit`

fulghum:
Previously, when `dolt_cherry_pick` encountered a data or schema conflict, or a constraint violation, it would return an error with a message about the problem and how to resolve. This works well when `@@autocommit=0`, but when `@@autocommit=1` the error causes the automatic transaction management to automatically rollback the changes, which prevents users from being able to examine and fix the conflicts. 

This PR changes `dolt_cherry_pick` to be more similar to how `dolt_merge` reports conflicts – instead of returning an error message, `dolt_cherry_pick` now returns additional fields that show the number of tables with data conflicts, number of tables with schema conflicts, and number of tables with constraint violations.


JCOR11599 (approved):
LGTM

PR #6176, opened 2023-06-16
Enable authentication in the CLI commands migrated to sql backend

macneale4:
Enable Authentication in the following ways:
 * New Global Flag --password
 * DOLT_CLI_PASSWORD environment variable
 * Ask for a password with a prompt.
 * Automatic authentication to a server using the secret in the sql-server.lock file

One significant change in behavior is that previously if a user presented a non-sense username/password, we'd accept it as a super user identity. If a real user was specified, we would promote that user to a super user - regardless of if the password was correct.

Now, if a --user flag is presented, the user must present a password by flag, env var or prompt. If the user/pwd combination is not a known user, the command will fail. This applies to both local and remote mode.

Important to call out that this isn't about security. If you want to be a super user, you can just _not_ provide a --user. This behavior is to enable consistent behavior of client applications where they need to test permissions. We were making this impossible before because bad credentials were being promoted.


go/cmd/dolt/cli/credentials.go
```diff
@@ -0,0 +1,95 @@
...
+		return newParsedArgs, &UserPassword{Username: userId, Password: password, Unspecified: false}, nil
+	}
+
+	if hasUserId && !hasPassword {
+		password = ""
+		val, hasVal := os.LookupEnv(DOLT_ENV_PWD)
+		if hasVal {
+			password = val
+		} else {
+			Printf("Enter password: ")
```
PavelSafronov, line 67: This is a very simple prompt which doesn't convey enough information to the user, like what the app is, which user are we getting the password for, etc. (If I saw this random prompty, I'd probably ctrl+C and try to figure out who's asking me and why.)

What about something like this?
`Please enter the password to Dolt DB FancyDB1, user account spacelove:`

go/cmd/dolt/cli/credentials.go
```diff
@@ -0,0 +1,95 @@
...
+			_, err = fmt.Scan(&password)
+			if err != nil {
+				return nil, nil, err
+			}
+			Println()
+		}
+		return newParsedArgs, &UserPassword{Username: userId, Password: password, Unspecified: false}, nil
+	}
+
+	testOverride, hasTestOverride := os.LookupEnv(DOLT_SILENCE_USER_REQ_FOR_TESTING)
```
PavelSafronov, line 87: Should this override happen before we prompt for the password?

go/cmd/dolt/cli/credentials.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+
+	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+	"golang.org/x/term"
+)
+
+type UserPassword struct {
+	Username    string
+	Password    string
+	Unspecified bool // If true, the user and password were not provided by the user.
```
PavelSafronov, line 28: Nit/personal pref: remove unnecessary negation here, it's confusing. Maybe change the field to "IsUserSupplied"?

go/cmd/dolt/cli/flags.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+// Constants for command line flags names. These tend to be used in multiple places, so definding
```
PavelSafronov, line 17: Nit: "defining"

go/cmd/dolt/cli/flags.go
```diff
@@ -0,0 +1,64 @@
...
+	SetUpstreamFlag  = "set-upstream"
+	ShallowFlag      = "shallow"
+	ShowIgnoredFlag  = "ignored"
+	SkipEmptyFlag    = "skip-empty"
+	SoftResetParam   = "soft"
+	SquashParam      = "squash"
+	TheirsFlag       = "theirs"
+	TrackFlag        = "track"
+	UpperCaseAllFlag = "ALL"
+	UserFlag         = "user"
```
PavelSafronov, line 63: Non-blocking question: this constant is being renamed from `UserParam` to `UserFlag`. What's the conceptual difference and why the rename?

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -358,41 +390,47 @@ func SetupDoltConfig(dEnv *env.DoltEnv, apr *argparser.ArgParseResults, config S
 	return nil
 }
 
-// getCommandLineServerConfig sets server config variables and persisted global variables with values defined on command line.
-// If not defined, it sets variables to default values.
-func getCommandLineServerConfig(apr *argparser.ArgParseResults) (ServerConfig, error) {
-	serverConfig := DefaultServerConfig()
+// getCommandLineConfig sets server config variables and persisted global variables with values defined on command line.
+// If not defined, it sets variables to default values. The creds option is available when building a client config, which
+// is used for most commands. `dolt sql-server` is special, and it's user/pwd is passed in vie apr, so creds must be nil
```
PavelSafronov, line 395: Nit:
- `it's` -> `its`
- `vie` -> `via`

go/libraries/utils/argparser/args_test.go
```diff
@@ -280,3 +280,48 @@ func TestValidation(t *testing.T) {
...
+	ap.SupportsString("string", "", "string_value", "A string")
+	ap.SupportsFlag("flag", "", "A flag")
+
+	apr, err := ap.Parse([]string{"--string", "str", "--flag", "1234"})
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	newApr1 := apr.DropValue("string")
+	if apr.Equals(newApr1) {
```
PavelSafronov, line 296: Nit: avoid manual comparisons and use `require.NotEqual`?

PavelSafronov (approved):
Good changes and tests. I've left a few minor comments and a few question, but otherwise this looks great, approving.

go/cmd/dolt/cli/credentials.go
macneale4: I changes the `Specifcied` because I think it's a good middle ground. It's an internal struct with a comment. Seems ok to me.

go/cmd/dolt/cli/flags.go
macneale4: Seemed like the right thing to do? Make them consistent? Not a hard opinion.

PR #6185, opened 2023-06-20
Add `dolt_schema_diff` table function

PavelSafronov:
`dolt_schema_diff` will return the schema diffs between refs and optionally will filter those changes to a specific table.

This is a new table function that provides us with the information we need for `dolt diff`.

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
@@ -332,7 +332,6 @@ func (ds *SchemaDiffTableFunction) RowIter(ctx *sql.Context, row sql.Row) (sql.R
 			"",         // 4
 		}
 
-		schemaDiffsFound := false
 		statements, err := diff.GetNonCreateNonDropTableSqlSchemaDiff(delta, toSchemas, fromSchema, toSchema)
 		if err != nil {
 			return nil, err
@@ -341,11 +340,6 @@ func (ds *SchemaDiffTableFunction) RowIter(ctx *sql.Context, row sql.Row) (sql.R
 			row := baseRow.Copy()
 			row[4] = stmt
 			dataRows = append(dataRows, row)
-			schemaDiffsFound = true
-		}
-
-		if !schemaDiffsFound {
-			dataRows = append(dataRows, baseRow)
 		}
 	}
 


go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
```diff
@@ -0,0 +1,420 @@
...
+var _ sql.TableFunction = (*SchemaDiffTableFunction)(nil)
+var _ sql.ExecSourceRel = (*SchemaDiffTableFunction)(nil)
+
+type SchemaDiffTableFunction struct {
+	ctx *sql.Context
+
+	fromCommitExpr sql.Expression
+	toCommitExpr   sql.Expression
+	dotCommitExpr  sql.Expression
+	tableNameExpr  sql.Expression
```
nicktobey, line 42: Can you add comments describing what these expressions are for?

nicktobey (commented):
This also needs documentation at https://docs.dolthub.com/sql-reference/version-control/dolt-sql-functions and tests for failure states.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
```diff
@@ -0,0 +1,420 @@
...
+func (ds *SchemaDiffTableFunction) WithChildren(children ...sql.Node) (sql.Node, error) {
+	if len(children) != 0 {
+		return nil, fmt.Errorf("unexpected children")
+	}
+	return ds, nil
+}
+
+// CheckPrivileges implements the interface sql.Node.
+func (ds *SchemaDiffTableFunction) CheckPrivileges(ctx *sql.Context, opChecker sql.PrivilegedOperationChecker) bool {
+	if ds.tableNameExpr == nil || !types.IsText(ds.tableNameExpr.Type()) {
```
nicktobey, line 137: Why does the table name not being text mean that the user has privileges? This might warrant a comment.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
```diff
@@ -0,0 +1,420 @@
...
+
+	newDstf := *ds
+	if strings.Contains(expression[0].String(), "..") {
+		newDstf.dotCommitExpr = expression[0]
+		if len(expression) > 1 {
+			newDstf.tableNameExpr = expression[1]
+		}
+	} else {
+		if len(expression) < 2 {
+			return nil, sql.ErrInvalidArgumentNumber.New(newDstf.Name(), "1 to 3", len(expression))
```
nicktobey, line 190: This is if the user has one argument that isn't a dot expression, right? I don't think InvalidArgumentNumber is the right error here. More like ErrInvalidArgumentDetails.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
PavelSafronov: Adding.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
PavelSafronov: Good point. In other cases, if the text isn't a string, we return `false`. Updating.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
PavelSafronov: Thanks for pointing out ErrInvalidArgumentDetails, changing this error to be clearer.

go/libraries/doltcore/sqle/dolt_schema_diff_table_function.go
nicktobey: My question is less "what should we do here?" but "why do we do what we do?" It's not clear to me why we should return any particular value in this case. Is it because we expect the tableName to be text but the function doesn't allow us to return an error?

PavelSafronov: Very good question, but I don't have an answer yet.
We are definitely checking the type in `WithExpressions`, so in theory there should be no way for a wrong type of expression to make it into this command. But I'm not, and all of our other implementations of `CheckPrivileges` return `false` when the expressions are of the wrong type, and all of those functions have those same checks as well.
@fulghum , do you know?

fulghum: Generally, when I see this in a `CheckPrivileges` function, it's because the user could pass in other expressions that don't have a string type (technically we would/should coerce anything to a string type, but we're being a little restrictive here since we know we need a valid string value that isn't coerced from another type) so before we cast to a string, we do this check and fail the privilege check if it's not a string. This is a little bit of extra defense, since validation errors in other places should catch this before the plan is going to be executed and starts checking privileges. In this case, it doesn't look like we ever cast the value to a string, so we probably don't need this extra check here.

go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
```diff
@@ -4862,6 +4862,99 @@ var CommitDiffSystemTableScriptTests = []queries.ScriptTest{
...
+		Name: "basic schema changes",
+		SetUpScript: []string{
+			"create table test (pk int primary key, c1 int, c2 int);",
+			"call dolt_add('.')",
+			"call dolt_commit('-m', 'commit 1');",
+			"alter table test drop column c2, add column c3 varchar(10);",
+			"call dolt_add('.')",
+			"call dolt_commit('-m', 'commit 2');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
```
macneale4, line 4876: There seems to be a high reliance on HEAD^ and HEAD in these tests. Adding more commits would be a good idea, and tests with commit ids and tags would be a good addition as well.

The reason I call this out is that there are default situations where we use HEAD^ and HEAD - by virtue of that being the top of the history. Having more changes, then calculating the schema changes between them would surface more problems. For example, what if you compare two commits which don't have any schema differences even though they are different commits?

PR #6187, opened 2023-06-20
Add `dolt_foreign_key_status` table function.

PavelSafronov:
`dolt_foreign_key_status` will return the status/information for a foreign key, including its resolved status and the names of the columns involved.
This is a new table function that provides us with the information we need for `dolt diff`.

### DIFF
--- a/go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
@@ -16,12 +16,13 @@ package sqle
 
 import (
 	"fmt"
-	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
-	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
-	"github.com/dolthub/go-mysql-server/sql"
-	"github.com/dolthub/go-mysql-server/sql/types"
 	"io"
 
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/types"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
+	"github.com/dolthub/dolt/go/libraries/doltcore/schema"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess"
 )
 
--- a/go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
+++ b/go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
@@ -1,0 +1,341 @@
+// Copyright 2022 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


fulghum:
Any chance of being able to pull this information from `information_schema` instead? I think the `is_resolved` bit is probably the only thing that wouldn't be there. How badly do we need to expose that bit to customers? Seems like it'd be nice to keep `is_resolved` internal so customers don't need to know about that detail, but maybe we need it for the diff?

PavelSafronov:
> Any chance of being able to pull this information from `information_schema` instead? I think the `is_resolved` bit is probably the only thing that wouldn't be there. How badly do we need to expose that bit to customers? Seems like it'd be nice to keep `is_resolved` internal so customers don't need to know about that detail, but maybe we need it for the diff?

You're right that most of this information is accessible from `information_schema`, but for `dolt diff` we do need to know if a FK is resolved or not, and this information is definitely not in SQL. Daylon suggested this very function:
image

nicktobey:
This definitely needs to be documented in https://docs.dolthub.com/sql-reference/version-control/dolt-sql-functions

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -4906,3 +4906,100 @@ END`,
 		},
 	},
 }
+
+var DoltForeignKeyStatus = []queries.ScriptTest{
+	{
+		Name: "Show unresolved foreign key status",
```
nicktobey, line 4912: You should also have tests that check for the different failure conditions in your implementation, like calling the function with inputs of the wrong type, or with unresolvable references, etc.

go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
```diff
@@ -0,0 +1,342 @@
...
+	return rows, nil
+}
+
+func addColToAllRows(rows []sql.Row, colIndex int, data ...string) []sql.Row {
+	newRows := []sql.Row{}
+	for _, datum := range data {
+		for _, row := range rows {
+			newRow := row.Copy()
+			newRow[colIndex] = datum
+			newRows = append(newRows, newRow)
```
nicktobey, line 297: I don't think you need to make a new row array each time;  just pass in a pointer to the row array and modifying the rows.

nicktobey (commented):
Looks good, just a couple of comments about tests and documentation.

go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
```diff
@@ -0,0 +1,342 @@
...
+		ctx:      ctx,
+		database: db,
+	}
+
+	node, err := newInstance.WithExpressions(expressions...)
+	if err != nil {
+		return nil, err
+	}
+
+	ds = node.(*ForeignKeyStatusTableFunction)
```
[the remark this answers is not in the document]
PavelSafronov: Removed the cast, returning `node` now.

go/libraries/doltcore/sqle/dolt_foreign_key_status_table_function.go
PavelSafronov: We do want to make copies of rows.
The idea here is that we need to return a list of fields for one of the columns.
We're doing this for multiple columns, but let's just consider `table_column`. So if we have 2 values for this column, "foo" and "bar", we want to produce 2 almost-identical  rows of values: (a,b,c,foo,d) and (a,b,c,bar,d).

PavelSafronov:
After offline discussion, it was decided that foreign key status (resolved vs not resolved) should not be surfaced to users, so I'm closing this PR without merging.

PR #6188, opened 2023-06-20
Reverting changes to database name management: branch-qualified databases now included in info_schema tables

zachmu:
Fixes https://github.com/dolthub/dolt/issues/6173


fulghum (approved):
Core logic looks good. Looks like it's just fixing up the BATS and MySQL client tests now. 🚢

PR #6190, opened 2023-06-20
Ignore the sql-server.lock file if the PID in the file matches current dolt process

macneale4:
https://github.com/dolthub/dolt/issues/6183


reltuk (approved):
LGTM!

PR #6191, opened 2023-06-21
Use a table's `workingRoot` when getting declared FKs.

PavelSafronov:
Fixes https://github.com/dolthub/dolt/issues/6178

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
@@ -49,7 +49,7 @@ var skipPrepared bool
 // SkipPreparedsCount is used by the "ci-check-repo CI workflow
 // as a reminder to consider prepareds when adding a new
 // enginetest suite.
-const SkipPreparedsCount = 83
+const SkipPreparedsCount = 82
 
 const skipPreparedFlag = "DOLT_SKIP_PREPARED_ENGINETESTS"
 
--- a/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
@@ -49,7 +49,7 @@ var skipPrepared bool
 // SkipPreparedsCount is used by the "ci-check-repo CI workflow
 // as a reminder to consider prepareds when adding a new
 // enginetest suite.
-const SkipPreparedsCount = 82
+const SkipPreparedsCount = 83
 
 const skipPreparedFlag = "DOLT_SKIP_PREPARED_ENGINETESTS"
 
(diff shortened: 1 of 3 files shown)


macneale4 (changes requested):
The behavior change looks simple enough - I don't really know that code. Create new tests and I think I'll have the confidence that the change works.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -99,11 +99,15 @@ var ShowCreateTableScriptTests = []queries.ScriptTest{
 			"set @Commit2 = '';",
 			"set @Commit3 = '';",
 			"set @Commit0 = hashof('main');",
-			"create table a (pk int primary key, c1 int);",
```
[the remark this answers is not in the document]
PavelSafronov: Existing tests kept, introduced these as a new set of tests.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -1474,6 +1474,24 @@ func TestDescribeTableAsOf(t *testing.T) {
...
+	h := newDoltHarness(t)
+	defer h.Close()
+	enginetest.TestScript(t, h, DescribeTableAsOfWithFkScriptTest)
+}
+
+func TestDescribeTableAsOfWithFkPrepared(t *testing.T) {
+	h := newDoltHarness(t)
+	defer h.Close()
+	enginetest.TestScriptPrepared(t, h, DescribeTableAsOfWithFkScriptTest)
+}
```
fulghum, line 1493: (minor) It would be nice to add this one new script test to the existing `TestDescribeTableAsOf` test, just to keep the top level Test functions a little more consolidated. (If we had just one ScriptTest in each top level Test function, it would be a bit unwieldily)

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -284,6 +299,59 @@ var DescribeTableAsOfScriptTest = queries.ScriptTest{
 	},
 }
 
+var DescribeTableAsOfWithFkScriptTest = queries.ScriptTest{
+	Name: "Describe table as of with foreign key constraints",
```
fulghum, line 303: Where are the foreign key constraints in this test?

fulghum (commented):
The change in `tables.go` looks correct to me – it will use the locked root for the table if present. I'm confused about the tests though. I think Neil's feedback was asking you to not change the existing `SHOW CREATE TABLE` test and to add a new test for this issue, but the new test doesn't look correct. It mentions testing FKs, but never actually does anything with FKs and when I revert the fix in `tables.go`, the new test does not catch the bug. 

To respect Neil's feedback, I think you should revert the changes to the ShowCreateTable test, put the new test in the existing set of `DescribeTableAsOf` test scripts, and make sure that new test triggers the bug and validates the fix.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
PavelSafronov: Wrong tests added :(
These are reverted and correct tests are in place. Sorry about that.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
PavelSafronov: Tests consolidated.

fulghum (approved):
Looks good! 🚢

PR #6203, opened 2023-06-21
Bug fixes for resolving the default DB name to the correct branch, was broken in several cases

zachmu:


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -115,36 +115,135 @@ func TestSingleQuery(t *testing.T) {
 
 // Convenience test for debugging a single query. Unskip and set to the desired query.
 func TestSingleScript(t *testing.T) {
-	t.Skip()
+	// t.Skip()
```
fulghum, line 118: Might as well keep this skipped for check-in

```suggestion
	t.Skip()
```

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -137,7 +137,7 @@ func (d *DoltSession) lookupDbState(ctx *sql.Context, dbName string) (*branchSta
 	if dbStateFound {
 		// If we got an unqualified name, use the current working set head
```
fulghum, line 138:
```suggestion
		// If we got an unqualified name, use the checked out working set
```

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -1243,22 +1243,15 @@ func (d *DoltSession) CWBHeadRef(ctx *sql.Context, dbName string) (ref.DoltRef,
...
-	baseName, rev = SplitRevisionDbName(dbName)
-	if rev != "" {
-		return "", false, fmt.Errorf("invalid database name: %s", dbName)
-	}
-
+	baseName := strings.ToLower(dbName)
+	
 	d.mu.Lock()
 	dbState, ok := d.dbStates[baseName]
 	d.mu.Unlock()
```
fulghum: (minor) `baseName` seems _slightly_ misleading here... right? Maybe just inline it? 
```suggestion
	d.mu.Lock()
	dbState, ok := d.dbStates[strings.ToLower(dbName)]
	d.mu.Unlock()
```

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -2119,6 +2119,135 @@ var DoltCheckoutScripts = []queries.ScriptTest{
...
+				Expected: []sql.Row{{"b2"}},
+			},
+			{
+				Query:    "select * from t;",
+				Expected: []sql.Row{{2, 2}},
+			},
+			{
+				Query:    "select * from mydb.t;",
+				Expected: []sql.Row{{1, 1}},
+			},
```
fulghum, line 2163: huh? `select * from t` is not equivalent to qualifying the table with the database name? That seems really surprising and likely to break basic assumptions SQL users have. I think you gotta have `use` change the checked out branch or it's just gunna come back to us as customer questions/issues.

fulghum (approved):
Fix looks good, but I think it's going to *really* confuse customers when they qualify a table name and suddenly pull data from a different branch than they told Dolt to `use`.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
zachmu: Discussed offline. Agree this could be surprising, but after a lot of rounds of design and experiments around this, I'm convinced this is the only way this can work and be consistent in all contexts. I'll keep thinking about how it could be made easier to understand.

fulghum: Agreed; thanks for discussing. I think my mental model for `use mydb/branch` was thinking of it as using the `mydb` database and checking out the `branch` branch. I know we're actively changing that of course, and if you think of `mydb` as a completely separate/distinct database from `mydb/branch` then this new approach does make a lot of sense. I think we just have to explain that in the right way in docs and see if customers still trip over it.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -2217,11 +2441,54 @@ var DoltInfoSchemaScripts = []queries.ScriptTest{
...
+				Expected: []sql.Row{{"b2"}},
+			},
+			{
+				Query:    "/* b2 */ select column_name from information_schema.columns where table_schema = 'mydb' and table_name = 't' order by 1;",
+				Expected: []sql.Row{{"a"}, {"b"}},
+			},
+			{
+				Query:    "select column_name from information_schema.columns where table_schema = 'mydb/b2' and table_name = 't' order by 1;",
+				Expected: []sql.Row{{"a"}, {"b"}, {"c"}},
+			},
```
fulghum, line 2489: (minor) Maybe worth also adding an assertion that runs `SELECT COUNT(*) FROM information_schema.columns where table_schema='mydb/b3'` and verifies no rows?

zachmu: Good call

PR #6204, opened 2023-06-21
sqle: cluster: Add dolt_cluster_transition_to_standby stored procedure.

reltuk:
This new stored procedure can be used to gracefully transition a primary to a standby while not all replicas in the cluster are up and able to become current. In contrast to `dolt_assume_cluster_role`, this stored procedure can only transition to role `standby`, but takes a second integer parameter representing the number of replica servers in the cluster which must be made current in order for the transition to succeed.


zachmu (approved):
LGTM!

PR #6209, opened 2023-06-22
Migrate `dolt checkout` to new CLI framework.

nicktobey:
This PR migrates dolt branch to invoke SQL commands instead of manipulating the database directly. This allows it to work even on remote connections.

We implement this by adding an additional flag to the `dolt_checkout` stored proceudre: `--global`. This tells the SQL environment to persist this checkout for subsequent connections, making it the new default branch for future connections.

If a server is currently running, `dolt_checkout("--global")` will return an error. This is because we try to avoid changing the default branch on a server until we can properly consider the indented effect on existing connections.

Because working sets behave differently in our SQL environment than the command line (command line has a single working set, SQL land has a working set per-branch), we only permit changing the default branch when there would be no observable different in behavior: that is, when both the former and the new branch are clean. If this isn't the case, we return an error. This guarentees that a user calling `dolt checkout` from the command line will never see behavior that differs from the original CLI implementation. They will, at worst, get a helpful error message.


go/cmd/dolt/commands/checkout.go
```diff
@@ -82,237 +83,66 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
-	err := actions.CreateBranchWithStartPt(ctx, dEnv.DbData(), newBranch, startPt, false, nil)
-	if err != nil {
-		return errhand.BuildDError(err.Error()).Build()
+	var branchName string
+	if apr.NArg() > 0 {
+		branchName = apr.Arg(0)
 	}
-
-	return checkoutBranch(ctx, dEnv, newBranch, false)
+	return HandleVErrAndExitCode(handleErrors(branchName, err), usagePrt)
```
macneale4, line 110: The way this code reads is confusing. `HandleVErrAndExitCode(handleErrors(branchName, err)` really makes it sound like we know there was an error, and we need to handle it. I had to dig into both functions to know that this is actually the happy path. What makes this particularly awkward is that branchName was pulled out of the args to just pass it into this error handling method.

What I suggest instead is pull branchName out branch name from the args before executing the query. Then print the success message if you don't receive and error. If you do, then call it the way you are now.

go/cmd/dolt/commands/checkout.go
```diff
@@ -82,237 +83,66 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
+	buffer.WriteString(")")
 
-	return nil
+	return dbr.InterpolateForDialect(buffer.String(), queryValues, dialect.MySQL)
 }
 
-func checkoutBranch(ctx context.Context, dEnv *env.DoltEnv, name string, force bool) errhand.VerboseError {
-	err := actions.CheckoutBranch(ctx, dEnv, name, force)
-
+func handleErrors(name string, err error) errhand.VerboseError {
```
macneale4, line 129: s/name/branchName/g

Also, enforce that an error is always passed in. Your verb/subject in the name says you are going to handle errors. It would never occur to me that there were no errors to handle based on those two words.

macneale4 (changes requested):
Looking good. A couple small style/flow suggestions. Ping me again when the existing and new bats tests are passing.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -0,0 +1,182 @@
...
+	wsMeta := &datas.WorkingSetMeta{
+		Name:        username,
+		Email:       email,
+		Timestamp:   uint64(time.Now().Unix()),
+		Description: "updated from dolt environment",
+	}
+
+	err = doltDb.UpdateWorkingSet(ctx, newWs.Ref(), newWs, wsHash, wsMeta, nil)
+
+	// err = dSess.SetWorkingSet(ctx, dbName)
```
macneale4, line 176: Delete

integration-tests/bats/sql-checkout.bats
macneale4, line 1: I think we need a significant number of tests in sql-local-remote.bats. My thinking is that as you go in and out of each context, we need to be sure that the checked out branch stays consistent. That seems to be a lot of what you had to fix anyway, so I'd like to see tests about it.

macneale4 (changes requested):
Main concern is that we need more tests for testing the state of the current branch when switching between remote and local mode. I'm not convinced that the current local-remote.bash approach covers this well enough.

go/cmd/dolt/commands/checkout.go
```diff
@@ -84,28 +85,46 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
 	}
 
 	_, err = getRowsForSql(queryEngine, sqlCtx, sqlQuery)
 
-	var branchName string
-	if apr.NArg() > 0 {
-		branchName = apr.Arg(0)
+	// This command doesn't modify `dEnv` which could break tests that call multiple commands in sequence.
+	// We must reload it so that it includes changes to the repo state.
+	if dEnv != nil {
```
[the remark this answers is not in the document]
nicktobey: If we migrate every command, we'll be able to remove this.

This only exists because without it, when a go test runs a CliContext command followed by a DoltEnv command, the DoltEnv command has stale information about the repo state. Once there are no more DoltEnv commands, this problem goes away.

go/cmd/dolt/commands/reset.go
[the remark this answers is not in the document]
nicktobey: This didn't cause a conflict; the function was previously used by both reset and checkout.

I could potentially reimplement the codepath to not use it, but I think it's fine the way it is now.

go/cmd/dolt/commands/checkout.go
nicktobey: Done.

go/cmd/dolt/commands/checkout.go
nicktobey: Done.

integration-tests/bats/sql-checkout.bats
nicktobey: I added a test for remote `dolt checkout` that also demonstrates the current limitation that changes to the active branch won't be seen until the server is stopped.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -35,57 +35,108 @@ var ErrEmptyBranchName = errors.New("error: cannot checkout empty string")
...
+func doDoltCheckout(ctx *sql.Context, args []string) (int, string, error) {
 	currentDbName := ctx.GetCurrentDatabase()
 	if len(currentDbName) == 0 {
-		return 1, fmt.Errorf("Empty database name.")
+		return 1, "", fmt.Errorf("Empty database name.")
 	}
 
-	apr, err := cli.CreateCheckoutArgParser().Parse(args)
+	argParser := cli.CreateCheckoutArgParser()
+	argParser.SupportsFlag(cli.MoveFlag, "m", "mimic the behavior of the `dolt checkout` command line, moving the working set into the new branch.")
```
macneale4, line 88: This is a hidden flag which we only expect the CLI to use, so it seems like a comment is justified here.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -35,57 +35,108 @@ var ErrEmptyBranchName = errors.New("error: cannot checkout empty string")
...
+		if err != nil {
+			return false, err
+		}
+		defaultBranch := repoState.CWBHeadRef().GetPath()
+		isModification = isModification || (defaultBranch != branchName)
+	}
+	return isModification, nil
+}
+
+func doDoltCheckout(ctx *sql.Context, args []string) (int, string, error) {
```
macneale4, line 81: gDoc. Function name makes it pretty clear what this does, but what are the outputs?

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -35,57 +35,108 @@ var ErrEmptyBranchName = errors.New("error: cannot checkout empty string")
...
-func doDoltCheckout(ctx *sql.Context, args []string) (int, error) {
+func generateSuccessMessage(newBranch, upstream string) string {
+	result := fmt.Sprintf("Switched to branch '%s'", newBranch)
+	if upstream != "" {
+		result += fmt.Sprintf("\nbranch '%s' set up to track '%s'.", newBranch, upstream)
+	}
+	return result
+}
+
+// willModifyDb determines whether or not this operation is a no-op and can return early with a helpful message.
```
macneale4, line 53: This is a helper function for a helper function. Push lower in the file. Keep the highest abstractions at the top, when possible. Non-blocking

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -179,51 +253,70 @@ func createWorkingSetForLocalBranch(ctx *sql.Context, ddb *doltdb.DoltDB, branch
...
+
+		return remoteRef.GetPath(), nil
 	} else {
-		return fmt.Errorf("'%s' matched multiple (%v) remote tracking branches", branchName, len(remoteRefs))
+		return "", fmt.Errorf("'%s' matched multiple (%v) remote tracking branches", branchName, len(remoteRefs))
 	}
 }
 
-func checkoutNewBranch(ctx *sql.Context, dbName string, dbData env.DbData, apr *argparser.ArgParseResults, rsc *doltdb.ReplicationStatusController) error {
+func checkoutNewBranch(ctx *sql.Context, dbName string, dbData env.DbData, apr *argparser.ArgParseResults, rsc *doltdb.ReplicationStatusController, isGlobal bool) (string, string, error) {
```
macneale4, line 319: Input - isGlobal -> isMove

Output - You added two new strings? What are they? gDocs help, but just naming the output variables goes a long way too.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -179,51 +253,70 @@ func createWorkingSetForLocalBranch(ctx *sql.Context, ddb *doltdb.DoltDB, branch
...
 	}
 
 	if len(remoteRefs) == 0 {
-		return fmt.Errorf("error: could not find %s", branchName)
+		if doltdb.IsValidCommitHash(branchName) && apr.Contains(cli.MoveFlag) {
+
+			// User tried to enter a detached head state, which we don't support.
+			// Inform and suggest that they check-out a new branch at this commit instead.
+
+			return "", fmt.Errorf("dolt does not support a detached head state. To create a branch at this commit instead, run:\n\n"+
```
macneale4, line 268: Consider backticks for multi-line output. It's more readable. Non-blocking

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -179,51 +253,70 @@ func createWorkingSetForLocalBranch(ctx *sql.Context, ddb *doltdb.DoltDB, branch
 
 // checkoutRemoteBranch checks out a remote branch creating a new local branch with the same name as the remote branch
 // and set its upstream. The upstream persists out of sql session.
-func checkoutRemoteBranch(ctx *sql.Context, dbName string, dbData env.DbData, branchName string, apr *argparser.ArgParseResults, rsc *doltdb.ReplicationStatusController) error {
+func checkoutRemoteBranch(ctx *sql.Context, dSess *dsess.DoltSession, dbName string, dbData env.DbData, branchName string, apr *argparser.ArgParseResults, rsc *doltdb.ReplicationStatusController) (string, error) {
```
macneale4, line 256: Returning a string now. What is it?

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -282,18 +375,33 @@ func checkoutNewBranch(ctx *sql.Context, dbName string, dbData env.DbData, apr *
...
+		if err != nil {
+			return "", "", err
+		}
+	}
+
+	return newBranchName, remoteAndBranch, nil
 }
 
-func checkoutBranch(ctx *sql.Context, dbName string, branchName string) error {
+func checkoutBranch(ctx *sql.Context, dbName string, branchName string, apr *argparser.ArgParseResults) error {
```
macneale4, line 404: In a file which also has `doDoltCheckout` and `doltCheckout`, this function blends in. What this function is actually doing is updating the branch reference, so say that.

integration-tests/mysql-client-tests/node/index.js
```diff
@@ -50,7 +50,7 @@ const tests = [
   { q: "call dolt_add('-A');", res: [{ status: 0 }] },
   { q: "call dolt_commit('-m', 'my commit')", res: [] },
   { q: "select COUNT(*) FROM dolt_log", res: [{ "COUNT(*)": 2 }] },
-  { q: "call dolt_checkout('-b', 'mybranch')", res: [{ status: 0 }] },
+  { q: "call dolt_checkout('-b', 'mybranch')", res: [{ status: 0, message: "Switched to branch 'mybranch'"  }] },
```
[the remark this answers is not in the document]
nicktobey: It was necessary because I added the extra column, but you're welcome.

integration-tests/bats/remotes.bats
```diff
@@ -69,17 +69,17 @@ teardown() {
...
     dolt commit --allow-empty -m "a commit for main from repo2"
     dolt push
 
-    run dolt branch
-    [[ ! "$output" =~ "other" ]] || false
+     dolt branch
+
 
     run dolt checkout other
+    echo "$output"
```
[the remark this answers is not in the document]
nicktobey: Fixed.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -854,7 +854,34 @@ var MergeScripts = []queries.ScriptTest{
...
+			},
+		},
+	},
+	{
+		Name: "DOLT_MERGE(--abort) clears staged",
+		SetUpScript: []string{
+			"CREATE TABLE test (pk int primary key);",
+			"INSERT INTO test VALUES (0),(1),(2);",
+			"set autocommit = off;",
+			"CREATE TABLE one_pk (pk1 BIGINT NOT NULL, c1 BIGINT, c2 BIGINT, PRIMARY KEY (pk1));",
```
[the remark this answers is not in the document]
nicktobey: Somehow this test got duplicated. I didn't write it. Removed.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Added both docstring and variable names.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Done.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Since it goes alongside `checkoutNewBranch` and `checkoutRemoteBranch`, I renamed it to `checkoutExistingBranch` and added a docstring.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Moved into `dolt_checkout_helpers.go`

integration-tests/mysql-client-tests/python/mysql.connector-test.py
```diff
@@ -19,10 +19,10 @@
     {"call dolt_add('-A');": [(0,)]},
     {"call dolt_commit('-m', 'my commit')": [('',)]},
     {"select COUNT(*) FROM dolt_log": [(2,)]},
-    {"call dolt_checkout('-b', 'mybranch')": [(0,)]},
+    {"call dolt_checkout('-b', 'mybranch')": [(0, "Switched to branch 'mybranch'")]},
```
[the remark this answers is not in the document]
nicktobey: Confirmed that this file is already using CRLF line endings in main (even as the rest of the repo uses LF). It's weird, but this PR didn't change it.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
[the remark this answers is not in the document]
nicktobey: This code was moved from actions/checkout.go in order to break an import cycle. It does have some changes to use sqlCtx instead of dEnv, but is otherwise the same logic.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Added return variable names.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
nicktobey: Added comment explaining the flag.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -0,0 +1,191 @@
...
+		return err
+	}
+
+	ws, err := dSess.WorkingSet(ctx, dbName)
+	if err != nil {
+		return err
+	}
+
+	// For backwards compatibility we support the branch not having a working set, but generally speaking it already
+	// should have one
```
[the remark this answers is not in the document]
nicktobey: This is _old_. Databases using the old file format may not have had a working set for each branch. Given that we no longer support the old format, I'm going to just remove this.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -224,6 +224,40 @@ assert_has_key_value() {
...
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "starting remote mode" ]] || false
+
+    run dolt --verbose-engine-setup --user dolt --password "" branch
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "starting remote mode" ]] || false
+    [[ "$output" =~ "main" ]] || false
+    [[ "$output" =~ "other" ]] || false
+
+    # Due to a current limitation, subsequent commands won't use the new branch until the server is stopped
```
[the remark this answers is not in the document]
nicktobey: When we started the migration you suggested that issues should be for customer facing features, and that we should track migration issues separately.

But I created https://github.com/dolthub/dolt/issues/6315 and linked it in the comments.

go/cmd/dolt/commands/checkout.go
```diff
@@ -82,276 +87,137 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
-	var remoteBranchName string
-	var startPt = "head"
+	rows, err := GetRowsForSql(queryEngine, sqlCtx, sqlQuery)
 
-	if apr.NArg() == 1 {
-		startPt = apr.Arg(0)
+	if err != nil {
+		// In fringe cases the server can't start because the default branch doesn't exist, `dolt checkout <existing branch>`
+		// offers an escape hatch.
+		if dEnv != nil && !branchOrTrack && strings.Contains(err.Error(), "cannot resolve default branch head for database") {
```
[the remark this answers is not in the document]
nicktobey: This error occurs even when the queryist is a SQLEngine. It's a necessary escape hatch for the case where the checked-out branch is deleted.

I removed the dEnv nil check though, since as you pointed out it's always set.

integration-tests/bats/sql-local-remote.bats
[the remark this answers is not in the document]
macneale4: Perfect. Thanks!

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -0,0 +1,191 @@
...
+		err = actions.CleanOldWorkingSet(ctx, dbData, db, dSess.Username(), dSess.Email(), initialRoots, headRef, initialWs)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func transferWorkingChanges(
```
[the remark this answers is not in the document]
nicktobey: Added a gdoc warning.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -0,0 +1,191 @@
...
+	branchRef ref.BranchRef,
+	force bool,
+) error {
+	dSess := dsess.DSessFromSess(ctx.Session)
+	newRoots, err := actions.RootsForBranch(ctx, initialRoots, branchHead, force)
+	if err != nil {
+		return err
+	}
+
+	// important to not update the checked out branch until after we have done the error checking above, otherwise we
```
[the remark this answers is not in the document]
nicktobey: Done.

go/cmd/dolt/commands/checkout.go
```diff
@@ -82,276 +87,137 @@ func (cmd CheckoutCmd) Exec(ctx context.Context, commandStr string, args []strin
...
-	if setTrackUpstream {
-		verr = SetRemoteUpstreamForBranchRef(dEnv, remoteName, remoteBranchName, headRef)
-		if verr != nil {
-			return verr
-		}
-	} else if autoSetupMerge, err := dEnv.Config.GetString("branch.autosetupmerge"); err != nil || autoSetupMerge != "false" {
-		// do guess remote branch if branch.autosetupmerge is not 'false', or if it is not set, it should default to 'true'.
-		// if no remote, it should not return an error
-		remotes, err := dEnv.RepoStateReader().GetRemotes()
+	if dEnv != nil && strings.Contains(message, "Switched to branch") {
```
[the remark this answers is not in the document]
nicktobey: Removed the check. I think I was assuming dEnv would be nil when connecting to an off-box host, but we don't support that yet and may not for a while.

macneale4 (approved):
Ship when the tests pass!
```
 .  o ..
 o . o o.o
      ...oo
        __[]__
     __|_o_o_o\__
     \""""""""""/
      \. ..  . /
 ^^^^^^^^^^^^^^^^^^^^
```

PR #6212, opened 2023-06-22
Add `dolt_schema_diff` table function

PavelSafronov:
`dolt_schema_diff` will return the schema diffs between refs and optionally will filter those changes to a specific table.

This is a new table function that provides us with the information we need for dolt diff.

The [original PR](https://github.com/dolthub/dolt/pull/6185) for this change was accidentally merged too early. The comments from that PR have been integrated into this change.


macneale4 (approved):
The additional engine tests are great! I'm pretty sure there are more edge cases we could close, but I think it's OK to ship this with what you have.

PR #6214, opened 2023-06-22
updates dolt reset to use sql queries

stephkyou:
This change updates `dolt reset` to use the appropriate sql engine to generate results.

### DIFF
--- a/go/cmd/dolt/commands/reset.go
+++ b/go/cmd/dolt/commands/reset.go
@@ -106,10 +106,10 @@ func (cmd ResetCmd) Exec(ctx context.Context, commandStr string, args []string, 
 	if apr.ContainsAll(HardResetParam, SoftResetParam) {
 		verr := errhand.BuildDError("error: --%s and --%s are mutually exclusive options.", HardResetParam, SoftResetParam).Build()
 		return HandleVErrAndExitCode(verr, usage)
-	} else if apr.Contains(HardResetParam) {
-		if apr.NArg() > 1 {
-			return handleResetError(fmt.Errorf("--hard supports at most one additional param"), usage)
-		}
+	} else if apr.Contains(HardResetParam) && apr.NArg() > 1 {
+		return handleResetError(fmt.Errorf("--hard supports at most one additional param"), usage)
+	} else if apr.Contains(SoftResetParam) && apr.NArg() > 1 {
+		return handleResetError(fmt.Errorf("--soft supports at most one additional param"), usage)
 	}
 
 	// process query through prepared statement to prevent sql injection
@@ -159,20 +159,13 @@ func constructInterpolatedDoltResetQuery(apr *argparser.ArgParseResults) (string
 		}
 	} else if apr.Contains(SoftResetParam) {
 		writeToBuffer("--soft")
-		if apr.NArg() > 0 {
-			for _, input := range apr.Args {
-				param = true
-				writeToBuffer("?")
-				params = append(params, input)
-			}
+		if apr.NArg() == 1 {
+			param = true
+			writeToBuffer("?")
+			params = append(params, apr.Arg(0))
 		}
 	} else {
 		for _, input := range apr.Args {
-			if strings.ToLower(input) == "head" && apr.NArg() == 1 {
-				buffer.Reset()
-				buffer.WriteString("CALL DOLT_RESET(")
-				break
-			}
 			param = true
 			writeToBuffer("?")
 			params = append(params, input)
--- a/go/cmd/dolt/commands/reset.go
+++ b/go/cmd/dolt/commands/reset.go
@@ -15,6 +15,7 @@
 package commands
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strings"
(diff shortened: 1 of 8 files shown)


integration-tests/bats/sql-local-remote.bats
```diff
@@ -283,4 +283,21 @@ get_staged_tables() {
     [[ "$output" =~ "starting local mode" ]] || false
     [[ "$output" =~ "main" ]] || false
     [[ "$output" =~ "b2" ]] || false
-}
\ No newline at end of file
+}
+
+@test "sql-local-remote: verify simple dolt reset behavior" {
```
macneale4, line 288: This test doesn't fit the pattern of other tests in this file. 1) You need to stop the server, and verify that the steps you run in both contexts are identical. 2) don't bother passing in `--verbose-engine-setup` if you aren't going to check the output. You should only  pass this to calls of `dolt reset` - since that is the command under observation.

go/cmd/dolt/commands/reset.go
```diff
@@ -80,90 +83,109 @@ func (cmd ResetCmd) ArgParser() *argparser.ArgParser {
...
-
-		return handleResetSoftTables(ctx, apr.Args, roots, dEnv, usage)
 	}
-}
 
-func handleResetSoftTables(ctx context.Context, tables []string, roots doltdb.Roots, dEnv *env.DoltEnv, usage cli.UsagePrinter) int {
-	roots, err := actions.ResetSoft(ctx, dEnv.DbData(), tables, roots)
+	// process query through prepared statement to prevent sql injection
+	query, params := constructParametrizedDoltResetQuery(apr)
+	interpolatedQuery, err := dbr.InterpolateForDialect(query, params, dialect.MySQL)
```
macneale4, line 117: I suggest interpolating the string in the construct method. Easier than getting to values back only to pass them into another function on the next line.

macneale4 (changes requested):
Small changes requested. Looks good - curious to see how bats tests go!

integration-tests/bats/sql-local-remote.bats
```diff
@@ -438,4 +438,20 @@ get_staged_tables() {
     run dolt sql -q "show tables"
     [ "$status" -eq 1 ]
     [[ "$output" =~ "When a password is provided, a user must also be provided" ]] || false
+
+@test "sql-local-remote: verify simple dolt reset behavior" {
+    start_sql_server altDB
+    dolt --verbose-engine-setup --user dolt sql -q "create table test1 (pk int primary key)"
```
macneale4: I suspect these won't work after my authentication change that you just rebased onto. You should actually just drop the `--user` flag from all inputs to new tests. It doesn't add anything to the tests.

integration-tests/bats/helper/local-remote.bash
```diff
@@ -31,7 +30,6 @@ SKIP_SERVER_TESTS=$(cat <<-EOM
 ~remotes-file-system.bats~
 ~sql-charsets-collations.bats~
 ~sql-cherry-pick.bats~
-~sql-commit.bats~
```
macneale4, line 34: Nice to see this!

go/cmd/dolt/commands/reset.go
```diff
@@ -80,90 +83,111 @@ func (cmd ResetCmd) ArgParser() *argparser.ArgParser {
...
-	}
-	ws, err := dEnv.WorkingSet(ctx)
-	if err != nil {
-		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+	var buffer bytes.Buffer
+	var first bool
+	first = true
+	buffer.WriteString("CALL DOLT_RESET(")
+
+	writeToBuffer := func(s string) {
```
macneale4, line 138: This pattern is becoming familiar. Are there some functions we can move into common shared functions? I'm not going to block on a refactor for this, but there may be a way to simplify this and reuse code

go/cmd/dolt/commands/reset.go
```diff
@@ -80,90 +83,111 @@ func (cmd ResetCmd) ArgParser() *argparser.ArgParser {
...
+	if apr.Contains(HardResetParam) {
+		writeToBuffer("--hard")
+		if apr.NArg() == 1 {
+			param = true
+			writeToBuffer("?")
+			params = append(params, apr.Arg(0))
+		}
+	} else if apr.Contains(SoftResetParam) {
+		writeToBuffer("--soft")
+		if apr.NArg() > 0 {
```
macneale4, line 162: According to the docs, soft takes no more than one arg. Take the loop out of this.

go/cmd/dolt/commands/reset.go
```diff
@@ -80,90 +83,111 @@ func (cmd ResetCmd) ArgParser() *argparser.ArgParser {
...
+		for _, input := range apr.Args {
+			if strings.ToLower(input) == "head" && apr.NArg() == 1 {
+				buffer.Reset()
+				buffer.WriteString("CALL DOLT_RESET(")
+				break
+			}
+			param = true
+			writeToBuffer("?")
+			params = append(params, input)
+		}
```
macneale4, line 179: The logic of the block above is a little awkward. The branching logic should start with first is if --hard or --soft is used. Those two cases are handled the same way - they are exclusive and have an optional arg. The alternate first branch would be to pull any number of tables. It more closely follows how we describe the behavior in the docs too.

macneale4 (changes requested):
Love the added testing. My feedback is mainly about the logic when building the query. A little bit of a refactor on the logic ordering would help it

macneale4 (approved):
LGTM. The remote bats tests are failing on what look like `reset` specific issues.

PR #6215, opened 2023-06-23
fix error return code in commit

stephkyou:
fixes `commit` to return an error when there is an error processing result rows


macneale4 (approved):
LGTM! Testing this is hard (impossible?), so fine to ship.

PR #6221, opened 2023-06-23
Migrate `dolt diff` and `dolt show` to SQL implementation

PavelSafronov:
This change migrates the current `dolt diff` implementation to use only SQL.
This is done by offloading most of the diffing logic to `dolt_diff` and `dolt_schema_diff` table functions, and making some of the common code generic, so we don't have to rely on DoltDb objects (which rely on dolt internals).

This change also partially migrates `dolt show` to SQL.
`dolt show` has the ability to log some of the storage internals in two cases:
1. `--no-pretty` flag is specified
2. `dolt show` is invoked against non-commit hashes

When either of the above cases occurs, we use DoltEnv to get these structures and log them. In all other cases, we use SQL to get the list of changes.


go/cmd/dolt/commands/show.go
```diff
@@ -186,181 +321,255 @@ func parseHashString(hashStr string) (hash.Hash, error) {
...
+		diffDisplaySettings: opts.diffDisplaySettings,
+		diffDatasets:        datasets,
+		tableSet:            tableSet,
 	}
-	return nil
+
+	return diffUserTables(queryist, sqlCtx, dArgs)
 }
 
-func showCommit(ctx context.Context, dEnv *env.DoltEnv, opts *showOpts, comm *doltdb.Commit) error {
```
macneale4, line 272: can you s/show/print/ for all of the show* commands. Hoping to get more consistent with stuff like this over time

go/cmd/dolt/commands/show.go
```diff
@@ -186,181 +321,255 @@ func parseHashString(hashStr string) (hash.Hash, error) {
...
+	if err != nil {
+		return "", err
+	}
+	if len(rows) == 0 {
+		return "", fmt.Errorf("no commits found for ref %s", ref)
+	}
+	return rows[0][0].(string), nil
+}
+
+func getTimestampColAsUint64(col interface{}) (uint64, error) {
```
macneale4, line 564: Is this going to be useful to other cli commands? Perhaps move to utils.go? Up to you, but I'd like to make sure we don't forget this is here.

go/cmd/dolt/commands/show.go
```diff
@@ -111,16 +125,109 @@ func (cmd ShowCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		}
+		if closeFunc != nil {
+			defer closeFunc()
+		}
+		err = showObjectsPretty(queryist, sqlCtx, opts)
+		return handleErrAndExit(err)
+	}
+}
+
+func resolveNonCommitSpec(ctx context.Context, dEnv *env.DoltEnv, specRef string) (isNonCommitSpec bool, resolvedValue string, err error) {
```
macneale4, line 180: Is a Spec a reference? or a refspec? Add a godoc on this method, then I may understand what it is that it's doing :-) I would think that any non-commit hash id would never be referenced, but I'm scratching my head at this method so I think there is something I'm not grasping.

go/cmd/dolt/commands/show.go
```diff
@@ -111,16 +125,109 @@ func (cmd ShowCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+
+	if !allSpecRefsAreCommits && !allSpecRefsAreNonCommits {
+		err = fmt.Errorf("cannot mix commit hashes and non-commit spec refs")
+		return handleErrAndExit(err)
+	}
+
+	useDoltEnv := !opts.pretty || (len(opts.specRefs) > 0 && allSpecRefsAreNonCommits)
+
+	if useDoltEnv {
+		if !opts.pretty && !dEnv.DoltDB.Format().UsesFlatbuffers() {
```
macneale4, line 160: I think it would be a good idea to at least determine if a server is running, then print a warning or possible abort if so. We don't want our users to be surprised by the fact that this is not a supported code path for remote calls.

go/cmd/dolt/commands/diff_output.go
```diff
@@ -195,78 +170,37 @@ func (t tabularDiffWriter) Close(ctx context.Context) error {
...
-		_, _ = bold.Printf("--- a/%s @ %s\n", td.FromName, h1.String())
-
-		h2, err := td.ToTable.HashOf()
-
-		if err != nil {
-			panic(err)
-		}
-
-		_, _ = bold.Printf("+++ b/%s @ %s\n", td.ToName, h2.String())
+		_, _ = bold.Printf("diff --dolt a/%s b/%s\n", fromTableName, toTableName)
```
macneale4, line 182: is `diff --dolt` actually a thing?! I see we print that already, I'm just not sure what the purpose of --dolt is in diff output.

go/cmd/dolt/commands/diff.go
```diff
@@ -886,27 +1225,28 @@ func diffRows(
...
 
 		// reset the row iterator
-		err = rowIter.Close(ctx)
+		err = rowIter.Close(sqlCtx)
 		if err != nil {
 			return errhand.BuildDError("Error closing row iterator:\n%s", query).AddCause(err).Build()
 		}
-		_, rowIter, err = sqlEng.Query(ctx, query)
-		defer rowIter.Close(ctx)
+		//_, rowIter, err = sqlEng.Query(ctx, query)
```
macneale4, line 1239: delete dead line. Forget the past! On to the future!

go/cmd/dolt/commands/diff.go
```diff
@@ -779,26 +1070,74 @@ func diffDoltSchemasTable(
 	return nil
 }
 
+// ArePrimaryKeySetsDiffable checks if two schemas are diffable. Assumes the
```
macneale4, line 1073: lowercase "A" to match function name

I also don't understand what `if __DOLT__` means. Are you talking about the storage format?

go/cmd/dolt/commands/diff.go
```diff
@@ -709,7 +999,8 @@ func diffDoltSchemasTable(
 		"order by coalesce(from_type, to_type), coalesce(from_name, to_name)",
 		dArgs.fromRef, dArgs.toRef, doltdb.SchemasTableName)
 
-	_, rowIter, err := sqlEng.Query(sqlCtx, query)
+	//_, rowIter, err := sqlEng.Query(sqlCtx, query)
```
macneale4, line 1002: Delete

macneale4 (changes requested):

integration-tests/bats/diff.bats
```diff
@@ -668,18 +668,18 @@ EOF
...
+@test "diff: new foreign key added and resolves" {
     dolt sql <<SQL
 create table parent (i int primary key);
 create table child (j int primary key, foreign key (j) references parent (i));
 SQL
 
     run dolt diff
     [ "$status" -eq 0 ]
-    [[ "$output" =~ 'resolved foreign key' ]] || false
+    [[ "$output" =~ "+  CONSTRAINT \`npsbmr30\` FOREIGN KEY (\`j\`) REFERENCES \`parent\` (\`i\`)" ]] || false
```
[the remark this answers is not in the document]
PavelSafronov: This was brought up and discussed in #sql channel: we no longer need to specify that a foreign key was resolved (and we overall don't want to expose that keys are resolved or not), so the "resolved foreign key" check was dropped from the BATS tests.

go/cmd/dolt/commands/diff.go
```diff
@@ -268,75 +288,96 @@ func parseDiffArgs(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
...
-func (dArgs *diffArgs) applyDiffRoots(ctx context.Context, dEnv *env.DoltEnv, args []string, isCached, useMergeBase bool) ([]string, error) {
-	headRoot, err := dEnv.HeadRoot(ctx)
+var doltSystemTables = []string{
+	"dolt_procedures",
+	"dolt_schemas",
+}
+
+func getTableNamesAtRef(queryist cli.Queryist, sqlCtx *sql.Context, ref string) (map[string]bool, error) {
+	// query for user-created tables
+	q := fmt.Sprintf("SHOW FULL TABLES AS OF '%s'", ref)
```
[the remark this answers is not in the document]
PavelSafronov: Is InterpolateForDialect the best approach to avoiding injection attacks?

InterpolateForDialect seems good on the surface, but there is absolutely no documentation around this method and it has some unexpected behavior, at least with "SHOW CREATE TABLE..." queries. ( I've opened a bug against the repo for this: https://github.com/gocraft/dbr/issues/254)

go/cmd/dolt/commands/diff.go
```diff
@@ -498,100 +530,204 @@ var diffSummarySchema = sql.Schema{
...
 	}
 
-	toRootHash, err := dArgs.diffDatasets.toRoot.HashOf()
+	return summaries, nil
+}
+
+func diffUserTables(queryist cli.Queryist, sqlCtx *sql.Context, dArgs *diffArgs) errhand.VerboseError {
+	var err error
+
+	//diffSummaries, err := getDiffSummariesBetweenRefs(queryist, sqlCtx, dArgs.fromRef, dArgs.toRef)
```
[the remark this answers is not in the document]
PavelSafronov: Removing.

integration-tests/bats/rename-tables.bats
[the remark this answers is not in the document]
PavelSafronov: Adding sql-local-remote.bats changes and removing the multiple `*diff.bats` lines from local-remote.bash

go/cmd/dolt/commands/diff.go
PavelSafronov: Removing.

go/cmd/dolt/commands/diff.go
PavelSafronov: Removing.

go/cmd/dolt/commands/diff_output.go
PavelSafronov: I guess it's to "match" similar `git diff` output:

```
$ git diff
diff --git a/go/cmd/dolt/commands/show.go b/go/cmd/dolt/commands/show.go
index 68057a357..360659a42 100644
--- a/go/cmd/dolt/commands/show.go
+++ b/go/cmd/dolt/commands/show.go
@@ -17,6 +17,8 @@ package commands
 import (
        "context"
        "fmt"
+       "github.com/gocraft/dbr/v2"
+       "github.com/gocraft/dbr/v2/dialect"
        "regexp"
        "strings"
        "time"
```

go/cmd/dolt/commands/show.go
PavelSafronov: Will standardize the function names.

go/cmd/dolt/commands/show.go
PavelSafronov: Moving to utils.go

go/cmd/dolt/commands/show.go
PavelSafronov: Adding comments to the code.

go/cmd/dolt/commands/show.go
PavelSafronov: Testing for this now.

go/cmd/dolt/commands/diff.go
PavelSafronov: Updating the casing.
The comment was taken from a different version of this function and is no longer relevant. Removing comment.

go/cmd/dolt/commands/show.go
PavelSafronov: Adding BATS tests for this behavior.

go/cmd/dolt/commands/diff.go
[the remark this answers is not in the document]
macneale4: It something we could fork and extend. We are being kind of odd in the way we use it, so let's see how it goes.

integration-tests/bats/rename-tables.bats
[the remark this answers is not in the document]
macneale4: Cool... and the tests pass? I guess we'll find out in CI!

macneale4 (approved):
Looking good. Get the tests passing and ship it!

integration-tests/bats/sql-local-remote.bats
```diff
@@ -328,8 +328,10 @@ assert_has_key_value() {
 
 @test "sql-local-remote: verify dolt diff behavior with data and schema changes" {
   start_sql_server defaultDB
-
   cd defaultDB
+
+  export DOLT_CLI_PASSWORD=""
```
macneale4, line 333: We unset the DOLT_CLI_PASSWORD var in this file, so I'd prefer you just set the `--password` everywhere, OR you can specify no user/pwd (which should work now that my auth changes are in).

integration-tests/bats/sql-local-remote.bats
```diff
@@ -384,6 +386,8 @@ EOF
 @test "sql-local-remote: verify dolt show behavior" {
   cd defaultDB
 
+  export DOLT_CLI_PASSWORD=""
```
macneale4, line 389: See previous comment

macneale4 (approved):
Small change requested for how you set the password in the tests. fix and ship!

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Cool, dropping `--user` arguments in these tests.

integration-tests/bats/sql-local-remote.bats
PavelSafronov: See previous response :)

PR #6223, opened 2023-06-24
Migrate `dolt conflicts cat` to SQL

PavelSafronov:
Migrate `dolt conflicts cat` to use SQL.

Currently only `dolt conflicts cat` is migrated, `dolt conflicts resolve` is not yet migrated.


integration-tests/bats/conflict-cat.bats
macneale4, line 1: Looks like you'll be able to enable this file in local-remotes.bash just as soon as checkout is done. Fingers crossed that lands soon

go/cmd/dolt/commands/utils.go
```diff
@@ -292,9 +292,25 @@ func getRowsForSql(queryist cli.Queryist, sqlCtx *sql.Context, query string) ([]
 	return rows, nil
 }
 
+// getTinyIntColAsBool returns the value of a tinyint column as a bool
+// This is necessary because Queryist may return a tinyint column as a bool (when using SQLEngine)
+// or as a string (when using ConnectionQueryist).
+func GetTinyIntColAsBool(col interface{}) (bool, error) {
```
macneale4, line 298: thanks for moving to utils.go!

go/cmd/dolt/commands/cnfcmds/conflictsplitter.go
```diff
@@ -73,6 +73,11 @@ func newConflictSplitter(conflictQuerySch sql.Schema, targetSch sql.Schema) (*co
 			}
 		}
 
+		if colName == "" {
```
macneale4, line 76: Under what circumstances does this happen?

macneale4 (commented):
Let's talk with nick and see if we can get checkout done before this is shipped. I'd like to start enabling more tests to build confidence

go/cmd/dolt/commands/cnfcmds/conflictsplitter.go
PavelSafronov: Right now, the only columns that fits this description are `from_root_ish` and `dolt_conflict_id`.

This code was originally written with the expectation that the fields are returned in a very specific order, but this would have required more work on our part and more queries (we would need to figure out the "base", "their", and "our" schemas, then use that info to craft the `select base_pk, our_pk, their_pk... from dolt_conflicts_FOO` query). Instead of doing that, I chose to slightly modify the code that parses this data, so it could accept whatever order of fields Dolt gives us. Getting all the fields in random order surprisingly *almost* worked (thanks Dhruv for writing such good code!), except we would have instances like these where we'd have to filter out the column.

macneale4: Got it. Resolving.

integration-tests/bats/conflict-cat.bats
macneale4: NM. Checkout it not done yet.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -431,6 +431,62 @@ EOF
   stop_sql_server 1
 }
 
+@test "sql-local-remote: verify dolt conflicts cat behavior" {
+  cd defaultDB
+
+  export DOLT_CLI_PASSWORD=""
```
macneale4, line 437: Drop this in this file, and remove all the `--user dolt` options. This can be a fast follow change

PavelSafronov: Good catch, updating.

PavelSafronov: Removed.

macneale4 (approved):
LGTM!

PR #6225, opened 2023-06-26
chore: remove refs to deprecated io/ioutil

testwill:


timsehn:
Thanks for the contribution. We will look today!

go/libraries/doltcore/table/typed/parquet/writer_test.go
```diff
@@ -83,12 +82,7 @@ John Johnson,21,
 Andy Anderson,27,
 `
 
-	file, err := ioutil.TempFile("", "parquet")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.Remove(file.Name())
```
zachmu, line 90: Looks like you dropped this defer block

zachmu (commented):
One line of dropped logic, otherwise this looks good

go/libraries/doltcore/table/typed/parquet/writer_test.go
zachmu: Actually, seems like the testing framework will remove this for us on test cleanup, so no need for this afterall, nevermind

zachmu (approved):

zachmu:
There's a go test failure on windows related to this change which may indicate that we're not cleaning up file resources appropriately:

```
Testing Package: github.com/dolthub/dolt/go/libraries/doltcore/table/typed/parquet
--- FAIL: TestWriter (1.54s)
    testing.go:1206: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestWriter4[294](https://github.com/dolthub/dolt/actions/runs/5374694074/jobs/9764597484?pr=6225#step:4:295)363021\001\parquet: The process cannot access the file because it is being used by another process.
```

There are retries in place as of go 1.19 or later, so this is potentially an actual problem.

https://github.com/golang/go/issues/51442

I can dig into this later today, but it needs more work for windows.

testwill:
Do I need to change t.TempDir() to os.MkdirTemp?

zachmu:
Closing this in favor of this:

https://github.com/dolthub/dolt/pull/6247

zachmu:
By the way this was the bug:

https://github.com/xitongsys/parquet-go/issues/543

PR #6229, opened 2023-06-26
Support for dolt_ procedures inside user stored procedures

zachmu:
These are just tests, the actual necessary changes are in https://github.com/dolthub/go-mysql-server/pull/1852

This addresses a large part of https://github.com/dolthub/dolt/issues/5829, but see the last comment there: because of how we analyze / execute stored procedures, things `dolt_checkout()` doesn't work in stored procedures in some cases (namely when using unqualified table names).


fulghum (approved):
Tests look solid. 

A case statement might be another good one to test and see if there's another spot that needs a tx init. 

Seems like triggers would have this same issue, and the fix should work for them, too, but might be nice to have a test case for it.

PR #6240, opened 2023-06-27
Add global arguments section to dump-docs command

tbantle22:


go/cmd/dolt/commands/dump_docs.go
```diff
@@ -34,6 +34,33 @@ const (
 		"title: CLI\n" +
 		"---\n\n" +
 		"# CLI\n\n"
+
+	globalArgumentsSection = `
+## Global Arguments
+
+Dolt subcommands are in transition to using the flags listed below as global flags.
+The sql subcommand is currently the only command that uses these flags. All other commands will ignore them.
```
zachmu, line 42: This is out of date, yeah?

zachmu: Can we define a var for where this is used in the main command and reference that here?

go/cmd/dolt/cli/command.go
```diff
@@ -309,6 +309,22 @@ func CheckUserNameAndEmail(config *env.DoltCliConfig) bool {
...
+			}
+		}
+
+		str += fmt.Sprintf("    %16s - %s\n", cmd.Name(), cmd.Description())
+	}
+
+	return str
+}
+
 func (hc SubCommandHandler) PrintUsage(commandStr string) {
```
fulghum, line 328: (minor) Looks like this function could use the new new `GetUsage` to create the string before it prints it.

zachmu (approved):
LGTM, some suggestions

go/cmd/dolt/commands/dump_docs.go
fulghum, line 42: @macneale4 – any other updates in here?

fulghum (approved):
Looks good! Neil may have some more updates for the global args section based on the migration progress, but those can be updated later. 

Thanks for fixing this! 🙏

zachmu (approved):
LGTM

go/cmd/dolt/commands/dump_docs.go
tbantle22: Talked to neil and updated this copy

PR #6241, opened 2023-06-27
Migrate `dolt conflicts resolve` to use only SQL

PavelSafronov:
This change migrates `dolt conflicts resolve` to only use SQL.

### DIFF
--- a/integration-tests/bats/conflicts-resolve.bats
+++ b/integration-tests/bats/conflicts-resolve.bats
@@ -1,0 +1,233 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
+
+setup() {
+    setup_common
+}
+
+basic_conflict() {
+    dolt sql -q "create table t (i int primary key, t text)"
+    dolt add .
+    dolt commit -am "init commit"
+    dolt checkout -b other
+    dolt sql -q "insert into t values (1,'other')"
+    dolt commit -am "other commit"
+    dolt checkout main
+    dolt sql -q "insert into t values (1,'main')"
+    dolt commit -am "main commit"
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+@test "conflicts-resolve: call with no arguments, errors" {
+    run dolt conflicts resolve
+    [ $status -eq 1 ]
+    [[ $output =~ "--ours or --theirs must be supplied" ]] || false
+}
+
+@test "conflicts-resolve: call without specifying table, errors" {
+    run dolt conflicts resolve --theirs
+    [ $status -eq 1 ]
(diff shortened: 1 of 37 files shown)


integration-tests/bats/sql-local-remote.bats
```diff
@@ -439,3 +451,48 @@ get_staged_tables() {
     [ "$status" -eq 1 ]
     [[ "$output" =~ "When a password is provided, a user must also be provided" ]] || false
 }
+
+@test "sql-local-remote: verify dolt conflicts resolve behavior" {
+  skip "This test relies on dolt commit being migrated"
```
macneale4, line 456: Commit has already been migrated. You mean checkout?

More to the point, let's not skip tests in this file for this project. In the worst case situation, you can do `dolt sql -q "call dolt_checkout()"`. There should always be some way to do what you want without using the CLI commands

integration-tests/bats/sql-local-remote.bats
```diff
@@ -439,3 +451,48 @@ get_staged_tables() {
...
+  start_sql_server defaultDB
+  cd defaultDB
+
+  basic_conflict
+  dolt --user dolt checkout main
+  run dolt --user dolt sql -q "select * from t"
+  [ $status -eq 0 ]
+  [[ $output =~ "main" ]] || false
+
+  run dolt --user dolt merge other
```
macneale4, line 467: You are going to need to drop all of the `--user` options now that authentication change is in.

integration-tests/bats/conflicts-resolve.bats
macneale4, line 1: I'm really glad you are writing these tests. Have you verified that they pass even if you run them on a version of dolt which doesn't have any of the sql changes?

`git rebase` can be really helpful for things like this. Reorder your history so that the addition of this file is all that your branch contains. build dolt, and verify that these tests work. Then you can do the migration with more confidence.

macneale4 (commented):
Verify that the new conflicts-resolve.bats tests work on an existing version of the server, then I'll be happy to sign off on this

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Removing.

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Removing `skip` directive and rewriting in such a way that we don't need to use a migrated version of `dolt checkout`.

integration-tests/bats/conflicts-resolve.bats
PavelSafronov: Just ran these tests against tip of main and they pass.

macneale4 (approved):
Ship It!

PR #6242, opened 2023-06-27
adds print statements for errors in commit

stephkyou:
adds missing print statements for errors occurring in `dolt commit`


macneale4 (changes requested):
Let's get it back to the point where it asks you to introduce yourself to dolt.

macneale4 (approved):

PR #6244, opened 2023-06-27
dolt clone: Fix dolt clone run against a sql-server where the database has been GCd.

reltuk:
A long-standing bug in the remotesapi which the sql-server exposes could cause a `dolt clone` to fail when running against a database which had been garbage collected. This change fixes the bug in the server. It also patches the client behavior so that it will tolerate responses from older versions of dolt.


zachmu (approved):
LGTM!

PR #6252, opened 2023-06-28
Set the DB name on the sql.Context when in remote mode

macneale4:
The SQL.Context used in the remote connection case is kind of an empty shell since the connection has all the details for working with the DB. One side effect of having a dummy sql.Context was not having the database name set in the context. The shell uses it to set it's prompt, but this could have ramifications elsewhere we haven't discovered.

https://github.com/dolthub/dolt/issues/6246


go/cmd/dolt/commands/sqlserver/sqlclient.go
```diff
@@ -498,7 +498,9 @@ func BuildConnectionStringQueryist(ctx context.Context, cwdFS filesys.Filesys, c
 	queryist := ConnectionQueryist{connection: conn}
 
 	var lateBind cli.LateBindQueryist = func(ctx context.Context) (cli.Queryist, *sql.Context, func(), error) {
-		return queryist, sql.NewContext(ctx), func() { conn.Conn(ctx) }, nil
+		sqlCtx := sql.NewContext(ctx)
+		sqlCtx.SetCurrentDatabase(database)
```
zachmu, line 502: I don't know if this is correct, the behavior of session initialization here is subtle and hard to get right. It looks to me like this gets called every time LateBindContext.QueryEngine gets called, which means that you're going to be setting the current DB to the initial one on every query, so if you `USE` another database it will get overwritten by this.

zachmu (commented):
Not sure if this is correct actually, have a look

go/cmd/dolt/commands/sqlserver/sqlclient.go
macneale4: This code is called once at the beginning of the command, and never again. We enforce this here:

https://github.com/dolthub/dolt/blob/1457b9b444bebca51a7672d31ae851140a92643a/go/cmd/dolt/cli/cli_context.go#L71

zachmu: I built this PR to play with it, and here's what I see when the server is running and I run `dolt sql`:

```sql
employees> show databases;
+--------------------+
| Database           |
+--------------------+
| employees          |
| information_schema |
| mysql              |
+--------------------+
3 rows in set (0.00 sec)

employees> use mysql;
Database changed
employees> select database(), active_branch();   -- notice the prompt doesn't change
Error 1105: database not found: mysql  -- another bug

employees> create database testdb;
employees> show databases;
+--------------------+
| Database           |
+--------------------+
| employees          |
| information_schema |
| mysql              |
| testdb             |
+--------------------+
4 rows in set (0.00 sec)

employees> use testdb;
Database changed
employees> select database(), active_branch();  -- prompt still not changed
+------------+---------------------+
| database() | active_branch()     |
+------------+---------------------+
| testdb     | globalInitialBranch |
+------------+---------------------+
1 row in set (0.00 sec)
```

zachmu: Also doesn't work with revision qualified db names:

```sql
employees> use employees/main;
Database changed
employees> select database(), active_branch();
+----------------+-----------------+
| database()     | active_branch() |
+----------------+-----------------+
| employees/main | main            |
+----------------+-----------------+
1 row in set (0.00 sec)
```

zachmu: `dolt sql` with no running server works as expected

```sql
% dolt sql
# Welcome to the DoltSQL shell.
# "exit" or "quit" (or Ctrl-D) to exit.
employees> use employees/main;
Database changed
employees/main> select database(), active_branch();
+----------------+-----------------+
| database()     | active_branch() |
+----------------+-----------------+
| employees/main | main            |
+----------------+-----------------+
1 row in set (0.00 sec)

employees/main>
```

nicktobey: @macneale4 is this fixed?

nicktobey (approved):

go/cmd/dolt/commands/sqlserver/sqlclient.go
macneale4: Adding the sql-shell.bats tests in remote mode demonstrated the bug. Latest change fixes it.

PR #6267, opened 2023-06-30
change commit to print with printCommitInfo

stephkyou:
Changes `commit` to print with `printCommitInfo`. This prevents seg fault errors from using `dolt log` as was previously done.

### DIFF
--- a/go/cmd/dolt/commands/commit.go
+++ b/go/cmd/dolt/commands/commit.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/dolthub/dolt/go/store/util/outputpager"
 	"github.com/dolthub/go-mysql-server/sql"
 	"github.com/fatih/color"
 	"github.com/gocraft/dbr/v2"
@@ -41,6 +40,7 @@ import (
 	"github.com/dolthub/dolt/go/libraries/utils/iohelp"
 	"github.com/dolthub/dolt/go/libraries/utils/set"
 	"github.com/dolthub/dolt/go/store/datas"
+	"github.com/dolthub/dolt/go/store/util/outputpager"
 )
 
 var commitDocs = cli.CommandDocumentationContent{
--- a/go/cmd/dolt/commands/commit.go
+++ b/go/cmd/dolt/commands/commit.go
@@ -171,12 +171,14 @@ func performCommit(ctx context.Context, commandStr string, args []string, cliCtx
 	}
 
 	commit, err := getCommitInfo(queryist, sqlCtx, "HEAD")
-	cli.ExecuteWithStdioRestored(func() {
-		pager := outputpager.Start()
-		defer pager.Stop()
+	if cli.ExecuteWithStdioRestored != nil {
+		cli.ExecuteWithStdioRestored(func() {
+			pager := outputpager.Start()
+			defer pager.Stop()
 
-		printCommitInfo(pager, 0, false, "auto", commit)
-	})
+			printCommitInfo(pager, 0, false, "auto", commit)
+		})
+	}
 
 	return 0, false
 }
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/commit.go
```diff
@@ -170,22 +170,14 @@ func performCommit(ctx context.Context, commandStr string, args []string, cliCtx
...
-			cli.Println(err.Error())
-			return 1, false
-		}
-		err = engine.PrettyPrintResults(sqlCtx, engine.FormatTabular, schema, rowIter)
-		if err != nil {
-			cli.Println(err.Error())
-			return 1, false
-		}
-	}
+	// TODO: when dolt log is migrated, remove this block printing out the commit and print with a dolt log call in Exec()
```
macneale4, line 173: Drop this comment. Don't see why we would ever use log in the future if this works.

macneale4 (changes requested):
Also, there was a skip added here which we need to remove:

https://github.com/dolthub/dolt/blob/c336e1063d044401326a4f063980063f817b053a/integration-tests/bats/sql-local-remote.bats#L279

integration-tests/bats/1pk5col-ints.bats
```diff
@@ -746,8 +746,8 @@ DELIM
     dolt checkout test-branch-m
     run dolt merge test-branch -m "merge"
     [ "$status" -eq 0 ]
-    [ "${lines[6]}" = "test | 1 +" ]
-    [ "${lines[7]}" = "1 tables changed, 1 rows added(+), 0 rows modified(*), 0 rows deleted(-)" ]
+    [ "${lines[5]}" = "test | 1 +" ]
```
PavelSafronov, line 749: What's the reason for this change? Do we need to add a `cli.Println()` somewhere for this test to pass?
Taylor recently fixed a similar issue with `status` (https://github.com/dolthub/dolt/pull/6243), wonder if this is another spot where we just need to log a newline.

PavelSafronov (approved):
Changes look good, thanks for tackling this so quickly.
Just have a minor question about a test.

integration-tests/bats/1pk5col-ints.bats
stephkyou: I think the output looks good, not missing any newlines so it seems like we're just formatting it slightly differently.

PavelSafronov: Looks like we're not *quite* showing everything.

I ran the above BATS test against tip of `main` and this PR, and here is the difference in output:

`main`
```txt
   >>>>>>>
   Updating m6rh21injvf192p8lidj7ir004dao7ns..gc93v5k5eju38on6qqf3t38ktarf06e1
   commit iaqghu0g7s8gcc7b0n8p9plfk072hbjp (HEAD -> test-branch-m)
   Merge: m6rh21injvf192p8lidj7ir004dao7ns gc93v5k5eju38on6qqf3t38ktarf06e1
   Author: Bats Tests <bats@email.fake>
   Date:  Fri Jun 30 10:06:58 -0700 2023

        merge

   test | 1 +
   1 tables changed, 1 rows added(+), 0 rows modified(*), 0 rows deleted(-)
   <<<<<<<
```

this pr
```txt
   >>>>>>>
   Updating ohrrnpi8qlcuupg3gilh61pi6h00n2fp..fq2u0jc670pcgofjsmbpsqkbeao8uroe
   commit jd2e3uo0momqq1ua7ufg1tp1t6ogt57u (HEAD -> test-branch-m) 
   Author: Bats Tests <bats@email.fake>
   Date:  Fri Jun 30 10:07:15 -0700 2023
   
        merge
   
   test | 1 +
   1 tables changed, 1 rows added(+), 0 rows modified(*), 0 rows deleted(-)
   <<<<<<<
```

Looks like we're not showing the merge information. Think we need to fix that.

PavelSafronov (changes requested):
Looks like the new commit message is missing merge info.

PavelSafronov (approved):
New changes look good! Thanks for finding and fixing the `show.go` bug, and for removing test modifications.

PR #6268, opened 2023-06-30
chore: slice replace loop

testwill:


fulghum (approved):
Thank you @testwill! 🙏 Change looks great! 👍 

One of the CI tests flaked, so I restarted it, and will merge when it completes.

PR #6270, opened 2023-06-30
Remove alter_statement column from dolt_schema_diff results.

PavelSafronov:
Remove `alter_statement` column from `dolt_schema_diff` results.

This change is a result of feedback about the `dolt_schema_diff` output: https://github.com/dolthub/docs/pull/1575#discussion_r1247212714

We're removing the `alter_statement` column, and requiring the user to get this information from `dolt_patch`.
The migrated command `dolt diff` is now doing this: we identify the changed tables using `dolt_schema_diff`, then get the alter statements from `dolt_patch`.

This update also resolves https://github.com/dolthub/dolt/issues/6265 , which is a feature request to return empty `to_*`/`from_*` columns in cases when the table does not exist at the `to` or `from` revision.


fulghum (approved):
Looks great! 🚢

tbantle22 (approved):
lgtm!

PR #6271, opened 2023-06-30
Revert transaction-unsafe implementation of `SessionStateAdapter::SetCWBHeadRef`

nicktobey:
This recently added implementation of `SetCWBHeadRef` was a designed to fix an issue when evalutating `call dolt_branch('-m', $ACTIVE_BRANCH, $NEW_BRANCH)` (which would cause a panic) by changing the active branch on the SQL session to match the new branch. However, this implementation has some problems.

When operating on the DB from a SQL-context, we use transactions to ensure that changes are applied atomically. However currently, renaming branches doesn't use transactions because they touch multiple branches at once and we can't really model that as a transaction. This becomes a problem when mixed with `DoltSession` operations, which do use transactions.

My original implementation attempted to work around this by calling `commitTransaction` before and after setting the new current working branch. This would allow `SwitchActiveBranch` to see the newly created branch. However, this is not a correct use of manual transaction management and is not generally safe.

No matter the solution to the original bug, doing manual transaction management in `SetCWBHeadRef` is not a good idea. It's been reverted, and the call to `SetCWBHeadRef` within `RenameBranch` has been removed.

The last commit in this PR is an attempt to implement the branch switch within `RenameBranch` itself. This also uses manual transaction management, but this is isolated to a single execution path. I welcome feedback on whether there's a better way to do this.

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
@@ -144,7 +144,13 @@ func renameBranch(ctx *sql.Context, dbData env.DbData, apr *argparser.ArgParseRe
 		return err
 	}
 
-	err := actions.RenameBranch(ctx, dbData, oldBranchName, newBranchName, sess.Provider(), force, rsc)
+	headRef, err := dbData.Rsr.CWBHeadRef()
+	if err != nil {
+		return err
+	}
+	activeSessionBranch := headRef.GetPath()
+
+	err = actions.RenameBranch(ctx, dbData, oldBranchName, newBranchName, sess.Provider(), force, rsc)
 	if err != nil {
 		return err
 	}
@@ -161,6 +167,29 @@ func renameBranch(ctx *sql.Context, dbData env.DbData, apr *argparser.ArgParseRe
 				repoState.Head.Ref = ref.NewBranchRef(newBranchName)
 				repoState.Save(fs)
 			}
+		}
+	}
+
+	// If the active branch of the SQL session was renamed, switch to the new branch.
+	if oldBranchName == activeSessionBranch {
+		currentTx := ctx.GetTransaction()
+		if currentTx != nil {
+			err := sess.CommitTransaction(ctx, currentTx)
+			if err != nil {
+				return err
+			}
+			newTx, err := sess.StartTransaction(ctx, sql.ReadWrite)
+			ctx.SetTransaction(newTx)
+		}
+
+		wsRef, err := ref.WorkingSetRefForHead(ref.NewBranchRef(newBranchName))
+		if err != nil {
+			return err
+		}
+
+		err = sess.SwitchWorkingSet(ctx, dbName, wsRef)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
@@ -170,13 +170,13 @@ func renameBranch(ctx *sql.Context, dbData env.DbData, apr *argparser.ArgParseRe
 		}
 	}
 
+	err = sess.RenameBranchState(ctx, dbName, oldBranchName, newBranchName)
+	if err != nil {
+		return err
+	}
+
 	// If the active branch of the SQL session was renamed, switch to the new branch.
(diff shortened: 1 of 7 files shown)


go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -144,7 +144,13 @@ func renameBranch(ctx *sql.Context, dbData env.DbData, apr *argparser.ArgParseRe
 		return err
 	}
 
-	err := actions.RenameBranch(ctx, dbData, oldBranchName, newBranchName, sess.Provider(), force, rsc)
+	headRef, err := dbData.Rsr.CWBHeadRef()
```
zachmu, line 147: This is kind of subtle... in addition to the current head ref, you also need to make sure to check the current checked out branch, which is what is returned by `DoltSession.CurrentHead()`. `CWBHeadRef()` will be different from the checked out branch in some cases, eg. after a `USE mydb/branch` statement. Your goal is that after this procedure is called, there are no remaining references to the old name in the session state.

Check out `DoltSession.RemoveBranchState()`, which is used when we delete a branch. We need something similar to use for this case, something like `DoltSession.RenameBranchState()`. Instead of checking out the default branch like `RemoveBranchState`, the new method should rename any references as appropriate.

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
```diff
@@ -164,6 +170,29 @@ func renameBranch(ctx *sql.Context, dbData env.DbData, apr *argparser.ArgParseRe
 		}
 	}
 
+	// If the active branch of the SQL session was renamed, switch to the new branch.
+	if oldBranchName == activeSessionBranch {
```
zachmu, line 174: If you implement the method suggested above, then you probably don't need to do any transaction management here at all. The transaction will get committed when this procedure completes anyway

zachmu (commented):
What you have here is mostly correct but misses a weird edge case. I have a suggestion for how to deal with it

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
nicktobey: Ah, I didn't realize that `DoltSession.CurrentHead()` and `CWBHeadRef()` could diverge.

The potential issue here is that when I don't commit the transaction, `DoltSession.lookupDbState` fails when it calls `addDb`: it can't add the initial state to the session because it can't find the new branch. I imagine this is because the branch was added by directly manipulating the DB after the transaction has already been started, and the function uses the initial DB to fetch the branch state.

So I could do this, but unless I rewrite how this creates the new branch only use DoltSession (which is doable but would be a learning spike), I'd still need to manually commit the transaction first.

nicktobey: Oh wait, I see what you're suggesting.

nicktobey: How's this?

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -260,6 +260,37 @@ func (d *DoltSession) RemoveBranchState(ctx *sql.Context, dbName string, branchN
...
+	if !ok {
+		return sql.ErrDatabaseNotFound.New(baseName)
+	}
+
+	d.mu.Lock()
+	branch, ok := checkedOutState.dbState.heads[strings.ToLower(oldBranchName)]
+
+	if !ok {
+		// nothing to rename
+		return nil
```
zachmu, line 280: Missing an unlock here

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
zachmu: LGTM, thanks

go/libraries/doltcore/sqle/dprocedures/dolt_branch.go
zachmu: You actually want to do this in all cases, regardless of the active branch

zachmu (approved):
LGTM, just a couple final small issues

go/libraries/doltcore/sqle/dsess/session.go
nicktobey: Fixed.

PR #6276, opened 2023-07-05
Use correct local branch name when setting upstream in `dolt_checkout`

nicktobey:
I found an issue in `dolt_checkout` that occurs in the narrow case where both the `-b` and `--track` flags are provided, in order to create a new branch that tracks an upstream branch with a different name.

In that case, we fail to set the upstream correctly because we accidentally look for a local branch to mark using the remote branch name, instead of the local branch name.

I added a regression test.

### DIFF
--- a/integration-tests/bats/remotes.bats
+++ b/integration-tests/bats/remotes.bats
@@ -2075,6 +2075,51 @@ SQL
     [[ "$output" =~ "Everything up-to-date." ]] || false
 }
 
+@test "remotes: dolt sql -q 'call dolt_checkout("-b", "newbranch", "--track", "origin/feature")'' checks out new local branch 'newbranch' with upstream set" {
+    mkdir remote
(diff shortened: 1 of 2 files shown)


integration-tests/bats/remotes.bats
```diff
@@ -2075,6 +2075,51 @@ SQL
...
+
+    cd ..
+    dolt clone file://./remote repo2
+
+    cd repo2
+    run dolt branch
+    [[ ! "$output" =~ "feature" ]] || false
+
+    dolt sql -q 'call dolt_checkout("-b", "newbranch", "--track", "origin/feature");'
+    dolt sql -q 'select * from dolt_branches;'
```
stephkyou, line 2097: extra line here

nicktobey: Good catch! Fixed.

PR #6277, opened 2023-07-05
adds test for reset handling ignored tables

stephkyou:
Adds tests for `dolt reset` and `dolt_reset()` to check their handling of ignored tables.

### DIFF
--- a/integration-tests/bats/reset.bats
+++ b/integration-tests/bats/reset.bats
@@ -316,8 +316,8 @@ SQL
 CREATE TABLE test2 (
     pk int primary key
 );
-INSERT INTO test2 VALUES (1);
-INSERT INTO test1 VALUES (1, 1, 1);
+INSERT INTO test2 VALUES (9);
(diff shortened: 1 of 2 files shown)


integration-tests/bats/reset.bats
```diff
@@ -308,3 +308,31 @@ SQL
...
+    run dolt reset --hard
+    [ "$status" -eq 0 ]
+
+    run dolt status
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "nothing to commit, working tree clean" ]]
+
+    run dolt sql -q "select * from test2"
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "1" ]] || false
```
macneale4, line 337: Good practice is to use a variety of values so you aren't getting false positives. You may want to insert 9 into test2 and 1,2,3 into test1.

macneale4 (approved):
Minor changes then ship it

PR #6281, opened 2023-07-05
return commit hash for `dolt_merge()` when `--no-ff` is not specified

JCOR11599:
Fix a bug where we didn't return the commit hash for `dolt_merge()` when the --no-ff argument wasn't specified, and we performed a non-ff merge.

Made it so that fast-forward merged will also return the new HEAD commit hash.

Additionally, changed many enginetests to use the new `DoltCommitType` to validate dolt commit hashes.

Also prevents panics when doing `call dolt_merge()`.

Companion PR:
https://github.com/dolthub/go-mysql-server/pull/1865


go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -5201,6 +5240,7 @@ var ThreeWayMergeWithSchemaChangeTestScripts = []MergeScriptTest{
 		Assertions: []queries.ScriptTestAssertion{
 			{
 				Query:    "call dolt_merge('right');",
+				SkipResultsCheck: true,
```
timsehn, line 5243: Is there a way to wildcard this? Seems sad to lose all the asserts.

fulghum: Right? I was thinking about this the other day... had a hacky idea where we could have a special/magic string like `"COMMIT HASH"` or `"SKIP VALIDATION"` and our test framework would skip validating that field. I'm not crazy about that idea, but it would be really nice to still assert the other values being returned are what we expect.

fulghum: Actually... I'm warming up to this, especially if we have the magic string tucked into a const and could just reference `SKIP_VALIDATION` instead of repeating the magic string. Given how many tests this change makes us needs to skip, I think an easy approach like this would be worth doing now.

go/libraries/doltcore/sqle/dprocedures/dolt_merge.go
```diff
@@ -245,15 +245,16 @@ func performMerge(ctx *sql.Context, sess *dsess.DoltSession, roots doltdb.Roots,
...
 		author := fmt.Sprintf("%s <%s>", spec.Name, spec.Email)
-		commit, _, err := doDoltCommit(ctx, []string{"-m", msg, "--author", author})
+		commit, _, err = doDoltCommit(ctx, []string{"-m", msg, "--author", author})
 		if err != nil {
 			return ws, commit, noConflictsOrViolations, threeWayMerge, fmt.Errorf("dolt_commit failed")
 		}
 	}
 
-	return ws, "", noConflictsOrViolations, threeWayMerge, nil
+	return ws, commit, noConflictsOrViolations, threeWayMerge, nil
```
fulghum, line 257: Looks like the return statement on line 216 is what gets returned when we have a fast-forward merge. I think we still want to return the commit there, too, since our return args will already indicate if it's FF or not. We've got a test for the FF return param being true, but the results check is skipped, so I think that's another argument to do some sort of "commit hash skipping" support in the test framework.

integration-tests/bats/sql-merge.bats
```diff
@@ -19,6 +19,24 @@ teardown() {
     teardown_common
 }
 
+@test "sql-merge: DOLT_MERGE with displays hash." {
```
fulghum, line 22: I think we could probably safely skip this BATS tests if we could get the golang engine tests to check more of the results when a commit is returned. We could at least check that the commit hash is not empty and looks like a commit. With those results checks turned back on, I think we'd have enough coverage of merge results and wouldn't need this. Alternatively, instead of a brand new BATS test function, you could just drop a line or two into an existing merge BATS test to assert that the returned commit hash matches HEAD.

go/libraries/doltcore/sqle/dprocedures/dolt_merge.go
```diff
@@ -92,6 +92,10 @@ func doDoltMerge(ctx *sql.Context, args []string) (string, int, int, error) {
 		return "", noConflictsOrViolations, threeWayMerge, err
 	}
 
+	if apr.NArg() < 1 {
```
JCOR11599, line 95: prevents panic caused by `branchName := apr.Arg(0)` when running query `call dolt_merge()`

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3599,12 +3599,12 @@ var DoltTagTestScripts = []queries.ScriptTest{
 				Expected: []sql.Row{{types.OkResult{RowsAffected: 2}}},
 			},
 			{
-				Query:            "CALL DOLT_COMMIT('-am','made changes in other')",
-				SkipResultsCheck: true,
+				Query:    "CALL DOLT_COMMIT('-am','made changes in other')",
+				Expected: []sql.Row{{doltCommit}},
```
fulghum, line 3603: This is so awesome to see us able to remove so many `SkipResultsCheck`s! 🙏 Great work!

fulghum (approved):
This turned into a really cool upgrade for our test framework. 👏 Thank you! 🚢

PR #6283, opened 2023-07-05
Remove temporary workaround for loopback bug in GMS

macneale4:
Resolves: https://github.com/dolthub/dolt/issues/6239


PavelSafronov (approved):
Changes look good, approving.

PR #6284, opened 2023-07-06
Autoincrement bug fixes

zachmu:
This PR adds the ability to manually reset a table's next auto increment to a lower value than those previously used, and fixes a number of other long standing bugs related to auto increment mangement

Fixes https://github.com/dolthub/dolt/issues/6253


go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3838,13 +3838,280 @@ var DoltAutoIncrementTests = []queries.ScriptTest{
...
+			"create table t (a int primary key auto_increment, b int)",
+			"call dolt_add('.')",
+			"call dolt_commit('-am', 'empty table')",
+			"insert into t (b) values (1), (2), (3), (4)",
+			"call dolt_commit('-am', 'two values on main')",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query:            "alter table t auto_increment = 2",
+				SkipResultsCheck: true,
```
fulghum, line 3885: (minor) should we just expect this to be a Query OK response with no rows updated, like MySQL returns? Might be nice just to assert that there isn't an error and use less `SkipResultsChecks`.

fulghum (approved):
Looks good! 🚢  Only a couple really minor suggestions.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
zachmu: SkipResultsCheck still asserts no error, it just doesn't inspect the return rows

PR #6287, opened 2023-07-06
Run backwards compatibility tests against more recent versions of dolt.

nicktobey:
We no longer support the old __LD_1_ file format, so the backwards compatibility tests right now are only testing against massively out of date and unsupported versions of DOLT.

This list includes v0.50.0, the first version where the new file format is the default, as well as v0.75.0, the other major release before v1.0.0, and then v1.0.0 and every minor release since.


coffeegoddd (approved):
LGTM

PR #6288, opened 2023-07-06
migrates dolt merge to use sql queries

stephkyou:
This change updates `dolt merge` to use the appropriate sql engine to generate results.


go/cmd/dolt/commands/merge.go
```diff
@@ -93,146 +99,274 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
 		defer closeFunc()
 	}
 
-	if apr.ContainsAll(cli.SquashParam, cli.NoFFParam) {
-		cli.PrintErrf("error: Flags '--%s' and '--%s' cannot be used together.\n", cli.SquashParam, cli.NoFFParam)
+	ok := validateDoltMergeArgs(apr, usage, cliCtx)
```
macneale4, line 102: Before we validate the args, verify that the queryist is a SQLEnginer instance. If it's not, then this is a remote usage, and we should print an error telling them to stop their local server.

go/cmd/dolt/commands/merge.go
```diff
@@ -93,146 +99,274 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
...
-	if apr.Contains(cli.AbortParam) {
-		mergeActive, err := isMergeActive(ctx, dEnv)
+	_, _, err = queryist.Query(sqlCtx, "set @@dolt_force_transaction_commit = 0")
+	if err != nil {
+		cli.Println(err.Error())
+		return 1
+	}
+
+	// calculate merge stats
+	if !apr.Contains(cli.AbortParam) && !apr.Contains(cli.NoCommitFlag) {
```
macneale4, line 137: Everything which fails from here on down should be scrutinized. Remember that the state of the DB has changed, and someone could be depending on the exit status of this program to control some other business logic. I wonder if failure to print statistics merits an exit status of 1. Thoughts?

macneale4: All of the stats logic should be in it's own function. I'd like to split out printing here, but that seems kind of tangled in with this.

macneale4 (changes requested):

go/cmd/dolt/commands/merge.go
```diff
@@ -369,6 +301,100 @@ func constructInterpolatedDoltMergeQuery(apr *argparser.ArgParseResults) (string
...
+			mergeStats[tableName] = &merge.MergeStats{ConstraintViolations: int(conflict[1].(uint64))}
+		}
+	}
+
+	return mergeStats, dataConflicts == nil && schemaConflicts == nil && constraintViolations == nil, 0
+}
+
+// calculateMergeStats calculates the table operations and row operations that occurred during the merge.
+func calculateMergeStats(queryist cli.Queryist, sqlCtx *sql.Context, mergeStats map[string]*merge.MergeStats) (map[string]*merge.MergeStats, int) {
+	diffSummaries, err := getDiffSummariesBetweenRefs(queryist, sqlCtx, "head^1", "head")
```
macneale4, line 353: Does this only get one side of the merge? I would think we want to get the merge stats of HEAD^1->HEAD and union them with the merge stats of HEAD^2 -> HEAD.

I don't want to rabbit hole on this, so fine to ship. But I'm curious what the current implementation does... and it definitely doesn't jump out at me when I read origin/main.

go/cmd/dolt/commands/merge.go
```diff
@@ -93,146 +100,299 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
 		defer closeFunc()
 	}
 
-	if apr.ContainsAll(cli.SquashParam, cli.NoFFParam) {
-		cli.PrintErrf("error: Flags '--%s' and '--%s' cannot be used together.\n", cli.SquashParam, cli.NoFFParam)
+	if _, ok := queryist.(*engine.SqlEngine); !ok {
+		cli.PrintErrln("`dolt merge` is only supported in local mode.")
```
macneale4, line 104: Add to the message that they need to shut down their `dolt sql-server` instance and try again.

macneale4 (changes requested):
Small wording change, but that's it!

Also, open this PR :-)

go/cmd/dolt/commands/merge.go
stephkyou: I think we only care about the `HEAD^1` -> `HEAD` diff because we want to know what changed in the branch we're currently in. `HEAD^2` -> `HEAD` diff would tell us how `HEAD` differs from `HEAD` in the branch we merged in.

go/cmd/dolt/commands/merge.go
```diff
@@ -301,12 +344,14 @@ func constructInterpolatedDoltMergeQuery(apr *argparser.ArgParseResults) (string
 	return interpolatedQuery, nil
 }
 
-// calculateMergeConflicts calculates the count of conflicts that occurred during the merge.
-func calculateMergeConflicts(queryist cli.Queryist, sqlCtx *sql.Context, mergeStats map[string]*merge.MergeStats) (map[string]*merge.MergeStats, bool, int) {
+// calculateMergeConflicts calculates the count of conflicts that occurred during the merge. Returns a map of table name to MergeStats,
+// a bool indicating whether there were any conflicts, and a bool indicating whether calculation was successful.
+func calculateMergeConflicts(queryist cli.Queryist, sqlCtx *sql.Context, mergeStats map[string]*merge.MergeStats) (map[string]*merge.MergeStats, bool, bool) {
```
macneale4, line 349: Please name the output fields:

(stats map[string]*merge.MergeStats, conflicts bool, ok bool)

macneale4: Actually, I think you should return the error rather than an ok bool. That way you can remove all of the println calls in this function, and bubble them up to the caller.

go/cmd/dolt/commands/merge.go
```diff
@@ -345,54 +392,68 @@ func calculateMergeConflicts(queryist cli.Queryist, sqlCtx *sql.Context, mergeSt
...
 	}
 
-	return mergeStats, dataConflicts == nil && schemaConflicts == nil && constraintViolations == nil, 0
+	return mergeStats, dataConflicts == nil && schemaConflicts == nil && constraintViolations == nil, true
 }
 
-// calculateMergeStats calculates the table operations and row operations that occurred during the merge.
-func calculateMergeStats(queryist cli.Queryist, sqlCtx *sql.Context, mergeStats map[string]*merge.MergeStats) (map[string]*merge.MergeStats, int) {
-	diffSummaries, err := getDiffSummariesBetweenRefs(queryist, sqlCtx, "head^1", "head")
+// calculateMergeStats calculates the table operations and row operations that occurred during the merge. Retyrns a map of
```
macneale4, line 398: returns

go/cmd/dolt/commands/merge.go
```diff
@@ -136,30 +138,65 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+		if !ok {
+			return 0
+		}
 
-		mergeStats, noConflicts, err := calculateMergeConflicts(queryist, sqlCtx, mergeStats)
-		if err != 0 {
-			return 1
+		if noConflicts {
+			if apr.Contains(cli.NoCommitFlag) {
+				mergeStats, ok = calculateMergeStats(queryist, sqlCtx, mergeStats, "STAGED")
```
macneale4, line 177: This is switched. As discussed, want HEAD -> STAGED

Add test to ensure we get the directionality right here.

go/cmd/dolt/commands/merge.go
```diff
@@ -345,54 +392,68 @@ func calculateMergeConflicts(queryist cli.Queryist, sqlCtx *sql.Context, mergeSt
...
-			diffStats[tableStats[0].TableName] = tableStats[0]
 		} else {
 			mergeStats[summary.TableName] = &merge.MergeStats{
 				Operation: merge.TableUnmodified,
 			}
 		}
 	}
 
+	if allUnmodified {
+		cli.Println("Already up to date.")
```
macneale4, line 445: moving all cli.Println out of the `calculateMergeStats` function would be ideal, but may be impractical. See what you can do.

macneale4 (changes requested):

go/cmd/dolt/commands/merge.go
```diff
@@ -102,7 +102,7 @@ func (cmd MergeCmd) Exec(ctx context.Context, commandStr string, args []string,
 	}
 
 	if _, ok := queryist.(*engine.SqlEngine); !ok {
-		cli.PrintErrln(cli.RemoteUnsupportedMsg)
+		cli.Println(cli.RemoteUnsupportedMsg)
```
macneale4, line 105: For this, you need to stick commandStr at the beginning. If you want to standardize this usage, this is the other example: https://github.com/dolthub/dolt/blob/1901dfeae2566a291e308a480802abc0339cdfcc/go/cmd/dolt/commands/checkout.go#L104

macneale4 (approved):
Small change needed, but you can ship with out  review when tests pass!

PR #6293, opened 2023-07-06
Migrate dolt table cp/mv/rm commands to SQL

PavelSafronov:
Migrate `dolt table cp`/`dolt table mv`/`dolt table rm` to use SQL.

This change also adds a new convenience method, InterpolateAndRunQuery, which can be used to interpolate a query, execute the query, and get all the result rows. Since this method does not return a schema, this method is useful only for fire-and-forget types of queries.

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/cp.go
+++ b/go/cmd/dolt/commands/tblcmds/cp.go
@@ -17,14 +17,16 @@ package tblcmds
 import (
 	"context"
 	"fmt"
+
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/gocraft/dbr/v2"
+
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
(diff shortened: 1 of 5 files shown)


macneale4 (changes requested):
What the test coverage status for this command? A new test in sql-local-remote.bats is needed

PavelSafronov:
> What the test coverage status for this command? A new test in sql-local-remote.bats is needed

Added sql-local-remote.bats tests

PavelSafronov:
This is only a partial migration of `dolt table`, as we're not yet migrating `dolt table import` and `dolt table export`.
This PR is being marked as Draft and will be resurrected when we are ready to migrate the entire `dolt table` command.

macneale4:
Preserving the branch: https://github.com/dolthub/dolt/tree/pavel/migrate-dolt-table-cp-mv-rm

There are several other issued to work out with the `table` command, and these will wait until we get there.

PR #6296, opened 2023-07-07
Migrate `dolt revert` to sql backend

macneale4:
* Update the `dolt revert` command to use dolt_revert() stored procedure.
* Update dolt_revert() stored procedure to respect the dolt_ignore table
* Add dolt_revert() dedicated engine tests
* `dolt revert` bats tests enabled for remote testing

https://github.com/dolthub/dolt/issues/3922


go/cmd/dolt/commands/revert.go
```diff
@@ -79,89 +85,66 @@ func (cmd RevertCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+		author, _ = apr.GetValue(cli.AuthorParam)
+	} else {
+		name, email, err := env.GetNameAndEmail(cliCtx.Config())
 		if err != nil {
-			return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+			cli.Println(err.Error())
+			return 1
 		}
-		commits[i] = commit
+		author = name + " <" + email + ">"
```
PavelSafronov, line 106: Nit/Q:
Why do manual concatenation here instead of `fmt.Sprintf`?
Also, since the author information is not available for SQL sessions and our commands need to start passing this information more often, should this be a method? `env.GetAuthor` or something?

go/cmd/dolt/commands/revert.go
```diff
@@ -79,89 +85,66 @@ func (cmd RevertCmd) Exec(ctx context.Context, commandStr string, args []string,
...
+		buffer.WriteString(", ?")
+		params = append(params, input)
 	}
-	opts := editor.Options{Deaf: dEnv.DbEaFactory(), Tempdir: tmpDir}
-	workingRoot, revertMessage, err := merge.Revert(sqlCtx, dEnv.DoltDB, workingRoot, headCommit, commits, opts)
+	buffer.WriteString(")")
+	query, err := dbr.InterpolateForDialect(buffer.String(), params, dialect.MySQL)
 	if err != nil {
-		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+		cli.Println(err.Error())
```
PavelSafronov, line 122: Should we be including other error information in these 3 cases? (Lines 122, 128, 133.) For example, on line 139, we're calling out that the main operation succeeded, but something else failed.
These error can be improved a bit by being clear about what failed: interpolation on line 122, query on line 128, reading the results on 131.

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -1597,6 +1597,28 @@ func TestDoltMergePrepared(t *testing.T) {
 	}
 }
 
+func TestDoltRevert(t *testing.T) {
+	for _, script := range RevertScripts {
+		// harness can't reset effectively. Use a new harness for each script
```
PavelSafronov, line 1602: Is this caused by how `revert` is implemented? How is this particular situation different from other similar tests?

macneale4: It's not. I'm follow patterns of other tests which make a habit of changing the contents of the branches.

go/cmd/dolt/commands/revert.go
macneale4: In the first and third cases, I don't expect us to every hit those in normal usage. I did update the second to state the sql query which failed. Will change the output a little, but should be ok to pass tests.

Issue #3922, opened 2022-07-25
`dolt sql` and other commands should connect to running database


zachmu:
It's not safe to issue CLI commands to update a database while a server is running. To prevent issues this might cause, we made any command that updates database state fail if a server is running. This makes the command line much less useful.

Instead of locking the database when a server is running, we should implement the following:

1) Implement all `dolt` CLI commands as SQL commands
2) Make `dolt` automatically connect to appropriate running server to issue its command
3) If no server is running, the command behaves exactly as if the command started a server, issued the single command, and quit the server.
4) The command line client is a single long-term session that persists between database invocations

This design necessarily also resolves odd differences in behavior between running a command and issuing its equivalent SQL statement.

We can make this change incrementally, but we shouldn't write new commands or make major changes to existing ones without considering this design.

fulghum:
Makes sense to me. I love unifying on the sql path and cleaning up the duplicated logic in the CLI commands.

The only downside I could brainstorm was that starting a server will slow down some of the CLI commands, but that seems negligible for a command line experience.

timsehn:
As you can see, this is heavily in progress.

timsehn:
Resoving this uber issue as we have sub issues for the commands that don't now.

PR #6299, opened 2023-07-07
Migrate `dolt tag` to use SQL

PavelSafronov:
This change updates `dolt tag` command to use only SQL.

Related: #3922


integration-tests/bats/sql-local-remote.bats
```diff
@@ -736,3 +741,168 @@ SQL
...
+
+  cd ../defaultDB
+  start_sql_server defaultDB
+
+  # get commit hashes
+  headCommit=$(get_commit_hash_at HEAD)
+  secondCommit=$(get_commit_hash_at HEAD~1)
+
+  # show tags
+  run dolt tag
```
macneale4, line 835: add the --verbose-engine-setup flag, and check the output to confirm we're really going remote. Alter an invocation in the local context as well.

go/cmd/dolt/commands/tag.go
```diff
@@ -76,6 +86,14 @@ func (cmd TagCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
+	if err != nil {
+		return handleStatusVErr(err)
+	}
+	if closeFunc != nil {
+		defer closeFunc()
+	}
+
 	// list tags
 	if len(apr.Args) == 0 {
 		var verr errhand.VerboseError
```
macneale4, line 99: There is no reason to have this line three times - lines 99, 115, 133. Just define once at the top of the function

integration-tests/bats/sql-local-remote.bats
macneale4, line 1: Looks like mosts tests in `commit_tags.bats` could be enabled in local-remote.bash. I see one test that we should skip until checkout is ready.

go/cmd/dolt/commands/tag.go
```diff
@@ -112,81 +136,113 @@ func (cmd TagCmd) Exec(ctx context.Context, commandStr string, args []string, dE
...
+			query = "call dolt_tag(?, ?)"
+			params = []interface{}{tagName, startPoint}
+		} else {
+			query = "call dolt_tag(?, ?, '--author', ?)"
+			params = []interface{}{tagName, startPoint, author}
+		}
 	} else {
-		name, email, err = env.GetNameAndEmail(dEnv.Config)
+		if len(author) == 0 {
+			query = "call dolt_tag('-m', ?, ?, ?)"
```
macneale4, line 169: You should always pass in an author. The stored procedure doesn't have the local user information available to it, so it puts in root@dolt (or something like that) as the author.

macneale4 (changes requested):

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Good idead. Adding `--verbose-engine-setup` flag for the first few invocations and verifying that we're getting the results we're expecting.

go/cmd/dolt/commands/tag.go
PavelSafronov: True, though it is nice to have self-contained sections of the code and not have to worry about fall-through. For example, lines 97-111 deal with listing tags, and that's all, no need to look elsewhere.
That being said, most of these blocks are just doing some param validation and then delegate to a small function, so I'm going to move the logic into those functions. Should make things a bit cleaner.

integration-tests/bats/sql-local-remote.bats
PavelSafronov: Enabling.

go/cmd/dolt/commands/tag.go
PavelSafronov: Hmm, that's not what I'm observing. This BATS test just succeeded for me. I'm specifically checking for local/remote execution, not specifying an author, and verifying that the author is `Bats Tests <bats@email.fake>`:

```sh
  # add tag without message
  run dolt --verbose-engine-setup tag v1
  [ $status -eq 0 ]
  [[ $output =~ "verbose: starting remote mode" ]] || false

  # list tags and check new tag is present
  run dolt tag
  [ $status -eq 0 ]
  [[ $output =~ "v1" ]] || false

  # list tags with verbose flag and check new tag is present
  run dolt --verbose-engine-setup tag --verbose
  [ $status -eq 0 ]
  [[ $output =~ "v1"$'\t'"$headCommit" ]] || false
  [[ $output =~ "Tagger: Bats Tests <bats@email.fake>" ]] || false
  [[ $output =~ "verbose: starting remote mode" ]] || false
```

macneale4: Huh. Interesting. I'm convinced by the test :)

macneale4 (approved):
```
    __4___
 _  \ \ \ \
<'\ /_/_/_/
 ((____!___/)
  \0\0\0\0\/
 ~~~~~~~~~~~
```

PR #6311, opened 2023-07-10
During CWBHeadRef/SetCWBHeadRef, surface RSLoadErr if it is present.

PavelSafronov:
This avoids the panic seen in https://github.com/dolthub/dolt/issues/6306 by surfacing the un-processed error `RSLoadErr` during CWBHeadRef/SetCWBHeadRef calls.

### DIFF
--- a/go/libraries/doltcore/env/environment.go
+++ b/go/libraries/doltcore/env/environment.go
@@ -663,10 +663,16 @@ type repoStateReader struct {
 }
 
 func (r *repoStateReader) CWBHeadRef() (ref.DoltRef, error) {
+	if r.RSLoadErr != nil {
+		return nil, r.RSLoadErr
+	}
 	return r.RepoState.CWBHeadRef(), nil
 }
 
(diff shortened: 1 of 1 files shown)


fulghum (approved):
Code fix looks good. Do you know if we have any tests that trigger `RSLoadErr` to be thrown?  Would be cool to have a test case covering `dolt status` throwing this error, but I'm not sure if our existing tests have a good example of RSLoadErr yet.

PavelSafronov:
> `RSLoadErr`

The customer included the "broken" dolt DB (it's ~1kb), so we can definitely add some integration tests that try to act on  the broken repo.

We are doing some testing of `RSLoadErr`, but it's not extensive: https://github.com/dolthub/dolt/blob/55a13e84523cb0059dfcc29f66bf647173075fd9/go/libraries/doltcore/env/environment_test.go#L86

I tried poking around in this area, but trying to create an empty env and then run a command against it is causing cyclical dependencies with the libraries and tests. I'll try to add checks like these during the next phase of the "CLI to SQL" project.

PR #6314, opened 2023-07-11
Migrating `dolt push` to use SQL

PavelSafronov:
Related: #3922


integration-tests/bats/remotes.bats
macneale4, line 1: Are there changes coming for sql-local-remote.bats? There is a heavy use of the `dolt push` command in `remotes-localbs.bats` - what happens when you remove that from `helper/local-remote.bash` ?

go/cmd/dolt/commands/push.go
```diff
@@ -85,95 +95,236 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
-		verr = printInfoForPushError(err, opts.Remote, opts.DestRef, opts.RemoteRef)
+		return nil, fmt.Errorf("failed to read dolt remotes: %w", err)
 	}
 
-	if opts.SetUpstream {
-		err := dEnv.RepoState.Save(dEnv.FS)
+	remotes := map[string]remoteInfo{}
+	for _, row := range rows {
+		name := row[0].(string)
+		url := row[1].(string)
```
macneale4, line 238: Pull out all columns here, that lets the reader see clearly the full structure of the response table. Don't use row[n] anywhere below.

go/cmd/dolt/commands/push.go
```diff
@@ -85,95 +95,236 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
-		status, ok := status.FromError(err)
-		if ok && status.Code() == codes.PermissionDenied {
-			cli.Println("hint: have you logged into DoltHub using 'dolt login'?")
-			cli.Println("hint: check that user.email in 'dolt config --list' has write perms to DoltHub repo")
-		}
-		if rpcErr, ok := err.(*remotestorage.RpcError); ok {
-			return errhand.BuildDError("error: push failed").AddCause(err).AddDetails(rpcErr.FullDetails()).Build()
-		} else {
-			return errhand.BuildDError("error: push failed").AddCause(err).Build()
+func getTrackingRefs(branchName string, info remoteInfo) (fromRef, toRef string, err error) {
```
macneale4, line 309: gDoc plz

go/cmd/dolt/commands/push.go
```diff
@@ -85,95 +95,236 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+		fetchSpecs := []string{}
+		for _, spec := range fetchSpecsArray {
+			text, ok := spec.(string)
+			if !ok {
+				return nil, fmt.Errorf("failed to read fetch specs for remote %s: %w", name, err)
+			}
+			fetchSpecs = append(fetchSpecs, text)
+		}
+
+		paramsJson, err := getJsonDocumentCol(sqlCtx, row[3])
```
macneale4, line 257: It's not clear to me what we support in this arbitrary hash map of parameters. There seems to be a lack of documentation on what this structure allows, which is not really a problem for this PR.

go/cmd/dolt/commands/push.go
```diff
@@ -85,95 +95,236 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+	if !changesMade {
+		cli.Println("Everything up-to-date")
 	}
-	var verr errhand.VerboseError
-	err = actions.DoPush(ctx, dEnv.RepoStateReader(), dEnv.RepoStateWriter(), dEnv.DoltDB, remoteDB, tmpDir, opts, buildProgStarter(defaultLanguage), stopProgFuncs)
+
+	return nil
+}
+
+func getChangesMade(remoteBranches map[string]remoteBranchInfo, postPushRemoteBranches map[string]remoteBranchInfo) bool {
```
macneale4, line 209: Docs please.

I would rename the arguments to be prePushBranches and postPushBranches. And, I'm pretty sure you just want to compare them for equality which doesn't require all the logic below.

macneale4 (changes requested):
I think the most important thing is to expand test coverage. There are a few nits in the code, but I don't think we'll have confidence on this until we expand test coverage

integration-tests/bats/remotes.bats
PavelSafronov: Adding those tests now.

go/cmd/dolt/commands/push.go
PavelSafronov: Restructuring now.

go/cmd/dolt/commands/push.go
PavelSafronov: Yeah, clear lack of docs. Since params aren't really supported (by any of the existing dolt code), we can remove the Params property and only add it back when we add params support. Thoughts?

go/cmd/dolt/commands/push.go
PavelSafronov: Sure thing. Will try a basic equality comparison, hopefully that'll be enough.

macneale4:
Shelving this for the time being. I'll revive in the next phase of CLI migrations.

PR #6317, opened 2023-07-11
Fix intermittent BATS failure

PavelSafronov:
Fix intermittent BATS failure that happens when a tag like "v2" appears in the commit hash.


macneale4 (approved):
LGTM

PR #6323, opened 2023-07-12
Bug Fix: Merge with convergent schema changes

fulghum:
There were a couple of gaps in our schema merge logic and testing when convergent schema changes happen on both sides of a merge. This change fixes an issue where columns could be omitted if they changed in exactly the same way, and also an issue with spurious FK conflicts if identical FKs were added on both sides of a merge.


JCOR11599 (approved):
LGTM, nice test!

PR #6324, opened 2023-07-12
dolt checkout message

macneale4:
Currently the `dolt checkout` can't work with a server running on a local database. There are more details to be sorted out with session persistence in the server which aren't addressed yet.

The SQLEngine execution for using SQL as the backend is complete, but now we simply print a message stating that the user should stop the server.

### DIFF
--- a/integration-tests/bats/sql-local-remote.bats
+++ b/integration-tests/bats/sql-local-remote.bats
@@ -1033,3 +1033,16 @@ SQL
 
   [[ "$localCherryPickOutput" == "$remoteCherryPickOutput" ]] || false
 }
+
+@test "sql-local-remote: verify checkout will fail early when a server is running" {
+  cd altDB
+  dolt reset --hard
+  start_sql_server altDB
+
+  dolt branch br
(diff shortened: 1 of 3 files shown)


integration-tests/bats/sql-local-remote.bats
```diff
@@ -1033,3 +1033,16 @@ SQL
 
   [[ "$localCherryPickOutput" == "$remoteCherryPickOutput" ]] || false
 }
+
+@test "sql-local-remote: verify checkout will fail early when a server is running" {
+  cd altDB
+  dolt reset --hard
```
stephkyou, line 1039: Is this `reset` just to make sure the workspace is clean before proceeding? Not too clear why we need this.

stephkyou (approved):

integration-tests/bats/sql-local-remote.bats
macneale4: Added comments in the test.

nicktobey (approved):

PR #6344, opened 2023-07-14
updates `dolt_merge` to allow dirty working sets with no conflicts

stephkyou:
Currently, the `dolt_merge` stored procedure doesn't allow any uncommitted changes in the working set, it will abort the merge if any are found. However the CLI (and git) allows uncommitted changes in the working set if they won't get stomped by the merge. This change updates the `dolt_merge` stored procedure to allow that.


integration-tests/bats/sql-merge.bats
macneale4, line 1: Weird that `dolt merge` was being used for sql-merge tests. Glad you are fixing this.

integration-tests/bats/sql-merge.bats
```diff
@@ -1130,6 +1130,173 @@ SQL
...
+
+    run dolt log
+    log_status_eq 0
+    [[ "$output" =~ "no-ff merge" ]] || false
+}
+
+@test "sql-merge: squash merge" {
+    dolt commit -m "initial commit"
+    dolt checkout -b merge_branch
+    dolt SQL -q "INSERT INTO test values (3)"
```
macneale4, line 1192: s/SQL/sql/

For this whole test.

integration-tests/bats/sql-merge.bats
```diff
@@ -1130,6 +1130,173 @@ SQL
...
+    dolt SQL -q "CREATE TABLE t2 (pk int)"
+    dolt SQL -q "INSERT INTO t2 values (9)"
+    run dolt status
+    log_status_eq 0
+    [[ "$output" =~ "t2" ]] || false
+    [[ ! "$output" =~ "test1" ]] || false
+
+    run dolt sql -q "CALL DOLT_MERGE('--squash', 'merge_branch', '--no-commit');"
+    log_status_eq 0
+    [[ "$output" =~ "| 0            | 0         |" ]] || false
```
macneale4, line 1210: Adding the header line would make the test more readable as a table. Some tests verify every line of output is there, but I'm not a huge fan of that.

integration-tests/bats/sql-merge.bats
```diff
@@ -1130,6 +1130,173 @@ SQL
...
+
+    run dolt status
+    log_status_eq 0
+    [[ "$output" =~ "t2" ]] || false
+    [[ "$output" =~ "test" ]] || false
+
+    # make sure the squashed commit is not in the log.
```
macneale4, line 1217: I don't understand this comment. Looks like this is intended to be down at line 1224

integration-tests/bats/sql-merge.bats
```diff
@@ -1130,6 +1130,173 @@ SQL
...
+    dolt checkout main
+    # dirty the working set with changes to test2
+    dolt SQL -q "CREATE TABLE t2 (pk int)"
+    dolt sql -q "INSERT INTO t2 VALUES (5);"
+
+    dolt merge other --no-commit
+    dolt sql -q "call dolt_merge('--abort');"
+
+    run dolt sql -q "SELECT * from dolt_merge_status"
+    [[ "$output" =~ "false" ]] || false
```
macneale4, line 1247: Put more of the table in there. "false" can show up in a lot of contexts, include stack traces - which is why you should verify the exit status as well.

integration-tests/bats/sql-merge.bats
```diff
@@ -1130,6 +1130,173 @@ SQL
...
+    # make sure all the commits make it into the log
+    dolt add .
+    dolt commit -m "squash merge"
+
+    run dolt log
+    log_status_eq 0
+    [[ "$output" =~ "add pk 9 to test" ]] || false
+    [[ "$output" =~ "add pk 8 to test" ]] || false
+}
+
```
macneale4, line 1299: I don't see any tests for fast forward merges with a dirty directory. Looks like you changed the behavior of the executeFFMerge method, and I'd expect that to need a new test or break an old test... or is this covered because several old tests we moved to the sql call rather than `dolt merge`?

go/libraries/doltcore/sqle/dprocedures/dolt_merge.go
```diff
@@ -455,41 +463,58 @@ func createMergeSpec(ctx *sql.Context, sess *dsess.DoltSession, dbName string, a
 
 // TODO: this copied from commands/merge.go because the latter isn't reusable. Fix that.
```
macneale4, line 464: This comment can be removed now :)

macneale4 (changes requested):

go/libraries/doltcore/sqle/dprocedures/dolt_pull.go
```diff
@@ -215,3 +219,22 @@ func stopProgFuncs(cancel context.CancelFunc, wg *sync.WaitGroup, statsCh chan p
...
+		return err
+	}
+
+	hrh, err := headRoot.HashOf()
+
+	if err != nil {
+		return err
+	}
+
+	if rh != hrh {
```
macneale4, line 236: It's not 1968, there is no shortage of storage. Longer names please. How about rootHash and headHash?

go/libraries/doltcore/sqle/dprocedures/dolt_pull.go
```diff
@@ -152,6 +152,10 @@ func doDoltPull(ctx *sql.Context, args []string) (int, int, error) {
 			if err != nil {
 				return noConflictsOrViolations, threeWayMerge, err
 			}
+			err = checkForUncommittedChanges(ctx, roots.Working, roots.Head)
```
macneale4, line 155: What's stopping us from supporting pull with dirty workspaces? It's no different from merge. Is this to comply with existing tests? Allowing it would be closer to git behavior and consistent with a pull being a fetch -> merge. Let's leave for now, and fast follow as priority allows.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -368,7 +367,7 @@ var MergeScripts = []queries.ScriptTest{
 			},
 			{
 				Query:    "SELECT * from dolt_status",
-				Expected: []sql.Row{{"test", true, "modified"}, {"test", false, "conflict"}},
```
macneale4, line 371: I _think_ I understand why this is correct now, but want to sleep on it.

macneale4 (changes requested):
Tiny change requested.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
macneale4: I'm ok with this since this is going to get a version bump.

macneale4 (approved):

PR #6345, opened 2023-07-14
Small changes to help improve the wait_for_connection failures

macneale4:
* Pick from a random port from 2048 - 6146
* Use `nc` to check if port is in use rather than `lsof`
* Wait longer, in case the server retries successfully.

I tested by running 10 changes through GH actions. 5 jobs were controls with no changes, 5 had this change. The control group had 2 jobs which failed waiting to connect, while the group with this change had no failures.


nicktobey (approved):
Ship it and hope this is the last we see of this issue.

PR #6346, opened 2023-07-15
Allow empty databases in the shell prompt

macneale4:
Easy to reproduce issue:

```
mkdir foobar
cd foobar
dolt sql-server
```

```
dolt sql
> select 1;
```

And panic() follows. This actually happens for any command, so `create database dba` included.

### DIFF
--- a/integration-tests/bats/sql-shell.bats
+++ b/integration-tests/bats/sql-shell.bats
@@ -71,7 +71,20 @@ teardown() {
     $BATS_TEST_DIRNAME/sql-works-after-failing-query.expect
 }
 
+@test "sql-shell: empty DB in prompt is OK" {
+    skiponwindows "Need to install expect and make this script work on windows."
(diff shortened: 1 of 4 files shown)


integration-tests/bats/sql-shell.bats
```diff
@@ -71,7 +71,20 @@ teardown() {
...
+    fi
+    # ignore common setup. Use an empty db with no server.
+    rm -rf .dolt
+    mkdir emptyDb
+    cd emptyDb
+    $BATS_TEST_DIRNAME/sql-shell-empty-prompt.expect
+}
+
 @test "sql-shell: delimiter" {
+
```
fulghum, line 87: (super minor) Need this empty line here? Feels like it makes the body of the test disconnected from the test header line.

```suggestion
```

fulghum (approved):
Looks great! 🚢

PR #6349, opened 2023-07-17
Removing old schema merge guard 

fulghum:
Before we added support for schema merge conflict reporting, we had a guard in place that would error out for schema changes in merges. This code prevented schema merges that Dolt couldn't handle, but it didn't provide much visibility into the exact schema conflicts. The new schema merge conflict reporting code is able to handle these same cases and provides better visibility into exactly what the schema conflicts are. We've already removed some of the checks from this guard, but there were still a few left in place. This PR completely removes the old schema merge guard code.

The code left in this guard was still protecting against a couple of cases that are now fixed and covered with tests:
- **Collation changes** – we weren't properly handling merging changes to a table's default collation. Fixed and tested now.
- **New, non-nullable columns with no default value** – new columns that are non-nullable and don't have a default value can't be auto merged since we don't know what value to safely use in the new column for any existing rows, so this now throws an error with instructions on how to fix.


zachmu (approved):
LGTM!

PR #6352, opened 2023-07-17
Bug fix: Encode binary values in hex for SQL patch statements

fulghum:
Current SQL formatting code converts binary values to strings, which won't round-trip back to the database correctly. This PR changes binary values to be hex-encoded. 

Fixes https://github.com/dolthub/dolt/issues/6350


JCOR11599 (approved):
LGTM

PR #6363, opened 2023-07-19
CLI remote connections

macneale4:
Enable the connection of SQL enabled CLI commands to remote servers with addition of three flags:
* --host - The server to connect to. Using this flag forces the use of a remote connection no matter what.
* --port - The port to use. Defaults to 3306. No effect when --host is not used.
* --no-tls - Disable the use of SSL. This is required to connect to your local host.

Also special cased the `sql` command to allow starting without a DB. Enables connecting to an empty repo so you can `CREATE DATABASE FOO`.


nicktobey:
Initial thought before actually reading any of the code: Can using --port without --host be an error instead of the --port flag having no effect?

go/cmd/dolt/dolt.go
```diff
@@ -558,11 +558,29 @@ func runMain() int {
 	return res
 }
 
+// buildLateBinder builds a LateBindQueryist for which is used to obtain the Querist used for the length of the
```
nicktobey, line 561: s/Querist/Queryist

go/cmd/dolt/commands/show.go
```diff
@@ -112,6 +112,10 @@ func (cmd ShowCmd) ArgParser() *argparser.ArgParser {
 	return ap
 }
 
+func (cmd ShowCmd) RequiresRepo() bool {
+	return false
```
nicktobey, line 116: This (and the change to diff) feel unrelated to enabling remote connections and probably belong in separate PR.

integration-tests/bats/cli-hosted.bats
```diff
@@ -0,0 +1,106 @@
...
+# Honestly, this test was created in haste, and because multiple parts of the default setup infrastructure don't work
+# for this use case.
+
+setup() {
+  setup_no_dolt_init
+
+  mkdir emptyDB
+  cd emptyDB
+
+  # Unusually situation - this is the only test that I know of where we want to start a server completely
```
nicktobey, line 19: s/Unusually/Unusual

integration-tests/bats/cli-hosted.bats
```diff
@@ -0,0 +1,106 @@
...
+  run dolt $TLS --host $HST --port $PRT --user bogus sql -q "create database bats_test_cli_hosted"
+  [ "$status" -eq 1 ]
+  [[ $output =~ "User not found 'bogus'" ]] || false
+}
+
+# This test will not work if you change DOLT_CLI_USER above. Be aware if you run against a hosted instance.
+@test "cli-hosted: bogus password rejected" {
+  run dolt $TLS --host $HST --port $PRT --user dolt --password bogus sql -q "create database bats_test_cli_hosted"
+
+  echo "-------------------"
```
nicktobey, line 99: Is this leftover from debugging?

go/cmd/dolt/commands/show.go
nicktobey: I see now that the bats test uses them. These should _probably_ be flipped on in a separate RP and then merged into this one... but I don't think it's a big deal either way.

nicktobey (approved):
Very cool. I only have a couple of nits.

go/cmd/dolt/dolt.go
stephkyou, line 561: spelling "Queryist"

stephkyou (approved):
LGTM!

go/cmd/dolt/commands/show.go
macneale4: They should have been enabled in the original migration. Making another PR seems like overkill.

integration-tests/bats/cli-hosted.bats
macneale4: doh.

macneale4:
> Initial thought before actually reading any of the code: Can using --port without --host be an error instead of the --port flag having no effect?

I was going to punt on that, but I went ahead and did it

nicktobey:
🚢

macneale4:
Enable the connection of SQL enabled CLI commands to remote servers with addition of three flags:

--host - The server to connect to. Using this flag forces the use of a remote connection no matter what.
--port - The port to use. Defaults to 3306. No effect when --host is not used.
--no-tls - Disable the use of SSL. This is required to connect to your local host.
Also special cased the sql command to allow starting without a DB. Enables connecting to an empty repo so you can CREATE DATABASE FOO.

PR #6367, opened 2023-07-20
Bug fixes for `dolt_patch`

fulghum:
Adding support in `dolt_patch()` to: 
- return data diff patch statements when there are schema changes
- return schema diff patch statements for charset/collation changes


JCOR11599 (approved):
LGTM

PR #6373, opened 2023-07-21
Support clone_admin grants for authenticating to remotesapi

tbantle22:
Fixes https://github.com/dolthub/dolt/issues/6354
GMS PR:  https://github.com/dolthub/go-mysql-server/pull/1882


go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -357,6 +363,30 @@ func Serve(
...
+
+func (r *remotesapiAuth) Authenticate(ctx context.Context, creds *remotesrv.RequestCredentials) bool {
+	if creds == nil {
+		return true
+	}
+	if r.serverConfig.User() == creds.Username {
+		return r.serverConfig.Password() == creds.Password
+	}
+
+	r.ctx.Session.SetClient(sql.Client{User: creds.Username, Address: "localhost", Capabilities: 0})
```
reltuk, line 384: This seems wrong to me. The caller isn't authenticated as the calling username unless they present credentials which match with the credentials stored in MySQLDb, but we don't seem to check those here. Am I just reading it wrong?

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -357,6 +363,30 @@ func Serve(
...
+func newAuthenticator(ctx *sql.Context, serverConfig ServerConfig, sqlEngine *engine.SqlEngine) remotesrv.Authenticator {
+	return &remotesapiAuth{ctx, serverConfig, sqlEngine}
+}
+
+func (r *remotesapiAuth) Authenticate(ctx context.Context, creds *remotesrv.RequestCredentials) bool {
+	if creds == nil {
+		return true
+	}
+	if r.serverConfig.User() == creds.Username {
+		return r.serverConfig.Password() == creds.Password
```
reltuk, line 381: This definitely needs to be constant time compare...

go/libraries/doltcore/remotesrv/interceptors.go
```diff
@@ -66,9 +69,10 @@ func (si *ServerInterceptor) Options() []grpc.ServerOption {
 }
 
 func (si *ServerInterceptor) authenticate(ctx context.Context) error {
-	if len(si.ExpectedUserAuth.User) == 0 && len(si.ExpectedUserAuth.Password) == 0 {
+	if !si.ShouldAuth || si.Authenticator == nil {
```
reltuk, line 72: What's going on here? Shouldn't we run this code even if a static username and password aren't configured? Isn't that the point of adding the grants support?

FWIW, if there are fast path exits like this, I think we should try to not install this interceptor at all instead. So, like, instead of ShouldAuth being a thing here, we could just do:

```go
if shouldAuth {
    args = args.WithUserPasswordAuth(...)
}
```

reltuk (commented):
Some initial questions trying to clear up some confusion.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -357,6 +366,42 @@ func Serve(
...
+
+func newAuthenticator(ctxFactory func() (*sql.Context, error), serverConfig ServerConfig, rawDb *mysql_db.MySQLDb) remotesrv.Authenticator {
+	return &remotesapiAuth{ctxFactory, serverConfig, rawDb}
+}
+
+func (r *remotesapiAuth) Authenticate(creds *remotesrv.RequestCredentials) bool {
+	if r.serverConfig.User() == creds.Username {
+		compare := subtle.ConstantTimeCompare([]byte(r.serverConfig.Password()), []byte(creds.Password))
+		return compare != 0
+	}
```
tbantle22, line 383: I might not need this check anymore?

PR #6374, opened 2023-07-21
adds support for `db/branch` syntax with `--use-db` global arg

stephkyou:
Adds support for `db/branch` syntax with `--use-db` global arg.


go/cmd/dolt/dolt.go
```diff
@@ -587,9 +587,10 @@ func buildLateBinder(ctx context.Context, cwdFS filesys.Filesys, mrEnv *env.Mult
 	}
 
 	if hasUseDb {
-		targetEnv = mrEnv.GetEnv(useDb)
+		dbName, _ := dsess.SplitRevisionDbName(useDb)
+		targetEnv = mrEnv.GetEnv(dbName)
 		if targetEnv == nil {
-			return nil, fmt.Errorf("The provided --use-db %s does not exist or is not a directory.", useDb)
+			return nil, fmt.Errorf("The provided --use-db %s does not exist or is not a directory.", dbName)
```
nicktobey, line 593: Is it still possible to use the `--use-db` flag to specify a directory? Is that something that we want to support? If not, we should probably edit this error message.

nicktobey (approved):

go/cmd/dolt/commands/sqlserver/sqlclient.go
```diff
@@ -481,11 +482,14 @@ func BuildConnectionStringQueryist(ctx context.Context, cwdFS filesys.Filesys, c
...
 	}
 
-	parsedMySQLConfig, err := mysqlDriver.ParseDSN(ConnectionString(clientConfig, database))
+	dbName, _ := dsess.SplitRevisionDbName(database)
+	parsedMySQLConfig, err := mysqlDriver.ParseDSN(ConnectionString(clientConfig, dbName))
 	if err != nil {
 		return nil, err
 	}
 
+	// ParseDSN currently doesn't support `/` in the db name
```
macneale4, line 491: This comment seems like it's in the wrong place. move it up.

Also, "database" is actually a databaseRevision (maybe dbRev?). Renaming it will help future readers.

go/cmd/dolt/dolt.go
macneale4: Yeah, that message doesn't make sense.

macneale4 (approved):
LGTM. Small refactor requested, but you can ship without another review

PR #6375, opened 2023-07-21
Bug fix: Commits created with `cherry-pick` should only have one parent commit

fulghum:
When a cherry-pick requires conflict resolution, we set MergeState metadata in the working set to mark that a merge is in progress. `dolt commit` sees this metadata and incorrectly records the cherry-picked commit as a parent of the new, created commit. This PR fixes this, so that commits created by cherry-pick always have only one parent. To do this, we record an extra flag in MergeState that the current merge is a cherry-pick.

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/session.go
+++ b/go/libraries/doltcore/sqle/dsess/session.go
@@ -657,7 +657,7 @@ func (d *DoltSession) newPendingCommit(ctx *sql.Context, branchState *branchStat
 	}
 
 	var mergeParentCommits []*doltdb.Commit
-	if branchState.WorkingSet().MergeActive() && branchState.WorkingSet().MergeState().IsCherryPick() == false {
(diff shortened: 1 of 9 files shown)


go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -657,7 +657,7 @@ func (d *DoltSession) newPendingCommit(ctx *sql.Context, branchState *branchStat
 	}
 
 	var mergeParentCommits []*doltdb.Commit
-	if branchState.WorkingSet().MergeActive() {
+	if branchState.WorkingSet().MergeActive() && branchState.WorkingSet().MergeState().IsCherryPick() == false {
```
zachmu, line 660: Maybe a method `workingSet.MergeCommitParents` or similar?

zachmu (approved):
LGTM!

PR #6389, opened 2023-07-25
fix panic in `dolt_history_<table>`

JCOR11599:
The `dolt_history_...` table derives its schema from the base table + 3 additional columns [historyTableSchema()](https://github.com/dolthub/dolt/pull/6389/files#diff-bde037bd1753c546c0e70386a347eeddca4756f13d3976419a2eb3472e783867L146).

However, when we have a projection over the table, the schema returned from `HistoryTable.Schema()` is trimmed to only the columns we look at. This is problematic when we're join planning as we look over all indexes and panic when we find an index over a column that isn't in the schema; in this case, it's the Primary key.

By having `HistoryTable` implement `PrimaryKeySchema`, we can return a schema with all the columns, and the index picking portion of join planning works.

Fix for: https://github.com/dolthub/dolt/issues/6382

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -2114,6 +2114,22 @@ ORDER BY
 					{300, doltCommit},
 				},
 			},
+			{
+				Query: `
+SELECT z
+FROM xyz
+WHERE z IN (
+  SELECT z
+  FROM dolt_history_xyz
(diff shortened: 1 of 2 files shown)


nicktobey:
Can you add some more detail to the PR description describing how `DoltHistory` not implementing `sql.PrimaryKeyTable` led to a panic? I'm a bit worried that this was able to happen in the first place. Standard operations on a keyless table should not cause panics.

max-hoffman (approved):
lgtm, thanks for the extra testing

PR #6390, opened 2023-07-25
Adds new `--profile` global arg

stephkyou:
Adds support for new `--profile` global arg which can group and apply a specified set of global args for the current command. Includes a new `--profile` flag and a new CLI command `dolt profile` to manage and list user profiles.


integration-tests/bats/profile.bats
```diff
@@ -0,0 +1,56 @@
...
+  # some, you need to ensure the repo is clean, and you should run `dolt reset --hard` at the beginning of the test.
+  cd ..
+}
+
+setup() {
+    setup_no_dolt_init
+    make_repo defaultDB
+    make_repo altDB
+
+    dolt config --global --add profile '{"defaultTest": {"use-db":"defaultDB"}}'
```
macneale4, line 33: This ends up being a string within a JSON object which is a serialized json object, right? That's kinda gross.

I tested putting an object into config_global.json, and `dolt config` panics attempting to marshal that data into a string. Effective makes it impossible to run dolt, so that's a wrinkle in this plan.

go/cmd/dolt/commands/profile.go
```diff
@@ -223,6 +235,24 @@ func prettyPrintProfile(profileName string, profile Profile) {
...
+func setGlobalConfigPermissions(dEnv *env.DoltEnv) error {
+	homeDir, err := env.GetCurrentUserHomeDir()
+	if err != nil {
+		return errhand.BuildDError("error: failed to get home directory: %s", err).Build()
+	}
+	path, err := dEnv.FS.Abs(filepath.Join(homeDir, dbfactory.DoltDir, env.GlobalConfigFile))
+	if err != nil {
+		return errhand.BuildDError("error: failed to get global config path: %s", err).Build()
+	}
+	err = os.Chmod(path, 0600)
```
macneale4, line 248: Did you verify that this works? Not clear why the sql-server.lock file permissions are wrong even though we do the same thing there

integration-tests/bats/profile.bats
```diff
@@ -0,0 +1,315 @@
...
+
+@test "profile: additional flag gets used" {
+    cd altDB
+    dolt sql -q "create table test (pk int primary key)"
+    dolt sql -q "insert into test values (999)"
+    dolt add test
+    dolt commit -m "insert initial value into test"
+    cd ..
+
+    dolt profile add --user dolt --password "" userProfile
```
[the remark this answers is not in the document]
macneale4: To take this a little further into the experience I would expect as a user.

`dolt --user pat sql` -  will prompt for a password because none was specified.

So if you create a profile:

`dolt profile add --user pat newprofile`

then 

`dolt --profile newprofile sql` should prompt for a password.

I believe it's impossible to do this with the current mechanism. You may need additional flags in the serialized object to ensure you have all the information you need.

integration-tests/bats/profile.bats
```diff
@@ -0,0 +1,315 @@
...
+    run dolt profile add --use-db altDB altTest
+    [ "$status" -eq 0 ] || false
+
+    run dolt profile
+    [ "$status" -eq 0 ] || false
+    [[ "$output" =~ "altTest" ]] || false
+    [[ "$output" =~ "altDB" ]] || false
+}
+
+@test "profile: dolt profile add overwrites an existing profile" {
```
macneale4, line 126: I believe we shouldn't allow this. Let's just throw an error if they attempt to do this. They can delete and create if they really want to overwrite.

go/cmd/dolt/dolt.go
```diff
@@ -420,8 +424,24 @@ func runMain() int {
...
+		return 1
+	}
 
+	globalConfig, ok := dEnv.Config.GetConfig(env.GlobalConfig)
+	if !ok {
+		cli.PrintErrln(color.RedString("Failed to get global config"))
+		return 1
+	}
+
+	apr, remainingArgs, err := parseGlobalArgsAndProfile(globalConfig, args)
```
macneale4, line 444: This is going to be invoked for all commands - even commands which don't support global arguments yet. When you have a default, I believe this will make using some commands impossible because it will insert arguments into the command line when they shouldn't be there. You should add a test for this. Test that `dolt log` works when there is a default configuration

go/cmd/dolt/dolt.go
```diff
@@ -742,6 +752,80 @@ func buildGlobalArgs() *argparser.ArgParser {
...
+			args = append([]string{"--profile", "default"}, args...)
+			apr, remaining, err = globalArgParser.ParseGlobalArgs(args)
+			if err != nil {
+				return nil, nil, err
+			}
+			profileName, hasProfile = apr.GetValue(commands.ProfileFlag)
+		}
+	}
+
+	if hasProfile {
```
macneale4, line 791: } else {

macneale4 (changes requested):

go/cmd/dolt/commands/profile.go
```diff
@@ -0,0 +1,291 @@
...
+	profileName := strings.TrimSpace(apr.Arg(1))
+
+	p := newProfile(apr)
+	profStr := p.String()
+
+	cfg, ok := dEnv.Config.GetConfig(env.GlobalConfig)
+	if !ok {
+		return errhand.BuildDError("error: failed to get global config").Build()
+	}
+	//TODO: enable config to retrieve json objects instead of just strings
```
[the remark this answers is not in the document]
stephkyou: The `ReadableConfig` and `WritableConfig` interfaces only have `GetString`/`SetString`. There would be a 5-6 implementations that would need to be updated if we added another method. It should be doable just a bit of time.

go/cmd/dolt/dolt.go
stephkyou: The logic is a little messy here but I check again for `hasProfile` because I re-parse the args with the `--profile default` flag added. The other option was to copy the code in `hasProfile` after adding the default profile. Which do you think makes more sense?

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -321,6 +321,7 @@ func CreateProfileArgParser() *argparser.ArgParser {
 	ap.SupportsString("privilege-file", "", "privilege-file", "DePath to a file to load and store users and grants.")
 	ap.SupportsString("branch-control-file", "", "branch-control-file", "Path to a file to load and store branch control permissions.")
 	ap.SupportsString("use-db", "", "use-db", "The name of the database to use when executing SQL queries.")
+	ap.SupportsString("profile", "", "profile", "The name of the profile to use when executing SQL queries. Run `dolt profile --help` for more information.")
```
macneale4, line 324: This options should only be available as a global argument. Wouldn't make sense to do the following:

`dolt profile --host foo --profile bar newprofile`

Add a boolean flag to the method to toggle this.

macneale4: Also, make the profile argument be the first one.

go/cmd/dolt/commands/profile.go
```diff
@@ -144,15 +143,27 @@ func addProfile(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) errhand.Verbo
 			return nil
 		}
 	}
+	profilesJSON, err := DecodeProfile(encodedProfiles)
+	if err != nil {
+		return errhand.BuildDError("error: failed to decode profiles, %s", err).Build()
+	}
+	cli.Println(profilesJSON)
```
macneale4, line 150: Remove!

go/cmd/dolt/commands/profile.go
```diff
@@ -172,13 +183,17 @@ func removeProfile(dEnv *env.DoltEnv, apr *argparser.ArgParseResults) errhand.Ve
...
 	if err != nil {
 		if err == config.ErrConfigParamNotFound {
 			return errhand.BuildDError("error: no existing profiles").Build()
 		}
 		return errhand.BuildDError("error: failed to get profiles, %s", err).Build()
 	}
+	profilesJSON, err := DecodeProfile(encodedProfiles)
+	if err != nil {
+		return errhand.BuildDError("error: failed to decode profiles, %s", err).Build()
+	}
```
macneale4: This code id duplicated twice. Put it in a helper function.

integration-tests/bats/profile.bats
```diff
@@ -318,11 +324,38 @@ teardown() {
...
+
+    run dolt log
+    [ "$status" -eq 0 ] || false
+    [[ "$output" =~ "insert initial value into test" ]] || false
+}
+
+@test "profile: profile with user but not password waits for password prompt" {
+    dolt profile add --use-db defaultDB -u "steph" defaultTest
+    run dolt --profile defaultTest sql -q "select * from table1" <<< ""
+    echo "$output"
```
macneale4, line 352: Remove echo.

This is a clever test :-)

macneale4 (changes requested):
Getting close! Pretty minor changes requested

go/cmd/dolt/commands/profile.go
```diff
@@ -51,10 +52,11 @@ Remove the profile named {{.LessThan}}name{{.GreaterThan}}.`,
...
 const (
-	addProfileId        = "add"
-	removeProfileId     = "remove"
-	GlobalCfgProfileKey = "profile"
-	DefaultProfileName  = "default"
+	addProfileId          = "add"
+	removeProfileId       = "remove"
+	GlobalCfgProfileKey   = "profile"
+	DefaultProfileName    = "default"
+	defaultProfileWarning = "Default profile has been added. All dolt commands taking global arguments will use this default profile until it is removed.\nWARNING: This will alter the behavior of command which specify no `--profile`.\nIf you are using dolt in contexts where you expect a `.dolt` directory to be accessed, the default profile will be used instead."
```
macneale4, line 59:
```suggestion
	defaultProfileWarning = "Default profile has been added. All dolt commands taking global arguments will use this default profile until it is removed.\nWARNING: This will alter the behavior of commands which specify no `--profile`.\nIf you are using dolt in contexts where you expect a `.dolt` directory to be accessed, the default profile will be used instead."
```

macneale4 (approved):
Single character change - well I guess two since you need to update the test :-)

```
         /|\
       /__| )
     /____| ))
   /______| )))
 /________|  )))
         _|____))
 \======| o o /
~~~~~~~~~~~~~~~~~~~
```

PR #6398, opened 2023-07-26
Implemented Full-Text indexes

Hydrocharged:
Technically functioning, although not the most ergonomic implementation. I need add a bit of user convenience and additional checking, such as removing the additional tables from `dolt status`, automatically adding the additional tables when the parent is added via `dolt add`, adding a user-facing write barrier on the tables, etc. Also need quite a few more tests, but the basics are in via the engine tests and very, very few bats tests.

I'm definitely going to expand the testing in the future, as well as implementing the aforementioned features and more. But for now, this should at least allow us to release unless there are major problems that I've not caught.


Hydrocharged:
Strangely, the failing Windows test is not consistent, and I cannot reproduce it locally. I'm going to chalk it up to flakiness for now, but I'll continue digging into it.

PR #6399, opened 2023-07-27
Bug fix: `dolt_diff_<tablename>` correctly shows historical values after narrowing types

fulghum:
The `dolt_diff_<tablename>` system table uses the schema @ HEAD of the table being diff'ed. If the schema of the table has changed over time and has narrowed column types (e.g. `varchar(100)` → `varchar(10)`), then some historical values may not be valid against the table's current schema, which causes an error to be returned and no result set. This PR fixes that by changing the schema of `dolt_diff_<tablename>` to widen all of the column types to their widest size in the same family (e.g. `varchar(100)` → `TEXT`).

Fixes https://github.com/dolthub/dolt/issues/6391


zachmu (approved):
LGTM!

PR #6404, opened 2023-07-27
Don't panic when attempting to publish metrics if config is unparsable

macneale4:
Currently if you have an unparsable config_global.json, we print a message say as much, then produce a panic attempting to use the config to print metrics. This just drops the attempt to produce the metrics.


stephkyou (approved):
lgtm

PR #6409, opened 2023-07-27
Bug fix for state working sets in replication

zachmu:
Fixes https://github.com/dolthub/dolt/issues/6400

### DIFF
--- a/go/libraries/doltcore/sqle/read_replica_database.go
+++ b/go/libraries/doltcore/sqle/read_replica_database.go
@@ -455,7 +455,7 @@ func pullBranches(
 
 						continue REFS
 					default:
-						ctx.GetLogger().Warnf("skipping replication for unhandled remote ref %s", remoteRef.Ref.String())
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/sqle/read_replica_database.go
```diff
@@ -302,6 +302,8 @@ func pullBranchesAndUpdateWorkingSet(
 	}
 
 	// update the current working set if necessary
+	// TODO: the current ref is wrong?
+	// Or we need to always update working sets?
```
fulghum, line 306: Is this TODO still needed?

fulghum (approved):
LGTM! 🚢

go/libraries/doltcore/sqle/read_replica_database.go
zachmu: Nope, that whole code block is obsolete, good catch!

PR #6438, opened 2023-08-02
Implemented --prune for fetch command procedure

zachmu:
`dolt fetch --prune` deletes any local remote refs that are not present on the remote being fetched. This is simpler than the equivalent git command, as it only operates on branches, not other kinds of refs.

Fixes https://github.com/dolthub/dolt/issues/6413


reltuk (approved):
LGTM!

PR #6465, opened 2023-08-07
`ANSI_QUOTES` support

fulghum:
Adds support for honoring [the `ANSI_QUOTES` SQL mode](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes). When this SQL mode is enabled, anything in double quotes is assumed to be an identifier, and not a string literal. 

Corresponding GMS PR: https://github.com/dolthub/go-mysql-server/pull/1915

Fixes: https://github.com/dolthub/dolt/issues/6305


go/libraries/doltcore/sqle/procedures_table.go
```diff
@@ -94,6 +101,88 @@ func DoltProceduresGetOrCreateTable(ctx *sql.Context, db Database) (*WritableDol
 	return tbl.(*WritableDoltTable), nil
 }
 
+// TODO: godocs
+// TODO: what's the best way to test this? backwards compat test suite?
+func migrateDoltProceduresSchema(ctx *sql.Context, db Database, oldTable *WritableDoltTable) (newTable *WritableDoltTable, rerr error) {
```
fulghum, line 106: I still need to come up with a way to test the migration code path. I saw some tests for dolt_schema migration yesterday, so I'll dig into those more next.

go/libraries/doltcore/doltdb/system_table.go
```diff
@@ -254,8 +254,10 @@ const (
 	// SchemasTablesExtraCol The name of the column that stores extra information about a schema element in the
 	// dolt_schemas table
 	SchemasTablesExtraCol = "extra"
-	//
-	SchemasTablesIndexName = "fragment_name"
+	// SchemasTablesSqlModeCol is the name of the column that stores the SQL_MODE string used when this fragment
+	// was originally defined. Mode settings, such as ANSI_QUOTES, are needed to correctly parse the fragment.
+	SchemasTablesSqlModeCol = "sql_mode"
```
zachmu, line 259: Put this at the end?

go/libraries/doltcore/sqle/procedures_table.go
```diff
@@ -94,6 +101,88 @@ func DoltProceduresGetOrCreateTable(ctx *sql.Context, db Database) (*WritableDol
...
+			return nil, err
+		}
+
+		newRow := make(sql.Row, ProceduresTableSchema().GetAllCols().Size())
+		newRow[0] = sqlRow[nameIdx]
+		newRow[1] = sqlRow[createStatementIdx]
+		newRow[2] = sqlRow[createdAtIdx]
+		newRow[3] = sqlRow[modifiedAtIdx]
+		if sqlModeIdx >= 0 {
+			newRow[4] = sqlRow[sqlModeIdx]
```
zachmu, line 142: Don't we want an explicit default for this column if the old values didn't have it?

go/libraries/doltcore/sqle/schema_table.go
```diff
@@ -265,12 +274,20 @@ func getSchemaFragmentsOfType(ctx *sql.Context, tbl *WritableDoltTable, fragType
 			continue
 		}
 
+		sqlModeString := ""
```
zachmu, line 277: Same comment, don't we want to fill in a non-ANSI value here?

zachmu (approved):
LGTM, but seems like this needs a feature version bump to be safe

go/libraries/doltcore/sqle/procedures_table.go
```diff
@@ -94,6 +101,88 @@ func DoltProceduresGetOrCreateTable(ctx *sql.Context, db Database) (*WritableDol
 	return tbl.(*WritableDoltTable), nil
 }
 
+// migrateDoltProceduresSchema migrates the dolt_procedures system table from a previous schema version to the current
+// schema version by adding any columns that do not exist.
+func migrateDoltProceduresSchema(ctx *sql.Context, db Database, oldTable *WritableDoltTable) (newTable *WritableDoltTable, rerr error) {
```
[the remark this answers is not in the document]
fulghum: I believe we do need explicit migration, otherwise the `dolt_procedures` table won't ever get the new column. `dolt_schemas` does also [have explicit migration code](https://github.com/dolthub/dolt/blob/75b80fd476a7b3d94fd33a627e6f7ae5ecd61ddd/go/libraries/doltcore/sqle/schema_table.go#L108).

go/libraries/doltcore/sqle/procedures_table.go
fulghum: Yeah, that's probably a good call. I was assuming the caller of this function would need to handle the default, but you're right that it'd be much cleaner to just apply it here.

PR #6468, opened 2023-08-07
updates global arg descriptions to be more verbose

stephkyou:
Restores more verbose descriptions for global arguments.


go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -315,16 +315,16 @@ func CreateGlobalArgParser(name string) *argparser.ArgParser {
 	if name == "dolt" {
 		ap.SupportsString("profile", "", "profile", "The name of the profile to use when executing SQL queries. Run `dolt profile --help` for more information.")
 	}
-	ap.SupportsString("user", "u", "user", "Defines the local superuser.")
-	ap.SupportsString("password", "p", "password", "Defines the password for the user.")
+	ap.SupportsString("user", "u", "user", "Defines the local superuser (defaults to `root`). If the specified user exists, will take on permissions of that user.")
+	ap.SupportsString("password", "p", "password", "Defines the password for the user. Defaults to empty string.")
```
macneale4, line 319: It will _only_ use the empty string by default when the user is `root` - Hopefully you can say that with not too many words.

macneale4 (approved):
One small change, then feel free to ship without another review.

PR #6470, opened 2023-08-08
Bug Fix: Error on merges that produce schema conflicts when `@@autocommit` is enabled

fulghum:
If `@@autocommit` is enabled, then a SQL commit will be created after every SQL statement. If a `dolt_merge()` operation results in data conflicts, then we return an error, revert the working set, and prevent the SQL commit. Schema conflicts should have the same behavior when `@@autocommit` is enabled, but they were not getting caught by our transaction logic. 

Fixes https://github.com/dolthub/dolt/issues/6469

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/transactions.go
+++ b/go/libraries/doltcore/sqle/dsess/transactions.go
@@ -565,8 +565,13 @@ func (tx *DoltTransaction) validateWorkingSetForCommit(ctx *sql.Context, working
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


tbantle22 (approved):
lgtm - thanks!

fulghum:
Labeling as `minor version bump` since this will cause some calls to `dolt_merge()` that previously completed without error (and reported schema conflicts) to return an error now when `@@autocommit` is enabled, especially since `@@autocommit` defaults to enabled.

PR #6474, opened 2023-08-08
`dolt_status` system table shows if merge is active when there are no conflicts

tbantle22:
**Breaking change**: Updates primary keys of `dolt_status` system table. We originally just had `table_name` in the primary key, but a table can already show up more than once in `dolt_status`. `status` and `staged` should also be part of the primary key.

This PR also adds a "merged" status row to the `dolt_status` system table when a schema conflict has been resolved but that table has not yet been committed, which fixes https://github.com/dolthub/dolt/issues/6473


fulghum (approved):
Looks great! Cool change. 🚢

PR #6475, opened 2023-08-08
updates dolt log to use sql backend

stephkyou:
Updates `dolt log` to use the appropriate sql engine to generate results.

### DIFF
--- a/go/cmd/dolt/commands/log.go
+++ b/go/cmd/dolt/commands/log.go
@@ -21,16 +21,14 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/dolthub/go-mysql-server/sql"
-	"github.com/gocraft/dbr/v2"
-	"github.com/gocraft/dbr/v2/dialect"
-
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
-	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
 	eventsapi "github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi/v1alpha1"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
 	"github.com/dolthub/dolt/go/store/util/outputpager"
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/gocraft/dbr/v2"
+	"github.com/gocraft/dbr/v2/dialect"
 )
 
 var logDocs = cli.CommandDocumentationContent{
@@ -105,7 +103,7 @@ func (cmd LogCmd) logWithLoggerFunc(ctx context.Context, commandStr string, args
 		defer closeFunc()
 	}
 
-	query, err := constructInterpolatedDoltLogQuery(apr)
+	query, err := constructInterpolatedDoltLogQuery(apr, queryist, sqlCtx)
 	if err != nil {
 		return handleErrAndExit(err)
 	}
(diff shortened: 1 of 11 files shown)


go/cmd/dolt/commands/log.go
```diff
@@ -283,56 +296,67 @@ func getCommitSpec(commit string) (*doltdb.CommitSpec, error) {
...
 		tagName = fmt.Sprintf("\033[33;1mtag: %s\033[0m", tagName) // tags names are bright yellow (33;1m)
-		cHashToRefs[t.Hash] = append(cHashToRefs[t.Hash], tagName)
+		cHashToRefs[hash.Parse(t[1].(string))] = append(cHashToRefs[hash.Parse(t[1].(string))], tagName)
 	}
 	return cHashToRefs, nil
 }
 
-func logCommits(ctx context.Context, dEnv *env.DoltEnv, opts *logOpts) int {
-	hashes := make([]hash.Hash, len(opts.commitSpecs))
+func logCommits(ctx context.Context, dEnv *env.DoltEnv, opts *logOpts, sqlResult []sql.Row, queryist cli.Queryist, sqlCtx *sql.Context) int {
```
macneale4, line 359: Drop the dEnv since you don't want it. Keeps you honest :-)

go/cmd/dolt/commands/log.go
```diff
@@ -399,34 +424,59 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, opts *logOpts) int {
...
+			Description:   row[4].(string),
+			UserTimestamp: 0,
+		}
+		/*meta, mErr := comm.GetCommitMeta(ctx)
 		if mErr != nil {
 			cli.PrintErrln("error: failed to get commit metadata")
 			return 1
+		}*/
+
+		parents, err := GetRowsForSql(queryist, sqlCtx, fmt.Sprintf("select parent_hash from dolt_commit_ancestors where commit_hash = %s", row[0]))
```
macneale4, line 449: ready row[0] into a variable above. Generally when you have an array at the top of the block and you have static indexes for for particular values, you want to copy them all into named values right away. When I read the lines 435 - 438, my first thought was 'where is row[0]`

go/cmd/dolt/commands/log.go
```diff
@@ -399,34 +424,59 @@ func logCommits(ctx context.Context, dEnv *env.DoltEnv, opts *logOpts) int {
...
+		}*/
+		cmHash := hash.Parse(row[0].(string))
 
 		commitsInfo = append(commitsInfo, logNode{
 			commitMeta:   meta,
 			commitHash:   cmHash,
 			parentHashes: pHashes,
 			branchNames:  cHashToRefs[cmHash],
-			isHead:       cmHash == *cwbHash})
+			isHead:       row[0] == headHash})
```
macneale4, line 480: double down on my suggestion above

go/cmd/dolt/commands/log.go
```diff
@@ -131,8 +154,87 @@ func (cmd LogCmd) logWithLoggerFunc(ctx context.Context, commandStr string, args
...
+// constructInterpolatedDoltLogQuery generates the sql query necessary to call the DOLT_LOG() function.
+// Also interpolates this query to prevent sql injection.
+func constructInterpolatedDoltLogQuery(apr *argparser.ArgParseResults) (string, error) {
+	var params []interface{}
+
+	var buffer bytes.Buffer
+	var first bool
+	first = true
+
+	if apr.Contains(cli.ParentsFlag) && apr.Contains(cli.DecorateFlag) {
```
macneale4, line 170: You can simplify this. For example, just alway send the --decorate flag, and always expect that column to exist. There are other examples of this too. Eg, if they don't specify a max number, then we should supply one.

stephkyou:
Closing to re-work `dolt_log` table function

PR #6478, opened 2023-08-09
[no-release-notes] Fixing missing "|| false" in BATS files

fulghum:
I ran into one of these missing ` || false` statements yesterday in another PR. It's confusing when the BATS tests pass locally and you can't repro a CI failure because of this quirk. I figured I'd do a quick audit to fix up any others. 

The command I used, was:
```
sed -E -i '' 's/\]\][[:space:]]*$/]] || false/' *.bats 
```


timsehn (approved):
Love it. Ship it.

PR #6487, opened 2023-08-10
feat: adding zsh plugin

iloveitaly:
This isn't perfect, but it's a starting point. What would be better:

* Create a separate repo for just this script (`dolthub/zsh`) for easier installation
* Change gen-zsh to output the script to stdout like `gh`, `op`, and other popular tools

You can use this via a `for` loop in zinit like:

```
zinit wait lucid for \
  'https://github.com/iloveitaly/dolt/blob/zsh-plugin/dolt.plugin.zsh'
```

### DIFF
--- a/dolt.plugin.zsh
+++ b/dolt.plugin.zsh
@@ -1,0 +1,13 @@
+if (( $+commands[dolt] )); then
+  # gen-zsh writes a file to the command line, and does not output the completion script like
+  # most other cli tools. For this reason, we write the file to the local directory and source it
+  local completion_path=${0:h}/_dolt
+
+  if [[ ! -e "$completion_path" ]]; then
+    # creates a completion file in the current directory
+    dolt gen-zsh --file="$completion_path"
+  fi
+
+  source "$completion_path"
+  compdef _dolt dolt
+fi


timsehn:
We have one of these.

https://www.dolthub.com/blog/2021-11-15-zsh-completions-with-subcommands/

timsehn:
It looks like there is a hidden command

dolt gen-zsh

timsehn:
I see. You know about that :-) This is a shell script to make it even easier :-)

timsehn:
@zachmu wrote the `gen-zsh` command. He should discuss this solution with you.

zachmu (commented):
Thanks for the contribution!

I wasn't aware that outputting completion scripts to stdout was more common. Would it be better to change the gen-zsh command to do that?

iloveitaly:
@zachmu I think so! Most popular cli tools do that (at least, gh cli and 1password)

zachmu:
I made that change here:

https://github.com/dolthub/dolt/pull/6492

Want to add to that PR so your plugin works correctly? Your zsh skills are beyond mine :)

zachmu:
I went ahead and merged that in, so you can simplify your zsh plugin to read from STDOUT at your convenience.

zachmu:
It looks like this PR is now merged because I merged its commit as part of my PR, so feel free to open another one to get rid of the `--file` flag in the zsh plugin.

PR #6490, opened 2023-08-10
[no-release-notes] gms bump for name res

max-hoffman:


max-hoffman:
#benchmark

max-hoffman:
#import-benchmark

max-hoffman:
#systab-benchmark

max-hoffman:
#merge-benchmark

max-hoffman:
#benchmark

go/libraries/doltcore/mvdata/engine_table_writer.go
```diff
@@ -265,52 +265,70 @@ func (s *SqlEngineTableWriter) createOrEmptyTableIfNeeded() error {
 
 // createTable creates a table.
 func (s *SqlEngineTableWriter) createTable() error {
-	cr := plan.NewCreateTable(sql.UnresolvedDatabase(s.database), s.tableName, false, false, &plan.TableSpec{Schema: s.tableSchema})
-	analyzed, err := s.se.Analyze(s.sqlCtx, cr)
-	if err != nil {
-		return err
+	// TODO don't use internal interfaces to do this, we had to have a sql.Schema somewhere
+	// upstream to make the dolt schema
+	sqlCols := make([]string, len(s.tableSchema.Schema))
```
zachmu, line 270: Surely there's a library function that can generate a create table statement

If we need an engine to do that fine, but this likely to break as we add features

go/libraries/doltcore/mvdata/engine_table_writer.go
```diff
@@ -265,52 +265,70 @@ func (s *SqlEngineTableWriter) createOrEmptyTableIfNeeded() error {
...
+	createTable := sql.GenerateCreateTableStatement(s.tableName, sqlCols, sql.CharacterSet_utf8.String(), sql.Collation_Default.String())
+	sch, iter, err := s.se.Query(s.sqlCtx, createTable)
+	if err != nil {
+		return err
 	}
+	_, err = sql.RowIterToRows(s.sqlCtx, sch, iter)
+	return err
 }
 
 // createInsertImportNode creates the relevant/analyzed insert node given the import option. This insert node is wrapped
```
zachmu, line 293: method doc out of date

go/libraries/doltcore/sqle/sqlddl_test.go
```diff
@@ -1091,6 +1103,30 @@ func TestParseCreateTableStatement(t *testing.T) {
 	}
 }
 
+func newTestEngine(ctx context.Context, dEnv *env.DoltEnv) (*gms.Engine, *sql.Context) {
```
zachmu, line 1106: This must be defined somewhere else right?

zachmu (approved):
LGTM, only minor comments

go/libraries/doltcore/mvdata/data_loc.go
```diff
@@ -89,7 +89,7 @@ type DataLocation interface {
 	Exists(ctx context.Context, root *doltdb.RootValue, fs filesys.ReadableFS) (bool, error)
 
 	// NewReader creates a TableReadCloser for the DataLocation
-	NewReader(ctx context.Context, root *doltdb.RootValue, fs filesys.ReadableFS, opts interface{}) (rdCl table.SqlRowReader, sorted bool, err error)
+	NewReader(ctx context.Context, dEnv *env.DoltEnv, opts interface{}) (rdCl table.SqlRowReader, sorted bool, err error)
```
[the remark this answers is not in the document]
max-hoffman: yeah i agree, but the alternative was adding databases and config to this interface. Passing 4 fields from env felt dumb, If you want a one-off table reader it might make sense to just have the dEnv

go/libraries/doltcore/mvdata/engine_table_writer.go
max-hoffman: I am concerned about all of the internal engine interface uses, but it didn't feel right to fix all of those in this PR. This change in particular seems pretty unnecessary because we do some BS to convert sql.Schema-> dolt Schema and then this hack to go backwards again.

go/libraries/doltcore/mvdata/engine_table_writer.go
max-hoffman: I think the doc is still accurate? We are still making an analyzed insert node, it's just using a different interface

go/libraries/doltcore/sqle/sqlddl_test.go
max-hoffman: kind of, the other one does the reverse, constructing an engine from a database. This is an empty dolt directory that needs a new DB. I probably could put more effort into figuring out how to manually create a dolt database and join it with the env, seemed like an equal amount of extra work

PR #6496, opened 2023-08-11
Bug fix for schema merge

zachmu:
The bug is triggered when merging the schema change here back into main:

https://www.dolthub.com/repositories/dolthub/test-idearoom/data/bkelley%2Fsc-30879%2Fremove-preventseed-column-from-the-client/branchMetadata

The trigger is that the schema change drops a column, and when we copy the table data into the new schema, we end up putting a varchar value into a text column. This panic when we run `mutator.Map()`, because it's expecting an address hash in that place in the schema where we placed character data. My fix is to introduce a new mutator method that uses the new schema earlier in the process.

But there are a couple things that worry me about this:

1) I can't get a simple repro in schema_merge_test based on my understanding of the issue, and 
2) My fix breaks a couple existing tests

This should point somebody in the right direction hopefully.


go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -88,12 +88,12 @@ func mergeProllyTable(ctx context.Context, tm *TableMerger, mergedSch schema.Sch
 		// After we migrate the data on the left-side to the new, merged schema, we reset
 		// the left mapping to an identity mapping, since it's a direct mapping now.
 		valueMerger.leftMapping = val.NewIdentityOrdinalMapping(len(valueMerger.leftMapping))
-	}
-
-	// After we've migrated the existing data to the new schema, it's safe for us to update the schema on the table
-	mergeTbl, err = tm.leftTbl.UpdateSchema(sqlCtx, mergedSch)
-	if err != nil {
-		return nil, nil, err
+	} else {
```
fulghum, line 91: I don't think this should be in an else block... it seems like the first part of this function would migrate the table data, but then it wouldn't go into the else block to actually update the schema metadata of the table, so the schema and data would be out of sync.

PR #6506, opened 2023-08-15
Drop the aws flags from the `dolt_remote()` procedure

macneale4:
Apparently there are no tests for these flags because the tests don't blow up if I remove them. We still allow the flags in the CLI context though.


reltuk (approved):
LGTM!

Issue #6273, opened 2023-07-04
Create flag for specifying the initial branch for a sql client.


nicktobey:
Something like `dolt --branch=foo sql ...`. Using this flag would change the initial branch for the `sql` command, as well as any commands that make sql queries under the hood.

Currently, the only way to do this is with the `dolt sql-client` command, which accepts a db/revision pair for its --use-db parameter.

Implementing this is mostly a convenience feature that eliminates the need for `use` or `call dolt_checkout` commands in the sql, but is also important for the rare case where `repo_state.json` is in a bad state that prevents the server from starting. (We have a regression test for this in `deleted_branches.bats`, but that test depends on the current implementation of `dolt checkout` which we are migrating away from.

Nithinps021:
@nicktobey The way I taught to implement this is by appending `use database/branch` or calling `dolt_checkout("branch")` at the start of the user `input query`. I don't know whether its clean way or not.

macneale4:
We do have branch support in the db name, so you can do the following:

`dolt --use-db=database_foo/branch_bar status`

A separate flag would be a nice addition though.

PR #6507, opened 2023-08-15
Added a `--branch` parameter to Dolt global arguments on the command line.

Nithinps021:
Screenshot 2023-08-15 at 12 05 23 PM


timsehn:
Thanks for the contribution @macneale4 will look at this since he is the master of global args.

timsehn:
Addresses #6273

macneale4:
Hi @Nithinps021 - Thanks for the PR! The addition of the --branch flag is something we considered, but hadn't pulled the trigger yet because we do have support for it in the specification of the DB name. If you want start your session on the `dev` branch, you would specify `--use-db=database_name/dev`

If you would like to continue and add the flag, I suggest you leverage that code path which is earlier in processing. Essentially adding it to the `useDB` variable here: https://github.com/dolthub/dolt/blob/559cbcbefc261e0f9119c28bb4893d4dcaaf1135/go/cmd/dolt/dolt.go#L611

The benefit of doing it this way is that all global argument enabled commands can take advantage of it.

Thanks again!

timsehn:
Closing this pending the rework Neil suggested.

PR #6511, opened 2023-08-15
Bug fix: panic in schema merge when schema contains BLOB columns

fulghum:
When processing a schema merge, we need to update existing, stored data in the primary index. We weren't updating the value descriptor in the prolly map to the new version from the changed schema, which started causing a problem with `BLOB`/`TEXT` types since validation checked that the value stored in those fields was a valid address pointing outside of the table to the real data. 

Related to https://github.com/dolthub/dolt/pull/6496 – the reason that PR wasn't able to cleanly repro the exact failure was because the merge was getting short circuited, since the right and left sides weren't both making changes, so essentially the merge was getting fast-forwarded instead.

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
@@ -4518,6 +4518,34 @@ var ThreeWayMergeWithSchemaChangeTestScripts = []MergeScriptTest{
 		},
 	},
 	{
+		// Repro for issue in: https://github.com/dolthub/dolt/pull/6496
+		Name: "drop a column, schema contains BLOB/AddrEnc columns",
+		AncSetUpScript: []string{
+			"set autocommit = 0;",
+			"CREATE table t (pk int primary key, col1 int, col2 text, col3 varchar(10));",
+			"INSERT into t values (1, 10, 'a', 'b'), (2, 20, 'c', 'd');",
+		},
+		RightSetUpScript: []string{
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4517,6 +4517,34 @@ var ThreeWayMergeWithSchemaChangeTestScripts = []MergeScriptTest{
...
+		Name: "drop a column, schema contains BLOB/AddrEnc columns",
+		AncSetUpScript: []string{
+			"set autocommit = 0;",
+			"CREATE table t (pk int primary key, col1 int, col2 text, col3 varchar(10));",
+			"INSERT into t values (1, 10, 'a', 'b'), (2, 20, 'c', 'd');",
+		},
+		RightSetUpScript: []string{
+			"INSERT into t values (300, 30, 'e', 'f');",
+		},
+		LeftSetUpScript: []string{
```
zachmu, line 4531: Maybe I'm confused, but wasn't the bug triggered by a drop on the right-side of the merge?

zachmu (approved):
LGTM!

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
fulghum: good eye! the test runner runs each schema merge test in both directions. it's been pretty helpful in catching several bugs. you're right that this bug only shows up in one direction.

PR #6513, opened 2023-08-15
dolt sql-server: Add a system_variables: key to config.yaml, which allows setting global system variables.

reltuk:
For example, this can be used to set `secure_file_priv` when the server starts up. That variable is non-dynamic, so it cannot be set with `SET @@GLOBAL...`


zachmu (approved):
LGTM!

PR #6514, opened 2023-08-16
adding `query-diff` cli command

JCOR11599:
Companion PR: https://github.com/dolthub/go-mysql-server/pull/1955


macneale4:
This seems like a lot of heavy lifting on the CLI side. Is there a reason we aren't doing the work in a table function? Then it would be available to SQL only users. Adding a CLI wrapper around it would be easy.

JCOR11599:
I agree, I'm currently exploring ways to implement this as a table function or a stored procedure.
One challenge I'm facing is how to parse, analyze, and execute the queries from within a table function/stored procedure.

max-hoffman (commented):
James and I talked about this yesterday, I think this is fine I just am worried about how limited it is in the broad search space of query diff. As is this will only return results for two queries that are identical and have the same underlying table, but different data versions. So like a slightly expanded form of diff that lets us do aggregation and join diffs.

Some of the other things in the space of query diff:
- queries are the same, schemas are different
- queries are the same, but aggregation or join precludes using dolt diff
- queries are different (schemas are different is arguably a subset of this)

It is almost like we are introducing a third diff dimension: schema->query plan->data. We have to diff the two query plans after diffing the schema in the general case to determine whether there are smarter ways to diff the data. Even if the queries are the same, we still have to walk the AST or plan to see whether materializing nodes prevent using dolt diff on underlying tables.

And re: bisection/temporary table materialization, both of those require reading the entire tables, which makes them more expensive than the naive algo.

Other thoughts:
- None of the current logic is really Dolt-specific. We could put all of this into GMS, have the schema comparison only test column shape+types (ignore table name), and have this be a generic query differ.
- The Dolt side could re-use our schema merge logic to go through the left/right schemas, match columns on name/type or tag, and then interpolate new query (strings or nodes) that will be schema-comparable. This would let us finesse queries to handle schema diff, with the generic query_diff living and testing independently in GMS.

I am not against merging what we have. It is just unclear to me what customers expect from this feature.

go/libraries/doltcore/sqle/dolt_query_diff_table_function.go
```diff
@@ -0,0 +1,319 @@
...
+func (tf *QueryDiffTableFunction) WithExpressions(expression ...sql.Expression) (sql.Node, error) {
+	if len(expression) != 2 {
+		return nil, sql.ErrInvalidArgumentNumber.New(tf.Name(), "2", len(expression))
+	}
+
+	for _, expr := range expression {
+		if !expr.Resolved() {
+			return nil, ErrInvalidNonLiteralArgument.New(tf.Name(), expr.String())
+		}
+		// prepared statements resolve functions beforehand, so above check fails
```
max-hoffman, line 142: is this new or copied? I think expressions are going to be resolved by default now, even if not literal

go/libraries/doltcore/sqle/dolt_query_diff_table_function.go
```diff
@@ -0,0 +1,319 @@
...
+// CheckPrivileges implements the sql.Node interface
+func (tf *QueryDiffTableFunction) CheckPrivileges(ctx *sql.Context, opChecker sql.PrivilegedOperationChecker) bool {
+	return opChecker.UserHasPrivileges(ctx, sql.NewPrivilegedOperation(tf.database.Name(), "", "", sql.PrivilegeType_Select))
+}
+
+// Schema implements the sql.Node interface
+func (tf *QueryDiffTableFunction) Schema() sql.Schema {
+	if !tf.Resolved() {
+		return nil
+	}
```
max-hoffman, line 299: i don't think this should be possible?

go/libraries/doltcore/sqle/dolt_query_diff_table_function.go
```diff
@@ -0,0 +1,319 @@
...
+	if tf.sqlSch == nil {
```
max-hoffman, line 300: also seems like we should have errored upstream to reach this

max-hoffman (approved):
LGTM as a resurrection of the old query diff in a SQL context. We'll try to review options with people when you get back, or when a customer asks us for more.

PR #6517, opened 2023-08-16
Automatically merge schema changes that widen `VARCHAR` fields to `TEXT`

fulghum:
This enhancement allows more schema changes to merge cleanly (instead of being reported as a schema conflict) when we can determine that a type is being widened and all existing data is guaranteed to be safe to convert. This includes:
- `VARCHAR` → `TINYTEXT`, `TEXT`, `MEDIUMTEXT`, `LONGTEXT` 
- `CHAR` → `TINYTEXT`, `TEXT`, `MEDIUMTEXT`, `LONGTEXT`
-  `BINARY` → `TINYBLOB`, `BLOB`, `MEDIUMBLOB`, `LONGBLOB`
-  `VARBINARY` → `TINYBLOB`, `BLOB`, `MEDIUMBLOB`, `LONGBLOB`

Trying to merge a change to a `VARCHAR` column larger than 255 characters to make it a `TINYTEXT` column will still generate a schema conflict, since `TINYTEXT` has a maximum size of 255 characters. 

One of the more interesting conversions is converting to `VARBINARY(N)` or `TINYTEXT` to `BINARY(N)`. MySQL right-pads all `BINARY(N)` values with null bytes, up to `N`, which requires us to call `Type.Convert` as part of remapping values in a schema merge, and also to rebuild secondary indexes in some cases. 

Fixes: https://github.com/dolthub/dolt/issues/6442


go/libraries/doltcore/merge/type_compatibility.go
```diff
@@ -29,142 +30,260 @@ import (
...
+	// convert the existing rows into the new schema.
+	//
 	// For the DOLT storage format, very few cases (outside of the types being exactly identical) are considered
-	// compatible, but we can widen them over time to optimize additional type changes that can be automatically
-	// merged. The older LD_1 storage format, has a more forgiving storage layout, so more type changes are
-	// considered compatible, generally as long as they are in the same type family/kind.
-	IsTypeChangeCompatible(from, to typeinfo.TypeInfo) bool
+	// compatible without requiring a full table rewrite. The older LD_1 storage format, has a more forgiving storage
+	// layout, so more type changes are considered compatible, generally as long as they are in the same type family/kind.
+	IsTypeChangeCompatible(from, to typeinfo.TypeInfo) (bool, bool)
```
zachmu, line 41: Name the out params?

go/libraries/doltcore/merge/type_compatibility.go
```diff
@@ -29,142 +30,260 @@ import (
...
+// typeChangeHandler has the logic to determine if a specific change from one type to another is a compatible
+// type change that can be automatically converted, possibly also requiring a full table rewrite to update data.
+type typeChangeHandler interface {
+	// canHandle returns true if this typeChangeHandler is able to check type change compatibility for the specified
+	// |from| and |to| sql.Type instances.
+	canHandle(fromSqlType, toSqlType sql.Type) bool
+	// isCompatible returns two booleans – the first boolean response parameter indicates if a type change from
+	// |fromType| to |toType| is compatible and safe to perform automatically. The second boolean response parameter
+	// indicates if the type conversion requires a full table rewrite.
+	isCompatible(fromSqlType, toSqlType sql.Type) (bool, bool)
```
zachmu, line 140: name out params?

go/libraries/doltcore/merge/type_compatibility.go
```diff
@@ -29,142 +30,260 @@ import (
...
+		toStringType.Collation() == fromStringType.Collation()
+
+	rewriteRequired := false
+	if compatible {
+		// Because inline string types (e.g. VARCHAR, CHAR) have the same encoding, the only time
+		// a table rewrite is required is when moving between an inline string type and an
+		// out-of-band string type (e.g. TEXT).
+		fromTypeOutOfBand := outOfBandType(fromSqlType)
+		toTypeOutOfBand := outOfBandType(toSqlType)
+		if !fromTypeOutOfBand && toTypeOutOfBand {
```
zachmu, line 184: Isn't the same true in the other direction as well (TEXT -> VARCHAR)

Should be `fromTypeOutOfBand != toTypeOutOfBand`?

zachmu: Test for this case?

go/libraries/doltcore/merge/type_compatibility_test.go
```diff
@@ -128,69 +144,161 @@ func TestDoltIsTypeChangeCompatible(t *testing.T) {
...
+			to:         varchar10,
+			compatible: false,
+		}, {
+			name:       "type widening: varchar to TEXT",
+			from:       varchar10,
+			to:         text,
+			compatible: true,
+			rewrite:    true,
+		}, {
+			name:       "type narrowing: TEXT to varchar",
```
zachmu, line 222: Seems wrong, need to rewrite the table in this case

end-to-end test would be nice too

zachmu (approved):
Like the new abstraction, very clean

My one major comment is that it seems like TEXT -> VARCHAR needs a table rewrite, no?

go/libraries/doltcore/merge/type_compatibility.go
fulghum: Great catch! I was originally assuming that any TEXT to VARCHAR conversion would error with a schema conflict and not get this far, but TINYTEXT to VARCHAR(300) is totally valid, and does require a table rewrite. Fixed in commit 
133c88b4568152da03dadb678bdcf2310522dcaa and added more tests for this.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1266,12 +1272,56 @@ func remapTupleWithColumnDefaults(ctx *sql.Context, keyTuple, valueTuple val.Tup
...
+	if newTypeInfo.Equals(previousTypeInfo) {
+		return value, nil
+	}
+
+	newValue, inRange, err := newTypeInfo.ToSqlType().Convert(value)
+	if err != nil {
+		return nil, err
+	}
+	if !inRange {
+		return nil, fmt.Errorf("out of range conversion for value %v to type %s", value, newTypeInfo.String())
```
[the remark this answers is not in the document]
fulghum: I'll think about this one a little more... I don't see a great way to trigger this path currently since we shouldn't ever execute this code unless we've already verified that the type we're moving to is larger than the type we're coming from and `StringType.Convert` will only return out of range when we give it a value larger than the type's width.

PR #6518, opened 2023-08-16
Support datetime and timestamp precision

zachmu:
Fixes https://github.com/dolthub/dolt/issues/6503

In addition to datetime precision support, this PR also significantly simplifies the process of deserializing and caching schemas. Previously on load, we always did two round trips to dereference table's schema hash, then re-hash the object, then serialized it into an intermediate in-memory format and back to the schema.Schema, before returning it. The new code eliminates this redundancy.


zachmu:
Companion PR: https://github.com/dolthub/go-mysql-server/pull/1950

go/libraries/doltcore/schema/typeinfo/typeinfo.go
```diff
@@ -252,18 +275,47 @@ func FromSqlType(sqlType sql.Type) (TypeInfo, error) {
...
+			return nil, err
+		}
 		return &setType{setSQLType}, nil
 	default:
 		return nil, fmt.Errorf(`no type info can be created from SQL base type "%v"`, sqlType.String())
 	}
 }
 
+func fillInSetCollationWithDefault(st sql.SetType) (sql.SetType, error) {
+	if st.Collation() == sql.Collation_Unspecified {
```
Hydrocharged, line 299: Far better way to do this:
```go
func someFuncName(typ sql.Type) (sql.Type, error) {
	if collationType, ok := typ.(sql.TypeWithCollation); ok && collationType.Collation() == sql.Collation_Unspecified {
		return collationType.WithNewCollation(sql.Collation_Default)
	}
	return typ, nil
}
```

Hydrocharged: Instead of making this a function, this logic could also go at the top of `FromSqlType` and it'll automatically apply to all types. That `queryType := sqlType.Type()` line could do this instead. Then, any added types that handle collations won't have a coverage gap.

Hydrocharged (approved):
LGTM!

go/libraries/doltcore/schema/typeinfo/typeinfo.go
zachmu: Good idea

PR #6522, opened 2023-08-17
update `dolt_log` table function to support more features

stephkyou:
Updates the `dolt_log` table function to support more features, e.g. accepting more than 2 revisions and accepting table names. This change is motivated by the ongoing CLI migration (https://github.com/dolthub/dolt/issues/3922) as the updated table function will help migrate the `dolt log` command.


go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -266,106 +263,75 @@ func (ltf *LogTableFunction) WithExpressions(expression ...sql.Expression) (sql.
...
-		return nil, err
-	}
-
-	return &newLtf, nil
-}
 
-func (ltf *LogTableFunction) invalidArgDetailsErr(expr sql.Expression, reason string) *errors.Error {
-	return sql.ErrInvalidArgumentDetails.New(ltf.Name(), fmt.Sprintf("%s - %s", expr.String(), reason))
-}
+		if !strings.Contains(exStr, "--") && !(i > 0 && strings.Contains(expression[i-1].String(), "--")) {
```
macneale4, line 274: Has argument parsing not happened at this point?

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3225,36 +3225,67 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
+				},
 			},
 			{
-				Query:    "SELECT count(*) from dolt_log(@Commit2);",
-				Expected: []sql.Row{{4}},
+				Query: "SELECT message from dolt_log(@Commit2);",
+				Expected: []sql.Row{
+					{"inserting into t"},
+					{"creating table t"},
+					{"Initialize data repository"},
```
macneale4, line 3255: why are the counts changing? These are for existing tests, so I'd expect them to stay the same

stephkyou: I updated the test harness to setting a setup commit since we're only doing reads in this test. Typically, there would be a checkpoint commit set with message "checkpoint enginetest database <db>" so that tests can reset to this point between query assertions. This would also show up on dolt log calls so I think it's less confusing to just skip setting it.

stephkyou: L2049 in `dolt_engine_test.go`

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3683,6 +3683,58 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
+				Query: "select message from dolt_log('--tables', 'test', 'test2');",
+				Expected: []sql.Row{
+					{"merged test-branch"},
+					{"inserted 1 into test"},
+					{"created table test2"},
+					{"created table test"},
+				},
+			},
+			{
+				Query: "select message from dolt_log('test-branch', '--tables', 'test');",
```
macneale4, line 3732: I'd expect the options to come before revision numbers. Prior to your change, was there any restriction there? I'd kind of expect it to be like the CLI - but if it's not that's ok.

stephkyou: I think it shouldn't matter either way, both should parse fine in SQL or on the CLI.

go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -267,104 +266,106 @@ func (ltf *LogTableFunction) WithExpressions(expression ...sql.Expression) (sql.
...
-			return sql.ErrInvalidArgumentDetails.New(ltf.Name(), fmt.Sprintf("%s - %s", ltf.notRevision, "--not revision cannot contain '^'"))
+		if strings.Contains(notRevStr, "^") {
+			return ltf.invalidArgDetailsErr("--not revision cannot contain '^'")
 		}
 	}
 
 	return nil
 }
 
+func mustExpressionsToString(ctx *sql.Context, expr []sql.Expression) ([]string, error) {
```
macneale4, line 329: gDoc

go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -267,104 +266,106 @@ func (ltf *LogTableFunction) WithExpressions(expression ...sql.Expression) (sql.
...
+			return nil, err
+		}
+
+		valStrs = append(valStrs, valStr)
+	}
+
+	return valStrs, nil
+}
+
+func expressionToString(ctx *sql.Context, expr sql.Expression) (string, error) {
```
macneale4, line 344: Do we have any tests which test actually expressions? All of our tests inputs are strings, so there is no evaluation actually being done. Please add a test which has a select or a function call in it as an argument to dolt_log()

macneale4 (changes requested):
Need more tests!

go/libraries/doltcore/sqle/dolt_log_table_function.go
macneale4: nm. discussed in person that @CommitA is evaled.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3543,63 +3696,63 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
 			{
-				Query:    "SELECT commit_hash = @MergeCommit, committer, email, message from dolt_log('main', '--min-parents', '2');",
-				Expected: []sql.Row{{true, "billy bob", "bigbillieb@fake.horse", "Merge branch 'branch2' into main"}},
+				Query:    "SELECT committer, email, message from dolt_log('main', '--min-parents', '2');",
+				Expected: []sql.Row{{"root", "root@localhost", "Merge branch 'branch2' into main"}},
 			},
 			{
 				Query:    "SELECT count(*) from dolt_log('main');",
-				Expected: []sql.Row{{6}},
+				Expected: []sql.Row{{5}},
```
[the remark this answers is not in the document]
stephkyou: Takes out the setup commit.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3543,63 +3696,63 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
 			},
 			{
 				Query:    "SELECT count(*) from dolt_log('main', '--min-parents', '1');", // Should show everything except first commit
-				Expected: []sql.Row{{5}},
+				Expected: []sql.Row{{4}},
```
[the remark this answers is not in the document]
stephkyou: Takes out the setup commit.

go/libraries/doltcore/sqle/dolt_log_table_function.go
```diff
@@ -623,78 +624,84 @@ func (ltf *LogTableFunction) NewDotDotLogTableFunctionRowIter(ctx *sql.Context,
...
+		h, cm, err = itr.child.Next(ctx)
 		if err != nil {
 			return nil, err
 		}
 
 		if itr.tableNames != nil {
-			if itr.prevCommit == nil {
-				itr.prevCommit = cm
-				itr.prevHash = h
+			if cm.NumParents() == 0 {
```
macneale4, line 637: Why continuing here? In the case where you weren't filtering on tables, that commit would be returned. I believe this will result in skipping a commit, but I may be reading this wrong.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3552,6 +3552,7 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
 			"call dolt_add('.')",
 			"call dolt_commit('-m', 'commit 1 MAIN')",
 			"call dolt_commit('--allow-empty', '-m', 'commit 2 MAIN')",
+			"call dolt_tag('tagM')",
```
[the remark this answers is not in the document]
macneale4: Reading this will be a lot easier if you have the short identifiers (1M, 2A, etc) in the commit description.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3644,16 +3661,26 @@ var LogTableFunctionScriptTests = []queries.ScriptTest{
...
 			"insert into test values (1)",
 			"call dolt_add('.')",
 			"call dolt_commit('-m', 'inserted 1 into test')",
 			"call dolt_merge('test-branch', '-m', 'merged test-branch')",
+			"drop table test3",
+			"call dolt_add('.')",
+			"call dolt_commit('-m', 'dropped table test3')",
+			"insert into test values (2)",
+			"call dolt_add('.')",
+			"call dolt_commit('-m', 'inserted 2 into test')",
```
macneale4, line 3677: You hit the "needs ascii art" threshold here

macneale4 (changes requested):
Getting close. I'd like to talk some more at the whiteboard about walking the graph.

macneale4 (approved):
```
              |    |    |
             )_)  )_)  )_)
            )___))___))___)\
           )____)____)_____)\\
         _____|____|____|____\\\__
---------\                   /---------
  ^^^^^ ^^^^^^^^^^^^^^^^^^^^^
    ^^^^      ^^^^     ^^^    ^^
         ^^^^      ^^^
```

PR #6523, opened 2023-08-17
Changing primary index to use the Rewriter function, instead of Mutate

fulghum:
Ensures that the tuple descriptors use the correct types, and don't do address validation on unnecessary data.


zachmu (approved):
LGTM

PR #6526, opened 2023-08-17
`dolt version` only prints the dolt version if there is no dolt database in the directory.

timsehn:
Remove "no valid database in this directory" from `dolt version` in the case of no dolt database found.


bheni (approved):
A good start

PR #6531, opened 2023-08-17
Empty Servers allow `dolt sql` connections without arguments

macneale4:
WIth this change, starting `sql-server` will result in creating a .dolt/sql-server.lock file in the --data-dir (CWD by default). This file will be used for two purposes:
1) Ensuring another `dolt sql-server` is not started with the same data dir.
2) Allowing the `dolt sql` command to connect to it. See related issue: https://github.com/dolthub/dolt/issues/6508

### DIFF
--- a/go/libraries/doltcore/env/environment.go
+++ b/go/libraries/doltcore/env/environment.go
@@ -1268,6 +1268,13 @@ func (dEnv *DoltEnv) Unlock() error {
 
 // WriteLockfile writes a lockfile encoding the pid of the calling process.
 func WriteLockfile(fs filesys.Filesys, lock *DBLock) error {
+	// if the DoltDir doesn't exist, create it.
+	doltDir, _ := fs.Abs(dbfactory.DoltDir)
+	err := fs.MkDirs(doltDir)
+	if err != nil {
+		return err
+	}
+
 	lockFile, _ := fs.Abs(filepath.Join(dbfactory.DoltDir, ServerLockFile))
 
 	portStr := strconv.Itoa(lock.Port)
@@ -1286,7 +1293,7 @@ func WriteLockfile(fs filesys.Filesys, lock *DBLock) error {
 		}
 	}
 
-	err := fs.WriteFile(lockFile, []byte(fmt.Sprintf("%d:%s:%s", lock.Pid, portStr, lock.Secret)))
+	err = fs.WriteFile(lockFile, []byte(fmt.Sprintf("%d:%s:%s", lock.Pid, portStr, lock.Secret)))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


integration-tests/bats/sql-server.bats
```diff
@@ -1923,3 +1924,51 @@ behavior:
...
+    cd "$baseDir"
+    run dolt sql-server
+    [ $status -eq 1 ]
+    [[ "$output" =~ "Database locked by another sql-server; Lock file" ]] || false
+
+    run dolt sql -q "select current_user"
+    [ $status -eq 0 ]
+    [[ "$output" =~ "__dolt_local_user__@localhost" ]] || false
+
+    teardown
```
fulghum, line 1973: (minor) I don't think you need to explicitly call `teardown`, right? Seems like it gets called automatically after each test.

go/libraries/doltcore/env/environment.go
```diff
@@ -1267,15 +1268,22 @@ func (dEnv *DoltEnv) Unlock() error {
 }
 
 // WriteLockfile writes a lockfile encoding the pid of the calling process.
-func WriteLockfile(fs filesys.Filesys, lock DBLock) error {
+func WriteLockfile(fs filesys.Filesys, lock *DBLock) error {
+	// if the DoltDir doesn't exist, create it.
+	doltDir, _ := fs.Abs(dbfactory.DoltDir)
+	err := fs.MkDirs(doltDir)
```
fulghum, line 1274: Creating the `.dolt` dir without initializing a db may cause problems for other commands... I believe other code looks for the presence of the `.dolt` dir and if found, assumes that it's a valid db and that other data files will be present in `.dolt`, so those commands may start erroring out in weird ways if that assumption doesn't hold. Definitely worth testing that ahead of time to see if it's going to create other problems.

fulghum (commented):
Looks good, but I'm concerned that creating a partial `.dolt` dir may cause problems for other commands that (I believe) assume there will be valid database files present if the `.dolt` dir exists. I think it's worth a little testing to see if it's going to cause errors if other commands are run in that directory.

zachmu (commented):
I think this is fine provided that other commands still work as they currently do after I run `dolt sql-server` and then immediately exit without creating any DBs. I believe there's a lot of logic about assuming that a directory contains a database if it has a .dolt directory. I would write a bunch of tests for things like `dolt init`, `dolt sql`, etc. Make sure they don't throw obsure errors or panic in the presence of this .dolt directory.

macneale4:
I met @zachmu half way. `dolt init` will work fine if there is an empty .dolt directory. If there is anything in it, including a stale sql-server.lock file, it will give up. Added tests for init and sql commands.

fulghum (approved):
Looks good! 🚢 

I did some manual testing with other `dolt` subcommands, too. Most will give a reasonable error message when a `sql-server` is running from a non-init'ed dir:
```shell
❯ dolt log 
The current directories repository state is invalid
open /tmp/aug22/.dolt/repo_state.json: no such file or directory
```

It is a little odd that `status` and `add` (and probably a few other subcommands) will pick the first valid database dir (by name) and use that, when most other subcommands do not. That's separate from your change of course, but just calling it out since I noticed it while testing and seems like a nice future change for the commands to operate more consistently.

PR #6532, opened 2023-08-17
When checking out a branch, consider it clean if the working set differs from the head only by feature version.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6524

`RootValue` is backed by a flatbuffer, but access to the flatbuffer is hidden behind interfaces. We check if two `RootValues` are equal by checking the RootValue's hash. This makes it awkward to check to see if two `RootValues` differ only by a single field.

The solution here is both simple and noninvasive: we create modified root values with the same feature version and hash them again.

This also adds an escape hatch when creating new branches, since we know that the new branches will be clean. We only do this extra check when checking out an existing branch.

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
@@ -29,7 +29,7 @@ import (
 
 // MoveWorkingSetToBranch moves the working set from the currently checked out branch onto the branch specified
 // by `brName`. This is a POTENTIALLY DESTRUCTIVE ACTION used during command line checkout
-func MoveWorkingSetToBranch(ctx *sql.Context, brName string, force bool) error {
+func MoveWorkingSetToBranch(ctx *sql.Context, brName string, force bool, isNewBranch bool) error {
 	branchRef := ref.NewBranchRef(brName)
 	dSess := dsess.DSessFromSess(ctx.Session)
 	dbName := dSess.GetCurrentDatabase()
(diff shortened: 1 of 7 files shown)


integration-tests/bats/feature-version.bats
```diff
@@ -130,3 +130,17 @@ SQL
...
+
+@test "feature-version: dolt checkout can be used when a working set has a different feature version than its head" {
+    rm -rf .dolt/
+    dolt --feature-version $OLD init
+    dolt --feature-version $NEW sql -q "create table a (b int);"
+    dolt --feature-version $NEW checkout -b other
+    run dolt --feature-version $NEW status
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "other" ]] || false
+    dolt --feature-version $NEW checkout main
```
zachmu, line 142: Should also have a test for the new branch case

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -78,8 +79,22 @@ func MoveWorkingSetToBranch(ctx *sql.Context, brName string, force bool) error {
 		if err != nil {
 			return err
 		}
-		if actions.CheckoutWouldStompWorkingSetChanges(currentRoots, newBranchRoots) {
-			return actions.ErrWorkingSetsOnBothBranches
+		if !isNewBranch && actions.CheckoutWouldStompWorkingSetChanges(currentRoots, newBranchRoots) {
```
zachmu, line 82: The idea here is fine but it seems like this logic should be encapsulated by `actions.CheckoutWouldStompWorkingSetChanges()`

zachmu (approved):
LGTM, just a comment about where the logic should live

integration-tests/bats/feature-version.bats
nicktobey: It does, on line 138.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
nicktobey: Fixed.

PR #6544, opened 2023-08-20
Moved AutoIncrementGetter to an internal interface

Hydrocharged:
Fixes https://github.com/dolthub/dolt/issues/6543.
Companion to https://github.com/dolthub/go-mysql-server/pull/1956.

The `AutoIncrementGetter` was moved from the editor to the table years ago, however the GMS interface remained. We had an internal reliance on the interface, even though GMS does not use it at all, so the interface has been moved to be an internal interface only. This fixes the panic as seen in the issue.


fulghum (approved):
LGTM!

PR #6550, opened 2023-08-21
Silence branch create

macneale4:
Previously `dolt branch foobar HEAD` would print a statement that HEAD is an invalid branch name. It would succeed though, and return an exit status of 0. So tests passed.

This is no longer the case.


nicktobey:
We can't just have the check in `createBranch`, since we also need to validate `moveBranch` and `copyBranch`.

nicktobey:
Also, does the `dolt_branch` stored procedure do this validation too? It probably should. We don't want people being able to call `dolt_branch('-b', 'HEAD');`

Other than these two notes, this looks good.

nicktobey (approved):

PR #6560, opened 2023-08-22
[no-release-notes] Refactoring schema merge tests to make room for more tests

fulghum:
No changes to test cases; just reorganizing a bit to prepare for adding more tests. 

Also fixes `TestThreeWayMergeWithSchemaChangeScriptsPrepared` so that they actually run as prepared tests.


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -2760,45 +2760,56 @@ func TestDoltStorageFormatPrepared(t *testing.T) {
...
+
+// runMergeScriptTestsInBothDirections creates a new test run, named |name|, and runs the specified merge |tests|
+// in both directions (right to left merge, and left to right merge). If
+// |runAsPrepared| is true then the test scripts will be run using the prepared
+// statement test code.
+func runMergeScriptTestsInBothDirections(t *testing.T, tests []MergeScriptTest, name string, runAsPrepared bool) {
+	t.Run(name, func(t *testing.T) {
+		t.Run("right to left merges", func(t *testing.T) {
+			for _, script := range tests {
+				func() {
```
max-hoffman, line 2789: does this do anything? I think the callback into the test handler is already parallelized

max-hoffman (approved):
lgtm

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
fulghum: the nested `t.Run`s just give a little more structure to the test results; that inner `func()` on line 2789 is really just there to scope the call to `defer` a couple lines down. It's not the most obvious pattern though, so I'll add a comment. Let me know if you meant something else.

PR #6562, opened 2023-08-22
Bug fix: Building secondary prefix indexes from existing data

fulghum:
When building a secondary index from existing table data and the secondary index contains a prefix of a TEXT/BLOB column, we don't currently dereference the TEXT/BLOB pointer to get the content. Instead, we take the TEXT/BLOB pointer address and incorrectly put that into the secondary index. This manifested as errors in collation sorting, since one string was not a valid, encoded string and was instead a pointer address. Additionally, we weren't applying the prefix to the content for any existing data when building a prefix index – i.e. any VARCHAR data was inserted in full to the prefix index without trimming. 

It's difficult to test indexes directly, so I updated the index validation code in our test suite to support validating prefix indexes. I added some more tests for prefix indexes in GMS, too: https://github.com/dolthub/go-mysql-server/pull/1971

Fixes https://github.com/dolthub/dolt/issues/6520


go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -115,75 +115,29 @@ func TestSingleQuery(t *testing.T) {
...
 // Convenience test for debugging a single query. Unskip and set to the desired query.
 func TestSingleScript(t *testing.T) {
-	t.Skip()
+	//t.Skip()
 
 	var scripts = []queries.ScriptTest{
 		{
-			Name: "datetime precision",
+			// panic: runtime error: slice bounds out of range [-1:]
+			Name: "Slice out of bounds panic repro",
```
zachmu, line 123: Why no new engine test that would catch this?

go/libraries/doltcore/sqle/index/key_builder.go
```diff
@@ -15,16 +15,23 @@
...
+	"github.com/dolthub/dolt/go/store/prolly/tree"
 	"github.com/dolthub/dolt/go/store/val"
 )
 
-func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.TupleDesc, p pool.BuffPool) (b SecondaryKeyBuilder) {
+// NewSecondaryKeyBuilder creates a new SecondaryKeyBuilder instance that can build keys for the secondary index |def|.
+// The schema of the source table is defined in |sch|, and |idxDesc| describes the tuple layout for the index's keys
+// (index value tuples are not used).
+func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.TupleDesc, p pool.BuffPool, nodeStore tree.NodeStore) (b SecondaryKeyBuilder) {
 	b.builder = val.NewTupleBuilder(idxDesc)
```
zachmu, line 30: Seems like an unnamed result param and a struct literal is more idiomatic for these constructor funcs

go/libraries/doltcore/sqle/index/key_builder.go
```diff
@@ -56,30 +63,66 @@ func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.Tup
...
+			//       prefix indexes) as well as custom handling for ZCell geometry fields.
 			b.builder.PutRaw(to, k.GetField(from))
 		} else {
+			// the "from" field comes from the value tuple fields
 			from -= b.split
-			buf := v.GetField(from)
+
+			// Copying the raw bytes from the source table to the secondary index is more efficient
+			// but some cases such as CellEnc or prefix indexes mean we have to manipulate the data.
+			copyRawBytes := true
```
zachmu, line 99: Maybe a method canCopyRawBytes?

go/libraries/doltcore/sqle/index/key_builder.go
```diff
@@ -56,30 +63,66 @@ func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.Tup
...
-				buf = cell[:]
+				copyRawBytes = false
+			} else if len(b.def.PrefixLengths()) > to && b.def.PrefixLengths()[to] > 0 {
+				copyRawBytes = false
+			}
+
+			if copyRawBytes {
+				b.builder.PutRaw(to, v.GetField(from))
+			} else {
+				value, err := GetField(ctx, b.sch.GetValueDescriptor(), from, v, b.nodeStore)
```
zachmu, line 109: Maybe a method for this branch?

zachmu (approved):
Looks pretty good. Only major issue is the lack of an engine test that triggers on the original issue

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
fulghum: Sorry, I should have highlighted that better – I added tests to the PrefixIndex tests in GMS PR linked from the description. Those tests do trigger the same collation error that happened in the linked issue. 
https://github.com/dolthub/go-mysql-server/pull/1971

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -744,14 +744,21 @@ func newUniqIndex(sch schema.Schema, def schema.Index, clustered, secondary prol
...
 func (idx uniqIndex) insertRow(ctx context.Context, key, value val.Tuple) error {
-	secondaryIndexKey := idx.secondaryBld.SecondaryKeyFromRow(key, value)
-	newValue := val.NewTuple(idx.secondary.NodeStore().Pool(), nil)
-	return idx.secondary.Put(ctx, secondaryIndexKey, newValue)
+	secondaryIndexKey, err := idx.secondaryBld.SecondaryKeyFromRow(ctx, key, value)
+	if err != nil {
+		return err
+	}
+
+	return idx.secondary.Put(ctx, secondaryIndexKey, val.EmptyTuple)
```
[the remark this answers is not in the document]
fulghum: Added a comment mentioning that secondary indexes don't use their value tuple and instead store everything in the key tuple.

go/libraries/doltcore/merge/mutable_secondary_index.go
```diff
@@ -98,40 +98,46 @@ func NewMutableSecondaryIdx(idx prolly.Map, sch schema.Schema, def schema.Index)
...
 func (m MutableSecondaryIdx) InsertEntry(ctx context.Context, key, newValue val.Tuple) error {
-	newKey := m.builder.SecondaryKeyFromRow(key, newValue)
-	err := m.mut.Put(ctx, newKey, val.EmptyTuple)
+	newKey, err := m.builder.SecondaryKeyFromRow(ctx, key, newValue)
 	if err != nil {
-		return nil
+		return err
+	}
+
+	err = m.mut.Put(ctx, newKey, val.EmptyTuple)
```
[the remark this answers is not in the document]
fulghum: Added a comment mentioning that secondary indexes don't use their value tuple and instead store everything in the key tuple.

PR #6568, opened 2023-08-23
updates `dolt log` to use sql backend

stephkyou:
Updates `dolt log` to use sql backend to generate results.

### DIFF
--- a/go/cmd/dolt/commands/log.go
+++ b/go/cmd/dolt/commands/log.go
@@ -252,8 +252,19 @@ func constructInterpolatedDoltLogQuery(apr *argparser.ArgParseResults, queryist 
 func getExistingTables(revisions []string, queryist cli.Queryist, sqlCtx *sql.Context) (map[string]bool, error) {
 	tableNames := make(map[string]bool)
 
+	if len(revisions) == 0 {
+		rows, err := GetRowsForSql(queryist, sqlCtx, "show tables")
+		if err != nil {
+			return nil, err
+		}
+		for _, r := range rows {
+			tableNames[r[0].(string)] = true
+		}
+		return tableNames, nil
+	}
+
 	for _, rev := range revisions {
-		rows, err := GetRowsForSql(queryist, sqlCtx, "show tables as of "+rev)
+		rows, err := GetRowsForSql(queryist, sqlCtx, "show tables as of '"+rev+"'")
 		if err != nil {
 			return nil, err
 		}
--- a/go/cmd/dolt/commands/log.go
+++ b/go/cmd/dolt/commands/log.go
@@ -142,34 +142,25 @@ func constructInterpolatedDoltLogQuery(apr *argparser.ArgParseResults, queryist 
 		}
 		tableNames := ""
 		for i := apr.PositionalArgsSeparatorIndex; i < apr.NArg(); i++ {
-			tableNames = tableNames + apr.Arg(i) + ","
+			tableNames = strings.Join([]string{tableNames, apr.Arg(i)}, ",")
 		}
 		if tableNames != "" {
-			tableNames = strings.TrimSuffix(tableNames, ",")
 			params = append(params, tableNames)
 			writeToBuffer("'--tables'")
 			writeToBuffer("?")
 		}
 	} else {
-		rows, err := GetRowsForSql(queryist, sqlCtx, "show tables")
-		if err != nil {
-			return "", err
-		}
-		existingTables := make(map[string]bool, len(rows))
-		for _, r := range rows {
-			existingTables[r[0].(string)] = true
-		}
+		var existingTables map[string]bool
 		seenRevs := make(map[string]bool, apr.NArg())
-
 		finishedRevs := false
 		tableNames := ""
-		for _, arg := range apr.Args {
+		for i, arg := range apr.Args {
 			// once we encounter a rev we can't resolve, we assume the rest are table names
 			if finishedRevs {
 				if _, ok := existingTables[arg]; !ok {
 					return "", fmt.Errorf("error: table %s does not exist", arg)
 				}
-				tableNames = tableNames + arg + ","
+				tableNames = strings.Join([]string{tableNames, arg}, ",")
 			} else {
 				if strings.Contains(arg, "..") || strings.HasPrefix(arg, "^") {
 					writeToBuffer("?")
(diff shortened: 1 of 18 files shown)


go/cmd/dolt/commands/log.go
```diff
@@ -110,576 +88,188 @@ func (cmd LogCmd) ArgParser() *argparser.ArgParser {
...
+		if !first {
+			buffer.WriteString(", ")
 		}
-
-		// `dolt log <ref>..<ref>`
-		refs := strings.Split(apr.Arg(0), "..")
-		notCs, err := getCommitSpec(refs[0])
-		if err != nil {
-			return err
+		if !param {
```
nicktobey, line 134: I think we can remove the `param` param if we just wrap `s` in quotes when it's not a parameter, like

```
writeToBuffer("\"--table\"")
```

integration-tests/bats/helper/local-remote.bash
```diff
@@ -136,6 +135,7 @@ SKIP_SERVER_TESTS=$(cat <<-EOM
 ~sql-checkout.bats~
 ~cli-hosted.bats~
 ~profile.bats~
+~log.bats~
```
macneale4, line 138: Would really like to have this one on. What doesn't work? How hard is it to fix?

go/libraries/utils/argparser/results.go
```diff
@@ -135,7 +137,7 @@ func (res *ArgParseResults) DropValue(name string) *ArgParseResults {
 		}
 	}
 
-	return &ArgParseResults{newNamedArgs, res.Args, res.parser}
+	return &ArgParseResults{newNamedArgs, res.Args, res.parser, -1}
```
macneale4, line 140: Having -1 as a special number is not great. I'd make a descriptive constant, like "NO_DASH_DASH_ARGS" or something like that. Future readers will easily know what that means.

macneale4 (changes requested):

integration-tests/bats/log.bats
```diff
@@ -254,14 +265,14 @@ teardown() {
     [[ "$output" =~ "BRANCH1" ]] || false
 
     # Should default to first argument as branch name, second argument as table name (if table exists) if two arguments provided
-    run dolt log myname myname 
+    run dolt log myname -- myname
```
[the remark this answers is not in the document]
stephkyou: Yeah we're no longer supporting table names without the separator. I think @macneale4 and I decided that would be best since it's really hard to resolve revisions otherwise.

integration-tests/bats/log.bats
```diff
@@ -254,14 +265,14 @@ teardown() {
...
     [ $status -eq 0 ]
     [[ ! "$output" =~ "MAIN" ]] || false
     [[ "$output" =~ "BRANCH1" ]] || false
 
     # Table main does not exist
-    run dolt log main main
-    [ $status -eq 1 ]
+    run dolt log main -- main
```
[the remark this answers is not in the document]
stephkyou: This follow's git's pattern where the log will just be empty if a table (file) doesn't exist.

integration-tests/bats/log.bats
```diff
@@ -479,15 +490,14 @@ teardown() {
     dolt add test
     dolt commit -m "Commit 2"
 
-    run dolt log test
-    [ $status -eq 1 ]
-    [[ "$output" =~ "error: table test does not exist" ]] || false
+    run dolt log -- test
+    [ $status -eq 0 ]
```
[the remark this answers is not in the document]
stephkyou: This follow's git's pattern where the log will just be empty if a table (file) doesn't exist.

integration-tests/bats/helper/local-remote.bash
stephkyou: Needs checkout to work remotely :(

integration-tests/bats/log.bats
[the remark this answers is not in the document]
stephkyou: After discussing with Neil I'm going to push a change to support table names without the separator. Should make us backwards compatible.

go/cmd/dolt/commands/log.go
```diff
@@ -110,576 +88,232 @@ func (cmd LogCmd) ArgParser() *argparser.ArgParser {
...
+	if apr.PositionalArgsSeparatorIndex >= 0 {
+		for i := 0; i < apr.PositionalArgsSeparatorIndex; i++ {
+			writeToBuffer("?")
+			params = append(params, apr.Arg(i))
 		}
-
-		cs, err := getCommitSpec(refs[1])
+		tableNames := ""
+		for i := apr.PositionalArgsSeparatorIndex; i < apr.NArg(); i++ {
+			tableNames = tableNames + apr.Arg(i) + ","
```
macneale4, line 145: Use join so you don't need to trim the comma: https://pkg.go.dev/strings#example-Join

go/cmd/dolt/commands/log.go
```diff
@@ -110,576 +88,232 @@ func (cmd LogCmd) ArgParser() *argparser.ArgParser {
...
+		}
+		if tableNames != "" {
+			tableNames = strings.TrimSuffix(tableNames, ",")
+			params = append(params, tableNames)
+			writeToBuffer("'--tables'")
+			writeToBuffer("?")
+		}
+	} else {
+		rows, err := GetRowsForSql(queryist, sqlCtx, "show tables")
```
macneale4, line 154: `show tables` will give you the table on the current HEAD. If you are logging on a different branch, this will possibly give incorrect results. This would be a really good test to add.

integration-tests/bats/helper/local-remote.bash
macneale4: How many of the tests required checkout? It would be worth discussing if we can just skip the tests in remote mode which depend on it.

integration-tests/bats/sql-local-remote.bats
macneale4, line 1: Add a test for log to ensure the local and remote produce the same results.

macneale4 (changes requested):
Getting so close! Just a couple bits of feedback and one additional test needed!

integration-tests/bats/sql-local-remote.bats
stephkyou: I think 9/24 tests require checkout. All the tests using branch name as a revision use checkout for setup.

integration-tests/bats/sql-local-remote.bats
```diff
@@ -1103,3 +1103,18 @@ SQL
...
+    run dolt --verbose-engine-setup log
+    [ $status -eq 0 ]
+    [[ "$output" =~ "starting local mode" ]] || false
+    [[ "$output" =~ "tables table1, table2" ]] || false
+
+    start_sql_server altDB
+    run dolt --verbose-engine-setup log
+    [ $status -eq 0 ]
+    [[ "$output" =~ "starting remote mode" ]] || false
+    [[ "$output" =~ "tables table1, table2" ]] || false
```
macneale4, line 1119: A more complete test would take the output of both command executions, and verify that they match. This is impossible to do with the --verbose-engine-setup flag, so you probably need to expand this test to have more command invocations.

macneale4 (approved):
Woot! You've reached the point where I don't need to review this anymore. There are a few updates requested though, some optional. Take one last pass and ship when the tests pass. I'm so happy that this command has been migrated. Great work!

PR #6570, opened 2023-08-23
Bug fix:  Correctly detect uncommitted changes for `dolt_pull` when roots have different feature versions

fulghum:
When pulling between two Dolt databases that have different feature versions, `dolt_pull()` can mistakenly interpret the difference in root value hash as being uncommitted changes. This change updates `dolt_pull()` to use the existing `actions.RootHasUncommittedChanges` function and changes that function to account for differing feature version when determining if roots have uncommitted changes.

### DIFF
--- a/go/libraries/doltcore/env/actions/checkout.go
+++ b/go/libraries/doltcore/env/actions/checkout.go
@@ -536,8 +536,15 @@ func clearFeatureVersion(ctx context.Context, roots doltdb.Roots) (doltdb.Roots,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


macneale4 (approved):
LGTM

max-hoffman:
are there any other dolt commands that are subject to this issue? we did checkout and pull, merge? clone? gc?

fulghum:
> are there any other dolt commands that are subject to this issue? we did checkout and pull, merge? clone? gc?

Good question. I tested stored procedures and commands for `gc`, `merge`, `cherry-pick`, `clone`, `stash` and couldn't repro the issue anywhere else.

PR #6584, opened 2023-08-24
Migrate the dolt_procedures table whenever stored procedures are used

nicktobey:
Fixes: https://github.com/dolthub/dolt/issues/6579

Before, we were only migrating the table when a new stored procedure was added. This means that any branches that predate our recent change to dolt_procedures would be broken until a new stored procedure was added. This fixes that by checking to see if the table needs to be migrated whenever the stored procedures are listed or used, not just when they're added.


fulghum (approved):
Nice fix! Looks great. Thank you! 🚢

PR #6595, opened 2023-08-29
Feature: Allow `*` wildcard in `@@dolt_replicate_heads`

fulghum:
This change allows customers to use `*` as a wildcard in the [`@@dolt_replicate_heads` system variable](https://docs.dolthub.com/sql-reference/version-control/dolt-sysvars#dolt_replicate_heads) to match zero or more characters in a branch name. For example, to replicate the `main` branch and all branches that start with `feature`, the following configuration can now be applied:
```
    dolt config --local --add sqlserver.global.dolt_replicate_heads main,feature*
```

Documentation: https://github.com/dolthub/docs/pull/1698

Fixes: https://github.com/dolthub/dolt/issues/6486

### DIFF
--- a/go/libraries/doltcore/sqle/wildcard.go
+++ b/go/libraries/doltcore/sqle/wildcard.go
@@ -19,9 +19,9 @@ import (
 	"strings"
 )
 
-// match returns true if |s| fully matches |pattern|. |pattern| may contain '*' wildcards
-// which match 0 or more characters.
-func match(pattern string, s string) bool {
+// matchWildcardPattern returns true if |s| fully matches |pattern|. |pattern| may contain '*' wildcards
+// which matchWildcardPattern 0 or more characters.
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/wildcard.go
```diff
@@ -0,0 +1,52 @@
...
+package sqle
+
+import (
+	"regexp"
+	"strings"
+)
+
+// match returns true if |s| fully matches |pattern|. |pattern| may contain '*' wildcards
+// which match 0 or more characters.
+func match(pattern string, s string) bool {
```
max-hoffman, line 24: matchWildcard? matchGlob?

max-hoffman (approved):
lgtm, regex conversion seems more reliable if not as fast as a more direct glob checking

PR #6598, opened 2023-08-29
Move storage format version to only be displayed with `dolt version --verbose`

fulghum:
`dolt version` displayed the storage format version by default, but this information is less relevant now that almost all Dolt databases have been migrated to the new storage format. This change adds a new `--verbose` /  `-v` flag that customers can specify if they do want to see the storage format version for a repository. 

Fixes https://github.com/dolthub/dolt/issues/6528


macneale4 (approved):
LGTM

PR #6619, opened 2023-09-05
Temporarily restrict `dolt_conflicts_resolve()` from resolving schema conflicts

fulghum:
Until we fix the workflow around merging data **after** schema conflicts have been resolved, customers may be surprised by the results of `dolt_conflicts_resolve()` when there are schema changes, so putting this block in out of extra caution. 

For more details, see issue https://github.com/dolthub/dolt/issues/6616


zachmu (approved):
LGTM

PR #6621, opened 2023-09-06
fixes profiles not working with multiple profiles added

stephkyou:
Fixes an issue with adding and using multiple profiles


macneale4 (approved):
LGTM

PR #6631, opened 2023-09-07
go: sqle: async replication: Fix a bug in the logic which decides what to push to stop pushing deleted branches after we push them once.

reltuk:


zachmu (approved):
Seems reasonable, but what about a test?

zachmu (approved):
LGTM!

PR #6652, opened 2023-09-12
migrates fetch to use sql queries

stephkyou:
This change updates `dolt fetch` to use the appropriate sql engine to generate results.

### DIFF
--- a/go/cmd/dolt/commands/fetch.go
+++ b/go/cmd/dolt/commands/fetch.go
@@ -79,7 +79,8 @@ func (cmd FetchCmd) Exec(ctx context.Context, commandStr string, args []string, 
 
 	queryist, sqlCtx, closeFunc, err := cliCtx.QueryEngine(ctx)
 	if err != nil {
-		handleErrAndExit(err)
+		cli.PrintErrln(err)
+		return 1
 	}
 	if closeFunc != nil {
 		defer closeFunc()
--- a/go/cmd/dolt/commands/fetch.go
+++ b/go/cmd/dolt/commands/fetch.go
@@ -15,16 +15,16 @@
 package commands
 
 import (
+	"bytes"
 	"context"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
 	eventsapi "github.com/dolthub/dolt/go/gen/proto/dolt/services/eventsapi/v1alpha1"
-	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env"
-	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
-	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
+	"github.com/gocraft/dbr/v2"
+	"github.com/gocraft/dbr/v2/dialect"
 )
 
 var fetchDocs = cli.CommandDocumentationContent{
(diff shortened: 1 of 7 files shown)


go/cmd/dolt/dolt.go
```diff
@@ -453,6 +452,15 @@ func runMain() int {
 		return 1
 	}
 
+	if subcommandName == "fetch" || subcommandName == "pull" || subcommandName == "push" {
```
macneale4, line 455: move down to line 628.

go/cmd/dolt/commands/fetch.go
```diff
@@ -67,54 +68,69 @@ func (cmd FetchCmd) ArgParser() *argparser.ArgParser {
...
+func constructInterpolatedDoltFetchQuery(apr *argparser.ArgParseResults) (string, error) {
+	var params []interface{}
+
+	var buffer bytes.Buffer
+	var first bool
+	first = true
+
+	buffer.WriteString("call dolt_fetch(")
+
+	writeToBuffer := func(s string) {
```
macneale4, line 113: I think this function is simple enough that you could just use a strings.join() for this.

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
+
+setup() {
+    setup_common
+    TMPDIRS=$(pwd)/tmpdirs
```
macneale4, line 6: TMPDIRS is very close to the standard TMPDIR. How about FETCHROOT?

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,322 @@
...
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "added table" ]] || false
+
+    run dolt sql -q "show tables as of hashof('origin/main')" -r csv
+    [ "${#lines[@]}" -eq 2 ]
+    [[ "$output" =~ "Table" ]] || false
+    [[ "$output" =~ "t1" ]] || false
+}
+
+@test "fetch: fetch main" {
```
macneale4, line 68: This test isn't valid is there isn't a check for the absence of something. You should create another branch on the remote, and verify that it doesn't get fetched when you just fetch main.

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,322 @@
...
+    dolt init
+    dolt remote add origin file://../rem1
+    dolt remote add test-remote file://../rem1
+    dolt push origin main
+
+    cd $TMPDIRS
+    dolt clone file://rem1 repo2
+    cd $TMPDIRS/repo2
+    dolt branch feature
+    dolt remote add test-remote file://../rem1
```
macneale4, line 20: This remote should point to a remote that has different content. Then in tests which pull from `test-remote`, you can verify that you didn't fetch from origin.

macneale4 (changes requested):
Lookin' good. A few minor things

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,342 @@
...
+
+@test "fetch: basic fetch" {
+    cd repo2
+    dolt fetch
+
+    run dolt diff main origin/main
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "added table" ]] || false
+
+    run dolt sql -q "show tables as of hashof('origin/main')" -r csv
```
zachmu, line 48: can just say `as of 'origin/main'` here

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,342 @@
...
+    [[ "$output" =~ "t1" ]] || false
+}
+
+@test "fetch: fetch specific ref" {
+    cd repo2
+    dolt fetch test-remote refs/heads/main:refs/remotes/test-remote/main
+
+    run dolt diff main test-remote/main
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "added table" ]] || false
```
zachmu, line 114: This test has the same problem Neil pointed out: the assertions here look exactly the same as origin.

You need different contents in the thing you're fetching so you can be sure you got the right thing

integration-tests/bats/fetch.bats
```diff
@@ -0,0 +1,342 @@
...
+    cd ../repo2
+    dolt fetch origin feature
+
+    run dolt diff main origin/feature
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "added table" ]] || false
+
+    run dolt sql -q "show tables as of hashof('origin/feature')" -r csv
+    [ "${#lines[@]}" -eq 2 ]
+    [[ "$output" =~ "Table" ]] || false
```
zachmu, line 135: Same comment

It's not enough to just push / fetch different refs, the assertions need to rule out that you fetched a different one than you intended

zachmu: Same applies to all the tests below. Put a new commit / table on the thing you're fetching so you know you got the right thing

zachmu (approved):
This looks pretty good, but the tests still need some work. You should have a different set of tables on the different remotes, or some other way to assert that you fetched the right one. Once you fix that you're good to go.

PR #6669, opened 2023-09-14
go: sqle: cluster: First pass at replicating branch control permissions.

reltuk:


go/libraries/doltcore/sqle/cluster/branch_control_replica.go
```diff
@@ -0,0 +1,251 @@
...
+	// At this point, none of the callbacks will ever be called again and
+	// ch.setWaitNotify grabs a lock and so establishes the happens before.
+	all := true
+	for _, b := range caughtup {
+		if !b {
+			wg.Done()
+			all = false
+		}
+	}
+	<-done
```
zachmu, line 248: Seems odd to do this here, is the concern there's a channel leak otherwise? This will always either instantly succeed or block forever

zachmu (approved):
LGTM, no major comments

PR #6688, opened 2023-09-19
Retain display width for `TINYINT(1)`

fulghum:
MySQL allows integer fields to specify a display width (e.g. TINYINT(1)). Dolt currently parses that information, but doesn't retain it anywhere. This PR changes that behavior to match MySQL and retain the display width setting so that it can be passed back to callers.

As of MySQL 8.1.0, the display width setting is ONLY retained for signed TINYINT fields and ONLY when the display width is set to 1.

Fixes: https://github.com/dolthub/dolt/issues/6682

GMS PR: https://github.com/dolthub/go-mysql-server/pull/2017


zachmu (approved):
LGTM!

PR #6694, opened 2023-09-20
Support for fulltext index rewrites inline with table rewrites

zachmu:
Also includes changes required from always closing table editors, even on an error.

Companion PR with support for GMS table rewrites, and similar changes to GMS fulltext index management:

https://github.com/dolthub/go-mysql-server/pull/2008


go/libraries/doltcore/sqle/dtables/column_diff_table.go
```diff
@@ -109,17 +113,17 @@ func (dt *ColumnDiffTable) PartitionRows(ctx *sql.Context, partition sql.Partiti
...
-//func (dt *ColumnDiffTable) IndexedAccess(lookup sql.IndexLookup) sql.IndexedTable {
-//	nt := *dt
-//	return &nt
-//}
+// // todo fix indexed paths, these were missing tests in CI
+// // GetIndexes implements sql.IndexAddressable
+// func (dt *ColumnDiffTable) GetIndexes(ctx *sql.Context) ([]sql.Index, error) {
+// 	return index.DoltCommitIndexes(dt.Name(), dt.ddb, true)
+// }
+//
```
Hydrocharged, line 121: Should we delete these comments?

go/libraries/doltcore/sqle/tables.go
```diff
@@ -598,8 +616,142 @@ func (t *WritableDoltTable) getFullTextEditor(ctx *sql.Context) (fulltext.TableE
 			RowCount:    rowCountTable.(fulltext.EditableTable),
 		})
 	}
+	return configTable, sets, nil
+}
 
-	return fulltext.CreateEditor(ctx, t, configTable, sets...)
+// tableSetsForRewrite returns the fulltext.TableSet for each Full-Text index in the table, truncated and modified
+// for a table rewrite operation. Returns the root given with all full-set pseudo tables updated with their new
```
Hydrocharged, line 623: I think you meant Full-Text instead of Full-Set? You're using Full-Set in multiple places here lol

go/libraries/doltcore/sqle/tables.go
```diff
@@ -1470,6 +1596,226 @@ func (t *AlterableDoltTable) RewriteInserter(
...
+	dbState dsess.SessionState,
+	workingRoot *doltdb.RootValue,
+) (sql.RowInserter, error) {
+
+	newTable, err := t.db.newDoltTable(t.Name(), newSch, dt)
+	if err != nil {
+		return nil, err
+	}
+
+	updatedRoot, configTable, tableSets, err := newTable.(*AlterableDoltTable).tableSetsForRewrite(ctx, workingRoot)
```
Hydrocharged, line 1615: Forgot to check this error

Hydrocharged (commented):
Overall looks good, I don't see anything that looks like it needs reworking or improving. It looks like the bats tests do _not_ like this change, so it looks like this needs a pass over the merge functionality.

go/libraries/doltcore/sqle/dtables/column_diff_table.go
zachmu: Yeah good call

go/libraries/doltcore/sqle/tables.go
zachmu: good catch

PR #6699, opened 2023-09-20
go/libraries/doltcore/sqle: cluster: When a database is dropped, shutdown the replication goroutines and stop returning its status in the dolt_cluster_status table.

reltuk:


zachmu (approved):
Seems reasonable

PR #6702, opened 2023-09-21
reduce release sizes and speed up release time

phanirithvij:
- Uses pigz instead of gzip for faster gzipping
- Uses 7zip to create zip and 7z archives for windows (7z archive is ~49% smaller than zip archive)
- add ldflags='-s' to go build to reduce binary size by strip (reduces
  to 89MB)
- Upload 7z to releases
- tested action on fork, works 
https://github.com/phanirithvij/dolt/releases/tag/v1.16.2

If pigz seems unnecessary feel free to remove it, it works well though.


timsehn:
Thanks for the contribution. I think @reltuk and @coffeegoddd should comment on this one.

reltuk:
Hi @phanirithvij, thanks for the contribution.

Changes for `7z` and `pigz` seem fine, and its nice that they're already available in the package repositories of the github action runner image.

Can you comment at all on the effect of `ldflags="-w -s"` on the ability of the resulting binary to collect heap and cpu profiles and/or generate stack traces for things like panics or `runtime/debug.Stack() []byte`? We like the idea of shipping smaller release artifacts, but we do have customers collect profiles on site for us sometimes, for example.

Thanks for your interest and contribution!

phanirithvij:
I haven't considered those aspects, I apologize.

**Tl;DR**
- go tool nm won't work with -s.
- stacktraces work, both panic and runtime/debug.Stack().
- I'm not familiar with profiling but based on a basic attempt I made, it seems to work fine with `-s` though has a less granular pprof graph and http profiles are less granular aswell.

Quoting a stackoverflow answer from Russ Cox. [Link](https://stackoverflow.com/a/22276273/8608146)
<details><summary>Expand</summary>

> You will get the smallest binaries if you compile with -ldflags '-w -s'.

> The -w turns off DWARF debugging information: you will not be able to use gdb on the binary to look at specific functions or set breakpoints or get stack traces, because all the metadata gdb needs will not be included. You will also not be able to use other tools that depend on the information, like pprof profiling.

> The -s turns off generation of the Go symbol table: you will not be able to use go tool nm to list the symbols in the binary. strip -s is like passing -s to -ldflags but it doesn't strip quite as much. go tool nm might still work after strip -s. I am not completely sure.

> None of these — not -ldflags -w, not -ldflags -s, not strip -s — should affect the execution of the actual program. They only affect whether you can debug or analyze the program with other tools.
</details>

From what I can read -w doesn't allow profiling, so maybe just `-s` can be used, unless `go tool nm` is directly/indirectly required for dolt?

```
$ go tool nm -size ./dolttrimstrip
reading ./dolttrimstrip: no symbol section
reading ./dolttrimstrip: no symbols
```

I tested whether the stack traces print properly with 'ldflags="-w -s"' and they do, for both `panic` and `runtime/debug.Stack()`. So stack traces work with those flags.

<details><summary>Sample to demonstrate stacktraces</summary>

```go
package main

import (
	"runtime/debug"
	"time"
)

func main() {
	go func() {
		panic("No!!")
	}()
	go func() {
		debug.PrintStack()
	}()
	time.Sleep(time.Second)
}
```

```
go mod init example.com/sample
go build -trimpath -o sample.trimpath.out
go build -ldflags="-s -w" -trimpath -o sample.strippedNtrimmed.out
```

```
$ for i in $(ls -1 *.out); do echo -e "\n........"$i".......\n"; sh -c ./$i; done

........sample.strippedNtrimmed.out.......

panic: No!!

goroutine 6 [running]:
main.main.func1()
        example.com/sample/main.go:10 +0x25
created by main.main in goroutine 1
        example.com/sample/main.go:9 +0x1a

........sample.trimpath.out.......

goroutine 19 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x5e
runtime/debug.PrintStack()
        runtime/debug/stack.go:16 +0x13
main.main.func2()
        example.com/sample/main.go:13 +0xf
created by main.main in goroutine 1
        example.com/sample/main.go:12 +0x26
panic: No!!

goroutine 18 [running]:
main.main.func1()
        example.com/sample/main.go:10 +0x25
created by main.main in goroutine 1
        example.com/sample/main.go:9 +0x1a

```

</details>

Also `-trimpath` is useful to remove the absolute paths at buildtime and use package paths instead. (new commit)

I built dolt with the '-s' flag (no -w) and ran `./dolt --prof mem --pprof-server sql-server` (found these flags from https://github.com/dolthub/dolt/issues/5305#issuecomment-1421446044) vs without these flags.
```
go build -trimpath -ldflags="-s" -o dolttrimstrip ./cmd/dolt
go build -trimpath -o dolt ./cmd/dolt
./dolttrimstrip --prof mem --pprof-server sql-server
# ./dolt --prof mem --pprof-server sql-server

go tool pprof -http=:8080 /tmp/........./mem.pprof #mem prof output by dolt
curl "http://localhost:6060/debug/pprof/allocs?debug=1" > dolt.txt
# curl "http://localhost:6060/debug/pprof/allocs?debug=1" > dolttrimstrip.txt
```

I can see some slight differences but I'm not familiar with profiling much, so I can't add to that. It seems to profile but with less granularity from what I can see.

Please decide whether or not to strip the binary `-ldflags='-s'` after checking if profiling works.
Please also factor the `-trimpath` changes when profiling (new commit I made)

reltuk:
Landing this for now. We will be testing the `ldflags` and may iterate a bit on the `-w` in particular depending on what we see for profiling data on all platforms we support.

Thank you for the contribution and for your quick and thorough response to our follow up :).

PR #6703, opened 2023-09-22
When a three-way merge contains an altered column, convert values to the new column type in order to prevent conflicts.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6660

I thought this was going to be a simple tweak, but it turns out that three-way merges in the presence of altered columns has some interesting subtleties.

First off, we don't just compare the bytes of cells when merging, we actually de-serialize them and check for equality. I suspect this doesn't matter in most cases, but is done because:
- Some types may have multiple ways to represent the same value, and
- In the event of an altered column, two distinct values (of different types) may have the same representation.

This leads to some subtle edge cases: for example, if once branch alters a column type, and then modifies a row so the cell as the same representation as the original, unaltered value, we still need to detect that as a modification, otherwise we might silently drop it during the merge.

On the other side of the coin, if one branch alters a column and the other branch modifies that column, that's not necessarily a conflict even though the representations of both cells have changed.

The logic for merging cells is more complicated now than it was before, but I added comments that should make it easy to follow. In brief, we detect when columns have been altered and convert values as appropriate before checking to see if they've been modified by the branch.

In slightly more detail, if we have a cell with value `base` (of type A), a branch `right` which alters the column to have type B, and a branch `left` which modifies the cell (but is still type A):

- We convert `base` to type B in order to detect whether the branch `right` also modified the cell.
- We also convert the value on `left` to type B in order to detect whether a conflict exists with the value on `right`.

I'm still writing extra tests, and the `Convert` method doesn't support every type yet, but the design is ready for feedback.

### DIFF
--- a/go/libraries/doltcore/merge/merge_prolly_rows.go
+++ b/go/libraries/doltcore/merge/merge_prolly_rows.go
@@ -1373,13 +1373,13 @@ func mergeTableArtifacts(ctx context.Context, tm *TableMerger, mergeTbl *doltdb.
 // key but with conflicting values. A successful resolve produces
 // a three-way cell edit (tree.DiffOpDivergentModifyResolved).
 type valueMerger struct {
-	numCols                                           int
-	baseVD, rightVD, resultVD                         val.TupleDesc
-	baseSchema, leftSchema, rightSchema, resultSchema schema.Schema
-	leftMapping, rightMapping, baseMapping            val.OrdinalMapping
-	syncPool                                          pool.BuffPool
-	keyless                                           bool
-	ns                                                tree.NodeStore
+	numCols                                int
+	baseVD, rightVD, resultVD              val.TupleDesc
+	resultSchema                           schema.Schema
+	leftMapping, rightMapping, baseMapping val.OrdinalMapping
+	syncPool                               pool.BuffPool
+	keyless                                bool
+	ns                                     tree.NodeStore
 }
 
 func newValueMerger(merged, leftSch, rightSch, baseSch schema.Schema, syncPool pool.BuffPool, ns tree.NodeStore) *valueMerger {
@@ -1390,9 +1390,6 @@ func newValueMerger(merged, leftSch, rightSch, baseSch schema.Schema, syncPool p
 		baseVD:       baseSch.GetValueDescriptor(),
 		rightVD:      rightSch.GetValueDescriptor(),
 		resultVD:     merged.GetValueDescriptor(),
-		baseSchema:   baseSch,
-		leftSchema:   leftSch,
-		rightSchema:  rightSch,
 		resultSchema: merged,
 		leftMapping:  leftMapping,
 		rightMapping: rightMapping,
@@ -1549,9 +1546,9 @@ func (m *valueMerger) tryMerge(ctx context.Context, left, right, base val.Tuple)
 // based on the |left|, |right|, and |base| schema.
 func (m *valueMerger) processColumn(ctx context.Context, i int, left, right, base val.Tuple) ([]byte, bool) {
 	// missing columns are coerced into NULL column values
-	baseCol, baseColIdx, baseColExists := getColumn(&base, m.baseSchema, &m.baseMapping, i)
-	leftCol, leftColIdx, leftColExists := getColumn(&left, m.leftSchema, &m.leftMapping, i)
-	rightCol, rightColIdx, rightColExists := getColumn(&right, m.rightSchema, &m.rightMapping, i)
+	baseCol, baseColIdx, baseColExists := getColumn(&base, &m.baseMapping, i)
+	leftCol, leftColIdx, leftColExists := getColumn(&left, &m.leftMapping, i)
+	rightCol, rightColIdx, rightColExists := getColumn(&right, &m.rightMapping, i)
 	resultType := m.resultVD.Types[i]
 
 	if base == nil || !baseColExists {
@@ -1633,7 +1630,7 @@ func (m *valueMerger) processColumn(ctx context.Context, i int, left, right, bas
 	}
 }
 
-func getColumn(tuple *val.Tuple, schema schema.Schema, mapping *val.OrdinalMapping, idx int) (col []byte, colIndex int, exists bool) {
+func getColumn(tuple *val.Tuple, mapping *val.OrdinalMapping, idx int) (col []byte, colIndex int, exists bool) {
 	colIdx := (*mapping)[idx]
 	if colIdx == -1 {
 		return nil, -1, false
--- a/go/libraries/doltcore/merge/merge_prolly_rows.go
+++ b/go/libraries/doltcore/merge/merge_prolly_rows.go
@@ -1588,8 +1588,8 @@ func (m *valueMerger) processColumn(ctx context.Context, i int, left, right, bas
 		return nil, false
 	}
 
-	b := m.baseMapping[i]
-	baseVal := base.GetField(b)
+	baseColumnIndex := m.baseMapping[i]
+	baseVal := base.GetField(baseColumnIndex)
 
 	// We can now assume that both left are right are modifications to an existing column.
 	// We want to know if they're the same modification, otherwise there's a conflict.
@@ -1617,7 +1617,7 @@ func (m *valueMerger) processColumn(ctx context.Context, i int, left, right, bas
 		if rightSchemaChanged {
 			// Attempt to convert the left column to match the changed right schema
 			var conflict bool
-			leftCol, conflict = Convert(ctx, m.leftVD, m.resultVD, m.resultSchema, i, left, m.ns)
+			leftCol, conflict = convert(ctx, m.leftVD, m.resultVD, m.resultSchema, leftColumnIndex, i, left, m.ns)
 			if conflict {
 				return nil, true
 			}
@@ -1625,7 +1625,7 @@ func (m *valueMerger) processColumn(ctx context.Context, i int, left, right, bas
 			leftModified = m.resultVD.Comparator().CompareValues(i, leftCol, baseVal, leftType) != 0
 			// To determine if the right value changed from the base value, we need to convert the base value.
 
-			convertedBaseValue, conflict := Convert(ctx, m.baseVD, m.rightVD, m.rightSchema, i, base, m.ns)
+			convertedBaseValue, conflict := convert(ctx, m.baseVD, m.rightVD, m.rightSchema, baseColumnIndex, rightColumnIndex, base, m.ns)
 			if conflict {
 				return nil, true
 			}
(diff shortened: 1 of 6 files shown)


go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1532,33 +1542,124 @@ func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
...
+			return rightCol, false
+		}
 	}
 
-	if base == nil {
-		// Conflicting insert
-		return nil, true
+	// Since we now know the column existed at base, if either left or right was not in the tuple,
+	// then the column must have either been deleted, it's default-NULL and the value was NULLed.
+	// TODO: For now, assume it was deleted. Is it possible for us to distinguish between these cases?
```
zachmu, line 1586: This doesn't seem right. If the column exists in base but not in one of the two sides, then it matters which side it was deleted on.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1532,33 +1542,124 @@ func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
...
+		return nil, false
 	}
 
-	var baseVal []byte
-	if b := m.baseMapping[i]; b != -1 {
-		baseVal = base.GetField(b)
+	baseColumnIndex := m.baseMapping[i]
+	baseVal := base.GetField(baseColumnIndex)
+
+	// We can now assume that both left are right are modifications to an existing column.
```
zachmu, line 1594: You should expand the method doc to clarify that the contract is: there was a byte-level value changed detected in column i

zachmu (commented):
Didn't read that carefully, but I think this is a good approach that should give substantially better results, even if it makes merges of schema changes slower.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1532,33 +1542,124 @@ func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
...
+		rightModified := rightCol != nil
+
+		switch {
+		case leftModified && rightModified:
+			// conflicting inserts
+			return nil, true
+		case leftModified:
+			return leftCol, false
+		default:
+			return rightCol, false
```
fulghum, line 1580: So, if a row with the same PK was inserted on both sides of the merge, then we look at each column and if one column is nil, then we always select the other column? I'm not sure if that's the right thing to do... it seems like it should just be a conflict that users have to resolve, since we don't know if the value or nil is the right choice to keep. Let me know if I'm just missing something. 😄

go/libraries/doltcore/sqle/index/prolly_fields.go
```diff
@@ -129,6 +130,111 @@ func GetField(ctx context.Context, td val.TupleDesc, i int, tup val.Tuple, ns tr
 	return v, err
 }
 
+// Serialize writes an interface{} to the ith field of the Tuple being built.
+func Serialize(ctx context.Context, ns tree.NodeStore, t val.Type, v interface{}) []byte {
```
fulghum, line 134: Is it possible to use `PutField` or refactor it so it's usable instead of copy'ing pasting it to this new function that returns `[]byte`?

go/libraries/doltcore/merge/merge_prolly_rows.go
nicktobey: You're right.

I also realized that this case currently doesn't need to be handled by `processColumns`, because this function only gets called for columns that exist in the merged schema.

This case has been removed.

go/libraries/doltcore/merge/merge_prolly_rows.go
nicktobey: Method doc was removed in the rewrite.

go/libraries/doltcore/merge/merge_prolly_rows.go
nicktobey: Yeah, that wasn't correct. The new implementation should handle that better.

go/libraries/doltcore/merge/merge_prolly_rows.go
nicktobey: Nevermind, it wasn't. I updated it.

go/libraries/doltcore/sqle/index/prolly_fields.go
nicktobey: Unfortunately, not really.  PutField does enough other things that there was no clean way to separate out just the serialization logic, unless PutField calls Serialize and then switches on the type _again_. It was simpler to just do this.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1532,40 +1544,131 @@ func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
...
+// convert takes the `i`th column in the provided tuple and converts it to the type specified in the provided schema.
+// returns the new representation, and a bool indicating success.
+func convert(ctx context.Context, fromDesc, toDesc val.TupleDesc, toSchema schema.Schema, fromIndex, toIndex int, tuple val.Tuple, originalValue []byte, ns tree.NodeStore) ([]byte, bool) {
+	if fromDesc.Types[fromIndex] == toDesc.Types[toIndex] {
+		// No conversion is necessary here.
+		return originalValue, true
+	}
+	parsedCell, err := index.GetField(ctx, fromDesc, fromIndex, tuple, ns)
+	if err != nil {
+		return nil, true
```
fulghum, line 1663: Looks like this should be returning `false` instead of `true`

It's probably also worth doing something with that `err`. If we're trying to debug a customer issue in this code, having that err logged somewhere could be handy. One idea is grabbing the server log from `ctx` and logging it as a warning or error.

fulghum: Similar comment for the other `err` on line 1667. Seems like the only way to see that currently would be to run the code through the debugger, but it would probably be helpful to have that in the logs or somewhere a customer could easily report to us.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1532,40 +1544,131 @@ func (m *valueMerger) tryMerge(left, right, base val.Tuple) (val.Tuple, bool) {
...
+	var baseColIdx = -1
+	var baseColExists = false
+	if base != nil {
+		baseCol, baseColIdx, baseColExists = getColumn(&base, &m.baseMapping, i)
 	}
-	var rightCol []byte
-	if r := m.rightMapping[i]; r != -1 {
-		rightCol = right.GetField(r)
+	leftCol, leftColIdx, leftColExists := getColumn(&left, &m.leftMapping, i)
+	rightCol, rightColIdx, rightColExists := getColumn(&right, &m.rightMapping, i)
```
fulghum, line 1557: Can `left` and `right` be nil, too? We're checking `base` for nil, but seems like `left` and `right` could also be nil. Would be nice if `getColumn` could handle that just so that the variable initialize code doesn't have to take up 5 lines for each side of the three way merge.

go/libraries/doltcore/sqle/index/prolly_fields.go
```diff
@@ -129,6 +130,68 @@ func GetField(ctx context.Context, td val.TupleDesc, i int, tup val.Tuple, ns tr
 	return v, err
 }
 
+// Serialize writes an interface{} into the byte string representation used in val.Tuple.
+func Serialize(ctx context.Context, ns tree.NodeStore, t val.Type, v interface{}) ([]byte, bool) {
```
fulghum, line 134: I think it's worth seeing if we can avoid all this copied code. It seems like we could change this function to something like:
- create new TupleDesc that only has one field of type `t`
- call `NewTupleBuilder` with that new TupleDesc
- set that field in the new TupleBuilder and call `Build`
- return the bytes of that field

I know that's not going to be quite as efficient as this code, but I think it's a win to have less code / less duplicated code, plus I don't think we need to optimize this particular portion of merge where we merge conflicting rows cell-wise together. There are bigger perf issues with merge that would dwarf anything here anyway. 

Plus, it'll avoid us having to jump on firedrills from customers when they get panics because this function doesn't support all the types yet. For example, the first time a customer hits a cell-wise merge conflict resolution with a float, this will panic and we'll have to jump on it. We could fill in all the other data types now of course, but it just seems like a lot of copied code.

fulghum (commented):
Looks very good. You did a really great job being thorough with the tests! Very good job finding those schema merge edge cases! 🙌🏻

I'm happy with this, but just had a couple small questions I wanted to ask before I hit the approve button.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -89,7 +89,13 @@ func mergeProllyTable(ctx context.Context, tm *TableMerger, mergedSch schema.Sch
 
 		// After we migrate the data on the left-side to the new, merged schema, we reset
 		// the left mapping to an identity mapping, since it's a direct mapping now.
-		valueMerger.leftMapping = val.NewIdentityOrdinalMapping(len(valueMerger.leftMapping))
+		// However, columns that did not exist on the left schema shouldn't be updated, because they
+		// still don't exist on the migrated data.
```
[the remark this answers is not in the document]
nicktobey: > I'm mostly just wondering if the logic will be simpler later if we could always assume left had all the new columns after we migrated the data.

For most cases it doesn't matter, because of what you point out:  when nullable columns are added to the end of the schema, we don't rewrite the table. The logic for using mappings to read from tuples is doesn't care if the mapping points to an index past the end of the tuple or if the mapping contains `-1` (indicating that the column doesn't exist in the tuple.) In both cases, attempting to retrieve the value from the tuple returns nil/Null.

However, there's one case where the distinction matters: when attempting to automatically resolve data conflicts in `processColumn`, we need to be able to differentiate between the case where the left schema has the column (and it's NULL) vs the case where the left schema doesn't have the column. We can't inspect the tuple because null values at the end of the tuple are omitted. And we can't use the schema anymore because by migrating the tuple here it no longer matches the schema. So we need to distinguish these cases in the mapping in order to correctly do data merges.

go/libraries/doltcore/sqle/index/prolly_fields.go
nicktobey: Good idea. I'll give that a shot.

> For example, the first time a customer hits a cell-wise merge conflict resolution with a float

This isn't currently possible because we only allow specific type changes to be mergible, and those are conversions between string types, set types, and enum types. Presumably as we add support for more types, we would add corresponding tests that will fail until we add more conversions here. So I'm not worried about user panics from unsupported types.

But I like your proposal for implementing this on top of PutField. If it works it could save us some trouble down the line.

nicktobey: I think I was initially skeptical because of the inelegance/inefficiency, but I think this was the right approach. Thanks for the suggestion.

go/libraries/doltcore/merge/merge_prolly_rows.go
nicktobey: Good point. I think that we should log the error somehow and then return that the conversion failed, since it likely indicates an unforeseen schema conflict preventing us from converting to the new column type.

I couldn't find anything about a server log, can you point me to it?

nicktobey: On second thought, I'm just going to propagate the error.

It's possible that as we allow more type changes in schema merges, some of these errors should be treated as data conflicts and shouldn't be propagated. But we can treat those on a case-by-case basis by inspecting the type of the error. For now, always propagating the error ensures that we're not suppressing important errors that *aren't* do to type conversion conflicts.

go/libraries/doltcore/sqle/index/prolly_fields.go
fulghum: Sweet! Thanks for trying it out! 😁

fulghum (approved):
Very nice work! Thanks for all the thorough answers to my questions. 🚢

PR #6720, opened 2023-09-26
Fixed a bug in table writers where error state wasn't being appropriately cleared

zachmu:
New tests for this bug in https://github.com/dolthub/go-mysql-server/pull/2035


fulghum (approved):
Looks good! 🚢

PR #6731, opened 2023-09-27
Bug fix: Allow `dolt_checkout()` and `dolt status` when running in read-only mode

fulghum:
The stored procedures for `dolt_checkout` and `dolt_count_commits` weren't marked as being read-only, so when run against a read-only sql-server, they resulted in an error about the database being read-only. 

Fixes: https://github.com/dolthub/dolt/issues/6730

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/init.go
+++ b/go/libraries/doltcore/sqle/dprocedures/init.go
@@ -23,14 +23,14 @@ var DoltProcedures = []sql.ExternalStoredProcedureDetails{
 	{Name: "dolt_add", Schema: int64Schema("status"), Function: doltAdd},
 	{Name: "dolt_backup", Schema: int64Schema("status"), Function: doltBackup},
 	{Name: "dolt_branch", Schema: int64Schema("status"), Function: doltBranch},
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/sqle/dprocedures/init.go
```diff
@@ -23,14 +23,14 @@ var DoltProcedures = []sql.ExternalStoredProcedureDetails{
 	{Name: "dolt_add", Schema: int64Schema("status"), Function: doltAdd},
 	{Name: "dolt_backup", Schema: int64Schema("status"), Function: doltBackup},
```
timsehn, line 24: This one should probably work in read only mode too.

timsehn (approved):
Clone and remote we may also want to work in read-only mode.

This fixes the direct bug so we could discuss the other ones.

fulghum:
> Clone and remote we may also want to work in read-only mode.

I was on the fence about clone – I could see it being marked as supported for read-only since it doesn't write to any *existing* databases, but at the same time, it does modify the data in the sql-server. Could be handy though. 

Managing remotes seems like it may make more sense. Adding a new remote and pushing to it, seems like a valid read only use case.

PR #6738, opened 2023-09-28
migrate dolt push to use sql queries

stephkyou:
This change updates `dolt push` to use the appropriate sql engine to generate results.


go/cmd/dolt/commands/push.go
```diff
@@ -87,93 +89,128 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+	}
+	if len(rows) == 0 {
+		return "", env.ErrNoRemote
+	}
+	if len(rows[0]) == 0 {
+		return "", env.ErrNoRemote
+	}
+	if len(rows[0]) == 1 {
+		return rows[0][0].(string), nil
+	}
```
macneale4, line 207: if it's not 0 len, it will always be len == 1 because of you're query specifying the single column. Probably don't need to check for len == 1.

go/cmd/dolt/commands/push.go
```diff
@@ -87,93 +89,128 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+	for _, row := range rows {
+		if row[0].(string) == "origin" {
+			return row[0].(string), nil
```
macneale4, line 210: return "origin", nil

Is more readable.

go/libraries/doltcore/sqle/dprocedures/dolt_push.go
```diff
@@ -83,7 +83,7 @@ func doDoltPush(ctx *sql.Context, args []string) (int, error) {
 	if err != nil {
 		switch err {
 		case doltdb.ErrUpToDate:
-			return cmdSuccess, nil
+			return cmdSuccess, err
```
macneale4, line 86: why was this nil before? Seems like this could change some tests, but I suspect you are working through that.

stephkyou: it makes some bats tests a little more wordy, needing to catch the up to date error in places where we didn't have to catch anything before. will probably also need a minor version bump for this change since we're returning an error where we didn't before.

macneale4: But being up-to-date means that it's a no-op, right?

integration-tests/bats/sql-push.bats
```diff
@@ -212,7 +212,7 @@ SQL
     dolt sql -q "call dolt_push('-u', 'origin', 'other')"
     # upstream should be set still
     run dolt push
-    [ "$status" -eq 0 ]
+    [ "$status" -eq 1 ]
```
macneale4, line 215: If everything is up to date, exit status 0 is the right thing to do. I verified that this is what git push does.

go/cmd/dolt/commands/push.go
```diff
@@ -199,20 +167,111 @@ func getDefaultRemote(sqlCtx *sql.Context, queryist cli.Queryist) (string, error
...
+			return "origin", nil
 		}
 	}
 	return "", env.ErrCantDetermineDefault
 }
 
+// processFetchSpecs takes a string of fetch specs and returns the destination ref and remote ref
+// Assumes the fetch specs look something like: ["refs/heads/*:refs/remotes/origin/*"]
+func processFetchSpecs(fetchSpecs string, branch string) (destRef, remoteRef string) {
+	// TODO: replace string parsing with better logic when more JSONDocument functionality is available
```
macneale4, line 184: I don't know what JSONDocument functionality is. Is that something that's actually planned?

stephkyou: it looks like there's some functions that haven't been implemented yet in GMS for `JSONDocument`, i'm not sure what the priority is to get those implemented.

macneale4: I'd just drop that comment. I don't think that is a TODO that will ever get done, and it will probably be so far in the future that the landscape will have changed.

go/cmd/dolt/commands/push.go
```diff
@@ -87,93 +93,212 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
-		err := dEnv.RepoState.Save(dEnv.FS)
+	return
+}
+
+// printPushMessage prints the appropriate message for the given push output
+func printPushMessage(rows []sql.Row) error {
+	var statusCode int64
+	if intCode, ok := rows[0][0].(int64); ok {
+		statusCode = intCode
+	} else if strCode, ok := rows[0][0].(string); ok {
```
macneale4, line 215: Are the types not agreeing between local and remote executions? I would think the string version would come back as an integer of some type - is it not?

macneale4 (changes requested):
A couple questions, but this is getting close. Also, wanted to ask if the new tests you wrote pass without any of your changes.

go/cmd/dolt/commands/push.go
stephkyou: Yeah it seems like the remote execution is coming back as a string, it won't cast as an int.

go/cmd/dolt/commands/push.go
```diff
@@ -87,93 +93,212 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
...
+	cli.Print(spinner.next() + " Uploading...")
+	defer func() {
+		cli.DeleteAndPrint(len(" Uploading...")+1, "")
+	}()
+
+	for {
+		select {
+		case err := <-errChan:
+			return handlePushError(err, usage, apr, queryist, sqlCtx)
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
stephkyou: I'm not super sure but from what I understand if someone attempts to ^C a push it will result in Done().

go/cmd/dolt/commands/push.go
macneale4: Can you add a comment to this effect? I don't know what the root cause is, and I'm not going to ask you to sort it out. But this will confuse future readers

go/cmd/dolt/commands/push.go
```diff
@@ -138,6 +138,14 @@ func (cmd PushCmd) Exec(ctx context.Context, commandStr string, args []string, d
 		case err := <-errChan:
 			return handlePushError(err, usage, apr, queryist, sqlCtx)
 		case <-ctx.Done():
+			switch ctx.Err() {
+			case context.DeadlineExceeded:
+				cli.Println("timeout exceeded")
+				return 1
+			case context.Canceled:
+				cli.Println("push cancelled by force")
+				return 1
```
macneale4, line 147: There should be a default too. Don't want to fall through is you don't have a matching case.

macneale4 (approved):
One small change requested on the Done() handling. Also there is a request to add a comment. You can address those things and ship it!

PR #6740, opened 2023-09-28
Support for `DOLT_COMMITTER_DATE` and `DOLT_AUTHOR_DATE` environment vars

zachmu:
These two new environment vars perform the same function as `GIT_COMMITTER_DATE` and `GIT_AUTHOR_DATE`, respectively. They set the two timestamps associated with newly created commits on `dolt commit` and `dolt merge`.

Setting these two environment variables allows the deterministic creation of commit hashes, as requested in https://github.com/dolthub/dolt/issues/6724


nustiueudinastea:
Hey @zachmu, this is awesome stuff! I was wondering, will it be possible to set the committer date programmatically somehow or only as an env variable via dolt cli? It would be very useful to have something like the '--date' parameter for the DOLT_COMMIT procedure.

zachmu:
> Hey @zachmu, this is awesome stuff! I was wondering, will it be possible to set the committer date programmatically somehow or only as an env variable via dolt cli? It would be very useful to have something like the '--date' parameter for the DOLT_COMMIT procedure. Perhaps an extra parameter on the same commit procedure.

We are copying the git interface for now. If you check the bats tests in this PR, you'll see an example of setting the env var for the context of a single invocation. That's how I would approach this for now.

We'll do another pass at a later date, including making this possible to use in the sql server context.

reltuk (approved):
LGTM!

PR #6744, opened 2023-09-28
go/store/nbs: store.go: Clean up how we update hasCache so that we only update it after successfully writing the memtable.

reltuk:


reltuk:
An alternative approach compared to #6743

zachmu (approved):
LGTM, I prefer this approach as well

PR #6753, opened 2023-09-29
adds message column to dolt_push output

stephkyou:
Adds a `message` column to the output of `dolt_push()` to provide additional information.


macneale4 (approved):
LGTM. I'm curious to see if this break other tests.

PR #6754, opened 2023-09-29
go/libraries/doltcore/sqle/cluster: Replicate `DROP DATABASE` statements so that a dropped database is also dropped on replicas.

reltuk:


go/libraries/doltcore/sqle/cluster/replication_service.go
```diff
@@ -38,6 +40,8 @@ type replicationServiceServer struct {
 
 	branchControl        BranchControlPersistence
 	branchControlFilesys filesys.Filesys
+
+	dropDatabaseProvider func(context.Context, string) error
```
zachmu, line 44: Unfortunate name since `databaseProvider` has a well established meaning

Maybe `dropDatabaseHandler`

go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -94,6 +96,8 @@ type Controller struct {
 	branchControlController *branch_control.Controller
 	branchControlFilesys    filesys.Filesys
 	bcReplication           *branchControlReplication
+
+	dropDatabaseProvider func(context.Context, string) error
```
zachmu, line 100: Same comment about name

zachmu (approved):
`DropDatabaseProvider` is an unfortunate name, otherwise LGTM!

PR #6760, opened 2023-10-02
dolt backup: Fix dropped error on the final stage of updating the destination with the newly pushed backup.

reltuk:


go/libraries/doltcore/env/actions/remotes.go
```diff
@@ -679,7 +679,22 @@ func SyncRoots(ctx context.Context, srcDb, destDb *doltdb.DoltDB, tempTableDir s
...
+			destRoot, err = destDb.NomsRoot(ctx)
+			numRetries += 1
+		}
+	}
+	if err != nil {
+		return err
+	}
+
+	if !success {
+		return errors.New("could not destination root to the same value as this database's root. the destination database received too many writes while we were pushing and we exhausted our retries.")
```
zachmu, line 696: could not set

zachmu (approved):
LGTM!

PR #6762, opened 2023-10-03
Feature: New stored procedure `dolt_undrop()` to restore dropped databases

fulghum:
This change introduces a new stored procedure, `dolt_undrop(<database name>)` that can be used to restore a dropped database. Dropped databases are now moved to a new `dolt_dropped_databases` directory until they are either restored or permanently purged with the new `dolt_purge_dropped_databases` stored procedure.

**Usage:**
```sql
# Use dolt_undrop() with the database name to restore a dropped database
drop database mydb;
call dolt_undrop('mydb');
...
# Use dolt_purge_dropped_databases() with no arguments to permanently delete all dropped databases
drop database mydb;
call dolt_purge_dropped_databases;
```

**Required Privileges:** Calling `dolt_undrop()` does not require any special privileges. Calling `dolt_purge_dropped_databases()` requires `SUPER` privileges. 

**Testing Notes:** Tests are in BATS, since Dolt enginetests use `InMemoryBlobStore` instead of writing db data to a filesystem, so there isn't currently a good way to undrop databases in enginetests. If the noms data files were stored through the `InMemFS`, then the same code would work in our tests to undrop databases. Currently the repo state files are able to be undropped, but not the actual database data, since it disappears with the `InMemoryBlobStore` reference when we drop the database. 

Fixes: https://github.com/dolthub/dolt/issues/6484

Documentation Updates: https://github.com/dolthub/docs/pull/1768


go/libraries/doltcore/sqle/database_provider.go
```diff
@@ -676,20 +612,101 @@ func (p *DoltDatabaseProvider) DropDatabase(ctx *sql.Context, name string) error
...
+
+// registerNewDatabase registers the specified DoltEnv, |newEnv|, as a new database named |name|. This
+// function is responsible for instantiating the new Database instance and updating the tracking metadata
+// in this provider. If any problems are encountered while registering the new database, an error is returned.
+func (p *DoltDatabaseProvider) registerNewDatabase(ctx *sql.Context, name string, newEnv *env.DoltEnv) (err error) {
+	// This method MUST be called with the provider's mutex locked. TryLock allows us to validate that the
+	// mutex is locked (without actually locking it and causing a deadlock) and to error out if we detect
+	// that the mutex is NOT locked.
+	if p.mu.TryLock() {
+		defer p.mu.Unlock()
```
zachmu, line 665: This seems like an anti-pattern. Fine to check if the lock is held, but if we didn't take the lock ourselves we probably shouldn't be releasing it either. Up to the caller

go/libraries/doltcore/sqle/dropped_databases.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		destinationDirectory = filepath.Join(newSubdirectory, file)
+	} else {
+		destinationDirectory = filepath.Join(droppedDatabaseDirectoryName, file)
+	}
+
+	// Add the final directory segment and convert all hyphens to underscores in the database directory name
+	dir, file := filepath.Split(destinationDirectory)
+	if strings.Contains(file, "-") {
+		destinationDirectory = filepath.Join(dir, strings.ReplaceAll(file, "-", "_"))
```
zachmu, line 92: This means that the drop / undrop won't round-trip, which isn't great

There can also be other characters (anything non-alphanum) that get converted in a db name from its directory name, seems odd to special-case only the hyphen

zachmu: Maybe this is finally the time to stop messing with database names and just use the directory name verbatim (would be a breaking change, but I think it's a good idea)

go/libraries/doltcore/sqle/dropped_databases.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (dd *droppedDatabaseManager) ListDroppedDatabases(_ *sql.Context) ([]string, error) {
+	if err := dd.initializeDeletedDatabaseDirectory(); err != nil {
+		return nil, fmt.Errorf("unable to list undroppable database: %w", err)
+	}
+
+	databaseNames := make([]string, 0, 5)
+	callback := func(path string, size int64, isDir bool) (stop bool) {
+		_, lastPathSegment := filepath.Split(path)
+		lastPathSegment = strings.ReplaceAll(lastPathSegment, "-", "_")
```
zachmu, line 178: Same comment about naming, this is going to break for some database names

Our sanitization of db names can't be round-tripped safely, seems silly to try

zachmu (approved):
This mostly looks good, but consider unwinding the mess of renaming dirs to db names, maybe as a followup.

The other non-trivial comment is that it seems pretty useful to be able to test this in engine tests as well as bats. I think this is not too hard to pull off, just need some methods on `DoltHarness` to make it use an on-disk filesystem instead of an in-mem one.

go/libraries/doltcore/sqle/database_provider.go
fulghum: This is probably worth a comment since it's not a common pattern... `TryLock` will attempt to lock the mutex and if it is able to, then it executes this block. That means we need to error out, but we've also grabbed the mutex in the process of calling `TryLock` to test the mutex, so this function needs to release the lock and not just leave it locked. Pulling this out into a helper function would probably make the coder a little easier to read and avoid needing to explain that behavior in multiple places. 

I'm not committed this to this approach either, but it seems like doing something to ensure the lock is held is worth it.

fulghum: Cleaned this up in a47c4f8b7c656d302b307f9c3cce1413bc51a929. That removed some duplication, will make that util function easier to reuse in the future, and the comments should better explain why the lock needs to be released in that function (since it was grabbed in that function).

fulghum:
> The other non-trivial comment is that it seems pretty useful to be able to test this in engine tests as well as bats. I think this is not too hard to pull off, just need some methods on `DoltHarness` to make it use an on-disk filesystem instead of an in-mem one.

Great suggestion! I had briefly poked at swapping in a filesystem based blob-store that used the in-mem fs for tests, but it's **much** simpler just to switch the harness completely over to a local filesystem for these tests and sidestep the issue with in-memory blob store. I've got something hacked up and working now and will get it cleaned up and pushed.

go/libraries/doltcore/sqle/dropped_databases.go
fulghum: Regarding round-tripping... I think the most important requirement is that the database have the same logical/SQL name when it's restored. That would give us correct round-tripping from the SQL interface. I don't think round-tripping to the exact same physical directory name/location on disk is needed though, but it was something I was wondering about. Let me know if you see any potential issues with that. 

Good callout on non-alphanumeric chars also being translated. I'll find where we're handling that today for database names and see if I can pull something out to reuse. 

Stopping the logical DB renaming from physical directory name seems like a good idea. I think it's best done as a separate PR though – we should retain the old functionality for a while anyway, otherwise older DBs will not be able to load with newer Dolt versions (like we hit the other day).

go/libraries/doltcore/sqle/dropped_databases.go
fulghum: Fixed in d976b46b0d55c74853c37ad3477252d75c25f17b – we now normalize the db name (using the same code we use when we load a multi repo env and translate physical dir names to logical SQL db names) when it's moved to the dropped databases dir, so on restore, we don't need to try and reverse map it. Also added more test coverage for directory names with unsupported chars.

PR #6767, opened 2023-10-03
go: sqle: cluster: Improvements for DROP DATABASE replication.

reltuk:
Fix a bug in session handling for the replication api endpoint which would prevent a dropped database from being recreated on a replica.

Fix a race condition when a database is recreated after it is dropped. In that case, we stop attempting to replicate the drop, so that it does not replicate after the new database does.

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/server.go
+++ b/go/cmd/dolt/commands/sqlserver/server.go
@@ -278,85 +278,82 @@ func Serve(
 	var remoteSrv *remotesrv.Server
 	if serverConfig.RemotesapiPort() != nil {
 		port := *serverConfig.RemotesapiPort()
-		if remoteSrvSqlCtx, err := sqlEngine.NewDefaultContext(ctx); err == nil {
-			listenaddr := fmt.Sprintf(":%d", port)
-			args := sqle.RemoteSrvServerArgs(remoteSrvSqlCtx, remotesrv.ServerArgs{
-				Logger:         logrus.NewEntry(lgr),
-				ReadOnly:       true,
-				HttpListenAddr: listenaddr,
-				GrpcListenAddr: listenaddr,
-			})
-
-			ctxFactory := func() (*sql.Context, error) { return sqlEngine.NewDefaultContext(ctx) }
-			authenticator := newAuthenticator(ctxFactory, sqlEngine.GetUnderlyingEngine().Analyzer.Catalog.MySQLDb)
-			args = sqle.WithUserPasswordAuth(args, authenticator)
-
-			args.TLSConfig = serverConf.TLSConfig
-			remoteSrv, err = remotesrv.NewServer(args)
-			if err != nil {
-				lgr.Errorf("error creating remotesapi server on port %d: %v", port, err)
-				startError = err
-				return
-			}
-			listeners, err := remoteSrv.Listeners()
-			if err != nil {
-				lgr.Errorf("error starting remotesapi server listeners on port %d: %v", port, err)
-				startError = err
-				return
-			} else {
-				go remoteSrv.Serve(listeners)
-			}
-		} else {
+		listenaddr := fmt.Sprintf(":%d", port)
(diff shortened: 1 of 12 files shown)


go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -278,85 +278,82 @@ func Serve(
...
 
-			go clusterRemoteSrv.Serve(listeners)
-			go clusterController.Run()
+		clusterRemoteSrv, err = remotesrv.NewServer(args)
+		if err != nil {
+			lgr.Errorf("error creating remotesapi server on port %d: %v", *serverConfig.RemotesapiPort(), err)
+			startError = err
+			return
+		}
+		clusterController.RegisterGrpcServices(sqlEngine.NewDefaultContext, clusterRemoteSrv.GrpcServer())
```
zachmu, line 340: Wary of using NewDefaultContext. It's not really good for much of anything that context.Background isn't, beside having a different type. In particular it doesn't have a real session in it, which you need for most I/O.

Can we get away with a `context.Context` everywhere?

zachmu (approved):
What's here is fine, but if we're going to introduce state tracking for in-progress operations, my suggestion is either do it as simple as possible (prevent more than one concurrent operation), or all the way (add a queue).

go/cmd/dolt/commands/sqlserver/server.go
reltuk: I...don't think we should do that. In particular, there's a lot of SQL-y stuff that we're calling into, and we need a `*sql.Context` in various places, even when we don't have a corresponding `mysql.Conn`. I don't think the right approach is to try to move to `context.Context` everywhere we need to do something like that. Another place this comes up is event execution. In reality, our session factory and the sql engine itself has to be prepared to configure contexts and execute things on behalf of programatic interactions which are not coming from a mysql.Conn.

reltuk: I be fair, NewDefaultContext is good at things like:

* Having a logger
* Theoretically interacting nicely with session tracking, (show full processlist? kill ...? etc.), although that could be better.
* Having a session, which can track session local system variables and can allow for access to mutable global system variables, etc.

Lots of things really...

go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -64,6 +64,15 @@ const RoleDetectedBrokenConfig Role = "detected_broken_config"
 
 const PersistentConfigPrefix = "sqlserver.cluster"
 
+// State for any ongoing DROP DATABASE replication attempts we have
+// outstanding. When we create a database, we cancel all on going DROP DATABASE
+// replication attempts.
+type databaseDropReplication struct {
```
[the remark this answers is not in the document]
reltuk: I think ideally a different approach would be possible, but I don't think it's particularly near at hand. The current approach is quite near at hand and is probably good enough. In particular:

1. I considered blocking or failing creation until the drop was fully replicated. But it's not particularly compelling because the replica can be down for an arbitrary amount of time, in which case you're needlessly taking an availability hit on the primary. There doesn't seem to be a compelling reason to do this when it's easy enough to just cancel the drop database replication.
2. A queue of replication events isn't really what you want. When a replica comes back online, you want to true it up to whatever the current state is, without going through all the intermediate steps which: (a) are arbitrarily large, (b) are irrelevant to any given client and are ideally never seen by a client, (c) are more expensive to enact than just replicating the current state.

The presented approach basically shoots for having at most one ongoing replication attempt for a given database name. Either we're not replicating anything for that name, or we're replicating a delete, or we're replicating a new head. An alternative approach which I considered was to put the delete replication itself into the commithook, and to have the commit hook remove itself, or something like that, once it was successfully replicated, and to have the initdatabasehook cancel those things before adding the new commit hooks (or just set existing commit hooks back to replicating the new head...). It would look somewhat similar, but also somewhat different. commithook is currently quite large and quite subtle, so I decided to forego adding even more complications there for the time being and keep the drop database replication outside of it for now.

PR #6768, opened 2023-10-03
[no-release-notes] test for joining diff table function with alias

JCOR11599:
It looks like this issue was fixed, but there wasn't a test for this specific case.
This PR adds tests for joining two `diff_table_function(...)` with alias.

fixes https://github.com/dolthub/dolt/issues/4798


max-hoffman (approved):
lgtm, i think we might test this on GMS side (hopefully)

PR #6769, opened 2023-10-03
[no-release-notes] go: sqle: DoltDatabaseProvider: Change it to a pointer receiver so updates to its state are much easier to reason about.

reltuk:


fulghum (approved):
Looks great! 🚢

PR #6787, opened 2023-10-09
add `--all` flag for dolt push

jennifersp:
This PR adds `--all` flag to `dolt push` command and update some error messages as well as success push messages to be returned. 
Also includes:
- removing some duplicate tests from `dolt-push.bats`
- splitting remote dolt push and pull tests from `remotes.bats` into `remotes-push-pull.bats`


go/libraries/doltcore/env/remotes.go
```diff
@@ -114,105 +133,184 @@ func (r *Remote) GetRemoteDBWithoutCaching(ctx context.Context, nbf *types.NomsB
...
-	remotes, err := rsr.GetRemotes()
+func NewPushOpts(ctx context.Context, apr *argparser.ArgParseResults, rsr RepoStateReader, ddb *doltdb.DoltDB, force, setUpstream, pushAutoSetupRemote, all bool) ([]*PushOpts, Remote, error) {
+	if apr.NArg() == 0 {
+		return getPushOptsAndRemoteFromNoArg(ctx, rsr, ddb, force, setUpstream, pushAutoSetupRemote, all)
+	}
+
+	rsrBranches, err := rsr.GetBranches()
 	if err != nil {
-		return nil, err
+		return nil, NoRemote, err
```
fulghum, line 162: (minor/style) You might consider returning `*Remote` instead of a copy. The pointer allows you to easily return `nil` without having to use special-case instances like `NoRemote`, plus it's slightly more efficient with less memory copying, but that's not a huge deal here. It's not a big deal and really more of style/preference.

integration-tests/bats/remotes-push-pull.bats
```diff
@@ -0,0 +1,845 @@
...
+    dolt sql -q "CREATE TABLE test (pk INT PRIMARY KEY, col1 VARCHAR(10))"
+    dolt add .
+    dolt commit -am "create table"
+    dolt checkout -b branch1
+    dolt sql -q "INSERT INTO test VALUES (1, '1')"
+    dolt commit -am "add 1s"
+    run dolt push --all origin  # should not set upstream for new branches
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ " * [new branch]          branch1 -> branch1" ]] || false
+    [[ ! "$output" =~ "branch 'branch1' set up to track 'origin/branch1'." ]] || false
```
fulghum, line 646: Do we have any other ways to see the default upstream tracking branch? Does that show up in `dolt config --list --local`? This negative string matching is just a little fragile, since only this exact string will fail the test, so if the message changes slightly, we wouldn't catch it that a default upstream had been set. Being able to check that more directly would help make this test a little sturdier.

fulghum (commented):
Logic and tests look good! My comments were all generally around some more polishing. Let me know if anything doesn't make sense.

integration-tests/bats/remotes-push-pull.bats
jennifersp: Currently, the branch config is not shown with `dolt config --list`. I think is only shows ones in `Valid configuration variables`, which can be found with `dolt config -h`.
I updated it to more specific check for upstream set for this branch.

integration-tests/bats/sql-pull.bats
```diff
@@ -106,7 +106,7 @@ teardown() {
     cd repo2
     run dolt sql -q "call dolt_pull('unknown')"
     [ "$status" -eq 1 ]
-    [[ "$output" =~ "unknown remote" ]] || false
+    [[ "$output" =~ "fatal: 'unknown' does not appear to be a dolt repository" ]] || false
```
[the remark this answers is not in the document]
jennifersp: I updated it to `fatal: remote 'unknown' not found` error message instead of `fatal: 'unknown' does not appear to be a dolt repository`

go/libraries/doltcore/env/remotes.go
```diff
@@ -53,10 +53,8 @@ var ErrCurrentBranchHasNoUpstream = goerrors.NewKind("fatal: The current branch
 	"\tdolt push --set-upstream %s %s\n" +
 	"To have this happen automatically for branches without a tracking\n" +
 	"upstream, see 'push.autoSetupRemote' in 'dolt config --help'.")
-var ErrInvalidRepository = goerrors.NewKind("fatal: '%s' does not appear to be a dolt repository\n" +
-	"fatal: Could not read from remote repository.\n\n" +
-	"Please make sure you have the correct access rights\n" +
-	"and the repository exists.")
+var ErrInvalidRepository = goerrors.NewKind("fatal: remote '%s' not found.\n" +
+	"Please make sure the repository exists.")
```
fulghum, line 57: (super minor) "remote" or "database" would be a little better than "repository" here (mostly since we're trying to train people to think "database" instead of "repository").

```suggestion
	"Please make sure the remote exists.")
```

go/libraries/doltcore/env/remotes.go
```diff
@@ -133,17 +131,20 @@ func (r *Remote) GetRemoteDBWithoutCaching(ctx context.Context, nbf *types.NomsB
 	return doltdb.LoadDoltDBWithParams(ctx, nbf, r.Url, filesys2.LocalFS, params)
 }
 
-type PushMeta struct {
-	Opts   []*PushOpts
+// PushOptions contains information needed for push for
+// one or more branches or a tag for a specific remote database.
+type PushOptions struct {
+	Opts   []*PushTarget
```
fulghum, line 137: Mind changing `Opts` to `Targets`? That'll read a little better. Looks like there are only a couple of references to `Opts` so should be easy to change.

go/libraries/doltcore/env/remotes.go
```diff
@@ -133,17 +131,20 @@ func (r *Remote) GetRemoteDBWithoutCaching(ctx context.Context, nbf *types.NomsB
...
-	Remote Remote
+	Remote *Remote
 	SrcDb  *doltdb.DoltDB
 	DestDb *doltdb.DoltDB
 	TmpDir string
 }
 
-type PushOpts struct {
+// PushTarget contains information needed for push per branch or tag.
+type PushTarget struct {
```
fulghum, line 147: Yay! Thank you! I think these type names make a lot more sense now.

fulghum (approved):
Looks great! Thanks for that last round of changes from feedback.

PR #6798, opened 2023-10-11
Testing bumping default server version to `8.0.33-Dolt`

fulghum:


fulghum:
Closing now that vitess change is bumping into Dolt

PR #6803, opened 2023-10-11
Feature: Add a `flake.nix` file for Nix users

sbrow:
# Overview
As discussed with @timsehn in discord, this pull request allows [Nix](https://nixos.org/) users to build dolt using the nix package manager. This behavior can be verified by running:

```bash
# acquires, builds, and runs the latest version of dolt.
nix --experimental-features 'nix-command flakes' run github:sbrow/dolt --
```
Note: if you are *not* a nix user, you will want to run `nix-store --gc` afterwards to clean up the packages, as this will download and install a separate copy of the go toolchain.

# Maintenance

The issues with a PR like this lie in the added maintenance cost of needing to frequently update the package files in order for the build to continue working. In this case there is a `vendorHash` field in `flake.nix` that needs to be valid in order for the build to succeed, and is dependent upon the following globs:

```bash
go/**
flake.{nix,lock}
.gitignore
```
 
Since the files in the  `go` folder are likely to change frequently, this is a pretty significant issue.

As far as I can tell, the only way to generate this hash is to run `nix build`, copy the correct hash from the error message into `vendorHash`, and then run `nix build` again. Any commit that doesn't have this process run on it won't work with nix.

There are a few ways around this:
1. **Run a script in a pre-commit hook to ensure that the commit is valid.** This would require **all committers** to have nix installed, which seems like a pretty bad idea. 
2. **Create a GitHub action**. The action would run the script and add a new commit on top of the previous commit if the hashes don't match. However, this would mean running the build in an ephemeral machine for every commit, which would be pretty time consuming unless you cached some of `/nix/store`.
3. **Vendorize the build dependencies**. According to [the docs](https://nixos.org/manual/nixpkgs/stable/#sec-language-go), if you use `go mod vendor`,  then you can set `vendorHash = null; ` and it becomes a non-issue. But vendoring all of your dependencies has issues of its own.

# Conclusion
Adding nix support has some serious downsides. As a NixOS user, I think it's an amazing tool, but maintaining support for it is not trivial, and it's hard to advocate running nix alongside another build chain rather than a replacement of it, as you're effectively duplicating all of your dependencies.

Let me know what you think, and if it's something you'd be willing to pursue.


reltuk:
@sbrow Thank you for the contribution. Sorry for the delay in my response – I was traveling last week.

I love Nix, although I'm not too familiar with nix flakes – it's admittedly been a while since I used it in anger. In general, I would love to see better support for `dolt` in the nix ecosystem, and I appreciate your suggestion and your documentation of the tradeoffs above.

On the balance, given some of the tradeoffs, we're not currently ready to put a `flake.nix` and the lock file itself in the root of the `dolt` repository. I don't want to lose this work though, and I do want to make `dolt` as easy as possible to consume from `nix`. So, I made the following proof-of-concept, which might be a middle ground for now:

https://github.com/dolthub/dolt-nix-flake

We can have a separate public repository where we publish the nix flake. That flake can take its source from `fetchFromGitHub` instead of the local file tree.

Eventually I would like to:
1) make a small CI workflow that bumps `dolt-nix-flake/main`'s `flake.nix` every time `dolt/main` receives a commit. 
2) make a small CI workflow that creates a release commit and a release in `dolt-nix-flake` for any new dolt release.
3) manually create a few dolt-nix-flake releases corresponding to dolt releases going back a few weeks/months in time.

In order to hopefully make those things achievable in the near-term, I took a pass at writing a small program which would render an appropriately `flake.nix` file for a given `dolt` commit. It's admittedly pretty gross, but it works for me in some local testing. Hopefully hooking up the CI machinery won't be too hard from there.

Does this sound like a useful middle ground? I realize it's not *as* awesome – having the `flake.nix` in the `dolt` repository itself let consumers easily take a dependency on any commit hash directly from the `dolt` repository, and it would let them take non-main branch heads and the likes. It's definitely not perfect. But I do think it might still be useful to people. I would love to hear you thoughts.

Thanks again,
Aaron

sbrow:
@reltuk You guys are the nicest team of developers I've ever interacted with on GitHub!

After attempting to keep my fork in line with this repo, I was actually going to recommend the same thing you suggested, as CI is pretty much a requirement for making this feature run smoothly.

I think having a flake.nix available to nix users is a big :+1: , but maintaining a `flake.nix` file in the main repo doesn't make sense unless everyone on your team is using nix flakes, or you're using [Hydra](https://github.com/NixOS/hydra).

Part of the beauty of flakes is that it makes it easy to version control remote dependencies, so the workflow you outlined works perfectly. I'll make a PR to [dolthub/dolt-nix-flake](https://github.com/dolthub/dolt-nix-flake) to make the flake a bit more idiomatic.

Perhaps once this is off the ground, you could link to it from somewhere in the README, or the docs.

Really love your whole team's positive responses and blazing fast response times :heart:

Feel free to close this PR, as we can move the conversation to [dolthub/dolt-nix-flake](https://github.com/dolthub/dolt-nix-flake), and thanks for being positive, (but reasonable) to this request!

reltuk:
Thanks for the great followup. Closing this for now and following up in the new repo.

PR #6810, opened 2023-10-12
Prevent `dolt_checkout` from creating new branches in read-only databases

fulghum:
Fixes: https://github.com/dolthub/dolt/issues/6785


JCOR11599 (approved):
LGTM!

PR #6822, opened 2023-10-16
Add defaults for dolt table row counts

max-hoffman:
We could do a better job keeping track of row counts for special tables. In the meantime, we want tables to implement `sql.StatisticsTable` and return row count values high enough that we use HASH_JOIN instead of INNER_JOIN.


go/libraries/doltcore/schema/data_length.go
```diff
@@ -0,0 +1,49 @@
...
+
+func SchemaAvgLength(schema sql.Schema) uint64 {
+	var numBytesPerRow uint64 = 0
+	for _, col := range schema {
+		switch n := col.Type.(type) {
+		case sql.NumberType:
+			numBytesPerRow += 8
+		case sql.StringType:
+			numBytesPerRow += uint64(n.MaxByteLength())
+		case sqltypes.BitType:
```
zachmu, line 30: Not really what the bit type does

zachmu: Probably want to treat it the same as an int

https://dev.mysql.com/doc/refman/8.0/en/bit-type.html

zachmu (approved):
LGTM, seems like the average length func could use a little touchup

One thought on costing: you need to be treating refed types like blobs and text differently than normal strings, since the cost of fetching them is astronomically higher. Probably worth a second pass at some point

go/libraries/doltcore/schema/data_length.go
max-hoffman: whole function was preexisting, i can update this one though

max-hoffman:
Costing will get an actual average row length for text and blobs, it will be a lot more accurate.

PR #6826, opened 2023-10-16
Support `AS OF` with `dolt_ignore` table.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6823

To make this work cleanly, this PR creates a new interface, `VersionedTable`, which is implemented by DoltTable and classes that wrap it (Like `WriteableDoltTable` and `IgnoreTable`.)


tbantle22 (approved):
Could use test coverage but otherwise lgtm - thanks for fixing!

PR #6827, opened 2023-10-17
Automatically create `dolt_docs` table

tbantle22:
Currently if you are using a SQL-only interface like Hosted, you have to manually create the `dolt_docs` system table to use it. This changes that functionality to be the same as the [`dolt_ignore` system table](https://github.com/dolthub/dolt/pull/5809), which is created on writes and returns an empty index on reads. The table will always exist as far as the mysql engine is concerned.

Related to https://github.com/dolthub/dolt/issues/6809


zachmu (approved):
Looks great, no notes

PR #6831, opened 2023-10-17
go: sqle: cluster: When performing a graceful transition to standby, take mysql and dolt_branch_control replication state into account.

reltuk:
Graceful transitions to standby block on the primary until a certain number of replicas are trued up. They then return a status of whether each database on each replica is caught up, so that a control plane agent can pick a caught up server to be next primary, for example.

The newly added replication of users and grants and branch control permissions did not used to participate in this graceful transition logic. This makes it so that the state of mysql and dolt_branch_control replication is reported in the rows returned from `call dolt_cluster_transition_to_standby` and those databases must also be caught up on a replica for the replica to be considered fully caught up.


zachmu (approved):
Seems pretty reasonable@

PR #6833, opened 2023-10-17
JSON and stats interface refactor

max-hoffman:
Bring Dolt into line with the new stats and json interfaces.


go/libraries/doltcore/sqle/json/noms_json_value_test.go
```diff
@@ -154,14 +154,17 @@ func TestJSONCompare(t *testing.T) {
...
 	for _, test := range tests {
 		name := fmt.Sprintf("%v_%v__%d", test.left, test.right, test.cmp)
 		t.Run(name, func(t *testing.T) {
 			left, right := MustNomsJSON(test.left), MustNomsJSON(test.right)
-			cmp, err := left.Compare(ctx, right)
+			cmp, err := gmstypes.CompareJSON(left, right)
 			require.NoError(t, err)
-			assert.Equal(t, test.cmp, cmp)
+			if test.left == `[1,2]` && test.right == `[1,9]` {
+				assert.NotEqual(t, 0, cmp)
```
fulghum, line 164: A little strange that we have to special case this one test in here. `CompareJSON` doesn't return `1` for this case and we can't update `test.cmp` in the test case definition?

fulghum (approved):

go/libraries/doltcore/sqle/json/noms_json_value_test.go
max-hoffman: if you scroll up there's a note about how mysql array ordering is non-deterministic, only guarantee is not 0. i guess i could change the answer to -1. I'm also pretty sure this code isn't used anywhere, would be nice to deprecate most of the noms types.

PR #6834, opened 2023-10-18
migrate dolt ls to use sql queries

stephkyou:
This change updates `dolt ls` to use the appropriate sql engine to generate results. This change also creates a system variable that, when enabled, shows dolt system tables in `show tables` and `information_schema.tables`.

Resolves: https://github.com/dolthub/dolt/issues/2073


integration-tests/bats/ls.bats
```diff
@@ -0,0 +1,199 @@
...
+    [[ "$output" =~ "dolt_diff_t1" ]] || false
+    [[ "$output" =~ "dolt_commit_diff_t1" ]] || false
+    [[ "$output" =~ "dolt_constraint_violations_t2" ]] || false
+    [[ "$output" =~ "dolt_history_t2" ]] || false
+    [[ "$output" =~ "dolt_conflicts_t2" ]] || false
+    [[ "$output" =~ "dolt_diff_t2" ]] || false
+    [[ "$output" =~ "dolt_commit_diff_t2" ]] || false
+}
+
+@test "ls: --verbose doesn't show row count for system tables" {
```
macneale4, line 86: Not sure if this is a test which matters. The fact that we don't show the row count is a result of the way we implement the behavior, so it's true that we don't show the count but it's not a spec we expect customers to depend on.

integration-tests/bats/helper/local-remote.bash
```diff
@@ -134,6 +134,7 @@ SKIP_SERVER_TESTS=$(cat <<-EOM
 ~sql-checkout.bats~
 ~cli-hosted.bats~
 ~profile.bats~
+~ls.bats~
```
macneale4, line 137: Why skip your shiny new tests? Do they not work with the server running?

go/cmd/dolt/commands/ls.go
```diff
@@ -83,198 +80,142 @@ func (cmd LsCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
...
-
+	query, err := constructInterpolatedDoltLsQuery(apr)
 	if err != nil {
-		return "", nil, errhand.BuildDError(`Unable to resolve "%s"`, csStr).AddCause(err).Build()
+		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 	}
 
-	r, err := cm.GetRootValue(ctx)
-
+	_, _, err = queryist.Query(sqlCtx, "set @@dolt_show_system_tables = 1")
```
macneale4, line 101: Question - does it work if we just put this set call as a prefix on the main query?

"set @@dolt_show_system_tables = 1; show tables ... "

go/cmd/dolt/commands/ls.go
```diff
@@ -83,198 +80,142 @@ func (cmd LsCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
...
+		return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
 	}
-
-	return h.String(), r, nil
-}
-
-func printUserTables(ctx context.Context, root *doltdb.RootValue, label string, verbose bool) errhand.VerboseError {
-	tblNames, err := doltdb.GetNonSystemTableNames(ctx, root)
-
+	_, _, err = queryist.Query(sqlCtx, "set @@dolt_show_system_tables = 0")
```
macneale4, line 109: or 

"set @@dolt_show_system_tables = 1; show tables ... ; set @@dolt_show_system_tables = 0"

go/cmd/dolt/commands/ls.go
```diff
@@ -83,198 +80,142 @@ func (cmd LsCmd) Exec(ctx context.Context, commandStr string, args []string, dEn
...
-func printSystemTables(ctx context.Context, root *doltdb.RootValue, ddb *doltdb.DoltDB, verbose bool) errhand.VerboseError {
-	perSysTbls, err := doltdb.GetPersistedSystemTables(ctx, root)
-	if err != nil {
-		return errhand.BuildDError("error retrieving persisted table names").AddCause(err).Build()
+func printUserTables(tableNames []string, apr *argparser.ArgParseResults, queryist cli.Queryist, sqlCtx *sql.Context) error {
+	var label string
+	if apr.NArg() == 0 {
+		label = "working set"
+	} else {
+		row, err := GetRowsForSql(queryist, sqlCtx, "select hashof('"+apr.Arg(0)+"')")
```
macneale4, line 170: I tend to avoid string addition. I think this is more readable:

		query := fmt.Sprintf("select hashof('%s')", apr.Arg(0))
		row, err := GetRowsForSql(queryist, sqlCtx, query

go/libraries/doltcore/sqle/database.go
```diff
@@ -673,7 +673,15 @@ func (db Database) GetTableNames(ctx *sql.Context) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	return filterDoltInternalTables(tblNames), nil
+	showSystemTables, err := ctx.GetSessionVariable(ctx, dsess.ShowSystemTables)
```
macneale4, line 676: You can simplify this by creating a function which returns a boolean for if it's enabled. That function can probably swallow all errors and just return false. Up to you.

macneale4 (changes requested):
Very close. The only objection I have at the moment is that we don't test the SQL directly with the new flag. I think a couple enginetests are in order because we are testing it exclusively as a part of `dolt ls` - there are going to be people who want this in the sql conext so we should test that way

integration-tests/bats/helper/local-remote.bash
stephkyou: they use `checkout` :(

go/cmd/dolt/commands/ls.go
stephkyou: No they don't work all in one string. I think the way `dolt sql -q` is set up doesn't allow multiple statements in one call.

macneale4: huh. Sounds like a feature we should have. ok. never mind for now!

macneale4 (approved):
```
    __|__ |___| |\
    |o__| |___| | \
    |___| |___| |o \
   _|___| |___| |__o\
  /...\_____|___|____\_/
  \   o * o * * o o  /
~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #6840, opened 2023-10-18
migrate dolt merge-base to use sql queries

stephkyou:
This change updates dolt merge-base to use the appropriate sql engine to generate results.

### DIFF
--- a/integration-tests/bats/sql-local-remote.bats
+++ b/integration-tests/bats/sql-local-remote.bats
@@ -1215,3 +1215,24 @@ SQL
     [ $status -eq 0 ]
     [[ "$output" =~ "cm3" ]] || false
 }
+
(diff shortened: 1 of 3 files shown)


integration-tests/bats/sql-local-remote.bats
```diff
@@ -1215,3 +1215,24 @@ SQL
     [ $status -eq 0 ]
     [[ "$output" =~ "cm3" ]] || false
 }
+
```
macneale4, line 1218: There is a bats/merge-base.bats file. Seems like you should remove that from bats/helper/local-remote.bash and see if it all works with these changes.

macneale4 (approved):
This was an easy one! Enable the remote server execution on merge-base.bats, and if the tests pass, ship it.

PR #6857, opened 2023-10-20
Improve import table error message

stephkyou:
This changes improves the error message for `dolt table import` to be more readable and useful. Notably, this change adds the column names, table name, and file line number to the error output.

Resolves: https://github.com/dolthub/dolt/issues/5718


fulghum (approved):
Looks great! 🚢

PR #6862, opened 2023-10-23
Ensure `dolt_diff_<tablename>` reports most-specific commit that changed a row

fulghum:
Changes parent order processing in commit iterator so that we prioritize traversing the parent of a merge commit that supplied the commits merged in that commit, before we traverse the other merge parent. 

Fixes: https://github.com/dolthub/dolt/issues/6861

### DIFF
--- a/go/libraries/doltcore/doltdb/commit_itr.go
+++ b/go/libraries/doltcore/doltdb/commit_itr.go
@@ -16,6 +16,7 @@ package doltdb
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	"github.com/dolthub/go-mysql-server/sql"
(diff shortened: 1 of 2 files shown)


zachmu (approved):
This seems ok, but I worry that this "fixes too much". Changing the order of our commit walk is going to have other farther reaching consequences and I'm not sure we have test coverage sufficient to say this yields the behavior we want in all cases. Might need different commit walk strategies for different use cases? @reltuk should take a pass and give his opinion here

fulghum:
I did a little more testing, audited the other uses of this commit iterator, and also followed up with Aaron to discuss any concerns. This is a subtle change to how we walk the commit graph, but it seems like we should use this same logic consistently for the other uses of this commit iterator (e.g. the other system tables, such as dolt_diff, dolt_history, etc).

PR #6866, opened 2023-10-24
Store source database info in tables and columns.

nicktobey:
This allows the engine to differentiate between tables with the same name from different databases, in line with MySQL. This also moves us toward a less error-prone model for resolving table references in queries, where we generate IDs for tables and columns during plan building and use those instead of doing lots of string comparisons.

The main reason for supporting this is that it allows writing queries like:

``SELECT * from `db/main`.table join `db/dev`.table using (pk)``

without having to put each branch in its own subquery.

This is the dolt side of https://github.com/dolthub/go-mysql-server/pull/2090


integration-tests/bats/sql.bats
```diff
@@ -2904,3 +2904,28 @@ SQL
 
     [[ ! -f .dolt/sql-server.lock ]] || false
 }
+
+@test "sql: join same table at two commits" {
```
max-hoffman, line 2908: might be easier in `DoltBranchScripts` or `DoltScripts`

max-hoffman (approved):
lgtm

PR #6880, opened 2023-10-25
[no-release-notes] include creating myview to creating mytable data for enginetests

jennifersp:
Depends on https://github.com/dolthub/go-mysql-server/pull/2101


fulghum (approved):
Looks great! 🚢

PR #6882, opened 2023-10-25
Feature: Dolt reflog support

fulghum:
A new `dolt_reflog()` table function allows you to query Dolt's chunk journal to view the history of named refs, such as branches and tags.

**Usage Example:**
```sql
> select * from dolt_reflog('branch1');
+--------------------+---------------------+----------------------------------+-------------------------------+
| ref                | ref_timestamp       | commit_hash                      | commit_message                |
+--------------------+---------------------+----------------------------------+-------------------------------+
| refs/heads/branch1 | 2023-10-25 20:54:37 | v531ptpmv2tquig8v591tsjghtj84ksg | inserting row 42              |
| refs/heads/branch1 | 2023-10-25 20:53:12 | rvt34lqrbtdr3dhnjchruu73lik4e398 | inserting row 100000          |
| refs/heads/branch1 | 2023-10-25 20:53:06 | v531ptpmv2tquig8v591tsjghtj84ksg | inserting row 42              |
| refs/heads/branch1 | 2023-10-25 20:52:43 | ihuj1l7fmqq37sjhtlrgpup5n76gfhju | inserting row 1 into table xy |
+--------------------+---------------------+----------------------------------+-------------------------------+
4 rows in set (0.00 sec)
```

**Documentation:** https://github.com/dolthub/docs/pull/1815

**Compatibility:** 
This change needs to be released as a minor version bump because it changes the chunk journal record format for root hash records to include a timestamp field. Older versions of Dolt's chunk journal reader aren't forward compatible to read this new format (but the chunk journal reader is backwards compatible to still read older journals). If a customer needs to downgrade after using a version with this support, they can stop their sql-server, run `dolt gc` to move the chunk journal content into oldgen, delete the chunk journal file if necessary (it should be removed automatically by `dolt gc`), and then restart their sql-server with an older Dolt version. 

**Future Extensions:**
- **root author** – I started looking at recording the SQL user who created the new root value, but was hesitant to put too much data into the chunk journal records. This does seem like useful information to track though, so I'm still thinking about whether this is worth adding now or later. 
- **reflog rows for deletion** – Currently, `dolt_reflog()` results show when a ref is changed to a new value, but there is not a row in the result set marking when the ref is deleted. This would be shown as a NULL commit_hash and commit_message. Combining this with the point above would allow the reflog to show who deleted a ref. 
- **show all refs** – The current implementation requires a ref name and filters results on that ref. We should allow viewing the full reflog without filtering on a ref.


go/store/nbs/journal.go
```diff
@@ -171,8 +184,33 @@ func trueUpBackingManifest(ctx context.Context, root hash.Hash, backing *journal
 	return mc, nil
 }
 
+// IterateRoots iterates over the in-memory roots tracked by the ChunkJournal and passes the root and associated
+// timestamp to a callback function, |f|. If |f| returns an error, iteration is stopped and the error is returned.
+func (j *ChunkJournal) IterateRoots(f func(root string, timestamp time.Time) error) error {
+	j.mu.Lock()
+	roots := j.roots
+	rootTimestamps := j.rootTimestamps
```
zachmu, line 192: Might want to make a note about what this locking is actually protecting here (you're making a consistent copy of both slice vars that won't change as other threads call `append` on them)

go/store/nbs/journal.go
```diff
@@ -171,8 +184,33 @@ func trueUpBackingManifest(ctx context.Context, root hash.Hash, backing *journal
...
+	length := len(roots)
+	j.mu.Unlock()
+
+	if len(roots) != len(rootTimestamps) {
```
zachmu, line 196: I think this needs to go inside the lock, it's a race otherwise. The locals vars are references to the slices, other threads could be appending them concurrently

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return fmt.Errorf("unable to look up references for root hash %s: %s",
+				hashof.String(), err.Error())
+		}
+
+		return datasets.IterAll(ctx, func(id string, addr hash.Hash) error {
+			// If the caller has supplied a branch or tag name, without the fully qualified ref path,
+			// take the first match and use that as the canonical
+			if strings.HasSuffix(strings.ToLower(id), "/"+strings.ToLower(refName)) {
+				refName = id
```
zachmu, line 97: What happens if the name matches both a tag and a branch e.g.?

zachmu: Just matches the first one alphabetically?

zachmu (approved):
Looks pretty good, just a couple comments

go/store/nbs/journal.go
fulghum: oh, good catch! yup, I'm trying to capture the slice references and the length in the mutex, so that length check has got to be protected as well. Also another good reason why some documentation explaining the intent of the locking would be helpful, like your comment above suggested. thank u!

go/libraries/doltcore/sqle/reflog_table_function.go
fulghum: Yup, if the reflog contains entries for a tag and a branch with the same name, the first reflog entry that matches will be used for the fully qualified ref path. We iterate over the reflog from newest to oldest records, so the most recently used ref will match.

go/store/nbs/journal.go
fulghum: Ended up pulling out a helper method for this (mostly to ensure the mutex is always unlocked, through `defer`) in 5484c3f34aedcb117976e62614924c624af8cab3. That was also a good place to explain the intent of the locking.

PR #6883, opened 2023-10-25
Remotes: AWS: Fix a bug where uploading tables files to S3 could have unbounded memory usage.

reltuk:


fulghum (approved):
Looks great! 🚢

PR #6893, opened 2023-10-26
Bug fix: for `dolt_history_` system tables

fulghum:
When filtering on an indexed column from the primary index (i.e. a pk column) if a column in the underlying table has been modified and changed its tag in previous versions, we weren't able to create the lookup builder and caused a segfault. This PR changes to use the primary index as a covering index, since it's the closest we have. This is also consistent with [how we had implemented Noms range lookups](https://github.com/dolthub/dolt/blob/1eda29268649a30ae68606f59760959e0a62ed8e/go/libraries/doltcore/sqle/index/index_lookup.go#L86). 

Related to: https://github.com/dolthub/dolt/issues/6891

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -1962,6 +1962,13 @@ var HistorySystemTableScriptTests = []queries.ScriptTest{
 				Query:    "select pk, c2 from dolt_history_t where commit_hash=@Commit2 order by pk;",
 				Expected: []sql.Row{{1, 3}, {4, 6}},
 			},
+			{
+				// When filtering on a column from the original table, we use the primary index here, but because
+				// column tags have changed in previous versions of the table, the index tags don't match up completely.
+				// https://github.com/dolthub/dolt/issues/6891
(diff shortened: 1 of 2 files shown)


max-hoffman (approved):
LGMT. I get why it doesn't make sense to use non-covering, but it wasn't immediately obvious to me why the non-covering panics. If I had to make an addition to this, I'd have `ordinalMappingFromIndex`, `newProllyIndexIter`/ `newNonCoveringLookupBuilder` error if input a primary key, because those constructors are only compatible with secondary indexes. I.e. it's not just sensible to use covering for PK, the other constructors should maybe error (instead of panicking) if given a PK.

zachmu (approved):
LGTM

fulghum:
> If I had to make an addition to this, I'd have `ordinalMappingFromIndex`, `newProllyIndexIter`/ `newNonCoveringLookupBuilder` error if input a primary key, because those constructors are only compatible with secondary indexes. I.e. it's not just sensible to use covering for PK, the other constructors should maybe error (instead of panicking) if given a PK.

Thanks for confirming the fix here! Good suggestion to have `newNonCoveringLookupBuilder` return an error if a primary index is passed in. I added that in 3f2e09e107a50aadfa28a37ccc23ba6acb4baefa as well as a little bit of godocs for the function.

PR #6909, opened 2023-10-27
Remove schema overlap check for tables with same name

stephkyou:
This change takes out the schema check for tables with the same name that get modified. Previously, only tables that get dropped and re-added with the same name and same schema would get labelled as modified. Now, tables that get dropped and re-added with the same name regardless of schema will get labelled as modified.

Fixes: https://github.com/dolthub/dolt/issues/5738

### DIFF
--- a/go/libraries/doltcore/diff/table_deltas.go
+++ b/go/libraries/doltcore/diff/table_deltas.go
@@ -270,12 +270,10 @@ func matchTableDeltas(fromDeltas, toDeltas []TableDelta) (deltas []TableDelta) {
 	for _, name := range matchedNames {
 		t := to[name]
 		f := from[name]
-		if schemasOverlap(t.ToSch, f.FromSch) {
-			matched := match(t, f)
-			deltas = append(deltas, matched)
-			delete(from, f.FromName)
-			delete(to, t.ToName)
-		}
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/diff/table_deltas.go
```diff
@@ -270,12 +270,10 @@ func matchTableDeltas(fromDeltas, toDeltas []TableDelta) (deltas []TableDelta) {
 	for _, name := range matchedNames {
```
fulghum, line 270: (minor) Might be worth a comment for this block and possibly the similar block starting on line 281 to help document the intent of this code. We were using `schemasOverlap` in both blocks, but we're taking it out of just this one. It's probably a little hard to reverse engineer the meaning behind that, so adding a little context could be helpful in the future the next time someone goes into this code. It seems like when we match a table by name, we want to skip the column tag match, but when we don't match by name, we want to try and be clever and see if they are the same tags (so we can try to catch a table rename presumably). Would be nice to leave some of that context about why the code works this way to help document the expected behavior.

fulghum (approved):
Fix looks good and looks like tests are happy. 😄 Just one suggestion to leave some comments explaining why the code works this way to help out future readers.

PR #6912, opened 2023-10-30
add warning message to dolt sql-client

stephkyou:
Adds a warning message to `dolt sql-client` that this command will be deprecated. Command will be fully deprecated in 2 weeks.

Related: https://github.com/dolthub/dolt/issues/6886


go/cmd/dolt/commands/sqlserver/sqlclient.go
```diff
@@ -109,6 +109,8 @@ func (cmd SqlClientCmd) Exec(ctx context.Context, commandStr string, args []stri
 	var serverController *ServerController
 	var err error
 
+	cli.Println(color.YellowString("WARNING: this command is being deprecated and is not recommended for general use."))
```
macneale4, line 112: You should tell them what to use instead.

integration-tests/bats/branch-control.bats
```diff
@@ -47,8 +47,8 @@ setup_test_user() {
 
     run dolt sql-client -u dolt -P $PORT --use-db "dolt_repo_$$" --result-format csv -q "select * from dolt_branch_control"
```
macneale4, line 48: Update the tests to use the supported command.

macneale4 (changes requested):
Drop usage of the command. I'd say the only place that should run `dolt sql-client` should be in sql-client.bats

README.md
```diff
@@ -234,7 +234,7 @@ Your terminal will just hang there. This means the server is running. Any errors
 In the new terminal, we will now connect to the running database server using a client. Dolt also ships with a MySQL compatible client. 
 
 ```bash
-% dolt sql-client -u root
+% dolt -u root -p "" sql
```
macneale4, line 237: just say `dolt sql`

In the line before, you may want to clarify that they should run this in the same directory where they started `dolt sql-server`

macneale4 (changes requested):
oh man, I wish this was a rubber stamp, but I want you to simplify simplify simplify. The --host option is almost never required in bats tests. It's behavior is covered by specific tests. So you should only use it in key places. Otherwise, you should just run the command in the server directory without specifying any host

macneale4 (approved):
LGTM! Just get those two tests to pass!

```
       _~
    _~ )_)_~
    )_))_))_)
    _!__!__!_
    \______t/
  ~~~~~~~~~~~~~
```

PR #6917, opened 2023-10-31
Dolt admin procedures

macneale4:
Currently we encourage our users to grant execute permissions at the DB level, which gives users the ability to run somewhat privileged operations such as `dolt_gc` and `dolt_push`. This change marks sensitive procedures such that explicit grants are required for sensitive features.

Related: https://github.com/dolthub/go-mysql-server/pull/2110


integration-tests/bats/sql-procedure-privs.bats
```diff
@@ -0,0 +1,157 @@
...
+    dolt sql -q "GRANT EXECUTE ON PROCEDURE mydb.dolt_gc TO mike@localhost"
+    dolt -u mike -p pwd sql -q "call dolt_gc()"
+}
+
+mike_blocked_check() {
+    run dolt -u mike -p pwd sql -q "call $1"
+    [ $status -eq 1 ]
+    [[ $output =~ "command denied to user 'mike'@'localhost" ]] || false
+}
+@test "sql-procedure-privs: admin procedures all block" {
```
max-hoffman, line 65:
```suggestion
}

@test "sql-procedure-privs: admin procedures all block" {
```

max-hoffman (approved):
lgtm, only thing i would have done different is maybe a handful of bats tests to verify storage, and have logical enginetests for thoroughness. The logical enginetests on the GMS side seem good enough though, hard to ship a GMS change that breaks these .

PR #6919, opened 2023-10-31
Partial support for virtual columns

zachmu:
Merges don't work yet, coming in a separate PR.

But storage format details should be final.


go/libraries/doltcore/schema/col_coll.go
```diff
@@ -220,16 +241,38 @@ func (cc *ColCollection) GetByTag(tag uint64) (Column, bool) {
...
+}
+
 // Size returns the number of columns in the collection.
 func (cc *ColCollection) Size() int {
 	return len(cc.cols)
 }
 
+// StoredSize returns the number of non-virtual columns in the collection
+func (cc *ColCollection) StoredSize() int {
+	num := 0
```
reltuk, line 267: Just `len(cc.storedIndexes)` ?

go/libraries/doltcore/schema/column.go
```diff
@@ -89,15 +90,24 @@ type Column struct {
...
+	col, err := NewColumnWithTypeInfo(name, tag, typeInfo, partOfPK, "", "", false, false, "", constraints...)
 	if err != nil {
 		panic(err)
 	}
 	return col
 }
 
 // NewColumnWithTypeInfo creates a Column instance with the given type info.
-func NewColumnWithTypeInfo(name string, tag uint64, typeInfo typeinfo.TypeInfo, partOfPK bool, defaultVal string, autoIncrement bool, comment string, constraints ...ColConstraint) (Column, error) {
+func NewColumnWithTypeInfo(
```
reltuk, line 101: Maybe the signature of this function stays the same and we get a new function like:

```
func NewVirtualColumnWithTypeInfo(name string, tag uint64, typeInfo typeinfo.TypeInfo, partOfPK bool, defaultVal generatedVal string, autoIncrement bool, comment string, constraints ...ColConstraint)
```

or something like that...

go/libraries/doltcore/schema/encoding/serialization.go
```diff
@@ -205,11 +205,18 @@ func serializeSchemaColumns(b *fb.Builder, sch schema.Schema) fb.UOffsetT {
...
 		to := b.CreateString(typeString)
 		no := b.CreateString(col.Name)
 
+		defVal := ""
+		if col.Default != "" {
+			defVal = col.Default
+		} else {
+			defVal = col.Generated
+		}
+		do := b.CreateString(defVal)
```
reltuk, line 218: Is there a reason we moved this from its prior location? Seems odd to change the serialization of existing schemas unnecessarily.

go/libraries/doltcore/schema/schema_impl.go
```diff
@@ -417,6 +417,10 @@ func (si *schemaImpl) GetKeyDescriptor() val.TupleDesc {
 	useCollations := false // We only use collations if a string exists
 	var collations []sql.CollationID
 	_ = si.GetPKCols().Iter(func(tag uint64, col Column) (stop bool, err error) {
+		if col.Virtual {
```
reltuk, line 420: What does it mean for a Virtual column to be in PK? How is this going to interact with logic which expects clustered indexes or whatever? Is this really valid?

zachmu (commented):
Thanks for the feedback

zachmu:
PTAL @reltuk

go/libraries/doltcore/schema/column.go
zachmu: I'm just using the struct directly where it's needed now

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4273,6 +4273,217 @@ var DoltVerifyConstraintsTestScripts = []queries.ScriptTest{
...
+			{
+				Query:    "select id from t1 where v3 = 8",
+				Expected: []sql.Row{{4}},
+			},
+		},
+	},
+	{
+		Name: "merge a virtual column",
+		SetUpScript: []string{
+			"create table t1 (id bigint primary key, v1 bigint, v2 bigint, v3 bigint as (v1 + v2), index (v3))",
```
max-hoffman, line 4373: was there a different PR for index support? do we have standalone tests for that?

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4273,6 +4273,217 @@ var DoltVerifyConstraintsTestScripts = []queries.ScriptTest{
...
+			},
+			{
+				Query:    "select id from t1 where v3 = 8",
+				Expected: []sql.Row{{4}},
+				Skip:     true,
+			},
+		},
+	},
+	{
+		Name: "merge a virtual column created on another branch",
```
max-hoffman, line 4428: what happens if you have a schema merge conflict between virtual/generated/real column that are otherwise the identical in type/name/value? do you have to pick one?

go/cmd/dolt/commands/diff.go
```diff
@@ -864,6 +864,8 @@ func getTableSchemaAtRef(queryist cli.Queryist, sqlCtx *sql.Context, tableName s
 	return sch, createStmt, nil
 }
 
+// schemaFromCreateTableStmt returns a schema for the CREATE TABLE statement given
+// TODO: this is substantially incorrect, doesn't handle primary key ordering, probably other things too
```
max-hoffman, line 868: This probably bothers me the most in this PR. Is this func silently breaking things?

max-hoffman (approved):
seems good to me, i might be missing some context on the virtual index support/testing

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
zachmu: You're not missing anything, index merges are broken. They're non trivial to fix because we need the engine's help to rebuild them. I just want to get this in before it gets any bigger, going to do another pass to fix indexes / merge and add more tests.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
zachmu: Good question for the next PR

go/cmd/dolt/commands/diff.go
zachmu: Yes, we are for sure printing schema diffs wrong. Will add bats tests on my next pass

PR #6923, opened 2023-11-01
fix empty Field in `prolly.Range`

JCOR11599:
When creating `prolly.Range`s for `pointLookupPartitions`, we never set the Field member variable. As a result, point lookups wouldn't return correctly.

This impacts specifically keyless multi-arity lookups. Indexes on regular tables and single key keyless lookups were not impacted by this bug.

Companion PR: https://github.com/dolthub/go-mysql-server/pull/2118


max-hoffman (approved):
change looks good, is there a way to trigger this bug with just a two table join? That test would be easier to understand than the 3 table one.

PR #6925, opened 2023-11-01
[no-release-notes] integration-tests/go-sql-server-driver: Change our approach to JWTs and X.509 certificates so that we generate what we need when we run the tests.

reltuk:
Fixes a bug where we had expiring certificates and JWTs in the repository and old commits of the repository would stop passing integration tests and some time in the future.


JCOR11599 (approved):
LGTM

PR #6933, opened 2023-11-01
store `GEOMETRY` types as `BLOB`s

JCOR11599:
To get around the 65k limit on our fields, we should store `GEOMETRY` types as `BLOB`s.
Additionally, MySQL stores these as out of band BLOBs.

This will also probably worsen the performance of `GEOMETRY` types as it just takes longer to reference.

Fixes https://github.com/dolthub/dolt/issues/6927


go/store/val/tuple_descriptor.go
```diff
@@ -439,6 +439,11 @@ func (td TupleDesc) GetGeometry(i int, tup Tuple) (v []byte, ok bool) {
 	return
 }
 
+func (td TupleDesc) GetGeometryAddr(i int, tup Tuple) (hash.Hash, bool) {
+	td.expectEncoding(i, GeometryEnc)
```
zachmu, line 443: I'm a little bothered by the fact that there isn't a separate encoding for this

zachmu (commented):
Mostly looks good, like the backwards compatibility

But I think we need a new encoding type if we're going to support both encodings

go/libraries/doltcore/sqle/index/prolly_fields.go
```diff
@@ -91,12 +91,27 @@ func GetField(ctx context.Context, td val.TupleDesc, i int, tup val.Tuple, ns tr
 			err = json.Unmarshal(buf, &doc.Val)
 			v = doc
 		}
+	// TODO: eventually remove this, and only read GeomAddrEnc
 	case val.GeometryEnc:
```
zachmu, line 95: Are any tests going to hit this logic to see if it works?

zachmu (approved):
LGTM, just make sure there are tests that exercise the backward compatibility logic. You may need to manually construct tuples with PutGeometry, because everything on the SQL path is just going to use the new encoding.

go/libraries/doltcore/sqle/index/prolly_fields.go
JCOR11599: added some in prolly_field_tests.go

PR #6935, opened 2023-11-02
Remove bulk_insert CI benchmark

max-hoffman:


timsehn (commented):
Does this still run it but just not report the stats?

timsehn:
We should definitely merge this. It's better than having 0 and making us look better than we are.

max-hoffman:
> Does this still run it but just not report the stats?

Pretty sure it will not run, but not 100% confident.

PR #6936, opened 2023-11-02
More flexibility for `filter-branch` command

zachmu:
Fixes https://github.com/dolthub/dolt/issues/6895

The command can now process multiple queries by default, and will ignore errors executing them with a `--continue` flag. This is a breaking change:

1. Previously, certain errors were detected and ignored by default. Now this behavior is controlled by the --continue flag.
2. Previously, the query to execute was specified as an argument. Now it's a flag, `--query`, or can be read via STDIN.


fulghum (approved):
Looks good! 🚢

PR #6947, opened 2023-11-03
Switch reflog data to be stored in a ring buffer

fulghum:
Changes the in-memory reflog buffer to be a ring buffer, to limit on how much memory is used. 

Although we do have an `async.RingBuffer` type already, it has slightly different use case. This implementation is more tailored to the reflog buffer's needs, including supporting iteration, and less locking and synchronization.

### DIFF
--- a/go/store/nbs/reflog_ring_buffer.go
+++ b/go/store/nbs/reflog_ring_buffer.go
@@ -1,0 +1,156 @@
+// Copyright 2023 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


go/store/nbs/reflog_ring_buffer.go
```diff
@@ -0,0 +1,200 @@
...
+	insertIndex   int
+	itemCount     int
+	epoch         uint
+}
+
+// newReflogRingBuffer creates a new reflogRingBuffer that allows the reflog to query up to |size| records.
+// Internally, the ring buffer allocates extra storage so that |size| records can be read while new root entries
+// are still being recorded.
+func newReflogRingBuffer(size int) *reflogRingBuffer {
+	if size < 0 {
```
zachmu, line 55: <=?

zachmu (approved):
LGTM!

go/store/nbs/reflog_ring_buffer.go
fulghum: I actually do allow 0 to be passed in – if the `DOLT_DISABLE_REFLOG` env var is set, then the reflog ring buffer is still created, but created with a 0 record capacity. It could be changed to not instantiate the reflog ring buffer in that case if we do want to change this.

PR #6955, opened 2023-11-06
Optimize when secondary indexes need to be fully rebuilt after a merge

fulghum:
At the end of a merge, Dolt was rebuilding some secondary indexes that did not need to be rebuilt. https://github.com/dolthub/dolt/issues/6951 shows one example where the destination side of a merge has an existing index that does not exist on the source side of the merge. Dolt's merge code correctly updates that secondary index in place as the merge diffs are processed, but we were still doing an unnecessary full rebuild on the secondary index after processing all the diffs for that table. 

Making this optimization uncovered another bug where column default values weren't being applied when merging a new row into a secondary index. This PR also fixes that bug, by ensuring that we remap a tuple and apply the column default value, before updating the secondary index with the new tuple.

### DIFF
--- a/go/libraries/doltcore/merge/merge_prolly_indexes.go
+++ b/go/libraries/doltcore/merge/merge_prolly_indexes.go
@@ -31,8 +31,9 @@ import (
 // mergeProllySecondaryIndexes merges the secondary indexes of the given |tbl|,
 // |mergeTbl|, and |ancTbl|. It stores the merged indexes into |tableToUpdate|
 // and returns its updated value. If |rebuildIndexes| is true, then all indexes
-// will be rebuilt from the table's data, instead of merging in incremental
-// changes from the other side of the merge. This is safer, but less efficient.
+// will be rebuilt from the table's data, instead of relying on incremental
+// changes from the other side of the merge to have been merged in before this
+// function was called. This is safer, but less efficient.
 func mergeProllySecondaryIndexes(
 	ctx context.Context,
 	tm *TableMerger,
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/merge/merge_prolly_indexes.go
```diff
@@ -31,21 +31,18 @@ import (
 // mergeProllySecondaryIndexes merges the secondary indexes of the given |tbl|,
 // |mergeTbl|, and |ancTbl|. It stores the merged indexes into |tableToUpdate|
 // and returns its updated value. If |rebuildIndexes| is true, then all indexes
```
timsehn, line 33:
```suggestion
// and returns its updated value. If |forceIndexRebuild| is true, then all indexes
```

timsehn (commented):
Was curious and found something you forgot to change in a comment

go/libraries/doltcore/merge/merge_prolly_indexes.go
fulghum: Good catch! Thank you!

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1148,8 +1150,12 @@ func (m *secondaryMerger) merge(ctx context.Context, diff tree.ThreeWayDiff, sou
 					return fmt.Errorf("cannot merge keyless tables with reordered columns")
 				}
 			} else {
-				valueMappedToMergeSchema := remapTuple(diff.Right, sourceSch.GetValueDescriptor(), m.valueMerger.rightMapping)
-				newTupleValue = val.NewTuple(m.valueMerger.syncPool, valueMappedToMergeSchema...)
+				tempTupleValue, err := remapTupleWithColumnDefaults(ctx, diff.Key, diff.Right, sourceSch.GetValueDescriptor(),
```
max-hoffman, line 1153: do we have a test for this one?

max-hoffman (approved):
nice optimization! tests for the default would be nice

go/libraries/doltcore/merge/merge_prolly_rows.go
fulghum: yup – after I made the change to stop rebuilding secondary indexes as aggressively after a merge, two of our schema merge tests started failing and showed me we had this bug here. Specifically... because we run the schema merge tests left-to-right, and then flip them to run them right-to-left, and because we have the secondary index validation check at the end of our tests, that's what was able to catch this.

PR #6956, opened 2023-11-06
fix `sql_mode` column access of `dolt_schemas` table

jennifersp:
The databases that were created before `sql_mode` column was added to `dolt_schemas` table does not have this column, so access this column needs to be check beforehand.


fulghum (approved):
Looks good! Thanks for fixing this one!

PR #6973, opened 2023-11-08
make `GeomAddrEnc` known as an address encoding

JCOR11599:
The addition of the new `GeomAddrEnc` needs to be explicitly known as an address type, so that `dolt gc` doesn't delete the chunks created.

fixes https://github.com/dolthub/dolt/issues/6969


reltuk (approved):
LGTM

zachmu (commented):
Where's the test for pushing geometry types to catch the actual bug?

zachmu (approved):
LGTM

PR #6974, opened 2023-11-08
Support merging schemas with virtual / generated columns

zachmu:
Substantially addresses https://github.com/dolthub/dolt/issues/6945, although there are a couple remaining edge cases that are hard to crack, in this PR as skipped tests.


go/libraries/doltcore/sqle/enginetest/dolt_queries_schema_merge.go
```diff
@@ -2003,3 +2094,236 @@ var SchemaChangeTestsSchemaConflicts = []MergeScriptTest{
...
+		LeftSetUpScript: []string{
+			"alter table t add column col1 int as (pk + 100) stored;",
+			"insert into t (pk) values (5), (6);",
+			"alter table t add index (col1);",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query:    "call dolt_merge('right');",
+				Expected: []sql.Row{{doltCommit, 0, 0}},
+				Skip:     true, // this fails merging right into left
```
fulghum, line 2274: Interesting! 

I ran this locally and it looks like it's actually failing on the left-to-right merge with:
```
index key ( 101, 1 ) not found in index col1

Secondary index contents:
  - k: [1 0 0 0 0 0 2 0] 
  - k: [2 0 0 0 0 0 2 0] 
  - k: [3 0 0 0 0 0 2 0] 
  - k: [4 0 0 0 0 0 2 0] 
  - k: [5 0 0 0 0 0 2 0] 
  - k: [6 0 0 0 0 0 2 0] 
```

Looks like the index created after the merge only has `pk` and doesn't contain `col1` at all, right? 

A couple other clues I noticed while playing around with this one:
- when I changed both sides to have the same virtual column expression (`pk + 100`), it only failed when going from left-to-right, probably because left still specified the virtual column was `stored`. 
- when I changed both sides to be virtual, non-stored columns, with the same expression, it failed with: `panic: runtime error: index out of range [0] with length 0` (coming from `merge_prolly_rows.go` on line 93).

go/libraries/doltcore/sqle/enginetest/dolt_queries_schema_merge.go
```diff
@@ -2003,3 +2094,236 @@ var SchemaChangeTestsSchemaConflicts = []MergeScriptTest{
...
+			},
+			{
+				Query: "show create table t;",
+				Expected: []sql.Row{{"t",
+					"CREATE TABLE `t` (\n" +
+						"  `pk` int NOT NULL,\n" +
+						"  `col1` int NOT NULL,\n" +
+						"  `col3` int GENERATED ALWAYS AS ((pk + 1)),\n" +
+						"  PRIMARY KEY (`pk`)\n" +
+						") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin"}},
```
fulghum, line 2321: Shouldn't the index `idx1` show up in here?

fulghum (commented):
One more file left to go through, but figured I'd share a few comments so far. 

A couple other testing thoughts about generated columns not really related to merging... I didn't go look up the answers yet, but figured I'd just go ahead and share: 
- do we have tests for virtual columns in keyless tables? (I noticed there was some logic that switched on generated keyless tables for key building)
- do we test modifying a column to change it to/from being virtual/non-virtual? (or stored/non-stored?)

I noticed we don't support using the `DEFAULT` keyword for generated columns, but MySQL does, e.g:
```
create table t3 (pk int primary key, c1 int as (pk * 3));
insert into t3 values (1, DEFAULT);
```

Dolt generates an error (`The value specified for generated column "c1" in table "t3" is not allowed.`), but MySQL will allow it.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -1078,12 +1056,111 @@ func (m *primaryMerger) merge(ctx *sql.Context, diff tree.ThreeWayDiff, sourceSc
 	case tree.DiffOpRightDelete:
 		return m.mut.Put(ctx, diff.Key, diff.Right)
 	case tree.DiffOpDivergentModifyResolved:
-		return m.mut.Put(ctx, diff.Key, diff.Merged)
+		// any generated columns need to be re-resolved because their computed values may have changed as a result of
+		// the merge
```
fulghum, line 1060: Do we need this same logic for `secondaryMerge.merge`?

fulghum (approved):
Finished looking over the last file. Looks good! Would be nice to dig into those skipped tests, but seems like they only affect merges with generated columns, so seems fine to fix those as follow ups.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4325,6 +4325,55 @@ var GeneratedColumnMergeTestScripts = []queries.ScriptTest{
...
+			"call dolt_commit('-Am', 'branch1 commit')",
+			"call dolt_checkout('branch2')",
+			"update t1 set v2 = 5 where id = 1",
+			"call dolt_commit('-Am', 'branch2 commit')",
+			"call dolt_checkout('main')",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query:            "call dolt_merge('branch1')",
+				SkipResultsCheck: true,
```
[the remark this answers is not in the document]
zachmu: Done

go/libraries/doltcore/sqle/index/key_builder.go
```diff
@@ -44,10 +55,28 @@ func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.Tup
...
 		if !ok {
-			if keyless {
+			col := sch.GetNonPKCols().TagToCol[tag]
+			if col.Virtual {
+				if len(virtualExpressions) == 0 {
+					virtualExpressions = make([]sql.Expression, len(def.AllTags()))
+				}
+				sqlCtx, ok := ctx.(*sql.Context)
+				if !ok {
+					sqlCtx = sql.NewContext(ctx)
```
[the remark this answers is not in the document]
zachmu: Hmm, I'll make a note of this for a second pass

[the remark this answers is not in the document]
zachmu: Fixed actually

go/libraries/doltcore/sqle/index/key_builder.go
```diff
@@ -57,20 +86,105 @@ func NewSecondaryKeyBuilder(sch schema.Schema, def schema.Index, idxDesc val.Tup
...
+func parseCreateTable(ctx *sql.Context, tableName string, sch schema.Schema) (*plan.CreateTable, error) {
+	createTable, err := sqlfmt.GenerateCreateTableStatement(tableName, sch, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	query := createTable
+
+	mockDatabase := memory.NewDatabase("mydb")
+	mockProvider := memory.NewDBProvider(mockDatabase)
```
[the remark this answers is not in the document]
zachmu: They're just stubs to get this technique to work at all, should work fine.

We explicitly exclude any foreign keys from the CREATE TABLE statement we generate, so should be no issue there either.

go/libraries/doltcore/sqle/enginetest/dolt_queries_schema_merge.go
zachmu: Thanks, making a note of this

go/libraries/doltcore/sqle/enginetest/dolt_queries_schema_merge.go
zachmu: Good catch

go/libraries/doltcore/merge/merge_prolly_rows.go
zachmu: Probably, I'll add a note

PR #6976, opened 2023-11-08
go/store/nbs: file_table_persister: fsync whole table file writes.

reltuk:
Dolt sometimes persists whole table files, instead of appending blocks to the chunk journal. In order to be as crash resistent as the chunk journal, we should fsync the writes before we add the new table file to the database.

Cases which use this code path and could have seen reduced durability across crashes include: dolt_fetch, dolt_pull, dolt_clone, dolt_gc, and dolt_push/dolt_backup to a file remote.


zachmu (approved):
LGTM

PR #6977, opened 2023-11-08
Improve the persistence of the file that stores USERs and GRANTs.

reltuk:
The USERs and GRANTs on a sql-server instance are stored in a separate file from the Merkle DAG table data that makes up the Dolt databases themselves. Previously, the contents of this file were not written in a crash resistant way, and they could be lost or corrupted after a crash or when taking a block-device snapshot.

This change also changes the file's permission bits to be `0600`, instead of `0777`, which was much more permissive than intended.


go/libraries/doltcore/sqle/mysql_file_handler/file_handler.go
```diff
@@ -53,7 +58,44 @@ func (p *Persister) Persist(ctx *sql.Context, data []byte) error {
 		}
 	}
 
-	return os.WriteFile(p.privsFilePath, data, 0777)
+	dir := filepath.Dir(p.privsFilePath)
```
zachmu, line 61: Method for this?

zachmu (approved):

PR #6978, opened 2023-11-08
Change to use a crash resistant, atomic file update protocol for a number of file writes.

reltuk:
Makes updates to branch_control, repo_state.json and some persisted config files atomic and crash resistant.


zachmu (approved):
LGTM

PR #6980, opened 2023-11-09
Correctly identify and report data conflicts when deleting rows and/or columns during a three-way merge.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6747
Also fixes https://github.com/dolthub/dolt/issues/6766

https://github.com/dolthub/dolt/issues/6747 is important to fix because ignoring data conflicts means that a merge could put a user in a state that wouldn't be possible from a linear sequence of transactions. Issues like this (especially ones that don't display any warning to the user) need to be addressed *before* they cause problems.

Most of this PR is refactoring how we do three way merging in a way that makes these fixes possible.

Prior to this PR, we would handle schema merges by converting the left side of the merge to match the result schema, and then processing each diff. However, this causes ambiguity when the right side drops a column: it is no longer possible to distinguish between merges where the left side modified the dropped column and merges where it didn't. It is no longer possible to reliably determine whether or not there was a data conflict.

This PR changes the approach so that, in the event of a schema merge, each row gets a Diff object which is processed during three way merging. This requires every part of the merge process to handle `DiffOpLeftAdd` and `DiffOpLeftModify`, diff types that were previously handled before the merge and thus could be safely ignored.

This shouldn't cause an impact in runtime: running the three way diff in the event of a schema merge is now O(N) on the size of the left table, but prior to this PR, *preparing to run the three way diff in the event of a schema merge* was O(N) on the size of the left table: the total runtime should be unaffected.

I added several more tests to gain confidence in the correctness of these changes. It's possible that, if there's an oversight, diffs where there *isn't* a schema change may also result in O(Left Rows) runtime. If this is the case it should be easy to fix, but it's something to keep an eye out for.


fulghum (approved):
Nice work on this one Nick. I think you found a good approach here and had clean changes. Only a few very minor comments from me.

PR #6992, opened 2023-11-10
go/libraries/doltcore/remotesrv: Ensure we stop the gRPC server and cancel inflight requests when we are multiplexing one port for HTTP and gRPC traffic.

reltuk:


go/libraries/doltcore/dtestutils/sql_server_driver/cmd.go
```diff
@@ -364,7 +364,27 @@ func (s *SqlServer) Restart(newargs *[]string, newenvs *[]string) error {
...
+				os.Stdout.Write([]byte("] "))
+			}
+			multiOut.Write(line)
+			if isPrefix {
+				wantsPrefix = false
+			} else {
+				multiOut.Write([]byte("\n"))
+				wantsPrefix = true
+			}
+		}
```
fulghum, line 387: Might be worth considering popping this out into a named function so you could reuse it here and in the `runSqlServerCommand` function. Looks like those two blocks of code have the same logic.

fulghum (approved):
Nice change for the server shutdown!

PR #6994, opened 2023-11-10
Concurrent remotes map

nustiueudinastea:
This is a first stab at converting the remotes map to a concurrent map. A new generic concurrent map has been added to the utils folder, as I couldn't find a widespread generic implementation of a concurrent map yet. 

I am creating the PR to get some feedback, even though the following still need to be done:
- tests for the concurrent map
- probably a conversion of backups and branches needs to be done as well

This completely fixes the issue I've seen in my project tests when high concurrency is involved.

This PR attempts to fix https://github.com/dolthub/dolt/issues/6965. @zachmu when you have some time, let me know what you think.

### DIFF
--- a/go/libraries/utils/concurrentmap/concurrentmap.go
+++ b/go/libraries/utils/concurrentmap/concurrentmap.go
@@ -66,7 +66,7 @@ func (cm *Map[K, V]) DeepCopy() *Map[K, V] {
 	return &Map[K, V]{m: newMap}
 }
 
-func (cm *Map[K, V]) Range(f func(key K, value V) bool) {
+func (cm *Map[K, V]) Iter(f func(key K, value V) bool) {
 	cm.mu.RLock()
 	defer cm.mu.RUnlock()
 	for k, v := range cm.m {
--- a/go/libraries/utils/concurrentmap/concurrentmap.go
+++ b/go/libraries/utils/concurrentmap/concurrentmap.go
@@ -1,0 +1,60 @@
+package concurrentmap
+
+import "sync"
+
+func New[K comparable, V any]() *Map[K, V] {
+	return &Map[K, V]{m: make(map[K]V)}
+}
+
+type Map[K comparable, V any] struct {
+	mu sync.RWMutex
+	m  map[K]V
+}
+
+func (cm *Map[K, V]) Get(key K) (V, bool) {
+	cm.mu.RLock()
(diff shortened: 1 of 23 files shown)


timsehn:
@zachmu is reviewing this now. Starting CI.

go/libraries/utils/concurrentmap/concurrentmap.go
```diff
@@ -0,0 +1,72 @@
...
+	cm.mu.RLock()
+	defer cm.mu.RUnlock()
+	newMap := make(map[K]V, len(cm.m))
+	for k, v := range cm.m {
+		newMap[k] = v
+	}
+	return &Map[K, V]{m: newMap}
+}
+
+func (cm *Map[K, V]) Range(f func(key K, value V) bool) {
```
zachmu, line 54: We tend to call this `Iter` in other parts of the code base

go/libraries/utils/concurrentmap/concurrentmap.go
```diff
@@ -0,0 +1,72 @@
...
+	cm.mu.RLock()
+	defer cm.mu.RUnlock()
+	for k, v := range cm.m {
+		if !f(k, v) {
+			break
+		}
+	}
+}
+
+func (cm *Map[K, V]) Snapshot() map[K]V {
```
zachmu, line 64: Method doc for this?

zachmu (commented):
Interface looks fine, but looks like you have broken quite a few test, including many bats tests. Ping me when those work and I'll take a closer look.

nustiueudinastea:
hey @zachmu can you run the tests again please. I fixed all the remotes related tests I found, but still getting some bats related failures. Most likely I don't have an identical testing environment. (still need to add the tests for the concurrent map)

timsehn:
Running tests again.

nustiueudinastea:
Hey @zachmu, I added tests for the concurrent map implementation. I also took the liberty to slightly chance the CI tests in order to introduce the race tests for the new concurrent map. So please check .github/workflows/ci-go-race-tests.yaml to see if you agree with this, otherwise I can revert it.

Now, since this done and all tests pass, the question is if RepoState.Backups and RepoState.Branches should also be changed to this concurrent map implementation? Most likely the same issue will happen with those as well, so if you like me to, I can change them as well in this PR.

PR #6995, opened 2023-11-10
allow hyphen in db name to match its dir name

jennifersp:
The PR adds support for allowing  hyphens,`-`, in a database name. 
Before we replaced hyphens with underscores when creating database using directory name. E.g.  for`test-db` directory name, database name will be `test_db`. Now, the database name for this directory will be `test-db`, the same as directory name. This change will be the default behavior, but it can be disabled(switch to previous behavior of replacing hyphens in database name with underscore) by ENV variable `DOLT_DBNAME_REPLACE_HYPHENS` set to non empty value.

### DIFF
--- a/go/libraries/doltcore/dbfactory/dirtodbname.go
+++ b/go/libraries/doltcore/dbfactory/dirtodbname.go
@@ -25,7 +25,8 @@ import (
 // DirToDBName takes the physical directory name, |dirName|, and replaces any unsupported characters to create a
 // valid logical database name. For example, spaces are replaced with underscores.
 func DirToDBName(dirName string) string {
-	var translateHyphensToUnderscores = os.Getenv(dconfig.EnvDisableDatabaseRenaming) != ""
+	// this environment variable is used whether to replace hyphens in the database name with underscores.
+	var translateHyphensToUnderscores = os.Getenv(dconfig.EnvDbNameReplaceHyphens) != ""
 	dbName := strings.TrimSpace(dirName)
 	dbName = strings.Map(func(r rune) rune {
 		if unicode.IsSpace(r) || (translateHyphensToUnderscores && r == '-') {
--- a/go/libraries/doltcore/dbfactory/dirtodbname.go
+++ b/go/libraries/doltcore/dbfactory/dirtodbname.go
@@ -15,16 +15,19 @@
 package dbfactory
 
 import (
+	"github.com/dolthub/dolt/go/libraries/doltcore/dconfig"
+	"os"
 	"strings"
 	"unicode"
 )
 
 // DirToDBName takes the physical directory name, |dirName|, and replaces any unsupported characters to create a
 // valid logical database name. For example, spaces are replaced with underscores.
-func DirToDBName(dirName string, disableHyphen bool) string {
+func DirToDBName(dirName string) string {
+	var translateHyphensToUnderscores = os.Getenv(dconfig.EnvDisableDatabaseRenaming) != ""
 	dbName := strings.TrimSpace(dirName)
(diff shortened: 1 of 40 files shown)


integration-tests/bats/db-revision-specifiers.bats
```diff
@@ -3,6 +3,7 @@ load $BATS_TEST_DIRNAME/helper/common.bash
 
 setup() {
     setup_common
+    dolt config --add database.disablehyphen true
```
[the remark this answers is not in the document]
fulghum: Although... thinking about this some more... an env var has the advantage of always being "global" for a SQL server. By using a config setting, we need to check each individual database's setting (or use `dolt config --global`). There's an edge case with `undrop` where the dropped database manager uses one global setting no matter what the dropped database requested. I'm wondering if using an env var might actually be simpler. It would also mean we didn't have to pipe the boolean value around through a bunch of calls, which would be a minor advantage. We can discuss more in our call, too.

fulghum (commented):
Looking very good! I just wanted to brainstorm on the decision between a per-database config setting versus a global env var setting.

go/libraries/doltcore/dbfactory/dirtodbname.go
```diff
@@ -15,16 +15,20 @@
...
 	"unicode"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/dconfig"
 )
 
 // DirToDBName takes the physical directory name, |dirName|, and replaces any unsupported characters to create a
-// valid logical database name. For example, hyphens and spaces are replaced with underscores.
+// valid logical database name. For example, spaces are replaced with underscores.
 func DirToDBName(dirName string) string {
+	var translateHyphensToUnderscores = os.Getenv(dconfig.EnvDisableDatabaseRenaming) != ""
```
fulghum, line 28: This logic is a little hard to parse... If `DOLT_DISABLE_DATABASE_RENAMING` is set to any value, then we translate hyphens to underscores? That seems backwards, right? Doesn't this mean if the user sets `DOLT_DISABLE_DATABASE_RENAMING` that we **will** rename the database? 

I think we need a clearer env var name here. Sorry if this was a name I brainstormed with you earlier. How about `DOLT_DBNAME_REPLACE_HYPHENS` or something similar? That seems more accurate in a couple of ways (e.g. we're still doing some db renaming after all) and it also avoids the negative of `DISABLE` which is hard to grok when the new change is to stop doing something (i.e. feels like a double negative to disable stopping doing something).

integration-tests/bats/db-revision-specifiers.bats
```diff
@@ -3,6 +3,7 @@ load $BATS_TEST_DIRNAME/helper/common.bash
 
 setup() {
     setup_common
+    export DOLT_DISABLE_DATABASE_RENAMING="false"
```
fulghum, line 6: The value of `false` is a little misleading, because we don't ever actually check the value, just whether the env var is set or not. Instead, just use `true` or `1` for settings where we're only checking the existence of the setting, and not checking the actual value. 

It's also a bit odd to see `disable=false` – avoid double negatives like that whenever you see them. They easily lead to confusion in code and in prose.

fulghum (commented):
I don't think the env var name (`DOLT_DISABLE_DATABASE_RENAMING`) is right yet. It seems like you set that env var to actually **enable** DBs to be renamed to remove hyphens, which seems backwards from its name. 🤔 

Looks like there are one or two other comments from yesterday's review that still need to be addressed, too.

fulghum (approved):
Looks great! Thanks for iterating on this one.

PR #6996, opened 2023-11-10
go: sqle: cluster: commithook: When shutting down the server, the cluster replication commithook could deadlock after the replication thread missed a wakeup signal.

reltuk:
Fix it so that the wakeup thread is guaranteed to see the canceled context or the wakeup signal.

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -214,7 +214,10 @@ func (nbs *NomsBlockStore) UpdateManifest(ctx context.Context, updates map[hash.
 		return
 	}
 
-	nbs.checkAllManifestUpdatesExist(ctx, updates)
+	err = nbs.checkAllManifestUpdatesExist(ctx, updates)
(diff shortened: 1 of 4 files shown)


go/store/nbs/store.go
```diff
@@ -214,7 +214,10 @@ func (nbs *NomsBlockStore) UpdateManifest(ctx context.Context, updates map[hash.
 		return
 	}
 
-	nbs.checkAllManifestUpdatesExist(ctx, updates)
+	err = nbs.checkAllManifestUpdatesExist(ctx, updates)
+	if err != nil {
+		return
+	}
```
fulghum, line 220: oooh! nice catch! I've been curious about swallowed errors like this that a linter could find in our code base. I'm gunna play with [errcheck](https://github.com/kisielk/errcheck) and see what it can find.

fulghum (approved):
Nice work finding these! 🙌

PR #7004, opened 2023-11-15
cmd/dolt/commands/sqlserver: Restructure the start up sequence for sql-server.

reltuk:
We explicitly model Services, which can have an Init step, a Run step and a Stop step. Every registered service get initialized in the order they were registered in, then they all run concurrently until Stop is called, when they all get Stopped in reverse order. It's possible for clients to wait for init to be completed and be delivered any errors encountered on startup. They can also wait for stop, to be delivered any errors encountered on shutdown.


go/libraries/utils/svcs/controller.go
```diff
@@ -0,0 +1,189 @@
...
+	Stop func() error
+}
+
+// A Controller is responsible for initializing a number of registered
+// services, running them all, and stopping them all when requested. Services
+// are registered with |Register(*Service)|. When |Start| is called, the
+// services are all initialized,  in the order of their registration, and if
+// every service initializes successfully, they are |Run| concurrently. When
+// |Stop| is called, services are stopped in reverse-registration order. |Stop|
+// does not block for the goroutines spawned by |Start| to complete, although
```
reltuk, line 41: Probably document the fact that |Stop| *does* block until Start is done calling Stop() on all services.

go/libraries/utils/svcs/controller.go
```diff
@@ -0,0 +1,189 @@
...
+			continue
+		}
+		err := s.Init(ctx)
+		if err != nil {
+			for j := i - 1; j >= 0; j-- {
+				if svcs[j].Stop != nil {
+					svcs[j].Stop()
+				}
+			}
+			c.mu.Lock()
```
reltuk, line 155: This seems to have a weird contract with Stop(). In particular, Stop() claims that its not safe to call it until after WaitForStart() returns successfully, but some calls to `Stop()` before `WaitForStart()` returns will succeed, even if `WaitForStart()` is going to return an error. However, if an Init function fails, it will write `stopped = true` here and then the call to `Stop` will fail.

Should `Stop` be changed to return an error in those cases as well? I.e., the Controller has to be fully inited in order for calling Stop() to not return an error?

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -63,335 +64,470 @@ func Serve(
...
+
+				metSrv = &http.Server{
+					Addr:    fmt.Sprintf("%s:%d", serverConfig.MetricsHost(), serverConfig.MetricsPort()),
+					Handler: mux,
+				}
 
-		go func() {
-			_ = metSrv.ListenAndServe()
-		}()
+				_ = metSrv.ListenAndServe()
```
reltuk, line 366: TODO: Dropped error here if Listen fails. Might be better to separate Listen from Serve and Listen in Init.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -63,335 +64,470 @@ func Serve(
...
+			}
+			return nil
+		},
+		Run: func(ctx context.Context) {
+			if remoteSrv == nil {
+				return
+			}
+			remoteSrv.Serve(remoteSrvListeners)
+		},
+		Stop: func() error {
```
reltuk, line 422: TODO: There's a state where Init() and Stop() get called but Run() never gets called. In that case, the grpc.Server instance never took ownership of the listeners and we should close them here.

AFAICT, to correctly handle this would require some thread-safe coordination between `Run` and `Stop`, so that `Stop` took responsibility for the listeners only if Serve had never been called, and otherwise the listeners were handled as they are now, by `GracefulStop`.

This state also exists in `RunClusterRemoteSrv` just below.

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -63,335 +64,470 @@ func Serve(
...
+			clusterController.GracefulStop()
+			return nil
+		},
 	}
+	controller.Register(RunClusterController)
 
-	if err = mrEnv.Lock(serverLock); err != nil {
-		startError = err
-		return
+	RunSQLServer := &svcs.Service{
```
reltuk, line 513: This is broken up from InitSQLServer, but it's possible they should be together. We need access to the mySQLServer instance for the cluster controller, but maybe there are other ways to handle that.

AFAICT, InitSQLServer opens the listeners, and so it should be paired up with the Close at least. But we want the Close() of these listeners to be the first thing we do on shutdown, before we shutdown the cluster controller for example.

We could add some more code to wrangle things further given the current setup – for example, we could set a flag or move the pointer to a new variable so that InitSQLServer is responsible for closing these things until we get far enough in the boot process, and after that point RunSQLServer becomes responsible.

zachmu (commented):
So overall I think there's a good abstraction here, but I'm not wild about the current implementation details. Basically boils down to 3 big points:

1) There's leakage from one service's Init to another, e.g. clusterController gets Init'ed by one service then used by the Init of another one. These things should be reasonably self-contained
2) There are many Init-only "Services", which is another sign of a leaky abstraction. A thing that just runs an Init block and doesn't even need the cleanup of Close() isn't a service and shouldn't be modeled as one
3) The actual services should be better encapsulated with their own types and probably their own files, not just anonymous functions. `Service` should probably be an interface, not a struct with funcs.

Basically I think you need to do a better job modeling and separating the concerns of initialization from the concerns of running a service and cleaning up when it shuts down.

go/libraries/utils/svcs/controller.go
reltuk: Changed this so that Stop never returns an error and can be called multiple times and concurrently.

reltuk:
I moved Service to an interface but I ended up leaving all the existing uses as instances of a struct implementation. When I teased things apart a bit further, if felt like I was getting less readable code, à la http://number-none.com/blow/john_carmack_on_inlined_code.html ...which, to be fair, we do have some gross mutations in here.

PR #7015, opened 2023-11-17
Update docs for merge's `--abort` param

fulghum:
The merge CLI command and SQL stored procedure both validate that the working set is clean before it will start a merge, so the warning about uncommitted changes isn't relevant.


jennifersp (approved):
Looks good!

PR #7016, opened 2023-11-17
Allow `dolt config` to run in folders without write permissions

stephkyou:
This change allows`dolt config` to run in folders without write permissions.

Fixes: https://github.com/dolthub/dolt/issues/2313

### DIFF
--- a/integration-tests/bats/config.bats
+++ b/integration-tests/bats/config.bats
@@ -290,7 +290,6 @@ teardown() {
 
 @test "config: config doesn't need write permission in current dir" {
     chmod 111 .
-    run dolt config --list
-    [ "$status" -eq 0 ]
+    dolt config --list
     chmod 755 .
 }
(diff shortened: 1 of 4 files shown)


integration-tests/bats/config.bats
```diff
@@ -287,3 +287,10 @@ teardown() {
     [ "$status" -eq 0 ]
     [[ "$output" =~ "* vegan-btw" ]] || false
 }
+
+@test "config: config doesn't need write permission in current dir" {
+    chmod 111 .
+    run dolt config --list
+    [ "$status" -eq 0 ]
```
fulghum, line 294: (minor) when you only need to check that a command executed without returning an error return code, you can skip using `run` and checking `$status` and just call the command directly. If it fails, it'll fail the BATS test. Handy to know for cases like this where you just want to make sure the command doesn't blow up.

fulghum (approved):
Looks good! 🚢

PR #7020, opened 2023-11-17
use database name that is the same as directory name on disk

jennifersp:
As of Dolt version `1.27.0`, we allow hyphens in database name and replace space characters with underscores and reduce multiple underscore characters side by side into single underscore character.

This PR resolves: 
- Using the database name the same as directory name on disk allows round-trip to the database location on disk.

Note:
- A potential issue might arise when the directory name has trailing space characters. MySQL do not allow database name with trailing space characters such as `mydb `, whereas Dolt allows. This will cause failing import of dolt dump to MySQL.


fulghum (approved):
Looks good!

PR #7032, opened 2023-11-21
CLI command for reflog

stephkyou:
Adds the CLI command `dolt reflog` which displays results form the `dolt_reflog()` table function


go/cmd/dolt/commands/reflog.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	args := ""
+	if refPlaceholder == "" && allFlag != "" {
+		args = allFlag
+	} else if refPlaceholder != "" && allFlag == "" {
+		args = refPlaceholder
+	} else if refPlaceholder != "" && allFlag != "" {
+		args = strings.Join([]string{refPlaceholder, allFlag}, ", ")
+	}
```
fulghum, line 128: (minor) You could make this code a little bit shorter/simpler if you created a `var args []string` and then just added the args to it in the if blocks on line 113 and line 117. That would avoid the messy conditional logic on lines 122-128. `args` would always have the args for the dolt_reflog function call, so you could just call `strings.Join(args, ", ")` on it to get the arguments plugged in. Give that a try and see what you think. I think it'll make this function a bit simpler/smaller/easier-to-read.

go/cmd/dolt/commands/reflog.go
```diff
@@ -0,0 +1,206 @@
...
+
+	query := strings.Join([]string{"SELECT ref, commit_hash, commit_message FROM DOLT_REFLOG(", args, ")"}, "")
```
fulghum, line 130: (minor) this would be just a _little_ bit simpler as a `Sprintf` call, instead of creating a literal slice and using strings.Join. 

```suggestion
	query := fmt.Sprintf("SELECT ref, commit_hash, commit_message FROM DOLT_REFLOG(%s)"}, args)
```

go/cmd/dolt/commands/reflog.go
```diff
@@ -0,0 +1,206 @@
...
+	commitHash    string
+	commitMessage string
+}
+
+// printReflog takes a list of sql rows with columns ref, commit hash, commit message. Prints the reflog to stdout
+func printReflog(rows []sql.Row, queryist cli.Queryist, sqlCtx *sql.Context) int {
+	var reflogInfo []ReflogInfo
+
+	// Get the current branch
+	res, err := GetRowsForSql(queryist, sqlCtx, "SELECT active_branch()")
```
fulghum, line 150: I believe this will fail in some cases... for example, if you check out a revision database of a tag or commit on a database, then `active_branch()` will return an error if I remember correctly. You might want to test for that and have some defensive code here in case we aren't able to determine the current branch.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -204,20 +195,28 @@ func (rltf *ReflogTableFunction) IsReadOnly() bool {
 }
 
 func (rltf *ReflogTableFunction) Expressions() []sql.Expression {
-	if rltf.refExpr != nil {
-		return []sql.Expression{rltf.refExpr}
-	}
 	return []sql.Expression{}
 }
```
fulghum, line 199: This is a little bit unsafe – we need to keep the expressions around so that they can be processed correctly...

These query plan trees will go through the analyzer and get modified and as expressions/nodes are processed, they will be replaced with new instances. If we aren't able to get the expressions, then we can't accurately create a new instance and the argument will likely drop out and the function may end up running differently than what the user requested.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -204,20 +195,28 @@ func (rltf *ReflogTableFunction) IsReadOnly() bool {
...
-	if len(expression) > 1 {
-		return nil, sql.ErrInvalidArgumentNumber.New(rltf.Name(), "0 or 1", len(expression))
+	if len(expression) > 2 {
+		return nil, sql.ErrInvalidArgumentNumber.New(rltf.Name(), "0 to 2", len(expression))
 	}
 
 	new := *rltf
-	if len(expression) > 0 {
-		new.refExpr = expression[0]
+	args, err := getDoltArgs(rltf.ctx, expression, rltf.Name())
```
fulghum, line 207: A similar concern here with expressions... `getDoltArgs` is going to evaluate the expression, but we don't know that it's safe to do so yet. We can't evaluate the expression until we know everything is fully resolved. For example, if this expression were a subquery that selected the name of a reference from another table, we'd try to eval that here without the query plan being fully resolved and would fail. The calls to `Eval` need to be limited to when we're executing / in our RowIter code.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -172,14 +166,11 @@ func (rltf *ReflogTableFunction) Schema() sql.Schema {
...
 func (rltf *ReflogTableFunction) Resolved() bool {
-	if rltf.refExpr != nil {
-		return rltf.refExpr.Resolved()
-	}
 	return true
 }
 
 func (rltf *ReflogTableFunction) String() string {
-	return fmt.Sprintf("DOLT_REFLOG(%s)", rltf.refExpr.String())
+	return fmt.Sprintf("DOLT_REFLOG(%s)", rltf.refName)
```
fulghum, line 173: We want the result of this function to be a string representation of what query the user requested, but because we've already evaled the expression, this will show up as the result, not the actual expression. For example, if I call `select * from dolt_reflog(concat('h', 'e', 'll', 'o'));` then this will return `DOLT_REFLOG('hello')`, which loses some of the information about what query is being run. 

This function also needs to represent if the `--all` flag has been specified as a parameter.

integration-tests/bats/reflog.bats
```diff
@@ -1,56 +1,231 @@
 #!/usr/bin/env bats
 load $BATS_TEST_DIRNAME/helper/common.bash
 
+setup() {
+    setup_common
```
fulghum, line 5: This was originally put into the tests (as opposed to in a `setup` function) so that the remote/local engine tests could run (we need to set the env vars before the `setup_common` code runs, otherwise the sql-server is already started and it's too late. That would probably have been a good comment! 😉 ). I noticed that those remote engine tests were turned off for the reflog bats test. I think it's generally better to keep them on if we at all can so that we have more testing coverage, but if you think they really shouldn't run for reflog let me know and I'm happy to chat more about it.

go/cmd/dolt/commands/reflog.go
```diff
@@ -0,0 +1,206 @@
...
+		pager := outputpager.Start()
+		defer pager.Stop()
+
+		for pos, info := range reflogInfo {
+			// TODO: use short hash instead
+			line := []string{fmt.Sprintf("\033[33m%s\033[0m", info.commitHash)} // commit hash in yellow (33m)
+
+			processedRef := processRefForReflog(info.ref, curBranch)
+			line = append(line, fmt.Sprintf("\033[33m(%s\033[33m)\033[0m", processedRef)) // () in yellow (33m)
+			line = append(line, fmt.Sprintf("HEAD@{%d}: %s\n", pos, info.commitMessage))
```
fulghum, line 183: I don't think we should include the `HEAD@{%d}` syntax... In Git, those are actually usable ways to reference commits, but we don't support that syntax, so it's not really doing anything helpful. Also, the name should be the name of the ref who's log you're searching. So... in Git, by default when you run `git reflog`, you are looking at the special `HEAD` reference (which we don't support in Dolt), so you see `HEAD@{1}`, etc. For example, when I select a different reflog to view with Git, by running `git reflog refs/heads/main`, I'll see `refs/heads/main@{11}` instead of `HEAD@{11}`. 

Another reason not to show these is because Dolt's reflog works a little differently than Gits – it allows you to see all the reflogs interwoven together with the `--all` flag, it would be difficult to make those labels meaningful/correct and even if they did match the ref and use the correct ID, they still wouldn't be useable for anything, so it just feels like noise that will get in customers' way.

fulghum (commented):
Looking good so far! Mostly some minor comments and some clarification on how Dolt's reflog is a little different than Git's and has some different requirements. 

There are also some comments about how the analyzer processes queries that you probably haven't seen before. I'm happy to explain more about how that works if you wanna know more details there or if any of those comments just don't make sense.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -211,14 +225,28 @@ func (rltf *ReflogTableFunction) Expressions() []sql.Expression {
...
-		new.refExpr = expression[0]
+
+	if len(expression) == 2 {
+		if expression[0].String() == "'--all'" && expression[1].String() == "'--all'" {
+			return nil, fmt.Errorf("error: multiple values provided for `all`")
+		}
+		if expression[0].String() != "'--all'" && expression[1].String() != "'--all'" {
+			return nil, fmt.Errorf("error: %s has too many positional arguments. Expected at most %d, found %d: %s", rltf.Name(), 1, 2, expression)
+		}
 	}
```
fulghum, line 241: Calling `expression[0].String()` and looking at the literal string value of the expression is going to be buggy. It works for basic cases, but it's fragile and will break. For example, it's totally valid to pass in `"--all"` in addition to `'--all'`. It's also possible that customers may want to pass in a SQL variable as the expression, and have that SQL variable hold the argument. There's lots of these potential edge cases, and it's better to make this code resilient from the start, instead of having to come back to it when a customer hits one of these. 

The best way to do this is to have the `WithExpressions` function simple assign the expressions as is to the new rftf instance and don't try to evaluate or directly inspect the expression values here. Do that evaluation and argument validation as part of the `PartitionRows` code – that's when we know the query plan will be fully resolved and it's safe to evaluate the expressions. (Technically... it's safe to evaluate the expressions whenever `Resolved()` returns true, but if you do that here, you'll still have to do it in the `PartitionsRows` function in case something wasn't resolved yet here, so it doesn't really make sense to do that anywhere else except for inside `PartitionsRows`).

fulghum (commented):
Looking very, very close now! Just a couple of things to follow up on. Hit me up if the analyzer-related comments don't make sense and I'm more than happy to explain in more detail.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -29,10 +29,11 @@ import (
...
 type ReflogTableFunction struct {
-	ctx      *sql.Context
-	database sql.Database
-	refExpr  sql.Expression
-	showAll  bool
+	ctx            *sql.Context
+	database       sql.Database
+	refAndArgExprs []sql.Expression
+	tabId          sql.TableId
+	colset         sql.ColSet
```
fulghum, line 36: These two members don't seem to be used anywhere, so should be removed.

go/libraries/doltcore/sqle/reflog_table_function.go
```diff
@@ -199,19 +213,17 @@ func (rltf *ReflogTableFunction) Schema() sql.Schema {
 }
 
 func (rltf *ReflogTableFunction) Resolved() bool {
-	if rltf.refExpr != nil {
-		return rltf.refExpr.Resolved()
+	for _, expr := range rltf.refAndArgExprs {
+		return expr.Resolved()
```
fulghum, line 217: You don't want to always return on the first expression – you need to check that all the expressions are resolved. 

You probably want something like `if expr.Resolved() == false { return false }` instead.

fulghum (approved):
Nice work! 🙌🏻 Just two really small things I noticed that should be super quick to fix before you merge this one into main.

go/libraries/doltcore/sqle/reflog_table_function.go
stephkyou: These were added in a separate PR, looks like they're being used a couple lines down.

PR #7035, opened 2023-11-21
Detect data conflicts even when one branch has the same binary representation as the ancestor.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/6746

This fixes a bug where the following situation fails to detect a conflict:

| Branch      | Value |
| ----------- | ----------- |
| Ancestor      | `(1, 2)`       |
| Left   | `(1, 2, NULL)`       |
| Right   | `(1, 2, 3)`        |

Both branches add a column but add different values to it. This should be a data conflict, but we currently don't detect it because Ancestor and Left have the same bytes in storage.


go/store/prolly/tree/merge.go
```diff
@@ -51,12 +51,14 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
 	serializer S,
 ) (final Node, stats MergeStats, err error) {
 
-	ld, err := DifferFromRoots[K](ctx, ns, ns, base, left, order)
+	// TODO: This function is only ever called to merge artifacts. Is is possible for the tuple desc to change here?
+	// If it does, do we want every row to appear in the diff?
```
fulghum, line 55: You're right that this code is only ever currently invoked when merging artifacts. 

You might consider passing in a boolean param to wire through to the calls to `DifferFromRoots`. That way, at least the caller would be being explicit about what behavior they wanted. The spot using this for merging artifacts could safely pass `false`, since the tuple description won't change for artifacts, as far as I know. 

Would be good to resolve this `TODO` before merging in though.

go/store/prolly/tree/map.go
```diff
@@ -42,7 +42,8 @@ func DiffOrderedTrees[K, V ~[]byte, O Ordering[K]](
 	from, to StaticMap[K, V, O],
 	cb DiffFn,
 ) error {
-	differ, err := DifferFromRoots[K](ctx, from.NodeStore, to.NodeStore, from.Root, to.Root, from.Order)
+	// TODO: If the schema has changed, do we want every row to be considered changed?
```
fulghum, line 45: The only interesting case is when a new, nullable column is added to the end of a table, right? That's the one case where we don't rewrite the table for the significant schema change as far as I know. Other less significant schema changes that wouldn't require a full table rewrite would be things like changing the length of a `varchar` field. 

Sounds like you're saying you think we may have this same issue with diff generation for a table's history?

fulghum (approved):
Approach looks good. Just a couple comments around trying to resolve the remaining TODOs in there.

go/libraries/doltcore/diff/diff_stat.go
```diff
@@ -131,7 +131,10 @@ func diffProllyTrees(ctx context.Context, ch chan DiffStatProgress, keyless bool
 		}
 	}
 
-	err = prolly.DiffMaps(ctx, f, t, func(ctx context.Context, diff tree.Diff) error {
+	// TODO: Use `vMapping` to determine whether or not the schema has changed. If it has, then all rows should
+	// count as modifications in the diff.
```
fulghum, line 135: (super minor) I was thinking about this some more... probably more accurate to say "to determine whether columns have been added or removed" instead of "schema has changed". That's what `vMapping` is already tracking of course, but a little more accurate since for other schema changes, like changing a column's type, we wouldn't want to show all rows as modified. 

```suggestion
	// TODO: Use `vMapping` to determine whether columns have been added or removed. If so, then all rows should
	// count as modifications in the diff.
```

fulghum (approved):
Looks great! 🚢

PR #7042, opened 2023-11-22
Run nullness checking on merged rows during three-way merges.

nicktobey:
This fixes one of the failures from https://github.com/dolthub/dolt/issues/7034, but it's the more important one (the panic)

Before, we would panic when attempting to compute a merged row if that row would contain a null in a non-null column.

Now, we allow this row to be generated, and we validate it before merging it into the table.


fulghum (approved):
Looks good!

PR #7044, opened 2023-11-23
Track whether or not rows actually need to be remapped to the result schema.

nicktobey:
This prevents a bunch of expensive computation in places where it isn't necessary.

Basically when computing the schema diff, we now compute:

- Whether the left side needs to be remapped to the final schema
- Whether the right side needs to be remapped to the final schema
- Whether we should invalidate all secondary indexes because a column changed its type.

The last one's the odd one out, since it's less a fact about the merge and more this derived property that could change later if we figure out a good way to update secondary indexes in the presence of schema change. But we can bikeshed the specifics once users aren't blocked by this.

We store all this data in a new `MergeInfo` struct, which gets passed around the merge logic.


fulghum (approved):
LGTM

PR #7075, opened 2023-11-30
[no-release-notes] go/libraries/doltcore/env/actions: remotes.go: Small fix to SetHeadToCommit our local remote ref when pushing even in fast-forward mode.

reltuk:


zachmu (approved):
LGTM, any tests of this?

PR #7076, opened 2023-11-30
[no-release-notes] go/libraries/doltcore/remotestorage: Fix some spurious retries on authentication failures and the likes when interacting with remotestorage.

reltuk:


macneale4 (approved):
LGTM

PR #7088, opened 2023-12-01
add `--single-branch` flag for `dolt clone`

stephkyou:
Adds `--single-branch` flag for `dolt clone` to match [git](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---no-single-branch) behavior more closely.

Fixes: https://github.com/dolthub/dolt/issues/3873


go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -125,6 +125,7 @@ func CreateCloneArgParser() *argparser.ArgParser {
 	ap.SupportsString(dbfactory.OSSCredsFileParam, "", "file", "OSS credentials file.")
 	ap.SupportsString(dbfactory.OSSCredsProfile, "", "profile", "OSS profile to use.")
 	ap.SupportsString(UserFlag, "u", "user", "User name to use when authenticating with the remote. Gets password from the environment variable {{.EmphasisLeft}}DOLT_REMOTE_PASSWORD{{.EmphasisRight}}.")
+	ap.SupportsFlag(SingleBranchFlag, "", "Clone only the history leading to the tip of a single branch, either specified by --branch or the primary branch remote's HEAD points at.")
```
macneale4, line 128:
```suggestion
	ap.SupportsFlag(SingleBranchFlag, "", "Clone only the history leading to the tip of a single branch, either specified by --branch or the remote's HEAD (default).")
```

integration-tests/bats/remotes.bats
```diff
@@ -1008,6 +1008,21 @@ create_five_remote_branches_main_and_master() {
     [[ "$output" =~ "remotes/origin/branch-two" ]] || false
 }
 
+@test "remotes: clone --single-branch does not create remote refs for all remote branches" {
```
macneale4, line 1011: Another test with the `--branch` argument seems appropriate.

go/libraries/doltcore/env/actions/clone.go
```diff
@@ -215,25 +215,25 @@ func CloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
 	// create remote refs corresponding to each of them. We delete all of
 	// the local branches except for the one corresponding to |branch|.
 	for _, brnch := range branches {
```
macneale4, line 217: can you change brnch to br or something more different from `branch`- someone could pretty easily screw that up in the future.

go/libraries/doltcore/env/actions/clone.go
```diff
@@ -215,25 +215,25 @@ func CloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
...
-			return fmt.Errorf("%w: %s; %s", ErrFailedToCreateRemoteRef, remoteRef.String(), err.Error())
-
-		}
-
 		if brnch.GetPath() != branch {
 			err := dEnv.DoltDB.DeleteBranch(ctx, brnch, nil)
 			if err != nil {
 				return fmt.Errorf("%w: %s; %s", ErrFailedToDeleteBranch, brnch.String(), err.Error())
 			}
+		} else if !singleBranch || brnch.GetPath() == branch {
```
macneale4, line 223: The only way you can reach this conditional is if the branches are equal (because we checked for inequality in the first conditional), so this block will always be entered regardless of the singleBranch flag, because `anything || TRUE` is always true.

You can convince me this is right if you can point me to tests which verify the remote tracking branches work without the --single-branch (or write a test for that if we don't have one).

macneale4 (changes requested):

go/libraries/doltcore/env/actions/clone.go
```diff
@@ -214,25 +214,26 @@ func CloneRemote(ctx context.Context, srcDB *doltdb.DoltDB, remoteName, branch s
 	// every branch in the remote. We iterate through local branches and
 	// create remote refs corresponding to each of them. We delete all of
 	// the local branches except for the one corresponding to |branch|.
-	for _, brnch := range branches {
-		cs, _ := doltdb.NewCommitSpec(brnch.GetPath())
-		cm, err := dEnv.DoltDB.Resolve(ctx, cs, nil)
-		if err != nil {
-			return fmt.Errorf("%w: %s; %s", ErrFailedToResolveBranchRef, brnch.String(), err.Error())
+	for _, br := range branches {
+		if !singleBranch || (singleBranch && br.GetPath() == branch) {
```
macneale4, line 218: small logical redundancy we talked about.

macneale4 (approved):
Ship when the nit we talked about is resolved and the tests continue to pass.

PR #7090, opened 2023-12-02
Added a periodic heartbeat event metric to `sql-server`

zachmu:
Also adds new constants for Doltgres as an application


zachmu:
Tested this manually, a little tricky to test via bats but open to suggestions

PR #7092, opened 2023-12-04
migrate `dolt gc` to use sql queries

stephkyou:
This change updates `dolt gc` to use the appropriate sql engine to generate results.


fulghum (approved):
Looks good! 🚢

PR #7094, opened 2023-12-04
fix inf loop for creds search on windows

JCOR11599:
There was an infinite loop for `dolt init` and `dolt sql` when there isn't a `.dolt` directory in current directory or any children directory. This issue was specific to windows, because we were only checking for `/` and not `\`.


reltuk (approved):
LGTM!

PR #7104, opened 2023-12-05
Feature: Support `BLOB/TEXT` columns in unique indexes, without requiring a prefix length

fulghum:
Depends on: ...

Related to: https://github.com/dolthub/dolt/issues/7040


go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -842,11 +842,41 @@ func (di *doltIndex) HandledFilters(filters []sql.Expression) []sql.Expression {
...
+// stores a hash of the content, instead of the content itself. This is currently limited to unique indexes, which can
+// use this property to store hashes of TEXT or BLOB fields and still efficiently detect uniqueness.
+func (di *doltIndex) HasContentHashedField() bool {
+	// content-hashed fields can currently only be used in unique indexes
+	if !di.IsUnique() {
+		return false
+	}
+
+	contentHashedField := false
+	di.indexSch.GetPKCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
```
max-hoffman, line 855: we're returning true even for non-zero prefixes here, is that the intended behavior?

max-hoffman (approved):
lgtm, just the one comment about when blob prefixes matter. My understanding of your last PR was that blob indexes with prefixes should follow the behavior of other secondary indexes, but I might be mistaken

fulghum:
> Does this work with case-insensitive collations? There could be two rows with different hashes that need to compare equal for unique indexes? (I wonder what MariaDB does in this case.)

Good feedback! Sure seems like case-insensitivity (or accent-insensitivity) would mess up the hashing matching. I'll play with this a bit today and if I can't figure out a fix I'll at least add a skipped test and mention this in a tracking issue for any follow ups.

go/libraries/doltcore/sqle/index/dolt_index.go
fulghum: Oh, great catch! Thank you! Added the missing prefixLength check in 09540a664454c07516d497473e309b6fa7ea35eb.

fulghum:
> Thoughts for how we could deal with collations:
> 
> * The secondary index could be keyed on a normalized value. If we can't find a library for normalizing every collation, we could just find a library that does standard unicode normalization plus case-folding, which will probably be correct for most users. Here's one library that does that, but only for UTF-8: https://github.com/JuliaStrings/utf8proc
> * If we need to be 100% correct, we could just only allow these indexes on binary collations.

I just pushed a commit to the GMS PR for this feature that errors out if you try to use a `TEXT` column with a case-insensitive or accent-insensitive collation in a unique index without specifying a prefix length. I think that's a good compromise for right now – we detect this problematic case and prevent users from hitting it, but we still allow use of `TEXT` columns with this feature when they are case-sensitive. I'll create a tracking issue for a couple follow ups for this feature and we can prioritize those as customers need them. For that work, I think you're right that the best way to address this is to normalize the value before hashing it. I don't think it's a huge change, but it's also not a small change, mostly because we are currently relying on the hash being a content-address and that won't be true anymore if we have to normalize the content before hashing.

PR #7115, opened 2023-12-06
serialize `on update` expressions

JCOR11599:
This PR add serialization for `ON UPDATE` expressions on columns. 

This change requires changes to the flatbuffer schema as we have to store an additonal field.
Older versions are able to read and modify databases in the newer version without any error, but any `ON UPDATE` functionality is missing (which makes sense). Tables with `ON UPDATE ...` defined on them won't be displayed and performing an `UPDATE` won't trigger any of the defined expressions to run. So, this should be a Feature Verison bump.

companion pr: https://github.com/dolthub/go-mysql-server/pull/2188

fixes https://github.com/dolthub/dolt/issues/6030


go/cmd/dolt/commands/diff.go
```diff
@@ -909,6 +909,7 @@ func schemaFromCreateTableStmt(createTableStmt string) (schema.Schema, error) {
 			TypeInfo:      typeInfo,
 			Default:       defBuf.String(),
 			Generated:     "",    // TODO
+			OnUpdate:      "",    // TODO
```
zachmu, line 912: I don't remember why this is difficult to fix. Why can't we use a  SqlEngine here and issue two "show create table" statements?

zachmu: If it's not trivial, can you do as a followup?

zachmu (approved):
LGTM but see comments on GMS PR

go/cmd/dolt/commands/diff.go
JCOR11599: I'm not sure what's wrong here (at least with Generated and Virtual). I'm able to create tables with generated and virtual columns and the `SHOW CREATE ...` statement displays them just fine when calling `dolt diff`.

However, there is a bug with `ON UPDATE`:
We can parse ...`ON UPDATE CURRENT_TIMESTAMP()`, but not `ON UPDATE (CURRENT_TIMESTAMP())`

Should be a simple vitess parser fix, will be fixed in a follow.
Tracking issue here: https://github.com/dolthub/dolt/issues/7129#issuecomment-1854798812

integration-tests/bats/helper/common.bash
```diff
@@ -81,7 +81,7 @@ assert_feature_version() {
     # Tests that don't end in a valid dolt dir will fail the above
     # command, don't check its output in that case
     if [ "$status" -eq 0 ]; then
-        [[ "$output" =~ "feature version: 7" ]] || exit 1
+        [[ "$output" =~ "feature version: 8" ]] || exit 1
```
zachmu, line 84: Revert, right?

zachmu (approved):
LGTM

PR #7124, opened 2023-12-07
[no-release-notes] Refactored server initialization code

zachmu:
Separated initialization and configuration from serving. This is primarily to support Doltgres.


Hydrocharged (approved):
LGTM!

PR #7125, opened 2023-12-08
Clear out reflog contents consistently after GC

fulghum:
When GC is executed, the in-memory reflog data buffer was being cleared out so that only the one, most recent entry was kept. For a `sql-server`, this means you can log back in and see one entry in the reflog. For a `sql` CLI command, since it's a new process running now, it doesn't have the reflog data buffer in memory anymore, so it has an empty reflog. This meant there was a slight behavior difference between using GC and checking the reflog depending on whether you are connecting to a `sql-server` or using the `sql` CLI (or silently connecting to a `sql-server` through the `sql` CLI command when running in local-remote mode). 

To smooth this small inconsistency out, the reflog data buffer is now completely cleared out during GC.


stephkyou (approved):
LGTM

PR #7126, opened 2023-12-08
fixes `dolt version` out of date warning

stephkyou:
Changes `dolt version` out of date warning to do a check against GitHub if the current build version is ahead of the stored latest release version. Fixes inconsistencies where the current build is running a version that was released within the past week.

### DIFF
--- a/go/cmd/dolt/commands/version.go
+++ b/go/cmd/dolt/commands/version.go
@@ -148,24 +148,12 @@ func checkAndPrintVersionOutOfDateWarning(curVersion string, dEnv *env.DoltEnv) 
 			return errhand.BuildDError("error: failed to read version check file").AddCause(err).Build()
 		}
 
-		vCheckData := strings.Split(string(vCheck), ",")
-		if len(vCheckData) != 2 {
-			// formatting or data is wrong, so just overwrite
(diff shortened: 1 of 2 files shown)


go/cmd/dolt/commands/version.go
```diff
@@ -174,6 +175,17 @@ func checkAndPrintVersionOutOfDateWarning(curVersion string, dEnv *env.DoltEnv)
...
+	if verr != nil {
+		return verr
+	}
+	if isEarlier {
+		latestRelease, verr = getLatestDoltReleaseAndRecord(path, dEnv)
+		if verr != nil {
+			return verr
+		}
+	}
 	if curVersion != latestRelease {
```
zachmu, line 189: This fix is too complicated and targets the wrong thing. The bug was we were checking for equality, when we should have been checking for earlierThan. Just do that instead, the logic to re-fetch the latest release if `isEarlier` is unnecessary.

Also add a bats test for this -- write a `latestRelease` file with an older version and assert that it doesn't trigger the warning.

zachmu (commented):
I think this could be improved, see comments

zachmu (approved):
LGTM, thanks for the changes

PR #7127, opened 2023-12-08
Disable performance bats test.

nicktobey:
I originally settled on a 50-second timeout for these tests because a 100-second timeout was too high to consistently detect the performance regression that motivated https://github.com/dolthub/dolt/pull/7044.

I suspect that any timeout high enough to avoid spurious failures is too high to reliably detect actual regressions. Detecting performance regressions is important, but this isn't the way to do it.

We shouldn't be blocking CI on this. We should incorporate these performance tests into our daily benchmarking instead.


fulghum (approved):
🚢

PR #7135, opened 2023-12-08
Removing `--old-format` option from `dolt init`

fulghum:
Removes the `--old-format` flag from `dolt init`. I decided to keep the flag for `--new-format` for a little longer (even though it's the default, so that's a no-op) just in case people are still using it in scripts. 

This one should technically be released as a minor version bump, so we can hold off on merging it in if we want to coordinate it with another minor version bump change so they go out together.


stephkyou (approved):
LGTM

PR #7137, opened 2023-12-09
Adding tests for column name metadata through MySQL C++ and Rust Connector

fulghum:
Added assertions that result set column name metadata is populated for the C++ and Rust MySQL Connector libraries.


stephkyou (approved):
LGTM

PR #7143, opened 2023-12-11
Poorly formatted version doesn't error

stephkyou:
Fixes an issue where bad data in `version_check.txt` returns an error. This change makes version overwrite any bad data in this file instead.

Fixes: https://github.com/dolthub/dolt/issues/7138


fulghum (approved):
Looks good! 🚢

PR #7153, opened 2023-12-13
proto/third_party/flatbuffers,go/gen/fb/serial: Bump flatc to pick up the version that only has Try accessors.

reltuk:


zachmu (approved):
LGTM

PR #7158, opened 2023-12-13
/{go,integration-tests}: add oci blobstore implementation

coffeegoddd:
This PR adds an Oracle Cloud Infrastructure (OCI) `blobstore` implementation. 

`OCIBlobstore.Concatenate` is unimplemented as OCI doesn't currently support building composite objects.

`OCIBlobstore.Put` will use the `objectstorage.PutObject` method for uploads smaller than the minimum multipart upload part size of 10MB, but will use the Multipart Upload process for all uploads larger than 10MB.

Objects have a `VersionId` as well as an `ETag`, however the `ETag` value is used for the conditional match header, not the `VersionId`, see https://github.com/oracle/oci-go-sdk/blob/v65.54.0/objectstorage/get_object_request_response.go?s=719:5196#L39

UPDATE: this PR also contains changes so that an Oracle bucket can be used as a remote and backup


go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+		req.Range = &byteRange
+	}
+
+	res, err := bs.client.GetObject(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Content.Close()
+
+	p, err := createTempFile()
```
reltuk, line 409: I don't think this should be structured to use temporary files and a final copy. Instead, I think:

1) the destination path should be opened for with `os.Create()` before we start calling downloadPart.
2) each `downloadPart` invocation should be given an `io.Writer` to dump its body into.
3) That `io.Writer` instance should be `io.NewOffsetWriter()` constructed against the `*os.File` and the total offset of the range which is being downloaded.

Does that make sense? This works on macOS, Linux and Windows AFAICT at least...

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+		f:    f,
+	}, nil
+}
+
+func downloadParts(ctx context.Context, objectName, namespace, etag string, requestedSize, totalSize, maxBatchSize, offSet, minPartSize, maxPartNum int64, concurrentListeners int, downloadF downloadFunc) ([]*toDownload, error) {
+	numParts, partSize := getNumPartsAndPartSize(requestedSize, minPartSize, maxPartNum)
+	completedPartChan := make(chan *toDownload, numParts)
+	completedParts := make([]*toDownload, numParts)
+
+	batch := make([]*toDownload, 0)
```
reltuk, line 500: I don't think we need much if any channels here. Given the suggestion on OffsetWriter, this should be something like:

```go
f, err := os.Create(file)
if err != nil {
  return err
}
defer f.Close()
eg, egCtx := errgroup.WithContext(ctx)
eg.SetLimit(MaxConcurrency)
for i := 0; i < totalSize && egCtx.Err() == nil; i+=batchSize {
  wr := io.NewOffsetWriter(f, i)
  blobrange := clampedBlobRange(i, batchSize, totalSize)
  eg.Go(func() error {
    return downloadPart(egCtx, wr, objectName, namespace, etag, blobrange)
  })
}
return eg.Wait()
```

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+				return nil
+			}
+		}
+	})
+
+	return eg.Wait()
+}
+
+func uploadParts(ctx context.Context, objectName, uploadID string, numParts, concurrentListeners int, totalSize, maxBatchSize int64, reader io.Reader, uploadF uploadFunc) ([]objectstorage.CommitMultipartUploadPartDetails, error) {
+	completedPartChan := make(chan objectstorage.CommitMultipartUploadPartDetails, numParts)
```
reltuk, line 597: Again, I don't know if need much in the way of channels here. Just make the plan, and then iterate of the batches. For each batch, just make the byte slice, read it fully into memory, and then call `eg.Go()` to upload it. Use `eg.SetLimit()` to put a limit on the number of concurrent uploads and to keep memory from blowing out.

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+			offSet = totalSize + br.offset
+		} else {
+			offSet = br.offset
+		}
+	}
+
+	return requestedSize, totalSize, offSet, nil
+}
+
+func getUploadInfo(partSize, maxPartNum int, r io.Reader) (int, int64, io.Reader, error) {
```
reltuk, line 745: This buffers the entire file into memory. I don't think we want to do this. We either know the size of the upload or we can get it by checking for like a Size() method on the Reader, or we can just go forward with a conservative estimate on the chunk size.

In this case, we do always know the size, I think...CopyTableFile is given it explicitly, and Persist is given the memtable in memory, so it's already a bytes.Buffer. Whether that's easy to thread through to this blobstore, I'm not sure, but we already made an entirely new table persister for this...

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+			if serr.GetHTTPStatusCode() == 404 {
+				return nil, "", NotFound{"oci://" + path.Join(bs.bucketName, absKey)}
+			}
+		}
+		return nil, "", err
+	}
+
+	// handle negative offset and positive length
+	if br.offset < 0 && br.length > 0 {
+		trimmedR, err := trimContent(res, br.length)
```
reltuk, line 204: Why isn't this just a LimitReader wrapper? Because we might not get the full response and we want to consider that an error?

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+	}
+	_, err = bs.client.HeadObject(ctx, objectstorage.HeadObjectRequest{
+		NamespaceName: &namespace,
+		BucketName:    &bs.bucketName,
+		ObjectName:    &absKey,
+	})
+	if err == nil {
+		return true, nil
+	}
+	return false, nil
```
reltuk, line 125: This seems a little disingenuous. Seems like we should do the dance:

```go
	if serr, ok := common.IsServiceError(err); ok {
		// handle not found code
		if err.GetHTTPStatusCode() == 404 {
			return false, nil
		}
	}
	return false, err
```

or something like that...

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+		return bs.nameSpace, nil
+	}
+	request := objectstorage.GetNamespaceRequest{}
+	r, err := bs.client.GetNamespace(ctx, request)
+	if err != nil {
+		return "", err
+	}
+	bs.nameSpace = *r.Value
+	return bs.nameSpace, nil
+}
```
reltuk, line 106: This makes `bs` non-intuitively non-thread-safe. Can we do this at construction time instead?

reltuk: Or put it behind a `(*sync.Once).Do()` I suppose...

go/store/blobstore/oci.go
```diff
@@ -0,0 +1,796 @@
...
+// CheckAndPut will check the current version of a blob against an expectedVersion, and if the
+// versions match it will update the data and version associated with the key
+func (bs *OCIBlobstore) CheckAndPut(ctx context.Context, expectedVersion, key string, reader io.Reader) (string, error) {
+	return bs.upload(ctx, expectedVersion, key, reader)
+}
+
+// At the time of this implementation, Oracle Cloud does not provide a way to create composite objects
+// via their APIs/SDKs.
+func (bs *OCIBlobstore) Concatenate(ctx context.Context, key string, sources []string) (string, error) {
+	panic("concatenate is unimplemented on the oci blobstore")
```
reltuk, line 228: Better to just return an error? I agree something is misconfigured if this is being called...

Or can we restructure the implementations and interfaces so that this isn't necessary here at all, and only the GCS blobstore is implementing a different interface that actually supports this and the table persister that works with that thing take that?

reltuk (commented):
OK, some comments! I'm going to let you rework it a bit. Hopefully these comments are helpful :).

go/store/blobstore/oci.go
coffeegoddd: theres something im not understanding about how to implement this correctly... even after writing tests that i think correctly copy data in parts correctly, somethings not quite right, im always getting the `invalid or corrupt tablefile` error...

go/store/blobstore/oci.go
coffeegoddd: havent been able to get this to work

go/store/blobstore/oci.go
coffeegoddd: im not really sure what you mean here either... how do i do this without channels in a way that lets me collect all the etags of the uploaded parts in order to call CommitMultipartUpload?

PR #7159, opened 2023-12-13
Bug fix: `dolt_blame_<tablename>` view should backtick quote identifiers

fulghum:
Fixes the `dolt_blame_<tablename>` view definitions so that table names and primary key column names are backtick quoted. Without this, customers can't use the `dolt_blame_<tablename>` views if the table name or primary key columns contain characters (e.g. "-") that require identifier quoting.


jennifersp (approved):
Looks good!

PR #7161, opened 2023-12-14
Return a different error message for tx commit conflicts depending on `@autocommit`

fulghum:
When a transaction commit conflict occurs, the next steps to take are slightly different depending on whether `@autocommit` is enabled or not. This PR changes the error message we return to describe those steps in more detail when `@autocommit` is enabled.


timsehn (approved):
Error messages and bats tests look great. Go code is out of my lane though.

PR #7169, opened 2023-12-14
More detailed warning message for import schema not matching

stephkyou:
Updates the warning message for `dolt table import` to print when the import schema does not match the destination table's schema, not just when there are fewer columns in the import file. Also updates the warning message to list the column names that differ between the import file and the destination table's schema.

Resolves: https://github.com/dolthub/dolt/issues/6113


macneale4 (approved):
```
                                \
                                  \   O,
                        \___________\/ )_________/
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ \~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                        \
```

PR #7173, opened 2023-12-14
go: sqle: cluster: Fix MySQLDb and dolt_branch_control replication for the cases where you have more than one standby server in the cluster.

reltuk:
Previously these changes would only get replicated to the first database in the cluster.


zachmu (approved):
LGTM!

PR #7178, opened 2023-12-15
go: sqle: cluster: Create missing remotes for cluster replication on startup if they do not already exist.

reltuk:
Previously `dolt sql-server` would error if there were existing databases and they were missing any of the remotes listed as cluster replication remotes in the config.yaml. This changes it so that those remotes are automatically created at startup instead.

### DIFF
--- a/go/libraries/doltcore/sqle/cluster/controller.go
+++ b/go/libraries/doltcore/sqle/cluster/controller.go
@@ -299,9 +299,14 @@ func (c *Controller) applyCommitHooks(ctx context.Context, name string, bt *sql.
 	dialprovider := c.gRPCDialProvider(denv)
 	var hooks []*commithook
 	for _, r := range c.cfg.StandbyRemotes() {
+		remoteUrl := strings.Replace(r.RemoteURLTemplate(), dsess.URLTemplateDatabasePlaceholder, name, -1)
 		remote, ok := remotes.Get(r.Name())
 		if !ok {
-			return nil, fmt.Errorf("sqle: cluster: standby replication: destination remote %s does not exist on database %s", r.Name(), name)
+			remote = env.NewRemote(r.Name(), remoteUrl, nil)
+			err := denv.AddRemote(remote)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


integration-tests/go-sql-server-driver/tests/sql-server-cluster.yaml
```diff
@@ -1736,3 +1736,89 @@ tests:
...
+        columns: ["count(*)"]
+        rows:
+        - ["2"]
+    - exec: 'SET @@GLOBAL.dolt_cluster_ack_writes_timeout_secs = 1'
+    - exec: "create table vals (id int primary key)"
+    - query: "show warnings"
+      result:
+        columns: ["Level", "Code", "Message"]
+        rows:
+        - ["Warning", "3024", "Timed out replication of commit to 1 out of 2 replicas."]
```
reltuk, line 1817: Would this test be better if I just brought up the third server as well?

reltuk: Answered my own question with a "yes" here. Update the test.

go/libraries/doltcore/sqle/cluster/controller.go
```diff
@@ -299,9 +299,14 @@ func (c *Controller) applyCommitHooks(ctx context.Context, name string, bt *sql.
...
 	var hooks []*commithook
 	for _, r := range c.cfg.StandbyRemotes() {
+		remoteUrl := strings.Replace(r.RemoteURLTemplate(), dsess.URLTemplateDatabasePlaceholder, name, -1)
 		remote, ok := remotes.Get(r.Name())
 		if !ok {
-			return nil, fmt.Errorf("sqle: cluster: standby replication: destination remote %s does not exist on database %s", r.Name(), name)
+			remote = env.NewRemote(r.Name(), remoteUrl, nil)
+			err := denv.AddRemote(remote)
+			if err != nil {
+				return nil, fmt.Errorf("sqle: cluster: standby replication: could not create remote %s for database %s: %v", r.Name(), name, err)
```
bheni, line 308: %w for the error?

bheni (approved):
lgtm

PR #7179, opened 2023-12-15
go: sqle: DoltDatabaseProvider: If we encounter an error while creating a database, try to clean up after ourselves so we do not leave partially initialized database state around.

reltuk:


zachmu (approved):
LGTM!

PR #7181, opened 2023-12-16
Warning for nonsense variables in config

stephkyou:
Prints a warning if any nonsense variables are found in global or local config files whenever any CLI command is run. Also restricts users from adding any nonsense variables with `dolt config --add` or `dolt config --set`.

Resolves: https://github.com/dolthub/dolt/issues/7165


fulghum (approved):
Looks good! Just a couple minor suggestions.

PR #7191, opened 2023-12-19
Correctly resolve merge conflicts for types where two values with different representations can compare equal.

nicktobey:
Prior to this PR, we used the column's comparison operator when attempting to detect if concurrent changes to a column could be resolved during a merge operation.

However, this can cause a problem when a type has two different representations that compare equal.

- Some types have different representations that compare equal but can exhibit different behavior. For example, strings in a case-insensitive text column compare equal but behave differently when used in a context with a different collation and when displayed to the user. In this case, the merge algorithm should not consider them equal.
- Other types have different representations that compare equal and exhibit identical behavior. For example, we store Decimal types as a (integer, exponent) pair, but we don't normalize this pair. So pairs like (45, 1), (450, 0), and (4500, -1) are all indistinguishable, including comparing equal. Since there's no way to distinguish these and the user should not have to know or care which representation is used, there is no danger in resolving potential merge conflicts by simply choosing a value. However, care must be taken that the same value is used regardless of the direction of the merge, in order to guarentee that the final table hash is the same.

In order to accomplish both of these, this PR does two things:
- Forces the three-way merge algorithm to use a default comparator instead of the table's comparator. This defaults to a binary collation for strings.
- In the event that a merge conflict is auto-resolvable because the two values being assigned are equal to each other, sort them and use the higher one instead of choosing one branch arbitrarily.

### DIFF
--- a/go/libraries/doltcore/doltdb/durable/index.go
+++ b/go/libraries/doltcore/doltdb/durable/index.go
@@ -15,6 +15,7 @@
 package durable
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -50,6 +51,8 @@ type Index interface {
 	// Returns the serialized bytes of the (top of the) index.
 	// Non-public. Used for flatbuffers Table persistence.
 	bytes() ([]byte, error)
+
+	DebugString(ctx context.Context) string
 }
 
 // IndexSet stores a collection secondary Indexes.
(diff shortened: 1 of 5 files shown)


fulghum (approved):
Looks great! Nice work on this one! 🚢

fulghum:
I restarted the sysbench runner job, but it failed with the same error. Seems like there might have been a regression in our parser? Here's a simpler repro of the syntax that's failing; it's the `NOT NULL` appearing after the check constraint. This works in MySQL, but doesn't seem to work in the latest Dolt builds:
`create table t123 (c1 varchar(5) check (c1 in ('v1', 'v2')) NOT NULL);`

PR #7194, opened 2023-12-19
[no-release-notes] Changing sysbench_runner CI test to explicitly specify script tests

fulghum:
We have a separate CI job to test that the `sysbench_runner` util is working. It wasn't specifying any tests, so all tests in the `ScriptDir` were being run. Once postgres specific tests landed there, it started failing this CI job. 

This changes the configuration for the `sysbench_runner` CI test job to explicitly specify which tests to run, and includes three basic tests, just for a smoke test to ensure `sysbench_runner` is working correctly (not actually for the perf results).

### DIFF
--- a/go/performance/continuous_integration/config.json
+++ b/go/performance/continuous_integration/config.json
@@ -18,6 +18,22 @@
     }
   ],
   "ScriptDir":"/sysbench-lua-scripts",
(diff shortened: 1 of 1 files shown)


zachmu (approved):
Thanks for fixing!

PR #7205, opened 2023-12-22
[no-release-notes] go/libraries/doltcore/sqle: remotesrv.go: Small cleanup so that CreateUnknownDatabases is a property of the SQL DBCache, not remotesrv.ServerArgs.

reltuk:
remotesrv.ServerArgs should be purely concerned with providing the values to the remotesrv implementation that it needs to actually run the gRPC server. This setting, CreateUnknownDatabases, it not in the control of the gRPC server itself, as currently implemented, but is purely a behavior of the DBCache implementation given the remotesrv Server.

Move this argument to (only) place where it actually has an effect.


macneale4 (approved):
LGTM

PR #7206, opened 2023-12-26
upgrade crypto

macneale4:
Dependabot sent two proposed changes to update crypto dependency, but they don't build unless they are combined.

https://github.com/dolthub/dolt/pull/7190
https://github.com/dolthub/dolt/pull/7188


stephkyou (approved):
LGTM

PR #7214, opened 2023-12-27
Additional `dolt schema import` types

stephkyou:
This change adds support for `JSON` and `TEXT` columns in `dolt schema import`.

Fixes: https://github.com/dolthub/dolt/issues/2401


go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -181,7 +182,19 @@ func leastPermissiveType(strVal string, floatThreshold float64) typeinfo.TypeInf
 		return typeinfo.BoolType
 	}
 
-	return typeinfo.StringDefaultType
+	if strings.Contains(strVal, "{") {
+		var json map[string]interface{}
+		err := json2.Unmarshal([]byte(strVal), &json)
```
fulghum, line 187: `json2` could benefit from a better name. I'd recommend not renaming the package import, so you can keep that as `json`, and just use a different variable name for `json` on line 186. It looks like that single variable is the only conflict with the `json` package name, so should be easy to change. 

Alternatively though, if you do need to rename the package, something like `jsonenc` or `encjson` would be a little more readable than `json2`.

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -181,7 +182,19 @@ func leastPermissiveType(strVal string, floatThreshold float64) typeinfo.TypeInf
 		return typeinfo.BoolType
 	}
 
-	return typeinfo.StringDefaultType
+	if strings.Contains(strVal, "{") {
```
fulghum, line 185: JSON data could also come in as a JSON array, instead of a JSON map, for example:

```sql
mysql> create table json1 (pk int primary key, j1 json);
Query OK, 0 rows affected (0.00 sec)

mysql> insert into json1 values (1, "[1,2,3]");
```

It would be good to look for `[` as another hint that a column's data may be in JSON format and attempt a test parse, too.

fulghum (commented):
Looks pretty good. Would be nice to add a test for a JSON array in addition to the test for a JSON map. Would also be good to test out the `TEXT` support with some longer data to assert `TEXT` columns are generated correctly.

go/libraries/doltcore/env/actions/infer_schema.go
```diff
@@ -181,7 +182,25 @@ func leastPermissiveType(strVal string, floatThreshold float64) typeinfo.TypeInf
...
+		var jsonMap map[string]interface{}
+		err := json.Unmarshal([]byte(strVal), &jsonMap)
+		if err == nil {
+			return typeinfo.JSONType
+		}
+	} else if strings.Contains(strVal, "[") {
```
fulghum, line 191: I *think* you may want to remove the `else` here and just have this be a separate, free-standing `if` block. 

Otherwise... if someone uses a JSON string like "[{a:'foo'}, {b:'bar'}]", it would attempt to parse it as a map and fail, and then not take the `else` path in the second block. 

Worth at least a quick test to see how the code acts in that case. 

Alternatively... you could probably collapse these two if blocks into one that contains `{` or `[` and just pass in a `interface{}` for the second arg to Unmarshall, especially since you don't need to look at the unmarshalled value.

fulghum (approved):
Very nice! 🙌 

Just one more idea of another test case to try.

PR #7230, opened 2024-01-02
Allow for three way merges to automatically resolve concurrent changes to a JSON object when the changes affect different keys.

nicktobey:
I recommend looking at one commit at a time.

This PR implemented the "Primary Goal" outlined in [this doc](https://docs.google.com/document/d/1yVIOm7vVzW868dk0h2r6vwJYPyVhPcqIjUgOxhVmUzA/edit).

Basically, when two branches in a merge both modify the same JSON value, but touch different keys, dolt should be able to automatically resolve the merge by doing a Three Way Merge that compares the value on both branches to the value at a common ancestor.

This implementation deviates from the strategy outlined in the doc in one major way: when a branch contains an edit to a nested object, we don't run the diff or merge algorithms recursively. Instead, we generate a diff that contains a [JSON path](https://dev.mysql.com/doc/refman/8.0/en/json.html#json-path-syntax) to the modified object. Doing this allows us to generate all the diffs in a single linear pass, and is still easy to reason about what changes are conflicts.

We discussed only performing this merge on JSON objects below a certain size for performance reasons, but the size of the object isn't actually visible to the merge algorithm. Given that this algorithm only runs for rows that would otherwise have a merge conflict, all dolt merges that succeed before this PR should see zero performance impact from this change, so I'm not worried about performance.


go/libraries/doltcore/merge/three_way_json_differ.go
```diff
@@ -0,0 +1,204 @@
...
+	Op tree.DiffOp
+	// a partial set of document values are set
+	// depending on the diffOp
+	Key                       string
+	Base, Left, Right, Merged *types.JSONDocument
+}
+
+func (differ *ThreeWayJsonDiffer) Next() (ThreeWayJsonDiff, error) {
+	for {
+		differ.loadNextDiff()
```
fulghum, line 51: missing an error check here

go/store/prolly/tree/json_diff_test.go
```diff
@@ -0,0 +1,163 @@
...
+		name: "nested insert",
+		from: types.JsonObject{"a": types.JsonObject{}},
+		to:   types.JsonObject{"a": types.JsonObject{"b": 1}},
+		expectedDiffs: []JsonDiff{
+			{
+				Key:  "$.\"a\".\"b\"",
+				To:   &types.JSONDocument{Val: 1},
+				Type: AddedDiff,
+			},
+		},
```
fulghum, line 88: (minor) Just an idea, not critical... might be nice to have multiple expected diffs in a test case so that you could assert they get returned in the correct/expected order.

go/store/prolly/tree/json_diff.go
```diff
@@ -0,0 +1,182 @@
...
+	From, To *types.JSONDocument
+	Type     DiffType
+}
+
+type jsonKeyPair struct {
+	key   string
+	value interface{}
+}
+
+// Differ computes the diff between two json objects.
```
fulghum, line 38:
```suggestion
// JsonDiffer computes the diff between two json objects.
```

fulghum (approved):
Looks great! Just a few minor comments. Overall, very clean.

go/libraries/doltcore/merge/schema_merge_test.go
```diff
@@ -1142,6 +1145,195 @@ var simpleConflictTests = []schemaMergeTest{
...
+			{
+				// TODO: Should we be able to merge this?
+				name:         "object inside array conflict",
+				ancestor:     singleRow(1, 1, 1, `{ "key1": [ { } ] }`),
+				left:         singleRow(1, 2, 1, `{ "key1": [ { "key2": "value2" } ] }`),
+				right:        singleRow(1, 1, 2, `{ "key1": [ { "key3": "value3" } ] }`),
+				dataConflict: true,
+			},
+			{
+				// TODO: Should we be able to merge this?
```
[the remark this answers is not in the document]
nicktobey: I'd be excited to merge additions to arrays too, and there's definitely low-hanging fruit that would cover most use cases: typically cases where each side of the merge is changing non-overlapping regions of the array.

The trouble is that even with a lot of constraints and assumptions, there's going to be a lot of degenerate cases where the diff is ambiguous or wrong or slow (NP hard). I need to think more about the best way to do this.

go/store/prolly/tree/z_encoding.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package index
```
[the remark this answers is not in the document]
nicktobey: It got moved along with the `PutField` and `GetField` functions because `PutField` depends on `ZCell` for writing geometry values.

Those functions got moved because the improved `OutputProllyNode`, which is used for printing prolly tree nodes for debugging, now depends on `GetField` in order to convert the rows into human-readable values.

In the commit description I acknowledge that `prolly/tree` isn't the best place for them and it might make more sense to move them to either `store/val` or to their own package. But if they're only used for indexes as you say it might make the most sense to keep them `sqle/index`, and then find another way to break the dependency cycle.

fulghum (approved):
Flag/var to opt-out of json merging looks good!

PR #7233, opened 2024-01-02
Delete branch message as a push result

macneale4:
Currently every push results in the message "[new branch] ..." which is misleading. Unfortunately fixing this for all updates is way harder than I want to dig into at the moment, so I'm going to make the only case we can detect easily (delete) print a better message.


stephkyou (approved):
LGTM

PR #7252, opened 2024-01-04
dolt_pull performance improvement

macneale4:
`dolt_pull` currently performs unnecessary FastForwards for remote branches. There is a more optimal mechanism to fetch all remotes, actions.FetchRefSpecs, and by using that we greatly increase the speed of `dolt_pull` when there are lot of remote branches.


reltuk (approved):
LGTM!

PR #7256, opened 2024-01-04
[no-release-notes] bump and test for required commit_diff index predicates

max-hoffman:
re: https://github.com/dolthub/go-mysql-server/pull/2238


go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
```diff
@@ -5506,6 +5506,24 @@ var SystemTableIndexTests = []systabScript{
 			},
 		},
 	},
+	{
+		name: "required index lookup in join",
```
zachmu, line 5510: Test is a bit fragile, relies on the hint system in ways that will silently make the test useless if we ever change the semantics even slightly. This is especially bad because it relies on us ignoring the hint entirely and choosing a different join plan.

It would be better to have a larger set of tests involving joins onto the dolt_commit_diff table with various circumstances we think are likely to trigger various join plans. Fine to supplement that with hints / special engine setup as well but those shouldn't be the only tests of this behavior.

zachmu (approved):
LGTM, see comment about test

go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
max-hoffman: It works without the hint if that makes you feel any better. I agree that more system tables and scenarios are subject to the same issue which we probably haven't fixed yet.

PR #7257, opened 2024-01-04
Update MySQL Connector/J integration tests to cover table-qualified field lookups

fulghum:
When using the MySQL Connector/J library to pull out a field value from a row, you can specify the column name, or you can specify the table-qualified column name (e.g. `myTable.col1`). https://github.com/dolthub/go-mysql-server/pull/2235 fixed the missing metadata, and this PR adds an integration test for the MySQL Connector/J library to ensure we don't regress. 

Related to https://github.com/dolthub/dolt/issues/7247


jennifersp (approved):
Looks good!

PR #7266, opened 2024-01-05
Bug fix: Race condition in `NewBranchAtCommit`

fulghum:
Changing `NewBranchAtCommit` to persist the branch HEAD and working set hashes as an atomic operation, instead of as two separate operations.

My only remaining concern is that we're taking in the `ReplicationStatusController` param and not using it. I've dug through the code paths that call this function and left some temporary notes in the code showing that nothing actually seems to use this parameter here. I can remove the param from `NewBranchAtCommit`, but wanted to get some feedback on this direction, first.


fulghum:
Closing this out until I get some time to get back into this one and dig into why `ReplicationStatusController` doesn't seem to ever be populated in these code paths.

PR #7278, opened 2024-01-09
default to text option for `dolt table import`

stephkyou:
Adds the `--all-text` option to `dolt table import` which will default all columns to `TEXT` type.

Resolves: https://github.com/dolthub/dolt/issues/6471


fulghum (approved):
Looks great! Nice work! 🚢

PR #7284, opened 2024-01-09
`dolt table import --all-text` correctly infers pk if none given

stephkyou:
`dolt table import` will use the first column in the import file as the primary key if none is explicitly defined. This change updates `--all-text` to also correctly do this inference.


go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -91,7 +91,7 @@ var importDocs = cli.CommandDocumentationContent{
...
 
-The schema for the new table can be specified explicitly by providing a SQL schema definition file, or will be inferred from the imported file.  All schemas, inferred or explicitly defined must define a primary key.  If the file format being imported does not support defining a primary key, then the {{.EmphasisLeft}}--pk{{.EmphasisRight}} parameter must supply the name of the field that should be used as the primary key.
+The schema for the new table can be specified explicitly by providing a SQL schema definition file, or will be inferred from the imported file.  All schemas, inferred or explicitly defined must define a primary key.  If the file format being imported does not support defining a primary key, then the {{.EmphasisLeft}}--pk{{.EmphasisRight}} parameter must supply the name of the field that should be used as the primary key. If not primary key is explicitly defined, the first column in the import file will be used as the primary key.
```
fulghum, line 94:
```suggestion
The schema for the new table can be specified explicitly by providing a SQL schema definition file, or will be inferred from the imported file.  All schemas, inferred or explicitly defined must define a primary key.  If the file format being imported does not support defining a primary key, then the {{.EmphasisLeft}}--pk{{.EmphasisRight}} parameter must supply the name of the field that should be used as the primary key. If no primary key is explicitly defined, the first column in the import file will be used as the primary key.
```

fulghum (approved):
Looks good! 🚢

PR #7285, opened 2024-01-10
Bug fix: keyless table constraint violations

fulghum:
Keyless tables didn't expose a unique row identifier in the `dolt_constraint_violations_<table>` system tables. This manifested as a panic when trying to delete the violation artifacts from the system table, but it also meant that we couldn't uniquely identify rows in the system table by the system table's PK for keyless tables. To enable keyless tables to use constraint violation artifacts, we now return the row's identifying hash from the source keyless table in the system table. Note that the `dolt_conflicts_<table>` system table did not have a similar issue because its primary key is a unique ID generated in the `dolt_conflict_id` field. 

Related to: https://github.com/dolthub/dolt/issues/7275

### DIFF
--- a/go/libraries/doltcore/doltdb/table.go
+++ b/go/libraries/doltcore/doltdb/table.go
@@ -34,6 +34,8 @@ import (
 
 var ErrNoConflictsResolved = errors.New("no conflicts resolved")
 
+const dolt_row_hash_tag = 0
+
 // IsValidTableName checks if name is a valid identifer, and doesn't end with space characters
 func IsValidTableName(name string) bool {
 	if len(name) == 0 || unicode.IsSpace(rune(name[len(name)-1])) {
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -66,6 +66,58 @@ func (dcv *doltCommitValidator) Validate(val interface{}) (bool, error) {
...
+				// Fix the data
+				Query:    "UPDATE aTable SET aColumn = 2 WHERE bColumn = 2;",
+				Expected: []sql.Row{{types.OkResult{RowsAffected: uint64(1), Info: plan.UpdateInfo{Matched: 1, Updated: 1}}}},
+			},
+			{
+				// clear out the violations
+				Query:    "DELETE FROM dolt_constraint_violations_aTable;",
+				Expected: []sql.Row{{types.NewOkResult(2)}},
+			},
+		},
```
zachmu, line 119: Merge and check result at the end?

go/libraries/doltcore/doltdb/table.go
```diff
@@ -354,18 +356,19 @@ func (t *Table) GetConstraintViolationsSchema(ctx context.Context) (schema.Schem
...
-		colColl = colColl.Append(sch.GetNonPKCols().GetColumns()...)
-		colColl = colColl.Append(infoCol)
+	// the commit hash or working set hash of the right side during merge
+	colColl = colColl.Append(schema.NewColumn("from_root_ish", 0, types.StringKind, false))
+	colColl = colColl.Append(typeCol)
+	if schema.IsKeyless(sch) {
+		// If this is a keyless table, we need to add a new column for the keyless table's generated row hash.
+		// We need to add this internal row hash value, in order to guarantee a unique primary key in the
+		// constraint violations table.
+		colColl = colColl.Append(schema.NewColumn("dolt_row_hash", dolt_row_hash_tag, types.BlobKind, true))
```
zachmu, line 366: What happens if there is more than one copy of a row in this keyless table? Is the idea that this couldn't happen because the only constraint violations possible would be on a unique key, so all the rows would need to be the same? What about if I add two copies of a row on one branch and then merge them to a branch that has an additional schema constraint they violate?

zachmu (approved):
My only real comment here is to make sure that this works when there is more than one copy of a given row with a violation, in which case you'll have two rows with the same hash in the violations table.

go/libraries/doltcore/doltdb/table.go
fulghum: I had briefly thought about this, but couldn't think of a case where we'd need to distinguish the row copies. I spent some more time thinking through it today, and I think this PR still works correctly for this case, although there's room to debate if that's the right workflow. It also turned up an issue with not-null constraints on keyless tables not being validated correctly in merges (https://github.com/dolthub/dolt/issues/7320). 

The possible constraint violations that could be reported are:
- not null
- unique 
- foreign key
- check constraint

For each of those, any duplicate rows that violate a constraint will violate it in the exact same way for each row. We *could* list the constraint violation multiple times in the `dolt_constraint_violations_<table>` system tables, but I'm not sure it's very helpful – it would require adding new column the primary key in `dolt_constraint_violations_<table>` for keyless tables to indicate some new ID/count for the row that violated the constraint (to keep it unique), and it seems like it adds more noise for customers to deal with when they are looking at constraint violations and potentially need to understand more about keyless table row hashes and row copies. 

With the behavior in this PR, customers can still correct one of the duplicate rows in a keyless table to fix the violation, they just can't/shouldn't clear the violation from the system table until they've fixed all the copies of that initial row. Having another ID/count field to unique identify each copy of the row could enable them to clear a violation for each individual row, but also makes it a little harder to clear the violations. I'm happy to explore that more if you think that's the right workflow though; seems like a good separate PR. 

I'll go ahead and add a test case that triggers duplicate rows in a keyless table that each violate a constraint, and we can keep discussing the right behavior if you think there does need to be one row per row copy in the constraint violations system table.

PR #7297, opened 2024-01-10
dolt push: Only assert on a clean working set when doing a push if we are running against a server which requests it.

reltuk:


macneale4 (approved):

macneale4:
2nd user to run into this: https://github.com/dolthub/dolthub-issues/issues/524

PR #7311, opened 2024-01-12
go/store/nbs: Flush a journal writer to fsync and write index records when it builds up substantial unflushed data, even if there is no commit yet.

reltuk:


zachmu (approved):
LGTM, no notes

PR #7373, opened 2024-01-19
add `dolt rebase` command

stephkyou:
Adds the `dolt rebase` command which is supported by the `dolt_rebase()` procedure.

Related: https://github.com/dolthub/ld/pull/16323


coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
<TR><TD>99.997743 to 99.997743</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c784959</TD>
<TD>not ok</TD>
<TD>134</TD>
</TR>
<TR><TD>c784959</TD>
<TD>ok</TD>
<TD>5937323</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c784959</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.997743</TD>
</TR>
</table>

coffeegoddd:
@stephkyou __DOLT__
 <table>
</TR>
<TR><TD>99.997743 to 99.997743</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a8f58bd</TD>
<TD>not ok</TD>
<TD>134</TD>
</TR>
<TR><TD>a8f58bd</TD>
<TD>ok</TD>
<TD>5937323</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a8f58bd</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.997743</TD>
</TR>
</table>

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+			status, err := getInt64ColAsInt64(rows[0][0])
+			if err != nil {
+				return HandleVErrAndExitCode(errhand.VerboseErrorFromError(err), usage)
+			}
+			if status == 1 {
+				return HandleVErrAndExitCode(errhand.VerboseErrorFromError(errors.New("error: "+rows[0][1].(string))), usage)
+			}
+
+			cli.Println("Interactive rebase aborted.")
```
fulghum, line 164: I wasn't immediately sure why `rebasePlan` or `Steps` would be nil. I'll dig into the `getRebasePlan` function next. It seems like whatever case this is, we may want to display a different error message explaining that something went wrong and the rebase was aborted (i.e. explain why the rebase was aborted).

It may even be worth a comment on line 151 to explain when this case happens (if it's not clear from an updated cli.Println message).

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+	if apr.Contains(cli.ContinueFlag) {
+		args = append(args, "'--continue'")
+	}
+	if apr.Contains(cli.AbortParam) {
+		args = append(args, "'--abort'")
+	}
+
+	/*if apr.ContainsAny(cli.ContinueFlag, cli.AbortParam) {
+		checkoutQuery = fmt.Sprintf("CALL DOLT_CHECKOUT('dolt_rebase_%s'); ", rebaseBranchName)
+	}*/
```
fulghum, line 213: Was this for the case you mentioned where when we abort a rebase, the CLI's active branch doesn't get moved back to the original branch?

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+
+	query := fmt.Sprintf("%sCALL DOLT_REBASE(%s);", checkoutQuery, strings.Join(args, ", "))
+	interpolatedQuery, err := dbr.InterpolateForDialect(query, params, dialect.MySQL)
+	if err != nil {
+		return "", err
+	}
+
+	return interpolatedQuery, nil
```
fulghum, line 221: (minor) you could simplify this code a little bit and get rid of those last five lines in the function if you just directly returned the results from `dbr.InterpolateForDialect`. It doesn't change the behavior materially, but it's nice to save five lines of code in this file!

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rebasePlan, err := parseRebaseMessage(rebaseMsg)
+	if err != nil {
+		return nil, err
+	}
+
+	return rebasePlan, nil
```
fulghum, line 260: (minor) Similar comment here. This could be simplified to just `return parseRebaseMessage(rebaseMsg)`

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+		return "", err
+	}
+	rows, err = GetRowsForSql(queryist, sqlCtx, "SELECT COUNT(*) FROM dolt_rebase")
+	if err != nil {
+		return "", err
+	}
+	numSteps, err := getInt64ColAsInt64(rows[0][0])
+	if err != nil {
+		return "", err
+	}
```
fulghum, line 301: I think you can avoid another round trip/query to the sql engine here... if you look at the len(rows) that should give you the count of the steps in the rebase plan.

go/cmd/dolt/commands/rebase.go
```diff
@@ -0,0 +1,369 @@
...
+		return dprocedures.RebaseActionEnumType.At(int(col.(uint16)))
+	}
+}
+
+// parseRebaseMessage parses the rebase message from the editor and adds all uncommented out lines as steps in the rebase plan.
+func parseRebaseMessage(rebaseMsg string) (*rebase.RebasePlan, error) {
+	plan := &rebase.RebasePlan{}
+	splitMsg := strings.Split(rebaseMsg, "\n")
+	for i, line := range splitMsg {
+		if !strings.HasPrefix(line, "#") && strings.ReplaceAll(line, " ", "") != "" {
```
fulghum, line 337: [strings.TrimSpace](https://pkg.go.dev/strings#TrimSpace) would probably be a better fit here. That will remove all leading and trailing whitespace (including tabs)

fulghum (commented):
I still need to dig through the tests and play with a local build a little bit, but wanted to send out a first batch of feedback in case you want to get started on any of these while I continue reviewing.

integration-tests/bats/helper/local-remote.bash
```diff
@@ -136,6 +136,7 @@ SKIP_SERVER_TESTS=$(cat <<-EOM
 ~cli-hosted.bats~
 ~profile.bats~
 ~ls.bats~
+~rebase.bats~
```
fulghum, line 139: What happens if we run the `rebase.bats` tests when there's a local sql-server running already? I see that we have a separate bats test file for local-remote mode, but I wanted to understand if there's some functionality that doesn't work in local-remote mode. Is it possible to get the same test file running for both local-remote and local modes?

integration-tests/bats/rebase.bats
```diff
@@ -0,0 +1,396 @@
...
+    [[ "$output" =~ "fatal: A branch named 'dolt_rebase_b1' already exists." ]] || false
+}
+
+@test "rebase: verify custom script" {
+    touch rebaseScript.sh
+    echo "#!/bin/bash" >> rebaseScript.sh
+    echo "mv rebasePlan.txt \$1" >> rebaseScript.sh
+    chmod +x rebaseScript.sh
+    export EDITOR=$PWD/rebaseScript.sh
+    export DOLT_TEST_FORCE_OPEN_EDITOR="1"
```
fulghum, line 98: You might consider putting some of this code into a helper function. It looks like it's used nine times in the file, so if you're able to reuse 5 of these 6 lines and not have to copy them a bunch of times in this file, I think it'd be a nice win and keep each test just a little bit simpler. Imaging how this file will change over time... we'll probably need to add more tests, so that helper function will probably help more in the future, too.

fulghum (commented):
A couple more quick comments... I'll hop back on this and finish up when I get back into the office in a bit.

integration-tests/bats/rebase.bats
```diff
@@ -0,0 +1,396 @@
...
+    run dolt rebase -i main
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Successfully rebased and updated refs/heads/b1" ]] || false
+
+    run dolt log
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "main commit 2" ]] || false
+}
+
+@test "rebase: invalid rebase plan" {
```
fulghum, line 135: When the rebase plan is invalid, I think we should clean up the rebase state, so that the user can rerun `dolt rebase` to try again. It looks like currently, the rebase branch is created and is left around, so users would have to delete that branch or switch over to the SQL shell to try and finish the rebase. It would be nice if they could just quickly try again with `dolt rebase`.

fulghum (commented):
I finished going through the tests and played with a local build a little bit. Things are looking good! Just a couple more small comments. Hit me up after you get a chance to go through the feedback and I'm happy to take one more quick look before we ship this one.

go/cmd/dolt/commands/rebase.go
stephkyou: `rebasePlan.Steps` could be nil if the user deletes all uncommented out lines from the editor, which will abort the rebase. `getRebasePlan` will add nothing to the `Steps` if that's the case. The `rebasePlan == nil` check is mostly a safety precaution for checking `rebasePlan.Steps`. I can definitely add a comment here to give context for this check.

go/cmd/dolt/commands/rebase.go
stephkyou: Oops, this was leftover from some testing I was doing, I'll remove since it shouldn't be needed.

integration-tests/bats/helper/local-remote.bash
stephkyou: The problem is that the rebase tests use `dolt checkout` for setup which won't work remotely :(

integration-tests/bats/rebase.bats
```diff
@@ -0,0 +1,396 @@
...
+    dolt checkout b1
+
+    touch invalidRebasePlan.txt
+    echo "foo" >> invalidRebasePlan.txt
+    run dolt rebase -i main
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "invalid line 0: foo" ]] || false
+
+    dolt branch -D dolt_rebase_b1
+    > invalidRebasePlan.txt
```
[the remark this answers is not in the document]
stephkyou: This should be a `touch`, the intention was to get an empty file.

integration-tests/bats/helper/local-remote.bash
fulghum: Ahh... yes... I think that was the same issue last time I asked about this list, too. Thanks for gently reminding me 😉 

Bummer that we keep hitting that snag. I can't think of a great way to avoid having to use `dolt checkout` for the tests. The only idea that comes to mind is using `main` as the branch to rebase, since that happens to already be the default branch for CLI-mode. You could temporarily switch to another branch and make commits if you sent a batch of SQL statements to `dolt sql -q`, but I don't think it's worth it to rewrite the tests like that.

coffeegoddd:
@stephkyou __DOLT__
 <table>
</TR>
<TR><TD>99.998097 to 99.997743</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b3d5705</TD>
<TD>not ok</TD>
<TD>134</TD>
</TR>
<TR><TD>b3d5705</TD>
<TD>ok</TD>
<TD>5937323</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b3d5705</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.997743</TD>
</TR>
</table>

fulghum (approved):
Nice work on this one! Looks great! 🚢

coffeegoddd:
@stephkyou __DOLT__
 <table>
</TR>
<TR><TD>99.998097 to 99.998097</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>89c8e80</TD>
<TD>not ok</TD>
<TD>113</TD>
</TR>
<TR><TD>89c8e80</TD>
<TD>ok</TD>
<TD>5937344</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>89c8e80</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.998097</TD>
</TR>
</table>

PR #7389, opened 2024-01-22
[no-release-notes] integration-tests/bats: Ensure we do not run `dolt` from a parent directory of the remotesrv dir.

reltuk:
Doing so can cause `dolt` to open the remotesrv database as a DoltDB, which may race with remotesrv opening it and may cause it to appear to read-only in the remotesrv process.


macneale4 (approved):
LGTM

PR #7412, opened 2024-01-25
When aborting a merge/cherry-pick, avoid resetting ignored tables.

nicktobey:
Tables that match `dolt_ignore` in the working set should be ignored by most operations.

(CAVEAT: If we want to match git, these tables should only be ignored if there are not staged/HEAD versions of those tables. But we appear to not match git in that regard right now.)

This PR changes the logic for aborting merges to avoid modifying any tables match `dolt_ignore`

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -5669,10 +5669,14 @@ var DoltCherryPickTests = []queries.ScriptTest{
 		SetUpScript: []string{
 			"INSERT INTO dolt_ignore VALUES ('generated_*', 1);",
 			"CREATE TABLE generated_foo (pk int PRIMARY KEY);",
+			"CREATE TABLE generated_bar (pk int PRIMARY KEY);",
+			"insert into generated_foo values (1);",
+			"insert into generated_bar values (1);",
 			"SET @@autocommit=1;",
 			"SET @@dolt_allow_commit_conflicts=1;",
 			"create table t (pk int primary key, v varchar(100));",
 			"insert into t values (1, 'one');",
+			"call dolt_add('--force', 'generated_bar');",
 			"call dolt_commit('-Am', 'create table t');",
 			"call dolt_checkout('-b', 'branch1');",
 			"update t set v=\"uno\" where pk=1;",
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
<TR><TD>99.998653 to 99.998653</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>826d9fc</TD>
<TD>not ok</TD>
<TD>80</TD>
</TR>
<TR><TD>826d9fc</TD>
<TD>ok</TD>
<TD>5937377</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>826d9fc</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.998653</TD>
</TR>
</table>

nicktobey:
It appears that prior to this PR, we were leaving all tables staged after an aborted merge, except for ignored tables. This seems incorrect: we should not be leaving any tables staged after aborting a merge, given that you shouldn't be able to start a merge while there are staged tables, and aborting the merge should return the user to the state they were in before the merge began.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
<TR><TD>99.998653 to 99.998653</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>1f12466</TD>
<TD>not ok</TD>
<TD>80</TD>
</TR>
<TR><TD>1f12466</TD>
<TD>ok</TD>
<TD>5937377</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>1f12466</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.998653</TD>
</TR>
</table>

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -5709,9 +5723,22 @@ var DoltCherryPickTests = []queries.ScriptTest{
 				Expected: []sql.Row{{1, "one"}},
 			},
 			{
+				// An ignored table should still be present (and unstaged) after aborting the merge.
 				Query:    "select * from dolt_status;",
-				Expected: []sql.Row{},
+				Expected: []sql.Row{{"generated_foo", false, "new table"}},
```
zachmu, line 5728: Maybe beyond the scope of this bug fix, but I would expect that an ignored table doesn't appear in `dolt_status`, similar to how ignored files don't appear in `git status`.

zachmu (approved):
LG, but see comment about behavior of dolt_status

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
nicktobey: I think when `dolt_ignore` was being added, the decision was to add a `ignored` column to the status table.

Either way, I agree that something should be done, and also agree that this is beyond the scope of the bug fix.

PR #7490, opened 2024-02-14
Bug fix: `IndexedDoltTable` didn't fully implement `DoltTable`'s interface

fulghum:
`IndexedDoltTable` wasn't fully compatible with `DoltTable`'s interface, which was causing some queries to execute incorrectly due to using the wrong field indexes. This only happened with queries where the table was used in a read-only context (such as an `AS OF` query), since in a read-write context `WritableIndexedDoltTable` is used, which is composed of a `WritableDoltTable`, so fully inherits the interface. `IndexedDoltTable` is now consistent with that, and is now composed of a `DoltTable` instance. This enables the GMS code in `assignExecIndexes` to use the correct, limited schema of the secondary index and to apply the correct field indexes. 

Fixes: https://github.com/dolthub/dolt/issues/7488

### DIFF
--- a/go/cmd/dolt/commands/engine/sqlengine.go
+++ b/go/cmd/dolt/commands/engine/sqlengine.go
@@ -115,7 +115,7 @@ func NewSqlEngine(
 		return nil, err
 	}
 
-	all := append(dbs)
+	all := dbs[:]
 
 	clusterDB := config.ClusterController.ClusterDatabase()
 	if clusterDB != nil {
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries.go
@@ -862,6 +862,51 @@ var DoltScripts = []queries.ScriptTest{
 		},
 	},
 	{
+		Name: "test AS OF indexed queries (https://github.com/dolthub/dolt/issues/7488)",
+		SetUpScript: []string{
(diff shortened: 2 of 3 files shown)


coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
<TR><TD>99.999562 to 99.999562</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>fd8fcd5</TD>
<TD>not ok</TD>
<TD>26</TD>
</TR>
<TR><TD>fd8fcd5</TD>
<TD>ok</TD>
<TD>5937431</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>fd8fcd5</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>99.999562</TD>
</TR>
</table>

max-hoffman (approved):
LGTM, the source of this bug here is that table pruning does not record metadata on the GMS side. We currently expect table sources to be able to tell us both (1) all of the logical columns in the table, and (2) what columns will actually be returned (projected) at execution time. Ideally pruning columns would shift some of this burden for execution indexing into the engine. `PrimaryKeySchema` is maybe a misleading name for the way this interface is used here. We could have something like `LogicalSchema` as a subinterface separate from the `PkOrdinals()` method. As long as there are other places where we depend on `PrimaryKeySchema` for the logical/execution schema difference, having table sources implement the interface for that info will be good.

PR #7516, opened 2024-02-18
Less docker layers and more compact Docker image

sonjek:
Install both packages on one Docker layer and delete package cache after installation.
Docker image size decreased from 259 MB to 210 MB.


fulghum (approved):
Thanks for the contribution @sonjek! 🙏 This looks good!

PR #7617, opened 2024-03-14
go/libraries/doltcore/merge: fulltext_rebuild.go: Add some comments, refactor some things into functions, even if they are still too big.

reltuk:


Hydrocharged (approved):
Refactor makes a ton of sense!

PR #7644, opened 2024-03-25
go/store/datas/pull: pull_chunk_fetcher: Move chunk fetching to a streaming interface instead of batch.

reltuk:
We want to better pipeline I/O when pulling from a remote, and moving a streaming interface where storage can see more of the addresses that are needed at once will allow us to achieve it.

For now, we implement the streaming interface just by calling the existing batch get interface.

### DIFF
--- a/go/store/datas/pull/pull_chunk_fetcher.go
+++ b/go/store/datas/pull/pull_chunk_fetcher.go
@@ -1,0 +1,174 @@
+// Copyright 2024 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pull
+
+import (
+	"context"
+	"io"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/dolthub/dolt/go/store/hash"
+	"github.com/dolthub/dolt/go/store/nbs"
+)
+
+type GetManyer interface {
+	GetManyCompressed(ctx context.Context, hashes hash.HashSet, found func(context.Context, nbs.CompressedChunk)) error
+}
+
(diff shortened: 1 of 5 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b05d3d9</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b05d3d9</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/datas/pull/pull_chunk_fetcher.go
```diff
@@ -0,0 +1,174 @@
...
+
+			var mu sync.Mutex
+			missing := batch.Copy()
+
+			// Blocking get, no concurrency, only one fetcher.
+			err := f.getter.GetManyCompressed(f.ctx, batch, func(ctx context.Context, chk nbs.CompressedChunk) {
+				mu.Lock()
+				missing.Remove(chk.H)
+				mu.Unlock()
+				select {
```
macneale4, line 113: This select block is a little odd to me. Why aren't you just sticking chk into f.resCh and calling it a day? These are unlimited channels, so they aren't going to block, right? Does this avoid getting a panic when the channel has been closed prematurely?

go/store/datas/pull/puller.go
```diff
@@ -296,139 +300,82 @@ func (p *Puller) Pull(ctx context.Context) error {
...
+				// thread. Calling wr.Close() here will block
+				// on uploading any table files and will write
+				// the new table files to the destination's
+				// manifest.
+				return p.wr.Close()
+			}
+			if err != nil {
+				return err
+			}
+			if len(cChk.FullCompressedChunk) == 0 {
```
macneale4, line 346: The only way this would happen would be if a CompressedChunk{} was returned, which should always be accompanied by an error. Is this out of an abundance of caution, or is there another case where this will come up?

macneale4 (approved):
I had a couple questions, but free to ship

go/store/datas/pull/pull_chunk_fetcher.go
reltuk: The channels are blocking and unbuffered...it's a synchronous exchange. So here we don't have a way to bail early but the reader might have already shut down, so we need to get through the batch of responses by ignoring them.

PR #7679, opened 2024-04-02
`dolt_verify_constraints()` support for all constraint types

fulghum:
The `dolt_verify_constraints()` stored procedure now supports verifying all constraint types: foreign keys, not null, unique, and check constraints. This new implementation reuses merge logic to calculate the constraint violations. 

Note that the `dolt constraints verify` CLI command next needs to be migrated to use this stored procedure so that it can also verify all constraint violation types. Until then, customers can use `dolt sql -q "CALL dolt_verify_constraints(...)"` if they want to verify all constraint types from the CLI. 

Doc updates: https://github.com/dolthub/docs/pull/2093

Fixes: https://github.com/dolthub/dolt/issues/6320


coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>68e8033</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>68e8033</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/merge/merge_rows.go
```diff
@@ -158,43 +179,73 @@ func (rm *RootMerger) MergeTable(ctx *sql.Context, tblName string, opts editor.O
...
 		}
 	}
 
-	tm.ancTbl, ok, err = rm.anc.GetTable(ctx, tblName)
+	// If we need to re-verify all constraints, then we need to stub out tables
+	// that don't exist, so that the diff logic can compare an empty table to
+	// the table containing the real data. This is required by dolt_verify_constraints()
+	// so that we can run the merge logic on all rows in all tables.
+	if mergeOpts.ReverifyAllConstraints {
+		if !leftSideTableExists && rightSideTableExists {
```
zachmu, line 227: This is pretty convoluted and mixes a bunch of different concerns together. My primary objection is that we're altering the `tm` object in place here, which changes the semantics of what comes next for an unrelated concern. Seems like it would be better to special case just this logic and do a separate pass than try to cram this use case into the same merge code path. Or if that's not easy, then call the merge code with a copy of the table merger with these alterations.

go/libraries/doltcore/sqle/dprocedures/dolt_verify_constraints.go
```diff
@@ -75,37 +77,110 @@ func doDoltConstraintsVerify(ctx *sql.Context, args []string) (int, error) {
...
+		}
+	}
+
 	if tablesWithViolations.Size() == 0 {
 		// no violations were found
 		return 0, nil
 	}
 
-	// violations were found
+	// TODO: We only return 1 or 0 to indicate if there were any constraint violations or not. This isn't
```
zachmu, line 103: Seems like a fast-follow

go/libraries/doltcore/sqle/dprocedures/dolt_verify_constraints.go
```diff
@@ -75,37 +77,110 @@ func doDoltConstraintsVerify(ctx *sql.Context, args []string) (int, error) {
...
+	return 1, nil
+}
 
-	if !outputOnly {
-		err = dSess.SetRoot(ctx, dbName, newRoot)
+// calculateViolations calculates all constraint violations between |workingRoot| and |comparingRoot| for the
+// tables in |tableSet|. Returns the new root with the violations, and a set of table names that have violations.
+// Note that constraint violations detected for ALL existing tables will be stored in the dolt_constraint_violations
+// tables, but the returned set of table names will be a subset of |tableSet|.
+func calculateViolations(ctx *sql.Context, workingRoot, comparingRoot *doltdb.RootValue, tableSet *set.StrSet) (*doltdb.RootValue, *set.StrSet, error) {
```
zachmu, line 114: This just seems like an odd way to structure this process. Naively I would expect that constraint violation checks are their own library methods that are used by both the merge code as necessary, as well as here (where we do a full table scan).

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
```diff
@@ -4045,7 +4045,34 @@ var OldFormatMergeConflictsAndCVsScripts = []queries.ScriptTest{
 	},
 }
 
-var verifyConstraintsSetupScript = []string{
+var verifyConstraintsUniqueViolationsSetupScript = []string{
```
zachmu, line 4048: Maybe these belong in their own file?

zachmu (approved):
I think this is directionally correct and fixes the immediate problem, so feel free to merge.

However, I don't think this refactoring goes far enough, or in the right direction. Conflating the concerns of constraint verification with merging is confusing and is going to lead to more bugs. Verification is a logically separate operation that deserves its own library package.

go/libraries/doltcore/sqle/enginetest/dolt_queries_merge.go
fulghum: totally agree. I was going to move them after review – I just figured the diff would be easier to read before that. I should'a mentioned that in the PR description.

go/libraries/doltcore/merge/merge_rows.go
max-hoffman: calling merge code with an edited copy sounds like a good near term idea

max-hoffman: maybe delete the tables whose constraints we don't want to verify? you wouldn't need `recordViolations` in that case also

max-hoffman: and then maybe `ReverifyAllConstraints` -> `shortCircuit = false`?

max-hoffman (approved):
I understand why you used the merger for this...otherwise you'd just be copying most of it and then deleting the subset related to primary and secondary index editing. I like zach's suggestion for altering the table inputs and then calling merge in mostly it's pre-existing format, writing the artifacts back from the merge result.

One thing andy felt strongly about was that violations shouldn't be materialized ever...I'm looking into how violations affect transaction throughput, but it also might affect the UX semantics in a way that may or may not be desirable if you have thoughts on that.

go/libraries/doltcore/sqle/dprocedures/dolt_verify_constraints.go
fulghum: We don't always do a full table scan though. The default mode of `dolt_verify_constraints()` only checks changes in the working set, so at some point, we have to support diffing HEAD with the working set and only verifying those changes.

go/libraries/doltcore/sqle/dprocedures/dolt_verify_constraints.go
fulghum: Yeah, this is an easy one to change, but before changing the behavior, I think we should look at migrating `dolt constraints verify` over to use the stored procedure – I suspect there will be other change to the result data that we'll need to make to support that.

go/libraries/doltcore/merge/merge_rows.go
fulghum: I think it's valid to adjust `tm` in this function, since this is the constructor function for the `tm` instance and once created, this instance is stable for it's lifetime and doesn't get altered. I agree that the concerns are mixed together a bit though. 

Regarding deleting the tables with constraints we don't want to modify... I tried something similar first (skipping over the tables completely), but because the root gets written back in order to store the violation artifacts, this led to other problems in the new root. It did seem to me like this was going to work though, so maybe I just missed something in that attempt.

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8f7f74e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8f7f74e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #7699, opened 2024-04-05
Add per-table locking for AutoIncrementTracker

nicktobey:
This PR refactors the AutoIncrementTracker to hold a separate mutex for each table instead of a single mutex for the entire database.

# Why are we doing this?

This relates to supporting the `innodb_autoinc_lock_mode` variable, described here: https://github.com/dolthub/dolt/issues/7634

Currently, the engine acquires the AutoIncrementTracker lock once per inserted row and holds it only while that row is being inserted. However, when `innodb_autoinc_lock_mode` is set to any value other than its default, the engine must instead hold the lock for the duration of the insert statement.

Unfortunately, there is a specific corner case that causes `innodb_autoinc_lock_mode` to interact poorly with a single global lock: if a table has a trigger, and that trigger attempts to insert into a different table, then a second insert statement begins executing before the first insert statement completes. In this case, the engine may attempt to acquire the global lock while it's already being held.

There are four possible ways to fix this:
1 - Use reentrant/recursive locks
2 - Analyze the entire execution plan to determine whether it will require the lock, and then acquire the lock at most once.
3 - Add additional logic to determine whether a session already holds the lock and avoid reacquiring it
4 - Use per-table locks

Option 1 is controversial: golang does not support recursive mutexes by design, and their use is discouraged by many. Option 2 will hold the lock for longer than is required, which would negatively impact performance. Option 3 is error prone and may result in deadlocks if implemented incorrectly.

Thus, option 4 is the best option.

# What is the impact of this PR?

Concurrent inserts into two different tables with auto-increment columns should be faster because of reduced lock contention, at the cost of the additional overhead of managing multiple mutexes when the inserts are not concurrent. The magnitude of this impact is difficult to predict or measure, but is likely overall positive.

# Why are there no tests?

This change should have no observable effect on query results. Our existing integration test suite is comprehensive enough that if no existing integration tests break, this should be safe.

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/autoincrement_tracker.go
+++ b/go/libraries/doltcore/sqle/dsess/autoincrement_tracker.go
@@ -16,13 +16,13 @@ package dsess
 
 import (
 	"context"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dsess/mutexmap"
+	"github.com/dolthub/go-mysql-server/sql"
+	gmstypes "github.com/dolthub/go-mysql-server/sql/types"
 	"io"
 	"math"
 	"strings"
 	"sync"
-
-	"github.com/dolthub/go-mysql-server/sql"
-	gmstypes "github.com/dolthub/go-mysql-server/sql/types"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
@@ -35,8 +35,8 @@ import (
 
 type AutoIncrementTracker struct {
 	dbName    string
-	sequences map[string]uint64
-	mu        *sync.Mutex
+	sequences *sync.Map // map[string]uint64
+	mm        *mutexmap.MutexMap
 }
 
 var _ globalstate.AutoIncrementTracker = AutoIncrementTracker{}
@@ -48,8 +48,8 @@ var _ globalstate.AutoIncrementTracker = AutoIncrementTracker{}
 func NewAutoIncrementTracker(ctx context.Context, dbName string, roots ...doltdb.Rootish) (AutoIncrementTracker, error) {
 	ait := AutoIncrementTracker{
 		dbName:    dbName,
-		sequences: make(map[string]uint64),
-		mu:        &sync.Mutex{},
+		sequences: &sync.Map{},
+		mm:        mutexmap.NewMutexMap(),
 	}
 
 	for _, root := range roots {
@@ -71,8 +71,9 @@ func NewAutoIncrementTracker(ctx context.Context, dbName string, roots ...doltdb
 				return true, err
 			}
 
-			if seq > ait.sequences[tableName] {
-				ait.sequences[tableName] = seq
+			oldValue, loaded := ait.sequences.LoadOrStore(tableName, seq)
+			if loaded && seq > oldValue.(uint64) {
+				ait.sequences.Store(tableName, seq)
 			}
 
 			return false, nil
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>fc32126</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>fc32126</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/dsess/mutexmap/mutexmap.go
```diff
@@ -0,0 +1,82 @@
+// Source: https://stackoverflow.com/questions/40931373/how-to-gc-a-map-of-mutexes-in-go
```
zachmu, line 1: Probably just want to use Dolthub copyright here

Or at least include the Doltub copyright notice as well, before other attributions

go/libraries/doltcore/sqle/dsess/autoincrement_tracker.go
```diff
@@ -82,41 +83,47 @@ func NewAutoIncrementTracker(ctx context.Context, dbName string, roots ...doltdb
...
+	}
+	return current.(uint64)
+}
+
 // Current returns the next value to be generated in the auto increment sequence for the table named
 func (a AutoIncrementTracker) Current(tableName string) uint64 {
-	a.mu.Lock()
-	defer a.mu.Unlock()
-	return a.sequences[strings.ToLower(tableName)]
+	tableName = strings.ToLower(tableName)
```
zachmu, line 100: Duplicate lowering here

zachmu (approved):
LGTM, elegant solution

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>262a397</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>262a397</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>45493be</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>45493be</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>fef14f3</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>fef14f3</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/dsess/mutexmap/mutexmap.go
nicktobey: Replaced with my own implementation.

PR #7704, opened 2024-04-05
Implement `traditional` auto-increment lock mode hold the lock for the duration of the insert iter.

nicktobey:
Fixes https://github.com/dolthub/dolt/issues/7634

This is the dolt half of https://github.com/dolthub/go-mysql-server/pull/2439

This adds support for `innodb_autoinc_lock_mode=0` ("traditional"). When this system variable is set, the engine will guarantee that every insert statement generates consecutive IDs for `AUTO INCREMENT` columns.

This PR also allows the user to set `innodb_autoinc_lock_mode=1` ("consecutive"), although the behavior is currently identical to "traditional". This is acceptable because both modes make the same guarantees (that each statement gets consecutive IDs), and the observed behavior is the same in almost all cases.

(The "consecutive" contains an additional optimization: if there is a known upper bound on the number of IDs that must be generated for an insert, under "consecutive" mode the engine will just increment the counter by that upper bound and immediately release the lock. In places where not all of those IDs are actually used, the excess are wasted. This PR does not include that optimization. Thus, with this PR, `traditional` and `consecutive` behave the same.)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>fe4a219</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>fe4a219</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>ab54c2c</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>ab54c2c</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/globalstate/auto_increment_tracker.go
```diff
@@ -38,4 +38,6 @@ type AutoIncrementTracker interface {
 	// below the current value for this table. The table in the provided working set is assumed to already have the value
 	// given, so the new global maximum is computed without regard for its value in that working set.
 	Set(ctx *sql.Context, tableName string, table *doltdb.Table, ws ref.WorkingSetRef, newAutoIncVal uint64) (*doltdb.Table, error)
+
+	AcquireTableLock(ctx *sql.Context, tableName string) (func(), error)
```
fulghum, line 42: Give this function some godocs. The name is pretty self explanatory, but generally if something is important enough to be in an interface it merits having godoc. Could be helpful for future code readers to explain the `func()` return param and any rules for using it correctly, for example.

integration-tests/bats/auto_increment_lock_modes.bats
```diff
@@ -0,0 +1,107 @@
...
+      cat > config.yml <<EOF
+system_variables:
+  innodb_autoinc_lock_mode: 0
+EOF
+    start_sql_server_with_config "" config.yml
+    # We use run here so that we don't terminate if the command fails, since terminating without stopping the server will cause the test to hang.
+    # We also use a small sequence table because larger tables increase the risk of triggering https://github.com/dolthub/dolt/issues/7702
+    # If we detect interleaved writes to `timestamps` with a smaller sequence table, then we would see it on the larger table anyway.
+    run dolt sql -q "INSERT INTO test1 (c0) select v from sequence5bit; SELECT * from timestamps; COMMIT;" &
+    run dolt sql -q "INSERT INTO test2 (c0) select v from sequence5bit; SELECT * from timestamps; COMMIT;"
```
fulghum, line 97: (minor) Doesn't seem like the `COMMIT`s are needed here, right? Since `@@autocommit` is enabled? Not a big deal, but might be nice remove them just to simplify and not confuse future code readers into thinking there's manual tx management going on. 

Is it interesting to test behavior for concurrent transactions? If so, that's another good reason to use the sql server multi session tests in the Go code.

integration-tests/bats/auto_increment_lock_modes.bats
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bats
+load $BATS_TEST_DIRNAME/helper/common.bash
```
fulghum, line 2: Did you get a chance to check out the sql server multi session tests in Dolt's Go code? Those tests are somewhat easier to debug, although since they start up a separate sql server process, it's still not as direct as a non-server test. They also have the advantage of having multiple concurrent sessions running and being able to interleave commands from different sessions. They are a little hacky, but improving them so we could use them more would be pretty valuable.

integration-tests/bats/auto_increment_lock_modes.bats
```diff
@@ -0,0 +1,107 @@
...
+    dolt sql -q "SET @@autocommit=1; INSERT INTO test1 (c0) select 1 from sequence10bit; INSERT INTO ranges VALUES (1, LAST_INSERT_ID(), ROW_COUNT());"
+    wait $!
+
+    stop_sql_server
+    dolt sql -r csv -q "select
+          c0,
+          min(pk) = firstId,
+          rowCount = 1024,
+          max(pk) = firstId + rowCount -1
+        from test1 join ranges on test1.c0 = ranges.pk group by c0"
```
fulghum, line 74: Looks like this is just an extra copy/paste?

fulghum (approved):
Looks good, a few small comments. I skipped over the changes included from https://github.com/dolthub/dolt/pull/7699 since those have already been reviewed and you're working on some follow ups there. 

My biggest comment is that I think the test code would be easier to understand and easier to extend if they used the sql-server golang multi session test framework we have in Dolt already. Testing this in BATS feels more complicated to me. 

We have a few tests that use the sql-server multi session test framework in Dolt already (e.g. for testing multi-session behavior with branch deletion). There are definitely some gaps in that part of our testing framework, but it would be really valuable to improve it so we could use it more easily for more functionality like this in the future.

integration-tests/bats/auto_increment_lock_modes.bats
nicktobey: Autocommit is actually disabled when the server is started with the `start_sql_server_with_config` helper function, so these commits are necessary.

nicktobey:
In regards to your comments about the server multi session tests: I took a look at them, but (if I understood them correctly) their limitation is that while they allow multiple sessions, the test harness is still only executing one statement at a time. For these tests, I need to test multiple sessions that are attempting to execute a statement concurrently.

I'm sure we could improve them to allow for this kind of concurrency, but in the meantime it was easier to just implement the tests this way.

integration-tests/bats/auto_increment_lock_modes.bats
nicktobey: Fixed.

go/libraries/doltcore/sqle/globalstate/auto_increment_tracker.go
nicktobey: Fixed.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e311f89</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e311f89</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6513fd2</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6513fd2</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e4b3856</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e4b3856</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

fulghum:
> In regards to your comments about the server multi session tests: I took a look at them, but (if I understood them correctly) their limitation is that while they allow multiple sessions, the test harness is still only executing one statement at a time. For these tests, I need to test multiple sessions that are attempting to execute a statement concurrently.

I don't think the BATS test gives you any better guarantee about the concurrency of the two sessions here – you're spinning up two processes through the shell, so there's no guarantee that the sessions will even be concurrent depending on how the OS ends up scheduling those processes. If you really need two cores to run the same statement at the same time to test this logic, I don't think this BATS test is the way to get that.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5ba740f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5ba740f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -195,6 +197,125 @@ func newUpdateResult(matched, updated int) gmstypes.OkResult {
...
+
+	// Alternate the iterators until they're exhausted.
+	var err1 error
+	var err2 error
+	for err1 != io.EOF || err2 != io.EOF {
+		if err1 != io.EOF {
+			var row1 sql.Row
+			require.NoError(t, err1)
+			row1, err1 = iter1.Next(ctx)
+			_ = row1
```
fulghum, line 278: (minor) Could you get rid of the `var row1` completely by saying: `_, err1 = iter1.Next(ctx)` like you're doing for `iter2`?

go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
```diff
@@ -195,6 +197,125 @@ func newUpdateResult(matched, updated int) gmstypes.OkResult {
...
+			require.NoError(t, err2)
+			_, err2 = iter2.Next(ctx)
+		}
+	}
+	err = iter1.Close(ctx)
+	require.NoError(t, err)
+	err = iter2.Close(ctx)
+	require.NoError(t, err)
+
+	dsess.DSessFromSess(ctx.Session).CommitTransaction(ctx, ctx.GetTransaction())
```
fulghum, line 290: Do we want these to be executed in the same session?

nicktobey: In a real environment they wouldn't be, but I don't think it matters here. The engine tests have a single session baked in, and figuring out how to disentangle that didn't seem worth it for a single test.

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>0f5cc28</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>0f5cc28</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>19c89b3</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>19c89b3</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #7706, opened 2024-04-06
[store] use struct{} as a value in a hashset

arturmelanchyk:


macneale4:
Thanks for the contribution @arturmelanchyk ! We are working through a CI issue in order to run the tests, but I don't expect any changes to be required. I'll merge when that gets sorted out.

macneale4 (approved):

PR #7776, opened 2024-04-23
`dolt_history` tables: Issue a warning when types have changed 

fulghum:
Adding a session warning when `dolt_history` tables can't create a `rowConverter` for a row due to a field's type having changed. In these cases, the old value can't always be returned safely as the new type, so we return `NULL` and issue a session warning. In the future, we could attempt to convert the old value to the new type. 

Follow up to https://github.com/dolthub/dolt/pull/7771


go/libraries/doltcore/sqle/history_table.go
```diff
@@ -596,6 +600,9 @@ func rowConverter(srcSchema, targetSchema sql.Schema, h hash.Hash, meta *datas.C
 			// TODO: we could do a projection to convert between types in some cases
 			if srcSchema[srcIdx].Type.Equals(targetSchema[i].Type) {
 				srcToTarget[srcIdx] = i
+			} else {
+				ctx.Warn(1246, "Unable to convert field %s in historical rows because its type (%s) doesn't match "+
```
zachmu, line 604: Probably want to limit this warning to once per table / statement?

zachmu (approved):
LGTM

go/libraries/doltcore/sqle/history_table.go
fulghum: Good call – I change this to only emit one warning per affected column.

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8feee63</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8feee63</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #7818, opened 2024-05-02
Apply a factor to better estimate `information_schema.TABLES.DATA_LENGTH`

fulghum:
`information_schema.TABLES.DATA_LENGTH` currently reports the max possible table size for a table, and doesn't take into account table file compression or that variable length fields (e.g. TEXT) are not always fully used. Tools such as DBeaver use this metadata to display table sizes, and since the estimates can easily be orders of magnitude greater than the actual size on disk, it can cause customers to be concerned by the reported sizes (e.g. https://github.com/dolthub/dolt/issues/6624).

As a short-term fix to make these estimates more accurate, we apply a constant factor to the max table size. I came up with this scaling factor by measuring a best case scenario (where no fields are variable length) and a worst case scenario (were all fields are variable length and only use a few bytes), then picking a value roughly in the middle. Longer-term, a better way to estimate table size on disk will be to use statistics data.

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/statistics_table.go
+++ b/go/libraries/doltcore/sqle/dtables/statistics_table.go
@@ -52,7 +52,24 @@ func (st *StatisticsTable) DataLength(ctx *sql.Context) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	return numBytesPerRow * numRows, nil
+
+	// maxSize is the upper bound for how much space a table takes up on disk. It will typically
+	// greatly overestimate the actual size of the table on disk because it does not take into
+	// account that the data on disk is compressed and it assumes that every variable length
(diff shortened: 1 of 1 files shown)


macneale4 (approved):
LGTM

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>78150fb</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>78150fb</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #7947, opened 2024-06-04
Bug Fix: Index name case-insensitivity

fulghum:
MySQL index names are case-insensitive, but Dolt's index implementation wasn't handling them that way. This makes index names case-insensitive.

Customer issue: https://github.com/dolthub/dolt/issues/7945

New enginetests added in GMS PR: https://github.com/dolthub/go-mysql-server/pull/2530


coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>bb1f705</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>bb1f705</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

zachmu (commented):
I don't think this is right... this only works if we store the index names lower case, which we don't (we preserve the original case). And any existing customers with indexes that aren't lower-case named will be broken by this, their index lookups won't work anymore.

The right approach for case sensitivity is to do the same thing we do with table name resolution: enumerate all the names, then find a case-insensitive match from among them.

And in any case, I'm not sure this is the right layer of the stack to apply this fix. E.g. in doltdb, we first get the case-sensitive table name match, then ask the root for a table with that case-sensitive name.

go/go.sum
```diff
@@ -183,8 +183,12 @@ github.com/dolthub/fslock v0.0.3 h1:iLMpUIvJKMKm92+N1fmHVdxJP5NdyDK5bK7z7Ba2s2U=
 github.com/dolthub/fslock v0.0.3/go.mod h1:QWql+P17oAAMLnL4HGB5tiovtDuAjdDTPbuqx7bYfa0=
 github.com/dolthub/go-icu-regex v0.0.0-20230524105445-af7e7991c97e h1:kPsT4a47cw1+y/N5SSCkma7FhAPw7KeGmD6c9PBZW9Y=
 github.com/dolthub/go-icu-regex v0.0.0-20230524105445-af7e7991c97e/go.mod h1:KPUcpx070QOfJK1gNe0zx4pA5sicIK1GMikIGLKC168=
-github.com/dolthub/go-mysql-server v0.18.2-0.20240604161217-d1dca79a32b8 h1:F9tnktZhnglXHYiG/tjnVMiLQSBZ8iJiflqgHg+ldKo=
-github.com/dolthub/go-mysql-server v0.18.2-0.20240604161217-d1dca79a32b8/go.mod h1:GT7JcQavIf7bAO17/odujkgHM/N0t4b1HfAPBJ2jzXo=
+github.com/dolthub/go-mysql-server v0.18.2-0.20240603181224-60bbc0c2423d h1:2L1X++TRoNWSHmlNOxjGGaZzquBN2AIEGAkCjJgQWcM=
```
zachmu, line 186: Might wanna go mod tidy here

go/libraries/doltcore/doltdb/durable/index.go
```diff
@@ -584,3 +588,22 @@ func (is doltDevIndexSet) RenameIndex(ctx context.Context, oldName, newName stri
 
 	return doltDevIndexSet{is.vrw, is.ns, am}, nil
 }
+
+// searchForCaseInsensitiveIndexName searches through the index names in this index set looking for a case-insensitive
+// match against |targetName|. If found, the address is returned, along with the exact case name. If no match was
+// found, a nil address is returned, along with an empty string.
+func (is doltDevIndexSet) searchForCaseInsensitiveIndexName(ctx context.Context, targetName string) (foundAddr *hash.Hash, foundName string, err error) {
```
zachmu, line 595: Think we usually return the zero-hash instead of nil for this kind of thing

zachmu: Check with hash.IsEmpty()

zachmu (approved):
LGTM

go/libraries/doltcore/doltdb/durable/index.go
fulghum: yeah... I have a thing against zero-objects – `nil` just seems so much more obvious to me that the value is invalid. I agree that it's better to be consistent though, so I'll change to `hash.IsEmpty()` instead of a nil check.

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e642131</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e642131</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6bb62cd</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6bb62cd</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #7977, opened 2024-06-07
integration-tests/bats: Add some waits on the remotesrv_pid exit for bats tests which spawn a background remotesrv.

reltuk:
Attempts to fix some observed flakiness in bats tests.


macneale4 (approved):
LGTM

PR #8081, opened 2024-06-27
Fixed keyless secondary indexing for Doltgres

Hydrocharged:
* Companion PR: https://github.com/dolthub/doltgresql/pull/452

This PR fixes two issues with creating secondary indexes for Doltgres types. The first deals with handlers, as we were not adding a `nil` handler for the additional hash type, which would cause a panic as the counts were not equal (all non-Extended types should have a matching nil handler).

The second issue was due to the reuse of an `ExtendedTupleComparator`. When creating a new `ExtendedTupleComparator`, we pass in the previous `TupleTypeHandler` to handle all non-Extended types. If the previous `TupleTypeHandler` was `ExtendedTupleComparator` and the new one was also `ExtendedTupleComparator`, then we could end up with a misinterpretation of data that could lead to incorrect results, as the handler assumed a different type than the actual type. This has been changed so that `ExtendedTupleComparator` will always use the inner comparator of a previous `ExtendedTupleComparator`. For now this will always be the default comparator, but if we ever add another one, then this should properly handle that change.

### DIFF
--- a/go/store/prolly/tuple_map.go
+++ b/go/store/prolly/tuple_map.go
@@ -435,9 +435,15 @@ func DebugFormat(ctx context.Context, m Map) (string, error) {
 func ConvertToSecondaryKeylessIndex(m Map) Map {
 	keyDesc, valDesc := m.Descriptors()
 	newTypes := make([]val.Type, len(keyDesc.Types)+1)
+	handlers := make([]val.TupleTypeHandler, len(keyDesc.Types)+1)
 	copy(newTypes, keyDesc.Types)
+	copy(handlers, keyDesc.Handlers)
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@Hydrocharged __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e14c64b</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e14c64b</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #8082, opened 2024-06-27
/docker/{docker-entrypoint.sh,serverDockerfile}: change image to pass all args to dolt sql-server command

coffeegoddd:
This PR fixes https://github.com/dolthub/dolt/issues/8079. Now when running `dolthub/dolt-sql-server` if the args `dolt sql-server` are passed to the image, it will error. This will also prevent accidentally starting two Dolt servers in the container.


jennifersp (approved):
LGTM! Thank you!

PR #8092, opened 2024-07-01
[no-release-notes] Disable "stats: boostrap abort over 1mm rows" BATS test in Lambda

fulghum:
This BATS test doesn't seem to run well in Lambda BATS. It failed for me four times in a row today, and I saw it failing a bunch last week, too. Disabling it in Lambda BATS.


max-hoffman (approved):
lgtm

PR #8124, opened 2024-07-15
[stats] limit stats boostrap to server start

max-hoffman:
fixes: https://github.com/dolthub/dolt/issues/8123

### DIFF
--- a/go/libraries/doltcore/sqle/system_variables.go
+++ b/go/libraries/doltcore/sqle/system_variables.go
@@ -224,7 +224,7 @@ func AddDoltSystemVariables() {
 			Dynamic: true,
 			Scope:   sql.GetMysqlScope(sql.SystemVariableScope_Global),
 			Type:    types.NewSystemBoolType(dsess.DoltStatsBootstrapEnabled),
-			Default: int8(1),
+			Default: int8(0),
 		},
 		&sql.MysqlSystemVariable{
 			Name:    dsess.DoltStatsMemoryOnly,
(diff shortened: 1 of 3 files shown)


coffeegoddd:
@max-hoffman __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>474fd48</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>474fd48</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/system_variables.go
```diff
@@ -224,7 +224,7 @@ func AddDoltSystemVariables() {
 			Dynamic: true,
```
nicktobey, line 224: If this is set on server start, should we make Dynamic = false so it can't be reassigned while the engine is running?

I also may not be understanding this change.

max-hoffman: hmm, would that prevent us from PERSISTing a change to the variable? certainly changing it during a running server isn't going to have any effect, its lifecycle scope is engine startup. but maybe there's no harm letting people write it while the server is running? and perhaps sometimes we'd want to set persist from in a session? In contrast, scanning some of the `Dynamic=false` variables I see things like `admin_port`, `admin_address` that we really don't want users to be able to change

nicktobey: Alright, that makes sense.

nicktobey (approved):

go/libraries/doltcore/sqle/system_variables.go
max-hoffman: The edit I made below is that by default we do not bootstrap statistics, and unless otherwise specified in the static configuration during sql-server we will bootstrap

PR #8154, opened 2024-07-25
When removing from a secondary index during merging, use the pre-merge ordinal mapping, not the merged mapping, to identify the secondary key to be removed.

nicktobey:
Basically, we have a bug where dropping a column on the remote side of a merge can interfere with updating secondary indexes.

Example:

Base Schema: (pk INT PRIMARY KEY, a TINYINT, b INT, UNIQUE KEY b_idx (b))

"theirs" drops column a

Merged Schema: (pk INT PRIMARY KEY, b INT, UNIQUE KEY b_idx (b))

In effect, the merger would see that in the final table, `b` is the second column. Then, when updating `b_idx` for each resolved row, it would use the second column of "ours" to find the index entry to remove. But this is incorrect, because the second column of "ours" is `a`.

If the user is lucky, these two columns will be different sizes and the merger will panic. But if the two columns are the same size, merge proceeds with an incorrect value. This will cause it to either fail to remove the old row from the secondary index, or remove a different row. Either way, the secondary index is now incorrect.

### DIFF
--- a/go/libraries/doltcore/merge/mutable_secondary_index.go
+++ b/go/libraries/doltcore/merge/mutable_secondary_index.go
@@ -147,12 +147,6 @@ func (m MutableSecondaryIdx) UpdateEntry(ctx context.Context, key, currValue, ne
 		return err
 	}
 
-	defer func() {
-		if r := recover(); r != nil {
-			currKey, err = m.leftBuilder.SecondaryKeyFromRow(ctx, key, currValue)
-			_ = currKey
-		}
-	}()
 	newKey, err := m.mergedBuilder.SecondaryKeyFromRow(ctx, key, newValue)
 	if err != nil {
 		return err
--- a/go/libraries/doltcore/merge/mutable_secondary_index.go
+++ b/go/libraries/doltcore/merge/mutable_secondary_index.go
@@ -16,6 +16,7 @@ package merge
 
 import (
 	"context"
+
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
--- a/go/libraries/doltcore/merge/mutable_secondary_index.go
+++ b/go/libraries/doltcore/merge/mutable_secondary_index.go
@@ -16,7 +16,6 @@ package merge
 
 import (
 	"context"
-
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
@@ -27,7 +26,7 @@ import (
 )
 
 // GetMutableSecondaryIdxs returns a MutableSecondaryIdx for each secondary index in |indexes|.
-func GetMutableSecondaryIdxs(ctx *sql.Context, sch schema.Schema, tableName string, indexes durable.IndexSet) ([]MutableSecondaryIdx, error) {
+func GetMutableSecondaryIdxs(ctx *sql.Context, ourSch, sch schema.Schema, tableName string, indexes durable.IndexSet) ([]MutableSecondaryIdx, error) {
 	mods := make([]MutableSecondaryIdx, sch.Indexes().Count())
 	for i, index := range sch.Indexes().AllIndexes() {
 		idx, err := indexes.GetIndex(ctx, sch, nil, index.Name())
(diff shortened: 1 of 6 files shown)


coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c189d9e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c189d9e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>bf0e938</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>bf0e938</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/merge/mutable_secondary_index.go
```diff
@@ -140,12 +142,18 @@ func (m MutableSecondaryIdx) InsertEntry(ctx context.Context, key, newValue val.
...
 		return err
 	}
 
-	newKey, err := m.builder.SecondaryKeyFromRow(ctx, key, newValue)
+	defer func() {
+		if r := recover(); r != nil {
+			currKey, err = m.leftBuilder.SecondaryKeyFromRow(ctx, key, currValue)
+			_ = currKey
+		}
+	}()
```
fulghum, line 155: Is this needed or was this just for debugging?

go/libraries/doltcore/merge/schema_merge_test.go
```diff
@@ -1083,6 +1083,20 @@ var secondaryIndexTests = []schemaMergeTest{
 		skipOldFmt:          true,
 		skipFlipOnOldFormat: true,
 	},
+	{
+		name:     "dropping column on right shifts column index between compatible types (see pr/8154)",
```
fulghum, line 1087: It looks like the `incompatible types` test case below errors out with a panic without the fix for this issue, but the `compatible types` test case seems to still pass when I ran it against main. 

Got any ideas on a better way to test the correctness of the updates to the secondary index that would catch the data issue even though this one was able to merge correctly? I had thought the code in `ValidateDatabase()` / `validateSecondaryIndexes` would catch it, but it doesn't seem to be running for these tests. Turning that validation on for these tests could be one easy way to catch that. 

Might also be nice to cover insert/update/delete cases with the secondary index update if we can, too.

fulghum (approved):
Code changes look good, nice job finding and fixing this issue. 

I think one test case needs a little more work to make sure we detect the secondary index data update issue for compatible types.

go/libraries/doltcore/merge/mutable_secondary_index.go
nicktobey: Whoops. That was for debugging. Removed.

go/libraries/doltcore/merge/schema_merge_test.go
nicktobey: Fixed with a new test `dolt_queries_schema_merge`

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b1d6ae0</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b1d6ae0</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #8186, opened 2024-08-01
Add support for Doltgres indexes

Hydrocharged:
* https://github.com/dolthub/doltgresql/pull/561


coffeegoddd:
@Hydrocharged __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>17851bb</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>17851bb</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

max-hoffman (commented):
I like the doltgresql side, I mentioned this before but I think reimplementing map iters isn't the best separation of concerns. @zachmu is better at interface tradeoffs, but I think hiding all of the details behind a different `Range.Matches` is better than duplicating iterators. I could be missing some bespoke postgres logic that requires more control, but it looks like binary search callbacks to me. If "findStart" / "findStop" can't be rewritten for our existing star cursor/end cursor iters, I think it'd be straightforward to add those primitives to the tree layer rather than implementing in the `index` package. Even if we branch logic without interface niceties, it's just such a smaller surface area to test/way greater code reusability if we keep the map iter implementations narrow. Duplicating fixes to all of the noms iters for so long was inefficient, a bit of time spent avoiding that state again would be my vote. I apologize for the handwavy nature of these comments, hopefully it makes sense from prior context, and happy to dive into more specifics if you want.

max-hoffman (approved):
This seems like it's in a place that would be amenable to refactors. It'd be nice to maybe document why there is non-logical overlap between mysql and PG ranges, in case we want to try to make the arithmetic faster at some point. The start/stop semantics don't seem bad, but it is different from how we do it in MySQL, and if I had to guess it seems like both the MySQL and PG sides could use either interface with the right compensating filters.

go/libraries/doltcore/sqle/dtables/commit_diff_table.go
```diff
@@ -151,11 +151,15 @@ func (dt *CommitDiffTable) Partitions(ctx *sql.Context) (sql.PartitionIter, erro
 }
 
 func (dt *CommitDiffTable) LookupPartitions(ctx *sql.Context, i sql.IndexLookup) (sql.PartitionIter, error) {
-	if len(i.Ranges) != 1 || len(i.Ranges[0]) != 2 {
+	ranges, ok := i.Ranges.(sql.MySQLRangeCollection)
+	if !ok {
+		return nil, fmt.Errorf("commit diff table requires MySQL ranges")
```
zachmu, line 156: What's the plan here? Is the range abstraction in the GMS PR incomplete and will be added to later, so that we can use it here and other places that now require a type cast? Or do we need to re-implement range semantics everywhere a Doltgres range type is present?

Hydrocharged:
https://github.com/dolthub/doltgresql/pull/561

coffeegoddd:
@Hydrocharged __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>24b49d0</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>24b49d0</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #8227, opened 2024-08-07
[no-release-notes] Skip loading DB when creating the single-DB environment: it will get thrown out when we create the multi-db environment anyway.

nicktobey:


nicktobey:
Closed because the existence of `singletonDB` means there's not actually a performance penalty for loading a DB multiple times.

PR #8244, opened 2024-08-09
Reuse DoltEnv when creating MultiEnv

nicktobey:
I thought this would speed up startup latency by as much as 50% when garbage hasn't been collected in a while. Previously we were loading the environment (including the index) from disk twice. Now we only do it once.

It turns out that reading from the db is cached so this won't actually affect startup latency. But it simplifies the code flow in most cases.

### DIFF
--- a/go/cmd/dolt/commands/backup.go
+++ b/go/cmd/dolt/commands/backup.go
@@ -335,7 +335,7 @@ func restoreBackup(ctx context.Context, dEnv *env.DoltEnv, apr *argparser.ArgPar
 		return errhand.VerboseErrorFromError(err)
 	}
 
-	mrEnv, err := env.MultiEnvForDirectory(ctx, dEnv.Config.WriteableConfig(), dEnv.FS, dEnv.Version, dEnv)
+	mrEnv, err := env.MultiEnvForDirectory(ctx, dEnv.Config.WriteableConfig(), dEnv.FS, dEnv.Version, dEnv, false)
 	if err != nil {
 		return errhand.BuildDError("error: Unable to list databases").AddCause(err).Build()
 	}
--- a/go/cmd/dolt/commands/dump.go
+++ b/go/cmd/dolt/commands/dump.go
@@ -813,7 +813,7 @@ func addCreateDatabaseHeader(dEnv *env.DoltEnv, fPath, dbName string) errhand.Ve
 // TODO: find a more elegant way to get database name, possibly implement a method in DoltEnv
 // getActiveDatabaseName returns the name of the current active database
 func getActiveDatabaseName(ctx context.Context, dEnv *env.DoltEnv) (string, errhand.VerboseError) {
-	mrEnv, err := env.MultiEnvForDirectory(ctx, dEnv.Config.WriteableConfig(), dEnv.FS, dEnv.Version, dEnv)
+	mrEnv, err := env.MultiEnvForDirectory(ctx, dEnv.Config.WriteableConfig(), dEnv.FS, dEnv.Version, dEnv, false)
 	if err != nil {
 		return "", errhand.VerboseErrorFromError(err)
 	}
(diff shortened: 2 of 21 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>4b0928a</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>4b0928a</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e07e403</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e07e403</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c3506c8</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c3506c8</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>9330687</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>9330687</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

reltuk:
Just my opinion, but the naked boolean parameter, having to choose its value at the call site correctly, having to check for certain error conditions and nil out error states on the dEnv, etc. I think it's a little hard to argue this change simplifies the code.

I do think it's confusing to have multiple dEnv instances floating around for the same directory on disk, and for those to have been loaded at different times in the application's lifecycle, etc. I'm just not certain that this change in particular actually simplifies the code. That being said, it might still be worth it if we're concerned about the runtime behavior, the potential for confusion between the instances, etc.

Even with this change, things seem quite subtle and hard to get right. Are the less subtle now than before? As someone who's not close to it at all and has only read over the PR, it's hard for me in particular to say.

fulghum:
I agree with Aaron. The change looks correct of course, but I'm not convinced it's worth merging in unless there was some customer benefit like a performance improvement. For me, the extra boolean param makes it a little harder to call `MultiEnvForDirectory` correctly, and even after reading the code, I couldn't tell you why we need to pass `true` in one place, but nowhere else. I think the current version (where we reload the main env) is simpler to call, easier to call correctly, and still has the same performance characteristics.

nicktobey:
I agree with your comments. While this technically avoids some redundant logic, it adds its own complexity, and without a clear benefit (such as a performance improvement), there's no compelling reason to merge it.

PR #8260, opened 2024-08-13
Fixes for dolt installer bash-script

imcrazytwkr:
1. Fixed error message mistakenly implying that dolt can be installed on 32-bit systems
2. Fixed color printing on bash (and zsh running in bash compatibility mode) installed on an OS with non-GNU coreutils (Alpine Linux, macOS)
3. Linted `install.sh` and fixed all the errors and warning that had obvious fixes. The only one left is `local error_code="$1"` in `fail` function that should be removed but may be reused for debugging purposes by manually editing the script

### DIFF
--- a/go/utils/publishrelease/install.sh
+++ b/go/utils/publishrelease/install.sh
@@ -17,11 +17,11 @@ _() {
 
 set -euo pipefail
 
-DOLT_VERSION=__DOLT_VERSION__
-RELEASES_BASE_URL=https://github.com/dolthub/dolt/releases/download/v"$DOLT_VERSION"
-INSTALL_URL=$RELEASES_BASE_URL/install.sh
+DOLT_VERSION='__DOLT_VERSION__'
+RELEASES_BASE_URL="https://github.com/dolthub/dolt/releases/download/v$DOLT_VERSION"
+INSTALL_URL="$RELEASES_BASE_URL/install.sh"
 
-CURL_USER_AGENT=${CURL_USER_AGENT:-dolt-installer}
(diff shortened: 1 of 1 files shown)


timsehn:
@reltuk should look at this...

go/utils/publishrelease/install.sh
```diff
@@ -30,84 +30,88 @@ WORK_DIR=
...
 install_binary_release() {
-  local FILE=dolt-$PLATFORM_TUPLE.tar.gz
-  local URL=$RELEASES_BASE_URL/$FILE
-  echo "Downloading:" $URL
-  curl -A "$CURL_USER_AGENT" -fsL "$URL" > "$FILE"
+  local FILE="dolt-$PLATFORM_TUPLE.tar.gz"
+  local URL="$RELEASES_BASE_URL/$FILE"
+
+  echo "Downloading: $URL"
+  curl -a "$CURL_USER_AGENT" -fsL "$URL" > "$FILE"
```
reltuk, line 107: This change from `-A` to `-a` doesn't look right to me.

https://curl.se/docs/manpage.html#-A

reltuk (commented):
Thanks for the contribution. This looks great.

One question about a change in the `curl` invocation.

go/utils/publishrelease/install.sh
imcrazytwkr: Sorry, my bad on this one. Let me push a quick fix.

reltuk:
@imcrazytwkr Thank you for the contribution! Awesome to have the script a bit better and cleaned up like this :).

PR #8428, opened 2024-10-09
Minor changes to support Doltgres `TEXT` columns

fulghum:
Minor changes to support Doltgres' use of `TEXT` columns in secondary indexes and unique constraints. 
* ensures `GetRunningServer()` is available when `controller.WaitForStart()` completes (this triggers after all `init` functions have run.
* updates `sql.Index` implementations so that prefix lengths can be set.
* adds a note explaining the issue with trying to pass addresses to `SerializedCompare()` without first de-reff'ing them.

Depends on: https://github.com/dolthub/go-mysql-server/pull/2687


fulghum:
Closing this and switching to a different approach that requires less changes in GMS

PR #8474, opened 2024-10-22
[no-release-notes] Replace ConvertToSecondaryKeylessIndex with AddHashToSchema

nicktobey:
This is a refactor that eliminates a hack in how we deal with secondary indexes on keyless tables.

Keyless tables are unique in that the tuples stored in their prolly maps don't match their schemas: they have a hidden key column which doesn't appear in their schema; this key column stores a hash of the value tuple, and is used as a hidden primary key.

This mismatch between schema and structure poses some problems when creating secondary indexes. The current solution is to first make an index using the schema, and then conditionally create a second index with the correct descriptors. This is inefficient and error-prone: if we forget to perform this correction, we'll have an index that is invalid and unusable. It also ties the process of working with indexes to a specific implementation (one that uses `prolly.Map`), which causes issues with vector indexes, which have the same caveats around keyless tables but don't use `prolly.Map`.

### DIFF
--- a/go/cmd/dolt/commands/schcmds/import.go
+++ b/go/cmd/dolt/commands/schcmds/import.go
@@ -330,7 +330,7 @@ func putEmptyTableWithSchema(ctx context.Context, tblName string, root doltdb.Ro
 		return nil, errhand.BuildDError("error: failed to get table.").AddCause(err).Build()
 	}
 
-	empty, err := durable.NewEmptyIndex(ctx, root.VRW(), root.NodeStore(), sch)
+	empty, err := durable.NewEmptyIndex(ctx, root.VRW(), root.NodeStore(), sch, false)
 	if err != nil {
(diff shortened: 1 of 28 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>010b7da</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>010b7da</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

max-hoffman (approved):
LGMT, the rearrangement looks harmless

PR #8538, opened 2024-11-06
Fix JSON merge issue that would report imprecise diffs in some situations

nicktobey:
If:
- A chunk begins with an object "A"
- A value "A.b" within this object was modified
- The previous chunk was also modified

Then the differ would incorrectly report that the entire "A" object had been modified, instead of the sub-value "A.b"

This could then lead to situations where the merge cannot be auto-resolved, because the differ claims that an object has been modified divergently, when it's actually two unrelated fields in the object that have been modified.

This PR fixes that scenario by properly detecting when the next chunk marks the start of an object.

### DIFF
--- a/go/store/prolly/tree/indexed_json_diff.go
+++ b/go/store/prolly/tree/indexed_json_diff.go
@@ -245,10 +245,24 @@ func (jd *IndexedJsonDiffer) Next(ctx context.Context) (diff JsonDiff, err error
 			case 0:
 				key := fromCurrentLocation.Clone().key
 
+				fromNextCharacter, err := jd.currentFromCursor.nextCharacter(ctx)
+				if err == io.EOF {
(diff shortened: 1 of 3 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c9b6269</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c9b6269</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

fulghum (approved):
LGTM

PR #8612, opened 2024-11-27
go/store/nbs: Fixing GCGen to be more correct.

reltuk:
The original purpose of gc gen was two fold. The first purpose was to avoid applying the garbage collection results if the store had changed due to multi-process concurrency for any reason. The second purpose was to fast-complete a `dolt gc` invocation if the store had not changed at all since the last GC run.

For the first purpose, it is no longer necessary. We no longer allow multi-process access to the same NomsBlockStore.

For the second purpose, it was implemented slightly incorrectly, given the introduction of `dolt gc --full`. This change fixes the implementation to be more correct.

In particular, the semantics are:

* After a `dolt gc --full`, an immediate invocation of `dolt gc` or `dolt gc --full` fast-completes as no collection being necessary.

* After a `dolt gc`, only a `dolt gc` fast-completes as no collection being necessary. A `dolt gc --full` will run a full GC to completion.


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>adfa851</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>adfa851</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/manifest.go
```diff
@@ -501,7 +501,7 @@ func formatSpecs(specs []tableSpec, tableInfo []string) {
 // persisted manifest against the lock hash it saw last time it loaded the
 // contents of a manifest. If they do not match, the client must not update
 // the persisted manifest.
-func generateLockHash(root hash.Hash, specs []tableSpec, appendix []tableSpec) hash.Hash {
+func generateLockHash(root hash.Hash, specs []tableSpec, appendix []tableSpec, extra []byte) hash.Hash {
```
nicktobey, line 504: I'm a bit skeptical about this. Using arbitrary extra bytes makes it extensible but feels messy. I'm going to keep reading and circle back on this.

integration-tests/bats/garbage_collection.bats
```diff
@@ -476,3 +469,114 @@ SQL
...
+    dolt gc --full
+    rm -rf .dolt/stats
+    ls -laR > ../../files_after_second_gc
+    # This should exit 0 because the gc should have changed things.
+    if cmp ../../files_before_gc ../../files_after_first_gc; then
+        echo "expected dolt gc to change things, but it didn't."
+        diff ../../files_before_gc ../../files_after_first_gc || true
+        false
+    fi
+    # This should exit non-0 because the gc should NOT have changed things.
```
nicktobey, line 575: This comment is incorrect: we expect things to change here

integration-tests/bats/garbage_collection.bats
```diff
@@ -476,3 +469,114 @@ SQL
...
+    if cmp ../../files_after_first_gc ../../files_after_second_gc && cmp ./.dolt/noms/manifest ../../manifest_after_first_gc; then
```
nicktobey, line 576: Do we expect both the files and the manifest to change or just one of them? We should break this into two separate checks.

go/store/nbs/manifest.go
nicktobey: I think this is fine for now. If we ever need to change the manifest format again we can think about it more.

nicktobey (approved):
LGTM, a couple of nits in the bats tests.

go/store/nbs/manifest.go
reltuk: Yeah, agreed this is not the greatest, but it works for now.

integration-tests/bats/garbage_collection.bats
reltuk: In reality, they can both change, or only the manifest can change. But the manifest must change.

The file listing for this particular case only changes if the order that the chunks are written to the stores is different or if the created table files as a result of the GC operation have two different create timestamps because the operations are run in different minutes.

So best to check for the lack of differences in both in the no-change case, and sufficient to check for a change in just the manifest in this case. So I've changed it to just look at the manifest.

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>60ca378</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>60ca378</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #8630, opened 2024-12-03
Fixed issues with branch control

Hydrocharged:
This is primarily to fix issues with:
* https://github.com/dolthub/dolt/issues/8622
* https://github.com/dolthub/dolt/issues/8623

For the first issue, we did not actually update the longest match counter, and it was therefore being ignored. The variable existed, but it was not being updated. I think this went so long without being caught (even with all of the tests in `enginetests` and `bats`) due to how we handle expression folding, where subset entries are completely ignored if a superset already exists.

The second issue is somewhat similar to the first, in that we were not checking the permission set when determining whether a new row is actually a subset. If an existing row is a superset, but the permissions are more restrictive than the new row, then it's _not an actual superset_, since it has more restrictive permissions. This is even in the documentation, so it's a simple mistake of forgetting to add that check to the code itself.

Lastly, the `bats` tests have been updated such that they rely on sockets now, and Windows does not support sockets. This was not caught since most Windows developers have moved to primary development on [Doltgres](https://github.com/dolthub/doltgresql), so this at least allows such developers to run the `bats` tests locally again.

### DIFF
--- a/go/libraries/doltcore/branch_control/access.go
+++ b/go/libraries/doltcore/branch_control/access.go
@@ -78,6 +78,7 @@ func (tbl *Access) Match(database string, branch string, user string, host strin
 	for _, result := range results {
 		if result.Length > length {
 			perms = result.Permissions
+			length = result.Length
 		} else if result.Length == length {
 			perms |= result.Permissions
 		}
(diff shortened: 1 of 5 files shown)


coffeegoddd:
@Hydrocharged __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b78310b</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b78310b</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

zachmu (approved):
LGTM!

PR #8632, opened 2024-12-03
fix error message for duplicate unique key violation on out of band types

jycor:
Certain types are stored out of band (like `TEXT` and `BLOB`), so we compare their addresses to test for existence. Consequently, the error message would contain the hash rather than the contents.

This PR fixes the formatting to print the prefix.

fixes: https://github.com/dolthub/dolt/issues/8629

### DIFF
--- a/go/libraries/doltcore/sqle/writer/prolly_index_writer.go
+++ b/go/libraries/doltcore/sqle/writer/prolly_index_writer.go
@@ -16,6 +16,7 @@ package writer
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"strings"
 
@@ -125,7 +126,12 @@ func (m prollyIndexWriter) ValidateKeyViolations(ctx context.Context, sqlRow sql
 	if err != nil {
 		return err
 	} else if ok {
-		keyStr := FormatKeyForUniqKeyErr(k, m.keyBld.Desc)
+		remappedSqlRow := make(sql.Row, len(sqlRow))
+		for to := range m.keyMap {
+			from := m.keyMap.MapOrdinal(to)
+			remappedSqlRow[to] = sqlRow[from]
+		}
+		keyStr := FormatKeyForUniqKeyErr(k, m.keyBld.Desc, remappedSqlRow)
 		return m.uniqueKeyError(ctx, keyStr, k, true)
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


coffeegoddd:
@jycor __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>43f57d8</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>43f57d8</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
```diff
@@ -422,7 +438,13 @@ func FormatKeyForUniqKeyErr(key val.Tuple, d val.TupleDesc) string {
 			sb.WriteString(",")
 		}
 		seenOne = true
-		sb.WriteString(d.FormatValue(i, key.GetField(i)))
+		switch d.Types[i].Enc {
+		// address encodings should be printed as strings
+		case val.BytesAddrEnc, val.StringAddrEnc:
```
fulghum, line 443: Do we need to include `jsonAddrEnc`? 

I'm not sure if we should print out the blob content as a string... perhaps as hex bytes or something? I'd guess that more often than not, text in a `BLOB` is binary data and won't be valid string data.

fulghum (approved):
Looks good! Nice work making sure that error message has user content in it and not an address that users wouldn't know what to do with.

go/libraries/doltcore/sqle/writer/prolly_index_writer.go
jycor: MySQL's converts blob types to string representation in just the error message.
```
mysql> insert into t values ('hello');
ERROR: 1062 (23000): Duplicate entry 'hello' for key 't.i'

mysql> insert into t values (0x68656c6c6f);
ERROR: 1062 (23000): Duplicate entry 'hello' for key 't.i'
```

JSON indexes must be over generated columns, so I don't think that applies here

Issue #7980, opened 2024-06-07
Create a config.yaml file when database server is first run


zachmu:
This is standard practice in the database world: you get a config file with the defaults when you install the server.

https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-CONFIGURATION-FILE

Dolt makes you write your own.

Instead, Dolt should drop a config.yaml file with the default values (modified as necessary for the current execution) in the same location as the database directories when `sql-server` is run.

farhanhubble:
Same goes for config.json. It is mentioned in the docker instructions mapping host directories to docker directories but no standard template is provided anywhere.

zachmu:
Thanks for the feedback, agree we could do better here.

milogreg:
I'd like to work on this, and would appreciate a little help before I get started. 

I plan to use the defaults described in sqlServerDocs in go/cmd/dolt/commands/sqlserver/sqlserver.go. Is there a better canonical set of defaults I should be sourcing from?

I am also not sure how the options with no defined defaults should be handled. Should a comment be left in the file as a placeholder, should they not be included at all, or is there a way to explicitly set a config option to undefined/unset?

Lastly, I'm assuming by "modified as necessary for the current execution" you mean the config file should include the corresponding config options for the flags sql-server is executed with. Is this assumption correct?

Thanks.

zachmu:
This is implemented in the latest release.

PR #8644, opened 2024-12-05
Generate config.yaml when running sql-server 

milogreg:
If sql-server is ran without a specified config file and there is no config.yaml in the database directory, one will be generated.

The rules for how the default config.yaml file is generated are as follows:
- If a field has a value defined by the sql-server execution (such as through CLI args), then that value will be used.
- If a field has no set value but has a default value, then that default will be used.
- If a field has no set value and no default, a commented-out line setting the field to null will be included as a placeholder.

Part of #7980


timsehn:
Very cool! 

My first comment is that it could use some bats tests. It probably should have its own test file.

Here's an example to crib from:

https://github.com/dolthub/dolt/blob/main/integration-tests/bats/sql-server.bats

Also, the bats readme can help you get started with running your tests:

https://github.com/dolthub/dolt/tree/main/integration-tests/bats#readme

milogreg:
Made some bats tests.

However, I noticed some other tests using sql-server skip when running windows due to issues I'm unaware of, so I'm not sure if the tests will behave properly on windows yet (I've only ran them on Mac).

timsehn:
@jycor will review this today.

zachmu:
Some of these bats tests are failing because we just removed a couple of config keys the tests are expecting. Merge main, run locally, and fix the tests.

zachmu:
Also @milogreg you should join the discord to talk in real time for faster turnaround on these issues.

jycor (commented):
Thanks for this contribution! I like the high level idea of this PR.
The BATS are very thorough and well tested!

However, I don't like the use of `reflect` for `fillDefault`. I think it would be better to explicitly check each field of the struct and fill accordingly.

Lastly, I'm concerned about how much longer BATS is taking. I'm not sure if it's a side effect of github actions on a PR from a fork or the the config generation is slowing it down signigicantly.

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -245,6 +244,11 @@ func StartServer(ctx context.Context, versionStr, commandStr string, args []stri
 		return err
 	}
 
+	err = generateYamlConfigIfNone(ap, help, args, dEnv, serverConfig)
```
zachmu, line 247: This should only run in the case that an explicit value for --config wasn't provided

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -432,3 +436,51 @@ func setupDoltConfig(dEnv *env.DoltEnv, apr *argparser.ArgParseResults, config s
 
 	return nil
 }
+
+func generateYamlConfigIfNone(
```
zachmu, line 440: Doc comments for new funcs plz

go/cmd/dolt/commands/sqlserver/sqlserver.go
```diff
@@ -432,3 +436,51 @@ func setupDoltConfig(dEnv *env.DoltEnv, apr *argparser.ArgParseResults, config s
...
+
+	if specifiesConfigFile {
+		return nil
+	}
+
+	path := filepath.Join(serverConfig.DataDir(), yamlConfigName)
+	exists, isDir := dEnv.FS.Exists(path)
+	if exists {
+		if isDir {
+			cli.PrintErrf("Couldn't generate YAML config at %s: directory with same name already exists", path)
```
zachmu, line 461: This should just be silently ignored

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -249,7 +247,113 @@ func clusterConfigAsYAMLConfig(config ClusterConfig) *ClusterYAMLConfig {
...
+	withDefaults.fillDefaults()
+
+	formatted := formattedYAMLMarshal(removeOmitemptyTags(withDefaults))
+	formatted = commentNullYAMLValues(formatted)
+
+	return formatted
+}
+
+// Assumes YAMLConfig has no circular references.
+func (cfg *YAMLConfig) fillDefaults() {
```
zachmu, line 267: For methods like this, we prefer to use the convention `withDefaultsFilledIn` and return a copy of the object. Also write a descriptive method comment describing what it does.

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -249,7 +247,113 @@ func clusterConfigAsYAMLConfig(config ClusterConfig) *ClusterYAMLConfig {
...
+	defaults := defaultServerConfigYAML()
+	recursiveFillDefaults(reflect.ValueOf(cfg), reflect.ValueOf(defaults))
+}
+
+func recursiveFillDefaults(cfgValue, defaultsValue reflect.Value) {
```
zachmu, line 272: Like James said, I think we would prefer to just explicitly list all the fields here, rather than use reflection.

go/libraries/doltcore/servercfg/yaml_config_test.go
```diff
@@ -415,3 +415,34 @@ listener:
 	err = ValidateConfig(cfg)
 	assert.Error(t, err)
 }
+
+// Tests if a circular reference exists in YAMLConfig.
+//
+// This test can be safely removed if:
+//   - YAMLConfig.VerboseString is updated to properly handle circular references.
+//   - yaml.Marshal is checked to make sure it properly handles circular references.
+func TestNoCircularReferenceInYAMLConfig(t *testing.T) {
```
zachmu, line 424: I don't think I understand how a YAML file can have a circular reference in it. In any case it should be unnecessary if you don't use reflection.

go/cmd/dolt/commands/sqlserver/sqlserver.go
zachmu: The code you have now will correctly not overwrite a file named config.yaml, but it should never attempt to write a file if a config file was provided. Plz add a test for this too

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -279,6 +383,19 @@ func (cfg YAMLConfig) String() string {
 	return result
 }
 
+func commentNullYAMLValues(needsComments string) string {
```
zachmu, line 386: Would be nice to have a unit test of this behavior in `yaml_config_test.go`

integration-tests/bats/sql-server-config-file-generation.bats
```diff
@@ -0,0 +1,142 @@
...
+        run cat "$data_dir/$NOT_CONFIG_FILE_NAME"
+        [ $status -eq 0 ]
+        [[ "$output" =~ "Don't overwrite me!" ]] || false
+
+        rm "$data_dir/$NOT_CONFIG_FILE_NAME"
+        stop_sql_server
+    done
+}
+
+@test "sql-server-config-file-generation: generated config file has fields set to arg-provided values" {
```
zachmu, line 94: Should check for commented out values in at least one of these tests

zachmu (commented):
Functionality looks good but see comments

milogreg:
> Thanks for this contribution! I like the high level idea of this PR. The BATS are very thorough and well tested!
> 
> However, I don't like the use of `reflect` for `fillDefault`. I think it would be better to explicitly check each field of the struct and fill accordingly.
> 
> Lastly, I'm concerned about how much longer BATS is taking. I'm not sure if it's a side effect of github actions on a PR from a fork or if the config generation is slowing it down signigicantly.

I'll remove that usage of reflect, and look into potential performance issues.

I didn't put much thought into performance when designing this, as I falsely assumed that sql-server wasn't something that would be ran repeatedly from a clean state (I clearly didn't take testing into account). So, I wouldn't be surprised at all if the config generation is the cause of the slowdown.

The first thing I'll try is removing the expensive usages of reflection (aside from the ones in the YAML marshaling itself), and see how that affects the runtime (hopefully this is all it will take to make it performant). If that isn't enough, I'll try to actually nail down what the bottleneck is and redesign accordingly.

However, I don't think I know enough to determine what a reasonable target time per config generation would be, is there an approximate maximum time that would be acceptable? I'll be benchmarking it on my m3 MacBook pro.

integration-tests/bats/sql-server-config-file-generation.bats
milogreg: The test "generated config file has placeholders for unset fields with no default values" checks for commented-out values, unless that isn't what you are talking about.

go/cmd/dolt/commands/sqlserver/sqlserver.go
milogreg: The bats test "config file isn't generated if a config is specified in args" should cover this (it gives a config file not named config.yaml and makes sure a new config.yaml isn't generated).

milogreg:
From my extremely quick testing it looks like almost all of the overhead of the config generation comes from my usage of dEnv.FS.WriteFile, which is many times slower than os.WriteFile. I think I just did it like this because I saw other parts of the codebase write files in this way. Is it ok to switch to os.WriteFile, or is there another more proper way of writing files faster?

milogreg:
I looked into the dEnv.FS.WriteFile performance difference, it looks like it's significantly slower than os.WriteFile because it writes the file atomically. I think the implementation of dEnv.FS.WriteFile (WriteFileAtomically in go/libraries/utils/file/sync.go) is pretty optimal, so this overhead may be unavoidable unless we want to give up the atomic guarantee. I can't think of a major reason as to why it would be bad to leave an incomplete config.yaml in the event of a crash, so I'd assume it would be okay to use the non-atomic WriteFile, but I wont make that decision by myself. Another option, for if an atomic write isn't needed but using dEnv.FS is required, is to use dEnv.FS.OpenForWrite (which is about as performant as os.WriteFile).

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -249,7 +247,149 @@ func clusterConfigAsYAMLConfig(config ClusterConfig) *ClusterYAMLConfig {
...
+	}
+	if withDefaults.BranchControlFile == nil {
+		withDefaults.BranchControlFile = defaults.BranchControlFile
+	}
+
+	return withDefaults
+}
+
+// Assumes 'in' has no circular references.
+func removeOmitemptyTags(in any) any {
```
zachmu, line 333: I think we have the same comment here: we'd rather just avoid the use of reflection altogether and do this manually. Just list the fields that have `omitempty` manually rather than examining them via reflection.

integration-tests/bats/sql-server-config-file-generation.bats
zachmu: That works

go/libraries/doltcore/servercfg/yaml_config_test.go
```diff
@@ -415,3 +414,33 @@ listener:
 	err = ValidateConfig(cfg)
 	assert.Error(t, err)
 }
+
+func TestCommentNullYAMLValues(t *testing.T) {
```
zachmu, line 418: This test is good but this is important enough behavior we want to exercise it a little more thoroughly, and to be more end-to-end. These are nice because it lets a human reviewer see what the behavior is explicitly.

When I see it spelled out like this, I notice the behavior here isn't quite what we want. A commented-out line with `value1: null` tells the user there's a key called `value1` but doesn't tell them what a normal value should be if they decide they want to set it. The goal is to create a self-documenting process where a customer gets a file created that they can use to learn about the configuration options and change them themselves going forward.

Better end-to-end behavior would be:

* Every key explicitly specified via command line flags is set
* Every other key is commented out with its default value (including those in structs)
* Config elements that currently have no default defined need example values to server as placeholders for this use case, rather than `null`. You will need to define these explicitly using the docs as a guide.
* A comment at the top of the file indicating it was generated and that users should uncomment and edit lines as necessary.

With all that in place, what I want to see is one or more Go tests like this one, but for the entire process end-to-end, i.e. "given these command line options, the following config file is produced verbatim."

go/libraries/doltcore/servercfg/yaml_config.go
zachmu: See also my other comment in the test file. `omitempty` isn't the best determinant of whether to comment out a line of config in the file produced.

zachmu (commented):
Thanks for the tests, seeing a concrete example of the output made it clear how this could be improved.

This is getting close, appreciate the effort on this.

zachmu:
Looks like you'll need to merge main again.

zachmu:
@milogreg do you still want to work on this? We can probably run it home for you if not. Let us know either way.

milogreg:
Thanks for checking in, I plan to work on it over the next week.

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -276,6 +342,229 @@ func (cfg YAMLConfig) String() string {
...
+	}
+
+	if withPlaceholders.SystemVars_ == nil {
+		withPlaceholders.SystemVars_ = map[string]interface{}{
+			"dolt_transaction_commit": 1,
+			"dolt_log_level":          "info",
+		}
+	}
+
+	if withPlaceholders.Jwks == nil {
```
zachmu, line 454: Default for this field should be empty (it's advanced and poorly documented, currently only used for our hosting service)

go/libraries/doltcore/servercfg/yaml_config.go
```diff
@@ -127,21 +127,21 @@ type YAMLConfig struct {
...
+	BehaviorConfig    BehaviorYAMLConfig     `yaml:"behavior,omitempty"`
+	UserConfig        UserYAMLConfig         `yaml:"user,omitempty"`
+	ListenerConfig    ListenerYAMLConfig     `yaml:"listener,omitempty"`
 	PerformanceConfig *PerformanceYAMLConfig `yaml:"performance,omitempty"`
 	DataDirStr        *string                `yaml:"data_dir,omitempty"`
 	CfgDirStr         *string                `yaml:"cfg_dir,omitempty"`
-	MetricsConfig     MetricsYAMLConfig      `yaml:"metrics"`
-	RemotesapiConfig  RemotesapiYAMLConfig   `yaml:"remotesapi"`
+	MetricsConfig     MetricsYAMLConfig      `yaml:"metrics,omitempty"`
+	RemotesapiConfig  RemotesapiYAMLConfig   `yaml:"remotesapi,omitempty"`
```
zachmu, line 137: We want ClusterConfig and MetricsConfig and Jwks the bottom of the file, since they are only for advanced users. Since the yaml marshaler outputs in struct field order, that means we should move these fields to the end of the struct.

zachmu (commented):
This looks really good, just minor comments.

There's one field which is internal only / undocumented, so we don't want it showing up in the generated config file. Up to you how to best address that.

zachmu (approved):
This looks great. Assuming this passes CI, I'll merge it in later today.

zachmu:
@milogreg if you are interested in another interesting getting-started task I have one for you, let me know.

PR #8675, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


fulghum (approved):
Thanks for the contribution @RiceChuan! This looks good to me!

PR #8719, opened 2025-01-07
Replace min/max helpers with built-in min/max

Juneezee:
We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max


macneale4:
Thanks for the submission! I'm going to move this into a PR I own. I think that's what is causing the failed checks.

Juneezee:
> Thanks for the submission! I'm going to move this into a PR I own. I think that's what is causing the failed checks.

I think the `${{ github.head_ref }}` does not work for fork repositories.

https://github.com/dolthub/dolt/blob/80ad9a2e9b5dad7b8ef352d1e7e86a90943f705a/.github/workflows/sql-regressions.yaml#L28-L31

We can see from the job log that it is trying to checkout `refactor/minmax`, but this branch only exists in my fork repository.

https://github.com/dolthub/dolt/actions/runs/12662844114/job/35386960409#step:4:3

Perhaps we should also set the repository for `actions/checkout`? https://github.com/actions/checkout/issues/551#issuecomment-1113089390

Juneezee:
According to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context, it looks like `github.ref` is the correct one:

> The fully-formed ref of the branch or tag that triggered the workflow run. ... For workflows triggered by `pull_request`, this is the pull request merge branch. ... The ref given is fully-formed, for pull requests it is `refs/pull/<pr_number>/merge`

macneale4 (approved):
LGTM. And passed CI. Thanks for the contribution!

Issue #7984, opened 2024-06-09
Implement `dolt_help` set of system tables


timsehn:
We should have a `dolt_help` system table with the following schema:

target (key): ie. dolt_checkout, dolt_diff, dolt_hashof, etc 
type: system table, procedure, function, variable
short description: the short help text
long description: long help text
arguments: a json blob describing the arguments

This would help move our documentation in product and be generally useful. We could even generate the docs site documentation from this code.

milogreg:
Hey, I've been working on this a bit, but had a few blocking questions about how things should be structured.

---

Should the short and long help texts be copied directly existing markdown docs, or should new plaintext-readable docs be created based on them? If automatic docs site generation is a priority then I'm assuming at least the long help text will need to be markdown.

---

Storing procedure arguments in JSON would be straightforward as they pretty much fit the CLI model of {argument:description}, but It's a little ambiguous how function arguments should be structured in JSON. I'm assuming that the argument JSON for functions would contain a list of parameter layouts and a list of arguments. It's clear that individual arguments should have a description, but there is some argument behavior that doesn't fit neatly under a specific named argument or spans multiple arguments (see the argument description of [DOLT_REFLOG](https://docs.dolthub.com/sql-reference/version-control/dolt-sql-functions#dolt_reflog)). Should there be a global description that applies to all layouts and arguments, per-layout descriptions that apply to all arguments within a given layout, or both?

---

If the docs will eventually be generated using this table, then whatever populates the table will become a new original source for the docs, and will need to be edited directly to change them. Is there a preferred location/format for this docs source?

A few options for location are:
- Docs stored as string literals in the go code (like the CLI help text). This would be pretty ugly for markdown docs, as the commonly used `` ` `` character breaks a multiline string and can't be escaped.
- Docs stored as many external files (one for each string required in the go code), which are embedded in the go code.
- Docs stored in fewer external files (such as one for system tables, one for functions, etc), which are embedded in the go code and parsed into smaller components.
- Docs stored in the dolthub/docs repo, which are embedded in the go code. This would likely require an additional build step, as //go:embed can only embed local files.

A few options for format are:
- Docs stored as strings which will have minimal processing done on them.
- Docs stored as strings with templating to allow changes in the names of system tables, functions, function args, etc. in the code to apply in the docs.
- Arguments stored in their final JSON format.
- Arguments stored in a non-JSON format and parsed into JSON.

zachmu:
[...]

PR #8739, opened 2025-01-13
Create dolt_help system table

milogreg:
Created the dolt_help system table. This table is meant to store documentation for system tables, procedures, functions, and variables. Currently dolt_help is only populated with documentation for procedures, and only procedures that have equivalent CLI commands.

Part of #7984

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/help_table.go
+++ b/go/libraries/doltcore/sqle/dtables/help_table.go
@@ -1,0 +1,211 @@
+// Copyright 2025 Dolthub, Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+// 	http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dtables
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/dolthub/dolt/go/cmd/dolt/cli"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/dprocedures"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/index"
+	"github.com/dolthub/go-mysql-server/sql"
+	sqlTypes "github.com/dolthub/go-mysql-server/sql/types"
+)
+
+type HelpTable struct {
+	dbName    string
+	tableName string
+}
+
+var HelpTableTypes = []string{
+	"system_table",
+	"procedure",
+	"function",
+	"variable",
+}
+
+// NewHelpTable creates a HelpTable
+func NewHelpTable(_ *sql.Context, dbName, tableName string) sql.Table {
+	return &HelpTable{dbName: dbName, tableName: tableName}
+}
+
+// Name is a sql.Table interface function which returns the name of the table.
+func (ht *HelpTable) Name() string {
+	return ht.tableName
+}
+
+// String is a sql.Table interface function which returns the name of the table.
+func (ht *HelpTable) String() string {
+	return ht.tableName
+}
+
+// Schema is a sql.Table interface function that gets the sql.Schema of the help system table.
+func (ht *HelpTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{
+			Name:           "target",
+			Type:           sqlTypes.TinyText,
+			Source:         ht.tableName,
+			PrimaryKey:     true,
+			DatabaseSource: ht.dbName,
+		},
+		{
+			Name:           "type",
+			Type:           sqlTypes.MustCreateEnumType(HelpTableTypes, sql.Collation_Default),
+			Source:         ht.tableName,
+			PrimaryKey:     false,
+			DatabaseSource: ht.dbName,
+		},
+		{
+			Name:           "short_description",
+			Type:           sqlTypes.LongText,
+			Source:         ht.tableName,
+			PrimaryKey:     false,
+			DatabaseSource: ht.dbName,
+		},
+		{
+			Name:           "long_description",
+			Type:           sqlTypes.LongText,
+			Source:         ht.tableName,
+			PrimaryKey:     false,
+			DatabaseSource: ht.dbName,
+		},
+		{
+			Name:           "arguments",
+			Type:           sqlTypes.JSON,
+			Source:         ht.tableName,
+			PrimaryKey:     false,
+			DatabaseSource: ht.dbName,
+		},
+	}
+}
+
+// Collation implements the sql.Table interface.
+func (ht *HelpTable) Collation() sql.CollationID {
+	return sql.Collation_Default
+}
+
+// Partitions is a sql.Table interface function that returns a partition
+// of the data. Currently the data is unpartitioned.
+func (ht *HelpTable) Partitions(*sql.Context) (sql.PartitionIter, error) {
+	return index.SinglePartitionIterFromNomsMap(nil), nil
+}
+
+// PartitionRows is a sql.Table interface function that gets a row iterator for a partition.
+func (ht *HelpTable) PartitionRows(_ *sql.Context, _ sql.Partition) (sql.RowIter, error) {
+	return NewHelpRowIter(), nil
+}
+
+type HelpRowIter struct {
+	idx  *int
+	rows *[]sql.Row
+}
+
+func NewHelpRowIter() HelpRowIter {
+	idx := 0
+	var nilRows []sql.Row
+	return HelpRowIter{idx: &idx, rows: &nilRows}
+}
+
+// DoltCommand is set in cmd/dolt/dolt.go to avoid circular dependency.
+var DoltCommand cli.SubCommandHandler
+
+func (itr HelpRowIter) Next(_ *sql.Context) (sql.Row, error) {
+	helpRows := *itr.rows
+	if helpRows == nil {
+		var err error
+		helpRows, err = generateProcedureHelpRows(DoltCommand.Name(), DoltCommand.Subcommands)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if *itr.idx >= len(helpRows) {
+		return nil, io.EOF
+	}
+
+	row := helpRows[*itr.idx]
+	(*itr.idx)++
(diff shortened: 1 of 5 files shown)


go/libraries/doltcore/sqle/dtables/help_table.go
```diff
@@ -0,0 +1,210 @@
...
+					hasProcedure = true
+					break
+				}
+			}
+
+			docs := curr.Docs()
+
+			if hasProcedure && docs != nil {
+				argsMap := map[string]string{}
+				for _, argHelp := range curr.Docs().ArgParser.ArgListHelp {
```
zachmu, line 189: This is missing all the options (flags)

go/libraries/doltcore/sqle/dtables/help_table.go
```diff
@@ -0,0 +1,210 @@
...
+// String is a sql.Table interface function which returns the name of the table.
+func (ht *HelpTable) String() string {
+	return ht.tableName
+}
+
+// Schema is a sql.Table interface function that gets the sql.Schema of the help system table.
+func (ht *HelpTable) Schema() sql.Schema {
+	return []*sql.Column{
+		{
+			Name:           "target",
```
zachmu, line 61: Would call this `name` instead

go/libraries/doltcore/sqle/dtables/help_table.go
```diff
@@ -0,0 +1,210 @@
...
+			DatabaseSource: ht.dbName,
+		},
+		{
+			Name:           "type",
+			Type:           sqlTypes.MustCreateEnumType(HelpTableTypes, sql.Collation_Default),
+			Source:         ht.tableName,
+			PrimaryKey:     false,
+			DatabaseSource: ht.dbName,
+		},
+		{
```
zachmu, line 74: Probably want a `synopsis` column in here for example usage, like the CLI help

zachmu (commented):
This is the right idea, but it's missing a few things. Check out the difference between the command line and sql output here:

```sql
select arguments from dolt_help where target = 'dolt_add';
+-------------------------------------------------------------------------------------------------------------------------------------+
| arguments                                                                                                                           |
+-------------------------------------------------------------------------------------------------------------------------------------+
| {"table": "Working table(s) to add to the list tables staged to be committed. The abbreviation '.' can be used to add all tables."} |
+-------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

v. command line:

```bash
% dolt add --help
NAME
        dolt add - Add table contents to the list of staged tables

SYNOPSIS
        dolt add [<table>...]

DESCRIPTION

        This command updates the list of tables using the current content found in the working root, to prepare the content staged for the next commit. It adds the current content of existing
        tables as a whole or remove tables that do not exist in the working root anymore.

        This command can be performed multiple times before a commit. It only adds the content of the specified table(s) at the time the add command is run; if you want subsequent changes
        included in the next commit, then you must run dolt add again to add the new content to the index.

        The dolt status command can be used to obtain a summary of which tables have changes that are staged for the next commit.

OPTIONS
        <table>
          Working table(s) to add to the list tables staged to be committed. The abbreviation '.' can be used to add all tables.

        -A, --all
          Stages any and all changes (adds, deletes, and modifications) except for ignored tables.

        -f, --force
          Allow adding otherwise ignored tables.

        -p, --patch
          Interactively select changes to add to the staged set.
```

So relative to the CLI, this is missing:

* options (flags)
* synopsis
* `<table>` formatting for arguments

go/libraries/doltcore/sqle/dtables/help_table.go
```diff
@@ -0,0 +1,232 @@
...
+
+			docs := curr.Docs()
+
+			if hasProcedure && docs != nil {
+				argsMap := map[string]string{}
+				for _, usage := range cli.OptionsUsageList(docs.ArgParser) {
+					argsMap[usage[0]] = usage[1]
+				}
+
+				argsJson, err := json.Marshal(argsMap)
```
zachmu, line 197: I'm not sure why, but something about this seems to be producing invalid json in a subtle and weird way. Check this out:

```sql
db1/main> select json_pretty(arguments) from dolt_help where name = 'dolt_revert';
invalid data type for JSON data in argument 1 to function json_pretty; a JSON string or JSON type is required

db1/main> create table json_test (j json);
db1/main*> insert into json_test (select arguments from dolt_help where name = 'dolt_revert');
db1/main*> select json_pretty(j) from json_test;
+-----------------------------------------------------------------------------------------------------------------------------------+
| json_pretty(j)
  |
+-----------------------------------------------------------------------------------------------------------------------------------+
| {
  |
|   "--author=\u003cauthor\u003e": "Specify an explicit author using the standard A U Thor \u003cauthor@example.com\u003e format.", |
|   "\u003crevision\u003e": "The commit revisions. If multiple revisions are given, they're applied in the order given."
  |
| }
  |
+-----------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

So whatever it is, it's getting handled by our round-trip into storage, but can't get processed by as returned by this table. This seems to affect all our json functions, so it needs to be fixed. Another example:

```sql
db1/main*> select arguments->>"$.<revision>" from dolt_help where name = 'dolt_revert';
invalid data type for JSON data in argument 1 to function json_extract; a JSON string or JSON type is required

db1/main*> select j->>"$.<revision>" from json_test;
+--------------------------------------------------------------------------------------------+
| j->>"$.<revision>"                                                                         |
+--------------------------------------------------------------------------------------------+
| The commit revisions. If multiple revisions are given, they're applied in the order given. |
+--------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

zachmu: I think it has something to do with the `<` and `>` characters. I would expect json.Marshall to do something reasonable there but it apparently is not.

go/libraries/doltcore/sqle/dtables/help_table.go
```diff
@@ -0,0 +1,232 @@
...
+				if err != nil {
+					return nil, err
+				}
+
+				synopsis, err := docs.GetSynopsis(cli.CliFormat)
```
zachmu, line 202: This is fun: something in this GetSynopsis call seems to be maintaining state between calls:

```sql
% dolt sql
# Welcome to the DoltSQL shell.
# "exit" or "quit" (or Ctrl-D) to exit. "\help" for help.
db1/main*> select synopsis from dolt_help where name = 'dolt_branch';
+-----------------------------------------------+
| synopsis                                      |
+-----------------------------------------------+
| dolt branch [--list] [-v] [-a] [-r]           |
| dolt branch [-f] <branchname> [<start-point>] |
| dolt branch -m [-f] [<oldbranch>] <newbranch> |
| dolt branch -c [-f] [<oldbranch>] <newbranch> |
| dolt branch -d [-f] [-r] <branchname>...      |
+-----------------------------------------------+
1 row in set (0.00 sec)

db1/main*> select synopsis from dolt_help where name = 'dolt_branch';
+-----------------------------------------------------------+
| synopsis                                                  |
+-----------------------------------------------------------+
| dolt branch dolt branch [--list] [-v] [-a] [-r]           |
| dolt branch dolt branch [-f] <branchname> [<start-point>] |
| dolt branch dolt branch -m [-f] [<oldbranch>] <newbranch> |
| dolt branch dolt branch -c [-f] [<oldbranch>] <newbranch> |
| dolt branch dolt branch -d [-f] [-r] <branchname>...      |
+-----------------------------------------------------------+
1 row in set (0.00 sec)

db1/main*> select synopsis from dolt_help where name = 'dolt_branch';
+-----------------------------------------------------------------------+
| synopsis                                                              |
+-----------------------------------------------------------------------+
| dolt branch dolt branch dolt branch [--list] [-v] [-a] [-r]           |
| dolt branch dolt branch dolt branch [-f] <branchname> [<start-point>] |
| dolt branch dolt branch dolt branch -m [-f] [<oldbranch>] <newbranch> |
| dolt branch dolt branch dolt branch -c [-f] [<oldbranch>] <newbranch> |
| dolt branch dolt branch dolt branch -d [-f] [-r] <branchname>...      |
+-----------------------------------------------------------------------+
1 row in set (0.00 sec)
```

Please add a test to make sure that successive select to `dolt_help` don't do this.

zachmu (commented):
This looks great with the exception of a couple weird bugs I found during testing.

The other high level comment is you should add a couple spot check tests in the `enginetest` package. Add a new method in `dolt_engine_test.go` that verifies the output of a couple commands (and verifies the synopsis behavior bug below isn't there anymore).

Fix those things and you're good to merge this.

go/libraries/doltcore/sqle/enginetest/dolt_queries_help.go
```diff
@@ -0,0 +1,231 @@
...
+			},
+		},
+	},
+
+	{
+		Name:        "dolt_help long descriptions are correct",
+		SetUpScript: []string{},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				Query: "select long_description from dolt_help where name='dolt_add'",
```
zachmu, line 150: This test is probably over specified -- we tweak these pretty often, don't want to check the whole thing.

I would probably do something like `select INSTR(long_description, "add") > 0`, the idea is just a spot test to be sure this returns the correct contents, not to check the contents verbatim.

go/libraries/doltcore/sqle/enginetest/dolt_queries_help.go
```diff
@@ -0,0 +1,231 @@
...
+					{`Removes the changes made in a commit (or series of commits) from the working set, and then automatically commits the result. This is done by way of a three-way merge. Given a specific commit (e.g. <b>HEAD~1</b>), this is similar to applying the patch from <b>HEAD~1..HEAD~2</b>, giving us a patch of what to remove to effectively remove the influence of the specified commit. If multiple commits are specified, then this process is repeated for each commit in the order specified. This requires a clean working set.
+
+Any conflicts or constraint violations caused by the merge cause the command to fail.`},
+				},
+			},
+		},
+	},
+
+	{
+		Name:        "dolt_help arguments are correct",
```
zachmu, line 184: Same comment here, loosen up these test to make them less likely to break when we change things in the future.

zachmu: What you have in the bats tests is much closer to what we want here.

zachmu (commented):
This is looking pretty good. One major comment about this, which is that we need configurable format options for the `OptionUsageList` method.

Fix that and we're good to merge.

zachmu:
Also fix the copyright header on your new file, it needs to match other files exactly.

zachmu (approved):
LGTM, nice work

zachmu:
I'll merge this assuming CI passes (ignoring the two checks that can't run)

PR #8746, opened 2025-01-14
Allow `dolt sql` to always log in as the `root` superuser

fulghum:
From the command line, when a sql-server is not running, `dolt sql` implicitly uses the `root` account to log, but if the `root` account exists with a password, `dolt sql` will fail to log in. Since the user has access to the host and the database data directory, we should allow `dolt sql` to log into the SQL shell, even if the `root` user has a password set. This change also makes this behavior match when a sql-server is running, and we allow superuser login through the `__dolt_local_user__` account (which only exists while a sql-server is running).

### DIFF
--- a/go/cmd/dolt/commands/utils.go
+++ b/go/cmd/dolt/commands/utils.go
@@ -287,20 +287,9 @@ func newLateBindingEngine(
 			}
 
 		} else {
+			// Ensure a root user exists, with superuser privs
 			dbUser = DefaultUser
 			ed := rawDb.Editor()
-			user := rawDb.GetUser(ed, dbUser, config.ServerHost, false)
-			ed.Close()
-			if user != nil {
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>17c65a7</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>17c65a7</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4 (approved):

PR #8752, opened 2025-01-15
go/store/{nbs,types}: GC: Move the reference walk from types to nbs.

reltuk:
Make the ChunkStore itself responsible for the reference walk, being given handles for walking references and excluding chunks as part of the GC process. This is an incremental step towards adding dependencies on read chunks during the GC process. The ChunkStore can better distinguish whether the read is part of the GC process itself or whether it came from the application layer. It also allows better management of cache impact and the potential for better memory usage.

This transformation gets rid of parallel reference walking and some manual batching which was present in the ValueStore implementation of reference walking. The parallel reference walking was necessary for reasonable performance in format __LD_1__, but it's actually not necessary in __DOLT__. For some use cases it's a slight win, but the simplification involved in getting rid of it is worth it for now.


go/store/nbs/store.go
```diff
@@ -1647,18 +1647,127 @@ func markAndSweepChunks(ctx context.Context, hashes <-chan []hash.Hash, nbs *Nom
 		destNBS = nbs
 	}
 
-	specs, err := copyMarkedChunks(ctx, hashes, src, destNBS)
+	tfp, ok := destNBS.p.(tableFilePersister)
+	if !ok {
+		return nil, fmt.Errorf("NBS does not support copying garbage collection")
+	}
+
+	gcc, err := newGarbageCollectionCopier()
```
max-hoffman, line 1655: somewhat unrelated, but if this embedded `tfp` their relationship might be clearer

go/store/nbs/store.go
```diff
@@ -1647,18 +1647,127 @@ func markAndSweepChunks(ctx context.Context, hashes <-chan []hash.Hash, nbs *Nom
...
+		gcc:      gcc,
+		mode:     mode,
+	}, nil
+}
+
+type markAndSweeper struct {
+	src      NBSCompressedChunkStore
+	dest     *NomsBlockStore
+	getAddrs chunks.GetAddrsCurry
+	filter   chunks.HasManyFunc
```
max-hoffman, line 1676: i'm hazy on what filter does, when would we discard hashes?

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f5a69f5</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f5a69f5</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/store_test.go
```diff
@@ -334,14 +334,18 @@ func TestNBSCopyGC(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, ok)
 
-	keepChan := make(chan []hash.Hash, numChunks)
+	require.NoError(t, st.BeginGC(nil))
```
max-hoffman, line 337: is our GC testing this sparse? or do we have tests at other interface levels somewhere else

max-hoffman (approved):
LGTM, the simplification is nice just a few related questions I noticed while getting up to speed

go/store/nbs/store.go
reltuk: Great question. It's used for generational GC. So, when we collect newgen -> oldgen, we're walking refs and we want to stop the walk anytime we walk into the old gen. Then, after those chunks are in the old gen, when we collect newgen -> newgen, we want to stop the walk once again anytime we walk into the old gen.

go/store/nbs/store.go
reltuk: Good suggestion! I'll take a pass and potentially send out a separate PR :)

PR #8760, opened 2025-01-16
[no-release-notes] go/store/nbs: During a GC process, take dependencies on chunks that are read through the ChunkStore.

reltuk:
Previously GC was constructed to walk the transitive closure of reachable chunks from when it started, and to take additional dependencies on any chunks that were written to the ChunkStore during the GC process. This change makes it so that we can additionally take dependencies on any chunks that are read from the ChunkStore during the GC process.

For the `nbs` layer to signal a new dependency during a GC, it makes use of a `keeperFunc`, which is passed into the `BeginGC` call. The contract is that, if `keeperFunc` is non-`nil` and it returns `false`, the chunk address which was given to the function will make its way back to a call to `SaveHashes` on some `MarkAndSweeper` whose contents  will end up in the final store once the GC is completed. `keeperFunc`, however, is allowed to return `true`, which signals to the block store that the in-progress GC has passed a certain point in time and can no longer make that guarantee. As a result, anytime `keeperFunc` returns `true`, the block store implementation needs to block until the GC process is over, and then it should resume the operation from the beginning.

That machinery already exists for the `Put()` and `Commit()` use case. This PR adds machinery to do all of that on various read paths that explicitly touch chunks, such as `Has`, `HasMany`, `Get`, `GetMany` and `GetManyCompressed`. In order to add dependency tracking on the read path, it's important for the GC process itself to not cause the chunks it has read to form additional dependencies. This PR makes a slight change to `markAndSweeper` so that it can fetch chunks from the source store without recording unnecessary duplicate dependencies and potentially becoming mired in deadlocks.

### DIFF
--- a/go/store/nbs/generational_chunk_store.go
+++ b/go/store/nbs/generational_chunk_store.go
@@ -145,14 +145,18 @@ func (gcs *GenerationalNBS) GetMany(ctx context.Context, hashes hash.HashSet, fo
 }
 
 func (gcs *GenerationalNBS) GetManyCompressed(ctx context.Context, hashes hash.HashSet, found func(context.Context, CompressedChunk)) error {
+	return gcs.getManyCompressed(ctx, hashes, found, gcDependencyMode_TakeDependency)
+}
+
+func (gcs *GenerationalNBS) getManyCompressed(ctx context.Context, hashes hash.HashSet, found func(context.Context, CompressedChunk), gcDepMode gcDependencyMode) error {
 	var mu sync.Mutex
 	notInOldGen := hashes.Copy()
-	err := gcs.oldGen.GetManyCompressed(ctx, hashes, func(ctx context.Context, chunk CompressedChunk) {
+	err := gcs.oldGen.getManyCompressed(ctx, hashes, func(ctx context.Context, chunk CompressedChunk) {
 		mu.Lock()
 		delete(notInOldGen, chunk.Hash())
 		mu.Unlock()
 		found(ctx, chunk)
-	})
+	}, gcDepMode)
 	if err != nil {
 		return err
 	}
@@ -161,12 +165,12 @@ func (gcs *GenerationalNBS) GetManyCompressed(ctx context.Context, hashes hash.H
 	}
 
 	notFound := notInOldGen.Copy()
-	err = gcs.newGen.GetManyCompressed(ctx, notInOldGen, func(ctx context.Context, chunk CompressedChunk) {
+	err = gcs.newGen.getManyCompressed(ctx, notInOldGen, func(ctx context.Context, chunk CompressedChunk) {
 		mu.Lock()
 		delete(notFound, chunk.Hash())
 		mu.Unlock()
 		found(ctx, chunk)
-	})
+	}, gcDepMode)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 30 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e423a5c</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e423a5c</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6ad9b37</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6ad9b37</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5c04d5f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5c04d5f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/generational_chunk_store.go
```diff
@@ -202,14 +208,30 @@ func (gcs *GenerationalNBS) Has(ctx context.Context, h hash.Hash) (bool, error)
 }
 
 // HasMany returns a new HashSet containing any members of |hashes| that are absent from the store.
-func (gcs *GenerationalNBS) HasMany(ctx context.Context, hashes hash.HashSet) (absent hash.HashSet, err error) {
-	gcs.newGen.mu.RLock()
-	defer gcs.newGen.mu.RUnlock()
-	return gcs.hasMany(toHasRecords(hashes))
+func (gcs *GenerationalNBS) HasMany(ctx context.Context, hashes hash.HashSet) (hash.HashSet, error) {
+	absent, err := gcs.newGen.HasMany(ctx, hashes)
```
max-hoffman, line 212: is the diff just from collapsing `gcs.getMany`? or is there more to this?

go/store/nbs/store.go
```diff
@@ -171,22 +188,60 @@ func (nbs *NomsBlockStore) GetChunkLocations(ctx context.Context, hashes hash.Ha
...
+		}
+		return false, err
+	}
+	if gcb == gcBehavior_Block {
+		nbs.mu.Lock()
+		if endRead != nil {
+			endRead()
+		}
+		err := nbs.waitForGC(ctx)
+		nbs.mu.Unlock()
```
max-hoffman, line 235: i'm sure this is intentional, but is it better to block here than let other sessions try to do partial work before blocking on the same wait? ex: marked chunks cache hits under this miss

max-hoffman (approved):
LGTM, just the one question about locking store access on block

PR #8783, opened 2025-01-22
Avoid loading DB for commands where it's not necessary.

nicktobey:
This PR avoids loading dolt DBs at startup, instead waiting for either `DoltEnv.ReloadDB` or `MultiRepoEnv.ReloadDBs` to be called.

Based on the specific subcommand being run:
- `MultiRepoEnv.ReloadDBs` is called prior to command execution if the command will always need DB access.
- Commands that conditionally require a DB can choose whether to call `DoltEnv.ReloadDB`
- Commands that never need the DB will never load it.

Commands that make use of `CliContext` don't need to load the DB, as the underlying query engine will load the DB if it needs database access. As a result, commands that are implemented purely in terms of the `CliContext` will load the DB when Dolt is runnning an embedded SQL engine, and will skip loading the DB when Dolt is connecting to a running server.


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5355928</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5355928</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4 (changes requested):

go/libraries/doltcore/env/environment.go
```diff
@@ -97,6 +97,34 @@ type DoltEnv struct {
...
+		doltDB:    doltDB,
+		FS:        fs,
+	}
+}
+
+func (dEnv *DoltEnv) DoltDB(ctx context.Context) *doltdb.DoltDB {
+	if dEnv.doltDB == nil {
+		LoadDoltDB(ctx, dEnv.FS, dEnv.urlStr, dEnv)
+	}
+	return dEnv.DoltDB(ctx)
```
macneale4, line 114: huh? Why doesn't the recurrse forever? shouldn't it be `return dEnv.doltDb`

go/libraries/doltcore/env/multi_repo_env.go
```diff
@@ -226,6 +224,31 @@ func MultiEnvForDirectory(
 	return mrEnv, nil
 }
 
+func (mrEnv *MultiRepoEnv) ReloadDBs(
+	ctx context.Context,
+) {
+	for _, namedEnv := range mrEnv.envs {
+		dEnv := namedEnv.env
+
+		if dEnv.DoltDB(ctx) == nil {
```
macneale4, line 233: When does this return nil? I'm not really sure what the DoltDb function is supposed to do, but my expectation would be that the LoadDoltDB call in there would succeed and the DoltDB function would return a non-nil value always

go/libraries/doltcore/merge/schema_integration_test.go
```diff
@@ -541,12 +541,12 @@ func newColTypeInfo(name string, tag uint64, typeInfo typeinfo.TypeInfo, partOfP
...
 }
 
-func fkCollection(fks ...doltdb.ForeignKey) *doltdb.ForeignKeyCollection {
-	fkc, err := doltdb.NewForeignKeyCollection(fks...)
-	if err != nil {
-		panic(err)
-	}
-	return fkc
+func fkCollection(fks ...DoltDB(ctx).ForeignKey) *doltdb.ForeignKeyCollection {
+fkc, err := doltdb.NewForeignKeyCollection(fks...)
```
macneale4, line 545: what the heck is happening with the formatter here?

go/libraries/doltcore/migrate/integration_test.go
```diff
@@ -217,41 +217,41 @@ func runMigration(t *testing.T, ctx context.Context, preEnv *env.DoltEnv) (postE
...
 }
 
 func initTestMigrationDB(ctx context.Context) (*doltdb.DoltDB, error) {
-	var db datas.Database
-	storage := &chunks.MemoryStorage{}
-	cs := storage.NewViewWithFormat("__DOLT__")
-	vrw := types.NewValueStore(cs)
-	ns := tree.NewNodeStore(cs)
-	db = datas.NewTypesDatabase(vrw, ns)
+var db datas.Database
```
macneale4, line 229: formatting again? Not sure why this is happening

go/libraries/doltcore/sqle/dsess/session.go
```diff
@@ -866,7 +866,7 @@ func (d *DoltSession) ReleaseSavepoint(ctx *sql.Context, tx sql.Transaction, sav
 	return nil
 }
 
-// GetDoltDB returns the *DoltDB for a given database by name
+// GetDoltDB returns the *doltDB for a given database by name
```
macneale4, line 874: revert

macneale4 (changes requested):
still a few questions. Overall, this is definitely an improvement.

Also, looks like all the tests are failing. Maybe because of the infinite recurssion?

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f78cc0d</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f78cc0d</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8b10e7f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8b10e7f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

integration-tests/bats/remotes-file-system.bats
```diff
@@ -176,3 +176,23 @@ SQL
...
 }
+
+@test "remotes-file-system: disallow cloning directly from a repo" {
+    mkdir repo1 repo2
+    cd repo1
+    dolt init
+    cd ../repo2
+    run dolt clone file://../repo1/.dolt/noms repo2
+    [ "$status" -eq 1 ]
+    [[ "$output" =~ "cannot create NBS store for directory containing chunk journal" ]] || false
```
macneale4, line 187: Why is this changing? Looks like we are taking away functionality verified in replication.bats

go/libraries/doltcore/env/environment.go
```diff
@@ -164,16 +194,21 @@ func LoadWithoutDB(_ context.Context, hdp HomeDirProvider, fs filesys.Filesys, v
...
+	dEnv := LoadWithoutDB(ctx, hdp, fs, urlStr, version)
+	LoadDoltDB(ctx, dEnv.FS, dEnv.urlStr, dEnv)
+	return dEnv
+}
 
+func LoadDoltDB(ctx context.Context, fs filesys.Filesys, urlStr string, dEnv *DoltEnv) {
 	ddb, dbLoadErr := doltdb.LoadDoltDB(ctx, types.Format_Default, urlStr, fs)
 
-	dEnv.DoltDB = ddb
+	dEnv.doltDB = ddb
```
macneale4, line 211: Given that we are lazy loading the db, there are probably race conditions here, right?

I'm pretty sure you want a "Once" here, but you could also put a mutex around it, since ReloadDB may violate the once approach

go/libraries/doltcore/sqle/testutil.go
```diff
@@ -103,9 +103,13 @@ func ExecuteSql(ctx context.Context, dEnv *env.DoltEnv, root doltdb.RootValue, s
 		}
 	}
 
-	err = dsess.DSessFromSess(sqlCtx.Session).CommitTransaction(sqlCtx, sqlCtx.GetTransaction())
-	if err != nil {
-		return nil, err
+	// commit leftover transaction
+	trx := sqlCtx.GetTransaction()
```
macneale4, line 107: Why does this need to change?

macneale4 (changes requested):
Only remaining concern is that we'll have multiple threads initializing the DB at the same time. Convince me that's not the case, or guard against it.

integration-tests/bats/remotes-file-system.bats
nicktobey: Basically, I checked with Aaron and this behavior _isn't_ supported, because we don't currently do it correctly (we don't account for the journal file or new-generation table tables). There are supposed to be checks against it, but those checks were inadvertently being skipped if the repo being cloned was already loaded into the dEnv.

By deferring the load, we now correctly identify that cloning this way is not safe and prevent it.

The removed test only happens to behave correctly because the new-generation is empty.

go/libraries/doltcore/env/environment.go
nicktobey: Good catch. Added.

go/libraries/doltcore/sqle/testutil.go
nicktobey: This is odd. This change comes from https://github.com/dolthub/dolt/pull/8787, which has already been merged into main.

I merged from main again and this diff disappears.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>d27b998</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>d27b998</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b1aa361</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b1aa361</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>2911085</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>2911085</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4 (approved):
LGTM! There is a Doltgres integration test failing which you should fix before you ship it.

PR #8796, opened 2025-01-28
[no-release-notes] go/store/nbs: generational_chunk_store.go: In GCMode_Full, also take dependencies on chunks read from OldGen.

reltuk:


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>1b59420</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>1b59420</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

max-hoffman (approved):
LGTM. Just so I'm following, the case where this would be useful is something like, the rootvalue does not reach certain chunks in oldgen, but there is a commit in progress who's new rootvalue does ref an oldgen chunk, and we want to include that refs oldgen closure in the new table file rather that GC'ing it and leaving the new root hanging. And this only matters for a full GC because that's the only case the old oldgen tablefiles are removed.

reltuk:
> LGTM. Just so I'm following, the case where this would be useful is something like, the rootvalue does not reach certain chunks in oldgen, but there is a commit in progress who's new rootvalue does ref an oldgen chunk, and we want to include that refs oldgen closure in the new table file rather that GC'ing it and leaving the new root hanging. And this only matters for a full GC because that's the only case the old oldgen tablefiles are removed.

Totally correct understanding. The theory is some inflight work is going to take a dep on oldgen, but the roots we saw when we started the GC don't reach those chunks.

PR #8824, opened 2025-02-05
[no-release-notes] go: nbs,chunks: Have AddTableFilesToManifest check all dependencies are present.

reltuk:


go/store/nbs/journal.go
```diff
@@ -196,7 +195,7 @@ func trueUpBackingManifest(ctx context.Context, root hash.Hash, backing *journal
 	if err != nil {
 		return manifestContents{}, err
 	} else if !ok {
-		return manifestContents{}, fmt.Errorf("manifest not found when opening chunk journal")
+		return manifestContents{}, nil
```
reltuk, line 198: We changed the semantics here so that we can open a journal file without a manifest already existing.

We do this as part of `Clone` in some cases, where we copy the journal from a remote into our local store, and then we AddTableFilesToManifest to add it to our manifest. The new sanity checking logic wants to open the journal and read its chunks before we agree to add it to the store. This change allows us to do that.

The change below to `ParseIfExists` stitches it together, by continuing to go to the backing manifest as long as we have never loaded an actual manifest here.

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c07900c</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c07900c</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/remotesrv/grpc.go
```diff
@@ -654,6 +654,18 @@ func (rs *RemoteChunkStore) AddTableFiles(ctx context.Context, req *remotesapi.A
 	return &remotesapi.AddTableFilesResponse{Success: true}, nil
 }
 
+func (rs *RemoteChunkStore) getAddrs(version string) chunks.GetAddrsCurry {
```
zachmu, line 657: Could use a method doc

go/store/datas/pull/puller.go
```diff
@@ -101,12 +101,25 @@ func NewPuller(
 		return nil, ErrIncompatibleSourceChunkStore
 	}
 
+	// walkAddrs can be used for getAddrs on the AddTableFile call as well.
+	getAddrs := func(c chunks.Chunk) chunks.GetAddrsCb {
+		return func(ctx context.Context, ins hash.HashSet, filter chunks.PendingRefExists) error {
+			return walkAddrs(c, func(h hash.Hash, _ bool) error {
+				if !filter(h) {
```
zachmu, line 108: Probably too difficult to change but this is the opposite of the usual filter idiom

go/store/nbs/journal.go
zachmu: Probably worth a comment to that effect

go/store/nbs/store.go
```diff
@@ -273,12 +272,13 @@ func (nbs *NomsBlockStore) conjoinIfRequired(ctx context.Context) (bool, error)
...
-		return
+		return manifestContents{}, err
 	}
-	return nbs.updateManifestAddFiles(ctx, updates, nil)
+	mi, _, err = nbs.updateManifestAddFiles(ctx, updates, nil, nil, nil)
+	return mi, err
 }
 
-func (nbs *NomsBlockStore) updateManifestAddFiles(ctx context.Context, updates map[hash.Hash]uint32, appendixOption *ManifestAppendixOption) (mi ManifestInfo, err error) {
+func (nbs *NomsBlockStore) updateManifestAddFiles(ctx context.Context, updates map[hash.Hash]uint32, appendixOption *ManifestAppendixOption, gcGen *hash.Hash, sources chunkSourceSet) (mi ManifestInfo, gcGenDifferent bool, err error) {
```
zachmu, line 281: Is it worth defining a type alias for gcGen? Probably useful to distinguish from other random hashes

go/store/nbs/table_set.go
```diff
@@ -425,9 +425,63 @@ func (ts tableSet) flatten(ctx context.Context) (tableSet, error) {
...
+			if err != nil {
+				cleanup()
+				return nil, err
+			}
+			ret[h] = cloned
+		}
+	}
+	eg, ctx := errgroup.WithContext(ctx)
+	var mu sync.Mutex
+	for h, c := range files {
```
zachmu, line 462: I'm unfamiliar with the semantics here so it's not obvious what value c represents or how it feeds into the Open call below

zachmu (approved):
LGTM as far as I understand this layer. Might want somebody who has been in here more recently take a pass as well.

go/store/nbs/store.go
```diff
@@ -1635,26 +1640,154 @@ func (nbs *NomsBlockStore) WriteTableFile(ctx context.Context, fileId string, nu
...
+		for _, r := range recs {
+			if !r.has {
+				ret.Insert(*r.a)
+			}
+		}
+		return ret, nil
+	}
+}
+
+// For each chunk in sources, walk all its addresses with |getAddrs| and ensure that none are missing from a call to |refCheck|.
```
max-hoffman, line 1684: so basically, the chunkstores are self-consistent and do not include any intermediate nodes with missing refs?

go/store/nbs/store.go
```diff
@@ -1635,26 +1640,154 @@ func (nbs *NomsBlockStore) WriteTableFile(ctx context.Context, fileId string, nu
...
+		// time so that we can ensure our view of the store is still
+		// accurate when we add these files to the store if all the sanity
+		// checks pass.
+		chunkSources, gcGen, err := nbs.openChunkSourcesForAddTableFiles(ctx, fileIdHashToNumChunks)
+		if err != nil {
+			return err
+		}
+		// If these sources get added to the store, they will get cloned.
+		// Either way, we want to close these instances when we are done.
+		defer chunkSources.close()
```
max-hoffman, line 1750: does the newest go version do this correctly? or does it just try to close the last very n-times?

go/store/nbs/store.go
```diff
@@ -1635,26 +1640,154 @@ func (nbs *NomsBlockStore) WriteTableFile(ctx context.Context, fileId string, nu
...
+			// A gcGenMismatch means that the store has changed out from under
+			// us as we were running these checks. We want to retry.
+			continue
+		}
+
+		// At this point, the files have been added to the
+		// store and the store was in a consistent state when
+		// we added them.
+		//
+		// We currently do not take any chunk dependencies
```
max-hoffman, line 1774: GC is irrelevant here because we're checking the self-consistency of whole file, which additively will be consistent w/ GC results?

max-hoffman (commented):
Looks good, I don't fully understand the different clone paths and the process for incorporating a server's journal file into another's. Like does this just prevent cloning a corrupt journal file? Also do we just dump the journal file into `noms/` to access in the future or do we reformat it?

PR #8852, opened 2025-02-11
Split DB loading into two stages.

nicktobey:
This is a follow up to https://github.com/dolthub/dolt/pull/8783, which was supposed to defer loading the DB for commands where it wasn't necessary. The main motivation was for situations where the Dolt CLI doesn't need to load a local DB because it connects to an already-running server instead.

In practice, the prior PR didn't actually accomplish much because in order to determine whether Dolt should connect to an already running server, it checks to see if there's a lock on the local DB, and in order to do that it tries to load it.

In order for CLI commands against an already-running server to avoid loading the DB, we need to either:
- Detect a lock on the DB without loading it
- Look for a ServerLocalCredsFile before determining whether or not to connect to a remote server.

This PR tries to tackle the former, although in hindsight the latter might be a cleaner, simpler fix.

Basically, this PR allows us to detect locks without loading the DB by splitting the DB load into two phases:
Phase 1: Acquire any necessary exclusive resources (like locks)
Phase 2: Finish loading the DB

We accomplish this by modifying the `DBFactory` interface. Previously, it had a `CreateDB` method that fully loads the DB. Now, it has a `GetDBLoader` method that performs the first phase and returns an instance of the `DBLoader` interface. `DBLoader` has a method `LoadDB` that performs the second phase and returns the fully loaded database.

This allows us to add other methods to `DBLoader` to inspect qualities of the DB that were ascertained during the first phase, such as whether or not we were able to acquire an exclusive lock.

Notes:
- This still needs additional tests
- As written, for multigenerational file stores, we only acquire the lock on the newgen during the first phase, and only acquire the oldgen lock during the second phase. I don't think this can lead to any race conditions: both phases are guarded by the same mutex and their results are cached. But it is potentially confusing and it might be better to acquire both file locks at the same time.

### DIFF
--- a/go/libraries/doltcore/dbfactory/file.go
+++ b/go/libraries/doltcore/dbfactory/file.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/dolthub/fslock"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -62,12 +63,14 @@ type FileFactory struct {
 }
 
 type singletonDB struct {
-	ddb datas.Database
-	vrw types.ValueReadWriter
-	ns  tree.NodeStore
+	lockFile *fslock.Lock
+	ddb      datas.Database
+	vrw      types.ValueReadWriter
+	ns       tree.NodeStore
 }
 
 var singletonLock = new(sync.Mutex)
+var singletonFileLocks = make(map[string]*fslock.Lock)
 var singletons = make(map[string]singletonDB)
 
 func CloseAllLocalDatabases() (err error) {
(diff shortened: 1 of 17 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f506927</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f506927</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/dbfactory/file.go
```diff
@@ -192,6 +189,51 @@ func (fact FileFactory) CreateDB(ctx context.Context, nbf *types.NomsBinFormat,
...
+	if params != nil {
+		_, useJournal = params[ChunkJournalParam]
+	}
+
+	var lock *fslock.Lock
+
+	if s, ok := singletonFileLocks[urlObj.Path]; ok {
+		lock = s
+	} else {
+		lock, err = nbs.AcquireManifestLock(path)
```
reltuk, line 220: These lifecycle changes seem weird and complicated and potentially wrong for non-journal stores (if those still exist...in that world the file manifest manager only hold the lock while it makes the switch...). Is there an alternative world where we change CreateDB() so that we can request ExeclusiveAccess and if it can't give it to us it can bail early and we can make lazy queryist use that instead?

reltuk: I guess we're not actually looking for Exclusive. We're looking for non-ReadOnly. Same thought though...

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>2af0836</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>2af0836</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

nicktobey:
With https://github.com/dolthub/dolt/pull/9579 as a workaround, this is no longer necessary.

PR #8890, opened 2025-02-25
feat:(logging) Added Log Format provision

damanV5:
Added `log_format` feature for logger.

### DIFF
--- a/go/libraries/doltcore/servercfg/yaml_config.go
+++ b/go/libraries/doltcore/servercfg/yaml_config.go
@@ -127,6 +127,7 @@ type UserSessionVars struct {
 // YAMLConfig is a ServerConfig implementation which is read from a yaml file
 type YAMLConfig struct {
 	LogLevelStr       *string                `yaml:"log_level,omitempty"`
+	LogFormatStr      *string                `yaml:"log_format,omitempty"`
 	MaxQueryLenInLogs *int                   `yaml:"max_logged_query_len,omitempty"`
 	EncodeLoggedQuery *bool                  `yaml:"encode_logged_query,omitempty"`
 	BehaviorConfig    BehaviorYAMLConfig     `yaml:"behavior,omitempty"`


damanV5:
@zachmu

damanV5:
@timsehn Can you take a look whether I am heading in the right direction?

timsehn:
You are heading in the right direction. It looks like you have most of the places to actually get the value to the server. Now the server needs to read it when it wants to emit a log line and emit the right format. 

@zachmu suggested that logic is here:

https://github.com/dolthub/dolt/blob/main/go/cmd/dolt/commands/sqlserver/server.go#L117

Looking at logrus (https://github.com/sirupsen/logrus), it seems they have JSON formatting capability. IN their README, they suggest

```
log.SetFormatter(&log.JSONFormatter{})
```

timsehn:
That looks like it is working. I'm not sure the best way to write tests for this. @zachmu should answer.

zachmu:
Hi @damanV5,

This looks really good. You should add a test for the new option in the bats file here, similar to this one:

https://github.com/dolthub/dolt/blob/main/integration-tests/bats/sql-server.bats#L123

You can check the output format is correct by redirecting the log to a file, like this test does:

https://github.com/dolthub/dolt/blob/main/integration-tests/bats/sql-server.bats#L1531

lucasfcnunes:
> @timsehn I have configured the logger for log_format but i am not sure how can I test the same with a logger.
> So far,  I am able to write a test-case to verify DefaultLogFormat.
> 
> Update:-
> I have built the go folder locally and tested with `log_format: json` in the yaml. Following is the output:-
> 
> image
> 

@timsehn @damanV5 

Can this very first line too be in json?

damanV5:
@zachmu  I have tried adding the following test case in sql-server.bats:-

```
@test "sql-server: logformats are case insensitive" {
    # assert that logformat on command line is not case sensitive
    cd repo1
    PORT=$( definePORT )
    dolt sql-server --logformat json --port=$PORT --socket "dolt.$PORT.sock" > log.txt 2>&1 &
    SERVER_PID=$!
    wait_for_connection $PORT 8500
    dolt sql -q "show databases;"
    stop_sql_server

    # assert that logformat in yaml config is not case sensitive
    cat >config.yml <<EOF
log_format: text
behavior:
  disable_client_multi_statements: true
listener:
  host: "0.0.0.0"
  port: $PORT
EOF
    dolt sql-server --config ./config.yml --socket "dolt.$PORT.sock" &
    SERVER_PID=$!
    wait_for_connection $PORT 8500
    dolt sql -q "show databases;"
    stop_sql_server
}
```

And used the following command to run the test:-
`bats -f "sql-server: logformats are case insensitive" sql-server.bats`

But, I am getting this error:-
image


Insights on what I did:-

I build dolt server using these commands :- 
cd go  > go build ./cmd/dolt

Executing this command from root:-
./go/dolt sql-server --config config.yaml

damanV5:
Figured something out.

The following test is working fine for insensitivity when checked from command line but from config.yaml it is behaving sensitive. So, that I need to figure out. 

Test:-
```
@test "sql-server: logformats are case insensitive" {
    # assert that logformat on command line is not case sensitive
    cd repo1
    PORT=$( definePORT )
    dolt sql-server --logformat jSon --port=$PORT --socket "dolt.$PORT.sock" > log.txt 2>&1 &
    SERVER_PID=$!
   wait_for_connection $PORT 8500 || { 
    echo "Server failed to start, dumping logs:"
    cat log.txt
    exit 1
}
    dolt sql -q "show databases;"
    stop_sql_server

    # assert that logformat in yaml config is not case sensitive
    cat >config.yml <<EOF
log_format: teXt
behavior:
  disable_client_multi_statements: true
listener:
  host: "0.0.0.0"
  port: $PORT
EOF
    dolt sql-server --config ./config.yml --socket "dolt.$PORT.sock" &
    SERVER_PID=$!
    wait_for_connection $PORT 8500
    dolt sql -q "show databases;"
    stop_sql_server
}
```

timsehn:
Running tests now.

integration-tests/bats/sql-server.bats
```diff
@@ -145,6 +145,32 @@ EOF
     dolt sql -q "show databases;"
     stop_sql_server
 }
+@test "sql-server: logformats are case insensitive" {
```
timsehn, line 148: You need a test here to assert here that the log outputted in json format.

timsehn (changes requested):

timsehn:
Also, you must fix the Go tests this broke.

timsehn:
It's odd that the Go tests passed with No race but failed with race on.

integration-tests/bats/sql-server.bats
damanV5: @timsehn  Added.

damanV5:
@timsehn  I have fixed the go tests for sqlserver.

Do i also need to update `minver_validation.txt` with `LogFormatStr *string 0.0.0 log_format,omitempty` as the header says _"file automatically updated by the release process"._

timsehn:
Running tests again.

integration-tests/bats/sql-server.bats
```diff
@@ -145,6 +145,34 @@ EOF
...
```
timsehn, line 148: I would make this two tests:

1.  Assert the json functionality works
2. Assert log format is case insensitive

The reason to make it two tests is because this is going to be misleading if the json logging fails. The test that fails is going to be about case insensitivity.

integration-tests/bats/sql-server.bats
```diff
@@ -145,6 +145,34 @@ EOF
...
+    cd repo1
+    PORT=$( definePORT )
+    dolt sql-server --logformat jSon --port=$PORT --socket "dolt.$PORT.sock" > log.txt 2>&1 &
+    SERVER_PID=$!
+    wait_for_connection $PORT 8500 
+    dolt sql -q "show databases;"
+    stop_sql_server
+
+     # Assert that log is in JSON format (checking if logs contain `{...}`)
+    grep -q '^{.*}$' log.txt
```
timsehn, line 159: I kinda think this test leaves a lot of leeway to fail.

I think I would prefer to parse the log using a json parser. But maybe this is ok?

timsehn (commented):
Just comments on the tests.

timsehn:
> @timsehn I have fixed the go tests for sqlserver.
> 
> Do i also need to update `minver_validation.txt` with `LogFormatStr *string 0.0.0 log_format,omitempty` as the header says _"file automatically updated by the release process"._ ?

This may be an artifact of our automation. I think we're good. Just factor that test into two tests and I tghink we can take this contribution.

Thanks for the hard work.

integration-tests/bats/sql-server.bats
damanV5: I agree ; have added it as a separate test case.

integration-tests/bats/sql-server.bats
damanV5: I also tried using `jq` but that was not working as expected.

timsehn:
@fulghum will give this a final review and get it merged. Thanks for the contribution.

timsehn (approved):

fulghum:
Thanks for the contribution @damanV5 🎉

PR #8894, opened 2025-02-26
[no-release-notes] go: statspro: auto_refresh: Fix session lifecycle leak in auto refresh thread.

reltuk:
Deferring the SessionCommandEnd calls at the top level inside an infinite loop meant that the session tracker saw these as in progress forever.

### DIFF
--- a/go/libraries/doltcore/sqle/statspro/auto_refresh.go
+++ b/go/libraries/doltcore/sqle/statspro/auto_refresh.go
@@ -16,6 +16,7 @@ package statspro
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


max-hoffman (approved):
looks good, just how do i make sure to port this over to my new branch that has a different lifecycle process? just make sure every sqlCtx when cancelled calls session end?

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c429cc4</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c429cc4</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

reltuk:
> looks good, just how do i make sure to port this over to my new branch that has a different lifecycle process? just make sure every sqlCtx when cancelled calls session end?

So the contract is something like:

1) every `ctx` that makes it to storage should come from the `commands/engine`-aware `ctxFactory`.
2) anytime that `ctx` makes it to storage, the `*dsess.DoltSession` associated with that context should have had `sql.SessionCommandBegin(dsess)` called on it.
3) any `*dsess.DoltSession` that does have `sql.SessionCommandBegin(dsess)` called on it should eventually, and preferably in a very timely manner, have `sql.SessionCommandEnd(dsess)` called on it as soon as its unit of work is done.
4) if a `*dsess.DoltSession` is finished at the application layer, `sql.SessionEnd(dsess)` should be called on it.

By surrounding your usage of storage with these lifecycle callbacks, you guarantee the following:

1) If your command was in progress when GC started, GC will not complete until your session is completed.
2) If your command was in progress when GC started, the session itself will have an opportunity to dump its GC roots into the GC process so that GC will be sure they end up in the collected store.

The behavior of the system if you violate the contract depends on the mode we're in and is currently in flux. As things currently stand:

1) If you use a non-`engine`-aware Context, nothing will seem amiss.
2) If you fail to call `SessionCommandBegin`, interaction with storage will work. The GC process may not visit your session to collect its roots. The GC process may concurrently visit your session when it is not in a quiesced state, creating a race, to collect its roots. The GC process may complete while your unit of work is ongoing, causing chunks on which your session depends to disappear.
3) If you fail to call `SessionCommandEnd`, GC processes may become unable to complete on the server, and will block indefinitely waiting for the session to quiesce and become visitable, which it never will do.
4) If you fail to call `SessionEnd`, you will leak a reference to the session in some book keeping structures.

In paranoid mode, which I'm currently working on:

1) If you use a non-`engine`-aware Context, the storage layer will `panic`.
2) If you fail to call `SessionCommandBegin`, the storage layer will `panic`.
3) If you fail to call `SessionCommandEnd`, the GC process will be unable to complete on the server, but it will complain with a stack trace of where the corresponding `SessionCommandBegin` calls which we are blocked on came from.
4) If you fail to call `SessionEnd`, you will leak a reference to the session in some book keeping structures.

We can revisit 4, but 1-3 are how paranoid mode works today.

PR #8908, opened 2025-02-28
[no-release-notes] integration-tests/bats: garbage_collection.bats: Fix flakiness stemming from times appear in ls output that we use for cmp.

reltuk:
These bats tests perform some GCs and inspect filesystem state to assert that either nothing changed for a noop GC or that things actually did change for a non-noop GC. When the inspect the filesystem state, the `ls` output includes mtimes for the files and directories. `.dolt/stats` actually does, or at least can, change across these runs, and so it is explicitly removed before the comparison is done. That means the mtime on the `.dolt` directory itself changes. These timestamps are at minute resolution, and so the tests normally pass, but if the interactions happen in two separate minutes, they fail.

This change removes the directories themselves from the listing, leaving the files all in place. This asserts that the files' sizes and mtimes haven't (or have) changed across the `dolt gc` runs.

### DIFF
--- a/integration-tests/bats/garbage_collection.bats
+++ b/integration-tests/bats/garbage_collection.bats
@@ -475,16 +475,16 @@ SQL
     cd one/two
     dolt init
     rm -rf .dolt/stats
-    ls -laR > ../../before_gc
+    ls -laR | grep -v '^d' > ../../before_gc
     dolt gc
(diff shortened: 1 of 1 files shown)


macneale4 (approved):
I made a suggestion. Welcome to take it or leave it.

PR #8927, opened 2025-03-04
[no-release-notes] /.github/workflows/cd-push-docker-image.yaml: use gh api

coffeegoddd:
referencing this https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release and this https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows


reltuk (approved):
LGTM

PR #8957, opened 2025-03-07
feat: Added disclaimer for branch checkout

### LINKED ISSUE #6876: Warn users about running `dolt sql -q "call dolt_checkout(...);"`

dds05:
To fix #6876

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
@@ -17,10 +17,11 @@ package dprocedures
 import (
 	"errors"
 	"fmt"
-
+	"os"
+	"strings"
 	"github.com/dolthub/go-mysql-server/sql"
 	"github.com/dolthub/go-mysql-server/sql/types"
-
+	"github.com/fatih/color"
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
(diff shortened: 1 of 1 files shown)


go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -176,6 +178,10 @@ func doDoltCheckout(ctx *sql.Context, args []string) (statusCode int, successMes
 		if err != nil {
 			return 1, "", err
 		}
+		terminal_command:=strings.Join(os_args, " ")
```
timsehn, line 181: I think we need to be a bit more clever here. I think we need to check and make sure that call dolt_checkout() with the -b option is the only sql statement being run.

timsehn (commented):
This needs to be a bit more robust.

Also we need a BATS test in:

https://github.com/dolthub/dolt/blob/main/integration-tests/bats/sql-checkout.bats

Some of those tests will need to be updated because we use `dolt sql -q "call dolt_checkout('-b', ...)"` a bunch of places.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
dds05: @timsehn 
Okay but what if the user calls the command like ` dolt sql --q  "call dolt_checkout('myBranch2');"` i.e. without the `-b` option.
I believe this also needs to be handled.

macneale4 (changes requested):
The stored procedure can't return color escape codes in its warning (or any other message) since ORMs and other tools don't consistently handle them correctly. Remember that this is an programming interface :-)

I assume you want the terminal experience to show the warning, and in color? That should all be handled either in the CLI commands or in the sql shell.

macneale4:
A possible better first step would be: https://github.com/dolthub/dolt/issues/8875

dds05:
@macneale4  So should I remove the color from :-
 `cli.PrintErrln(color.YellowString("WARNING: The checked out branch is only changed for this current session."))`

macneale4:
I'm going to close this PR since it's gone dormant.

PR #8990, opened 2025-03-16
Remove unnecessary `hashicorp/go-uuid` dependency

Juneezee:
We are already using `github.com/google/uuid` for UUIDs. I think this dependency was accidentally introduced.

https://github.com/dolthub/dolt/blob/b7c0ddb4aac3702114c4e060651007d590a8be8c/go/go.mod#L24

https://github.com/dolthub/dolt/blob/b7c0ddb4aac3702114c4e060651007d590a8be8c/go/go.mod#L71


macneale4:
Seems reasonable enough. Going to run in through CI with a PR of my own.

PR #8991, opened 2025-03-17
Remove duplicate `github.com/vbauerster/mpb` dependency

Juneezee:
https://github.com/dolthub/dolt/blob/b7c0ddb4aac3702114c4e060651007d590a8be8c/go/cmd/dolt/cli/stdio.go#L28

https://github.com/dolthub/dolt/blob/b7c0ddb4aac3702114c4e060651007d590a8be8c/go/cmd/dolt/cli/stdio_test.go#L23

It's weird that we use the tagged `v8` version in the source code but non-tagged version in the test.

### DIFF
--- a/go/cmd/dolt/cli/stdio_test.go
+++ b/go/cmd/dolt/cli/stdio_test.go
@@ -20,7 +20,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/vbauerster/mpb/cwriter"
 )
 
 func TestEphemeralPrinter(t *testing.T) {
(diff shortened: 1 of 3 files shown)


macneale4:
What's the motivation for removing this? Seems nice to have fewer dependencies, but the code is harder to read and maintain.

go/cmd/dolt/cli/stdio_test.go
```diff
@@ -64,6 +63,8 @@ func TestEphemeralPrinter(t *testing.T) {
 	})
 }
 
+// clearLinesTxt moves cursor up n lines, and erases from cursor position to end of screen.
+// https://github.com/vbauerster/mpb/blob/v8.0.2/cwriter/writer.go#L11-L15
 func clearLinesTxt(n int) string {
-	return fmt.Sprintf("%c[%dA", cwriter.ESC, n) + fmt.Sprintf("%c[J", cwriter.ESC)
+	return fmt.Sprintf("\x1b[%dA\x1b[J", n)
```
Juneezee, line 69: > but the code is harder to read and maintain.

I added 2 lines of comment in case future readers find it hard to understand.

macneale4: yep, I saw that. But that package isn't the spec, so kind of an odd artifact. More appropriate would be this: https://en.wikipedia.org/wiki/ANSI_escape_code#C0_control_codes

But even given that, the code to start with is poor. What does "[%dA" do for example? I already needed to know/read the spec to figure that out. I'd be happy to take a change which makes it overall more readable. The change you proposed doesn't meet that requirement.

go/cmd/dolt/cli/stdio_test.go
```diff
@@ -64,6 +63,15 @@ func TestEphemeralPrinter(t *testing.T) {
...
 func clearLinesTxt(n int) string {
-	return fmt.Sprintf("%c[%dA", cwriter.ESC, n) + fmt.Sprintf("%c[J", cwriter.ESC)
+	// These are ANSI escape codes, see 
+	//  - https://en.wikipedia.org/wiki/ANSI_escape_code#C0_control_codes1
+	//  - https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands
+	//
+	// \x1b: ESC (Escape)
+	// [%dA: CUU (Cursor Up). [5A means moves the cursor up 5 lines
+	// [J  : ED (Erase in Display)
+	return fmt.Sprintf("\x1b[%dA\x1b[J", n)
```
Juneezee, line 76: @macneale4 I have added more comments. Hope you find this version easier to read.

macneale4:
Looks good to me. Gonna run though CI with a pr on my end

macneale4 (approved):

PR #8998, opened 2025-03-18
Replace `golang.org/x/exp/slices` with stdlib `slices`

Juneezee:
The experimental functions in `golang.org/x/exp/slices` are now available in the standard library in Go 1.21.

Reference: https://go.dev/doc/go1.21#slices

### DIFF
--- a/go/cmd/dolt/commands/add.go
+++ b/go/cmd/dolt/commands/add.go
@@ -20,14 +20,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"sort"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


go/cmd/dolt/commands/add.go
```diff
@@ -191,7 +190,7 @@ func patchWorkflow(sqlCtx *sql.Context, queryist cli.Queryist, tables []string)
 		return 0
 	}
 
-	sort.Strings(tables)
+	slices.Sort(tables)
```
Juneezee, line 193: Since Go 1.22, `sort.Strings` simply calls `slices.Sort`

https://pkg.go.dev/sort#Strings

PR #9006, opened 2025-03-19
Replace `jpillora/backoff` with `cenkalti/backoff/v4`

Juneezee:
This is a drop-in replacement.

Both [`github.com/jpillora/backoff`](https://pkg.go.dev/github.com/jpillora/backoff) and [`github.com/cenkalti/backoff/v4`](https://pkg.go.dev/github.com/cenkalti/backoff/v4) provide exponential backoff implementations. Since we already use `github.com/cenkalti/backoff/v4` elsewhere in the codebase, this change standardizes our dependency by using a single implementation consistently.

### DIFF
--- a/go/Godeps/LICENSES
+++ b/go/Godeps/LICENSES
@@ -10943,34 +10943,6 @@ Exhibit B - "Incompatible With Secondary Licenses" Notice
 ================================================================================
 
 ================================================================================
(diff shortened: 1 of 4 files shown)


zachmu (approved):
Seems fine, thanks for the contribution.

PR #9044, opened 2025-03-27
refactor: use the built-in max/min to simplify the code

evenevent:
Use the built-in max/min from the standard library in Go 1.21 to simplify the code.  https://pkg.go.dev/builtin@go1.21.0#max


macneale4 (approved):
LGTM. Running CI

macneale4:
Copied into a PR owned by me: https://github.com/dolthub/dolt/pull/9089 - I have permissions which should help with the CI.

PR #9093, opened 2025-04-08
[no-release-notes] Have TextStorage and ByteArray implement driver.Value

nicktobey:
These types can now get returned by the engine, as alternate representations of text and blob types respectively.

Implementing the `driver.Value` interface ensures that they work correctly with go's SQL drivers.


tbantle22 (approved):
Don't really have any context for these changes but seems reasonable to me

PR #9204, opened 2025-05-12
feat(import,csv,psv): Add support for importing CSV and PSV files without header rows

codeaucafe:
## Summary

- Add --no-header flag to treat the first row in CSV/PSV files as data instead of column names
- Add --columns option to specify column names when importing files without headers
- Fix nil pointer panic when importing from stdin with --create-table

In short, this feature makes Dolt more compatible with MySQL/SQLite workflows and provides users with more flexibility when importing data.

## Problem

Previously, Dolt always expected the first row of CSV/PSV files to contain column names. This differs from MySQL and SQLite which support importing files where the first row contains data. Users migrating from these systems or working with headerless data files couldn't import them without modifying their files.

Additionally, when users attempted to import data from stdin using --create-table, they would encounter a nil pointer panic instead of receiving a error message.

## Solution

The implementation adds:
1. A new --no-header flag that treats the first row as data instead of column headers
2. A complementary --columns option to specify column names when headers aren't present
3. Proper validation to ensure correct flag combinations
4. Comprehensive error handling for stdin imports with clear error messages
5. Integration tests for both CSV and PSV files

## Testing

- Added integration tests for both CSV and PSV files that verify:
- Importing files with --no-header and --columns options
- Error cases when required options are missing
- Original behavior is maintained when not using --no-header
- Behavior of --columns with and without --no-header
- Edge cases like stdin imports

### DIFF
--- a/go/cmd/dolt/commands/tblcmds/import.go
+++ b/go/cmd/dolt/commands/tblcmds/import.go
@@ -73,6 +73,8 @@ const (
 	ignoreSkippedRows = "ignore-skipped-rows" // alias for quiet
 	disableFkChecks   = "disable-fk-checks"
 	allTextParam      = "all-text"
+	noHeaderParam     = "no-header" // for CSV files without header row
+	columnsParam      = "columns"   // for specifying column names
 )
 
 var jsonInputFileHelp = "The expected JSON input file format is:" + `
@@ -114,10 +116,10 @@ During import, if there is an error importing any row, the import will be aborte
 In create, update, and replace scenarios the file's extension is used to infer the type of the file.  If a file does not have the expected extension then the {{.EmphasisLeft}}--file-type{{.EmphasisRight}} parameter should be used to explicitly define the format of the file in one of the supported formats (csv, psv, json, xlsx).  For files separated by a delimiter other than a ',' (type csv) or a '|' (type psv), the --delim parameter can be used to specify a delimiter`,
 
 	Synopsis: []string{
-		"-c [-f] [--pk {{.LessThan}}field{{.GreaterThan}}] [--all-text] [--schema {{.LessThan}}file{{.GreaterThan}}] [--map {{.LessThan}}file{{.GreaterThan}}] [--continue]  [--quiet] [--disable-fk-checks] [--file-type {{.LessThan}}type{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
-		"-u [--map {{.LessThan}}file{{.GreaterThan}}] [--continue] [--quiet] [--file-type {{.LessThan}}type{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
-		"-a [--map {{.LessThan}}file{{.GreaterThan}}] [--continue] [--quiet] [--file-type {{.LessThan}}type{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
-		"-r [--map {{.LessThan}}file{{.GreaterThan}}] [--file-type {{.LessThan}}type{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
+		"-c [-f] [--pk {{.LessThan}}field{{.GreaterThan}}] [--all-text] [--schema {{.LessThan}}file{{.GreaterThan}}] [--map {{.LessThan}}file{{.GreaterThan}}] [--continue] [--quiet] [--disable-fk-checks] [--file-type {{.LessThan}}type{{.GreaterThan}}] [--no-header] [--columns {{.LessThan}}col1,col2,...{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
+		"-u [--map {{.LessThan}}file{{.GreaterThan}}] [--continue] [--quiet] [--file-type {{.LessThan}}type{{.GreaterThan}}] [--no-header] [--columns {{.LessThan}}col1,col2,...{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
+		"-a [--map {{.LessThan}}file{{.GreaterThan}}] [--continue] [--quiet] [--file-type {{.LessThan}}type{{.GreaterThan}}] [--no-header] [--columns {{.LessThan}}col1,col2,...{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
+		"-r [--map {{.LessThan}}file{{.GreaterThan}}] [--file-type {{.LessThan}}type{{.GreaterThan}}] [--no-header] [--columns {{.LessThan}}col1,col2,...{{.GreaterThan}}] {{.LessThan}}table{{.GreaterThan}} {{.LessThan}}file{{.GreaterThan}}",
 	},
 }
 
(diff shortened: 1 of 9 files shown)


timsehn:
running workflows

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -402,6 +441,8 @@ func (cmd ImportCmd) ArgParser() *argparser.ArgParser {
 	ap.SupportsString(fileTypeParam, "", "file_type", "Explicitly define the type of the file if it can't be inferred from the file extension.")
 	ap.SupportsString(delimParam, "", "delimiter", "Specify a delimiter for a csv style file with a non-comma delimiter.")
 	ap.SupportsFlag(allTextParam, "", "Treats all fields as text. Can only be used when creating a table.")
+	ap.SupportsFlag(noHeaderParam, "", "Treats the first row of a CSV file as data instead of a header row with column names.")
+	ap.SupportsString(columnsParam, "", "columns", "Comma-separated list of column names if the file has no header row.")
```
nicktobey, line 445: It looks like if `--columns` is provided *without* `--no-header`, it overrides the column names in the provided file. This should be reflected in the argument description, and we should have tests for that combination.

go/libraries/doltcore/mvdata/file_data_loc.go
```diff
@@ -95,21 +95,54 @@ func (dl FileDataLocation) NewReader(ctx context.Context, dEnv *env.DoltEnv, opt
...
 			if len(csvOpts.Delim) != 0 {
 				delim = csvOpts.Delim
 			}
+
+			if csvOpts.NoHeader {
+				hasHeaderLine = false
+			}
+
+			if len(csvOpts.Columns) > 0 {
+				columns = csvOpts.Columns
```
nicktobey, line 113: It looks like this logic for initializing a CsvOptions struct appears in four places, and is more or less identical each time. We can probably outline this into its own function.

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -248,9 +265,23 @@ func getImportMoveOptions(ctx *sql.Context, apr *argparser.ArgParseResults, dEnv
 			srcLoc = val
 		}
 
+		noHeaderFlag := apr.Contains(noHeaderParam)
```
nicktobey, line 268: This is nearly identical to the added code above: we may be able to move it to before the switch block and just have it run as long as the data location isn't a TableDataLocation.

integration-tests/bats/import-no-header.bats
```diff
@@ -0,0 +1,104 @@
...
+3,Bob,Johnson,42,Chicago
+DELIM
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "import with --no-header and --columns options" {
+    # Test regular import with no header (should use first row as column names)
```
nicktobey, line 20: "(should use first row as column names)"

This comment seems to say the opposite of what it should: if we're passing in the `--columns` option, we *shouldn't* use the first row as column names.

nicktobey (changes requested):
Thank you, David! This is a very clean fix.

I left a couple of comments pertaining to documentation, tests, and code duplication. After that, this should be good to merge.

integration-tests/bats/import-no-header.bats
```diff
@@ -0,0 +1,104 @@
...
+2,Jane,Smith,28,Los Angeles
+3,Bob,Johnson,42,Chicago
+DELIM
+}
+
+teardown() {
+    teardown_common
+}
+
+@test "import with --no-header and --columns options" {
```
nicktobey, line 19: Bats test names are required to begin with the test suite. These all need to begin with "import-no-header: ..."

nicktobey:
Don't worry about the "Benchmark SQL Correctness" check failing: that's a known issue for PRs that come from forks.

If the other checks are green, we'll be good.

integration-tests/bats/import-no-header.bats
codeaucafe: doh; good catch

codeaucafe:
hi @nicktobey I think I got it in a better position, so its "ready for review". FYI I realized trying to use no-headers and columns while streaming in the csv was panicking. I adding error handling where `// todo: capture stream data to file so we can use schema inference` was. Not sure if that's right or not.

thanks for reviewing.

go/libraries/doltcore/mvdata/csv_helper.go
```diff
@@ -0,0 +1,51 @@
...
+package mvdata
+
+import (
+	"github.com/dolthub/dolt/go/libraries/doltcore/table/untyped/csv"
+)
+
+// CreateCSVInfo creates a CSVInfo object based on the provided options.
+// This is a helper function that extracts and processes CSV-related options
+// from the generic options interface.
+func CreateCSVInfo(opts interface{}, defaultDelim string) *csv.CSVFileInfo {
```
codeaucafe, line 24: not sure how y'all feel about this file name. Wasn't sure if I should name it after helper func, or keep it a generic helper file. 🤷‍♂️

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -325,6 +330,14 @@ func validateImportArgs(apr *argparser.ArgParseResults) errhand.VerboseError {
 		return errhand.BuildDError("parameters %s and %s are mutually exclusive", allTextParam, schemaParam).Build()
 	}
 
+	if apr.Contains(noHeaderParam) && !apr.Contains(columnsParam) {
```
codeaucafe, line 333: similar handling in the issues I had testing with std streaming of csv

go/cmd/dolt/commands/tblcmds/import.go
```diff
@@ -317,6 +316,12 @@ func validateImportArgs(apr *argparser.ArgParseResults) errhand.VerboseError {
 		return errhand.BuildDError("fatal: " + schemaParam + " is not supported for update or replace operations").Build()
 	}
 
+	if apr.Contains(createParam) && apr.NArg() <= 1 {
```
codeaucafe, line 319: similar panic handling handling from streaming file. Not sure this is needed. I wasn't sure if I read the docs correctly on this handling (I'll review better tomorrow)

nicktobey: Yeah. I'm a bit confused why there's three new places where this can return an error. If they're all caused by the same issue (importing via stdin with no schema flag), I imagine that there would be one place to catch this and return an error, not three.

nicktobey:
Good catch noticing the panic, and thanks for including the fix for that in your PR.

I left one comment about some confusion on my part, but otherwise I like this. I'm running the CI workflows now.

go/cmd/dolt/commands/tblcmds/import.go
codeaucafe: Hi Nick, I realized this is the only check we need, the others are redundant given this and it'll fail earlier (I believe). I reverted the other places I put the error return for nil

nicktobey:
This looks good!

I forgot to mention: since the newly added bats test files are testing functionality that only works when run against a local database, we need to make sure that they don't get run when testing remote functionality. (That's why the "Test Bats Unix Remote" test is failing.)

We can address that by adding entries to the `SKIP_SERVER_TESTS` variable defined in `integration-tests/bats/helper/local-remote.bash`.

After that, I think this PR is good to go!

(As a side note, I noticed that you've been rebasing your branch in response to PR feedback. While rebasing creates a cleaner commit history, it makes the PR process more confusing because it makes it harder to see what changed, and comments left on the old diff may not be visible on the new diff. Once a PR has comments on it, we try to avoid rebasing.)

codeaucafe:
Thank you. I'll change that tonight. 

Sorry about the rebases, I dislike unorganized clean commits, but it makes sense to avoid after first PR comment. I am sorry if I missed that in the Contributing markdown.

Thanks again

codeaucafe:
Awesome! Glad to hear that.

Thank you for letting me contribute and reviewing my first Dolt contribution. I really appreciate it.

cheers

PR #9208, opened 2025-05-12
When a table is deleted on one branch and has its data modified on the other, record this as a schema conflict.

nicktobey:
Currently, attempting to merge when a table has been drop on one branch and had its data modified on the other produces an immediate error that aborts the merge. This error does not necessarily have enough context to determine the affected table, which makes resolving the merge difficult.

Given that we already handle a simultaneous drop + schema modification (by treating it as a schema conflict), we should be able to handle the simpler case of data modifications as well.

Since dropping a table is technically a schema change, this should now get reported as a schema conflict, and pause the merge, allowing the user to query dolt_schema_conflicts for more information.

After this PR, conflicts involving a dropped table will present the same as other schema conflicts: the user still won't be able to use `dolt conflicts resolve` to resolve them because of https://github.com/dolthub/dolt/issues/6616, but the user can query `dolt_schema_conflicts` to see the cause of the conflict.

### DIFF
--- a/go/libraries/doltcore/doltdb/root_val.go
+++ b/go/libraries/doltcore/doltdb/root_val.go
@@ -1369,7 +1369,7 @@ func ValidateTagUniqueness(ctx context.Context, root RootValue, tableName string
 	err = sch.GetAllCols().Iter(func(tag uint64, col schema.Column) (stop bool, err error) {
 		oldTableName, ok := existing.Get(tag)
 		if ok && oldTableName != tableName {
-			return true, schema.ErrTagPrevUsed(tag, col.Name, tableName, oldTableName)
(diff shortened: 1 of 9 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6ce5905</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6ce5905</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4 (approved):

Issue #8200, opened 2024-08-02
Support more flags with `dolt log --graph`


liuliu-dev:
add support to `dolt log --graph`:

- [x] `--all`
- [x] `--oneline`

macneale4:
Support for the `--all` operation would be amazing. We can currently do this with the following:

```
$ dolt branch
main
br
foobar
$ dolt log --graph main br foobar
...
```

But `dolt log --graph --all` would be way better.

Worth noting that `--all` makes sense even out of the context of `--graph`

PR #9246, opened 2025-05-23
--all option for dolt log

NathanGabrielson:
Issue #8200 

Adds a `--all` option for dolt log. You can also specify tables after the all option, with `dolt lot --all table1 table2`. 

I still want to work on:

- [x] The code's a bit messy still, I think a helper function might be nice, especially to set the existingTables map.
- [x] Maybe make it work with `--not` to exclude branches?
- [x] I'm sort of wondering if the code handling the generation of the "query" we make could be improved. It might not be worth it, but for example if you have a table and branch with the same name, we reinsert that value into the `params` slice. We also reuse a bunch of code for the code handling the first non-revision argument.

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -289,6 +289,7 @@ func CreateLogArgParser(isTableFunction bool) *argparser.ArgParser {
 	ap.SupportsFlag(ParentsFlag, "", "Shows all parents of each commit in the log.")
 	ap.SupportsString(DecorateFlag, "", "decorate_fmt", "Shows refs next to commits. Valid options are short, full, no, and auto")
 	ap.SupportsStringList(NotFlag, "", "revision", "Excludes commits from revision.")
+	ap.SupportsFlag(AllFlag, "", "Automatically select every branch in database")
 	ap.SupportsFlag(ShowSignatureFlag, "", "Shows the signature of each commit.")
 	if isTableFunction {
 		ap.SupportsStringList(TablesFlag, "t", "table", "Restricts the log to commits that modified the specified tables.")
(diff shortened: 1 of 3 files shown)


NathanGabrielson:
The only parts I'm a little iffy on are:

1. I'm abusing the fact that you can call the dolt_log() thing on the same branch a bunch (i.e. dolt_log('br1','br1','br1','br1') flags no errors), so if you use `--all` I just dump all the branches in there, and then include duplicates if you include branch names with `--all`. It's definitely a bit of an edge case, but it seems a little weird. 
2. The code getting the table names could probably be abstracted to a helper function. I think you could use something like `updateValidTables` which returns a slice to update the `existingTables`. Then we could pass in either a list of every branch if you use `--all`, or just the specific branches if that was your query. It's a little ugly as is.
3. Finally `--not` doesn't really work, but it behaves really weirdly in general so that's probably best left for another improvement.

coffeegoddd:
@NathanGabrielson __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>785e6e1</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>785e6e1</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@NathanGabrielson __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b769464</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b769464</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4 (approved):
```
 .  o ..
 o . o o.o
      ...oo
        __[]__
     __|_o_o_o\__
     \""""""""""/
      \. ..  . /
 ^^^^^^^^^^^^^^^^^^^^
```

PR #9251, opened 2025-05-27
Allow dumping wrapper values to SQL

nicktobey:
This prevents an issue where commands like `dolt dump` would fail when trying to export a table as SQL statements if the table has a `BLOB` column, since it would fail to unwrap the text wrapper value to access the underlying bytes.


jycor (approved):
LGTM

PR #9275, opened 2025-05-29
Fix merge test to properly test for a constraint violation

timsehn:
Identified and fixed by Claude


zachmu:
What's with the .backup files? Seems like this needs a human pass

timsehn:
accidentally added those will fix and merge

PR #9278, opened 2025-05-29
Two additional unskipped bats tests: one that said flakey, one in git status

timsehn:
One says it was skipped because it was flakey in CI. Let's see if it's now fine.


zachmu (approved):
LGTM

Issue #9281, opened 2025-05-30
`CREATE TABLE` doesn't generate `Query OK` confirmation message


angelamayxie:
Dolt
```
tmp/main*> create table MY_TABLE (ID int not null primary key, MY_BOOL bool not null);
tmp/main*> insert into MY_TABLE values (0, false);
Query OK, 1 row affected (0.00 sec)
```

MySQL
```
mysql> create table MY_TABLE (ID int not null primary key, MY_BOOL bool not null);
Query OK, 0 rows affected (0.01 sec)

mysql> insert MY_TABLE values (0, false);
Query OK, 1 row affected (0.00 sec)
```

jycor:
This is specific to `dolt sql` shell and `dolt sql -q "..."`.

I'm not sure why, but the function `processParsedQuery` from `dolt/go/cmd/dolt/commands/sql.go`, special cases `DDL` statements and returns `nil` for schema and rowIter?

This might be the fix, not sure what tests it'll break
```diff
diff --git a/go/cmd/dolt/commands/sql.go b/go/cmd/dolt/commands/sql.go
index 4c85367b52..4d20ca5277 100644
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -1199,15 +1199,11 @@ func processParsedQuery(ctx *sql.Context, query string, qryist cli.Queryist, sql
                }
                return nil, nil, nil, nil
        case *sqlparser.DDL:
-               _, ri, _, err := qryist.Query(ctx, query)
-               if err != nil {
-                       return nil, nil, nil, err
-               }
-               _, err = sql.RowIterToRows(ctx, ri)
+               sch, ri, _, err := qryist.Query(ctx, query)
                if err != nil {
                        return nil, nil, nil, err
                }
-               return nil, nil, nil, nil
+               return sch, ri, nil, nil
        case *sqlparser.Load:
                if s.Local {
                        return nil, nil, nil, fmt.Errorf("LOCAL supported only in sql-server mode")
```

macneale4:
We also don't show the success confirmation for create trigger.

angelamayxie:
We also don’t show a confirmation message for alter table and set statements

macneale4:
Went down a bit of a rabbit hole on this.

[...]

PR #9291, opened 2025-05-31
(#9281)fix: return schema and row iterator for DDL queries to show confirmation

aaryansinhaa:
#9281 fix: return schema and row iterator for DDL queries to show confirmation

## What I did: 
In `processParsedQuery` inside `dolt/go/cmd/dolt/commands/sql.go` , DDL statements (like `CREATE TABLE` ) were returning nil for both schema and row iterator, which caused the shell to skip printing the usual confirmation message (`Query OK, 0 rows affected`).

I modified the DDL case to return the `schema` and `rowIter` from `qryist.Query(...)`, allowing the shell to process the result and print the confirmation message just like it does for other statements.

In `dolt/go/cmd/dolt/commands/sql.go` , inside `processParsedQuery` function :
``` go
        return nil, nil, nil, nil
	case *sqlparser.DDL:
		sch, ri, _, err := qryist.Query(ctx, query)
		if err != nil {
			return nil, nil, nil, err
		}
		return sch, ri, nil, nil
	case *sqlparser.Load:
```

## Tests: 

- Ran `go test ./cmd/dolt/commands -v` and confirmed all tests pass:
```
PASS
ok      github.com/dolthub/dolt/go/cmd/dolt/commands    0.872s
```
- Manually tested in the SQL shell and confirmed the confirmation message:
Screenshot From 2025-05-31 14-27-32

## Feedback:
Any feedback would be appreciated, It was my first time contributing to this project, looking forward to do more in the future!

Thanks,
Aaryan.

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -1199,15 +1199,11 @@ func processParsedQuery(ctx *sql.Context, query string, qryist cli.Queryist, sql
 		}
 		return nil, nil, nil, nil
 	case *sqlparser.DDL:
-		_, ri, _, err := qryist.Query(ctx, query)
+		sch, ri, _, err := qryist.Query(ctx, query)
 		if err != nil {
 			return nil, nil, nil, err
 		}
-		_, err = sql.RowIterToRows(ctx, ri)
-		if err != nil {
-			return nil, nil, nil, err
-		}
-		return nil, nil, nil, nil
+		return sch, ri, nil, nil
 	case *sqlparser.Load:
 		if s.Local {
 			return nil, nil, nil, fmt.Errorf("LOCAL supported only in sql-server mode")


coffeegoddd:
@macneale4 should review

jycor:
Upon further investigation, I think that the intention for this special case is to ensure that the rowIter's `Close()` method is called. 

It seems redundant, but maybe something like this
```go
case *sqlparser.DDL:
	sch, ri, _, err := qryist.Query(ctx, query)
	if err != nil {
		return nil, nil, nil, err
	}
	rows, err := sql.RowIterToRows(ctx, ri)
	if err != nil {
		return nil, nil, nil, err
	}
	newRowIter := sql.RowsToRowIter(rows)
	return sch, newRowIter, nil, nil
...
```

Additionally, this should happen for `*sqlparser.AlterTable` and `*sqlparser.Set` (the case statement above this one)

PR #9300, opened 2025-06-03
Implement tree-based merging

nicktobey:
## High Level Concept

We want to optimize merges by taking into account the shape of the prolly tree. In many cases we can assemble a merged tree out of nodes from either branch, and avoiding unnecessarily recursing into the leaf nodes. But in order for that to work, the diff and merge operations need to be able to operate on internal tree nodes, and not just on leaf key-value pairs.

This project contains several independent-but-related changes in order to make that possible.

The three-way merge process has three distinct stages:
- The Two-Way Differ: compares two trees and produces an ordered stream of diffs describing the changes between them.
- The Three-Way Differ: Takes two two-way differs, each of which compares one of the branches with their common ancestor. Produces an ordered stream of patches, describing changes that must be applied to the `ours` tree in order to produce the merged tree.
- The Three-Way Merger: Applies the patches from the previous stage.

This PR modifies each of these stages in order to make tree-based diffing possible.

## Part 1: Two-Way Range Diffs

Relevant commit: 10e6cbb5d2301a651f87e983886ddc3b9476a524

If an internal tree node is added on one branch and has no conflicting changes on the other branch, that node will appear in the merged tree. We want to be able to treat that entire node as a single change, represented as a single `Diff` object.

This diff object is a new type of Diff called a `RangeDiff`: a diff that represents a change to a range of keys. In addition to the typical `Key` field that all diffs have, the range diff also has a `PreviousKey` field, which contains the key value that immediately precedes the beginning of the range. The diff thus represents changes on range `(PreviousKey,Key]` where the lower bound is open and the upper bound is closed. The `To` value of this diff is the hash address of the added tree node, or `nil` if all keys in that range were deleted. Note that if the address is not nil, it is not possible to determine whether the diff contains an addition, a modification, or a deletion without loading the referenced node, and in fact a range diff may contain all of the above within its range. But in many cases loading the node is not necessary for the merge.

Range diffs make it possible for a differ to produce diffs based on an intermediate level of the tree without loading the levels beneath it. For non-leaf levels of the tree, all produced diffs have a DiffType of `RangeDiff`.

By default, differs do not produce range diffs. The newly added function `RangeDifferFromRoots` returns a differ that can return range diffs.

The user of this "Range Differ" can call two different methods in order to get the next diff:
- Next(): returns the next diff, which may be at the same tree level as the previous diff, or a higher level.
- split(): recurses one level lower into the tree, returning a diff whose range is within the range of the previous diff. (If this causes the differ to reach level 0, it will start producing the standard AddedDiff, ModifiedDiff, and RemovedDiff values.)

When a call to Next hits a chunk boundary, it returns to the previous tree level. This results in the following invariant:

- If you list every node returned by the differ *except* for the nodes that preceded a call to `split()`, then none of those nodes have overlapping ranges, those nodes appear in order, and they collectively describe every change between the two trees.

As a requirement for implementing this, I had to make some changes to the AddrDiff and JsonDiff types. These were built on top of differs, but directly accessed Differ internals and made assumptions about their behavior. Changing the internal behavior broke these types, so I changed them to no longer rely on internal differ state for correctness.

## Part 2: Producing Three-Way Diffs (Range Patches)

Relevant commit: 02a242e95ee21d9ba6dc7372ce82de6ac584a988

Part 1 only describes two-way diffing. We want to use that to achieve optimal three-way diffing and merging.

Dolt currently has two different algorithms for performing three-way merging:
- The simpler `ThreeWayMerge` method, which takes two trees, and diffs them both against a common base tree. We use this for merging simple internal tree-based structures like artifact maps and commit closures.
- The more complicated `ThreeWayDiffer` type, which is an iterator that produces `ThreeWayDiff` values. `ThreeWayDiffs` contain additional information that is necessary to detect constraint violations, update artifact maps, etc. We use this for merging table data.

As of https://github.com/dolthub/dolt/pull/9229, both approaches write a stream of Patch values to a channel, where a Patch is a key-value pair describing a change that must be made to the left branch in order to produce the merged tree. The `ThreeWayDiffer` type does extra work before it produces these patches, but in both cases the patches are consumed and applied the same way.

In order to merge optimally, we need to allow Patches to represent a range, just like how Part 1 allows diffs to represent arrange. I accomplished this by just allowing a patch to wrap a diff.

If the table being merged has constraints, we have to look at every modified row to ensure that the constraints aren't being violated by the merge. So for now, we only allow for range patches in the case that there are no constraints to violate, and we can safely use the simpler `ThreeWayMerge` approach. We also currently only do this if there are no secondary indexes on the table, although we should be able to relax this in the future.

The new version of `ThreeWayMerge` uses the new range differs to produce the smallest set of diffs that describe the merged tree: if a range diff produced by one branch does not overlap with any of the diffs produced by the other branch, we can pass that entire range as a single Patch. If the two branches have overlapping range diffs, we must call `split()` on the differs to produce smaller range diffs that don't overlap.

## Part 3: Apply Range Patches via the Chunker

Relevant commits:
- 96889a846a842500a7e328dbd1f3168d119fd68e
- ffdce3f1e21ad7170d3978b4496a067739898817

Once we produce the stream of patches, we need to apply them to the tree. This is the most straightforward part of the PR: we change the API for the chunker to allow it to take a node address and write all of that node's rows into the new table. If the chunker is currently at a chunk boundary, it can write the address directly into the new tree without needing to load it. Otherwise, it loads the node and recursively writes it's children into the new tree.

## Putting it all together during Merge

Relevant commit: 083fdbe752ecb3e0e7ae95c4978299887c4775b9

This commit contains the changes to table merger, built on top of all the previous changes. It checks whether or not the table merge meets the current limitations for tree-based merging, and picks which algorithm to use to produce the stream of patches, then uses `ApplyMutations` to apply those patches to the `ours` tree, producing the merged tree.

## Impact

The impact is best seen when the two branches are making changes to completely separate regions of the key space, for example if every key modified by branch A is less than every key modified by branch B. This isn't an unreasonable use case: imagine a bulk import job getting merged back into a main branch, where all the imported keys are contiguous.

I created a benchmark: a table with a single int column.
Ancestor table: empty
Left branch table: contains the values 0 <= pk < 1,000,000
Right branch table: contains the values 1,000,000 <= pk < 2,000,000

The SerialMessage flatbuffer uses 8 bytes to store each row, so both branches contain around 8MB of new data.

Prior to this PR, calling `dolt_merge` took 1.38 seconds on my laptop, regardless of the direction of the merge.
After this PR, calling `dolt_merge` took 0.012 seconds on my laptop, regardless of the direction of the merge.

That's roughly a 100x speedup for this example. The actual speedup is an asymptotic increase in performance: previously, merge time was O(N+M), where N and M are the number of changed rows on each branch. The new time complexity is harder to measure, since it depends on the number of contiguous regions that are modified by one branch but not the other, and also on the height of tree, but it's approximately O(log(N+M)) in the best case, and no worse than the original algorithm even in the worst case.

I believe that we should always prefer this approach over the original implementation. The original implementation cares about the "direction" the merge, which can impact performance, while this implementation should have the same performance characteristics regardless of the direction of the merge. This is not immediately obvious given that both approaches pick one side of the merge and apply "patches" from the other, but it's still true: the tree-aware approach to merging should achieve the theoretical minimum number of loads and comparisons to complete the merge, regardless of merge direction.

In the worst case, this approach should have performance comparable to the original performance, although I'll be doing more benchmarks before submitting to verify this. The most likely way that this *may* cause a performance regression in the worst case is because the objects that we use to encode a tree patch is now larger, and these objects get sent through channels, which may have observable performance implications: See https://www.jtolio.com/2016/03/go-channels-are-bad-and-you-should-feel-bad/ for more details. If this turns out to be the case, we can optimize the `Patch` struct for this before submitting.

## Consequences / Caveats

One unintended consequence of this change is that it may not be possible to accurately calculate merge stats using this method.

Merging a table produces a `MergeStats` struct that contains a count of the number of rows added/modified/deleted. But this information can only be computed by visiting every row that has been changed, which this approach no longer does. If we skip recursing into a node because we see that node has only changed on one branch, we can't know the exact number of modifications.

However, if you try the above command line example, you'll notice that it correctly reports the number of added rows. You'll also notice that it's significantly slower than calling `DOLT_MERGE`. This is because the `MergeStats` produced by merging the table is actually ignored. Instead, the `dolt merge` CLI makes a subsequent select from the `dolt_diff_stats` system table function in order to get the number of modified rows, and this system table function does a conventional diff.

Thus, the fact that we're not calculating correct merge stats during the merge itself may not matter. We do have some tests for it, which are breaking. I'm going to discuss with Neil how we want to handle this.

## Limitations

Currently, this PR only produces RangeDiffs under the following narrow circumstances:
- The table is not keyless
- The table does not have secondary indexes
- The table does not have any constraints
- The merge does not introduce schema changes

We should be able to relax these constraints if needed. For example:
- It should be possible to merge secondary indexes using the same algorithm we use here to merge primary indexes.
- We need to ensure that the merge doesn't produce a table that violates constraints, but this doesn't necessarily require examining every leaf node: depending on the exact constraints / which branch adds a constraint / which branch adds rows that potentially violate constraints, we may be able to validate the table without having to process every modified row.

## Room for Improvement

The most obvious room for improvement is relaxing the limitations under which we can safely produce and apply range diffs.

Another possible improvement: The current design for the how the three-way differ uses the underlying two-way differs is interactive: after getting a two-way diff, the three-way differ has to decide whether to call `split()` to recurse into a lower level of the tree, or call `Next()` to accept the provided diff and move on. This design is inherently non-parallelizable. It's possible that an alternate approach may allow each two-way differ to have its own goroutine and produce diffs in parallel, and then choosing which ones to split separately.

### DIFF
--- a/go/store/prolly/patch_based_merging_test.go
+++ b/go/store/prolly/patch_based_merging_test.go
@@ -16,15 +16,16 @@ package prolly
 
 import (
 	"context"
-	"github.com/dolthub/dolt/go/store/prolly/message"
-	"github.com/dolthub/dolt/go/store/prolly/tree"
-	"github.com/dolthub/dolt/go/store/val"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"strconv"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"golang.org/x/sync/errgroup"
+
+	"github.com/dolthub/dolt/go/store/prolly/message"
+	"github.com/dolthub/dolt/go/store/prolly/tree"
+	"github.com/dolthub/dolt/go/store/val"
 )
 
 // getChunkBoundaries returns all the keys in a node, which indicate the chunk boundaries in the previous tree level.
--- a/go/store/prolly/tree/tree_patcher.go
+++ b/go/store/prolly/tree/tree_patcher.go
@@ -17,6 +17,7 @@ package tree
 import (
 	"bytes"
 	"context"
+
 	"github.com/dolthub/dolt/go/store/hash"
 	"github.com/dolthub/dolt/go/store/prolly/message"
 )
--- a/go/store/prolly/traditional_merging_test.go
+++ b/go/store/prolly/traditional_merging_test.go
@@ -17,10 +17,12 @@ package prolly
 import (
 	"bytes"
 	"context"
+	"io"
+
+	"golang.org/x/sync/errgroup"
+
 	"github.com/dolthub/dolt/go/store/prolly/message"
 	"github.com/dolthub/dolt/go/store/prolly/tree"
-	"golang.org/x/sync/errgroup"
-	"io"
 )
 
 func traditionalThreeWayMerge[K ~[]byte, O tree.Ordering[K], S message.Serializer](ctx context.Context, ns tree.NodeStore, left, right, base tree.Node, collide tree.CollisionFn, leftSchemaChange, rightSchemaChange bool, order O, serializer S) (final tree.Node, err error) {
--- a/go/store/prolly/tree/patch_test.go
+++ b/go/store/prolly/tree/patch_test.go
@@ -16,11 +16,13 @@ package tree
 
 import (
 	"context"
-	"github.com/dolthub/dolt/go/store/val"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"io"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/store/val"
 )
 
 func TestPatchGeneratorFromRoots(t *testing.T) {


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>21ecb1b</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>21ecb1b</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>84dc72e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>84dc72e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>083fdbe</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>083fdbe</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>7b6d01b</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>7b6d01b</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/indexed_json_diff.go
```diff
@@ -151,13 +151,17 @@ func (jd *IndexedJsonDiffer) Next(ctx context.Context) (diff JsonDiff, err error
...
 			// (ie, both chunks ended at the same JSON path). We can use `Differ.Next` to seek to the next difference.
-			// Passing advanceCursors=false means that instead of using the returned diff, we read the cursors out of
-			// the differ, and advance them manually once we've walked the chunk.
-			_, err := jd.differ.next(ctx, false)
+			_, err := jd.differ.Next(ctx)
 			if err != nil {
 				return JsonDiff{}, err
 			}
 
+			// Ordinarily, the Differ will advance the cursors the next time Next() is called.
```
macneale4: Why the hack? Either something is wrong with the model or it's not a hack. Honestly, if you hadn't commented on it this way I wouldn't have noticed. Probably just soften the comment.

macneale4 (commented):
I haven't looked at the full change. Just the three commits you requested.

go/store/prolly/tree/indexed_json_diff.go
nicktobey: Yeah, this is just a bad approach on my part. One of the other commits in the PR removed it (c8147bba17a8d5d05b2dd329f61e7878e62c432a). I should have just amended that commit onto this one.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>1294cbe</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>1294cbe</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>efe32db</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>efe32db</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>aa31569</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>aa31569</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

nicktobey:
Based on feedback, I've made some changes.

The biggest change is that the tree merging is no longer built into the existing diff and merge types. I chose this approach at the time because tree merging is in a sense a generalization of key-value merging. Having them use the same underlying algorithm minimized code duplication and provided extra test coverage, since all existing merge operations would be using the new merge logic.

But doing this came with some problems:
- The logic of merging became harder to reason about, even for existing use cases that would always walk the full tree and produce every key-value diff anyway.
- The existing use cases would potentially suffer a performance penalty due to the more complicated logic.
- There were certain inputs that can't currently happen that would need to be covered, and would be difficult to reason about. Example: what is the intended behavior when diffing trees on a subrange, where the user provides a "stop" key that occurrs in the middle of a leaf chunk? This wasn't an operation that made any sense.
- Tree-based diffing breaks some assumptions that are an important part of how we visualize key-value merging. For instance, if we swap the "to" and "from" tables in a key-value diff, we get the same number of diffs, but every addition becomes a removal and vice-versa. But for tree-based merging, that doesn't happen.

Ultimately, the nail in the coffin was that the operation I called "Range diffing" in the previous comment isn't actually a diff and it doesn't make sense to think of it as one. It's the wrong abstraction to describe the operation.

So while the algorithm hasn't changed, the way it's presented has.

The purpose of comparing two trees when performing tree-based merging isn't to diff them, but to produce a sequence of **patches** that, if applied to one of the trees, will produce the merged tree.

The PR now has its own set of types specifically designed for working with patches, called Patch, PatchBuffer, and PatchIter, and a function ApplyPatches that efficiently produces a merged tree given an input tree and an ordered sequence of patches to apply.

Hopefully, this makes the logic easier to follow.

Additionlly, I've added tests that assert both that the sequences of patches produced matches what we expect, and also that applying these patches produces the same tree as if the trees were merged conventionally.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>4a0783d</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>4a0783d</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>4cc14ce</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>4cc14ce</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8f7972a</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8f7972a</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/chunker.go
```diff
@@ -168,61 +175,62 @@ func (tc *chunker[S]) DeletePair(ctx context.Context, _, _ Item) error {
 //	anticipation of impending edits that may edit the current chunk. Note that
 //	processPrefix is only necessary for the "fast forward" case where we
 //	synchronized the tree level before reaching |next|.
-func (tc *chunker[S]) advanceTo(ctx context.Context, next *cursor) error {
-	cmp := tc.cur.compare(next)
+func (tc *chunker[S]) insertRange(ctx context.Context, start, end *cursor) error {
```
reltuk: This docstring might need a once over. For example `s/|next|/|end|/`

reltuk: Adding the `start` parameter here is weird to me, especially how its handled within the body. I wouldn't expect `chunker` to be setup to allow, for example, `tc.append()` calls to work appropriately when `tc.cur` hasn't been advanced to the correct insert point. And the same time, we're working with pointers and the two call sites are always the chunker's actual `cur`, so all these changes seem maybe unnecessary. What's going on here?

go/store/prolly/tree/tree_patcher.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		if err != nil {
+			return Node{}, err
+		}
+		prev := newMutation.EndKey
+		newMutation = edits.NextPatch(ctx)
+		nextKey := newMutation.EndKey
+		if nextKey == nil {
+			nextKey = newMutation.KeyBelowStart
+		}
```
reltuk: A comment explaining what's happening here? Why is `EndKey` missing on the first patch a fast-complete but `EndKey` missing here means that maybe `KeyBelowStart` matters?

go/store/prolly/tree/tree_patcher.go
```diff
@@ -0,0 +1,180 @@
...
+		err = insertNode(ctx, chkr, fromKey, toKey, hash.New(addr), subtree, level, order)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = Seek(ctx, chkr.cur, K(toKey), order)
+	if err != nil {
+		return err
+	}
```
reltuk, line 185: This seems like it would be part of `insertNode`. Leaving the chunker in a weird state which requires the caller to rectify it after the splice seems wrong.

go/store/prolly/tree/tree_patcher.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if err != nil {
+		return err
+	}
+	// Append all key-values from the Node.
+	// If we're on a chunk boundary, this will just copy the node in.
+
+	// If the start of the range is greater than the last key written, and the tree levels line up, we can just write the supplied address.
+	// If supplied tree level is *above* our current one, we need to load the chunk and write its children until we line up again.
+	// But it might be below, in which case we need to make sure that we write the address at the right level.
```
reltuk, line 173: Mybe slightly stale after insertNode was factored out or something.

go/store/prolly/patch_based_merging_test.go
```diff
@@ -0,0 +1,494 @@
...
+				},
+			})
+	})
+
+	t.Run("deleting an entire chunk at the end produces a single range", func(t *testing.T) {
+		testPatchBasedMerging(
+			t, ctx, ns, desc,
+			threeAndAHalfChunks,
+			mapWithUpdates(threeAndAHalfChunks, remove(1)),
+			mapWithUpdates(threeAndAHalfChunks, makeDeletePatches(831, 1024)...),
```
reltuk: Nit: Maybe slightly better here and 394, for example, to reference `chunkBoundaries`

go/store/prolly/tree/merge.go
```diff
@@ -88,172 +81,295 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
...
 		err, rok = nil, false
 	}
 	if err != nil {
-		return MergeStats{}, err
+		return err
 	}
 
 	for lok && rok {
-		cmp := l.order.Compare(ctx, K(left.Key), K(right.Key))
+		order := l.order
```
reltuk: Nit: slightly more clear to move this out of the loop probably.

go/store/prolly/tree/merge.go
```diff
@@ -88,172 +81,295 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
...
+				// This change is already on the left map, so we ignore it.
+				left, lDiffType, err = l.Next(ctx)
+				if err == io.EOF {
+					err, lok = nil, false
+				}
+				if err != nil {
+					return err
+				}
+			} else if nilCompare(ctx, order, K(right.EndKey), K(left.KeyBelowStart)) <= 0 {
+				// Right change is entirely before right change.
```
reltuk:
```suggestion
				// Right change is entirely before left change.
```

go/store/prolly/tree/merge.go
```diff
@@ -88,172 +81,295 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
...
-				return MergeStats{}, err
+			// Convergent edit:
+			if !bytes.Equal(left.To, right.To) {
+				resolvedPatch, ok := resolveCollision(left, lDiffType, right, rDiffType, cb)
+				if ok {
+					err = buf.SendPatch(ctx, resolvedPatch)
+				}
+				if err != nil {
+					return err
+				}
```
reltuk: Where did this `err` come from? What are we supposed to do if `!ok`?

go/store/prolly/tree/merge.go
```diff
@@ -88,172 +81,295 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
...
+
+				right, rDiffType, err = r.Next(ctx)
+				if err == io.EOF {
+					err, rok = nil, false
+				}
+				if err != nil {
+					return err
+				}
+			} else if bytes.Equal(left.To, right.To) {
+				// A concurrent change.
```
reltuk: This is a bit subtle. Might be worth commenting that `EndKey` is fully determined by `To` here because levels > 0, so that these two patches are pointing to identical contents of within their merkle trees.

go/store/prolly/tree/merge.go
```diff
@@ -88,172 +81,295 @@ func ThreeWayMerge[K ~[]byte, O Ordering[K], S message.Serializer](
...
+						return err
+					}
+				}
+			}
+			continue
+		}
+
+		// If one branch returns a range patch and the other returns a point patch, we need to see if they overlap and possibly split the range diff.
+		if rightLevel > 0 {
+			if nilCompare(ctx, l.order, K(left.EndKey), K(right.KeyBelowStart)) <= 0 {
```
reltuk: When do we use `order` and when do we use `l.order`? Should be `order` everywhere, right?

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+	return Patch{
+		From:   from,
+		EndKey: key,
+		To:     to,
+		Level:  0,
+	}
+}
+
+type PatchIter interface {
+	NextPatch(ctx context.Context) Patch
```
reltuk: Nit: I find it weird to take `context` and not return an `error`. `context` almost always means I/O or cancelable or something. I realize there are other iterators in `tree` and friends that already do this. Thoughts?

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+var _ PatchIter = PatchBuffer{}
+
+func NewPatchBuffer(sz int) PatchBuffer {
+	return PatchBuffer{buf: make(chan Patch, sz)}
+}
+
+func (ps PatchBuffer) SendPatch(ctx context.Context, patch Patch) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
```
reltuk: Maybe slightly better:

```suggestion
		return context.Cause(ctx)
```

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+	case ps.buf <- patch:
+		return nil
+	}
+}
+
+func (ps PatchBuffer) SendKV(ctx context.Context, key, value Item) error {
+	patch := Patch{
+		EndKey:       key,
+		To:           value,
+		SubtreeCount: 1,
```
reltuk: Is this expected? `newLeafPatch` currently sends `SubtreeCount: 0` for a point modification.

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+		Level:        0,
+	}
+	select {
```
reltuk: Nit: Slightly better as:

```go
return ps.SendPatch(ctx, Patch{
	EndKey:       key,
	To:           value,
	SubtreeCount: 1,
	Level:        0,
})
```

?

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case ps.buf <- patch:
+		return nil
+	}
+}
+
+func (ps PatchBuffer) SendDone(ctx context.Context) error {
```
reltuk, line 114: I would have expected `close(ps.buf)` here. Alternatively, if receiving `Patch{}` on `NextPatch` is a non-error condition, but receiving `!ok` on the `ps.buf` is an error condition, I would have expected the caller to be able to tell.

Is there a way for a caller of `NextPatch()` to know they didn't get all the patches because of an error in the sender?

go/store/prolly/tree/patch_generator.go
```diff
@@ -0,0 +1,533 @@
...
+
+// advanceFromPreviousPatch advances the cursors past the end of the previous patches.
+// In the event that the chunk boundaries have shifted between the two versions of the tree,
+// this process might produce additional patches until the boundaries line up again.
+func (td *PatchGenerator[K, O]) advanceFromPreviousPatch(ctx context.Context) (patch Patch, diffType DiffType, err error) {
+	if td.previousPatchLevel > 0 {
+		switch td.previousDiffType {
+		case AddedDiff:
+			td.previousKey = td.to.CurrentKey()
+			// If we've already exhausted the |from| iterator, then returning to the parent
```
reltuk: It might be safe and appropriate to still walk to the next not-at-end parent on the `to` cursor, but is it the case that `from.Valid() == false` here?

reltuk (approved):
LGTM! No large scale comments. A few questions and nits.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>2dee960</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>2dee960</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>abc88a4</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>abc88a4</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>1f84298</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>1f84298</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f1a8382</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f1a8382</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>3c1a83c</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>3c1a83c</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e70f947</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e70f947</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>444839e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>444839e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>ce3ceec</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>ce3ceec</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>44313de</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>44313de</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6fb0a8e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6fb0a8e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>730482f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>730482f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #9306, opened 2025-06-03
go: store/datas/pull: clone.go: Improve robust of Clone for certain remoteapi implementations when the remote Conjoins.

reltuk:
The clone code works by listing the remote table files and downloading them into the local table file store. When the remote is a remoteapi implementation, like a DoltHub repository, this resulting in listing the remote table files and using URLs to fetch each of them.

The URLs returned from these APIs can expire and they need to be refreshed. This refresh can happen in two ways:

1) There is explicit support in the TableFileSource representation returned by the API to include a mechanism to refresh it. DoltHub uses this, and the Dolt client will make use of that support to refresh expired URLs. 2) The heavy handed approach is to list the table files again and use the newly returned URLs.

The Clone code has explicit support for doing #2, and it is necessary for remoteapi implementations with expiring URLs but without explicit RefreshTableFileUrl support. dolt itself, when running a remote as part of sql-server for example, does not implement RefreshTableFileUrl support, and so the re-list support is still necessary.

This PR changes the Clone implementation so that, on a retry, it makes all the newly returned table file sources available for the next try, but it keeps the old sources around if they no longer come back from ListTableFiles. In this way, we get strictly more robust behavior than before.

The downside is that, when the remote file is actually gone, the Clone code will continue attempting to download it until it reaches a terminal download failure. This change in behavior is not as disruptive as the current behavior, and so we make this new trade off for now.

### DIFF
--- a/go/store/datas/pull/clone.go
+++ b/go/store/datas/pull/clone.go
@@ -204,11 +204,34 @@ func clone(ctx context.Context, srcTS, sinkTS chunks.TableFileStore, sinkCS chun
 		if failureCount >= maxAttempts {
 			return err
 		}
-		if _, sourceFiles, appendixFiles, err = srcTS.Sources(ctx); err != nil {
+		if _, refreshedSourceFiles, refreshedAppendixFiles, err := srcTS.Sources(ctx); err != nil {
 			return err
 		} else {
-			tblFiles = filterAppendicesFromSourceFiles(appendixFiles, sourceFiles)
-			_, fileIDToTF, _ = mapTableFiles(tblFiles)
+			refreshedTblFiles := filterAppendicesFromSourceFiles(refreshedAppendixFiles, refreshedSourceFiles)
+			_, refreshedFileIDToTF, _ := mapTableFiles(refreshedTblFiles)
+			// Sources() will refresh remote table file
+			// sources with new download URLs. However, it
+			// will only return URLs for table files which
+			// are in the remote manifest, which could
+			// have changed since the clone started. Here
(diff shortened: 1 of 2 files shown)


go/store/datas/pull/clone.go
```diff
@@ -204,11 +204,34 @@ func clone(ctx context.Context, srcTS, sinkTS chunks.TableFileStore, sinkCS chun
...
+			// conjoined away or have been the victim of a
+			// garbage collection run on the remote.
+			//
+			// If these files are no longer accessible,
+			// for example because the URLs expired
+			// without a RefreshTableFileUrlRequest being
+			// provided, or because the table files
+			// themselves have been removed from storage,
+			// then continuing to use these sources will
+			// fail termainally eventually. But in the
```
coffeegoddd:
```suggestion
			// fail terminally eventually. But in the
```

coffeegoddd (approved):
LGTM

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c5a5ec8</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c5a5ec8</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

Issue #9374, opened 2025-06-20
Support MYSQL_USER and MYSQL_PASSWORD environment variables in Dolt Docker image


bfontaine:
This issue is similar to https://github.com/dolthub/dolt/issues/9345: the MySQL Docker image supports setting `MYSQL_USER` and `MYSQL_PASSWORD` to create an initial user ([source](https://github.com/docker-library/mysql/blob/8ade9b2c9a32a79fbaa44b564d09a40744f1d105/8.4/docker-entrypoint.sh#L316)). The Dolt Docker image should support them to be compatible with it.

bfontaine:
For the context, I’m trying to figure out how to make a Dolt plugin for [Dokku](https://dokku.com/) (https://github.com/Bixoto/dokku-dolt) but for now I’m stuck at the step of creating the initial database because of this.

fulghum:
Thanks for the great feature request @bfontaine. It makes sense to support these same parameters for the Dolt Docker image. PR https://github.com/dolthub/dolt/pull/9346 applies a similar type of change, so will be a good reference for implementation. @elianddb will pick this up after he wraps up a couple things. 

Thanks for sharing about the Dokku plugin you're building, too. Sounds like a cool project! We'd love to play with it when you get it up and running.

PR #9376, opened 2025-06-21
9374 support mysql user mysql password env vars

elianddb:
Fixes: #9374
Add support for MYSQL_USER/MYSQL_PASSWORD environment variables


bfontaine:
If I’m not mistaken you could factorize the code by reusing `get_env_var` in `create_default_database_from_env`

Edit: or even merge `create_default_database_from_env` and `create_user_from_env`, so you get `database` only once

PR #9389, opened 2025-06-24
Archive cmd improvements

macneale4:
This change has a grab bag of improvements from working with media_wiki database.
1) All table files, not just oldgen, and converted to archives.
2) Updates to the DB are performed incrementally. Each table file conversion results in a swap of tables, allowing for large job termination and not losing all progress
3) Compression workers are parallelized.
4) Revert is incremental, and works on newgen.


coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>ab8bc20</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>ab8bc20</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/cmd/dolt/commands/archive.go
```diff
@@ -108,6 +108,15 @@ func (cmd ArchiveCmd) Exec(ctx context.Context, commandStr string, args []string
 	progress := make(chan interface{}, 32)
 	handleProgress(ctx, progress)
 
+	defer func() {
```
reltuk, line 113: Seems much more idiomatic to read until the end of the channel and block on a call to `go handleProgress` returning.

go/store/nbs/archive_build.go
```diff
@@ -39,163 +40,233 @@ import (
...
+		}
+		if isJournalAddr(id) {
+			continue
+		}
 
-					swapMap[arc.hash()] = hash.Parse(id)
-				}
-			}
+		orginTfId := revertMap[id]
+		exists, err := smd.oldGenTableExists(orginTfId)
```
reltuk: This seems misplaced given that we're no longer confined to oldgen.

go/store/nbs/archive_build.go
```diff
@@ -39,163 +40,233 @@ import (
...
-					if err != nil {
-						return err
-					}
+		arc, ok := cs.(archiveChunkSource)
+		if !ok {
+			continue
+		}
+		if isJournalAddr(id) {
+			continue
+		}
```
reltuk: This probably makes more sense immediately after the beginning of the `for` loop. In either case, it will never be true here I don't think – the journal is not an archiveChunkSource.

go/store/nbs/archive_build.go
```diff
@@ -239,46 +310,43 @@ func convertTableFileToArchive(
...
-	if grouped+singles != idx.chunkCount() {
-		// Leaving as a panic. This should never happen.
-		missing := idx.chunkCount() - (grouped + singles)
-		panic(fmt.Sprintf("chunk count mismatch. Missing: %d", missing))
+	chunkCount := grouped + singles
+	if chunkCount != idx.chunkCount() {
+		// Table files can have duplicate chunks, and the archive build will dedupe them. So we do end
+		// up with fewer total chunks in the archive.
+		dupes := idx.chunkCount() - (grouped + singles)
+		progress <- fmt.Sprintf("\nChunk count mismatch. Duplicate chunks: %d\n", dupes)
```
reltuk, line 328: If this is true then we can't use the chnkCnt from the archive file in the unarchive command above in the case that we're able to fast-path it based on the revert map, right?

go/store/nbs/archive_build.go
```diff
@@ -394,42 +463,107 @@ func writeDataToArchive(
...
-			if err != nil {
-				return 0, 0, 0, err
-			}
-			err = arcW.stageZStdChunk(h, dictId, id)
-			if err != nil {
-				return 0, 0, 0, err
+	workCh := make(chan hash.Hash, len(allChunks))
+	resultCh := make(chan compressedChunk, workerCount)
+	errCh := make(chan error, 1)
+	var wg sync.WaitGroup
```
reltuk: This should almost certainly be an errgroup instead. There's no reason to allocate a workCh as large as the HashSet, for example. Something more like:

```go
	eg, ctx := errgroup.WithContext(context.Background())
	eg.Go(func() error {
		for h := range allChunks {
			select {
			case workCh <- h:
			case <-ctx.Done():
				return context.Cause(ctx)
			}
		}
		close(workCh)
		return nil
	})
	for i := 0; i < workerCount; i++ {
		eg.Go(func() error {
			cmpBuff := make([]byte, 0, fourMb)
			for {
				select {
				case addr, ok := <-workCh:
					if !ok {
						return nil
					}
					c, err := chunkCache.get(ctx, addr, stats)
					if err != nil {
						return err
					}
					cmpBuff = gozstd.CompressDict(cmpBuff[:0], c.Data(), defaultDict)
					cp := append([]byte{}, cmpBuff...)
					select {
					case resultCh <- compressedChunk{h: h, data: cp}:
					case <-ctx.Done():
						return context.Cause(ctx)
					}
				case <-ctx.Done():
					return context.Cause(ctx)
				}
			}
		})
	}
	var completed int32
	eg.Go(func() error {
		for {
			select {
			case cc, ok := <-resultCh:
				if !ok {
					return nil
				}
				id, err := arcW.writeByteSpan(cc.data)
				if err != nil {
					return err
				}
				err = arcW.stageZStdChunk(cc.h, defaultSpanId, id)
				if err != nil {
					return err
				}
				completed++
				select {
				case progress <- ArchiveBuildProcessMsg{
					Stage: "Writing Ungrouped Chunks",
					Total: totalChunks,
					Completed: completed,
				}:
				case <-ctx.Done():
					return context.Cause(ctx)
				}
			case <-ctx.Done():
				return context.Cause(ctx)
			}
		}
	})

	err := eg.Wait()
	if err != nil {
		return 0, err
	}
	return uint32(totalChunks), nil
```

reltuk (approved):
Generally looks good. A couple comments.

coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6f8ea36</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6f8ea36</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/archive_build.go
macneale4: Good catch. The logic was busted.

go/store/nbs/archive_build.go
macneale4: yeah... and I think this means that the fast path is bunk. That's a bummer.

coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>588b790</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>588b790</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #9419, opened 2025-06-28
Add indexes to dolt_commit_diff_ tables that include the primary key of the underlying table

nicktobey:
The `dolt_commit_diff_` system tables allow for comparing a table at two different commits. They require a filter on the `to_commit` and `from_commit` fields, which the table exposes as an index.

This PR extends those indexes to have additional columns based on the primary keys of the underlying table. This allows a user to efficiently compare only a specific range of that table.

Specifically, if the table's primary key columns are named `pk1`, `pk2`, ... `pkN`, then the system table now provides two indexes:

- `to_commit`, `from_commit`, `from_pk1`, `from_pk2`... `from_pkN`
- `to_commit`, `from_commit`, `to_pk1`, `to_pk2`... `to_pkN`

If the select filter uses a prefix of either of these indexes, then we can implement the select as a Diff on only the intended range.

Because the `to_` columns are NULL for removed rows, and the `from_` columns are NULL for inserted rows, a filter on the `from_` columns will only return modifications and deletions, and a filter on the `to_` columns will only return modifications and additions. In hindsight, perhaps the primary key rows should have appeared in the diff as a single column instead of both a `to_` and `from_` column, but that ship has sailed.

As for the implementation, I don't love the fact that code paths that diff the entire table without accepting a range pass around a zero-valued `sql.Lookup`, which we then detect by checking to see if `lookup.index == nil`. It's a code smell for sure. But I couldn't think of a better way to do it.

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/diff_iter.go
+++ b/go/libraries/doltcore/sqle/dtables/diff_iter.go
@@ -206,6 +206,7 @@ type prollyDiffIter struct {
 	targetFromSch, targetToSch schema.Schema
 	fromConverter, toConverter ProllyRowConverter
 	keyless                    bool
+	ranges                     []prolly.Range
 
 	fromCm commitInfo2
 	toCm   commitInfo2
@@ -234,7 +235,7 @@ var _ sql.RowIter = prollyDiffIter{}
 // than |targetFromSchema| or |targetToSchema|. We convert the rows from the
 // schema of |from| to |targetFromSchema| and the schema of |to| to
 // |targetToSchema|. See the tablediff_prolly package.
-func newProllyDiffIter(ctx *sql.Context, dp DiffPartition, targetFromSchema, targetToSchema schema.Schema) (prollyDiffIter, error) {
+func newProllyDiffIter(ctx *sql.Context, dp DiffPartition, targetFromSchema, targetToSchema schema.Schema, lookup sql.IndexLookup) (prollyDiffIter, error) {
 	fromCm := commitInfo2{
 		name: dp.fromName,
 		ts:   (*time.Time)(dp.fromDate),
@@ -256,6 +257,18 @@ func newProllyDiffIter(ctx *sql.Context, dp DiffPartition, targetFromSchema, tar
 			return prollyDiffIter{}, err
 		}
 		if fsch, err = dp.from.GetSchema(ctx); err != nil {
+			return prollyDiffIter{}, err
+		}
+	}
+
+	var ranges []prolly.Range
+	var err error
+	if lookup.Index != nil {
+		for i := range lookup.Ranges.(sql.MySQLRangeCollection) {
+			lookup.Ranges.(sql.MySQLRangeCollection)[i] = lookup.Ranges.(sql.MySQLRangeCollection)[i][2:]
+		}
+		ranges, err = index.ProllyRangesFromIndexLookup(ctx, lookup)
+		if err != nil {
 			return prollyDiffIter{}, err
 		}
 	}
@@ -306,6 +319,7 @@ func newProllyDiffIter(ctx *sql.Context, dp DiffPartition, targetFromSchema, tar
 		keyless:       keyless,
 		fromCm:        fromCm,
 		toCm:          toCm,
+		ranges:        ranges,
 		rows:          make(chan sql.Row, 64),
 		errChan:       make(chan error),
 		cancel:        cancel,
@@ -339,8 +353,7 @@ func (itr prollyDiffIter) Close(ctx *sql.Context) error {
 
 func (itr prollyDiffIter) queueRows(ctx context.Context) {
 	// TODO: Determine whether or not the schema has changed. If it has, then all rows should count as modifications in the diff.
-	considerAllRowsModified := false
-	err := prolly.DiffMaps(ctx, itr.from, itr.to, considerAllRowsModified, func(ctx context.Context, d tree.Diff) error {
+	cb := func(ctx context.Context, d tree.Diff) error {
 		dItr, err := itr.makeDiffRowItr(ctx, d)
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>84932a3</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>84932a3</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>630454d</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>630454d</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>53b40b3</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>53b40b3</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>d570c8b</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>d570c8b</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/index/dolt_index.go
```diff
@@ -197,19 +197,32 @@ func DoltDiffIndexesFromTable(ctx context.Context, db, tbl string, t *doltdb.Tab
...
+		indexes = append(indexes, &doltIndex{
+			id:                            "commits_" + toFrom,
+			tblName:                       doltdb.DoltCommitDiffTablePrefix + tbl,
+			columns:                       cols,
+			unique:                        true,
+			comment:                       "",
+			order:                         sql.IndexOrderNone,
+			constrainedToLookupExpression: false,
+			// We pass a nil ValueStore into the key builder, because we don't have one. This would cause an issue
+			// if any of the columns uss Adaptive encoding, but that shouldn't be possible in the primary key.
```
fulghum:
```suggestion
			// if any of the columns use Adaptive encoding, but that shouldn't be possible in the primary key.
```

go/libraries/doltcore/sqle/dtables/diff_iter.go
```diff
@@ -260,6 +261,18 @@ func newProllyDiffIter(ctx *sql.Context, dp DiffPartition, targetFromSchema, tar
 		}
 	}
 
+	var ranges []prolly.Range
+	var err error
+	if lookup.Index != nil {
+		for i := range lookup.Ranges.(sql.MySQLRangeCollection) {
+			lookup.Ranges.(sql.MySQLRangeCollection)[i] = lookup.Ranges.(sql.MySQLRangeCollection)[i][2:]
```
fulghum, line 270: (minor) Might be worth a comment here explaining we're pulling out the to_commit and from_commit filters. It's not super obvious just from looking at this code.

integration-tests/bats/sql-commit-diff.bats
```diff
@@ -0,0 +1,91 @@
...
+CALL DOLT_BRANCH('update');
+SQL
+}
+
+teardown() {
+    assert_feature_version
+    teardown_common
+}
+
+@test "sql-commit-diff: DOLT_COMMIT_DIFF with a range on to_ key" {
```
fulghum, line 48: (minor) probably a little nicer to have these in Go tests, since those tend to run faster and get run more often during local dev. `ScriptTestAssertion.ExpectedIndexes` should make it pretty easy to assert index usage. Not a huge deal either way though.

integration-tests/bats/sql-commit-diff.bats
```diff
@@ -0,0 +1,91 @@
...
+    [[ "$output" =~ "| 20      | NULL  | removed   |" ]] || false
+    [[ "$output" =~ "| 21      | 21    | modified  |" ]] || false
+    run dolt sql -q "describe plan select * from dolt_commit_diff_test where from_commit = DOLT_HASHOF('create') and to_commit = DOLT_HASHOF('head') and from_pk > 0 and from_pk < 5;"
+    [ $status -eq 0 ]
+    [[ "$output" =~ "index: [dolt_commit_diff_test.to_commit,dolt_commit_diff_test.from_commit,dolt_commit_diff_test.from_pk]" ]] || false
+    [[ "$output" =~ "(0, 5)" ]] || false
+
+    run dolt sql -q "select from_pk1, from_pk2, to_pk1, to_pk2, diff_type from dolt_commit_diff_test_two_pk where from_commit = DOLT_HASHOF('initial') and to_commit = DOLT_HASHOF('update') and from_pk1 = 2;"
+    [ $status -eq 0 ]
+    echo "$output"
```
fulghum:
```suggestion
```

fulghum (approved):
Looks good!

integration-tests/bats/sql-commit-diff.bats
nicktobey: Oh, I agree. Unfortunately we want to be able to assert that not only is the index being used, but that the lookup has the correct range. ExpectedIndexes only lets us assert the name of the index. Our best other option is Plan tests, but those require us to match the output of DESCRIBE PLAN exactly, which we can't do here because the query includes commit hashes that change each run.

In lieu of a better option, I decided to just use bats tests.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b1e0f9c</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b1e0f9c</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/dtables/diff_iter.go
nicktobey: Added

Issue #483, opened 2020-03-20
Unable to checkout table if branch name and table name are identical


coffeegoddd:
Skipped bats test for this included here:  https://github.com/liquidata-inc/dolt/pull/482

After modifying a table, `dolt status` outputs:
`(use "dolt checkout <table>" to discard changes in working directory)`

If a branch and table share the same name, if you modify the table, then want to use `dolt checkout` to "discard changes", Dolt thinks the branch is being referenced, outputting:
`Already on branch <branch>`

datatalking:
While this sounds good on the face, in my experience you usually don't want to combine or mix table names and branch name schema... but I'm curious as to your case.

Can you give me some cases on why this would be a good idea, seems like you wouln'dnt want to have dual names so seems like we would be causing a bigger problem than solving one.

timsehn:
I think we want to do the exact same thing as Git in this circumstance.

Yes. Bad idea. But we need to match what the reference implementation does :-)

bpf120:
@datatalking Thanks for commenting. Are you looking at Dolt for any projects? We like learning what potential users are up to. :) Happy to chat over [email](mailto:brianf@dolthub.com) or on our [Discord](https://discord.gg/gqr7K4VNKe).

codeaucafe:
Any case this still needs a fix? Happy to help out if so.

nicktobey:
I think this has already been fixed.

If you want to drop changes from the working set but not staging, you can do `dolt checkout -- tablename`, which should work even if the table name matches a branch name.

If you want to drop changes from both staging and the working set, you can do `dolt checkout HEAD -- tablename`.

This distinction is confusing, but it matches Git's behavior.

nicktobey:
Actually, looking at this more closely I don't think this is fixed.

`dolt checkout` seems to not be implemented correctly. It ignores `--` completely, and always interprets the first other parameter as a branch name (or other refspec).

`dolt reset` has similar issues, but worse. It didn't seem to handle the case of providing both a branch/refspec and a list of tables at all: If multiple parameters are passed, it simply resets every table to match HEAD, which is clearly not the intended behavior.

@codeaucafe if you still want to tackle this, by all means.

nicktobey:
This might be tricky to get right.

Some context you might need:

The command is implemented in a file named `dolt_checkout.go`

Assuming we want to match Git's behavior (@coffeegoddd feel free to push back if you don't think we should), then the intended behavior is as follows:

[...]

PR #9436, opened 2025-07-02
Improve `dolt checkout` error handling for missing tables

### LINKED ISSUE #9392: `dolt checkout test0 test1` fails all if one is missing, but git checks out available ones
When running `dolt checkout test0 test1`, if `test0` does not exist, `dolt` prevents `test1` from being checked out as well. In contrast, Git processes each path independently: if one path is missing, it checks out the others that are valid.

`test1` should be checked out, and only `test0` should cause an error/warning. For further clarification these two elements would be tables.

AsheeHuang:
Fix #9392 

 ###  Summary

  Enhances the dolt_checkout stored procedure to handle missing table specifications
   more gracefully by processing valid tables while providing detailed warnings for
  invalid ones, instead of failing immediately on the first missing table.

###   Changes Made

  - Enhanced Error Handling: Modified checkoutTablesFromHead() to collect warnings
  for missing tables instead of failing fast
  - Improved User Experience: Valid tables are now processed successfully even when
  some specified tables don't exist
  - Better Error Reporting: Multiple missing tables now show aggregated error
  messages
  - Added Test Coverage: Comprehensive tests for single and multiple missing table
  scenarios

###   Behavior Changes

  Before: call `dolt checkout valid_table missing_table` would fail entirely
  After: call `dolt_checkout valid_table missing_table` processes valid_table
  and returns warning about missing_table

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
@@ -17,6 +17,7 @@ package dprocedures
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/dolthub/go-mysql-server/sql"
 	"github.com/dolthub/go-mysql-server/sql/types"
@@ -664,6 +665,7 @@ func doGlobalCheckout(ctx *sql.Context, branchName string, isForce bool, isNewBr
 // of HEAD commit
 func checkoutTablesFromHead(ctx *sql.Context, roots doltdb.Roots, name string, tables []string) error {
 	var tableNames []doltdb.TableName
+	var warningMsg strings.Builder
 	var err error
 
 	if len(tables) == 1 && tables[0] == "." {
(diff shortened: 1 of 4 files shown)


go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -664,6 +665,7 @@ func doGlobalCheckout(ctx *sql.Context, branchName string, isForce bool, isNewBr
 // of HEAD commit
 func checkoutTablesFromHead(ctx *sql.Context, roots doltdb.Roots, name string, tables []string) error {
 	var tableNames []doltdb.TableName
+	var warningMsg []string
```
elianddb: Change this to use Go's strings.Builder

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -672,33 +674,48 @@ func checkoutTablesFromHead(ctx *sql.Context, roots doltdb.Roots, name string, t
 			return err
 		}
 	} else {
-		tableNames = make([]doltdb.TableName, len(tables))
```
elianddb, line 675: You can set the capacity to `len(tables)` to avoid reconstructions with an initial size of 0

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
```diff
@@ -3486,6 +3486,30 @@ var DoltCheckoutScripts = []queries.ScriptTest{
...
+	{
+		Name: "dolt_checkout multiple tables with one missing",
+		SetUpScript: []string{
+			"create table t1 (pk int primary key, c int);",
+			"call dolt_commit('-Am', 'created table');",
+			"insert into t1 values (1,1);",
+			"call dolt_commit('-am', 'add row');",
+			"insert into t1 values (2,2);",
+		},
+		Assertions: []queries.ScriptTestAssertion{
```
elianddb, line 3498: Also provide a test with the missing tablespec in the front of the tables list, e.g., `dolt checkout missing t1 missing2`. You'll have to use `--` operator at the moment to disambiguate since Dolt isn't defaulting to tablespec but finding the branch instead.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
```diff
@@ -672,33 +674,48 @@ func checkoutTablesFromHead(ctx *sql.Context, roots doltdb.Roots, name string, t
...
+			return err
 		}
 	}
 
-	dSess := dsess.DSessFromSess(ctx.Session)
-	return dSess.SetRoots(ctx, name, roots)
+	if len(warningMsg) > 0 {
+		return errors.New(strings.Join(warningMsg, "\n"))
+	}
+	return nil
```
elianddb, line 719: Returning nil, no error, is technically correct? Git's behavior can be a bit weird depending on whether or not you modified an already tracked item here. I think it should be fine but CI will have the final say.

go/libraries/doltcore/sqle/enginetest/dolt_queries.go
elianddb: Also would update the bats files with similar tests to make sure the cli is handling things correctly. #483 has the names for them.

elianddb:
Some minor comments but looks good so far, I'll come back to this later. Also, I would recommend generally to use a separate branch and rebase so your changes to sit on top of main for best practices, but it doesn't matter as much in this scenario.

AsheeHuang:
Hi @elianddb.

I've revised some of my code based on your advice.
Thanks

elianddb:
LGTM! Just add or amend bats test and you should be good. You can find these under `dolt/integration-tests/bats` @AsheeHuang

AsheeHuang:
@elianddb,

I have added  bats test for missing table checkout.
If there are still any issues, please let me know.

elianddb (approved):
LGTM!

PR #9458, opened 2025-07-07
[no-release-notes] go: store/nbs: file_table_reader.go: Fix clone() so that it actually uses the reference count.

reltuk:
A bug in the usage of dynassert meant that clone() was always opening a new table file, instead of using the existing open file.

There were a few known consequences:

* the window of opportunity when unlocked reads could try to read a closed chunk source and receive an error was much larger than typical. As a result, auto_gc on a standby replica in a cluster was much more likely to fail spuriously.

* the number of file descriptors a running server would churn was higher than typical, and some operations were a bit slower / more prone to context switches than they needed to be.

This PR also updates some errors to be slightly more verbose about their origin.

This PR also updates TestAutoGC in go-sql-server-tests to run less iterations in CI and in -test.short mode. The iterations should be able to be lower since auto GC will fail less often now.

### DIFF
--- a/go/store/nbs/file_table_reader.go
+++ b/go/store/nbs/file_table_reader.go
@@ -196,8 +196,8 @@ type fileReaderAt struct {
 }
 
 func (fra *fileReaderAt) clone() (tableReaderAt, error) {
-	if dynassert.Assert(atomic.AddInt32(fra.cnt, 1) > 1, "attempt to clone a closed fileReaderAt") {
-		// Restore previous refcnt, even know we're in a weird state...
+	if !dynassert.Assert(atomic.AddInt32(fra.cnt, 1) > 1, "attempt to clone a closed fileReaderAt") {
+		// Restore previous refcnt, despite being in a weird state...
 		atomic.AddInt32(fra.cnt, -1)
 		return newFileReaderAt(fra.path)
 	}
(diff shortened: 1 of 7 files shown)


go/store/nbs/file_table_reader.go
```diff
@@ -196,8 +196,8 @@ type fileReaderAt struct {
 }
 
 func (fra *fileReaderAt) clone() (tableReaderAt, error) {
-	if dynassert.Assert(atomic.AddInt32(fra.cnt, 1) > 1, "attempt to clone a closed fileReaderAt") {
-		// Restore previous refcnt, even know we're in a weird state...
+	if !dynassert.Assert(atomic.AddInt32(fra.cnt, 1) > 1, "attempt to clone a closed fileReaderAt") {
```
macneale4, line 199: yikes

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>174b10e</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>174b10e</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #9488, opened 2025-07-09
Optimize dolt_diff_ tables for lookups on primary key ranges.

nicktobey:
The `dolt_diff_` tables expose an index that matches the primary keys of the underlying table. We even already have tests that this index gets chosen for select expressions that limit the key to a specific value or range.

However, in practice this index wasn't actually used.

As of this PR, the index will actually be used to optimize the ranges that are scanned at each commit.

I'm not sure of an obvious way to test this: ordinarily I'd add a test that a query selects the index and returns the correct answer (and I added such a test for the index on the `from_` keys)... but the engine was already using the index and then discarding it.

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_query_plans.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_query_plans.go
@@ -39,6 +39,15 @@ var DoltDiffPlanTests = []queries.QueryPlanTest{
 			"",
 	},
 	{
+		Query: `select * from dolt_diff_one_pk where from_pk>=10 and from_pk<=100`,
+		ExpectedPlan: "Filter\n" +
+			" ├─ ((dolt_diff_one_pk.from_pk >= 10) AND (dolt_diff_one_pk.from_pk <= 100))\n" +
+			" └─ IndexedTableAccess(dolt_diff_one_pk)\n" +
+			"     ├─ index: [dolt_diff_one_pk.from_pk]\n" +
+			"     └─ filters: [{[10, 100]}]\n" +
+			"",
+	},
+	{
 		Query: `select * from dolt_diff_two_pk where to_pk1=1`,
 		ExpectedPlan: "Filter\n" +
 			" ├─ (dolt_diff_two_pk.to_pk1 = 1)\n" +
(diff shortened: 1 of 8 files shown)


coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>ab0352f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>ab0352f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/enginetest/dolt_query_plans.go
```diff
@@ -38,6 +38,15 @@ var DoltDiffPlanTests = []queries.QueryPlanTest{
 			"     └─ filters: [{[10, 100]}]\n" +
 			"",
 	},
+	{
+		Query: `select * from dolt_diff_one_pk where from_pk>=10 and from_pk<=100`,
```
fulghum, line 42: Would be nice to have similar query plan tests for the `to_pks` index, too, just to make sure that one is also working correctly.

fulghum (approved):
Looks good! 

Might be worth a test with a left join, [similar to what we added recently](https://github.com/dolthub/dolt/pull/9465/files#diff-5e4fe8a7a80566910d5c54e08ad17cdaf8652f175db7872fbf1db9a47c1ef2a3R5418). I didn't see the same issue in the code, but wouldn't hurt to add one test just to make sure we didn't make the same miss again somehow.

coffeegoddd:
@nicktobey __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>34bba42</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>34bba42</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/enginetest/dolt_query_plans.go
nicktobey: That issue was caused by the dolt_commit_diff_ table lookup modifying the lookup to strip the to_commit_ and from_commit_ ranges. dolt_diff_ doesn't modify the lookup, so it shouldn't have the same issue.

PR #9490, opened 2025-07-10
Unskip feature version bats test that now passes

timsehn:
It looks like someone has fixed this feature version issue and Codex identified it:

https://github.com/dolthub/dolt/issues/6303


timsehn:
Nick reviewed this and approved.

PR #9562, opened 2025-07-22
go: sqle: auto_gc.go: Add some simple ergonomics to Auto GC.

reltuk:
Previously auto gc would run anytime the store had grown 128MB. This change adds some simple heuristics which are used to keep auto gc from running when it is unlikely to collect much relative to the amount of work it will need to do.

In particular, the heuristics added are:

1) After GC has run, we will not run another GC until as much time as elapsed since the end of the GC as it took to run the GC itself.

2) After GC has run, we will not run another GC until the store has grown in size by at least as many bytes as the new gen contained at the end of the last successful GC run.

### DIFF
--- a/go/libraries/doltcore/sqle/auto_gc.go
+++ b/go/libraries/doltcore/sqle/auto_gc.go
@@ -72,7 +72,7 @@ func NewAutoGCController(arcLevel chunks.GCArchiveLevel, lgr *logrus.Logger) *Au
 // submit one dolt_gc request at a time.
 type autoGCWork struct {
 	db   *doltdb.DoltDB
-	done chan struct{}
+	done chan *gcWorkReport
 	name string // only for logging.
 }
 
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6b87665</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6b87665</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a98ce8f</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a98ce8f</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/sqle/auto_gc.go
```diff
@@ -295,9 +313,33 @@ const checkInterval = 1 * time.Second
 const size_128mb = (1 << 27)
 const defaultCheckSizeThreshold = size_128mb
 
+func shouldRequestGC(currSz, lastSz doltdb.StoreSizes, lastGcReport *gcWorkReport, now time.Time) bool {
+	grew := currSz.TotalBytes > lastSz.TotalBytes
+	growth := currSz.TotalBytes - lastSz.TotalBytes
+	if lastGcReport == nil || lastGcReport.err != nil {
```
nicktobey, line 319: How can grew be true if there's no last report? Is lastSz necessarily 0 in that case?

nicktobey (approved):

go/libraries/doltcore/sqle/auto_gc.go
reltuk: `lastSz` always represents an actual size we saw. Immediately after a GC, or immediately after start up, we populate it with the first thing we see on the tick.

The reason `grew` exists separately from `growth` is just that these are uint64s, and so if `curSz` is less than `lastSz` for some reason (manual GC?), we don't want to treat it as if the store grew massively instead of shrunk a bit.

PR #9587, opened 2025-07-25
Bug fix: Update secondary index correctly during a merge with a schema change

fulghum:
During a merge, Dolt applies diffs to existing indexes to sync their data with the changes being merged in. If the table schema has changed between the merge base and the left/right branches of the merge, Dolt was using the incorrect value tuple to remove the row from the index. This resulted in a customer issue where the secondary index still contained rows and the foreign key validation logic was incorrectly detecting FK violations, even though the child table was empty. 

As a follow-up to this, we need to investigate other similar cases in the secondary index merger and see if other diff types have a similar issue.

### DIFF
--- a/go/libraries/doltcore/merge/violations_fk_prolly.go
+++ b/go/libraries/doltcore/merge/violations_fk_prolly.go
@@ -367,7 +367,10 @@ func createCVsForPartialKeyMatches(
 
 	kb := val.NewTupleBuilder(primaryKD, primaryIdx.NodeStore())
 
-	for k, _, err := itr.Next(ctx); err == nil; k, _, err = itr.Next(ctx) {
+	for k, _, err := itr.Next(ctx); err != io.EOF; k, _, err = itr.Next(ctx) {
+		if err != nil {
+			return err
+		}
 
 		// convert secondary idx entry to primary row key
 		// the pks of the table are the last keys of the index
@@ -390,13 +393,19 @@ func createCVsForPartialKeyMatches(
 			return err
 		}
 
+		// If a value wasn't found, then there is a row in the secondary index
+		// that can't be found in the primary index. This is never expected, so
+		// we return an error.
+		if value == nil {
+			return fmt.Errorf("unable to find row from secondary index in the primary index, with key: %v", primaryIdxKey)
+		}
+
+		// If a value was found, then there is a row in the child table that references the
+		// deleted parent row, so we report a constraint violation.
 		err = receiver.ProllyFKViolationFound(ctx, primaryIdxKey, value)
 		if err != nil {
 			return err
 		}
-	}
-	if err != nil && err != io.EOF {
-		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>0b36000</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>0b36000</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/merge/violations_fk_prolly.go
```diff
@@ -367,8 +367,8 @@ func createCVsForPartialKeyMatches(
 
 	kb := val.NewTupleBuilder(primaryKD, primaryIdx.NodeStore())
 
-	for k, _, err := itr.Next(ctx); err == nil; k, _, err = itr.Next(ctx) {
-
+	var k val.Tuple
+	for k, _, err = itr.Next(ctx); err == nil; k, _, err = itr.Next(ctx) {
```
nicktobey: Is this change necessary? It doesn't look like we use k after the loop.

fulghum: I'll add a comment to explain what's happening there... it was hiding in there... 

Previously, the code was assigning `itr.Next(ctx)` to two new vars, `k` and `err`. However... right after that loop, there is an `if err != nil` check that is meant to check the error from `itr.Next(ctx)`, but since that var is shadowed, it wasn't ever doing anything.

fulghum:
> It looks like there are two different issues fixed here: trying to remove the wrong key, and reporting conflicts when a key doesn't actually exist in the primary index. But there's only one test.
> 
> What is the impact of the latter change? Are there still cases where it's possible for the computed index key to not exist in the primary index? If so, how are those cases affected by this change? If not, should we return an error when value is nil, because we expect that to never happen?

Initially, I found the issue where we weren't finding a matching row in the index, but were still creating a constraint violation (with an invalid tuple that caused the table to error when read). I fixed that with the check you see there, but then kept digging deeper to see why the index had that data that couldn't be found in the primary key. 

I decided to keep that change to avoid writing nil entries to the constraint violation table, but you're right that we can't trigger that case anymore since the root cause is fixed now. Returning an error instead of just skipping creating the violation is a good idea, since we don't expect that to happen and if it does, we should dig in and debug.

go/libraries/doltcore/merge/violations_fk_prolly.go
nicktobey: I think a cleaner and less confusing fix might just be to add an err check at the first line of the loop, breaking if err == `io.EOF` and returning the err otherwise. This should avoid the shadowing issues entirely. WDYT?

nicktobey (approved):

coffeegoddd:
@fulghum __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c453ba5</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c453ba5</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #9610, opened 2025-07-30
[no-release-notes] proto/Makefile: Fix the generation of _grpc.pb.go files.

reltuk:
With the removal of the circular symlink in go/gen/proto/github.com, we need grpc-go_out path to be source relative as well.


elianddb (approved):
LGTM!

Issue #8879, opened 2025-02-21
`dolt diff` shows NULL values for virtual generated columns


nicktobey:
Repro steps:

```
create table test(pk int primary key, c0 int generated always as (pk + 1));
insert into test(pk) values (1), (2);
```

`dolt diff` will produce the following output:

```
diff --dolt a/test b/test
added table
+CREATE TABLE `test` (
+  `pk` int NOT NULL,
+  `c0` int GENERATED ALWAYS AS ((`pk` + 1)),
+  PRIMARY KEY (`pk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_bin;
+---+----+------+
|   | pk | c0   |
+---+----+------+
| + | 1  | NULL |
| + | 2  | NULL |
+---+----+------+
```

This is likely because `c0` is virtual, so the tuples in the table don't actually contain a value for it. Displaying the correct values would require evaluating the stored expression for that column, which requires a SQL engine.

As for the correct behavior here, I'm leaning towards excluding virtual generated columns from `dolt diff` entirely, since they aren't actually part of the data in the diff.

astorig:
hi, I would like to take this issue. could you assign it to me?

nicktobey:
Sure! Thank you for your time!

This likely happened because Dolt previously didn't support virtual columns. And adding virtual columns complicates things.

When there are no virtual columns, there's no difference between "table rows stored on disk" and "table rows seen by the engine". But once a table has virtual columns, that distinction matters, and any code that assumes that the two are identical may get incorrect results.

In this case, it's likely that the `dolt diff` code is reading the rows directly out of storage and assuming that they match the table schema. Due to the way we store rows, adding a `NULL` to the end of a row doesn't change the bytes on disk: so for example, the row `(1, NULL)` and the row `(1,)` have the exact same bytes. So `dolt diff` is likely reading the latter and treating it like it's the former.

I think that the best solution here is to simply omit any virtual generated columns from the output, since computing the correct value for the column would be more complicated and potentially slow.

[...]

PR #9653, opened 2025-08-11
refactor(diff): exclude virtual generated columns from dolt diff output

codeaucafe:
Virtual generated columns previously appeared in dolt diff with NULL values because they are computed columns not stored in table data. These columns are now completely excluded from diff output since they don't represent actual stored data differences.

Changes:
- Modified getColumnNames() to filter out columns with Virtual=true
- Updated test cases to verify virtual columns are excluded
- Added negative tests to confirm virtual columns are inaccessible

Refs: #8879


codeaucafe:
FYI I will work on fixes for the failing years tomorrow.

codeaucafe:
update: will work on fixing tests accordingly once get some free time from life and work this weekend

macneale4:
running ci here: https://github.com/dolthub/dolt/pull/9695

macneale4:
Apparently we have a sqllite download failure due to being a robot. We'll ignore the Test Benchmark Runner test for the time being.

macneale4:
See @nicktobey 's comments here: https://github.com/dolthub/dolt/issues/8879#issuecomment-3198656974

This is a bit harder than we thought. Thanks for taking a crack at it!

codeaucafe:
Thanks @macneale4. I'll close this PR for now, and circle back with some questions later this week(end) when I get some free time.

codeaucafe:
Closed give implementation needs to be different. See corresponding #8879 issue comments

PR #9694, opened 2025-08-18
chore(workflows): remove `cd-bump-homebrew`

chenrui333:
[cd-bump-homebrew](https://github.com/dolthub/dolt/actions/workflows/cd-bump-homebrew.yaml) has not been running for quite some time, and the release bumps are [handled on the homebrew side](https://github.com/Homebrew/homebrew-core/pulls?q=is%3Apr+dolt), thus removing the workflow.


fulghum:
Thanks for the PR @chenrui333!

PR #9706, opened 2025-08-19
[RFC] conjoin selection

macneale4:
The existing chooseConjoinees behavior doesn't seem to behave like the documentation says it should. Specifically, we will conjoin all tables aggressively if they are all very similar sizes (`sum <= next` is false so we never break from the loop).

This change doesn't pass CI because there are tests which depend on the old behavior. This makes me wonder if the description of the behavior is incorrect or the tests are incorrect. Feedback appreciated.

### DIFF
--- a/go/store/nbs/conjoiner.go
+++ b/go/store/nbs/conjoiner.go
@@ -50,9 +50,16 @@ func (c inlineConjoiner) conjoinRequired(ts tableSet) bool {
 	return ts.Size() > c.maxTables && len(ts.upstream) >= 2
 }
 
-// chooseConjoinees implements conjoinStrategy. Current approach is to choose the smallest N tables which,
-// when removed and replaced with the conjoinment, will leave the conjoinment as the smallest table.
-// We also keep taking table files until we get below maxTables.
+// chooseConjoinees picks tables to conjoin. There are two primary things to balance here. First, we want to
+// change as few tables as possible. Second, we want to normalize table sizes so we don't have any outliers.
+//
+// Strategy is to take the smallest tables, and conjoin them until we have: Fewer than |maxTables| tables left OR the
+// size of the final conjoined table would exceed the size of the smallest unconjoined table.
+//
+// It is easily possible that conjoining the two smallest tables will result in a table that is larger than the third,
+// and in that case we stop and conjoin only those two tables. The idea being that we want to improve the situation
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8b1d2cf</TD>
<TD>ok</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8b1d2cf</TD>
<TD>5937457</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

macneale4:
To give an example of why I think this is busted, we have two tests which are virtually identical in input, but produce very different outcomes:

```
		{"log, first two", []uint32{1, 2, 4, 8, 16, 32, 64}, []uint32{3, 4, 8, 16, 32, 64}},
		{"log, all", []uint32{2, 3, 4, 8, 16, 32, 64}, []uint32{129}},
```

Line 198 in conjoiner_test.go.

macneale4:
Worth noting, all of the unit tests set maxTables to 0, which never happens in real code. So the following conditional makes the unit tests not represent real usage very well.

https://github.com/dolthub/dolt/blob/3cb6c1fba69154ffee4a13cd3e57b18cbc335796/go/store/nbs/conjoiner.go#L69

macneale4:
Last commit fixes the tests. The maxTables == 0 check in the impl code should never be there IMO.

That commit should fix the tests, but the question remains if this is the right behavior.

coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>24f1491</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>24f1491</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #9832, opened 2025-09-12
Reverted Extended encoding from adaptive to ref

Hydrocharged:
While working on importing dumps in Doltgres, I found out that data wasn't being written correctly in some circumstances. Turns out that the new adaptive encoding is incomplete, and therefore broken, and any columns over the inline threshold will be truncated. This truncation results in a panic whenever the field is read, since the serialized size does not match the data, causing OOB access.

This should be safe to change since existing databases will simply swap to using the old ref approach until adaptive is fixed, where we can then swap back to the adaptive encoding (since the ref code is still in the codebase). I've already spoken to Nick who will lead the fix once he has returned next week.

### DIFF
--- a/go/libraries/doltcore/schema/serial_encoding.go
+++ b/go/libraries/doltcore/schema/serial_encoding.go
@@ -30,8 +30,8 @@ func EncodingFromSqlType(typ sql.Type) serial.Encoding {
 		case sql.ExtendedTypeSerializedWidth_64K:
 			return serial.EncodingExtended
 		case sql.ExtendedTypeSerializedWidth_Unbounded:
-			// Always uses adaptive encoding for extended types, regardless of the setting of UseAdaptiveEncoding below.
-			return serial.EncodingExtendedAdaptive
+			// TODO: should use serial.EncodingExtendedAdaptive, but it's currently broken
+			return serial.EncodingExtendedAddr
 		default:
 			panic(fmt.Errorf("unknown serialization width"))
 		}


coffeegoddd:
@Hydrocharged __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b468f17</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b468f17</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/schema/serial_encoding.go
```diff
@@ -30,8 +30,8 @@ func EncodingFromSqlType(typ sql.Type) serial.Encoding {
 		case sql.ExtendedTypeSerializedWidth_64K:
 			return serial.EncodingExtended
 		case sql.ExtendedTypeSerializedWidth_Unbounded:
-			// Always uses adaptive encoding for extended types, regardless of the setting of UseAdaptiveEncoding below.
-			return serial.EncodingExtendedAdaptive
+			// TODO: should use serial.EncodingExtendedAdaptive, but it's currently broken
+			return serial.EncodingExtendedAddr
```
zachmu, line 34: Doesn't this break lots of other things? Like we can't use TEXT as a key without this

PR #9843, opened 2025-09-15
fix archive int overflow

macneale4:
A integer overflow bug introduced by the mmap work. The bug was in the suffix offset calculation, which took a uint32 and multiplied it by 12 (suffix len), which resulted in the overflow.

This type of bug cropped up a couple times during archive development, but this is the first time we've regressed here. This only manifests if you have more than 357,913,942 chunks in an archive, which is a lot. So many that it would take days run an automated test to find it in CI.

### DIFF
--- a/go/libraries/utils/file/mmap.go
+++ b/go/libraries/utils/file/mmap.go
@@ -38,11 +38,11 @@ type MmapData struct {
 
 var mmapAlignment = int64(os.Getpagesize())
 
-func (m MmapData) GetUint64(offset int64) uint64 {
+func (m MmapData) GetUint64(offset uint64) uint64 {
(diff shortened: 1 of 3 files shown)


coffeegoddd:
@macneale4 __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e289a04</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e289a04</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

Issue #5861, opened 2023-05-03
`dolt_diff_summary` system table should hide tables ignored by `dolt_ignore`


nicktobey:
This will bring it in line with the behavior seen by `dolt diff`.

timsehn:
It would be nice to have a simple repro in here @nicktobey.

nicktobey:
It looks like DOLT_DIFF requires a single table name, so I'm not sure it makes sense to factor in dolt_ignore here. 

DOLT_DIFF_STAT can diff all tables (and accepts "WORKING" as an input), but only shows stats for tables that exist in both revisions. Because dolt_ignore is only supposed to filter tables that were added in the working set, it shouldn't affect the behavior of DOLT_DIFF_STAT either.

So the only function we need to consider is DOLT_DIFF_SUMMARY.  A repro would look like this:

Setup:

```sql
INSERT INTO dolt_ignore VALUE ('t', true);
CREATE TABLE t (pk int);
SELECT * FROM DOLT_DIFF_SUMMARY("HEAD", "WORKING");
```

Current output:

```sql
+-----------------+---------------+-----------+-------------+---------------+
| from_table_name | to_table_name | diff_type | data_change | schema_change |
+-----------------+---------------+-----------+-------------+---------------+
|                 | t             | added     | false       | true          |
+-----------------+---------------+-----------+-------------+---------------+
```

Expected output:

```
Empty set
```

codeaucafe:
Is that fine if I take this issue to work on?

PR #9946, opened 2025-10-10
#5861 refactor: make dolt_diff_summary respect dolt_ignore patterns

codeaucafe:
## Summary
Makes `dolt_diff_summary` table function respect `dolt_ignore` patterns to match `dolt diff` command behavior.

## Changes
- Added ignore pattern filtering to `dolt_diff_summary`
- Tables matching `dolt_ignore` patterns are now filtered out
- Added system table filtering for `dolt_*` prefixed tables
- Applied filtering to both general and specific table queries
- ensure dolt_ignore table itself is still shown in dolt_diff_summary

## Testing
- Added 5 integration tests and 4 bats tests

## Questions/Misc
@macneale4 I think I got this change good for #5861, finally (sorry took me awhile, went out of country and was busy with work to jump back on it immediately). two things to call out for clarity:
- dolt_ignore table itself should be included dolt_diff_summary (like .gitignore, for example), right?
- there are two failing bats tests, I'm not sure why exactly since couldn't repro on my side locally.

Closes: https://github.com/dolthub/dolt/issues/5861

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries_diff.go
@@ -6897,11 +6897,13 @@ var QueryDiffTableScriptTests = []queries.ScriptTest{
 	{
 		Name: "dolt_diff_summary respects dolt_ignore: dropped tables",
 		SetUpScript: []string{
+			"INSERT INTO dolt_ignore VALUES ('dummy', true);", // Create dolt_ignore in initial commit
 			"CREATE TABLE will_be_ignored (pk int primary key);",
 			"CREATE TABLE will_not_be_ignored (pk int primary key);",
 			"CALL DOLT_ADD('.')",
 			"CALL DOLT_COMMIT('-m', 'add tables');",
-			"INSERT INTO dolt_ignore VALUES ('will_be_ignored', true);",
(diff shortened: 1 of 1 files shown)


macneale4:
The tests that are failing require using an internal PR. running now.

go/libraries/doltcore/sqle/dtablefunctions/dolt_diff_summary.go
```diff
@@ -438,3 +452,45 @@ func getRowFromSummary(ds *diff.TableDeltaSummary) sql.Row {
...
+}
+
+// shouldIgnoreDelta determines if a table delta should be ignored based on dolt_ignore patterns.
+// This follows the same logic as the dolt diff command: only "added" or "dropped" tables
+// can be ignored, not modified/renamed tables.
+func shouldIgnoreDelta(delta diff.TableDelta, ignorePatterns doltdb.IgnorePatterns) bool {
+	if delta.IsAdd() {
+		ignoreResult, err := ignorePatterns.IsTableNameIgnored(delta.ToName)
+		if err != nil {
+			return false // On error, don't ignore
```
macneale4, line 481: Ignoring the error isn't ideal. This would effectively mean that the regex was garbage to begin with, and the user should probably know that. Both places this method is used can bubble up errors, so I'll add a commit which does that.

PR #9955, opened 2025-10-14
Update docker-entrypoint.sh

itdependsnetworks:
I suspect you need to do more than this, as this is only the sql file, but without it, the command is too large to be sent in, and if you have a large enough file, you end up with something like this

Failed to execute /docker-entrypoint-initdb.d/nautobot.sql: /usr/local/bin/docker-entrypoint.sh: line 63: /usr/local/bin/dolt: Argument list too long

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -191,8 +191,7 @@ docker_process_init_files() {
       ;;
     *.sql)
       mysql_note "$0: running $f"
-      sql=$(cat "$f")
-      exec_mysql "$sql" "Failed to execute $f: "
+      dolt sql < "$f"
       ;;
     *.sql.bz2)
       mysql_note "$0: running $f"


elianddb:
Hello @itdependsnetworks! Could you provide an example or repro that causes such an issue? I'm currently investigating this further.

itdependsnetworks:
Drop this into your `/docker-entrypoint-initdb.d/` folder as a .sql file 

https://github.com/datacharmer/test_db/blob/master/load_employees.dump

itdependsnetworks:
long and short of it, you can see the root of the issue here: https://www.geeksforgeeks.org/linux-unix/the-argument-list-too-long-error-in-linux-commands/

elianddb:
@itdependsnetworks I got a fix up for both this issue and a couple of others users have reported regarding the entry point #9961. I'll make a release once it's reviewed.

PR #9961, opened 2025-10-14
docker-entrypoint.sh: Add missing dependencies for SQL file extraction and amend MySQL entry point behavior

elianddb:
Fixes dolthub/dolt#9955
- `docker-entrypoint.sh` now matches the order of operations of MySQL's entry point, user and database environment variables are configured before the `docker-entrypoint-init.d/` directory.
- Added missing dependencies for SQL file extraction inside of `serverDockerfile`.
- `docker-entrypoint-initdb.d` files are now piped directly into `dolt sql` to fix `Argument list too long` error when parsing a large dump.
- `docker-entrypoint-initdb.d` files now output their query results.
- Some status messages have changed to match MySQL's entry point.
- Increased timeout on `run_container` in `bats` integration tests to reduce false positives.
image

### DIFF
--- a/integration-tests/bats/docker-entrypoint.bats
+++ b/integration-tests/bats/docker-entrypoint.bats
@@ -98,8 +98,8 @@ wait_for_log() {
   cname="${TEST_PREFIX}root-env"
   docker run -d --name "$cname" -e DOLT_ROOT_PASSWORD=rootpass -e DOLT_USER=root -e DOLT_PASSWORD=anything "$TEST_IMAGE"
   wait_for_log "$cname" "cannot be used for the root user"
-  docker logs "$cname" >/tmp/${cname}.log 2>&1
-  run grep -F "cannot be used for the root user" /tmp/"${cname}".log
+  docker logs "$cname" >$BATS_TMPDIR/${cname}.log 2>&1
+  run grep -F "cannot be used for the root user" $BATS_TMPDIR/"${cname}".log
   [ $status -eq 0 ]
 }
 
@@ -107,8 +107,8 @@ wait_for_log() {
 @test "docker-entrypoint: password without user warns and is ignored" {
   cname="${TEST_PREFIX}pass-no-user"
   run_container "$cname" -e DOLT_ROOT_PASSWORD=rootpass -e DOLT_PASSWORD=orphan
-  docker logs "$cname" >/tmp/${cname}.log 2>&1
-  run grep -i "password will be ignored" /tmp/"${cname}".log
+  docker logs "$cname" >$BATS_TMPDIR/${cname}.log 2>&1
+  run grep -i "password will be ignored" $BATS_TMPDIR/"${cname}".log
   [ $status -eq 0 ]
 }
 
@@ -253,7 +253,7 @@ wait_for_log() {
 # bats test_tags=no_lambda
 @test "docker-entrypoint: valid server config file handling" {
   cname="${TEST_PREFIX}valid-server-config"
-  temp_dir="/tmp/test-config-$$"
+  temp_dir="$BATS_TMPDIR/test-config-$$"
   mkdir -p $temp_dir
   cat > $temp_dir/test.yaml << 'EOF'
 log_level: info
@@ -516,7 +516,7 @@ EOF
 @test "docker-entrypoint: docker-entrypoint-initdb.d script execution" {
   cname="${TEST_PREFIX}initdb-scripts"
 
-  local temp_dir="/tmp/initdb-test-$$"
+  local temp_dir="$BATS_TMPDIR/initdb-test-$$"
   mkdir -p "$temp_dir"
 
   cat > "$temp_dir/01-create-table.sql" << 'EOF'
@@ -531,7 +531,6 @@ EOF
 echo "Bash script executed"
 dolt sql -q "USE testinit; INSERT INTO init_test VALUES (2, 'Bash script executed');"
 EOF
-  chmod +x "$temp_dir/02-bash-script.sh"
 
   cat > "$temp_dir/03-data.sql" << 'EOF'
 USE testinit;
@@ -558,7 +557,8 @@ EOF
 @test "docker-entrypoint: .sh shell script execution" {
   cname="${TEST_PREFIX}sh-script"
 
-  local temp_dir="/tmp/sh-script-test-$$"
+  # the normal tmpdir permissions can't unset chmod -x
+  local temp_dir="$BATS_TEST_TMPDIR/sh-script-test-$$"
   mkdir -p "$temp_dir"
 
   cat > "$temp_dir/01-executable.sh" << 'EOF'
@@ -569,14 +569,13 @@ dolt sql -q "CREATE DATABASE IF NOT EXISTS shell_test_db;"
 dolt sql -q "USE shell_test_db; CREATE TABLE script_test (id INT PRIMARY KEY, message VARCHAR(100));"
 dolt sql -q "USE shell_test_db; INSERT INTO script_test VALUES (1, 'Executable script ran');"
 EOF
-  chmod +x "$temp_dir/01-executable.sh"
 
   cat > "$temp_dir/02-sourced.sh" << 'EOF'
 # Non-executable script - will be sourced
 echo "Sourcing shell script"
 dolt sql -q "USE shell_test_db; INSERT INTO script_test VALUES (2, 'Sourced script ran');"
 EOF
-  # Do NOT chmod +x: it should be sourced instead of executed
+  chmod +x "$temp_dir/01-executable.sh"
 
   run_container_with_port "$cname" 3306 -e DOLT_ROOT_HOST=% -v "$temp_dir:/docker-entrypoint-initdb.d"
 
@@ -596,8 +595,8 @@ EOF
   [ $status -eq 0 ]
   [[ "$output" =~ "2" ]] || false
 
-  docker logs "$cname" >/tmp/"$cname".log 2>&1
-  run grep -F "sourcing /docker-entrypoint-initdb.d/02-sourced.sh" /tmp/"$cname".log
+  docker logs "$cname" > $temp_dir/"$cname".log 2>&1
+  run grep -F "sourcing /docker-entrypoint-initdb.d/02-sourced.sh" $temp_dir/"$cname".log
   [ $status -eq 0 ]
 
   rm -rf "$temp_dir"
@@ -607,7 +606,7 @@ EOF
 @test "docker-entrypoint: .sql.gz gzip compressed file processing" {
   cname="${TEST_PREFIX}gzip-sql"
 
-  local temp_dir="/tmp/gzip-sql-test-$$"
+  local temp_dir="$BATS_TMPDIR/gzip-sql-test-$$"
   mkdir -p "$temp_dir"
 
   cat > "$temp_dir/01-init.sql" << 'EOF'
@@ -645,7 +644,7 @@ EOF
 @test "docker-entrypoint: .sql.bz2 bzip2 compressed file processing" {
   cname="${TEST_PREFIX}bzip2-sql"
 
-  local temp_dir="/tmp/bzip2-sql-test-$$"
+  local temp_dir="$BATS_TMPDIR/bzip2-sql-test-$$"
   mkdir -p "$temp_dir"
 
   cat > "$temp_dir/01-init.sql" << 'EOF'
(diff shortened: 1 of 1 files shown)


itdependsnetworks:
A complete nit... but one difference I have noted in MySQL vs dolt docker is that the init sql does not seem to print to stdout on dolt. 

Complete non-issue, but if you happen to know the reason would be great to see that :)

elianddb:
@itdependsnetworks We should not have that output for the init.d/ files 👍

docker/docker-entrypoint.sh
```diff
@@ -171,48 +306,42 @@ get_config_file_path_if_exists() {
...
-          mysql_error "Failed to source init script '$f'"
+          mysql_error "Failed to execute $f: "
         fi
       fi
       ;;
     *.sql)
       mysql_note "$0: running $f"
-      sql=$(cat "$f")
-      exec_mysql "$sql" "Failed to execute $f: "
+      exec_mysql "" "Failed to execute $f: " < "$f" 1
```
fulghum: Why the change to pipe the file contents in, instead of `cat`ing them and sending them in? It makes the interface for `exec_mysql` a bit harder to read.

docker/docker-entrypoint.sh
```diff
@@ -1,47 +1,172 @@
...
+#   $n (optional) - Numeric index of metadata field to use for determining log type/color
+#
+# Description:
+#   Prints a structured log line with timestamp, metadata, and message.
+#   Colors the output based on log type:
+#     Warning - yellow
+#     ERROR   - red
+#     Debug   - blue
+#   The log type is determined from the metadata field at type_index (default 0).
+#   The message is always the second-to-last argument.
```
fulghum: This isn't actually true, right?

docker/docker-entrypoint.sh
```diff
@@ -1,47 +1,172 @@
...
+  local type="${meta[$type_index]}"   # pick type for color
+
+  local color=""
   case "$type" in
-  Warn) color="\033[1;33m" ;;  # yellow
-  ERROR) color="\033[1;31m" ;; # red
-  *) color="" ;;
+  Warning) color="\033[1;33m" ;;
+  ERROR)   color="\033[1;31m" ;;
+  Debug)   color="\033[1;34m" ;;
```
fulghum: Can you make the case consistent on these?

docker/docker-entrypoint.sh
```diff
@@ -55,24 +180,34 @@ exec_mysql() {
...
 
   start_time=$(date +%s)
 
   while true; do
-    if output=$(dolt sql -q "$sql_command" 2>&1); then
-      [ "$show_output" -eq 1 ] && echo "$output"
+    if [ -n "$sql_command" ]; then
+      output=$(dolt sql -q "$sql_command" 2>&1)
+      status=$?
+    else
```
fulghum, line 96: The docs for this method should explain this – if the sql command is not specified than we read from stdin. It's a bit of a nonstandard/suprising interface to pass an empty sql statement and get a different behavior. At minimum, it should be explained in the docs so future maintainers don't have to spend time reverse engineering what's going on. Is there a way to clean this up more and avoid having to switch on an empty sql string?

docker/docker-entrypoint.sh
```diff
@@ -319,18 +443,23 @@ dolt_server_initializer() {
 
   SERVER_PID=-1
 
-  trap 'mysql_note "Caught Ctrl+C, shutting down Dolt server..."; [ $SERVER_PID -ne -1 ] && kill "$SERVER_PID"; exit 1' INT TERM
```
fulghum, line 322: I'm not convinced about changing this behavior... I like matching the MySQL docker image behavior as a general rule, but some of these changes means that Dolt will behave differently in Docker versus when running directly, which will probably surprise some customers. 

I could be convinced to keep this behavior change, but I actually think that Dolt has a much better experience than MySQL in terms of shutting down the server. It's really annoying to go find the mysql admin script to kill a process, and I think customers generally enjoy that you can CTRL+C to quit a running sql-server. 

If we've gotten customer feedback directly about this behavior, I'd be interested to hear it.

integration-tests/bats/docker-entrypoint.bats
```diff
@@ -438,31 +434,31 @@ EOF
...
+
   run docker run --rm --network host mysql:8.0 mysql -h 127.0.0.1 -P 3306 -u "$kw_user" --password="$pwd" -e "USE \`$kw_db\`; SELECT * FROM test_table;"
   [ $status -eq 0 ]
   [[ "$output" =~ "1" ]] || false
 }
 
 # bats test_tags=no_lambda
 @test "docker-entrypoint: MySQL client custom database with root and user access" {
-  cname="${TEST_PREFIX}custom-db"
+  cname="${TEST_PREFIX}custom-d"
```
fulghum: Why the change to "custom-d"?

docker/serverREADME.md
```diff
@@ -180,5 +180,6 @@ The Dolt SQL Server image supports the following environment variables:
 - `DOLT_PASSWORD` / `MYSQL_PASSWORD`: Sets the password for the user specified in `DOLT_USER`/`MYSQL_USER`
 - `DOLT_USER_HOST` / `MYSQL_USER_HOST`: Specifies a host for the custom user (default: falls back to `DOLT_ROOT_HOST`, then localhost)
 - `DOLT_SERVER_TIMEOUT`: Sets the server startup timeout in seconds (default: 300). A value of 0 means no timeout.
+- `DOLT_RAW`: When set to 1, disables `dolt sql-server` parser to display raw log format.
```
fulghum: I'm not sure I like this change to reformat the log messages. It seems messy, it's a lot of string parsing work, and it's going to be inconsistent and not work for all log messages. 

I like the spirit of matching the behavior of MySQL's docker image, but it seems confusing that Dolt server logs would have a different format in Docker than they do when running `dolt sql-server` directly. I think if we want to match MySQL's log format, we'd be better off to do it directly in the dolt binary, and not do all this extra parsing for docker. That way the logging would be consistent no matter where you run Dolt, which I think is what customers would expect. Making that change could easily break customers who are using log scanning to detect errors though, so this would be a bigger change that we'd want to do in a bigger major version bump. 

My vote is to take out the log parsing/reformatting and any reference to `DOLT_RAW`.

integration-tests/bats/docker-entrypoint.bats
```diff
@@ -558,13 +554,298 @@ EOF
...
+  docker wait "$cname"
+  docker logs "$cname" >/tmp/${cname}.log 2>&1
+  run grep -F "ERROR" /tmp/${cname}.log
+  [ $status -eq 0 ]
+
+  rm -rf "$temp_dir"
+}
+
+# bats test_tags=no_lambda
+@test "docker-entrypoint: GRANT privileges in init script does not require CREATE USER" {
```
fulghum, line 814: nice test! 😎

docker/docker-entrypoint.sh
```diff
@@ -1,47 +1,172 @@
...
 
-# mysql_log logs messages with a timestamp and optional color formatting.
+# mysql_log prints a structured log line with optional color formatting
+#
 # Arguments:
-#   $1 - Log type (e.g., Note, Warn, ERROR)
-#   $@ - Log message (if empty, reads from stdin)
+#   $1            - Timestamp (empty string "" for current time)
+#   $2..$n-1      - Metadata fields (last argument before optional type index is the message)
+#   $n (optional) - Numeric index of metadata field to use for determining log type/color
```
fulghum: What does this mean? This was a lot more clear in the previous version ("Log Type (e.g. Note, Warn, Error)"). The new description is pretty opaque and much harder to call. Logging frameworks generally specify the log level as the first argument, so it feels a bit strange to see it as an optional argument that is an opaque index into a "metadata field". Cleaning this function interface up a bit will make this a lot easier for future maintainers to work with.

integration-tests/bats/docker-entrypoint.bats
elianddb: I thought I fixed that ...

docker/docker-entrypoint.sh
elianddb: This is a limitation on Linux commands which gives the error `Argument List Too Long`.
https://www.geeksforgeeks.org/linux-unix/the-argument-list-too-long-error-in-linux-commands/

docker/docker-entrypoint.sh
elianddb: It is but it's also a result of me trying to match the MySQL logs. I'll revert this one.

docker/docker-entrypoint.sh
itdependsnetworks: `and I think customers generally enjoy that you can CTRL+C to quit a running sql-server.` agreed.

docker/docker-entrypoint.sh
elianddb: new interface exec_mysql
```sh
#   exec_mysql [--show-result] "<ERROR_MESSAGE>" "<QUERY>"
#   exec_mysql [--show-result] "<ERROR_MESSAGE>" < /docker-entrypoint-initdb.d/init.sql
#   cat /docker-entrypoint-initdb.d/init.sql | exec_mysql [--show-result] "<ERROR_MESSAGE>"
```

docker/docker-entrypoint.sh
elianddb: Sounds good I shall keep this as it was then 🙂‍↕️

integration-tests/bats/docker-entrypoint.bats
```diff
@@ -258,8 +253,9 @@ wait_for_log() {
 # bats test_tags=no_lambda
 @test "docker-entrypoint: valid server config file handling" {
   cname="${TEST_PREFIX}valid-server-config"
-  mkdir -p /tmp/test-config
-  cat > /tmp/test-config/test.yaml << 'EOF'
+  temp_dir="/tmp/test-config-$$"
```
macneale4: you shouldn't put anything in `/tmp`. It's better to use the `$BATS_TMPDIR` as the base.

```suggestion
  temp_dir="$BATS_TMPDIR/test-config-$$"
```

integration-tests/bats/docker-entrypoint.bats
```diff
@@ -558,13 +554,328 @@ EOF
   rm -rf "$temp_dir"
 }
 
+# bats test_tags=no_lambda
+@test "docker-entrypoint: .sh shell script execution" {
+  cname="${TEST_PREFIX}sh-script"
+
+  local temp_dir="/tmp/sh-script-test-$$"
```
macneale4:
```suggestion
  local temp_dir="$BATS_TMPDIR/sh-script-test-$$"
```

macneale4: I see this in more places. I'm not going to comment on the other places that need updates, but I expect you to fix them.

docker/docker-entrypoint.sh
```diff
@@ -325,12 +353,13 @@ dolt_server_initializer() {
...
       SERVER_PID=-1
-      dolt sql-server --host=0.0.0.0 --port=3306 "$@" &
+      dolt sql-server --host=0.0.0.0 --port=3306 "$@" 2>&1 &
       SERVER_PID=$!
+
     fi
 
     if is_port_open "0.0.0.0" 3306; then
-      mysql_note "Dolt server started successfully (PID=$SERVER_PID)"
+      mysql_note "Dolt server started."
```
macneale4, line 362: Is there a reason to change this?

docker/docker-entrypoint.sh
```diff
@@ -369,47 +398,42 @@ _main() {
...
+  create_database_from_env
 
   create_user_from_env
 
-  exec_mysql "SELECT User, Host FROM mysql.user;" "Could not list users: " 1
+  exec_mysql --show-result "Could not list users: " "SELECT User, Host FROM mysql.user;"
 
-  mysql_note "Server initialization complete!"
+  if [[ ! -f $INIT_COMPLETED ]]; then
+    if ls /docker-entrypoint-initdb.d/* >/dev/null 2>&1; then
```
macneale4, line 428: Changing the order of processing could lead to subtle differences in behavior from previous versions. This seems like a breaking change to me. minor version bump possible?

docker/docker-entrypoint.sh
elianddb: I've started matching the log messages from MySQL's entry point

elianddb:
@itdependsnetworks Release should be out now for this fix. Apologies for the delays, we wanted to include a couple of fixes in the minor ver. bump.

PR #9990, opened 2025-10-23
go/store/nbs: store.go: Fix a bug which caused dolt gc --full to not collect new data sometimes.

reltuk:
If `dolt gc --full` was run, and then branches were changed and `dolt gc` was run, possibly multiple times, and then all of the state was returned to the state of the database during the initial `dolt gc --full`, then running `dolt gc --full` anew would falsely conclude that there was nothing to collect.

After the GC work is done, and we are inside swapTables, there is no need to check for whether we should swap the tables. We already have done the work to compute the new files and we should apply them regardless.

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -2255,13 +2255,6 @@ func (nbs *NomsBlockStore) swapTables(ctx context.Context, specs []tableSpec, mo
 		specs:   specs,
 	}
 
-	sameGcGen := newContents.gcGen == nbs.upstream.gcGen
-	hasLocalNovelty := nbs.hasLocalGCNovelty()
-	if sameGcGen && !hasLocalNovelty {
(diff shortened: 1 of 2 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f3adba5</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f3adba5</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10024, opened 2025-11-03
go/store/nbs: planRangeCopyConjoin: Account for quota when building the merged index.

reltuk:
In contexts where memory usage is quota'd and allocation is failable, ConjoinAll should take quota when building up O(n) data structures like the merged prefix index. This change reworks planRangeCopyConjoin so that large in-memory structures are allocated through or accounted in the memory quota provider.

This change should not have an impact on Dolt's current behavior, since Dolt currently always run stoage with a non-failable memory quota provider.

### DIFF
--- a/go/store/nbs/archive_writer.go
+++ b/go/store/nbs/archive_writer.go
@@ -848,11 +848,15 @@ func planArchiveConjoin(ctx context.Context, sources []sourceWithSize, q MemoryQ
 	if err != nil {
 		return compactionPlan{}, err
 	}
-	acquiredBytes := sz
+	// We acquire bytes throughout the conjoin and need to clean
+	// them up.  In the happy case, some of these acquired bytes
+	// will actually be the callers responsibility to clean up,
+	// but acquiredBytesOnStack will have been adjusted so that we
+	// always release what we ourselves are still responsible for
+	// at the time that we return from this function.
+	acquiredBytesOnStack := sz
 	defer func() {
-		if err != nil {
-			q.ReleaseQuotaBytes(acquiredBytes)
-		}
+		q.ReleaseQuotaBytes(acquiredBytesOnStack)
 	}()
 
 	aw.stagedBytes = make(stagedByteSpanSlice, 0, numByteSpans)
@@ -877,7 +881,7 @@ func planArchiveConjoin(ctx context.Context, sources []sourceWithSize, q MemoryQ
 			if err != nil {
 				return compactionPlan{}, err
 			}
-			acquiredBytes += tempSz
+			acquiredBytesOnStack += tempSz
 			chunkRecs := make([]tableChunkRecord, 0, chks)
 			for i := uint32(0); i < chks; i++ {
 				var h hash.Hash
@@ -907,7 +911,8 @@ func planArchiveConjoin(ctx context.Context, sources []sourceWithSize, q MemoryQ
 					data:       uint32(len(aw.stagedBytes)),
 				})
 			}
-			acquiredBytes -= tempSz
+			chunkRecs = nil
+			acquiredBytesOnStack -= tempSz
 			q.ReleaseQuotaBytes(tempSz)
 		} else {
 			footer := arcSrc.aRdr.footer
@@ -977,7 +982,7 @@ func planArchiveConjoin(ctx context.Context, sources []sourceWithSize, q MemoryQ
 	if err != nil {
 		return compactionPlan{}, err
 	}
-	acquiredBytes += idxSize
+	acquiredBytesOnStack += idxSize
 	bufBytes = bufBytes[:0]
 	buf := bytes.NewBuffer(bufBytes)
 	if err := writer.Flush(buf); err != nil {
@@ -987,8 +992,10 @@ func planArchiveConjoin(ctx context.Context, sources []sourceWithSize, q MemoryQ
 	stats.BytesPerConjoin.Sample(dataBlocksLen + uint64(len(buf.Bytes())))
 
 	writer = nil
-	// Our temporary slices used for serializing the index are no longer needed.
-	q.ReleaseQuotaBytes(acquiredBytes - idxSize)
+
+	// On return, we are going to release quota bytes we are still responsible for.
+	// We are transfering responsibility for idxSize bytes to our caller.
+	acquiredBytesOnStack -= idxSize
 
 	return compactionPlan{
 		sources:             orderedSrcs,
--- a/go/store/nbs/aws_table_persister.go
+++ b/go/store/nbs/aws_table_persister.go
@@ -281,7 +281,11 @@ func (s3p awsTablePersister) assembleTable(ctx context.Context, plan compactionP
 	}
 
 	// Separate plan.sources by amount of chunkData. Tables with >5MB of chunk data (copies) can be added to the new table using S3's multipart upload copy feature. Smaller tables with <5MB of chunk data (manuals) must be read, assembled into |buff|, and then re-uploaded in parts that are larger than 5MB.
-	copies, manuals, buff, err := dividePlan(ctx, plan, s3p.limits.partMin, s3p.limits.partMax, s3p.q)
+	copies, manuals, buffSize, err := dividePlan(ctx, plan, s3p.limits.partMin, s3p.limits.partMax)
+	if err != nil {
+		return nil, err
+	}
+	buff, err := s3p.q.AcquireQuotaByteSlice(ctx, int(buffSize))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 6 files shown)


go/store/nbs/archive_writer_test.go
```diff
@@ -75,3 +75,10 @@ func TestArchiveStreamWriterRemove(t *testing.T) {
 		require.Equal(t, 1, CountFilesInDir(t, dir))
 	})
 }
+
+func TestPlanArchiveConjoin(t *testing.T) {
+	t.Run("Quota", func(t *testing.T) {
+		t.Run("NoError", func(t *testing.T) {
+		})
```
macneale4: ?

go/store/nbs/table_persister.go
```diff
@@ -199,8 +208,19 @@ func planTableConjoin(ctx context.Context, sources []sourceWithSize, stats *Stat
...
 	suffixesPos := suffixesOffset(plan.chunkCount)
-	plan.mergedIndex = make([]byte, indexSize(plan.chunkCount)+footerSize)
+	sz := int(indexSize(plan.chunkCount) + footerSize)
+	plan.mergedIndex, err = q.AcquireQuotaByteSlice(ctx, sz)
+	if err != nil {
+		return compactionPlan{}, err
+	}
+	acquiredBytes += sz
 
+	prefixRecsSz := int(unsafe.Sizeof(prefixIndexRec{})) * int(plan.chunkCount)
```
macneale4: unsafe always piques some spidey senses. The only risk I can see here is in the future someone modifies the `prefixInexRec` struct to include an interface or whatever. Can you add a comment at the struct definition point to tell people to make their changes be unsafe.Sizeof friendly?

Do the same for `byteSpan`, `tableChunkRecord` and `stagedChunkRef`

Also, I found a little trick to future proof for these things changing sizes.

```go
var _ [16 - int(unsafe.Sizeof(byteSpan{}))]struct{}
```

will fail to compile if the size of byteSpan increases. Would at least be a mechanism to ensure future devs know they are touching something that shouldn't be touched. Falls apart if they remove a field tho.

Then again, I can't honestly imagine anyone would change these structs in a dangerous way. Take all of this with a grain of salt.

coffeegoddd:
@coffeegoddd __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>3206bcb</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>3206bcb</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e75ef2e</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e75ef2e</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/table_persister.go
```diff
@@ -285,9 +304,16 @@ func planTableConjoin(ctx context.Context, sources []sourceWithSize, stats *Stat
 
 	writeFooter(plan.mergedIndex[uint64(len(plan.mergedIndex))-footerSize:], plan.chunkCount, totalUncompressedData)
 
+	prefixIndexRecs = nil
```
macneale4, line 306: I wonder if we can skip the prefixIndex alloc entirely by working on the plan.MergedIndex memory. Probably risky, so definitely fine to leave this all alone.

go/store/nbs/byte_sink_test.go
```diff
@@ -75,6 +77,73 @@ func TestBufferedFileByteSink(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.Equal(t, uint64(77824), q.Usage())
+			assert.Equal(t, 77824, buf.Len())
+			sink.Close()
+			assert.Equal(t, uint64(0), q.Usage())
+		})
+		t.Run("Error", func(t *testing.T) {
+			t.Run("Init", func(t *testing.T) {
+				q := &errorQuota{NewUnlimitedMemQuotaProvider(), 0}
+				_, err := NewBlockBufferByteSink(t.Context(), 4096, q)
```
macneale4: Nice to differentiate this call from the one three lines down.
```suggestion
				_, err := NewBlockBufferByteSink(t.Context(), 1, q)
```

go/store/nbs/byte_sink.go
```diff
@@ -118,12 +119,25 @@ type BlockBufferByteSink struct {
...
+
+	ctx context.Context
+	q   MemoryQuotaProvider
 }
 
 // NewBlockBufferByteSink creates a BlockBufferByteSink with the provided block size.
-func NewBlockBufferByteSink(blockSize int) *BlockBufferByteSink {
-	block := make([]byte, 0, blockSize)
-	return &BlockBufferByteSink{pos: 0, blockSize: blockSize, blocks: [][]byte{block}}
+func NewBlockBufferByteSink(ctx context.Context, blockSize int, q MemoryQuotaProvider) (*BlockBufferByteSink, error) {
```
macneale4, line 131: Maybe worth a comment here stating that you need to call Close() on this returned value. Not sure how to protect against leaking otherwise, and it's not immediately obvious that it's necessary.

go/store/nbs/aws_table_persister.go
```diff
@@ -461,16 +479,18 @@ func dividePlan(plan compactionPlan, minPartSize, maxPartSize uint64) (copies []
...
 	var offset int64
 	for ; i < len(plan.sources.sws); i++ {
 		sws := plan.sources.sws[i]
 		manuals = append(manuals, manualPart{sws.source, offset, offset + int64(sws.dataLen)})
 		offset += int64(sws.dataLen)
 		buffSize += sws.dataLen
 	}
-	buff = make([]byte, buffSize)
-	copy(buff[buffSize-uint64(len(plan.mergedIndex)):], plan.mergedIndex)
+	buff, err = q.AcquireQuotaByteSlice(ctx, int(buffSize))
```
macneale4: A little dodgy because you are releasing the bytes up a level in the stack. Maybe return a function which will force the caller to do the right thing? Probably fine to just leave a comment

go/store/nbs/archive_writer.go
```diff
@@ -938,19 +965,34 @@ func planArchiveConjoin(sources []sourceWithSize, stats *Stats) (compactionPlan,
...
+	buf := bytes.NewBuffer(bufBytes)
 	if err := writer.Flush(buf); err != nil {
 		return compactionPlan{}, fmt.Errorf("failed to build index buffer while conjoining archives: %w", err)
 	}
 
 	stats.BytesPerConjoin.Sample(dataBlocksLen + uint64(len(buf.Bytes())))
 
+	writer = nil
+	// Our temporary slices used for serializing the index are no longer needed.
+	q.ReleaseQuotaBytes(acquiredBytes - idxSize)
```
macneale4: Above we have:

```go
defer func() {
		if err != nil {
			q.ReleaseQuotaBytes(acquiredBytes)
```

seems like we should remove the err != nil condition, and adjust `acquiredBytes -= idxSize` here.

acuiredBytes may be better named as `stackFrameOnlyBytes` or something like that. I dunno. This all seems like it's begging for future me to mess it up.

macneale4 (approved):
There are several comments for tweeks, but it all looks correct. Fine to ship without follow up review.

go/store/nbs/table_persister.go
reltuk: We probably can. We can also definitely skip the Flush of the sink into bufBytes if we have something like https://pkg.go.dev/go4.org/readerutil#NewMultiReaderAt. But to see any improvement we would also need to get rid of the the 2x peak memory utilization when we go to construct the chunkSource from it. I'm punting on it because the 2x utilization is present in archives and table file conjoins for now either way.

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8838a20</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8838a20</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10030, opened 2025-11-05
dolt/dolthub#1430: Add `--filter` option for dolt diff

### LINKED ISSUE #1430: `dolt diff` should support `--filter` option

codeaucafe:
### TL;DR

There was no action on the original #3499 for issue #1430; the PR was closed ~3 years ago. This PR fixes the open PR comments and updates the implementation details a bit for the RowWriting of filtered rows

### Description

Adds `--filter` option to `dolt diff` for filtering by change type (`added`, `modified`, `renamed`, `dropped`, or `removed` as alias for `dropped`); please note I added renamed and dropped alias to this filter list. Completes PR #3499 by fixing the issues that blocked it from merging, plus fixes an additional bug where empty table headers were
printed when all rows were filtered out.

Users reviewing large diffs often need to focus on specific change types - deletes may need extra scrutiny while inserts are routine. With diffs spanning thousands of rows across multiple tables, grep isn't enough since updates show
 both additions and deletions.

## Implementation

**Core changes:**
- Added `diffTypeFilter` struct with validation methods in `diff.go`
- Added `--filter` argument in `arg_parser_helpers.go`
- Table filtering uses `delta.SchemaChange` and `delta.IsRename()` for modification detection
- Row filtering checks `diff.ChangeType` in `writeDiffResults()`
- Fixed control flow - uses `continue` instead of `return nil` so all tables get processed

**Empty header bug fix:**
- Implemented `lazyRowWriter` that delays `BeginTable()` until first row write
- Added `shouldUseLazyHeader()` helper to determine when lazy initialization is needed
- Only applies to filtered data-only diffs (no schema changes or renames)
- Prevents empty table headers from printing when all rows are filtered out

**Tests:**
- `go/cmd/dolt/commands/diff_filter_test.go` - comprehensive unit tests:
  - `TestShouldUseLazyHeader` - validates lazy header logic (8 test cases)
  - `TestLazyRowWriter_*` - tests lazy initialization behavior (4 test functions)
  - Mock implementations of `diffWriter` and `SqlRowDiffWriter` interfaces
- BATS integration tests for all filter types with tables, rows, and schema changes
- Edge cases: invalid values, typos, case sensitivity

## What Was Broken in PR #3499

The original PR had three blocking issues from [code review comments](https://github.com/dolthub/dolt/pull/3499) that I fixed here:

1. **No row-level filtering** - only filtered tables, not individual rows
   - Fixed by adding filtering in `writeDiffResults()`

2. **Incomplete modification detection** - didn't check for schema changes properly
   - Fixed by using `delta.SchemaChange || delta.IsRename()` as requested

3. **Loop exits early** - `return nil` prevented processing remaining tables
   - Fixed by using `continue` instead

## Additional Bug (bug from my initial filter changes; not existing prior) Fixed

**Empty table headers when filtering:**
When using `--filter` with data-only changes, if all rows were filtered out, dolt would still print the table diff header with no content:
```bash
# Before fix:
$ dolt diff HEAD~1 --filter=modified  # (table has only added rows)
diff --dolt a/t b/t
--- a/t
+++ b/t
(empty - no rows printed)

# After fix:
$ dolt diff HEAD~1 --filter=modified
(completely empty - no output at all)
```

This occurred because `BeginTable()` was called before row filtering. The `lazyRowWriter` pattern delays header output until we confirm at least one row will be written.

## Usage/Examples

```
dolt diff --filter=added      # new tables/rows
dolt diff --filter=modified   # schema changes, row updates
dolt diff --filter=renamed    # renamed tables
dolt diff --filter=dropped    # dropped tables, deleted rows
dolt diff --filter=removed    # alias for dropped
dolt diff HEAD~1 --filter=dropped -r sql
```

Fixes #1430

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -311,7 +311,7 @@ func CreateDiffArgParser(isTableFunction bool) *argparser.ArgParser {
 		ap.SupportsString(FormatFlag, "r", "result output format", "How to format diff output. Valid values are tabular, sql, json. Defaults to tabular.")
 		ap.SupportsString(WhereParam, "", "column", "filters columns based on values in the diff.  See {{.EmphasisLeft}}dolt diff --help{{.EmphasisRight}} for details.")
 		ap.SupportsInt(LimitParam, "", "record_count", "limits to the first N diffs.")
-		ap.SupportsString(FilterParam, "", "diff_type", "filters results based on the type of modification (added, modified, removed).")
+		ap.SupportsString(FilterParam, "", "diff_type", "filters results based on the type of change (added, modified, renamed, dropped). 'removed' is accepted as an alias for 'dropped'.")
 		ap.SupportsFlag(StagedFlag, "", "Show only the staged data changes.")
 		ap.SupportsFlag(CachedFlag, "c", "Synonym for --staged")
 		ap.SupportsFlag(MergeBase, "", "Uses merge base of the first commit and second commit (or HEAD if not supplied) as the first commit")


codeaucafe:
I believe the two failing bats tests are unrelated to my changes. I believe a perms issue on my part?

also the one test failing in `Test Go/Go tests (ubuntu-22.04)` is for `TestMergeConcurrency` which fails for ubuntu with race detected error but passes for mac and windows. Is this a known flaky test? If its not I can look into this later when I get back from short vacation this week

go/cmd/dolt/cli/arg_parser_helpers.go
```diff
@@ -320,6 +320,7 @@ func CreateDiffArgParser(isTableFunction bool) *argparser.ArgParser {
 		ap.SupportsString(FormatFlag, "r", "result output format", "How to format diff output. Valid values are tabular, sql, json. Defaults to tabular.")
 		ap.SupportsString(WhereParam, "", "column", "filters columns based on values in the diff.  See {{.EmphasisLeft}}dolt diff --help{{.EmphasisRight}} for details.")
 		ap.SupportsInt(LimitParam, "", "record_count", "limits to the first N diffs.")
+		ap.SupportsString("filter", "", "diff_type", "filters results based on the type of modification (added, modified, removed).")
```
elianddb: Should follow the `NameParam` naming scheme

go/cmd/dolt/commands/diff.go
```diff
@@ -60,6 +60,11 @@ const (
 	TabularDiffOutput diffOutput = 1
 	SQLDiffOutput     diffOutput = 2
 	JsonDiffOutput    diffOutput = 3
+
+	FilterAdds     = "added"
```
elianddb: Keep the same post-tense wording, ie `FilterAdded`

go/cmd/dolt/commands/diff.go
```diff
@@ -220,6 +358,11 @@ func (cmd DiffCmd) validateArgs(apr *argparser.ArgParseResults) errhand.VerboseE
 		return errhand.BuildDError("invalid output format: %s", f).Build()
 	}
 
+	filterValue, _ := apr.GetValue("filter")
+	if filterValue != "" && filterValue != FilterAdds && filterValue != FilterModified && filterValue != FilterRemoved && filterValue != NoFilter {
+		return errhand.BuildDError("invalid filter: %s. Valid values are: added, modified, removed", filterValue).Build()
+	}
```
elianddb: Use the `IsValid` func here so we only have to update one place.

go/cmd/dolt/commands/diff.go
```diff
@@ -220,6 +358,11 @@ func (cmd DiffCmd) validateArgs(apr *argparser.ArgParseResults) errhand.VerboseE
 		return errhand.BuildDError("invalid output format: %s", f).Build()
 	}
 
+	filterValue, _ := apr.GetValue("filter")
```
elianddb: You can use the const that should be added to apr above here for these magic values.

go/cmd/dolt/commands/diff.go
```diff
@@ -268,6 +411,9 @@ func parseDiffDisplaySettings(apr *argparser.ArgParseResults) *diffDisplaySettin
 	displaySettings.limit, _ = apr.GetInt(cli.LimitParam)
 	displaySettings.where = apr.GetValueOrDefault(cli.WhereParam, "")
 
+	filterValue := apr.GetValueOrDefault("filter", "all")
```
elianddb: this relates to magic apr vals

go/cmd/dolt/commands/diff.go
```diff
@@ -816,6 +962,37 @@ func diffUserTables(queryist cli.Queryist, sqlCtx *sql.Context, dArgs *diffArgs)
...
+				if !isModified && delta.DataChange {
+					shouldIncludeTable = true
+				}
+			}
+
+			if !shouldIncludeTable {
+				continue // Skip this table but continue processing others
+			}
+		}
+
```
elianddb, line 979: Use the `TableDeltaSummary` object provided. This may require rewriting your impl for the filter struct above, but it keeps the logic in one place. Specifically, there's a `DiffType` field you can make use of.

Secondly, I think this enables you to keep track of enabled filters in a map, ie str -> bool. This also keeps the diff type strings synchronized in the future. One func can handle the checks, instead of multiple include helper funcs. Of course, impl const vars to avoid the magic vals. You should define the consts in `table_deltas.go` since it already defines its own names for `DiffType`.

go/cmd/dolt/commands/diff.go
```diff
@@ -130,6 +138,136 @@ type diffStatistics struct {
...
+	isAdd, isDrop bool,
+	createWriter func() (diff.SqlRowDiffWriter, error),
+) *lazyRowWriter {
+	return &lazyRowWriter{
+		createWriter:  createWriter,
+		dw:            dw,
+		fromTableName: fromTableName,
+		toTableName:   toTableName,
+		isAdd:         isAdd,
+		isDrop:        isDrop,
```
elianddb: This information should be in the delta object already, aka `TableDeltaSummary`. Also, you could alternatively store the `beginTable` since that's your main priority.
```go
type lazyRowWriter struct {
    realWriter  diff.SqlRowDiffWriter
    beginTable  func() error
    initialized bool
}

func (l *lazyRowWriter) WriteRow(ctx *sql.Context, row sql.Row, diffType diff.ChangeType, colDiffTypes []diff.ChangeType) error {
    if !l.initialized {
        if err := l.beginTable(); err != nil {
            return err
        }
        l.initialized = true
    }
    return l.realWriter.WriteRow(ctx, row, diffType, colDiffTypes)
}
```

go/cmd/dolt/commands/diff.go
```diff
@@ -1445,10 +1622,26 @@ func diffRows(
...
+			func() (diff.SqlRowDiffWriter, error) {
+				return dw.RowWriter(fromTableInfo, toTableInfo, tableSummary, unionSch)
+			},
+		)
+	} else {
+		// We always instantiate a RowWriter in case the diffWriter needs it to close off any work from schema output
+		rowWriter, err = dw.RowWriter(fromTableInfo, toTableInfo, tableSummary, unionSch)
+		if err != nil {
+			return errhand.VerboseErrorFromError(err)
+		}
```
elianddb: Instead of providing two different branchs to initialize rowWriter differently, wrap the original with your wrapper:
```go
rowWriter, err := dw.RowWriter(fromTableInfo, toTableInfo, tableSummary, unionSch)
if err != nil {
}

if shouldUseLazyHeader(dArgs, tableSummary) {
    rowWriter = &lazyRowWriter{
        rowWriter,
        func() error {
            return dw.BeginTable(
                tableSummary.FromTableName.Name,
                tableSummary.ToTableName.Name,
                tableSummary.IsAdd(),
                tableSummary.IsDrop(),
            )
        },
    }
}
```

go/cmd/dolt/commands/diff.go
```diff
@@ -1708,6 +1901,13 @@ func writeDiffResults(
 			return err
 		}
 
+		// Apply row-level filtering based on diff type
+		if dArgs.filter != nil && dArgs.filter.filterBy != NoFilter {
+			if shouldSkipDiffType(dArgs.filter, oldRow.RowDiff) || shouldSkipDiffType(dArgs.filter, newRow.RowDiff) {
```
elianddb: This goes back to using the existing summary `DiffType` field and adding consts for the strings. Make sure to update the helper funcs. You could check the map for the necessary fields and flip using `!`.

codeaucafe:
Thanks @elianddb. Sorry for the delay. I have not been able to take a look at your comments yet due to having to work very late last couple days and helping with some family stuff. I'll be sure to review your comments and ask you any questions I have when I get time to work on addressing your comments this weekend.

Thank you!

Cheers,
David

go/cmd/dolt/commands/merge.go
```diff
@@ -480,17 +480,17 @@ func calculateMergeStats(queryist cli.Queryist, sqlCtx *sql.Context, mergeStats
 		if strings.HasPrefix(summary.TableName.Name, diff.DBPrefix) {
 			continue
 		}
-		if summary.DiffType == "added" {
+		if summary.DiffType == diff.DiffTypeAdded {
```
codeaucafe: FYI, used the new diff type constants added here so we aren't used string literals here

go/cmd/dolt/commands/diff.go
```diff
@@ -130,6 +134,133 @@ type diffStatistics struct {
...
+
+// lazyRowWriter wraps a SqlRowDiffWriter and delays calling BeginTable
+// until the first row is actually written. This prevents empty table headers
+// when all rows are filtered out.
+type lazyRowWriter struct {
+	writer diff.SqlRowDiffWriter
+
+	// Callback to invoke before first write
+	// Set to nil after first call
+	onFirstWrite func() error
```
codeaucafe: how is this looking @elianddb? 

Also, instead of having initialized just added nil check on onFirstWrite

go/cmd/dolt/commands/merge.go
codeaucafe, line 483: FYI, used the new diff type constants added here so we aren't used string literals here

go/cmd/dolt/commands/diff.go
```diff
@@ -130,6 +134,133 @@ type diffStatistics struct {
...
+// all rows are filtered out in data-only diffs.
+func shouldUseLazyHeader(dArgs *diffArgs, tableSummary diff.TableDeltaSummary) bool {
+	return dArgs.filter != nil && dArgs.filter.filters != nil &&
+		!tableSummary.SchemaChange && !tableSummary.IsRename()
+}
+
+// lazyRowWriter wraps a SqlRowDiffWriter and delays calling BeginTable
+// until the first row is actually written. This prevents empty table headers
+// when all rows are filtered out.
+type lazyRowWriter struct {
```
codeaucafe, line 224: how is this looking @elianddb?

Also, instead of having initialized just added nil check on onFirstWrite

go/cmd/dolt/commands/diff.go
```diff
@@ -86,6 +87,8 @@ The diffs displayed can be limited to show the first N by providing the paramete
...
 
+To filter diff output by change type, use {{.EmphasisLeft}}--filter <type>{{.EmphasisRight}} where {{.EmphasisLeft}}<type>{{.EmphasisRight}} is one of {{.EmphasisLeft}}added{{.EmphasisRight}}, {{.EmphasisLeft}}modified{{.EmphasisRight}}, {{.EmphasisLeft}}renamed{{.EmphasisRight}}, or {{.EmphasisLeft}}dropped{{.EmphasisRight}}. The {{.EmphasisLeft}}added{{.EmphasisRight}} filter shows only additions (new tables or rows), {{.EmphasisLeft}}modified{{.EmphasisRight}} shows only schema modifications or row updates, {{.EmphasisLeft}}renamed{{.EmphasisRight}} shows only renamed tables, and {{.EmphasisLeft}}dropped{{.EmphasisRight}} shows only deletions (dropped tables or deleted rows). You can also use {{.EmphasisLeft}}removed{{.EmphasisRight}} as an alias for {{.EmphasisLeft}}dropped{{.EmphasisRight}}. For example, {{.EmphasisLeft}}dolt diff --filter=dropped{{.EmphasisRight}} shows only deleted rows and dropped tables.
```
codeaucafe, line 90: Updated to include renamed filter, as well as using removed for alias for dropped `DiffTypeDropped`

go/cmd/dolt/commands/diff.go
```diff
@@ -130,6 +134,141 @@ type diffStatistics struct {
...
+// Pass diff.DiffTypeAll or empty string to include all types.
+// Accepts "removed" as an alias for "dropped" for user convenience.
+func newDiffTypeFilter(filterType string) *diffTypeFilter {
+	if filterType == "" || filterType == diff.DiffTypeAll {
+		return &diffTypeFilter{filters: nil} // nil means include all
+	}
+
+	// Map "removed" to "dropped" (alias for user convenience)
+	internalFilterType := filterType
+	if filterType == "removed" {
```
codeaucafe, line 155: `removed` filter value alias for DiffTypeDropped. Figure for filtering `removed` is better terminology but shouldn't be specific diff type value.

codeaucafe (commented):
@elianddb I updated the code based on your comments. Let me know if I misunderstood something and implemented it wrong.

I believe the failing bats test are unrelated and failing for same reasons those other unrelated bats tests failed for

Thank you!

go/cmd/dolt/commands/diff_filter_test.go
```diff
@@ -0,0 +1,548 @@
...
+func (m *mockRowWriter) Close(_ /* ctx */ context.Context) error {
+	m.closeCalled = true
+	return nil
+}
+
+func TestLazyRowWriter_NoRowsWritten(t *testing.T) {
+	mockDW := &mockDiffWriter{}
+	realWriter := &mockRowWriter{}
+
+	beginTableCalled := false
```
elianddb, line 405: This already exists in your `mockDW` variable, similar to how it was done in the next test function.

integration-tests/bats/diff.bats
```diff
@@ -2245,3 +2245,223 @@ EOF
...
+    [[ $output = '' ]] || false
+
+    # filter=modified should not show the dropped table
+    run dolt diff HEAD~1 --filter=modified
+    [ $status -eq 0 ]
+    [[ $output = '' ]] || false
+
+    # filter=renamed should not show the dropped table
+    run dolt diff HEAD~1 --filter=renamed
+    [ $status -eq 0 ]
```
elianddb, line 2465: Awesome test coverage for Bats and Go!

elianddb (approved):
@codeaucafe lgtm! I've left a singular comment on a test, but more for future reference.

elianddb (changes requested):
Gotta run tests first, I'll come back once they're done.

elianddb:
I'm submitting on my PR #10097 to cleanup the release notes and have the CI run correctly. Looks like everything passed though! Thank you for the contribution @codeaucafe!

codeaucafe:
Thank you @elianddb !!!!!

go/cmd/dolt/commands/diff_filter_test.go
codeaucafe: Doh, sorry. Can't believe I missed that before submitting to review again

PR #10048, opened 2025-11-07
go/store/nbs: Fix race in chunk journal initialization.

reltuk:
With `dolt gc`, the store journal can be deleted and recreated as part of normal operation. When we bootstrap a new chunk journal, we always write 1MB of 0 bytes and a root hash record, reflecting the current root hash, at offset 0.

ChunkJournal takes an advisory exclusive lock when it opens the database data directory. It will operate in a read-only mode when it cannot take this lock.

Due to a bug, even when operating in read-only mode, the chunk journal bootstrap code would run if the store was requested to open the chunk journal but there were was no existing chunk journal on disk. This meant there was a race condition if a Dolt process opened the store in read-only mode while another process was going to GC the store. If the read-only store tried to open the journal file and did not find it, it could end up overwriting the beginning of the contents of a just-created journal file.

This PR adds two fixes:

1) When we create a journal file, we must be its actual creator. O_EXCL will
   cause os.Open() to return an error if this syscall did not create the file,
   which means we are not in charge and we need to bail.

2) When running in read-only mode, we should never create the journal file. We
   can try to read its contents if we learn of its existence, but if there is
   no file and we have been requested to open it, we should return an error.

### DIFF
--- a/go/store/nbs/journal.go
+++ b/go/store/nbs/journal.go
@@ -140,7 +140,9 @@ func (j *ChunkJournal) bootstrapJournalWriter(ctx context.Context) (err error) {
 		return err
 	}
 
-	if !ok { // create new journal file
+	canCreate := !j.backing.readOnly()
+
+	if canCreate && !ok { // create new journal file
 		j.wr, err = createJournalWriter(ctx, j.path)
 		if err != nil {
 			return err
--- a/go/store/nbs/journal_test.go
+++ b/go/store/nbs/journal_test.go
@@ -46,6 +46,18 @@ func makeTestChunkJournal(t *testing.T) *ChunkJournal {
 	return j
 }
 
+func openTestChunkJournal(t *testing.T, dir string) *ChunkJournal {
+	m, err := newJournalManifest(t.Context(), dir)
+	require.NoError(t, err)
+	q := NewUnlimitedMemQuotaProvider()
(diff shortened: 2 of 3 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>2ca93be</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>2ca93be</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/journal.go
```diff
@@ -140,7 +140,9 @@ func (j *ChunkJournal) bootstrapJournalWriter(ctx context.Context) (err error) {
 		return err
 	}
 
-	if !ok { // create new journal file
+	canCreate := !j.backing.readOnly()
```
macneale4, line 143: If the file doesn't exist, then openJournalWriter below will return !ok, and the error to the user will be "missing chunk journal" - that doesn't seem helpful to the user.

PR #10094, opened 2025-11-18
go/store/nbs: Fix NomsBlockStore Conjoin against AWS S3 when the AWS S3 endpoint requires a Content-Length header.

reltuk:
AWS Go SDK needs an io.ReadSeeker in the UploadPartInput in order to supply a Content-Length header. Supplying an io.MultiReader was resulting in an error.

Fix this for now by making a temporary copy of the data and using a bytes.NewReader instead.


reltuk:
#deploy [doltremoteapi]

coffeegoddd (commented):
hmmm... i'm wondering if we should add an explicit check that the reader we supply to the request is an io.ReadSeeker?

PR #10124, opened 2025-11-24
go/cmd/dolt: Allow the Dolt CLI to connect to a running dolt sql-server which is configured with require_secure_transport: true.

reltuk:
When Dolt CLI is running in a directory with a corresponding running sql-server process, it will connect to the server process and complete its work using SQL statements. Previously, the Dolt CLI was configured to always connect on a plaintext TCP connection for these connections. That meant it did not work for servers configured with require_secure_transport: true. One consequence was that the published Dolt dockerhub image did not work with require_secure_transport: true, since that image runs `dolt sql` against the running server as part of its entrypoint.

This changes `dolt` CLI to connect over (non-verified) TLS if such as an option is presented by the server. The CLI still falls back to plaintext as well.

Dolt CLI still does not work when the server is configured with require_client_certificate, since Dolt CLI does not currently have a way to configure its presented client certificate and private key. As a consequence, at least for the time being, the published DockerHub images for Dolt do not work with require_client_certificate: true.

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/queryist_utils.go
+++ b/go/cmd/dolt/commands/sqlserver/queryist_utils.go
@@ -16,6 +16,7 @@ package sqlserver
 
 import (
 	"context"
+	"crypto/tls"
 	sql2 "database/sql"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a2fe529</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a2fe529</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10183, opened 2025-12-09
go: store/datas/pull: pull_chunk_tracker.go: Optimize memory use when backing up to an AWS remote.

reltuk:
PullChunkTracker is responsible for making the HasMany calls against the destination and batching up absent hashes into HashSets which will be delivered to GetManyCompressed and eventually written into table files which are uploaded. This code is used for both pull and push, when the destination is the "local" database or when destination is the remote database respectively. It is used when the remote is both doltremoteapi, thus every HasMany call is an RPC, and when the remote is something like file:// or aws://, thus the table file indexes for the remote are in memory and HasMany calls are very quick.

Different operational characteristics of the various dependencies mean that sometimes a Pull is prone to build up large sets of hashes waiting for HasMany calls, whereas other times it is prone to build up large sets of absent hashes which are waiting for the fetcher thread(s) to take them.

Previously, PullChunkTracker was structured to accumulate HasMany responses and wait to batch them into appropriately-sized batches for GetManyCompressed until the fetcher threads asked for them. This meant that if HasMany batches were very small, because HasMany was very fast, we would accumulate a large number of very small HashSets. These HashSets would take up large amounts of memory. Accumulating the batches as the HasMany responses come in is more memory efficient and should be no slower - we will always accumulate the full batches, and in basically the same order.

Tested by pushing a large database to an AWS remote and memory profiling the result.

### DIFF
--- a/go/store/datas/pull/pull_chunk_fetcher_test.go
+++ b/go/store/datas/pull/pull_chunk_fetcher_test.go
@@ -190,15 +190,25 @@ func TestPop(t *testing.T) {
 		backing[i] = new(int)
 		*backing[i] = i
 	}
+	// s is pointers of [0, 1, 2, ..., 16]
 	s := backing[:]
+	assert.Len(t, s, 16)
+	// pop continuously and assert that we see
+	// the sequence 0, 1, 2, ..., 16 and each
+	// prior element at the end of the list is
+	// |nil|.
 	for i := range 16 {
-		assert.Len(t, s, 16-i)
 		var p *int
 		p, s = pop(s)
 		assert.Len(t, s, 16-i-1)
 		assert.Equal(t, i, *p)
+		// One off the end of the new |s| is now nil.
 		assert.Nil(t, backing[16-i-1], "i is %d", i)
 	}
+	// pop has a precondition of len > 0. This should panic.
+	assert.Panics(t, func() {
+		pop(backing[:0])
+	})
 }
 
 func TestAppendAbsent(t *testing.T) {
--- a/go/store/datas/pull/pull_chunk_tracker.go
+++ b/go/store/datas/pull/pull_chunk_tracker.go
@@ -186,11 +186,12 @@ func (t *PullChunkTracker) reqRespThread(ctx context.Context, initial hash.HashS
 				outstanding -= 1
 				if resp.err != nil {
 					err = errors.Join(err, resp.err)
+				} else if len(resp.hs) > 0 {
+					// Add all the resp.hs hashes, those which are not already present
+					// in dest, to our batches of absent hashes we will return through
+					// GetChunksToFetch.
+					absent = appendAbsent(absent, resp.hs, t.cfg.BatchSize)
 				}
-				// Add all the resp.hs hashes, those which are not already present
-				// in dest, to our batches of absent hashes we will return through
-				// GetChunksToFetch.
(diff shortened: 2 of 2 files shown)


coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f23f4c0</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f23f4c0</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/datas/pull/pull_chunk_tracker.go
```diff
@@ -169,61 +169,57 @@ func (t *PullChunkTracker) reqRespThread(ctx context.Context, initial hash.HashS
...
 				if len(unchecked) == 0 || len(unchecked[len(unchecked)-1]) >= t.cfg.BatchSize {
 					outstanding += 1
 					unchecked = append(unchecked, make(hash.HashSet))
 				}
 				unchecked[len(unchecked)-1].Insert(h)
 			case resp := <-hasManyRespCh:
+				// A hasMany response came back.
 				outstanding -= 1
 				if resp.err != nil {
 					err = errors.Join(err, resp.err)
```
bheni, line 188: May want a continue statement here.  I assume that if resp.err is not nil then resp.hs should be empty, but I guess I'd still opt to skip the append statement as I don't know there are any guarantees that that is true.

go/store/datas/pull/pull_chunk_tracker.go
```diff
@@ -236,6 +232,40 @@ func (t *PullChunkTracker) reqRespThread(ctx context.Context, initial hash.HashS
 	return eg.Wait()
 }
 
+// Pop returns the first element of s and the remaining elements of
```
bheni:
```suggestion
// pop returns the first element of s and the remaining elements of
```

go/store/datas/pull/pull_chunk_fetcher_test.go
```diff
@@ -183,3 +183,38 @@ var getManyerErr = fmt.Errorf("always return an error")
 func (errorGetManyer) GetManyCompressed(ctx context.Context, hashes hash.HashSet, found func(context.Context, nbs.ToChunker)) error {
 	return getManyerErr
 }
+
+func TestPop(t *testing.T) {
```
nicktobey, line 187: These tests could use some docstrings. It's not immediately clear what's being tested.

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>e63f912</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>e63f912</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10227, opened 2025-12-18
dolthub/dolt#5862: Add ignore system table

### LINKED ISSUE #5862: Add alternative to `dolt_status` system table that includes tables ignored by system_ignore
Basically an equivalent to `dolt status --ignored`, except we can add flag parameters to system tables, we have to make a new system table.

codeaucafe:
## Summary

Adds a new `dolt_status_ignored` system table that extends `dolt_status` functionality by including an `ignored` column to identify which unstaged tables match patterns defined in `dolt_ignore`. This provides a SQL interface equivalent to `dolt status --ignored`.

### New System Table

**`dolt_status_ignored`** - Same schema as `dolt_status` plus an `ignored` column:
- `table_name` (text): Name of the table
- `staged` (boolean): Whether the table is staged
- `status` (text): Status of the table (e.g., "new table", "modified", "conflict")
- `ignored` (boolean): Whether the table matches a `dolt_ignore` pattern

### Implementation Details

The implementation started with logic similar to `status_table.go` and was then refactored to share common code between both tables while adding the ignore-checking functionality.

**go/libraries/doltcore/doltdb/system_table.go**
- Added `StatusIgnoredTableName` constant
- Registered table in `GeneratedSystemTableNames()` and `DoltGeneratedTableNames`

**go/libraries/doltcore/sqle/database.go**
- Added switch case for `StatusIgnoredTableName` in `getTableInsensitiveWithRoot()`
- Extracted `getStatusTableRootsProvider()` helper to eliminate duplicate logic between `dolt_status` and `dolt_status_ignored` switch cases

**go/libraries/doltcore/sqle/dtables/status_table.go**
- Added `getStatusRowsData()` function using existing `statusTableRow` struct to share status collection logic between both tables
- Refactored `newStatusItr()` to use shared function

**go/libraries/doltcore/sqle/dtables/status_ignored_table.go** (new file)
- Implements `StatusIgnoredTable` and `StatusIgnoredItr`
- Uses shared `getStatusRowsData()` from `status_table.go`
- Uses adapter pattern via `DoltTableAdapterRegistry` similar to `StatusTable`
- Adds ignore-checking logic via helper functions:
  - `getIgnorePatterns()`: Fetches patterns from `dolt_ignore`
  - `buildUnstagedTableNameSet()`: Creates set for quick lookup
  - `checkIfIgnored()`: Checks if table matches ignore pattern (returns error on failure)

### Behavior

- **Unstaged tables**: `ignored=1` if table name matches a pattern in `dolt_ignore`, `ignored=0` otherwise
- **Staged tables**: Always `ignored=0` (staging overrides ignore patterns, matching git behavior)
- **Constraint violations, merge conflicts, schema conflicts**: Always `ignored=0`

### BATS integration tests

**integration-tests/bats/system-tables.bats:**
- Basic functionality: Verifies `ignored` column correctly identifies ignored tables while non-ignored tables have `ignored=0`
- Staged tables: Confirms staged tables always have `ignored=0` regardless of ignore patterns
- System table visibility: Ensures table appears in `dolt ls --system`

**integration-tests/bats/ls.bats:**
- Updated "ls: --system shows system tables" to include `dolt_status_ignored` (27 tables instead of 26)

### Go enginetests

**go/libraries/doltcore/sqle/enginetest/dolt_queries.go:**

Closes #5862


nicktobey:
It looks like the refactor may have changed the behavior of `dolt_status`, as seen in the newly failing`TestDoltMergeArtifacts`:

```
--- FAIL: TestDoltMergeArtifacts (0.22s)
    --- FAIL: TestDoltMergeArtifacts/divergent_type_change_causes_schema_conflict (0.01s)
        --- FAIL: TestDoltMergeArtifacts/divergent_type_change_causes_schema_conflict/select_*_from_dolt_status (0.00s)
            evaluation.go:854: 
                	Error Trace:	/Users/runner/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.20.1-0.20251218001312-084240aa6481/enginetest/evaluation.go:854
                	            				/Users/runner/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.20.1-0.20251218001312-084240aa6481/enginetest/evaluation.go:598
                	            				/Users/runner/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.20.1-0.20251218001312-084240aa6481/enginetest/evaluation.go:471
                	            				/Users/runner/go/pkg/mod/github.com/dolthub/go-mysql-server@v0.20.1-0.20251218001312-084240aa6481/enginetest/evaluation.go:442
                	            				/Users/runner/hostedtoolcache/go/1.25.3/arm64/src/runtime/asm_arm64.s:1268
                	Error:      	Not equal: 
                	            	expected: []sql.Row{sql.Row{"t", 0x0, "schema conflict"}}
                	            	actual  : []sql.Row{sql.Row{"t", 0x0, "merged"}}
                	            	
                	            	Diff:
                	            	--- Expected
                	            	+++ Actual
                	            	@@ -4,3 +4,3 @@
                	            	   (uint64) 0,
                	            	-  (string) (len=15) "schema conflict"
                	            	+  (string) (len=6) "merged"
                	            	  }
                	Test:       	TestDoltMergeArtifacts/divergent_type_change_causes_schema_conflict/select_*_from_dolt_status
                	Messages:   	Unexpected result for query select * from dolt_status
```

One alternative to having two system tables that share code would be to make `dolt_status` a system view equal to `SELECT table_name, staged, status FROM DOLT_STATUS WHERE ignored = false`. We do that in some other places: for example, `DOLT_BLAME` is implemented as a view over `DOLT_DIFF`, and you can see how that's implemented in `go/libraries/doltcore/sqle/dtables/blame_view.go`

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -115,6 +115,14 @@ type statusTableRow struct {
 	// of this table when you are using local vs remote sql connections.
 }
 
+// statusRowData contains the data for a single status row.
+type statusRowData struct {
```
nicktobey: This is identical to `statusTableRow` declared immediately above it. I don't think we need it.

go/libraries/doltcore/sqle/dtables/status_ignored_table.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	// Build a set of unstaged table names for quick lookup
+	unstagedTableNames := buildUnstagedTableNameSet(unstagedTables)
+
+	// Convert status rows to status_ignored rows, adding the ignored column
+	rows := make([]statusIgnoredTableRow, len(statusRows))
+	for i, row := range statusRows {
+		ignored := byte(0)
+		// Only check ignore patterns for unstaged tables
```
nicktobey: We also should only check the ignore patterns for tables whose status is being added, not tables being modified or removed. That's what git does, and we want to match git here.

go/libraries/doltcore/sqle/dtables/status_ignored_table.go
```diff
@@ -0,0 +1,196 @@
...
+	return result
+}
+
+// checkIfIgnored returns 1 if the table name matches an ignore pattern, 0 otherwise.
+func checkIfIgnored(tableName string, ignorePatterns doltdb.IgnorePatterns) byte {
+	tblNameObj := doltdb.TableName{Name: tableName}
+	result, err := ignorePatterns.IsTableNameIgnored(tblNameObj)
+	if err != nil {
+		return byte(0)
+	}
```
nicktobey: We don't want to silently swallow the error here. I'm not sure it makes sense to outline this function if it's small and  only gets called in one place.

codeaucafe:
I think this is good to be reviewed again now

go/libraries/doltcore/sqle/dtables/status_ignored_table.go
nicktobey: Propagating the error is good, but I'm still skeptical that this should be in it's own function, since it only gets called in one place, and casting it to a byte only really makes sense in the context of the status table.

go/libraries/doltcore/sqle/dtables/status_table.go
```diff
@@ -190,15 +183,15 @@ func getStatusRowsData(
 	if ws.MergeActive() {
 		ms := ws.MergeState()
 		for _, tbl := range ms.TablesWithSchemaConflicts() {
-			rows = append(rows, statusRowData{
+			rows = append(rows, statusTableRow{
 				tableName: tbl.String(),
 				isStaged:  byte(0),
-				status:    mergedStatus,
+				status:    "schema conflict",
```
nicktobey: This and "constraint violation" should probably be stored in a constant, like mergeConflictStatus does below.

codeaucafe: oddly enough I thought I made this change, but not sure what happened. good call out. I'll fix this and the other comments you made.

codeaucafe:
Thanks Nick, I'll update accordingly. Was almost done ambit computer messed up so I called it a night. I'll push the fixes tomorrow

go/libraries/doltcore/sqle/dtables/status_ignored_table.go
```diff
@@ -0,0 +1,204 @@
...
+// StatusIgnoredItr is a sql.RowIter implementation for the status_ignored table.
+type StatusIgnoredItr struct {
+	rows []statusIgnoredTableRow
+}
+
+type statusIgnoredTableRow struct {
+	tableName string
+	status    string
+	isStaged  byte // not a bool bc wire protocol confuses bools and tinyint(1)
+	ignored   bool
```
codeaucafe, line 116: @nicktobey I wasn't sure I read your comment correctly about "... casting it to a byte only really makes sense in the context of the status table...". I read interpreted it as: since statusIgnoredTableRow is a separate struct from statusTableRow, the ignored field doesn't need the byte workaround for wire protocol issues, so I changed ignored from byte to bool. Is this what you meant? or if not, then I can change back to byte

codeaucafe:
WARNING: please don't merge. there is a failing bats tests that appears tied to my changes that I missed. I'll work on fixing that and adding a comment when its fixed

PR #10252, opened 2025-12-30
feat: Structural output mode for `dolt status`

### LINKED ISSUE #6056: Structural output mode for `dolt status`
Requested by @macneale4.

Currently, the only option for `dolt status` is to output it in a human readable format, mirroring git.

However, now that we've migrated `dolt status` to be backed by the `dolt_status` system table, it should be trivial to allow more structured output, and would be useful for automation (cleaner tests being the motivating example.)

majiayu000:
## Summary
- Add `-r/--result-format` flag to `dolt status` command supporting JSON output
- Follow the pattern established by `dolt diff` for format flag handling

## Changes
- Add `FormatFlag` support to status command ArgParser
- Add `StatusJSON` and `StatusTableChange` structs for structured JSON output
- Add `printJSON` function to output status in JSON format
- Update `Exec` to validate and handle the format flag

The JSON output includes branch info, remote tracking info, ahead/behind counts, merge status, staged/unstaged/untracked tables, conflicts, constraint violations, and ignored tables.

Fixes #6056

### DIFF
--- a/go/cmd/dolt/commands/status.go
+++ b/go/cmd/dolt/commands/status.go
@@ -94,7 +94,7 @@ func (cmd StatusCmd) Docs() *cli.CommandDocumentation {
 func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 0)
 	ap.SupportsFlag(cli.ShowIgnoredFlag, "", "Show tables that are ignored (according to dolt_ignore)")
-	ap.SupportsString(FormatFlag, "r", "result output format", "How to format status output. Valid values are tabular (human readable) and json. Defaults to tabular.")
+	ap.SupportsString(FormatFlag, "r", "result output format", "How to format status output. Valid values are human readable (tabular) and json. Defaults to tabular.")
 	return ap
 }
 


go/cmd/dolt/commands/status.go
```diff
@@ -93,6 +94,7 @@ func (cmd StatusCmd) Docs() *cli.CommandDocumentation {
 func (cmd StatusCmd) ArgParser() *argparser.ArgParser {
 	ap := argparser.NewArgParserWithMaxArgs(cmd.Name(), 0)
 	ap.SupportsFlag(cli.ShowIgnoredFlag, "", "Show tables that are ignored (according to dolt_ignore)")
+	ap.SupportsString(FormatFlag, "r", "result output format", "How to format status output. Valid values are tabular, json. Defaults to tabular.")
```
macneale4: The default form isn't really a table though. A better description of the format would be "human readable".

macneale4 (changes requested):
Thanks for the contribution!

We'll need some demonstration that it works so we can maintain for the long haul. Please add tests here `integration-tests/bats/status.bats`

majiayu000:
Added integration tests in  for  (clean and changed/ignored cases). Also updated the flag help text to describe the default as human readable (tabular). Latest commits pushed.

majiayu000:
Added integration tests in `integration-tests/bats/status.bats` for `dolt status --result-format json` (clean and changed/ignored cases). Also updated the flag help text to describe the default as human readable (tabular). Latest commits pushed.

majiayu000:
@macneale4 Added integration coverage in integration-tests/bats/status.bats for dolt status --result-format json (clean + changes/ignored) and updated the flag help text to describe the default as human readable (tabular). Please take another look.

macneale4:
We are debating on the team about this feature. At a minimum, we should do this the way `git` does, which is with the `--porcelain` flag. This is kind of a special case with git, and I'm not sure it's something we should encourage with dolt. But the --porcelain flag just prints parsable text, not JSON, so it's not clear if what you want is parsable `dolt status` or JSON.

More generally, the only reason for doing this is to script around the CLI, and that's not a great approach. We would recommend any scripting be built around `dolt sql -r json -q <query>`. For status you would do:

`dolt sql -r json -q "select * from dolt_status"`

Which you can then consume however you see fit. Hold tight while we consider the ramifications of this change.

macneale4:
After a brief discussion, we've decided that the original issue should have been closed and the correct mechanism is what I suggested: `dolt sql -r json...`

I really appreciate your effort in sending this PR. I'm sorry we hadn't revisited the list of `good first issues` in a bit, but I'm going to go through and verify they are all legitimate feature requests. Sorry for wasting your time!

majiayu000:
Appreciate the feedback!

PR #10327, opened 2026-01-16
go/cmd/dolt: Make movable temp file configuration lazy.

reltuk:
Previously, early in the process life-cycle after dolt was run, dolt would immediately check if a file created in `os.TempDir()` was able to be `os.Rename`d into a subdirectory of the dolt process's data directory, by default `$PWD/.dolt`. If the rename failed, it would configure Dolt to use `.dolt/temptf` as the movable temp file directory instead.

This meant that for many `dolt` invocations, Dolt would do some local filesystem writes, including potentially `Mkdir(.dolt)` before it it was fully loaded. With the advent of things like dolt accessing the running server through sql-server.info and `dolt --host ... sql` this behavior was not ideal. It creates races with concurrently run `dolt` processes that try to use the `.dolt` directory, and it creates odd behavior when running something like `dolt sql` in a non-Dolt directory but on a filesystem mount where this rename from `os.TempDir()` does not work.

This PR changes the check and the configuration of a potential `.dolt/temptf` directory to be delayed until the first temporary movable file is actually requested by Dolt. At that point we know that we have a need for a renamable temp file and it is OK to go ahead and create `.dolt/temptf` if we need it.

This PR changes the error handling around some edge cases like permissions errors on the calling process creating `temptf` within `.dolt`. In particular, some errors which were previously early and fatal are now delayed until use site and may end up being persistent but non-fatal to the process, depending on the operation.

### DIFF
--- a/go/store/util/tempfiles/temp_files.go
+++ b/go/store/util/tempfiles/temp_files.go
@@ -80,6 +80,48 @@ func (tfp *TempFileProviderAt) Clean() {
 	}
 }
 
+type LazyTempFileProvider struct {
+	once     sync.Once
+	loader   func() (TempFileProvider, error)
+	provider TempFileProvider
+	perr     error
+}
+
+func NewLazyTempFileProvider(loader func() (TempFileProvider, error)) *LazyTempFileProvider {
+	return &LazyTempFileProvider{
+		loader: loader,
+	}
+}
+
+func (p *LazyTempFileProvider) loadit() {
+	p.once.Do(func() {
+		p.provider, p.perr = p.loader()
+	})
+}
+
+func (p *LazyTempFileProvider) Clean() {
+	// Don't load if we haven't already been loaded.
+	if p.provider != nil {
+		p.provider.Clean()
+	}
+}
+
+func (p *LazyTempFileProvider) GetTempDir() string {
+	p.loadit()
+	if p.perr != nil {
+		return os.TempDir()
+	}
+	return p.provider.GetTempDir()
+}
+
+func (p *LazyTempFileProvider) NewFile(dir, pattern string) (*os.File, error) {
+	p.loadit()
+	if p.perr != nil {
+		return nil, p.perr
+	}
+	return p.provider.NewFile(dir, pattern)
+}
+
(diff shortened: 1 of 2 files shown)


go/store/util/tempfiles/temp_files.go
```diff
@@ -80,6 +80,48 @@ func (tfp *TempFileProviderAt) Clean() {
...
+
+func (p *LazyTempFileProvider) NewFile(dir, pattern string) (*os.File, error) {
+	p.loadit()
+	if p.perr != nil {
+		return nil, p.perr
+	}
+	return p.provider.NewFile(dir, pattern)
+}
+
 // MovableTemFile is an object that implements TempFileProvider that is used by the nbs to create temp files that
```
angelamayxie:
```suggestion
// MovableTempFile is an object that implements TempFileProvider that is used by the nbs to create temp files that
```

angelamayxie: not related to your change but might as well fix it 🤷

go/store/util/tempfiles/temp_files.go
```diff
@@ -80,6 +80,48 @@ func (tfp *TempFileProviderAt) Clean() {
 	}
 }
 
+type LazyTempFileProvider struct {
```
angelamayxie, line 89: add a comment re: what this is for

zachmu (approved):
LGTM

angelamayxie (approved):

coffeegoddd:
@reltuk __DOLT__
 <table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5362a4c</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5362a4c</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10449, opened 2026-02-07
fix: reject vector indexes on nullable columns at DDL time

### LINKED ISSUE #10448: VECTOR INDEX: batch inserts fail with NULL embeddings after ~83 rows
## Bug Report

Inserting rows with NULL vector columns into a table with a VECTOR INDEX fails after approximately 83 rows with:

```
unable to cast <nil> of type <nil> to vector
```

### Reproduce

```sql
CREATE TABLE t (id INT PRIMARY KEY, v VECTOR(768) DEFAULT NULL);
ALTER TABLE t ADD VECTOR INDEX idx (v);
-- Insert rows without providing vector value
INSERT INTO t (id) VALUES (1), (2), ..., (83);  -- fails at row 83
```

### Root Cause

The bug spans two repos in the Dolt stack:

1. **GMS** (`sql/core.go`): `ConvertToVector()` has no nil guard — falls through to error default for nil input. Fixed in dolthub/go-mysql-server#3420.

2. **Dolt** (`go/store/prolly/proximity_map.go`): `getConvertToVectorFunction()` passes nil from `GetBytesAdaptiveValue` directly to `ConvertToVector` without checking for nil. The proximity map flush path (`visitNode`, `makePathMaps`) also doesn't handle nil vectors returned from the convert function.

```
INSERT (NULL embedding)
  → prollySecondaryIndexWriter.Insert()
  → mut.Put(key_with_null, EmptyTuple)
  → ~83 rows accumulate (64KB defaultMaxPending threshold)
  → flushPending()
  → visitNode() → convert(ctx, key)
  → GetBytesAdaptiveValue() → nil
  → ConvertToVector(ctx, nil) → CRASH
```

### Impact

Any table with a VECTOR INDEX and nullable VECTOR column cannot insert more than ~83 rows without providing vector values. This blocks use cases where embeddings are computed asynchronously (insert row first, add embedding later).

### Fix

Defense-in-depth approach across both repos:

- **GMS fix**: Nil guard in `ConvertToVector` — dolthub/go-mysql-server#3420
- **Dolt fix**: Nil guards in `getConvertToVectorFunction`, `visitNode`, `makePathMaps`, and `getClosestVector` — prevents nil from reaching distance calculations

### Related

- dolthub/go-mysql-server#3420 (GMS root cause fix)
- dolthub/go-mysql-server#3421 (GMS draft PR)

### Environment

- Dolt version: 1.81.5 (also reproducible on main)
- OS: macOS (also confirmed on Linux)

peterkc:
## Summary

Reject `CREATE VECTOR INDEX` on nullable columns at DDL time, matching MariaDB ERROR 1252 behavior. This replaces the previous nil-guard approach (7-site runtime checks) with a single DDL constraint that makes the invalid state unrepresentable.

### Maintainer Direction

Per @nicktobey's [review](https://github.com/dolthub/dolt/pull/10449#issuecomment-2942498498):

> "It looks like the real issue is that we're allowing vector indexes on a nullable vector column at all. MariaDB would reject this with ERROR 1252."

### Changes (2 files)

**`go/libraries/doltcore/schema/index_coll.go`**

`validateColumnIndexable()` (currently a no-op) now checks for vector + nullable:

```go
func validateColumnIndexable(c Column, isVector bool) error {
    if isVector && c.IsNullable() {
        return fmt.Errorf("all parts of a VECTOR index must be NOT NULL")
    }
    return nil
}
```

Call site in `AddIndexByColTags()` updated to pass `props.IsVector`. `UnsafeAddIndexByColTags` is unaffected (bypasses validation by design).

**`go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go`**

- New `TestVectorIndexDDL` with 3 test cases:
  - `CREATE VECTOR INDEX` on nullable column → error
  - `CREATE VECTOR INDEX` on NOT NULL column → success
  - `ALTER TABLE ADD VECTOR INDEX` on nullable column → error
- `TestVectorIndexes` updated to skip 2 GMS tests that create vector indexes on nullable columns (now correctly rejected)

### What's NOT in this PR (vs previous version)

The previous version had 7 nil-guard sites in `proximity_map.go` and `proximity_mutable_map.go`, plus a separate test file. All of that is removed — this PR is just the DDL constraint.

### Open question: pre-existing databases

Databases created before this fix may have vector indexes on nullable columns. The DDL constraint only blocks *new* index creation. Existing indexes would surface errors during `OPTIMIZE TABLE` or index rebuild. Should we add a runtime guard for this case, or is the DDL constraint sufficient for now?

Asked in [this comment](https://github.com/dolthub/dolt/pull/10449#issuecomment-3874798301).

### Test Plan

- [x] `TestVectorIndexDDL` — 3 new DDL constraint tests pass
- [x] `TestVectorIndexes` — existing error tests pass, 2 nullable-column tests correctly skipped
- [x] Local binary smoke test confirms constraint works end-to-end

Fixes #10448

### DIFF
--- a/go/libraries/doltcore/schema/index_coll.go
+++ b/go/libraries/doltcore/schema/index_coll.go
@@ -214,7 +214,7 @@ func (ixc *indexCollectionImpl) AddIndexByColTags(indexName string, tags []uint6
 	for _, tag := range tags {
 		// we already validated the tag exists
 		c, _ := ixc.colColl.GetByTag(tag)
-		err := validateColumnIndexable(c)
+		err := validateColumnIndexable(c, props.IsVector)
 		if err != nil {
 			return nil, err
 		}
@@ -243,7 +243,10 @@ func (ixc *indexCollectionImpl) AddIndexByColTags(indexName string, tags []uint6
 }
 
 // validateColumnIndexable returns an error if the column given cannot be used in an index
-func validateColumnIndexable(c Column) error {
+func validateColumnIndexable(c Column, isVector bool) error {
+	if isVector && c.IsNullable() {
+		return fmt.Errorf("all parts of a VECTOR index must be NOT NULL")
+	}
 	return nil
 }
 
--- a/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
@@ -883,7 +883,64 @@ func TestIndexes(t *testing.T) {
 func TestVectorIndexes(t *testing.T) {
 	harness := newDoltHarness(t)
 	defer harness.Close()
-	enginetest.TestVectorIndexes(t, harness)
+	// GMS VectorIndexQueries includes tests that create vector indexes on nullable columns.
+	// Dolt now rejects this at DDL time (see TestVectorIndexDDL), so we skip those tests here
+	// and only run the ones that are compatible with the NOT NULL constraint.
+	for _, script := range queries.VectorIndexQueries {
+		if script.Name == "basic JSON vector index" || script.Name == "basic VECTOR vector index" {
+			t.Run(script.Name, func(t *testing.T) {
+				t.Skip("Dolt rejects vector indexes on nullable columns; see TestVectorIndexDDL")
+			})
+			continue
+		}
+		enginetest.TestScript(t, harness, script)
+	}
+}
+
+func TestVectorIndexDDL(t *testing.T) {
+	harness := newDoltHarness(t)
+	defer harness.Close()
+	scripts := []queries.ScriptTest{
+		{
+			Name: "vector index on nullable column rejected",
+			SetUpScript: []string{
+				"CREATE TABLE t (id INT PRIMARY KEY, v VECTOR(3))",
+			},
+			Assertions: []queries.ScriptTestAssertion{
+				{
+					Query:          "CREATE VECTOR INDEX idx ON t (v)",
+					ExpectedErrStr: "all parts of a VECTOR index must be NOT NULL",
+				},
+			},
+		},
+		{
+			Name: "vector index on NOT NULL column succeeds",
+			SetUpScript: []string{
+				"CREATE TABLE t (id INT PRIMARY KEY, v VECTOR(3) NOT NULL)",
+			},
+			Assertions: []queries.ScriptTestAssertion{
+				{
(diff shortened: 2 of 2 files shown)


timsehn:
@nicktobey owns vectors on our side. He will review this.

nicktobey:
So this PR works by:
1. detecting nil inputs to ConvertToVector and returning nil
2. detecting nil outputs from vector conversion and skipping them when building the index

I don't think this is the correct logic:

1. The output of ConvertToVector is a slice of floats. If we return nil, that's the same as returning a zero-dimension vector. And while I don't believe zero-dimension vectors are allowed, there's nothing inherently wrong with them and treating them as a sentinel value could cause issues down the line.
2. I'm not convinced that skipping nil elements when building the index is the correct solution. This change effectively swallows the error without addressing the underlying cause.

I think the correct fix here is to match disallow creating vector indexes on nullable columns. In the event a user already has a vector index on a nullable column, we should return a useful error message if the index logic ever encounters a nil value.

peterkc:
@nicktobey Thanks for the detailed review — reworked the PR to use a DDL constraint in `validateColumnIndexable()` instead of the nil-guard approach. See updated PR description for details.

peterkc:
@nicktobey I've reworked this PR per your feedback — replaced the 7-site nil-guard approach with a single DDL constraint in `validateColumnIndexable()`. 

The PR now has just 1 commit touching 2 files (index_coll.go + dolt_engine_test.go). Would appreciate another look when you have a chance!

nicktobey:
This looks much better.

My one comment is that instead of enforcing the not-null constraint in Dolt and skipping the GMS tests that make a vector index on nullable columns, it would be better to fix this is GMS. It never makes sense to have a vector index on nullable columns, so if we disallow that behavior in GMS, we can also update the GMS tests instead of introducing new tests in Dolt and skipping the GMS ones.

zachmu:
Thank you for your contribution, @peterkc! I carried your commit forward into https://github.com/dolthub/dolt/pull/11785. Current GMS already enforces the NOT NULL restriction Nick requested, so I retained the Dolt schema check as defense in depth, added a direct schema regression test, and removed the obsolete vector-test skips. https://github.com/dolthub/go-mysql-server/pull/3858 extends the shared GMS coverage to JSON columns and ALTER TABLE. The vector tests pass in both GMS and Dolt, and the Dolt schema tests pass. Closing this PR in favor of those replacements; your contribution is preserved in the commit history.

PR #8681, opened 2024-12-15
Bump for go-icu-regex

coffeegoddd:
Bump for:
* https://github.com/dolthub/go-icu-regex/pull/2

Issue #1551, opened 2021-04-15
`dolt sql --multi-db-dir` does not ignore broken dolt repositories


timsehn:
`PS C:\Users\timse\dolthub\dolt> dolt sql --multi-db-dir ./
error loading environment 'broken_pipe' at path 'C:\Users\timse\dolthub\dolt\broken-pipe': open C:\Users\timse\dolthub\dolt\broken-pipe\.dolt\repo_state.json: The system cannot find the path specified.`

zachmu:
Duplicate of #809

PR #10453, opened 2026-02-08
Add Nix flake and CI job for build compatibility visibility

### LINKED ISSUE #10452: Add Nix flake and CI job for build compatibility visibility
## Summary

Would you consider adding a Nix flake (`flake.nix` + `default.nix`) and a lightweight CI workflow to Dolt? The goal isn't to gate releases on Nix compatibility, but to give maintainers early visibility when Go version bumps break Nix-based builds — before downstream projects discover it the hard way.

## Context

Dolt is available in [nixpkgs](https://search.nixos.org/packages?query=dolt) and has a meaningful Nix user base. The [beads](https://github.com/steveyegge/beads) project depends heavily on Dolt and recently went through a painful cycle of Nix build failures caused by Dolt's Go version bumps:

1. **Jan 26**: Dolt bumped Go to 1.25.6 for CVE-2025-61729 ([PR #10371](https://github.com/dolthub/dolt/pull/10371))
2. **Jan 30**: Beads inherited the requirement via `dolthub/driver` — Nix builds broke immediately because nixpkgs only had Go 1.25.1 at the time ([beads #1373](https://github.com/steveyegge/beads/issues/1373))
3. **Feb 3**: Multiple issues were filed ([beads #1422](https://github.com/steveyegge/beads/issues/1422), [beads #1476](https://github.com/steveyegge/beads/issues/1476)). Workarounds included `GOTOOLCHAIN=auto` and `continue-on-error: true` in CI
4. **Feb 6**: Beads made the drastic architectural decision to remove the embedded Dolt driver entirely — partly to decouple from this version churn ([commit e533064](https://github.com/steveyegge/beads/commit/e533064))
5. **Feb 7**: Beads bumped to Go 1.25.7 for CVE-2025-68121, Nix broke again ([beads #1551](https://github.com/steveyegge/beads/issues/1551))
6. **Ongoing**: [beads #1465](https://github.com/steveyegge/beads/issues/1465) tracks removing `continue-on-error` once nixpkgs catches up

The core issue is a lag between Go releases and nixpkgs uptake (typically 2-4 weeks). Dolt bumps Go versions aggressively for security — which is the right thing to do — but there's currently no signal in Dolt's own CI that these bumps affect Nix builds.

## Proposal

Add a minimal Nix flake and a `continue-on-error` CI job, modeled on what beads already does. Concretely:

### 1. `flake.nix` + `default.nix` at the repo root

A `default.nix` using `pkgs.buildGoModule` to build Dolt from source, and a `flake.nix` wrapping it for multi-platform support. The key detail from beads' approach is:

```nix
env.GOTOOLCHAIN = "auto";
```

This lets the Go toolchain auto-download the required version even when nixpkgs is behind, so `nix build` still works as a fallback.

### 2. A CI workflow (e.g., `ci-nix-build.yaml`)

A lightweight job that runs on PRs touching `go/go.mod` or the Nix files:

```yaml
name: CI Nix Build
on:
  pull_request:
    paths:
      - 'go/go.mod'
      - 'flake.nix'
      - 'default.nix'
      - 'flake.lock'
  push:
    branches: [main]

jobs:
  nix-build:
    name: Test Nix Build
    runs-on: ubuntu-latest
    continue-on-error: true  # informational, not a gate
    steps:
      - uses: actions/checkout@v4
      - uses: cachix/install-nix-action@v31
        with:
          nix_path: nixpkgs=channel:nixos-unstable
      - name: Build Dolt via Nix
        run: nix build .#default
      - name: Verify binary
        run: ./result/bin/dolt version
```

### What this would NOT do

- **Block releases or PRs** — the job uses `continue-on-error: true`, so it's purely informational
- **Slow down CI** — path-filtered to only run when `go.mod` or Nix files change
- **Require Nix expertise to maintain** — the flake is ~50 lines and rarely needs updating (mostly just `vendorHash` when dependencies change)

### What this WOULD do

- **Surface Nix breakage in Dolt's own CI** rather than letting downstream projects discover it weeks later
- **Let Nix users build Dolt from source** via `nix build github:dolthub/dolt` without waiting for nixpkgs updates
- **Create parity with beads** so both projects see the same compatibility signal
- **Show ecosystem citizenship** toward Nix users who already depend on Dolt via nixpkgs

## Prior art

- Beads' Nix CI: [ci.yml test-nix job](https://github.com/steveyegge/beads/blob/main/.github/workflows/ci.yml)
- Beads' flake: [flake.nix](https://github.com/steveyegge/beads/blob/main/flake.nix), [default.nix](https://github.com/steveyegge/beads/blob/main/default.nix)

I'm happy to put together a PR with the implementation if there's interest. Thank you for considering this — and thank you for maintaining Dolt. It's a fantastic project.

turian:
## Summary

- Adds a Nix flake (`flake.nix` + `default.nix`) so Nix users can build Dolt from source via `nix build`
- Adds a lightweight CI workflow (`ci-nix-build.yaml`) that builds Dolt via Nix on PRs touching `go.mod` or nix files
- The CI job uses `continue-on-error: true` — it is **informational only** and does not gate PRs or merges

## Motivation

Downstream projects like [beads](https://github.com/steveyegge/beads) depend on Dolt and have experienced repeated Nix build failures when Dolt bumps Go versions (see the full timeline in #10452). The core issue is that nixpkgs lags behind Go releases by 2-4 weeks, so Go version bumps — even justified ones for CVEs — silently break Nix builds.

This PR gives Dolt maintainers an early signal when that happens, rather than letting downstream projects discover it weeks later.

## What's included

| File | Purpose |
|---|---|
| `default.nix` | `buildGoModule` derivation for Dolt; uses `GOTOOLCHAIN=auto` so builds still work when nixpkgs Go is behind |
| `flake.nix` | Multi-platform flake wrapper (linux/mac × x86_64/aarch64) with dev shell |
| `.github/workflows/ci-nix-build.yaml` | CI job, path-filtered to `go.mod`/nix files, `continue-on-error: true` |
| `.gitignore` | Added `result` (nix build output symlink) |

## Known: vendorHash placeholder

The `vendorHash` in `default.nix` is set to `lib.fakeHash`. The first CI run will fail and **print the correct hash in the build logs**. I'll update it once CI runs. This is the [standard nix workflow](https://nixos.wiki/wiki/Import_From_Derivation#Extracting_the_hash) for new packages.

## Test plan

- [ ] CI runs and prints the correct `vendorHash` in the nix build logs
- [ ] Update `vendorHash` with the correct value
- [ ] CI nix build passes (or fails only due to nixpkgs Go version lag, which is expected)
- [ ] `dolt version` runs successfully from the nix-built binary

Closes #10452

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,18 +14,15 @@ pkgs.buildGoModule {
   subPackages = [ "cmd/dolt" ];
   doCheck = false;
 
-  # Placeholder hash — cannot be computed until nixpkgs ships a Go version
-  # that satisfies go.mod (currently requires >= 1.25.6, nixpkgs has 1.25.5).
-  # Once nixpkgs catches up, the build will fail with a hash mismatch and
-  # print the correct value in the "got:" line of the error output.
+  # vendorHash must be updated when go.mod/go.sum change. When incorrect,
+  # `nix build` will fail and print the correct hash in the "got:" line.
+  # To compute without nix installed:
+  #   docker run --rm -v $(pwd):/workspace -w /workspace nixos/nix \
+  #     sh -c 'echo "experimental-features = nix-command flakes" >> /etc/nix/nix.conf && nix build .#default 2>&1'
   vendorHash = pkgs.lib.fakeHash;
 
-  # Note: GOTOOLCHAIN=auto does NOT work inside the nix build sandbox because
-  # the sandbox has no network access, so Go cannot download a newer toolchain.
-  # When nixpkgs Go is older than what go.mod requires, the build will fail.
-  # This is intentional — it is the signal this CI job exists to surface.
-
   nativeBuildInputs = [ pkgs.git ];
+  buildInputs = [ pkgs.icu ];
 
   meta = with pkgs.lib; {
     description = "Dolt – a SQL database you can diff, branch, and merge";
--- a/flake.nix
+++ b/flake.nix
@@ -39,6 +39,7 @@
               git
               gopls
               gotools
+              icu
             ];
 
             shellHook = ''


turian:
Hi! Quick note on the `vendorHash` placeholder:

The `vendorHash` in `default.nix` is currently set to `lib.fakeHash` because you need to actually run `nix build` to compute the correct hash (it's a SHA256 of all Go module dependencies). This is the standard workflow for new nix packages — there's no way to compute it without running the build.

**What needs to happen:**

1. CI needs to run the nix build workflow (I don't have CI enabled on my fork, so it'll need to run here)
2. The build will fail and print the correct hash in the logs, something like:
   ```
   hash mismatch in fixed-output derivation:
     specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
     got:       sha256-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX=
   ```
3. I'll update `vendorHash` in `default.nix` with the `got:` value and push a fixup

So the first CI run failing on the nix job is expected and by design — just need that one run to get the correct hash. After that it should build cleanly (or fail only if nixpkgs Go is behind, which is exactly what the job is meant to detect).

turian:
## First CI run results

Ran the nix build on the fork ([turian/dolt#1](https://github.com/turian/dolt/pull/1)) and got the exact failure this CI job is designed to detect:

```
go: go.mod requires go >= 1.25.6 (running go 1.25.5; GOTOOLCHAIN=local)
```

nixpkgs unstable currently ships Go 1.25.5. The nix build sandbox forces `GOTOOLCHAIN=local` (no network access), so Go can't auto-download 1.25.6.

I've updated `default.nix` to remove the `GOTOOLCHAIN=auto` setting (it was a no-op in the sandbox) and documented why the build failure is the intended signal.

**Current status:**
- The `vendorHash` is still a placeholder (`lib.fakeHash`) because we can't get past the Go version check to compute it
- Once nixpkgs updates to Go 1.25.6+ (reportedly in staging, expected mid-February), the build will progress to the hash mismatch step, we'll fill in the correct hash, and the full build should work
- In the meantime, the CI job correctly reports that Dolt's Go version requirement is ahead of nixpkgs — which is exactly the signal downstream projects like beads need

This is a working proof-of-concept of the value: if this CI job had existed when Go was bumped to 1.25.6, the dolt team would have seen the amber light immediately rather than hearing about it weeks later from beads users.

turian:
## Why this matters — the downstream impact in detail

To give some more context on why we think this is worth adding:

### The beads saga

[Beads](https://github.com/steveyegge/beads) is an issue tracker that uses Dolt as a storage backend. When Dolt bumped Go to 1.25.6 on Jan 26 (for CVE-2025-61729), it triggered a cascade of breakage for beads' Nix users:

- [beads #1373](https://github.com/steveyegge/beads/issues/1373) — Nix build fails (Jan 30)
- [beads #1422](https://github.com/steveyegge/beads/issues/1422) — CI is failing on main: Lint, Build, and Nix errors (Feb 3)
- [beads #1476](https://github.com/steveyegge/beads/issues/1476) — Nix flake uses Go 1.25.1 but go.mod requires Go 1.25.6 (Feb 3)
- [beads #1465](https://github.com/steveyegge/beads/issues/1465) — Remove continue-on-error from Nix CI once nixpkgs has Go 1.25.6 (still open)
- [beads #1551](https://github.com/steveyegge/beads/issues/1551) — Nix flake build broken again after Go 1.25.7 bump (Feb 7)

Multiple community members reported the issue ([schickling](https://github.com/steveyegge/beads/issues/1373), [turian](https://github.com/steveyegge/beads/issues/1476), [others](https://github.com/steveyegge/beads/issues/1551)).

### The structural problem

The lag between Go releases and nixpkgs uptake is typically 2-4 weeks. Dolt bumps Go aggressively for security (which is the right call), but there's a window where:

1. Dolt's `go.mod` requires Go X.Y.Z
2. nixpkgs only has Go X.Y.(Z-1)
3. Any project that depends on Dolt via Go modules inherits the requirement
4. Nix builds break for all of them

Nobody finds out until downstream users file issues. This CI job closes that feedback loop.

### What this means practically for Dolt maintainers

- **Day-to-day**: you will almost never see this job. It only triggers on PRs that touch `go.mod`, `go.sum`, or the nix files
- **When you bump Go**: the job will go amber (fail but not block). You'll know immediately that nix builds are broken and can note it in release notes or give downstream projects a heads-up
- **When nixpkgs catches up**: the job goes green again with no action needed from you
- **Maintenance burden**: near zero. The only thing that needs updating is `vendorHash` when Go module dependencies change, and nix will tell you the correct value in the error output

It's essentially a free early-warning system for an ecosystem you already support (Dolt is [packaged in nixpkgs](https://search.nixos.org/packages?query=dolt)).

reltuk:
Thank you for this submission. We are definitely interested in making sure common development workflows for beads continue to work and that we have tools to know about any upcoming breakages and can communicate those clearly in the cases when they are necessary.

A few questions / notes:

1) Could/should beads just change its nix flake to run to use go-overlay with `go-bin.latest`? Since that picks up new versions of Go within approximately 4 hours of release, and Dolt is very unlikely to take a hard dependency on a new versions of Go within 4 hours of its release, that seems to fix any possible breakage from Go SDK version here.

2) The flake as written doesn't seem to work. I need to add `buildInputs = [ pkgs.icu ];` to the dolt go module derivation.

3) What are the semantics of and expectations on the `vendorHash` parameter in the derivation? Will that need to be set to the correct value every time go.mod changes? Could we get a CI job that trues it up whenever it needs it?

Thanks again. Please let me know your thoughts. For testing this flake, I was quite happy with the conversion to go-overlay:

```diff
aaronson@Aarons-MacBook-Pro dolt % git diff -- default.nix flake.nix
diff --git a/default.nix b/default.nix
index a2c1fd5a4c..115e171207 100644
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,5 @@
-{ pkgs, self }:
-pkgs.buildGoModule {
+{ pkgs, go, self }:
+(pkgs.buildGoModule.override { inherit go; }) {
   pname = "dolt";
   version =
     let
@@ -14,18 +14,10 @@ pkgs.buildGoModule {
   subPackages = [ "cmd/dolt" ];
   doCheck = false;
 
-  # Placeholder hash — cannot be computed until nixpkgs ships a Go version
-  # that satisfies go.mod (currently requires >= 1.25.6, nixpkgs has 1.25.5).
-  # Once nixpkgs catches up, the build will fail with a hash mismatch and
-  # print the correct value in the "got:" line of the error output.
-  vendorHash = pkgs.lib.fakeHash;
-
-  # Note: GOTOOLCHAIN=auto does NOT work inside the nix build sandbox because
-  # the sandbox has no network access, so Go cannot download a newer toolchain.
-  # When nixpkgs Go is older than what go.mod requires, the build will fail.
-  # This is intentional — it is the signal this CI job exists to surface.
+  vendorHash = "sha256-cQRIb5EbWg1BULyrFyHuRJmcx//72R15STEONRbgQ+A=";
 
   nativeBuildInputs = [ pkgs.git ];
+  buildInputs = [ pkgs.icu ];
 
   meta = with pkgs.lib; {
     description = "Dolt – a SQL database you can diff, branch, and merge";
diff --git a/flake.nix b/flake.nix
index d0dad5b816..e8c308b821 100644
--- a/flake.nix
+++ b/flake.nix
@@ -4,13 +4,15 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-  };
+    go-overlay.url = "github:purpleclay/go-overlay";
+};
 
   outputs =
     {
       self,
       nixpkgs,
       flake-utils,
+      go-overlay,
     }:
     flake-utils.lib.eachSystem
       [
@@ -22,8 +24,11 @@
       (
         system:
         let
-          pkgs = nixpkgs.legacyPackages.${system};
-          dolt = pkgs.callPackage ./default.nix { inherit pkgs self; };
+          pkgs = import nixpkgs {
+           inherit system;
+           overlays = [ go-overlay.overlays.default ];
+         };
+          dolt = pkgs.callPackage ./default.nix { inherit pkgs self; go = pkgs.go-bin.latest; };
         in
         {
           packages.default = dolt;
@@ -35,10 +40,11 @@
 
           devShells.default = pkgs.mkShell {
             buildInputs = with pkgs; [
-              go
+              go-bin.latest
               git
               gopls
               gotools
+             icu
             ];
```

turian:
Thanks for the thoughtful review and for actually building and testing the flake — really helpful.

**1) go-overlay with `go-bin.latest`**

I'd recommend against this for both beads and dolt:

- **Supply chain**: [purpleclay/go-overlay](https://github.com/purpleclay/go-overlay) is a 10-star repo with a single pseudonymous maintainer. It downloads pre-built Go binaries rather than building from source like nixpkgs. Adding it as a flake input means trusting that maintainer's CI pipeline and account security. Nix best practices [recommend minimizing non-nixpkgs inputs](https://nixcademy.com/posts/secure-supply-chain-with-nix/).
- **`go-bin.latest` includes release candidates**, so CI could break on a Go RC without any repo change.
- **It defeats the purpose of this CI job.** The whole point is visibility into when nixpkgs is behind. If we bypass nixpkgs' Go, we lose that signal — downstream nix users installing dolt from nixpkgs will still be affected.
- **Bus factor**: If purpleclay disappears, anything depending on the overlay breaks on the next flake update.

For beads, I already fixed Nix CI in [beads #1577](https://github.com/steveyegge/beads/pull/1577) using the durable approach: keep `flake.lock` fresh, avoid unnecessary Go version bumps, make CI honest.

**2) ICU dependency**

You're right. I confirmed via Docker — `dolthub/go-icu-regex` needs ICU headers (`unicode/uregex.h`). Your own `serverDockerfile` also installs `libicu-dev` for source builds. I'll add `buildInputs = [ pkgs.icu ];`.

**3) vendorHash semantics**

Yes, `vendorHash` needs updating whenever `go.mod`/`go.sum` change. This is the standard `buildGoModule` workflow. When wrong, the build fails and prints the correct value:

```
hash mismatch in fixed-output derivation:
  specified: sha256-AAAA...
  got:       sha256-XXXX...
```

For maintainers without nix installed, this can be done via Docker:

```bash
docker run --rm -v $(pwd):/workspace -w /workspace nixos/nix \
  sh -c 'echo "experimental-features = nix-command flakes" >> /etc/nix/nix.conf && nix build .#default 2>&1'
```

A CI job that auto-updates it is possible but complex. I'd suggest starting with the manual workflow and seeing how often it's needed.

**On the Go version situation**

The original bump commit cited [CVE-2025-61729](https://nvd.nist.gov/vuln/detail/CVE-2025-61729), but that was fixed in [Go 1.25.5](https://groups.google.com/g/golang-announce/c/8FJoBkPddm4). Go 1.25.6 fixed [6 different CVEs](https://www.openwall.com/lists/oss-security/2026/01/15/3) — real security fixes, but none introduce new language features or APIs.

The `go 1.25.6` directive is a security-policy floor, not a compilation requirement. For example, `dolthub/eventsapi_schema` was at `go 1.24.0` and got bumped to `go 1.25.6` as a side effect of `go mod tidy` — it doesn't use any 1.25.x features. nixpkgs-unstable currently ships 1.25.5; Go 1.25.6 was [merged to staging Jan 16](https://github.com/NixOS/nixpkgs/pull/480465) and [staging-next already has Go 1.25.7](https://github.com/NixOS/nixpkgs/commit/466bc97ad9) as of Feb 4. staging-next is ~2,800 commits ahead of unstable, so this should land in nixpkgs-unstable soon. If the ecosystem's go.mod files required only 1.25.5, the nix build would work today.

**What I'll update in this PR**

1. Add `buildInputs = [ pkgs.icu ];`
2. Keep nixpkgs Go (no go-overlay) — preserves the CI signal
3. Keep `vendorHash = lib.fakeHash` for now — can't compute until nixpkgs-unstable gets Go ≥ 1.25.6
4. Add a comment documenting the Docker trick for vendorHash updates
5. Keep `continue-on-error: true` — the job is informational

Once nixpkgs-unstable updates (should be soon given staging-next status), I'll push the correct vendorHash and we should get a full green build.

reltuk:
It's a fair point on supply chain for go-overlay and whether you want to take the dependency and/or fork.

I see the following problems with the submitted PR:

1) Without an automated way to keep flake.lock and `vendorHash` in default.nix up to date, and as a fallible CI job, I do not believe this job will almost ever been green. `go.mod` changes constantly and the job isn't even starting at green. I took a pass at fixing both of those problems here: #10464, but unfortunately the `vendorHash` is not even deterministic across environments. The CI fails in Github Actions but the package builds locally on macOS arm64 and on Linux amd64 and arm64.

2) `beads` does not technically care about the version of Go in `dolt`, but in `dolthub/driver`. The way we bump Go versions means we start at the top of the tree and work our way down. `dolthub/driver` gets an update before `dolthub/dolt`.

3) `beads` has its own lock file, its own snapshot of nixpkgs in its flake.lock. The system of record for whether a dolthub/driver bump in beads breaks CI is whether the PR which bumps the dependency breaks CI. Just because this job passes doesn't mean you can or should merge a dolthub/driver bump in beads.

4) Having a flake.nix and default.nix in the root of the dolt repository which is not held to a standard of always working causes confusion and misunderstanding for people who come across it and try to use it.

5) Currently `dolthub/driver` is not tagged for releases corresponding to `dolthub/dolt` releases. It is taken as a dependency on a given commit and unfortunately, for now, it is not immediately clear from the `go.mod` and `go.sum` entries which versions of `dolthub/dolt` a given `dolthub/driver` commit corresponds to. I'm not sure that the green/yellow status of a failing `nix` build CI job would ever make it into Dolt release notes.  Even if it did, I'm not sure if there would be a release corresponding to any given bump that `beads` was considering. Even if there was, as per point 1, a failing job here only means it failed with that particular snapshot of nixpkgs-unstable – when `beads` is bumping, maybe it works, or maybe when it did work for dolt because of its flake.lock but when `beads` is bumping it has an older nixpkgs-unstable and it won't work.

Which is all to say: I'm not convinced this approach provides enough value to support its cost. It seems like a better approach would be for `beads` to run its CI on its `dolthub/driver` bumps and to land those when the CI is green.

turian:
## Thank You & What I Learned

@reltuk — Thank you for actually cloning, building, and testing this flake, and for the detailed technical feedback. That's way more engagement than most maintainers give external PRs. Your counter-PR (#10464) also taught me a lot about what "doing this right" actually looks like in a high-churn repo — the automated vendorHash updates and weekly flake.lock refresh are the kind of infrastructure this needs. If DoltHub does want an in-repo flake, that's the right approach.

I dug into the Nix ecosystem, Dolt's CI culture, the `dolthub/driver` dependency chain, and how comparable Go projects handle packaging. Some of what I found surprised me and changes my position significantly.

---

## What I Found

### The real Nixpkgs blocker is ICU, not Go version lag

This was the biggest surprise. The [Nixpkgs Dolt package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/do/dolt/package.nix) has been stuck at v1.59.10 since September 2025. I assumed this was caused by Go 1.25.6 version lag. It's not. The [nixpkgs-update bot logs](https://nixpkgs-update-logs.nix-community.org/dolt/) show ~21 consecutive weekly failures, and the vast majority failed on:

```
unicode/uregex.h: No such file or directory
```

This started when `dolthub/go-icu-regex` was added as a CGO dependency in [PR #8681](https://github.com/dolthub/dolt/pull/8681) (Dec 2024). The Nixpkgs derivation is simply missing `buildInputs = [ icu ];`. The Go version mismatch only appeared in the [most recent failure](https://nixpkgs-update-logs.nix-community.org/dolt/) (Feb 4). My entire PR was addressing the wrong root cause.

### Beads uses Dolt as a library, not a binary — so this flake can't help Beads

This is the point you made that I initially pushed back on but now fully agree with. Beads embeds `dolthub/driver` via Go's `database/sql` interface ([`internal/storage/dolt/store.go`](https://github.com/steveyegge/beads/blob/main/internal/storage/dolt/store.go)). The `dolt` CLI binary is only needed for two optional features. A Nix flake that builds the dolt binary provides zero signal about whether the driver library compiles or is API-compatible with Beads. I was wrong to frame this as "early warning for Beads."

### dolthub/driver hasn't been tagged since July 2024

The [last real tag is v0.2.0](https://github.com/dolthub/driver/releases/tag/v0.2.0). Dozens of commits have landed since then without a release. Beads is forced to depend on pseudo-versions like `v0.2.1-0.20260129210815-dc59f6dbac23`, which means every upstream commit is a potential surprise. This is the actual root cause of the "Beads keeps breaking" problem — not Go versions, not Nix lag, but the lack of stable release points in the dependency chain.

### Dolt's CI culture is strictly pass/fail

Dolt has zero informational-only CI jobs across its 43 workflows. My proposal for `continue-on-error: true` was unprecedented in this repo. That reflects a deliberate engineering culture, and I should have respected it.

### Most Go projects don't carry in-repo flakes

In a survey of 15 popular Go CLIs ([gh](https://github.com/cli/cli), [fzf](https://github.com/junegunn/fzf), [terraform](https://github.com/hashicorp/terraform), [hugo](https://github.com/gohugoio/hugo), [grafana](https://github.com/grafana/grafana), [k9s](https://github.com/derailed/k9s), [prometheus](https://github.com/prometheus/prometheus), [lazygit](https://github.com/jesseduffield/lazygit), etc.), only 2 have in-repo flakes (lazygit, [gum](https://github.com/charmbracelet/gum)), and neither tests them in CI or automates vendorHash. The standard practice is to let Nixpkgs handle packaging.

### The Nix demand signal is real but lives in Nixpkgs, not here

Dolt has had only 2 Nix-related PRs in its entire history ([#6803](https://github.com/dolthub/dolt/pull/6803) in 2023 and this one). But the Nixpkgs side tells a different story: the bot has attempted [~21 weekly updates](https://nixpkgs-update-logs.nix-community.org/dolt/), there's an [open issue requesting updates](https://github.com/NixOS/nixpkgs/issues/352179), and Dolt was explicitly listed among packages broken by Go 1.22's vendorHash changes ([nixpkgs#300640](https://github.com/NixOS/nixpkgs/issues/300640)). The users exist — they just file issues in Nixpkgs, not here.

### The previous compromise repo rotted in 3 months

[PR #6803](https://github.com/dolthub/dolt/pull/6803) in 2023 proposed the same thing I did. You created [dolthub/dolt-nix-flake](https://github.com/dolthub/dolt-nix-flake) as a compromise. It has 3 stars and its last commit was January 2024. The compromise was reasonable at the time — out-of-repo packaging tends to fall behind without dedicated automation. That's another reason the Nixpkgs path is the right one.

### What I got wrong

- **Wrong root cause**: I blamed Go version lag; the real blocker is ICU.
- **Wrong framing**: "Early warning for Beads" doesn't work because Beads uses the library, not the binary.
- **Wrong cultural fit**: `continue-on-error` is not how this repo operates.
- **Incomplete submission**: `fakeHash`, missing ICU `buildInputs`, no automation.

### What we both missed

- Neither of us identified ICU as the actual Nixpkgs blocker until I dug into the [bot logs](https://nixpkgs-update-logs.nix-community.org/dolt/).
- The vendorHash mismatch in #10464 appears to be fixable — adding `proxyVendor = true` (matching what [Nixpkgs already uses](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/do/dolt/package.nix)) and computing the hash on Linux should resolve it ([relevant nixpkgs issue](https://github.com/NixOS/nixpkgs/issues/205842)). Happy to share debugging notes on the `buildGoModule` internals if useful.

---

## Recommendations

### 1. Close this PR

Feel free to close this — the premise was wrong (Go version lag → actually ICU), the framing was wrong (help Beads → Beads uses the library), and the execution was incomplete. I'll close it myself if you'd prefer not to deal with it.

### 2. dolthub/driver would benefit from tagged releases

This is the single highest-leverage improvement for downstream consumers. The [last tag is v0.2.0](https://github.com/dolthub/driver/releases/tag/v0.2.0) from July 2024. Beads (and any future consumer) is forced into pseudo-versions, which makes dependency management fragile. Even occasional tags aligned with Dolt releases would make a big difference. I understand this is a process/resourcing decision — just flagging it as the thing that would help most.

### 3. Beads should fix its own Nix build and pin to tagged driver releases

On the Beads side, the path forward is: pin to stable driver tags (once available), add automated `flake.lock` and vendorHash updates to CI, and stop trying to get upstream signal from Dolt's binary build.

### 4. Someone should submit the one-line ICU fix to Nixpkgs

Adding `buildInputs = [ icu ];` to [`pkgs/by-name/do/dolt/package.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/do/dolt/package.nix) would unblock 5 months of stalled updates for every Nix user of Dolt. I'm happy to submit this PR if DoltHub doesn't want to — it requires no changes to this repo.

---

Thanks again for the thorough review. I learned a lot from this process.

turian:
Following up on recommendation #4 above — I've submitted the one-line ICU fix to nixpkgs: [NixOS/nixpkgs#489380](https://github.com/NixOS/nixpkgs/pull/489380). Once merged, the [r-ryantm bot](https://nixpkgs-update-logs.nix-community.org/dolt/) should be able to bump dolt from v1.59.10 to the latest release.

reltuk:
@turian Thank you for your detailed response and for getting that PR out to nixpkgs. That is awesome. I think we have broad agreement on all of the points you make above.

I totally agree that a core piece that is causing pain is the missing `dolthub/driver` versions. It is a problem for `beads` and for any other consumer of `dolthub/driver`. I was going to bring it up in my previous comment, but I didn't have a near-at-hand suggestion for how we could fix it. In particular, I was looking for a concrete versioning scheme for the `dolthub/driver` releases. I was hoping to be able to come up with something that would do the following:

1) Cut a `dolthub/driver` release on every new `dolt` release.
2) Have the `dolthub/driver` version number reflect the `dolt` release which is embedded.
3) Have the possibility to release `dolthub/driver` only changes. These would bump the version number, but now in a way which obscured the `dolt` version.

Number 3 was the sticking point up to now, because Dolt already uses all 3 components of the Go module versioning scheme. I currently have some ideas and I am discussing with colleagues to determines pros and cons:

1) We could treat upcoming `dolthub/driver` releases as forever pegged at Dolt major version `1`. Then the first two components of the `driver` version number are are the last two components of the `dolt` version number, and the last version number starts at `0` and increments every time there is a change to `dolthub/driver` on an existing `dolt` release version.

2) Like number 1, but we explicitly make `dolthub/driver` into `dolthub/driver/v1`, since it will presumably become `dolthub/driver/v2` when Dolt 2.0 is released.

3) We just put the Dolt version number into the `dolthub/driver` release. We never release `dolthub/driver` changes without a corresponding `dolthub/dolt` release. If there is a critical bug we need to get a fix out for, we just cut ` dolthub/dolt` release as well.

One advantage of 1 and 2 is that release workflow is simpler. It also allows for naturally back porting `driver` specific fixes for previous versions of embedded `dolt`, if that is something someone is interested in or needs in the future. An advantage of `3` is that it keeps people on the automated upgrade train without any necessary code changes when something like Dolt 2.0 comes along.

A disadvantage of both is that they don't necessarily closely follow the Go module versioning guidelines for the semantics of the version numbers.

Anything is better than what we have today though.

Regardless of our decision, the above changes require some changes to our existing CI infrastructure. We can probably keep landing bumps of `dolt` on the `main` branch of `dolthub/driver` branch, but we need to detect or be notified about the dolt release and cut a corresponding appropriate release for `dolthub/driver` as well.

I have gone ahead and opened https://github.com/dolthub/driver/issues/86 in the `dolthub/driver` repository so we can track it.

I'm going to go ahead and close this pull request. Thank you again for you suggestions and contributions. Please feel free to continue the discussion around versioning on the `dolthub/driver` issue :).

PR #10459, opened 2026-02-10
[no-release-notes] adding perf regression label

jycor:
This PR adds a performance regression query.
It'll check if any of the mini benchmarks report a slow down of > 2.5% from main.
If the benchmarks show that the PR has negligible impact on performance, it'll mark the PR with a `performance_approved` label. This should take around 12 minutes to complete and automatically run when a PR is created.


.github/workflows/ci-check-performance.yaml
```diff
@@ -0,0 +1,21 @@
...
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 1
+          labels: "performance_approved"
+          add_comment: true
+          message: "This PR is being tested for SQL correctness. Please allow ~12 mins for this to complete. If this PR does not result in a performance regression, the `performance_approved` label will be automatically added to this PR."
```
coffeegoddd: This is for correctness

PR #10539, opened 2026-02-20
fix: add nil pointer checks to DoltDB initialization

zfogg:
## Summary

Add defensive nil pointer checks to prevent panics when DoltDB is uninitialized in certain code paths. These fixes address runtime crashes in embedded dolt usage.

## Changes

1. **SetCrashOnFatalError()** - Check if `ddb.db.Database == nil` before accessing
2. **GetRefsOfType() / VisitRefsOfType()** - Add nil checks before calling `ddb.db.Datasets()`
3. **PrependCommitHooks() / ExecuteCommitHooks()** - Add nil pointer validation before accessing DoltDB fields

## Motivation

These nil pointer dereferences occur when DoltDB is created but not fully initialized, which can happen in certain error recovery paths or early initialization stages. Adding checks prevents panics and allows graceful error handling.

## Testing

These changes maintain backward compatibility and add defensive programming practices. No behavioral changes to existing valid codepaths.

### DIFF
--- a/go/libraries/doltcore/doltdb/doltdb.go
+++ b/go/libraries/doltcore/doltdb/doltdb.go
@@ -2145,11 +2145,17 @@ func (ddb *DoltDB) DatasetsByRootHash(ctx context.Context, hashof hash.Hash) (da
 }
 
 func (ddb *DoltDB) PrependCommitHooks(ctx context.Context, hooks ...CommitHook) *DoltDB {
+	if ddb == nil || ddb.db.Database == nil {
+		return ddb
+	}
 	ddb.db = ddb.db.SetCommitHooks(ctx, append(hooks, ddb.db.PostCommitHooks()...))
 	return ddb
 }
 
 func (ddb *DoltDB) ExecuteCommitHooks(ctx context.Context, datasetId string) error {
+	if ddb == nil || ddb.db.Database == nil {
+		return nil
+	}
 	ds, err := ddb.db.GetDataset(ctx, datasetId)
 	if err != nil {
 		return err


coffeegoddd:
@zfogg you say "These nil pointer dereferences occur when DoltDB is created but not fully initialized, which can happen in certain error recovery paths or early initialization stages. Adding checks prevents panics and allows graceful error handling."

Can you provide any reproduction steps to the panics your encountered, either in Dolt or in the dolthub/driver? At the moment it's not clear to me that this is fixing anything, there's no test cases or context to determine if this is fixing a bug or masking it.

I think we need more details, any you can provide would be appreciated.

zfogg:
@coffeegoddd it was crashing for me when using with beads a version or two ago with gastown. i couldn't get dolt working. these fixes allowed dolt to run without crashing. i don't have exact version information but i was running the latest of everything when i made this pull request.

anyway you shouldn't access these things when they are nil so if you can't reason about why it might crash without them, just have the nil check as safety. it won't hurt and im telling you i needed these fixes for dolt to stop crashing with gastown and beads.

coffeegoddd:
@zfogg To clarify it's not that I don't believe your report, if that's how it came across, so if that's the case I'm sorry. We're really excited to be supporting the Beads and Gastown use-case for Dolt, and are committed to making sure everything works correctly.

It's important for our dev team to be able to triage and fixs bugs quickly, without introducing new ones, and so my hesitation re this PR is motivated by prevent any regressions or even potentially mask errors or other bugs from occuring. Returning `nil` errors when the underlying struct and database are `nil` would mask these.

I took a look at the latest Beads changes and from what I can tell Beads does not take a dependency on Dolt, only Gastown does. But without an logs, stack trace, reproduction steps, or tests, either recorded here or in Gastown, it's not clear to me that these changes fix a Dolt bug--it actually look like this change is more likely addressing a caller-side bug which is instantiating the DoltDB with nil values.

If you're able to provide any logs or reproduction steps, either here or in the Gastown repo, I'm happy to reconsider this PR, but we are not going to merge it at this time, without additional information.

timsehn:
We convinced Steve this was a bug in Gas Town, not Dolt.

PR #10549, opened 2026-02-23
dolthub/dolt#10488: Add --overwrite-ignore and --no-overwrite-ignore

### LINKED ISSUE #10488: Add `--overwrite-ignore` and `--no-overwrite-ignore` to `dolt checkout`
Dolt checkout now overwrites ignored tables when switching branches by default, matching Git.

Add explicit flags so users can control this: --overwrite-ignore to overwrite ignored tables (and make the default explicit) and --no-overwrite-ignore to abort when the target branch would overwrite ignored tables. Mirror the behavior described in [git-checkout(1)](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---overwrite-ignore) for these options.

codeaucafe:
This PR adds `--overwrite-ignore` and `--no-overwrite-ignore` to `dolt checkout` (for both CLI and SQL procedure). This attempts to mirror Git's behavior. By default, `--overwrite-ignore` is implied, and checkout silently overwrites ignored tables during a branch switch. When `--no-overwrite-ignore` is specified, checkout aborts if the target branch would overwrite ignored tables in the working set. Per Git docs, the check only triggers "when the new branch contains ignored files," so tables that only exist locally are not flagged.
- Register flag constants and argparser entries
- Add `overwriteIgnore bool` through all checkout paths: `doDoltCheckout` -> `checkoutNewBranch` / `checkoutExistingBranch` / `checkoutRemoteBranch` -> `doGlobalCheckout` -> `MoveWorkingSetToBranch` -> `transferWorkingChanges` -> `RootsForBranch`
- Add `FindOverwrittenIgnoredTables()` using `dolt_ignore` to detect ignored tables that differ between working set and target branch
- Add `ErrCheckoutWouldOverwriteIgnoredTables` error type with clear user-facing message

Closes #10488
Closes #10616

### DIFF
--- a/go/libraries/doltcore/env/actions/errors.go
+++ b/go/libraries/doltcore/env/actions/errors.go
@@ -149,7 +149,7 @@ func CheckoutWouldOverwriteTables(err error) []string {
 var ErrCheckoutWouldOverwriteIgnoredTables = goerrors.NewKind(
 	"The following ignored tables would be overwritten by checkout:\n\t%s\n" +
 		"Please move or remove them before you switch branches.\n" +
-		"Use --overwrite-ignore to force.")
+		"Use --overwrite-ignore to force.\n")
 
 type NothingStaged struct {
 	NotStagedTbls []diff.TableDelta


codeaucafe:
@elianddb i think this can be reviewed now: the one failing BATS test is that unrelated docked bats test that always fails for me, so I think we are good on that front. 

Thanks in advance for reviewing!

go/libraries/doltcore/env/actions/checkout.go
```diff
@@ -143,6 +143,10 @@ func RootsForBranch(ctx context.Context, roots doltdb.Roots, branchRoot doltdb.R
 		return doltdb.Roots{}, ErrCheckoutWouldOverwrite{conflicts.AsStringSlice()}
 	}
 
+	if err := CheckOverwrittenIgnoredTables(ctx, roots, branchRoot, overwriteIgnore); err != nil {
```
elianddb: `RootsForBranch` only handles root-move/conflicts mechanics, this is also seems to be already handled at a higher level `dolt_checkout_helpers.go`.

go/libraries/doltcore/env/actions/checkout.go
```diff
@@ -338,6 +342,64 @@ func moveModifiedTables(ctx context.Context, oldRoot, newRoot, changedRoot doltd
 	return resultMap, nil
 }
 
+// FindOverwrittenIgnoredTables returns ignored tables in |roots.Working| that have a different hash
+// in |branchRoot|. |branchRoot| is the HEAD root of the target branch being checked out.
+func FindOverwrittenIgnoredTables(ctx context.Context, roots doltdb.Roots, branchRoot doltdb.RootValue) ([]string, error) {
```
elianddb: This seems to only be used by another helper, you can inline its implementation for now. Unless, you find use for it under other existing infrastructure.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
```diff
@@ -118,11 +118,15 @@ func MoveWorkingSetToBranch(ctx *sql.Context, brName string, force bool, isNewBr
 	// If this is the case, then the destination branch must *not* have any uncommitted changes, as checked by
 	// checkoutWouldStompWorkingSetChanges
 	if hasChanges {
-		err = transferWorkingChanges(ctx, dbName, initialRoots, branchHead, branchRef, force)
+		err = transferWorkingChanges(ctx, dbName, initialRoots, branchHead, branchRef, force, overwriteIgnore)
```
elianddb: Same thing here, the functions signature is being extended for a feature that may not be applicable in other contexts. Checkout is responsible for this at a higher level.

integration-tests/bats/checkout.bats
```diff
@@ -1200,3 +1200,216 @@ SQL
     [[ "$output" =~ "* main" ]] || false
 }
 
+@test "checkout: --no-overwrite-ignore aborts when ignored table would be overwritten" {
```
elianddb, line 1203: These bats test are good, but add two `queries.ScriptTest` in the relevant testing infrastructure (i.e., `dolt_queries.go` checkout array) to make sure CLI and SQL procedures flags are both at least working. We typically now make CLI call the dolt procedures anyway to keep easy parity, so it may even be better to move some of the concrete tests that don't rely on CLI behavior to `dolt_queries.go`. Then keep any CLI-specific behavior here.

go/libraries/doltcore/sqle/dprocedures/dolt_checkout_helpers.go
elianddb: This seems to be already handled above (line 65),  not sure what's the purpose of this is here. I know it's unrelated to your changes, but you're working in here already so you can remove it.

elianddb (changes requested):
@codeaucafe Looks good! I've left specific nit picks on the implementation before merging.

codeaucafe:
Thank you @elianddb. I'll make corresponding changes tonight

codeaucafe:
@elianddb The --no-overwrite-ignore ScriptTest asserts an ExpectedErrStr containing the unqualified table name ignored_tbl. In DoltgreSQL, TableName.String() returns public.ignored_tbl, so the DoltgreSQL integration test fails on the contains check. Dolt's test harness uses exact match, so I can't shorten the expected string either. I see three options to handle here:                              
                                                                                                                                                                                           
1. Skip the test in DoltgreSQL harness                                                                                                                              
2. Use tbl.Name instead of tbl.String() in the error message so both environments produce ignored_tbl (though I don't think this is the better of the options since I think showing the schema-qualified name is more useful for DoltgreSQL users)
3. Something else you'd prefer                                                                                                                                                             
                                                                                                                                                                                           
Which approach works best here?

elianddb:
> @elianddb The --no-overwrite-ignore ScriptTest asserts an ExpectedErrStr containing the unqualified table name ignored_tbl. In DoltgreSQL, TableName.String() returns public.ignored_tbl, so the DoltgreSQL integration test fails on the contains check. Dolt's test harness uses exact match, so I can't shorten the expected string either. I see three options to handle here:
> 
> 1. Skip the test in DoltgreSQL harness
> 2. Use tbl.Name instead of tbl.String() in the error message so both environments produce ignored_tbl (though I don't think this is the better of the options since I think showing the schema-qualified name is more useful for DoltgreSQL users)
> 3. Something else you'd prefer
> 
> Which approach works best here?

Sorry for the late reply, I managed to get sick so I've been churning slower.

You don't have to match the error string exactly so you could just check for the part of the name after the ".". Also, if the error is an `error.Kind`, you should be able to use the `ExpectedErr` field in combination to match the error message.

codeaucafe:
ahh yea good call. I realized later today what I asked was kinda useless because of what you said. Thank you

hope you are feeling better

codeaucafe:
@elianddb i resolved and believe this is good to fully review again. The only failing workflows are the ones that generally fail for me as a contributor although it seems there is an issue with Maven install for Bats Unix workflow (have same issue in another PR). Ill check if that is fixed in master after work and if so pull/push updates

codeaucafe:
@elianddb thought me pulling your Maven fix from main would fix my BATS tests failing for Maven issue but i guess something is still up. I'm on vacation until early next week so I'll check it when I get back if I don't get to it tonight (still need to look at specific errors/logs)

codeaucafe:
@elianddb  I think this can be reviewed again. 
For failing workflow:
- sql-server integration: I believe TestCloneNoConjoin test failure is unrelated (could be wrong though?)
- Test bats unix remote: still having Maven failure issues. I thought there was recent merged PR that fixed this that I pulled in but still having this failure. Maybe its flaky or something else I'm missing?
- other failing workflows: these always fail for my PRs so I believe its expected due to be being outside contributor?

go/libraries/doltcore/env/actions/errors.go
```diff
@@ -144,6 +146,11 @@ func CheckoutWouldOverwriteTables(err error) []string {
 	return cwo.tables
 }
 
+var ErrCheckoutWouldOverwriteIgnoredTables = goerrors.NewKind(
+	"The following ignored tables would be overwritten by checkout:\n\t%s\n" +
+		"Please move or remove them before you switch branches.\n" +
+		"Use --overwrite-ignore to force.")
```
elianddb: Missing `\n` character here.

elianddb:
```suggestion
		"Use --overwrite-ignore to force.\n")
```

codeaucafe:
No problem at all, completely understandable.

Thank you for taking the time to review the PR!

PR #10571, opened 2026-02-25
[no-release-notes] Fix `mysql-client-tests` entry point to report status of all `bats` runs

elianddb:
The entry point was only reporting the failure of the last bats test ran. This left the failing `node` client tests unreported, these have also been fixed.


fulghum (approved):
Looks good!

PR #10590, opened 2026-02-27
Fix nil pointer panic in workspace table Update and Delete methods

nullun:
When StatementBegin encounters an error (e.g., table not found in staging root), it stores the error in wtu.err but leaves tableWriter as nil. The Update and Delete methods were dereferencing tableWriter before checking if it was nil, causing a panic.

This fix adds an early return to check for errors from StatementBegin before attempting to use tableWriter, preventing the nil pointer dereference.


jycor:
@nullun Thank for your the contribution. 
The change looks good, but do you have a way to reproduce the panic?

nullun:
> @nullun Thank for your the contribution. The change looks good, but do you have a way to reproduce the panic?

It happens when doing `dolt add -p` on a table that only exists in working. It's what lead me on to my other PR (https://github.com/dolthub/dolt/pull/10591), but I thought it would be better to keep this small fix on its own.

PR #10591, opened 2026-02-27
Support staging new tables with dolt add -p

nullun:
When using 'dolt add -p' to stage rows from a new table (one that exists in working but not in staging), the workspace table UPDATE mechanism previously failed with 'table not found' because GetTableWriter looked for the table in the staging root where it didn't exist yet.

This change adds ensureTableExistsInStaging() which:

1. Checks if the table exists in staging (fast path - no-op)
2. If not, checks if it exists in working
3. If yes, creates an empty table in staging with the same schema
4. Updates the session state to reflect the new staging root

The table is created empty (not copied with data) because 'dolt add -p' allows partial staging - each row selected by the user will be inserted individually into the staging table via the workspace table UPDATE.

Also adds unit tests and BATS integration tests for the new functionality.

Example usage: https://gist.github.com/nullun/e88cb2dab9568c6612c98d415a4a2efd

### DIFF
--- a/integration-tests/bats/add-patch.bats
+++ b/integration-tests/bats/add-patch.bats
@@ -349,12 +349,12 @@ teardown() {
   ! [[ "$output" =~ "charlie" ]] || false
 
   # Verify workspace table shows correct state
-  run dolt sql -q "SELECT staged, to_pk, to_name FROM dolt_workspace_new_table ORDER BY to_pk" -r csv
+  run dolt sql -q "SELECT IF(staged, 'true', 'false'), to_pk, to_name FROM dolt_workspace_new_table ORDER BY to_pk" -r csv
   [ $status -eq 0 ]
   # First row should be staged (true), others should not be (false)
-  [[ "$output" =~ "1,1,alice"   ]] || false
-  [[ "$output" =~ "0,2,bob"     ]] || false
-  [[ "$output" =~ "0,3,charlie" ]] || false
+  [[ "$output" =~ "true,1,alice"    ]] || false
+  [[ "$output" =~ "false,2,bob"     ]] || false
+  [[ "$output" =~ "false,3,charlie" ]] || false
 
   # Commit the staged changes
   dolt commit -m "partial staging test"


integration-tests/bats/add-patch.bats
```diff
@@ -324,3 +324,51 @@ teardown() {
...
+  # Verify the table is now in staging with partial changes
+  run dolt sql -q "SELECT * FROM new_table AS OF STAGED ORDER BY pk"
+  [ $status -eq 0 ]
+  [[ "$output" =~ "alice" ]] || false
+  # bob and charlie should NOT be in staged
+  ! [[ "$output" =~ "bob" ]] || false
+  ! [[ "$output" =~ "charlie" ]] || false
+
+  # Verify workspace table shows correct state
+  run dolt sql -q "SELECT staged, pk, name FROM dolt_workspace_new_table ORDER BY pk"
```
jycor:
```suggestion
  run dolt sql -q "SELECT staged, to_pk, to_name FROM dolt_workspace_new_table ORDER BY to_pk"
```

integration-tests/bats/add-patch.bats
```diff
@@ -324,3 +324,51 @@ teardown() {
...
+  [ $status -eq 0 ]
+  # First row should be staged (true/1), others should not be (false/0)
+  [[ "$output" =~ "1" ]] || false  # at least one staged
+  [[ "$output" =~ "0" ]] || false  # at least one not staged
```
jycor:
```suggestion
  [[ "$output" =~ "| true   | 1     | alice   |" ]] || false
  [[ "$output" =~ "| false  | 2     | bob     |" ]] || false 
  [[ "$output" =~ "| false  | 3     | charlie |" ]] || false 
```

integration-tests/bats/add-patch.bats
```diff
@@ -324,3 +324,51 @@ teardown() {
...
+  # Commit the staged changes
+  dolt commit -m "partial staging test"
+
+  # Verify only the first row is in the committed table
+  run dolt sql -q "SELECT * FROM new_table ORDER BY pk"
+  [ $status -eq 0 ]
+  [[ "$output" =~ "alice" ]] || false
+  ! [[ "$output" =~ "bob" ]] || false
+  ! [[ "$output" =~ "charlie" ]] || false
+
```
jycor: This will still select from the working set, which contains all three rows.

jycor (commented):
Looks like there are some failing tests.

integration-tests/bats/add-patch.bats
nullun: Whoops, thank you! I've updated this one to use `dolt show` instead, to verify what was commited.

btw, I like your avatar! 😄

nullun:
Hmmm, just got the email about the test failing again.

> 2026-03-02T18:50:45.2740284Z #   `[[ "$output" =~ "| true   | 1     | alice   |" ]] || false' failed

Not sure I understand why off the top of my head, unless the formatting is different somehow? I'll take a look in the morning if that's okay.

jycor:
Hmm... that should be right. It's possible it's something silly like a missing/extra space.
Tomorrow morning is totally fine, take your time.

The individual BATS statements can just be run the in the shell if that helps.

nullun:
Should have fixed it. I didn't realise there was a different when using `remote-engine`, which were what was failing. So my previous checks for "true" and "false" were actually "1" and "0". I thought simply using `-r csv` would make the output identical on both, but realised that wasn't fixing the actual problem, so now the SQL query itself will convert 1/0 to true/false.

The previously failing tests are now working, but let me know if you want me to tidy anything up or change it to be another way.
[Bats tests (ubuntu-22.04)](https://github.com/dolthub/dolt/actions/runs/22628204870/job/65570606707?pr=10591#logs)
> ok 61 add-patch: new table partial staging

[Bats tests](https://github.com/dolthub/dolt/actions/runs/22628204838/job/65570606024?pr=10591#logs)
> Using remote engine for tests
> ok 61 add-patch: new table partial staging

PR #10603, opened 2026-03-02
fix(binlog): handle empty TEXT/BLOB columns and missing chunks in serialization

### LINKED ISSUE #10601: BUG: Binlog serialization panics on TEXT/BLOB columns during DOLT_COMMIT
## Summary

When `log_bin=1` is enabled, `DOLT_COMMIT` on a table with TEXT or BLOB columns causes a server crash (lost connection). The crash occurs in `encodeBytesFromAddress` during binlog row serialization.

## Reproduction

1. Dolt v1.83.0 with `log_bin=1`, `gtid_mode=ON`, `enforce_gtid_consistency=ON`
2. Create a table with TEXT columns:
   ```sql
   CREATE TABLE test_text (id INT PRIMARY KEY, name VARCHAR(100), description TEXT);
   INSERT INTO test_text VALUES (1, 'test', 'hello world');
   INSERT INTO test_text VALUES (2, 'empty', '');
   INSERT INTO test_text VALUES (3, 'null', NULL);
   CALL DOLT_ADD('-A');
   CALL DOLT_COMMIT('-m', 'test commit');
   ```
3. Result: `Lost connection to MySQL server during query`

Observed on a production server with a table containing 11K+ rows and 4 TEXT columns.

## Stack Trace

```
assertion failed: empty chunk returned from ChunkStore
  node_cursor.go:639       (assertTrue — panic)
  node_store.go:103        (Read: assertTrue(c.Size() > 0, ...))
  node_store.go:212        (ReadBytes → Read)
  binlog_type_serialization.go:872 (encodeBytesFromAddress → ns.ReadBytes)
  binlog_type_serialization.go:739 (textSerializer.serialize)
  binlog_row_serialization.go:110  (serializer.serialize)
  binlog_producer.go:464           (DiffMaps callback, AddedDiff case)
```

## Root Cause

Two issues in `encodeBytesFromAddress`:

### 1. Empty TEXT/BLOB columns store a zero hash

When a TEXT column is empty (`''`) or the tuple encodes a default empty value, the tuple stores a zero hash (all zeros) as the blob address. `encodeBytesFromAddress` passes this zero hash to `ns.ReadBytes()`, which calls `ChunkStore.Get()` with the zero hash. The ChunkStore returns an `EmptyChunk`, triggering the assertion panic in `nodeStore.Read`.

### 2. Shared NodeStore cache masks ChunkStore misses

All NodeStores share a global `sharedCache` (`node_store.go:73`). When blobs are written via `BlobBuilder`, `nodeStore.Write()` inserts nodes into this cache. Under normal conditions, `Read()` finds nodes in the cache without hitting the ChunkStore.

On large tables with many TEXT columns, cache pressure causes evictions. When a blob's tree node is evicted, `Read()` falls through to `ChunkStore.Get()`, which can return `EmptyChunk` for chunks that are in-flight or in an intermediate NBS state during `writeWorkingSet`. The `assertTrue` assertion in `nodeStore.Read` converts this into an unrecoverable panic.

Existing tests (e.g., `TestTextSerializer`) pass because they use a single `MemoryStorage`-backed ChunkStore where blobs are always findable. The shared cache further masks any lookup issues in tests.

## Fix

I have a fix ready as a PR with:

1. **Zero-hash check**: `addr.IsEmpty()` guard in `encodeBytesFromAddress` that serializes empty TEXT/BLOB as a zero-length value, avoiding the ChunkStore lookup entirely
2. **Defensive error handling**: Convert the `assertTrue` panic in `nodeStore.Read` to a returned error so the server stays up and logs the issue
3. **Error context**: Wrap errors in `encodeBytesFromAddress` with address and type info for debugging
4. **Test**: `TestTextSerializer_MissingChunk` that reproduces the ChunkStore miss using separate storage backends and cache purging

Tested on a production server — INSERT + DOLT_COMMIT with empty, null, and populated TEXT columns all succeed with binlog enabled after the fix.

scbrown:
Fixes #10601

## Summary

When `log_bin=1` is enabled, `DOLT_COMMIT` on a table with TEXT/BLOB columns panics in `encodeBytesFromAddress` because:

1. Empty TEXT/BLOB columns store a zero hash in the tuple — the serializer tries to load this from the ChunkStore, which returns EmptyChunk
2. The `assertTrue` in `nodeStore.Read` converts ChunkStore misses into unrecoverable panics

## Changes

- **`binlog_type_serialization.go`**: Add `addr.IsEmpty()` check in `encodeBytesFromAddress` — serialize empty TEXT/BLOB as zero-length value with correct length prefix size for the blob type. Wrap errors with address and type context.
- **`node_store.go`**: Convert `assertTrue(c.Size() > 0, ...)` panic to a returned `fmt.Errorf` so the server stays up and logs the issue.
- **`binlog_type_serialization_test.go`**: Add `TestTextSerializer_MissingChunk` that reproduces the ChunkStore miss using separate storage backends and explicit cache purging.

## Testing

- Unit tests pass (go vet clean, TestTextSerializer suite, new TestTextSerializer_MissingChunk)
- Verified on a production Dolt v1.83.0 server with 11K+ row table containing 4 TEXT columns
- INSERT + DOLT_COMMIT with empty string, NULL, and populated TEXT columns all succeed with binlog enabled

### DIFF
--- a/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization.go
+++ b/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization.go
@@ -866,12 +866,25 @@ func encodeBytes(data []byte, typ sql.Type) ([]byte, error) {
 // in the returned |data| slice. The |typ| parameter is used to determine the maximum byte length of the serialized
 // type, in order to determine how many bytes to use for the length prefix.
 func encodeBytesFromAddress(ctx *sql.Context, addr hash.Hash, ns tree.NodeStore, typ sql.Type) (data []byte, err error) {
+	if addr.IsEmpty() {
+		// Empty TEXT/BLOB column — the tuple stores a zero hash (no out-of-band data).
+		// Serialize as a zero-length value instead of attempting a ChunkStore lookup.
+		blobType := typ.(sql.StringType)
+		if blobType.MaxByteLength() > 0xFFFFFF {
+			return []byte{0, 0, 0, 0}, nil
+		} else if blobType.MaxByteLength() > 0xFFFF {
+			return []byte{0, 0, 0}, nil
+		} else if blobType.MaxByteLength() > 0xFF {
+			return []byte{0, 0}, nil
+		}
+		return []byte{0}, nil
+	}
 	if ns == nil {
 		return nil, fmt.Errorf("nil NodeStore used to encode bytes from address")
 	}
 	bytes, err := ns.ReadBytes(ctx, addr)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("binlog serialization: failed to read out-of-band data at address %s for type %s: %w", addr.String(), typ.String(), err)
 	}
 
 	blobType := typ.(sql.StringType)
--- a/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization_test.go
+++ b/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization_test.go
@@ -1013,6 +1013,49 @@ func TestTextSerializer(t *testing.T) {
 	})
 }
 
+// TestTextSerializer_MissingChunk verifies that when a TEXT column's out-of-band
+// blob chunk is missing from the ChunkStore, the serializer returns an error instead
+// of panicking. This reproduces a production crash where encodeBytesFromAddress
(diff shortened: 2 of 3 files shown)


fulghum:
Thank you for sending this PR. This was helpful to quickly see what was going on in the code. I've made a slightly different change in https://github.com/dolthub/dolt/pull/10621, so I'll close out this PR.

Issue #9001, opened 2025-03-18
Server panic: empty chunk returned from ChunkStore


nicktobey:
This happens rarely but I've seen it twice now over the past couple days.

Platform: WSL
Version: 1.49.3
Stack trace:

```
panic: assertion failed: empty chunk returned from ChunkStore

goroutine 321 [running]:
github.com/dolthub/dolt/go/store/prolly/tree.assertTrue(...)
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:634
github.com/dolthub/dolt/go/store/prolly/tree.nodeStore.Read({{0x38af5e8, 0xc000130030}, {{0xc001bbcf90, 0xc001bbcff0, 0xc001bbd050, 0xc001bbd0b0, 0xc001bbd110, 0xc001bbd170, 0xc001bbd1d0, 0xc001bbd230, ...}}, ...}, ...)
        github.com/dolthub/dolt/go/store/prolly/tree/node_store.go:100 +0x5f9
github.com/dolthub/dolt/go/store/prolly/tree.fetchChild(...)
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:629
github.com/dolthub/dolt/go/store/prolly/tree.(*cursor).fetchNode(...)
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:560
github.com/dolthub/dolt/go/store/prolly/tree.(*cursor).advance(0xc0018de180, {0x3815028, 0xc034953c20})
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:512 +0x165
github.com/dolthub/dolt/go/store/prolly/tree.(*cursor).advance(0xc0018de200, {0x3815028, 0xc034953c20})
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:500 +0x76
github.com/dolthub/dolt/go/store/prolly/tree.(*cursor).advance(0xc0018de280, {0x3815028, 0xc034953c20})
        github.com/dolthub/dolt/go/store/prolly/tree/node_cursor.go:500 +0x76
github.com/dolthub/dolt/go/store/prolly/tree.(*OrderedTreeIter[...]).Next(0xc039860790, {0x3815028?, 0xc034953c20})
        github.com/dolthub/dolt/go/store/prolly/tree/map.go:579 +0x183
[...]

PR #10604, opened 2026-03-02
fix(binlog): add GC safepoint for binlog producer to prevent chunk deletion during traversal

### LINKED ISSUE #10602: BUG: Binlog producer does not participate in GC safepoints — concurrent DOLT_GC can delete chunks mid-traversal
## Summary

The binlog producer's `WorkingRootUpdated` method traverses Prolly trees via `prolly.DiffMaps` to generate row events for connected replicas. This traversal is not protected against concurrent garbage collection. If `DOLT_GC` runs while the binlog producer is mid-traversal, GC can delete chunks that the traversal still needs, causing "empty chunk returned from ChunkStore" panics.

## Reproduction

This is a race condition that requires specific timing:

1. Enable `log_bin=1` on a Dolt server
2. Have a workload generating frequent commits (binlog producer actively serializing diffs)
3. Run `DOLT_GC` concurrently
4. If GC deletes chunks that the binlog producer's `DiffMaps` iterator needs, the server panics

The window is wider on tables with many rows or large TEXT/BLOB columns where Prolly tree traversal takes longer.

## Analysis

The binlog producer registers as a `DatabaseUpdateListener`. When `WorkingRootUpdated` is called, it:

1. Calls `diff.GetTableDeltas(ctx, before, after)` to find changed tables
2. Iterates via `prolly.DiffMaps` which walks the Prolly tree
3. For each diff, serializes rows via `serializeRowToBinlogBytes`

This entire operation reads chunks from the ChunkStore. If GC runs concurrently and collects chunks that the diff iterator hasn't visited yet, the iterator panics when it tries to read the deleted chunk.

## Fix

I have a fix ready as a PR that adds a `GCPausableListener` interface:

1. **New interface `GCPausableListener`** in `doltdb` package with `Stop() chan struct{}` and `Resume()` methods
2. **`binlogProducer` implements it** using `sync.RWMutex` — `WorkingRootUpdated` holds a read lock during traversal, `Stop()` acquires a write lock (blocking until all in-flight traversals complete)
3. **`RunDoltGC` pauses listeners** before running GC and resumes them after
4. **Comprehensive test suite** covering: no-inflight drain, inflight wait, multiple concurrent inflight, stop/resume cycles, and concurrent stop+read contention

This ensures GC cannot delete chunks while any binlog producer traversal is in flight, and new traversals block while GC is running.

scbrown:
Fixes #10602

## Summary

The binlog producer traverses Prolly trees via `prolly.DiffMaps` during `WorkingRootUpdated`, but this traversal is not protected against concurrent `DOLT_GC`. If GC deletes chunks mid-traversal, the server panics.

## Changes

- **`doltdb.go`**: Add `GCPausableListener` interface with `Stop() chan struct{}` and `Resume()` methods
- **`binlog_producer.go`**: Implement `GCPausableListener` using `sync.RWMutex` — `WorkingRootUpdated` holds RLock during diff traversal, `Stop()` acquires write lock (blocks until in-flight traversals drain)
- **`dolt_gc.go`**: Before running GC, iterate `DatabaseUpdateListeners` and call `Stop()` on any that implement `GCPausableListener`. Resume all after GC completes.
- **`binlog_producer_gc_test.go`**: Comprehensive test suite covering no-inflight drain, inflight wait, multiple concurrent inflight, stop/resume idempotency, and concurrent stop+WorkingRootUpdated contention

## Design

Uses `sync.RWMutex` for minimal overhead on the hot path (WorkingRootUpdated only takes a read lock). The `Stop()` method returns a channel so the caller can wait asynchronously for all in-flight operations to drain before proceeding with GC.

## Testing

- All new tests pass including race detector (`go test -race`)
- Verified interface satisfaction at compile time

### DIFF
--- a/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization.go
+++ b/go/libraries/doltcore/sqle/binlogreplication/binlog_type_serialization.go
@@ -869,9 +869,15 @@ func encodeBytesFromAddress(ctx *sql.Context, addr hash.Hash, ns tree.NodeStore,
 	if ns == nil {
 		return nil, fmt.Errorf("nil NodeStore used to encode bytes from address")
 	}
-	bytes, err := ns.ReadBytes(ctx, addr)
-	if err != nil {
-		return nil, err
+
+	// A zero hash address means empty content (e.g. empty string stored without
+	// an out-of-band chunk). Skip the ChunkStore lookup and encode zero-length data.
+	var bytes []byte
+	if !addr.IsEmpty() {
+		bytes, err = ns.ReadBytes(ctx, addr)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	blobType := typ.(sql.StringType)


reltuk:
Hi @scbrown. Thanks for the report!

Can you provide an example server configuration where you saw the panic?

I do not believe the approach in this PR is the approach we would want to take. But looking at the code, it does seem possible that the retained roots tracking and GC safepoints on the binlogreplication diff path have been misrepresented in some way. Being able to test more deeply and dig into the root cause would be helpful.

scbrown:
Hi @reltuk, thanks for the quick response!

## Server configuration

```yaml
# config.yaml
listener:
  host: 0.0.0.0
  port: 3306
  max_connections: 500

remotesapi:
  port: 50051
```

```json
// .dolt/config.json (persisted system variables)
{
  "sqlserver.global.log_bin": "1",
  "sqlserver.global.gtid_mode": "ON",
  "sqlserver.global.enforce_gtid_consistency": "ON",
  "sqlserver.global.binlog_row_metadata": "FULL",
  "sqlserver.global.dolt_transaction_commit": "1"
}
```

Running Dolt v1.83.0 on Linux (Ubuntu 22.04, go1.23.4).

## What we actually hit

To be transparent: the panic we reproduced in production was **not** from a GC race. It was the zero-hash issue described in the companion PR #10603.

Our production table has ~11K rows with 7 TEXT columns, many containing empty strings (`''`). Empty strings store a zero hash as the blob address in the prolly tuple. `encodeBytesFromAddress` passes this zero hash to `ns.ReadBytes()`, which calls `ChunkStore.Get()` with hash zero → EmptyChunk → `assertTrue` panic.

We filed this GC PR separately because we noticed `WorkingRootUpdated` traverses prolly trees via `DiffMaps` and we weren't sure if this path participates in the existing GC safepoint infrastructure. It seemed worth flagging given the similarity to #9001/#9077 (stats path).

## Custom binary testing

We built a custom binary from our fork with both fixes applied and tested on our production server:

1. **Phase 1** (defensive): Converted the `assertTrue` panic in `nodeStore.Read` to a returned error + wrapped errors in `encodeBytesFromAddress` with address/type context. Result: server stays up, logs the error gracefully, commits succeed.

2. **Phase 2** (root cause): Added `addr.IsEmpty()` check in `encodeBytesFromAddress` — if the address is all zeros, serialize as a zero-length value with the correct length-prefix size for the blob type. Result: **no errors at all**, clean binlog serialization for empty, NULL, and populated TEXT columns.

Test matrix on production data (11K rows, 7 TEXT columns):
- INSERT with non-empty TEXT + DOLT_COMMIT → PASS
- INSERT with empty string TEXT (`''`) + DOLT_COMMIT → PASS
- INSERT with NULL TEXT + DOLT_COMMIT → PASS
- Mixed columns (some empty, some populated, some NULL) + DOLT_COMMIT → PASS
- Multiple consecutive commits with binlog ON → Server stable, no panics or errors

## On GC safepoints

Your point about retained roots tracking is well taken. If the binlog replication diff path already participates in GC safepoints through a mechanism we missed, this PR may be unnecessary. The zero-hash fix in #10603 is the one that addresses our actual crash. Happy to close this one if you'd prefer to handle the GC coordination through your existing infrastructure.

Let me know if you'd like any additional detail or testing on either PR!

timsehn:
Closing per discussion. We have a fix for the real bug you found.

PR #10618, opened 2026-03-03
dolthub/dolt#10058 feat(pull): add --rebase flag to dolt pull

### LINKED ISSUE #10058: [Feature Request] dolt pull --rebase
It would be kinda neat to be able to combine the new rebase functionality with the pull functionality like in Git, I almost lost data over this missing and it wasn't trivial to imitate it's behavior and save my data...

codeaucafe:
### Changes                                                                                          
- Add `--rebase`/`-r` flag to `dolt pull` (CLI) and `CALL dolt_pull` (SQL) so that after fetching, local commits are rebased onto the upstream branch instead of merging 
- Handles three cases: up-to-date, fast-forward, and diverged history (calls `startRebase` + `continueRebase`)                                                           
- Validates flag conflicts (`--rebase` vs `--squash`, `--no-ff`, `--ff-only`, `--no-commit`)                                                                             
- CLI conflict resolution sets `@@dolt_allow_commit_conflicts` and syncs to the temp rebase branch so users can resolve and `dolt rebase --continue`

### Tests:
- 21 CLI BATS tests in `pull.bats`
- 19 SQL BATS tests in `sql-pull.bats`
- covers: divergent history, fast-forward, up-to-date, flag conflicts, data/schema conflict resolution (--theirs/--ours), multi-commit rebase, empty commit drop,
uncommitted changes guard, abort and retry, branch collision, explicit remote+branch args, conflict state inspection during paused rebase

Closes #10058

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_pull.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_pull.go
@@ -326,35 +326,3 @@ func performRebaseAfterPull(ctx *sql.Context, upstreamPoint string, skipVerifica
 
 	return noConflictsOrViolations, threeWayMerge, result.message, nil
 }
-
-// TODO: remove this as it does not do anything useful
-func pullerProgFunc(ctx context.Context, statsCh <-chan pull.Stats) {
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case <-statsCh:
-		}
-	}
-}
-
-// TODO: remove this as it does not do anything useful
-func runProgFuncs(ctx context.Context) (*sync.WaitGroup, chan pull.Stats) {
-	statsCh := make(chan pull.Stats)
-	wg := &sync.WaitGroup{}
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		pullerProgFunc(ctx, statsCh)
-	}()
-
-	return wg, statsCh
-}
-
-// TODO: remove this as it does not do anything useful
-func stopProgFuncs(cancel context.CancelFunc, wg *sync.WaitGroup, statsCh chan pull.Stats) {
-	cancel()
-	close(statsCh)
-	wg.Wait()
-}


go/cmd/dolt/commands/pull.go
```diff
@@ -128,6 +145,13 @@ func (cmd PullCmd) Exec(ctx context.Context, commandStr string, args []string, d
 			errChan <- err
 			return
 		}
+		if apr.Contains(cli.RebaseParam) {
```
macneale4, line 154: We currently require users to set this flag themselves. Should we extend this to other types of merges for the cli? put another way, why is this just Rebase?

macneale4: (this is mainly a question for myself)

codeaucafe:
Ahh yea sorry I just realized the failing bats tests. Originally these tests weren't getting to the actual bats tests execution in the bats workflow, although I thought they were passing previously. I'll take a look and fix tonight

codeaucafe:
Small update. Started but didn't finish fixing those 6 failed tests. Should have something tomorrow though. Will post an update when it can be reviewed again.

go/cmd/dolt/commands/pull.go
codeaucafe: @macneale4 not sure I completely follow your question (to yourself). Are you trying to determine why we don't set @@dolt_allow_commit_conflicts = 1 for git pull merge? If so, wouldn't that allow users to commit conflicts (in dolt conflicts table) which is probably what we don't want to allow since it would allow conflicts to be committed, no?

codeaucafe:
@macneale4 I think you can review again. I think I handled the failing tests correctly. Regarding the still failing workflows:
- Bats unix remote: I'm not sure why I keep getting this Maven failure. I though there was a recent PR that resolved, but maybe I misunderstood, or its flaky/failing due to me being outside contributor?
- Test bingo worflow: I believe this is flaky failure unrelated?
- Test integration with DoltgreSQL: I think the panic failure is possible race condition?
- the others: always failing for my PRs (I believe this expected)

codeaucafe:
@macneale4  can you review again please. Thank you

integration-tests/bats/pull.bats
```diff
@@ -716,6 +716,7 @@ SQL
 }
 
 @test "pull: pull --rebase with data conflict pauses rebase" {
+    skip_if_remote
```
macneale4, line 722: why are we skipping this tests? Same goes for all the other skipped tests below. these should work.

go/cmd/dolt/commands/pull.go
macneale4: It's not clear to me why this required, actually. We've already set the `dolt_force_transaction_commit` flag, which should supercede the `dolt_allow_commit_conflicts` - I think. But ultimately, rebase and regular merges should be handled in the same way.

codeaucafe:
Thank you. I'll take a look tomorrow. I put the reason for skipping those tests in last commit msg BUT I'm starting to think my understanding was wrong when I was debugging. 

I'll circle back tomorrow with a better comment detailing why to make sure my understanding is wrong or right.

integration-tests/bats/pull.bats
codeaucafe: @macneale4 

I originally skipped these tests because `rebase.bats` is in `SKIP_SERVER_TESTS` in `local-remote.bash`, so I assumed rebase in remote-engine mode was a known unsupported case. But `pull.bats` isn't in that list, so I'm guessing these should run

From my understanding, the tests were failing because in remote-engine mode, `dolt pull --rebase` correctly writes `dolt_rebase_main` to `repo_state.json` after a conflict, but subsequent CLI commands (dolt conflicts resolve, `dolt rebase --continue`) each start as a fresh process and connect to the server without a branch qualifier, so the server puts them on main instead of `dolt_rebase_main`. Correct me if that's wrong. 
                                                                                                                                                                                                                            One possible fix: in `buildLateBinder` (`dolt.go`), when a CLI command auto-detects a local server and useDb has no branch qualifier, read the branch from `repo_state.json` and append it (e.g. repo/dolt_rebase_main). Tests pass with this, but I'm not confident it's right - it changes connection behavior for all CLI commands in local-server mode, not just rebase. Feels broad.

Is conflict resolution during rebase in remote-engine mode intended to be supported? Happy to implement if so, just want to make sure I'm approaching it correctly.

go/cmd/dolt/commands/pull.go
codeaucafe: @macneale4 you're right that we set `@@dolt_force_transaction_commit = 1` before the pull query runs, but from my understanding `validateConflictsCanBeResolved` in `dolt_rebase.go:1087` only checks `@@dolt_allow_commit_conflicts` and `@@autocommit` and never reads `@@dolt_force_transaction_commit`. So, even though it's set on the session, the rebase stored procedure ignores it.

Should I add a check for `@@dolt_force_transaction_commit` in `validateConflictsCanBeResolved` and remove the explicit set `@@dolt_allow_commit_conflicts = 1` from `pull.go` so the behavior is consistent?

integration-tests/bats/pull.bats
macneale4: I kind of got nerd-sniped on rebase not working. Turns out that the rebase cli command can't operate against a remote-server. We hadn't explicitly prevented it before, but we will now. (https://github.com/dolthub/dolt/pull/10818)

The reason for this is that rebase checks out a temporary branch to operate, and we don't have a way to do that on the CLI when talking to a remote-server. it's the same reason we can't `dolt checkout` when a server is running. 

So I think your original suggestion is actually pretty close. There should probably be a test to indicate that `dolt pull` and `dolt pull --rebase` both refuse to operate against a running server.

Sorry this has been a lot of back and forth. I think now that I understand that remote server is off the table, the end should be in sight. Thanks for all your help!!

codeaucafe: hey @macneale4 no problem at all. All the back and forth has only improved my understanding of Dolt so I enjoy it actually. Glad to hear my initial suggestion was close.

I'll try and get those changes in sometime this week. 

thank you.

go/cmd/dolt/commands/pull.go
macneale4: It's ok. we'll punt for the time being. Not urgent.

codeaucafe:
Sure thing. Thanks @macneale4. I'll take a look at your branch to for my understanding as well.

PR #10628, opened 2026-03-04
fix: use redirect URL to fetch latest Dolt version

nullun:
I saw my PR (#10591) failing to this test.

~~I believe docker is expanding the `DOLT_VERSION` ARG before bash execution. So when set to `latest`, the dynamically fetched version number was being overwritten, causing the script to try and download `/vlatest/install.sh` and fails.~~

~~This commit introduces an internal `ACTUAL_VERSION` variable in the bash script to bypass Docker's string substitution.~~

The `download-binary` stage fetches the latest Dolt version from the GitHub API. In CI, unauthenticated requests to the GitHub API can be rate limited, causing the request to fail silently. The version variable ends up empty, the install script URL is malformed, and the build fails with a confusing `Not: command not found` error.

This is a pre-existing issue unrelated to my other changes.

Hopefully this solves the mystery of the flakey test #10625

### DIFF
--- a/docker/serverDockerfile
+++ b/docker/serverDockerfile
@@ -43,15 +43,13 @@ FROM base AS download-binary
 ARG DOLT_VERSION
 RUN if [ "$DOLT_VERSION" = "latest" ]; then \
         # Fetch latest version number from GitHub API
-        ACTUAL_VERSION=$(curl -sI https://github.com/dolthub/dolt/releases/latest \
+        DOLT_VERSION=$(curl -sI https://github.com/dolthub/dolt/releases/latest \
             | grep -i location \
             | sed 's|.*/v||' \
             | tr -d '[:space:]'); \
-    else \
-        ACTUAL_VERSION=$DOLT_VERSION; \
     fi && \
-    if [ "$ACTUAL_VERSION" != "source" ]; then \
-        curl -L "https://github.com/dolthub/dolt/releases/download/v${ACTUAL_VERSION}/install.sh" | bash; \
+    if [ "$DOLT_VERSION" != "source" ]; then \
+        curl -L "https://github.com/dolthub/dolt/releases/download/v${DOLT_VERSION}/install.sh" | bash; \
     fi
 
 
--- a/docker/serverDockerfile
+++ b/docker/serverDockerfile
@@ -43,13 +43,16 @@ FROM base AS download-binary
 ARG DOLT_VERSION
 RUN if [ "$DOLT_VERSION" = "latest" ]; then \
         # Fetch latest version number from GitHub API
-        DOLT_VERSION=$(curl -s https://api.github.com/repos/dolthub/dolt/releases/latest \
+        # Use a new variable name so Docker doesn't overwrite it
+        ACTUAL_VERSION=$(curl -s https://api.github.com/repos/dolthub/dolt/releases/latest \
             | grep '"tag_name"' \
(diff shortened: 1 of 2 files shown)


nullun:
Just realised I should probably remove the `skip` that was introduced. Reverted 29dba59c89f0f620338ca4dbbc15ddac833b9771

nullun:
Well, that was a journey... Sorry for performing so many tests. My initial attempt didn't fix it, but after avoiding the API rate limit it appears to have solved the issue. On reflection I do still think my first commit fixes an issue, as one of the logs showed `/vlatest/` in the URL, so I'll add one more commit to use `ACTUAL_VERSION` again.

coffeegoddd:
i'm still thinking about this... I don't love that we fetch the latest binary from github at all in this test, where I think the intention is to  demonstrate that the server dockerfile on this branch has not had a regression. @elianddb can maybe chime in here. I think to fix this we should make building the docker image successfully one test, and we should build it from the source so its not making and http request, and if the build exits with 0, it passes. then in subesquent tests we can use that built image as a dependency, so if it exists, run tests using it, or if it doesnt exist, skip the tests. something like that...

docker/serverDockerfile
```diff
@@ -43,13 +43,15 @@ FROM base AS download-binary
 ARG DOLT_VERSION
 RUN if [ "$DOLT_VERSION" = "latest" ]; then \
         # Fetch latest version number from GitHub API
-        DOLT_VERSION=$(curl -s https://api.github.com/repos/dolthub/dolt/releases/latest \
-            | grep '"tag_name"' \
-            | cut -d'"' -f4 \
-            | sed 's/^v//'); \
+        ACTUAL_VERSION=$(curl -sI https://github.com/dolthub/dolt/releases/latest \
```
elianddb: You can leave this as `DOLT_VERSION` and keep your changes using the non-api path.

integration-tests/bats/docker-entrypoint.bats
```diff
@@ -925,13 +925,11 @@ EOF
...
 # bats test_tags=no_lambda
 @test "docker-entrypoint: latest binary build from dolt directory" {
-  skip "this is way too flaky in CI and needs investigation into why/ how to stabilize"
-
   BATS_TEST_RETRIES=5 # GitHub ver. retrieval can sometimes return Not Found on CI, could be some form of rate limiting
   cname="${TEST_PREFIX}latest-docker"
 
   LATEST_IMAGE="dolt-entrypoint-latest:test"
-  EXPECTED_VERSION=$(curl -s https://api.github.com/repos/dolthub/dolt/releases/latest | grep '"tag_name"' | cut -d'"' -f4 | sed 's/^v//')
+  EXPECTED_VERSION=$(curl -sI https://github.com/dolthub/dolt/releases/latest | grep -i location | sed 's|.*/v||' | tr -d '[:space:]')
```
elianddb, line 932: I've spoken with @coffeegoddd and the main issue with this is we're relying on an external call to github. Our best bet would be to check if the docker code contains this line itself. This also goes for the test below.

elianddb (changes requested):

elianddb:
@nullun I think it would be best so we don't rely on external Github calls to check for the existence of the specific links in the Dockerfile. I do think it's fine to update to not use the api link version tho, if it helps us avoid rate limiting.

PR #10686, opened 2026-03-16
sqle: skip dolt commit path for read-only transactions

### LINKED ISSUE #10685: dolt_transaction_commit=true causes empty commit warning storm on read-heavy workloads
## Summary

When `dolt_transaction_commit=true` is set in the Dolt SQL server config, every read-only SQL query (SELECT, SHOW, etc.) triggers an automatic `COMMIT` attempt at transaction close. Since reads have no changes, Dolt logs "nothing to commit" warnings. In a multi-client environment with frequent polling queries, this generates **100k+ warnings over 5 hours** and **718k connections**, eventually overwhelming the server.

## Reproduction

1. Start `dolt sql-server` with `dolt_transaction_commit: true` in config.yaml
2. Run multiple clients doing frequent SELECT queries (e.g., 6 agents polling every 30s)
3. Observe "nothing to commit" warning log entries accumulating rapidly
4. Over hours, connection IDs climb into the hundreds of thousands
5. Server eventually degrades — queries return `database not found` errors, then `connection refused`

## Expected behavior

Read-only transactions should not attempt a commit when `dolt_transaction_commit=true`. The setting should only trigger automatic commits when the transaction actually has write operations (INSERT, UPDATE, DELETE, etc.).

## Observed behavior

Every connection closure triggers a commit attempt regardless of whether writes occurred. The "nothing to commit" path still consumes resources — log I/O, connection tracking, internal locking — which compounds at scale.

## Timeline from production incident

- 6 concurrent SQL clients, each polling every 30s
- "nothing to commit" warnings: 451/hr → 11,000/hr → 41,000/hr (escalating)
- Connection IDs reached 717,000+
- Server began returning `database not found` for valid databases
- Server process crashed — `connection refused` on port 3307

## Workaround

Set `dolt_transaction_commit: false` and handle commits explicitly in application code. This eliminates the warning storm entirely.

## Environment

- Dolt SQL server (latest as of March 2026)
- Multi-client setup with 6 concurrent connections
- Linux (WSL2) and macOS
- Read-heavy workload (~95% SELECT, ~5% writes)

## Related

- dolthub/dolt#7187 — `dolt_transaction_commit` persistence issue
- dolthub/dolt#5704 — latency with `dolt_transaction_commit=1`

DreadPirateRobertz:
## Summary

- Skip the expensive `PendingCommitAllStaged` path for read-only transactions when `dolt_transaction_commit=true`
- Read-only transactions (SELECTs) go directly to `commitWorkingSet`, matching the behavior they'd get anyway but without staging overhead
- Prevents empty commit attempt storm in high-frequency read workloads

## Problem

When `dolt_transaction_commit=true`, every transaction close—including read-only SELECTs—runs `PendingCommitAllStaged` which calls `StageAllTables` and `newPendingCommit`. For read-only queries there are never changes to stage, so `pendingCommit` always returns nil and the code falls through to `commitWorkingSet` anyway.

In production with 6 agents polling every 30 seconds, we observed:
- 100k+ empty commit attempts over 5 hours  
- Connection IDs reaching 717,000+
- Server eventually returning `database not found` errors then crashing

## Fix

Check `DoltTransaction.IsReadOnly()` before entering the dolt commit path. The `IsReadOnly()` method already exists on `DoltTransaction` and checks `tx.tCharacteristic == sql.ReadOnly`. Read-only transactions short-circuit directly to `commitWorkingSet`.

## Test plan

- [ ] Existing `TestDoltTransactionCommitOneClient` and related tests pass (write transactions still commit)
- [ ] Read-only queries with `dolt_transaction_commit=1` no longer trigger `PendingCommitAllStaged`

Fixes #10685

### DIFF
--- a/go/libraries/doltcore/sqle/dsess/session.go
+++ b/go/libraries/doltcore/sqle/dsess/session.go
@@ -478,6 +478,18 @@ func (d *DoltSession) CommitTransaction(ctx *sql.Context, tx sql.Transaction) (e
 	}
 
 	dirtyBranchState := dirties[0]
+
+	// Skip the dolt commit path for read-only transactions. When dolt_transaction_commit=1,
+	// every transaction close attempts PendingCommitAllStaged (which stages all tables and
+	// checks for changes). For read-only queries this is wasted work — there are never
+	// changes to commit. In high-frequency read workloads (e.g. monitoring agents polling
+	// every few seconds), this overhead compounds: each read-only query pays the cost of
+	// StageAllTables + newPendingCommit only to get nil back and fall through to
+	// commitWorkingSet anyway. Short-circuiting here avoids that entirely.
+	if dtx, ok := tx.(*DoltTransaction); ok && dtx.IsReadOnly() {
+		return d.commitWorkingSet(ctx, dirtyBranchState, tx)
+	}
+
 	if peformDoltCommitInt == 1 {
 		// if the dirty working set doesn't belong to the currently checked out branch, that's an error
 		err = d.validateDoltCommit(ctx, dirtyBranchState)


timsehn:
Thanks for the contribution. I've kicked off the test suite. This seems ok at first glance. 

Could probably use a test.

PR #10726, opened 2026-03-21
fix: join hint doesn't work when table names are quoted

### LINKED ISSUE #10528: Join hint doesn't work when table names are quoted
Discovered this bug when trying to repro #10527. I was using quotes for the table name because the table name had periods in it.
```
tmp/main*> explain plan SELECT /*+ LOOKUP_JOIN(t1, `t2.with.dot`)*/ t1.id, t1.name, t1.location, `t2.with.dot`.country FROM t1 JOIN `t2.with.dot` ON t1.id = `t2.with.dot`.id WHERE t1.id = 1;
+-----------------------------------------------------------------+
| plan                                                            |
+-----------------------------------------------------------------+
| Project                                                         |
|  ├─ columns: [t1.id, t1.name, t1.location, t2.with.dot.country] |
|  └─ InnerJoin                                                   |
|      ├─ (t1.id = t2.with.dot.id)                                |
|      ├─ Table                                                   |
|      │   ├─ name: t2.with.dot                                   |
|      │   └─ columns: [id country]                               |
|      └─ IndexedTableAccess(t1)                                  |
|          ├─ index: [t1.id,t1.name,t1.location]                  |
|          ├─ filters: [{[1, 1], [NULL, ∞), [NULL, ∞)}]           |
|          └─ columns: [id name location]                         |
+-----------------------------------------------------------------+
11 rows in set (0.00 sec) 
```

Join hint works when table name is not wrapped in quotes. But this means we probably don't handle table names with a db prefix (`[db name].[table name]`) correctly.
```
tmp/main*> explain plan SELECT /*+ LOOKUP_JOIN(t1, t2.with.dot)*/ t1.id, t1.name, t1.location, `t2.with.dot`.country FROM t1 JOIN `t2.with.dot` ON t1.id = `t2.with.dot`.id WHERE t1.id = 1;
+--------------------------------------------------------------------------+
| plan                                                                     |
+--------------------------------------------------------------------------+
| Project                                                                  |
|  ├─ columns: [t1.id, t1.name, t1.location, t2.with.dot.country]          |
|  └─ LookupJoin                                                           |
|      ├─ IndexedTableAccess(t1)                                           |
|      │   ├─ index: [t1.id,t1.name,t1.location]                           |
|      │   ├─ filters: [{[1, 1], [NULL, ∞), [NULL, ∞)}]                    |
|      │   └─ columns: [id name location]                                  |
|      └─ IndexedTableAccess(t2.with.dot)                                  |
|          ├─ index: [t2.with.dot.name,t2.with.dot.id,t2.with.dot.country] |
|          ├─ columns: [id country]                                        |
|          └─ keys: t1.id, t1.id, t1.id                                    |
+--------------------------------------------------------------------------+
11 rows in set (0.00 sec) 
```

majiayu000:
Fixes #10528

## Summary

Join hints like `LOOKUP_JOIN(t1, `t2`)` silently fail when table names are backtick-quoted. The `argsRegex` in `parseJoinHints()` captures backticks as part of the argument value, but the AST table name (from the parser) has backticks already stripped. The `strings.EqualFold` comparison in `SetJoinOp()` fails, so the hint is ignored.

**Fix**: Strip surrounding backticks from parsed hint arguments using `strings.Trim(arg[1], "`")` in `go-mysql-server/sql/memo/select_hints.go`.

The actual code change is in go-mysql-server: dolthub/go-mysql-server#3477

This PR points the `go-mysql-server` dependency to the fork with the fix via a `replace` directive. The replace directive should be removed and replaced with a proper version bump once the upstream PR is merged.

## Test plan

- 4 new unit tests added upstream for backtick-quoted args in `LOOKUP_JOIN`, `HASH_JOIN`, `MERGE_JOIN`, and `JOIN_ORDER`
- All 28 existing + new hint parsing tests pass

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -208,4 +208,6 @@ require (
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 )
 
+replace github.com/dolthub/go-mysql-server => github.com/majiayu000/go-mysql-server v0.20.1-0.20260321223822-d93f4fe6386e
+
 go 1.25.6
--- a/go/go.sum
+++ b/go/go.sum
@@ -212,8 +212,6 @@ github.com/dolthub/fslock v0.0.0-20251215194149-ef20baba2318 h1:n+vdH5G5Db+1qnDC
 github.com/dolthub/fslock v0.0.0-20251215194149-ef20baba2318/go.mod h1:QWql+P17oAAMLnL4HGB5tiovtDuAjdDTPbuqx7bYfa0=
 github.com/dolthub/go-icu-regex v0.0.0-20250916051405-78a38d478790 h1:zxMsH7RLiG+dlZ/y0LgJHTV26XoiSJcuWq+em6t6VVc=
 github.com/dolthub/go-icu-regex v0.0.0-20250916051405-78a38d478790/go.mod h1:F3cnm+vMRK1HaU6+rNqQrOCyR03HHhR1GWG2gnPOqaE=
-github.com/dolthub/go-mysql-server v0.20.1-0.20260319021557-3b1ba29f1e79 h1:zN/+1a2tCSoNNARJa21dH8x8/Q/p9YvqXbx3wjPzRSo=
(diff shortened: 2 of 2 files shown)


nicktobey:
Thank you for the contribution.

I'm closing this PR since the actual change is in https://github.com/dolthub/go-mysql-server/pull/3477. Once that gets merged, we can introduce the change to dolt via our normal dependency bump process.

PR #10750, opened 2026-03-26
Add incremental mode for garbage collection

nicktobey:
This feature is enabled by providing the --incremental-file-size flag to `dolt gc` or `call dolt_gc()` with a positive value, and can be enabled for automatic GC with a setting in the server's `config.yaml`, eg:

```
behavior:
  auto_gc_behavior:
    incremental_file_size: 1000000
```

In this example, after GC has written 1000000 bytes of leaf chunks (chunks that do not reference other chunks), it will finish the chunk file and begin writing a new one.

There are two main reasons to do this:
- Chunks written this way don't need to be tracked in the GC "visited set", reducing memory usage.
- If GC is interrupted (for instance, it may be terminated by the OS for using too much memory), the already-processed chunks will be preserved, and subsequent runs of GC will have less work to do and require less memory. This provides an escape hatch for the scenario where a user has not run GC in so long that GC requires more memory to run than is available.


coffeegoddd:
@nicktobey __DOLT__
| covering_index_scan  | 0.75         | 0.75       | 0.0            |
| groupby_scan         | 10.84        | 10.65      | -1.75          |
| index_join           | 1.82         | 1.86       | 2.2            |
| index_join_scan      | 1.32         | 1.34       | 1.52           |
| index_scan           | 22.28        | 21.5       | -3.5           |
| oltp_point_select    | 0.25         | 0.25       | 0.0            |
| oltp_read_only       | 5.28         | 5.28       | 0.0            |
| select_random_points | 0.5          | 0.5        | 0.0            |
| select_random_ranges | 0.77         | 0.78       | 1.3            |
| table_scan           | 21.89        | 21.5       | -1.78          |
| types_table_scan     | 68.05        | 66.84      | -1.78          |


| oltp_delete_insert    | 6.43         | 6.32       | -1.71          |
| oltp_insert           | 3.07         | 3.07       | 0.0            |
| oltp_read_write       | 11.24        | 11.45      | 1.87           |
| oltp_write_only       | 5.99         | 5.99       | 0.0            |

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b3c85e3</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b3c85e3</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
| tpcc-scale-factor-1 | 57.87            | 57.87          | 0.0            |


| tpcc-scale-factor-1 | dolt             | ffd2bb5f174758ca9dab61835703526e4ea1004a | 41.7     | dolt           | 068c41ddadf53290ace69219b85b12d60132cbe7 | 40.96  | -1.77          |

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c10df95</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c10df95</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/cmd/dolt/commands/sqlserver/server.go
```diff
@@ -280,13 +280,14 @@ func ConfigureServices(
 		InitF: func(context.Context) error {
 			// Having AutoGCBehavior == nil means that they are using default behavior, which is enabled.
 			if cfg.ServerConfig.AutoGCBehavior() == nil {
-				config.AutoGCController = sqle.NewAutoGCController(chunks.SimpleArchive, lgr)
+				config.AutoGCController = sqle.NewAutoGCController(chunks.SimpleArchive, chunks.IncrementalTablesDisabled, lgr)
```
reltuk: Maybe better if this constant had something about GC in its name.

Maybe better if it was `const` instead of a `var`

go/libraries/doltcore/doltdb/doltdb.go
```diff
@@ -2044,7 +2044,7 @@ func pullHash(
 
 func (ddb *DoltDB) getAddrs(c chunks.Chunk) chunks.GetAddrsCb {
```
reltuk: Maybe better if this typedef was updated to something like `InsertAddrsCb`

go/store/nbs/cmp_chunk_table_writer.go
```diff
@@ -171,6 +174,18 @@ func (tw *CmpChunkTableWriter) AddChunk(tc ToChunker) (uint32, error) {
...
+// SeenChunk returns whether this writer has already written a certain chunk.
+// Since chunks are stored in insertion order and not sorted until Finish() is called,
+// this is slow for large table files.
+func (tw *CmpChunkTableWriter) SeenChunk(h hash.Hash) bool {
+	for _, prefix := range tw.prefixes {
+		if prefix.addr.Equal(h) {
+			return true
+		}
+	}
+	return false
```
reltuk: Relative tradeoffs compared to slices.Sort'ing it each time and then binary search? The internet says the underlying algorithm "is specifically designed to recognize and excel on already sorted or "nearly sorted" data", so we may make it up on just a few linear scans here...

Alternatively, keeping the separate HashSet in rotation writer for this case might be a better alternative...

go/store/nbs/gc_copier.go
```diff
@@ -169,3 +172,87 @@ func fileNameToAddr(fileName string) (hash.Hash, bool) {
...
+
+	var writer GenericTableWriter
+	writer, err = newTableWriterFromArchiveLevel(gcc.archiveLevel)
+
+	gcc.writer = writer
+	gcc.bytesWritten = 0
+	return err
+}
+
+func (gcc *rotatingGCCopier) finalize(ctx context.Context) error {
```
reltuk: For structured concurrency reason, we probably want a way for SaveHashes() to block on the eg.Wait() without actually doing the final copyTablesToDir().

go/store/nbs/gc_copier.go
```diff
@@ -169,3 +172,87 @@ func fileNameToAddr(fileName string) (hash.Hash, bool) {
...
+	if gcc == nil {
+		return nil
+	}
+
+	_, err := gcc.copyTablesToDir(ctx)
```
reltuk: An weird edge case when we go two-table-files and when we do the rotate which I don't think existed before (or didn't come up as often?) – we can get to the end of SaveHashes with a table file writer which we never gave any chunks.

go/store/nbs/store.go
```diff
@@ -2319,16 +2368,30 @@ func (i *markAndSweeper) SaveHashes(ctx context.Context, hashes []hash.Hash) err
...
 				return
 			}
-			addErr = i.getAddrs(c)(ctx, nextToVisit, func(h hash.Hash) bool { return false })
+			isLeaf := true
+			addErr = i.getAddrs(c, func(h hash.Hash) error {
+				isLeaf = false
+				nextToVisit.Insert(h)
+				return nil
+			})
+
```
reltuk, line 2363: We dropped addErr here I think.

go/store/nbs/store.go
```diff
@@ -2358,11 +2421,11 @@ func (i *markAndSweeper) Finalize(ctx context.Context) (chunks.GCFinalizer, erro
...
 		specs: specs,
-		mode:  i.mode,
+		mode:  i.gcConfig.Mode,
 	}, nil
 }
 
-func (i *markAndSweeper) Close(ctx context.Context) error {
+func (i *markAndSweeper) Close(ctx context.Context) (err error) {
 	if i.gcc != nil {
 		return i.gcc.cancel(ctx)
```
reltuk: Why did we need to cancel gcc but not incrementalGcc?

go/store/types/value_store.go
```diff
@@ -542,11 +559,6 @@ func (lvs *ValueStore) Commit(ctx context.Context, current, last hash.Hash) (boo
 
 type GCMode int
 
-const (
-	GCModeDefault GCMode = iota
```
reltuk, line 546: It's a little weird to move these to a layer that knows nothing about them. Still considering...

reltuk (commented):
Looks quite good! A few comments.

go/store/nbs/cmp_chunk_table_writer.go
nicktobey: There's different approaches we could do here.

The simplest approach would be to just not support incremental GC for the old chunk file format. After all, archives are the default now.

Sorting it each time sounds acceptable, if sorting already sorted sets is fast, as you say.

Keeping a separate HashSet in the rotation writer works, at the cost of more memory, and needlessly so for archive files, since it's literally a duplicate of the hash set in the archive file writer.

go/store/nbs/store.go
nicktobey: The incrementalGcc does not outlive the call to SaveHashes, so there's nothing to cancel.

go/store/types/value_store.go
nicktobey: These were duplicates, defined in two different packages, one of which depended on the other. So I removed the redundant implementation.

go/store/nbs/gc_copier.go
nicktobey: Is it a problem to write a table file with 0 chunks? I can add a check for it but I imagine that in the rare case of  0 chunk table file will still get correct behavior.

reltuk: Having never seen it, I wouldn't necessarily expect it to work. But it's possible it does! In other places where we do similar things, like `PullTableFileWriter`, we only construct the writer after we have the first chunk to put into it.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c4bb798</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c4bb798</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5e83e94</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5e83e94</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/gc_copier.go
```diff
@@ -142,3 +147,144 @@ func fileNameToAddr(fileName string) (hash.Hash, bool) {
...
 	return hash.Hash{}, false
 }
+
+// newlyWrittenSources is a set of newly written chunk sources.
+// It can be written to by multiple goroutines concurrently,
+// so access is controlled by a mutex
+type newlyWrittenSources struct {
+	// specs stores a list of table files produced by this copier.
+	// specNames stores just their content hashes.
+	// Since these are appended in a goroutine, access is controlled by the mutex specsMu
```
reltuk: Stale.

go/store/types/value_store.go
```diff
@@ -542,11 +559,6 @@ func (lvs *ValueStore) Commit(ctx context.Context, current, last hash.Hash) (boo
 
 type GCMode int
```
reltuk: Nit: stale.

go/store/nbs/store.go
```diff
@@ -2220,21 +2220,50 @@ func markAndSweepChunks(_ context.Context, nbs *NomsBlockStore, src CompressedCh
...
-	gcc  *gcCopier
-	mode chunks.GCMode
+	tfp      tableFilePersister
+	gcc      *gcCopier
+	gcConfig chunks.GCConfig
+
+	specs []tableSpec
+}
+
+func getHashSubset(hashes hash.HashSet, subsetSize int) (subset hash.HashSet, remaining hash.HashSet) {
```
reltuk: Nit: is this used?

go/store/nbs/store.go
```diff
@@ -2314,11 +2380,14 @@ func (i *markAndSweeper) SaveHashes(ctx context.Context, hashes []hash.Hash) err
...
-
 		toVisit = nextToVisit
+
 	}
-	return nil
+
+	// we don't need to acquire the mutex on the specs because all other tasks that reference it have completed.
+	i.specs = append(i.specs, incrementalGcc.specs.specs...)
+
+	return incrementalGcc.finalize(ctx)
```
reltuk: What's happening here? Why could we copy the incrementalGcc.specs.specs... into i.specs before the finalize call? Doesn't `finalize` itself generate a spec?

go/store/nbs/gc_copier.go
```diff
@@ -142,3 +147,144 @@ func fileNameToAddr(fileName string) (hash.Hash, bool) {
...
+			return err
+		}
+
+		return gcc.specs.append(ctx, specs, gcc.dest)
+	})
+
+	writer, err := newTableWriterFromArchiveLevel(gcc.archiveLevel)
+
+	gcc.gcCopier.writer = writer
+	gcc.bytesWritten = 0
```
reltuk: Did we test this? 0 chunk archive and 0 chunk table file are happy?

go/store/nbs/store.go
```diff
@@ -2314,11 +2380,14 @@ func (i *markAndSweeper) SaveHashes(ctx context.Context, hashes []hash.Hash) err
...
 
-		i.visited.InsertAll(toVisit)
-
 		toVisit = nextToVisit
+
 	}
-	return nil
+
+	// we don't need to acquire the mutex on the specs because all other tasks that reference it have completed.
+	i.specs = append(i.specs, incrementalGcc.specs.specs...)
```
reltuk: This doesn't look like everything we want to transfer here. We want the open chunkSources themselves to make it to the Finalize call, so they can be offered in the `openChunkSourcesForManifestUpdateAndRebase` call as existing sources.

go/store/nbs/store.go
```diff
@@ -2314,11 +2380,14 @@ func (i *markAndSweeper) SaveHashes(ctx context.Context, hashes []hash.Hash) err
 			return fmt.Errorf("dangling references requested during GC. GC not successful. %v", toVisit)
 		}
 
-		i.visited.InsertAll(toVisit)
-
 		toVisit = nextToVisit
+
 	}
-	return nil
+
```
reltuk, line 2362: I think on the error path here we need to make sure to close existing chunkSources on incrementalGcc.

reltuk (commented):
Generally looks good. Some comments.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8b79978</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8b79978</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/store.go
nicktobey: Removed

go/store/nbs/store.go
nicktobey: You're right, we should be finalizing the incremental copies before getting the specs out of it.

Curiously, this probably should be breaking the tests that I wrote today, but it isn't. I need to understand why.

go/store/nbs/gc_copier.go
nicktobey: 0 chunk files are never returned by the the finalize method. This is handled in the gcCopier. This is also covered by the added tests: one has an incremental table file size threshold of 1 byte, which guarentees that there will be an empty pending table when we finalize.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>ef608ea</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>ef608ea</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>90521e8</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>90521e8</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/cmp_chunk_table_writer.go
nicktobey: Even in the most optimal case, verifying that the data is already sorted is linear time and requires a linear scan. I don't think that's any faster than the current implementation I added to `cmp_chunk_table_writer.go`. If the user isn't using archives, the copier need to track the chunk set itself.

If the user is using archives, we extract a pointer to the archive writer's chunk set and use that. It breaks encapsulation, but given that the whole point of this mode is reducing memory by not inserting leaf chunk hashes into the visited set, it makes sense to not duplicate this set in memory.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>f944dbf</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>f944dbf</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/store.go
nicktobey: I now persist the incremental gc writer until we call markAndSweeper.Finalize, so it can use the incremental writer's source set.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>6333664</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>6333664</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>94459ca</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>94459ca</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8626345</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8626345</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/nbs/store.go
```diff
@@ -2349,40 +2345,50 @@ func (i *markAndSweeper) SaveHashes(ctx context.Context, hashes []hash.Hash) err
...
+		mode:  i.gcConfig.Mode,
 		srcs:  result.sources,
 	}, nil
 }
 
-func (i *markAndSweeper) Close(ctx context.Context) error {
+func (i *markAndSweeper) Close(ctx context.Context) (err error) {
 	if i.gcc != nil {
 		return i.gcc.cancel(ctx)
 	}
```
reltuk, line 2394: Cleanup here is still not what I would expect? What am I missing?

incrementalGcc potentially has a writer open, which needs to be canceled.

incrementalGcc has a chunkSourceSet which needs to be closed.

It seems like this path should have cleanup for i.incrementalGcc. It seems like Finalize should be closing the chunkSourceSet and setting incrementalGcc to nil on the success path.

reltuk (commented):
One comment about cleanup. Generally looks good!

go/store/nbs/store.go
nicktobey: You're right.

Previously, incrementalGcc was a local variable in SaveHashes that never outlived the call, so it wasn't alive by the time Close was called. But that's not the case anymore, and so it needs to be closed when the markAndSweeper is.

Fixed.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>38bcc80</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>38bcc80</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>4a073c4</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>4a073c4</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>de38946</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>de38946</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

reltuk (approved):
If you review and merge #10947, I think this is good to go.

coffeegoddd:
@nicktobey __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>068c41d</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>068c41d</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

reltuk (approved):
LGTM!

PR #10786, opened 2026-03-31
Add support for manually resolving conflicts with `dolt_revert()` and `dolt revert`

fulghum:
Previously any conflict or constraint violation would cause `dolt_revert()` to error out. With this change, it returns information on the conflicts or constraint violations so that customers can manually resolve them and then continue the revert action. 

Note: in order to return information about conflicts and constraint violations, the return schema of `dolt_revert()` has changed. It now matches the return schema for `dolt_cherry_pick()` and `dolt_merge()`. Revert's behavior also changes to create one revert commit per input commit (instead of a single revert commit) to match Git's behavior. Because of that, this change should be released as a minor version bump.

### DIFF
--- a/go/cmd/dolt/commands/revert.go
+++ b/go/cmd/dolt/commands/revert.go
@@ -141,6 +141,18 @@ func revert(ctx context.Context, apr *argparser.ArgParseResults, cliCtx cli.CliC
 		buffer.WriteString(")")
 	}
 
+	_, err = cli.GetRowsForSql(queryist.Queryist, queryist.Context, "set @@dolt_allow_commit_conflicts = 1")
+	if err != nil {
+		cli.Println(fmt.Errorf("error: failed to set @@dolt_allow_commit_conflicts: %w", err))
+		return 1
+	}
+
+	_, err = cli.GetRowsForSql(queryist.Queryist, queryist.Context, "set @@dolt_force_transaction_commit = 1")
+	if err != nil {
+		cli.Println(fmt.Errorf("error: failed to set @@dolt_force_transaction_commit: %w", err))
+		return 1
+	}
+
 	query, err := dbr.InterpolateForDialect(buffer.String(), params, dialect.MySQL)
 	if err != nil {
 		cli.Println(err.Error())
--- a/go/libraries/doltcore/sqle/enginetest/dolt_queries_revert.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_queries_revert.go
@@ -108,8 +108,45 @@ var RevertScripts = []queries.ScriptTest{
 		},
 	},
 	{
+		// With @@autocommit=1 (the default) and no allow vars set, dolt_revert() returns
+		// an error when it encounters conflicts. The revert is not persisted to disk —
+		// the engine's rollback leaves the repository in its original clean state.
+		// To use --abort or --continue across dolt sql -q invocations, users must set
+		// @@dolt_allow_commit_conflicts=1 or @@autocommit=0 so the merge state isn't rolled back.
+		Name: "dolt_revert() with @@autocommit=1 returns error on conflicts and leaves repo clean",
+		SetUpScript: []string{
+			"create table test (pk int primary key, c0 int)",
+			"insert into test values (1,1),(2,2),(3,3);",
+			"call dolt_commit('-Am', 'seed table');",
+			"update test set c0 = 42 where pk = 2;",
+			"call dolt_commit('-am', 'first change');",
+			"update test set c0 = 23 where pk = 2;",
+			"call dolt_commit('-am', 'second change');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				// Reverting HEAD~1 creates a conflict: both ours (HEAD) and theirs
+				// (HEAD~2, parent of the reverted commit) modified pk=2.
+				// With @@autocommit=1 and no allow vars, this returns an error.
+				Query:          "call dolt_revert('HEAD~1');",
+				ExpectedErrStr: "Merge conflict detected, @autocommit transaction rolled back. @autocommit must be disabled so that merge conflicts can be resolved using the dolt_conflicts and dolt_schema_conflicts tables before manually committing the transaction. Alternatively, to commit transactions with merge conflicts, set @@dolt_allow_commit_conflicts = 1",
+			},
+			{
+				// The rollback leaves the repository clean — no conflicts on disk.
+				Query:    "select `table` from dolt_conflicts;",
+				Expected: []sql.Row{},
+			},
+			{
+				// No new commit should have been created.
+				Query:    "select message from dolt_log limit 1;",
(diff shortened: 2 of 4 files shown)


coffeegoddd:
@fulghum __DOLT__
| covering_index_scan  | 0.54         | 0.55       | 1.85           |
| groupby_scan         | 9.91         | 9.91       | 0.0            |
| index_join           | 1.86         | 1.86       | 0.0            |
| index_join_scan      | 1.34         | 1.37       | 2.24           |
| index_scan           | 22.28        | 22.28      | 0.0            |
| oltp_point_select    | 0.27         | 0.27       | 0.0            |
| oltp_read_only       | 5.28         | 5.28       | 0.0            |
| select_random_points | 0.53         | 0.54       | 1.89           |
| select_random_ranges | 0.55         | 0.56       | 1.82           |
| table_scan           | 22.28        | 22.28      | 0.0            |
| types_table_scan     | 68.05        | 66.84      | -1.78          |


| oltp_delete_insert    | 6.32         | 6.32       | 0.0            |
| oltp_insert           | 3.13         | 3.13       | 0.0            |
| oltp_read_write       | 11.24        | 11.45      | 1.87           |
| oltp_write_only       | 5.99         | 5.99       | 0.0            |
| types_delete_insert   | 6.91         | 6.79       | -1.74          |

coffeegoddd:
@fulghum __DOLT__
| tpcc-scale-factor-1 | 54.83            | 54.83          | 0.0            |


| tpcc-scale-factor-1 | dolt             | 7d1ba2b26d2f2a388b0b627ce6a9ef10c2daaf02 | 43.98    | dolt           | 53275584998e0a35e2d4fcd1ef2ea710ea6a410f | 43.63  | -0.8           |

integration-tests/bats/helper/local-remote.bash
```diff
@@ -68,6 +68,7 @@ SKIP_SERVER_TESTS=$(cat <<-EOM
 ~arg-parsing.bats~
 ~dump.bats~
 ~rename-tables.bats~
+~revert.bats~
```
macneale4: Why are you removing the remote-server testing? `dolt revert` should definitely work in that scenario.

go/libraries/doltcore/sqle/enginetest/dolt_queries_revert.go
```diff
@@ -71,20 +72,390 @@ var RevertScripts = []queries.ScriptTest{
...
+			"call dolt_commit('-am', 'change row 2');",
+			// commit C - modifies same row as B's revert, causing conflict
+			"update test set c0 = 99 where pk = 2;",
+			"call dolt_commit('-am', 'change row 2 again');",
+		},
+		Assertions: []queries.ScriptTestAssertion{
+			{
+				// Reverting HEAD~2 (insert row 4) first, then HEAD~1 (change row 2).
+				// HEAD~2 reverts cleanly (commit created). HEAD~1 conflicts.
+				Query:    "call dolt_revert('HEAD~2', 'HEAD~1');",
```
macneale4, line 496: interesting. didn't know we supported this. Does the workflow work is there are conflicts in each? A test where we revert three commits, the first two have conflicts, and the last one applies cleanly would be nice. I'd expect to see two --continue calls, to ensure we can get out of sticky situations.

macneale4: Also, does --abort get you back to the state at the beginning of the entire workflow, or will it drop you back on the most recently successful revert? I'm not actually sure what git does in this situation. Can you verify, and write a test to ensure we are in sync?

integration-tests/bats/helper/local-remote.bash
fulghum: I was struggling with getting conflict resolution to work with the remote-engine and noticed that we also exclude merge and cherry-pick from the remote server tests and thought perhaps that was why those others were excluded. Looking closer at that exclusion list... it looks like we exclude about 75% of **all** BATS tests from the remote-engine testing currently, plus some other skips in the remaining tests. 😬 Probably worth digging into that to see why we have so many tests we've excluded there. 

I can poke around some more with conflict resolution today and see if I can figure it out. It seems like it should work. At minimum, I could skip individual BATS tests in remote-engine mode and leave the others to keep running.

go/libraries/doltcore/sqle/enginetest/dolt_queries_revert.go
fulghum: Great question about Git's behavior with `--abort`. Looks like Git will abort the entire operation, including any commits that were successfully reverted before the conflict. I'm updating the logic in this PR to match that now.

integration-tests/bats/helper/local-remote.bash
fulghum: I dug into this and found a `panic` when the remote-engine tests were running. I chased this down a bit, but it seemed to be a red herring – all remote-engine tests, successful or failing, seem to exhibit this panic as part of the test tear down, we just don't see it reported for passing tests. We should track that down, but it's not related to these changes. 

After that, I dug into the CLI code and found where it wasn't interpreting the query results correctly when they came back in remote-engine mode, due to type differences. That's fixed and remote engine tests are all passing now.

go/cmd/dolt/commands/revert.go
```diff
@@ -82,20 +82,12 @@ func (cmd RevertCmd) Exec(ctx context.Context, commandStr string, args []string,
...
 	if !cli.CheckUserNameAndEmail(cliCtx.Config()) {
 		return 1
 	}
 
-	if apr.Contains(cli.ContinueFlag) {
-		return revert(ctx, apr, cliCtx)
-	}
-
-	if apr.NArg() < 1 {
+	if apr.NArg() < 1 && !(apr.Contains(cli.ContinueFlag) || apr.Contains(cli.AbortParam)) {
```
macneale4: I'd write that a apr.NArg() == 0

It looks like nowhere in this code do we error out if you present both flags. Other commands give an error when both --abort and --continue are given.

I would also expect a usage error when `apr.Narg() > 0 && (apr.Contains(cli.ContinueFlag) || apr.Contains(cli.AbortParam))` because both of those flags only make sense where there are no arguments.

fulghum: Changed to `apr.NArg() == 0`. 

There is a check for when both `--abort` and `--continue` are specified and we test that error case in the Go tests.

coffeegoddd:
@fulghum __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>5327558</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>5327558</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10804, opened 2026-04-03
dolthub/dolt#10813 fix(checkout): enforce --no-overwrite-ignore in DOLT_CHECKOUT -b via direct SQL

codeaucafe:
## Summary

When calling `DOLT_CHECKOUT('-b', ...)` directly via SQL (without `--move`), the `--no-overwrite-ignore` flag was silently ignored. `CheckOverwrittenIgnoredTables` was never called for the `isMove=false` path in `checkoutNewBranch`, so creating a branch from a non-HEAD start point would succeed even when ignored tables in the working set differed from the start point.

The fix captures the current working roots before `commitTransaction` (which resets session state, making `GetRoots` return false afterward), then runs the ignored table check before switching the working set to the new branch.

## Changes

- Call `CheckOverwrittenIgnoredTables` in the `isMove=false` path of `checkoutNewBranch`
- Capture roots before `commitTransaction` to avoid session state reset
- Add engine script tests and bats tests covering `--no-overwrite-ignore` and
`--overwrite-ignore` with a non-HEAD start point

Closes: #10813

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_checkout.go
@@ -466,9 +466,17 @@ func checkoutNewBranch(
 		}
 	}
 
+	// Capture current roots before commitTransaction, which starts a new transaction and
+	// resets session state, causing GetRoots to return false afterwards.
+	sess := dsess.DSessFromSess(ctx.Session)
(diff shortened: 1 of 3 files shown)


nicktobey (approved):

codeaucafe:
Thank you @nicktobey

PR #10813, opened 2026-04-04
go/cmd/dolt: Add GOMEMLIMIT awareness via automemlimit and server config

stebbins:
Dolt has no GOMEMLIMIT or GOGC awareness. For containerized deployments, this means the Go runtime doesn't know about cgroup memory limits, leading to OOM kills when RSS grows past the container limit.

This adds three-tier memory limit configuration:

1. `DOLT_GOMEMLIMIT` env var — explicit, human-readable values like "1GiB"
2. `automemlimit` cgroup auto-detection — sets GOMEMLIMIT to 90% of the cgroup limit
3. Silent fallthrough on bare metal when neither applies

Also adds `max_go_memory` to the YAML server config, applied at sql-server startup via `debug.SetMemoryLimit()`, and `MaxGoMemory() int64` to the `ServerConfig` interface.

Files changed:
- `go/cmd/dolt/memlimit.go` (new) — `configureMemoryLimit()` with the three-tier logic
- `go/cmd/dolt/dolt.go` — calls `configureMemoryLimit()` early in `runMain()`
- `go/cmd/dolt/commands/sqlserver/server.go` — applies `max_go_memory` at sql-server startup
- `go/libraries/doltcore/servercfg/` — `MaxGoMemory() int64` on interface and implementations

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/command_line_config.go
+++ b/go/cmd/dolt/commands/sqlserver/command_line_config.go
@@ -604,6 +604,10 @@ func (cfg *commandLineServerConfig) AutoGCBehavior() servercfg.AutoGCBehavior {
 	return stubAutoGCBehavior{}
 }
 
+func (cfg *commandLineServerConfig) MaxGoMemory() int64 {
+	return 0
+}
+
 func (cfg *commandLineServerConfig) Overrides() sql.EngineOverrides {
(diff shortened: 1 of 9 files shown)


stebbins:

stebbins:

PR #10814, opened 2026-04-04
go/libraries/doltcore/servercfg: Add pprof_server to YAML server config

stebbins:
The `--pprof-server` CLI flag exists but there is no way to enable it via YAML config. This adds `pprof_server` as a boolean field in the server config, following the same pattern as existing bool fields like `read_only`.

- `PprofServer() bool` added to the `ServerConfig` interface
- YAML implementation: struct field, getter, `ValueSet`, serialization helpers, placeholder default
- `command_line_config` returns `false` (CLI has its own flag path via `--pprof-server`)
- `server.go` wires `cfg.ServerConfig.PprofServer()` into `ConfigureServices` to start pprof
- `testdata/minver_validation.txt` updated with the new field

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/command_line_config.go
+++ b/go/cmd/dolt/commands/sqlserver/command_line_config.go
@@ -600,6 +600,10 @@ func (cfg *commandLineServerConfig) ValueSet(value string) bool {
 	return ok
 }
 
+func (cfg *commandLineServerConfig) PprofServer() bool {
+	return false
+}
+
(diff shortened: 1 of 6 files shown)


reltuk (commented):
pprof-server is generally a developer tool and is not currently designed for running on anything like a production server. The inability to specify a listener and the lack of auth makes this very footgun-y.

What's the motivation for specifying that this should be running on config.yaml, versus an expert user who needed this just changing the command line arguments to the dolt process itself?

stebbins:
@reltuk you are absolutely right :+1: i'll close. that works for my use-case too.

PR #10815, opened 2026-04-04
go/store/prolly/tree: Add DOLT_NODE_CACHE_SIZE env var for prolly node cache

stebbins:
The prolly tree node cache is hardcoded at 256 MiB and is the dominant memory consumer in Dolt. This makes it configurable via the `DOLT_NODE_CACHE_SIZE` environment variable, following the exact same pattern as the existing `DOLT_COMMIT_CACHE_SIZE` env var.

For memory-constrained deployments (containers, small VMs), setting `DOLT_NODE_CACHE_SIZE=33554432` (32 MiB) significantly reduces RSS with minimal performance impact for small datasets.

No default behavior change — without the env var, cache remains 256 MiB.

- `go/store/prolly/tree/node_store.go` — `const cacheSize` changed to `var`, `init()` reads env var
- `go/libraries/doltcore/dconfig/envvars.go` — `EnvNodeCacheSize` constant
- `go/store/prolly/tree/node_cache_test.go` — tests for valid, invalid, zero, and negative values

### DIFF
--- a/go/libraries/doltcore/dconfig/envvars.go
+++ b/go/libraries/doltcore/dconfig/envvars.go
@@ -46,6 +46,7 @@ const (
 	EnvDoltRootHost                  = "DOLT_ROOT_HOST"
 	EnvDoltRootPassword              = "DOLT_ROOT_PASSWORD"
 	EnvDoltGCScheduler               = "DOLT_GC_SCHEDULER"
+	EnvNodeCacheSize                 = "DOLT_NODE_CACHE_SIZE"
 
 	// If set, must be "kill_connections" or "session_aware"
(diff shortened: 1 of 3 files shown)


stebbins:
@reltuk thanks for taking a look! I think I'll close these env var PRs in favor of this approach based on your feedback: https://github.com/dolthub/dolt/pull/10820

what do you think? it should scale gracefully based on GOMEMLIMIT, based on the ratios that were already implemented. 

additional configuration probably makes sense for followup, but maybe it makes sense to just give a tuner where you can set the ratio of reserved memory  to favor node cache for reads and memtable for writes?

stebbins:

stebbins:

PR #10816, opened 2026-04-04
go/libraries/doltcore/dbfactory: Add DOLT_MEMTABLE_SIZE env var for write buffer sizing

stebbins:
The NBS memtable write buffer defaults to 256 MiB in dbfactory and is the second-largest fixed memory allocation in Dolt. This makes it configurable via the `DOLT_MEMTABLE_SIZE` environment variable.

For memory-constrained deployments, tuning this down reduces RSS. For write-heavy workloads on machines with ample RAM, tuning it up can reduce flush frequency.

No default behavior change — without the env var, memtable remains 256 MiB.

- `go/libraries/doltcore/dbfactory/factory.go` — `const defaultMemTableSize` changed to `var`, `init()` reads env var
- `go/libraries/doltcore/dconfig/envvars.go` — `EnvMemTableSize` constant

### DIFF
--- a/go/libraries/doltcore/dbfactory/factory.go
+++ b/go/libraries/doltcore/dbfactory/factory.go
@@ -18,8 +18,11 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


stebbins:

stebbins:

PR #10820, opened 2026-04-07
go: derive internal cache sizes from GOMEMLIMIT

stebbins:
## Summary

- Solves https://github.com/dolthub/dolt/issues/10821
- Adds a `memlimit` package that reads standard Go `GOMEMLIMIT` and proportionally sizes Dolt's three largest caches (node cache, memtable, decoded chunks)
- When `GOMEMLIMIT` is unset, all defaults preserved exactly — zero behavior change
- Replaces the per-cache env var approach from #10815/#10816 with a single standard knob

## Usage

```bash
GOMEMLIMIT=512MiB dolt sql-server
```

## Design

| Cache | Default | At GOMEMLIMIT=512MiB |
|-------|---------|---------------------|
| Prolly node cache | 256 MiB | ~120 MiB |
| Memtable write buffer | 256 MiB | ~120 MiB |
| Decoded chunks cache | 32 MiB | ~15 MiB |

Budget partitioning: 50% usable (50% GC reserve), then 47/47/6 across caches — matching the ratio of the original hardcoded defaults (256:256:32 MiB). Minimums prevent pathologically small caches. Maximum caps ensure caches never exceed current defaults, even with high GOMEMLIMIT — we can loosen these later if needed.

## Changes

- **New:** `go/libraries/doltcore/memlimit/` — budget calculator with 8 unit tests
- **Modified:** `go/store/prolly/tree/node_store.go` — lazy `sync.Once` init via memlimit
- **Modified:** `go/store/types/value_store.go` — decoded chunks from memlimit
- **Modified:** `go/libraries/doltcore/dbfactory/` — all 8 factory impls call `memlimit.MemtableSize()` directly
- **Modified:** `go/cmd/dolt/dolt.go` — `memlimit.Init()` in `runMain()`

## What this doesn't do

- No custom `DOLT_*` env vars
- No YAML config additions
- No `automemlimit` dep yet (follow-up for container cgroup auto-detection)
- Doesn't touch commit cache (already has `DOLT_COMMIT_CACHE_SIZE`)

## Test plan

- [x] 8 unit tests: defaults, zero, 512 MiB, 128 MiB, 32 MiB (minimums), 4 GiB (capped at defaults), accessors
- [x] `store/prolly/tree`, `store/types`, `libraries/doltcore/dbfactory`, `store/prolly` test suites pass
- [ ] Full binary build + live RSS measurement (needs CI with ICU headers)

### DIFF
--- a/go/libraries/doltcore/memlimit/memlimit.go
+++ b/go/libraries/doltcore/memlimit/memlimit.go
@@ -14,6 +14,10 @@ const (
 	minNodeCacheSize     = 16 * 1024 * 1024 // 16 MiB
 	minMemtableSize      = 4 * 1024 * 1024  // 4 MiB
 	minDecodedChunksSize = 4 * 1024 * 1024  // 4 MiB
+
+	maxNodeCacheSize     = DefaultNodeCacheSize
+	maxMemtableSize      = DefaultMemtableSize
+	maxDecodedChunksSize = DefaultDecodedChunksSize
 )
 
 type Budget struct {
@@ -41,9 +45,9 @@ func compute() Budget {
 		return defaults()
 	}
 
-	// Partition 75% of the limit across caches, reserving the rest for
+	// Partition 50% of the limit across caches, reserving the rest for
 	// GC/stacks. Ratios mirror the original defaults (256:256:32 MiB).
-	usable := float64(limit) * 0.75
+	usable := float64(limit) * 0.50
 
 	b := Budget{
 		NodeCache:     uint64(usable * 0.47),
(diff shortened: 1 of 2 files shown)


go/libraries/doltcore/memlimit/memlimit.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if b.NodeCache < minNodeCacheSize {
+		b.NodeCache = minNodeCacheSize
+	}
+	if b.Memtable < minMemtableSize {
+		b.Memtable = minMemtableSize
+	}
+	if b.DecodedChunks < minDecodedChunksSize {
+		b.DecodedChunks = minDecodedChunksSize
+	}
```
reltuk, line 72: I like this approach, but I would be conservative here to not make these too big even if GOMEMLIMIT is set. Maybe `usable := float64(limit) * 0.5` instead, and maybe add something like:

```go
	maxNodeCacheSize     = DefaultNodeCacheSize
	maxMemtableSize      = DefaultMemtableSize
	maxDecodedChunksSize = DefaultDecodedChunkSize
```

with coincident bounds checks here.

at least for now? We can always roll out a setting for max... > Default... later if we need to, but the reality is that no existing Dolt user is running up against these caches being too small with default settings for now.

stebbins: sorry for the late response. sounds good to me. it does make me curious to see a graph of performance vs available memory although i don't have a convenient way to test with an interesting amount of memory right now

reltuk:
Awesome! Thank you for the update. Everything looked good. I've merged this to main :).

PR #10829, opened 2026-04-08
[no-release-notes] integration-tests/bats: Cleanup some background process lifecycle and teardown ordering warts.

reltuk:
Every now-and-then a bats test would fail with a panic from a kill'd-and-wait'd background process whose dolt database directory had been deleted out from under it before it was asked to shutdown. This is a cleanup where we make sure to always kill and wait background children we spawn before the test ends, and we do it before teardown_common / rm -rf test dirs happens.


zachmu (approved):
LGTM!

PR #10883, opened 2026-04-15
[no-release-notes] integration-tests/go-sql-server-driver: Cleanup some illegal require usage from within goroutines.

reltuk:


fulghum (approved):
Looks good!

PR #10810, opened 2026-04-03
Fix remotesapi push fail with "target has uncommitted changes" after any successful push

elianddb:
- Fixed regression where pushing to a SQL server left the server's staged and working state pointing at the previous commit instead of the new one.
- Pushes to servers with ignored tables correctly advances all three branch pointers together.

Fix #10807

PR #10946, opened 2026-04-25
sqle: repair stale working set after remotesapi push from older clients

### LINKED ISSUE #10945: RemotesAPI push: working-set-stale bug from #10807 still reproducible in v1.86.5
## Summary

[#10807](https://github.com/dolthub/dolt/issues/10807) was closed by [#10810](https://github.com/dolthub/dolt/pull/10810) ("Fix remotesapi push fail with 'target has uncommitted changes' after any successful push"). v1.86.5 is 231 commits ahead of that merge (0 behind), and the source change is visibly present in `actions/remotes.go` and `doltdb/doltdb.go` at this tag.

However, the same bug still reproduces in v1.86.5: a successful push to a remotesapi server leaves the server's working set pointing at the previous HEAD, and the next push fails with:

```
target has uncommitted changes. --force required to overwrite
```

Reproduces both via [beads](https://github.com/steveyegge/beads) (which uses Dolt as an embedded Go library) and via pure `dolt push` from the CLI — so the bug is on the dolt server side, not in the beads wrapper.

## Environment

- **Server**: `dolthub/dolt-sql-server:1.86.5` Docker image, started with `--remotesapi-port=7007` and no other flags
- **Clients tried**:
  - beads v1.0.3 in embedded mode (Dolt Go library push via remotesapi)
  - `dolt` CLI v1.86.5 (`dolt clone http://server:7007/db` then `dolt push origin main`)

Both clients exhibit identical behavior.

## Reproduction (via beads)

Server starts in a clean state (verified: `SELECT * FROM mydb.dolt_status` returns no rows).

1. From a beads project, create an issue (writes a commit locally):
   ```
   bd create "test"
   ```

2. Push to the remotesapi server:
   ```
   bd dolt push
   ```
   → succeeds: `Push complete.`

3. **Immediately** check the server's status:
   ```sql
   SELECT * FROM `mydb`.dolt_status;
   ```
   ```
   +------------+--------+----------+
   | table_name | staged | status   |
   +------------+--------+----------+
   | events     | 1      | modified |
   | issues     | 1      | modified |
   +------------+--------+----------+
   ```
   The working set diverged from HEAD even though no SQL client wrote to the server.

4. Confirm the divergence at the row level:
   ```sql
   SELECT id, status FROM `mydb`.issues;                  -- working set: previous state
   SELECT id, status FROM `mydb`.issues AS OF 'main';     -- HEAD:        post-push state
   ```
   For example, an issue we just closed appears as `closed` at HEAD but `open` in the working set.

5. Make another local change and push again:
   ```
   bd close test-id
   bd dolt push
   ```
   →
   ```
   Error: push to origin/main: Error 1105: target has uncommitted changes. --force required to overwrite
   ```

6. **Workaround** — hard-reset the server's working set:
   ```sql
   USE `mydb`; CALL DOLT_RESET('--hard');
   ```
   Subsequent push then succeeds — until the next push, when the cycle repeats.

## Reproduction (pure CLI, no beads)

To rule out a beads-side issue, the same bug reproduces with the dolt CLI alone, against the already-stale server:

```bash
dolt clone http://server:7007/mydb
cd mydb
dolt sql -q "INSERT INTO labels (issue_id, label) VALUES ('some-id', 'cli-test');"
dolt add labels
dolt commit -m "test: pure dolt CLI write"
dolt push origin main
# → target has uncommitted changes. --force required to overwrite
```

## Source observation

After PR #10810, `FastForwardWithWorkspaceCheck` in `go/libraries/doltcore/doltdb/doltdb.go` only includes a non-empty working-set path under a conditional:

```go
ws := ""
pushConcurrencyControl := chunks.GetPushConcurrencyControl(datas.ChunkStoreFromDatabase(ddb.db))
if pushConcurrencyControl == chunks.PushConcurrencyControl_AssertWorkingSet {
    wsRef, err := ref.WorkingSetRefForHead(branch)
    if err != nil {
        return err
    }
    ws = wsRef.String()
}
_, err = ddb.db.FastForward(ctx, ds, addr, ws, allowDirtyWorking)
```

`dolt sql-server` does set `ConcurrencyControl: PUSH_CONCURRENCY_CONTROL_ASSERT_WORKING_SET` in `cmd/dolt/commands/sqlserver/server.go`, so in principle the conditional should pass and `ws` should be populated. Empirically the working set is not getting updated, which suggests either:

- The remote chunk store isn't surfacing `AssertWorkingSet` to the client at runtime (something between the server's metadata advertisement and the client's `GetPushConcurrencyControl` call breaks), or
- A different push code path is being exercised that bypasses this update entirely.

## Asks

1. Is this a regression of #10807 via a different code path, or an incomplete fix?
2. Is there a way to verify what `PushConcurrencyControl` value the client sees from a given `dolt sql-server`?
3. Until this is resolved, is `CALL DOLT_RESET('--hard')` on the server after each push the recommended workaround for remotesapi-only deployments?

Happy to bisect, instrument, or run additional experiments.

alingenhag:
## Summary

Fixes #10945. When an older Dolt client (pre PR #10810 — e.g. beads v1.0.3, which pins `dolt@005921bdd8ca`, 166 commits before the fix) pushes to a `dolt-sql-server` over remotesapi, it advances only the branch HEAD ref and leaves the working set ref pointing at the pre-push `WorkingSet`. The server's `staged` root no longer matches `HEAD`, and subsequent pushes from any client — including up-to-date ones — fail with `target has uncommitted changes. --force required to overwrite`. The only recovery today is `CALL DOLT_RESET('--hard')` on the server.

The root cause on the old-client side: `WorkingSetContainsOnlyIgnoredTables` returns `true` for an empty diff (the remote's working set is clean, which is the normal case), so the old `Push()` takes the `FastForward` branch that does not include a working set update. PR #10810 fixed this client-side by always going through `FastForwardWithWorkspaceCheck`, but clients on older versions remain affected. This change makes the server tolerant of those clients.

## What changes

In `hooksFiringRemoteSrvStore.Commit` (the layer the `dolt-sql-server` wraps every remotesapi commit with), after the underlying chunk store commit succeeds, walk the post-commit dataset map and, for each branch whose HEAD advanced but whose corresponding working set ref was unchanged in the same commit, advance the working set so its `working` and `staged` roots both match the new HEAD's root. Implemented via `DoltDB.UpdateWorkingSet` with the existing CAS.

The repair is a no-op when:
- The client also updated the working set ref (modern client) — detected via `oldAddrs[wsID] != newAddrs[wsID]`
- No working set exists at the ref (file:// remotes, freshly created branches)
- The working set is already in sync with HEAD
- The CAS loses a race with another writer (best-effort; the next push will retry)

The repair runs **before** firing commit hooks so hooks observe the consistent post-repair state.

Real dirty workspaces from concurrent SQL writes still reject pushes — the dirty check fires client-side in `doFastForward` before the Commit RPC ever reaches the server, so the repair never runs in that path and the AssertWorkingSet contract is preserved.

## Test plan

- [x] New `TestRepairStaleWorkingSetAfterPush` constructs the stale state via `ddb.Commit` (which advances only HEAD, mirroring old-client behavior), replays the commit through `hooksFiringRemoteSrvStore.Commit`, and asserts the working set's `working`/`staged` roots end up at the new HEAD's root.
- [x] New `TestRepairWorkingSetSkippedWhenClientUpdated` verifies the repair leaves the working set untouched when the client already updated it (regression guard for modern clients).
- [x] Existing `TestHooks*` and `TestOnlyChangedDatasetsFireHooks` / `TestReplicaWriteFiltersHooks` still pass.
- [x] `./libraries/doltcore/sqle`, `./libraries/doltcore/env/actions`, `./libraries/doltcore/doltdb`, `./store/datas` test suites pass.
- [x] End-to-end manual repro against a patched `dolt-sql-server`:
  - Built `dolt` at `005921bdd8ca` (beads' pinned commit) as the "old" client.
  - Old-client push: succeeds, server `dolt_status` is empty (was: showed `staged | modified` for every changed table).
  - Subsequent old-client and modern-client pushes: succeed (was: all failed with `target has uncommitted changes`).
  - Real SQL-induced dirty workspace on the server: push still rejected (regression preserved).

### DIFF
--- a/go/libraries/doltcore/sqle/remotesrv.go
+++ b/go/libraries/doltcore/sqle/remotesrv.go
@@ -122,6 +122,10 @@ func (s hooksFiringRemoteSrvStore) Commit(ctx context.Context, current, last has
 		return nil
 	})
 
+	// Repair stale working sets left behind by older clients before firing
+	// hooks, so hooks see the consistent post-repair state.
+	s.repairStaleWorkingSets(ctx, oldAddrs, newAddrs)
+
 	executeHooks := s.ddb.ExecuteCommitHooks
 	if s.replicaWrite {
 		executeHooks = s.ddb.ExecuteReplicaCommitHooks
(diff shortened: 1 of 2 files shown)


reltuk:
In general I don't think we want to do this. The Dolt server should not change the value of the roots hash as it was pushed by the client. The client is in control of what it wants to land.

It's unfortunate that there are versions of Dolt that compute and push the wrong value. I'm not convinced there is a good way for the server to actually know or guess what the right value was or would have been. The client's themselves are wrong, they are doing the wrong CAS, the concurrency control and visibility of the smeared write is already there, etc.

alingenhag:
Thanks for taking a look, Aaron — appreciate the quick review.

I agree with your reasoning: the server faithfully landing what the client pushed is the right contract, and a server-side rewrite of the committed root crosses that boundary even if the "right" working set value is mechanically determined for `ASSERT_WORKING_SET`. The bug lives in the older client and that's where it should be fixed. I'll close this PR and pursue a client-side bump instead (in my case, getting beads off its old pinned Dolt commit).

reltuk:
Thanks Andreas. Sorry for the churn and the busted older version which made it into Beads!

PR #10952, opened 2026-04-27
adaptive encoding for JSON columns

zachmu:
This will be turned on for the 2.0 release.

### DIFF
--- a/go/libraries/doltcore/merge/merge_prolly_rows.go
+++ b/go/libraries/doltcore/merge/merge_prolly_rows.go
@@ -46,7 +46,7 @@ import (
 // correctly fill in a new column's value for existing table rows. This can happen when a column default value uses
 // references that need to be resolved by the analyzer (e.g. column references, function references).
 var ErrUnableToMergeColumnDefaultValue = errorkinds.NewKind("unable to automatically apply column default value " +
-		"in merge: %s for table '%s'; to continue merging, first manually apply the column alteration on this branch")
+	"in merge: %s for table '%s'; to continue merging, first manually apply the column alteration on this branch")
 
 // mergeProllyTable merges the table specified by |tm| using the specified |mergedSch| and returns the new table
 // instance, along with merge stats and any error. If |diffInfo.RewriteRows| is true, then any existing rows in the
@@ -54,11 +54,11 @@ var ErrUnableToMergeColumnDefaultValue = errorkinds.NewKind("unable to automatic
 // conflicts), migrates any existing table data to the specified |mergedSch|, and merges table data from both
 // sides of the merge together.
 func mergeProllyTable(
-		ctx *sql.Context,
-		tm *TableMerger,
-		mergedSch schema.Schema,
-		mergeInfo MergeInfo,
-		diffInfo tree.ThreeWayDiffInfo,
+	ctx *sql.Context,
+	tm *TableMerger,
+	mergedSch schema.Schema,
+	mergeInfo MergeInfo,
+	diffInfo tree.ThreeWayDiffInfo,
 ) (*doltdb.Table, *MergeStats, error) {
 	mergeTbl, err := mergeTableArtifacts(ctx, tm, tm.leftTbl)
 	if err != nil {
@@ -132,15 +132,15 @@ func mergeAutoIncrementValues(ctx context.Context, tbl, otherTbl, resultTbl *dol
 }
 
 func computeProllyTreePatches(
-		ctx *sql.Context,
-		tm *TableMerger,
-		finalSch schema.Schema,
-		mergeTbl *doltdb.Table,
-		valueMerger *valueMerger,
-		mergeInfo MergeInfo,
-		diffInfo tree.ThreeWayDiffInfo,
-		patchBuffer tree.PatchBuffer,
-		s *MergeStats) (*secondaryMerger, *conflictMerger, error) {
+	ctx *sql.Context,
+	tm *TableMerger,
+	finalSch schema.Schema,
+	mergeTbl *doltdb.Table,
+	valueMerger *valueMerger,
+	mergeInfo MergeInfo,
+	diffInfo tree.ThreeWayDiffInfo,
+	patchBuffer tree.PatchBuffer,
+	s *MergeStats) (*secondaryMerger, *conflictMerger, error) {
 	ns := tm.ns
 
 	iter, err := threeWayDiffer(ctx, tm, valueMerger, diffInfo)
@@ -243,13 +243,13 @@ func computeProllyTreePatches(
 	// either skip if there's secondary indexes, or merge secondary indexes.
 	needsSchemaMigration := mergeInfo.RightNeedsRewrite || mergeInfo.LeftNeedsRewrite
 	canFastMergeProllyTrees := !keyless &&
-			!needsUniquenessValidation &&
-			!needsCheckValidation &&
-			!needsNullValidation &&
-			!needsSecondaryIndexMerge &&
-			!needsSchemaMigration &&
-			!diffInfo.RightSchemaChange &&
-			!diffInfo.LeftSchemaChange
+		!needsUniquenessValidation &&
+		!needsCheckValidation &&
+		!needsNullValidation &&
+		!needsSecondaryIndexMerge &&
+		!needsSchemaMigration &&
+		!diffInfo.RightSchemaChange &&
+		!diffInfo.LeftSchemaChange
 	if canFastMergeProllyTrees {
 		lDiff, err := tree.PatchGeneratorFromRoots(ctx, ns, ns, ancRows.Node(), leftRows.Node(), leftRows.Tuples().Order)
 		if err != nil {
@@ -1048,13 +1048,13 @@ type nullValidator struct {
 }
 
 func newNullValidator(
-		ctx context.Context,
-		final schema.Schema,
-		tm *TableMerger,
-		vm *valueMerger,
-		artEditor *prolly.ArtifactsEditor,
-		patchBuffer tree.PatchBuffer,
-		secEditors []MutableSecondaryIdx,
+	ctx context.Context,
+	final schema.Schema,
+	tm *TableMerger,
+	vm *valueMerger,
+	artEditor *prolly.ArtifactsEditor,
+	patchBuffer tree.PatchBuffer,
+	secEditors []MutableSecondaryIdx,
 ) (nullValidator, error) {
 	theirRootish, err := tm.rightSrc.HashOf()
 	if err != nil {
@@ -1212,8 +1212,8 @@ func newConflictMerger(ctx context.Context, tm *TableMerger, ae *prolly.Artifact
 		}
 
 		equal := schema.ColCollsAreEqual(a.GetAllCols(), tm.ancSch.GetAllCols()) &&
-				schema.ColCollsAreEqual(l.GetAllCols(), tm.leftSch.GetAllCols()) &&
-				schema.ColCollsAreEqual(r.GetAllCols(), tm.rightSch.GetAllCols())
+			schema.ColCollsAreEqual(l.GetAllCols(), tm.leftSch.GetAllCols()) &&
+			schema.ColCollsAreEqual(r.GetAllCols(), tm.rightSch.GetAllCols())
 		if !equal {
 			return nil, ErrConflictsIncompatible
 		}
@@ -1645,16 +1645,16 @@ func remapTuple(tuple val.Tuple, desc *val.TupleDesc, mapping val.OrdinalMapping
 // |rightSide| indicates if the tuple came from the right side of the merge; this is needed to determine if the tuple
 // data needs to be converted from the old schema type to a changed schema type.
 func remapTupleWithColumnDefaults(
-		ctx *sql.Context,
-		keyTuple, valueTuple val.Tuple,
-		valDesc *val.TupleDesc,
-		mapping val.OrdinalMapping,
-		tm *TableMerger,
-		rowSch schema.Schema,
-		mergedSch schema.Schema,
-		defaultExprs []sql.Expression,
-		pool pool.BuffPool,
-		rightSide bool,
+	ctx *sql.Context,
+	keyTuple, valueTuple val.Tuple,
+	valDesc *val.TupleDesc,
+	mapping val.OrdinalMapping,
+	tm *TableMerger,
+	rowSch schema.Schema,
+	mergedSch schema.Schema,
+	defaultExprs []sql.Expression,
+	pool pool.BuffPool,
+	rightSide bool,
 ) (val.Tuple, error) {
 	tb := val.NewTupleBuilder(mergedSch.GetValueDescriptor(tm.ns), tm.ns)
 
@@ -1708,15 +1708,15 @@ func remapTupleWithColumnDefaults(
 // writeTupleExpression attempts to evaluate the expression string |exprString| against the row provided and write it
 // to the provided index in the tuple builder. This is necessary for column default values and generated columns.
 func writeTupleExpression(
-		ctx *sql.Context,
-		keyTuple val.Tuple,
-		valueTuple val.Tuple,
-		expr sql.Expression,
-		col schema.Column,
-		sch schema.Schema,
-		tm *TableMerger,
-		tb *val.TupleBuilder,
-		colIdx int,
+	ctx *sql.Context,
+	keyTuple val.Tuple,
+	valueTuple val.Tuple,
+	expr sql.Expression,
+	col schema.Column,
+	sch schema.Schema,
+	tm *TableMerger,
+	tb *val.TupleBuilder,
+	colIdx int,
 ) error {
 	if !expr.Resolved() {
 		return ErrUnableToMergeColumnDefaultValue.New(expr.String(), tm.name)
--- a/go/libraries/doltcore/merge/merge_prolly_rows.go
+++ b/go/libraries/doltcore/merge/merge_prolly_rows.go
@@ -46,7 +46,7 @@ import (
 // correctly fill in a new column's value for existing table rows. This can happen when a column default value uses
 // references that need to be resolved by the analyzer (e.g. column references, function references).
 var ErrUnableToMergeColumnDefaultValue = errorkinds.NewKind("unable to automatically apply column default value " +
-	"in merge: %s for table '%s'; to continue merging, first manually apply the column alteration on this branch")
+		"in merge: %s for table '%s'; to continue merging, first manually apply the column alteration on this branch")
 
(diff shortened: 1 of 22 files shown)


coffeegoddd:
@zachmu __DOLT__
| covering_index_scan  | 0.54         | 0.54       | 0.0            |
| groupby_scan         | 9.91         | 9.73       | -1.82          |
| index_join           | 1.79         | 1.86       | 3.91           |
| index_join_scan      | 1.32         | 1.34       | 1.52           |
| index_scan           | 21.5         | 21.5       | 0.0            |
| oltp_point_select    | 0.25         | 0.25       | 0.0            |
| oltp_read_only       | 5.09         | 5.0        | -1.77          |
| select_random_points | 0.5          | 0.5        | 0.0            |
| select_random_ranges | 0.57         | 0.56       | -1.75          |
| table_scan           | 21.11        | 21.5       | 1.85           |
| types_table_scan     | 65.65        | 65.65      | 0.0            |


| oltp_delete_insert    | 6.32         | 6.32       | 0.0            |
| oltp_insert           | 3.07         | 3.07       | 0.0            |
| oltp_read_write       | 11.04        | 11.04      | 0.0            |
| oltp_write_only       | 5.99         | 5.88       | -1.84          |

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c4578b2</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c4578b2</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
| tpcc-scale-factor-1 | 54.83            | 53.85          | -1.79          |


| tpcc-scale-factor-1 | dolt             | 6418260c483b1ac9fc90b5ba9674d9259cf37f63 | 44.54    | dolt           | a4d08b0f8b85180a391da3cea5f9890cbf240426 | 44.74  | 0.45           |

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -2216,6 +2219,53 @@ func (m *valueMerger) mergeJSONAddr(ctx context.Context, baseAddr []byte, leftAd
...
+	if conflict {
+		return nil, true, nil
+	}
+
+	jsonBytes, err := types.MarshallJson(ctx, mergedDoc)
+	if err != nil {
+		return nil, true, err
+	}
+
+	// Mirror the storage format of the left input: if the left field was stored
```
nicktobey, line 2265: This can lead to the same problems that we're currently seeing with merging adaptive encoding values resulting in a different representation than producing the row normally. But fixing it would require fixing both issues together, so we can submit this as-is and I'll incorporate the fix in my PR and rebase it on top of this.

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -2216,6 +2219,53 @@ func (m *valueMerger) mergeJSONAddr(ctx context.Context, baseAddr []byte, leftAd
...
+	if val.AdaptiveValue(leftField).IsOutOfBand() {
+		adaptiveVal, err := val.NewOutOfBandAdaptiveValue(ctx, m.ns, jsonBytes)
+		if err != nil {
+			return nil, true, err
+		}
+		return adaptiveVal, false, nil
+	}
+
+	// Inline adaptive value: [0x00 | json_bytes]
+	result = make([]byte, 1+len(jsonBytes))
```
nicktobey: Stuff like this should probably be encapsulated by common methods in adaptive_value.go

go/store/prolly/tree/prolly_fields.go
```diff
@@ -323,6 +325,23 @@ func GetFieldValue(ctx context.Context, td *val.TupleDesc, i int, tup val.Tuple,
 		}
 		return v, nil
 
+	case val.JsonAdaptiveEnc:
+		v.Typ = querypb.Type_JSON
+		b := td.GetField(i, tup)
+		if len(b) == 0 {
+			return v, nil // NULL
+		}
+		if b[0] == 0 {
```
nicktobey: This should probably also be encapsulated in a function in adaptive_value.go

go/store/prolly/tree/prolly_fields.go
```diff
@@ -489,6 +508,35 @@ func PutField(ctx context.Context, ns NodeStore, tb *val.TupleBuilder, i int, v
 			return err
 		}
 		tb.PutJSONAddr(i, h)
+	case val.JsonAdaptiveEnc:
+		switch value := v.(type) {
+		case *val.JsonStorage:
+			if !value.IsExactLength() {
```
nicktobey: The logic here is backwards, and the comments don't explain what's actually happening.

The purpose of IsExactLength is because the current non-adaptive address types don't store an inline length, and the only way to get the length is to deserialize the value. Adaptive types on the other hand always have a length that can be computed just from the inline value.

If we're writing a value into an adaptive encoded column whose exact length isn't known (because it came from a non-adaptive address type), we need to load the bytes so that we can record the length. But if the length is known, then we can just pass it through without de-serializing.

So, the if condition needs to be flipped, and the comments should be updated.

Additionally, JSON types add another consideration that isn't handled here: out-of-band JSON chunked with the JSON chunker will not have the same address as the equivalent out-of-band text because they get chunked differently. So if the value came from an adaptive text column and is being written to an adaptive JSON column, then simply passing the address through will result in the JSON column storing a value that was not chunked with the JSON chunker and won't get the performance benefits to JSON functions.

go/store/val/adaptive_value.go
```diff
@@ -202,6 +202,20 @@ func (v AdaptiveValue) convertToGeometryStorage(ctx context.Context, vs ValueSto
 	return NewGeometryStorageOutOfBand(ctx, addr, vs, int64(length)), nil
 }
 
+func (v AdaptiveValue) convertToJsonStorage(vs ValueStore) (*JsonStorage, error) {
```
nicktobey: This (and the convertToGeometryStorage method above) need to convert this value to an out-of-band value first, like we see in convertToTextStorage. Otherwise, passing in an inlined value will produce incorrect results.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>406f8ce</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>406f8ce</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/val/adaptive_value.go
```diff
@@ -202,6 +202,20 @@ func (v AdaptiveValue) convertToGeometryStorage(ctx context.Context, vs ValueSto
...
+	length, lengthBytes := uvarint.Uvarint(v)
+	addr := hash.New(v[lengthBytes:])
+	return NewJsonStorageOutOfBand(addr, vs, int64(length)), nil
+}
+
+// NewOutOfBandAdaptiveValue writes |data| to |vs| and returns an out-of-band AdaptiveValue
+// encoding [varint(len(data)) | content_hash]. This is used when writing adaptive values
+// outside the TupleBuilder (e.g. in the merge path).
+func NewOutOfBandAdaptiveValue(ctx context.Context, vs ValueStore, data []byte) (AdaptiveValue, error) {
+	inline := AdaptiveValue(append([]byte{0}, data...))
```
nicktobey: This is correct but inefficient: we're copying the data slice into an adaptive value, and then convertToOutOfBand is just going to pull it back out again in order to chunk it.

go/store/val/tuple_builder_test.go
```diff
@@ -272,6 +282,191 @@ func (t *TestValueStore) WriteBytes(_ context.Context, val []byte) (h hash.Hash,
...
+		require.NoError(t, err)
+		tup, err := tb.Build(testPool)
+		require.NoError(t, err)
+
+		result, ok, err := td.GetJsonAdaptiveValue(ctx, 0, vs, tup)
+		require.NoError(t, err)
+		require.True(t, ok)
+		require.NotNil(t, result)
+
+		// Inline values have IsExactLength() == true.
```
nicktobey: All adaptive values have IsExactLength() == true, not just inline ones. IsExactLength exists so that non-adaptive address types can still implement AnyWrapper.

go/store/val/value_store.go
```diff
@@ -301,6 +302,82 @@ func (g *GeometryStorage) Addr() hash.Hash {
...
+	}
+	var val interface{}
+	if err = json.Unmarshal(buf, &val); err != nil {
+		return nil, err
+	}
+	return val, nil
+}
+
+// IsExactLength returns true when the bytes are stored inline (exact length known without loading).
+func (j *JsonStorage) IsExactLength() bool {
```
nicktobey: I see what you're going for here, allowing JsonStorage to represent both inline and out-of-band values. But currently this has some limitations:
- It's not possible to express an out-of-band value with a known exact length, such as a value read from a JSON adaptive encoding column
- By using this type as the sql.JSONWrapper implementation instead of IndexedJsonDocument, we lose the ability to do efficient lookups and modifications.

It's possible that want to combine this type with IndexedJsonDocument, and give IndexedJsonDocument the ability to store the length of a document with an exact known length. But other than that, I don't think it makes sense to use this type instead of using JsonDocument for inline values and IndexedJsonDocument from out-of-band values.

go/libraries/doltcore/merge/merge_prolly_rows.go
zachmu: sgtm, you can address next

go/store/val/value_store.go
zachmu: Just robot poop, fixed.

go/store/prolly/tree/prolly_fields.go
zachmu: I believe this final concern is handled by JsonType.Convert, which explicitly handles sql.StringWrapper values. Only *JsonAdaptiveStorage gets passed through unchanged.

nicktobey: As long as Convert is getting called on all code paths, this should be safe.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b4de013</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b4de013</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>bca295e</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>bca295e</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>0c5845a</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>0c5845a</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/merge/merge_prolly_rows.go
```diff
@@ -2259,26 +2259,38 @@ func (m *valueMerger) mergeJSONAdaptive(ctx context.Context, baseField []byte, l
...
-	result[0] = 0
-	copy(result[1:], jsonBytes)
+	result = val.AdaptiveValueInlineBytes(jsonBytes)
 	return result, false, nil
 }
 
-func mergeJSON(ctx context.Context, ns tree.NodeStore, base, left, right sql.JSONWrapper) (resultDoc sql.JSONWrapper, conflict bool, err error) {
+func mergeJSON(ctx context.Context, ns tree.NodeStore, base, left, right any) (resultDoc sql.JSONWrapper, conflict bool, err error) {
+	var baseJson, leftJson, rightJson sql.JSONWrapper
+	baseJson, err = toJsonWrapper(base)
```
nicktobey: It seems like a code smell that the method takes `any` parameters instead of JSONWrappers, and then the immediate first thing it does is forcibly case them to JSONWrappers. It's possible that this is the best way to do it, because the alternative would be inserting lots of casts at the callsites, but I'm skeptical. You have more context here: is this really the best way to do this?

go/store/prolly/tree/prolly_fields.go
```diff
@@ -116,6 +116,18 @@ func GetField(ctx context.Context, td *val.TupleDesc, i int, tup val.Tuple, ns N
 		v, ok, err = td.GetGeomAdaptiveValue(ctx, i, ns, tup)
 	case val.JsonAdaptiveEnc:
 		v, ok, err = td.GetJsonAdaptiveValue(ctx, i, ns, tup)
```
nicktobey, line 126: None of the other adaptive encoding types require this extra step, and it's really weird that GetJsonAdaptiveValue can return either a JsonAdaptiveStorage or a []byte.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>cd320b9</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>cd320b9</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/prolly_fields.go
```diff
@@ -328,32 +340,35 @@ func GetFieldValue(ctx context.Context, td *val.TupleDesc, i int, tup val.Tuple,
 	case val.JsonAdaptiveEnc:
```
nicktobey, line 348: The different cases here are almost identical, and leak implementation details about the AdaptiveValues type by operating on the underlying slice. Can we move this logic into AdaptiveValues?

go/store/prolly/tree/prolly_fields.go
```diff
@@ -116,6 +116,18 @@ func GetField(ctx context.Context, td *val.TupleDesc, i int, tup val.Tuple, ns N
 		v, ok, err = td.GetGeomAdaptiveValue(ctx, i, ns, tup)
 	case val.JsonAdaptiveEnc:
 		v, ok, err = td.GetJsonAdaptiveValue(ctx, i, ns, tup)
+		if ok {
+			switch val := v.(type) {
+			case *val.JsonAdaptiveStorage:
+				v, err = NewJSONDoc(val.Addr(), ns).ToIndexedJSONDocument(ctx)
```
nicktobey, line 130: Thinking about it more, does it make more sense to just return the JsonAdaptiveStorage value unchanged here? Then, functions that want to take mutate or do lookups can convert it to an IndexedJSONDocument in order to call the extra methods on that type.

That conversion could happen multiple times when evaluating a query. But loading the root chunk should be cached.

The alternative option would be to have a single struct type that lazily fetches the root chunk, basically combining JsonAdaptiveStorage and IndexedJsonDocument, but that could get complicated.

go/store/val/value_store.go
```diff
@@ -302,56 +301,41 @@ func (g *GeometryStorage) Addr() hash.Hash {
...
-// JsonStorage wraps raw JSON bytes and defers deserialization until the value is needed.
+// JsonAdaptiveStorage wraps raw JSON bytes and defers deserialization until the value is needed.
 // The bytes may be stored inline or out-of-band via a content address.
 // It implements sql.JSONWrapper and types.JSONBytes.
-type JsonStorage struct {
-	// inlineBytes holds the raw JSON bytes when the value is stored inline.
-	// When nil, the value is out-of-band and must be loaded via outOfBand.
-	inlineBytes []byte
-	// outOfBand holds the lazily-loaded address. Only used when inlineBytes is nil.
+type JsonAdaptiveStorage struct {
```
nicktobey, line 307: I think having this as a separate implementation of JSONWrapper is going to be error-prone, because there's going to be places where we end up fully deserializing when we could be converting it into an IndexedJsonDocument.

I left a comment in `prolly_fields.go` discussing options.

go/store/prolly/tree/prolly_fields.go
zachmu: Done

go/store/prolly/tree/prolly_fields.go
zachmu: Discussed offline. Agreed that it's strange that Strings, Bytes, and now JSON can all return different value types from the storage layer. Addressing this requires standardizing on wrapper types and semantics, especially for JSON.

go/libraries/doltcore/merge/merge_prolly_rows.go
zachmu: I think it's better to perform these conversions at the call site before passing. There's only the one call site.

nicktobey (approved):

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a4d08b0</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a4d08b0</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #10956, opened 2026-04-28
feat: add -C/--directory global flag matching git -C behaviour

rjc123:
## Summary

Adds a `-C <dir>` / `--directory <dir>` global flag so dolt commands can be run against a database directory without cd-ing into it first. This matches the convention established by `git -C` and is useful for scripting and agent-driven workflows where the caller's cwd is not the dolt repo.

The existing `--chdir` flag is unchanged.

## Motivation

`git -C`, `make -C`, and `ninja -C` have established `-C` as the standard short flag for "run as if in this directory." Dolt users coming from git muscle memory expect it. The existing `--chdir` is undocumented and was added as an internal GoLand IDE workaround — it is not a substitute for a user-facing flag.

## Implementation

A lightweight `preScanDirFlag` helper scans raw args for `-C`/`--directory` before `createBootstrapConfig` creates `cwdFs`. The target directory is resolved via `LocalFilesysWithWorkingDir`, which creates a new `Filesys` instance rooted at the target path without mutating the process cwd. All downstream resolution (data-dir, local config, env loading) inherits the correct base directory through `cwdFs`.

## Changes

- `go/cmd/dolt/cli/arg_parser_helpers.go`: adds `--directory`/`-C` to `CreateGlobalArgParser`
- `go/cmd/dolt/dolt.go`: adds `preScanDirFlag` helper and directory resolution in `createBootstrapConfig`
- `go/cmd/dolt/dolt_test.go`: unit tests for `preScanDirFlag` (8 cases); integration tests gated on `DOLT_TEST_BUILD=1`

## Test plan

- [ ] `go test ./cmd/dolt/ -run TestPreScanDirFlag` — unit tests, no binary required
- [ ] `DOLT_TEST_BUILD=1 go test ./cmd/dolt/ -run TestDirectoryFlagIntegration` — end-to-end: valid dir, long form, nonexistent dir error, cwd-isolation guarantee
- [ ] `dolt -C /path/to/db status` works from any directory
- [ ] `dolt -C /nonexistent status` exits with a clean error

Note: the revision history on this PR reflects an overzealous code assistant — the `--chdir` cleanup has been correctly separated into #10957.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10957, opened 2026-04-28
fix: harden --chdir error handling; add deprecation warning

rjc123:
## Summary

Two independent commits — cherry-pick either or both:

**Commit 1 — bug fixes (uncontroversial):**
- `--chdir` without a following argument previously read `args[1]` out of bounds; now exits with a clear error
- `os.Chdir()` failure previously called `panic(err)`; now exits cleanly with a readable message

**Commit 2 — deprecation (for maintainers to evaluate):**
- Emits a yellow warning to stderr when `--chdir` is used, pointing users to `-C` (PR #10956)
- Registers `--chdir` in the global arg parser so it appears in `--help` with a `[deprecated]` note

**Note on commit 2:** `--chdir` is handled in the debug-flags pre-pass in `runMain()`, not by the global arg parser — so the parser registration is for help-text visibility only. The underlying `os.Chdir()` behaviour is unchanged. The tradeoffs for maintainers to consider:
- The runtime warning on stderr could break GoLand or any tooling that reads combined output — though `--chdir` was explicitly a GoLand workaround, so this risk may be low
- Registering in the arg parser makes `--chdir` visible in `--help` for the first time, which may be undesirable if the intent is to quietly retire it
- The deprecation message references `-C` which lands in #10956 — if this PR is merged first, the message still makes sense as a forward reference

We have what we came for with #10956. This PR is offered as a housekeeping contribution — accept, reject, or cherry-pick the bug-fix commit only as you see fit.

Note: the revision history on this and #10956 reflects an overzealous code assistant. The PRs were correctly separated and structured on reflection.

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -392,6 +392,7 @@ func CreateGlobalArgParser(name string) *argparser.ArgParser {
 	ap.SupportsString("port", "", "port", "Defines the port to connect to.")
 	ap.SupportsFlag("no-tls", "", "Disables TLS for the connection to remote databases.")
 	ap.SupportsString("data-dir", "", "data-dir", "Defines a data directory whose subdirectories should all be dolt data repositories accessible as independent databases. Defaults to the current directory.")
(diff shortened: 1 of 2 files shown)


zachmu:
We're going to reject this one. --chdir is not intended for external use, which is why it is undocumented. We'll get rid of it once -C becomes standard.

PR #10976, opened 2026-05-01
fix(sql): use LEFT JOIN in blame view to handle schema-only reverts

### LINKED ISSUE #10965: dolt_blame returns no rows after schema-only dolt_revert
## Problem

After a schema-only commit is reverted with `dolt_revert`, `dolt_blame_<table>` can return zero rows even though the table still has live rows.

Doltlite returns the expected blame row here, so this looks like a Dolt-side blame / lineage bug.

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
INSERT INTO t VALUES (1, 'base');
CALL dolt_add('-A');
CALL dolt_commit('-m', 'c1');

CREATE TABLE t_new(id INTEGER PRIMARY KEY, v TEXT CHECK (length(v) > 0));
INSERT INTO t_new SELECT * FROM t;
DROP TABLE t;
ALTER TABLE t_new RENAME TO t;
CALL dolt_add('-A');
CALL dolt_commit('-m', 'main_check');

CALL dolt_revert((SELECT commit_hash FROM dolt_log WHERE message='main_check' LIMIT 1));

SELECT * FROM t;
SELECT count(*) FROM dolt_blame_t;
SELECT CONCAT('BL|', id, '|', message) FROM dolt_blame_t ORDER BY id;
```

## Actual behavior

`SELECT * FROM t;` returns the live row:

```text
1,base
```

But blame is empty:

```text
count(*) = 0
```

and

```text
SELECT CONCAT('BL|', id, '|', message) FROM dolt_blame_t ORDER BY id;
```

returns no rows.

## Expected behavior

`dolt_blame_t` should return one row for the live row in `t`, attributed to the original commit `c1`.

wucm667:
## Description

When a schema-only commit (involving DROP TABLE + RENAME TO) is reverted,
`dolt_blame_<table>` returns zero rows even though the table has live data.

## Root Cause

The blame view used an implicit cross join (`FROM a, b WHERE a.x = b.y`) to
join `sorted_diffs_by_pk` with `dolt_log`. After a schema-only revert, the
`to_commit` hash from the diff table may not match any commit in `dolt_log`,
causing all rows to be filtered out.

## Fix

Changed the blame view SQL to use a `LEFT JOIN` between `sorted_diffs_by_pk`
and `dolt_log`, so rows from the diff table are returned even when the commit
hash lookup in `dolt_log` doesn't find a match.

Fixes #10965

### DIFF
--- a/go/libraries/doltcore/sqle/dtables/blame_view.go
+++ b/go/libraries/doltcore/sqle/dtables/blame_view.go
@@ -42,10 +42,10 @@ const (
 								 diff_type,
 				                 ROW_NUMBER() OVER (
 				                     PARTITION BY 
-										%s  -- pksPartitionByExpression
+											%s  -- pksPartitionByExpression
 				                     ORDER BY 
-										coalesce(to_commit_date, from_commit_date) DESC
-								) row_num
+											coalesce(to_commit_date, from_commit_date) DESC
+									) row_num
 				             FROM
 				                 ` + "`dolt_diff_%s`" + ` -- tableName
 				            )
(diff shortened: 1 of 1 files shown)


zachmu (commented):
LGTM!

zachmu (approved):

PR #11001, opened 2026-05-05
Store adaptive-encoded values in lexical order

zachmu:
In Dolt, all adaptive-encoded values are stored in an arbitrary order according to their raw bytes.

```go
func compareAdaptiveValue(l, r AdaptiveValue) int {
    return bytes.Compare(l, r)
}
```

This isn't a problem for Dolt, because keys on columns that will be adaptively encoded post 2.0 (TEXT, BLOB) all require a key prefix length, which is used by the secondary index to order the keys correctly.

But for Doltgres, it's possible to create indexes on TEXT and VARCHAR (no length) columns without a key prefix length. This means that adaptive-encoded values must be stored in a canonical order independent of whether the value is stored inline or out-of-band. Otherwise, various SQL engine optimizations, like removing the Sort operation on an index range scan, would not yield correct results in Doltgres. There are also many other known and unknown assumptions in Dolt and go-mysql-server that break if values in indexed columns are not stored in a consistent order. Additionally, enforcing a canonical order for such columns makes it possible to bring these enhancements to Dolt in the future (by relaxing MySQL's constraints on key lengths), or in new products we might build.

This PR sorts adaptive encoded values by adding a ValueStore param to the `Compare()` interface. It is unused except when comparing adaptive values.

The downside to this change is that we must perform 1 or 2 additional disk reads for each comparison of out-of-band values. This is a large performance penalty but is unavoidable, and should (we think) happen in a small minority of cases.


coffeegoddd:
@zachmu __DOLT__
| covering_index_scan  | 2.3          | 2.3        | 0.0            |
| groupby_scan         | 134.9        | 137.35     | 1.82           |
| index_join           | 1.93         | 2.0        | 3.63           |
| index_join_scan      | 1.34         | 1.34       | 0.0            |
| index_scan           | 215.44       | 219.36     | 1.82           |
| oltp_point_select    | 0.26         | 0.26       | 0.0            |
| oltp_read_only       | 5.18         | 5.18       | 0.0            |
| select_random_points | 0.55         | 0.56       | 1.82           |
| select_random_ranges | 0.67         | 0.64       | -4.48          |
| table_scan           | 211.6        | 207.82     | -1.79          |
| types_table_scan     | 484.44       | 484.44     | 0.0            |


| oltp_delete_insert    | 6.21         | 6.21       | 0.0            |
| oltp_insert           | 3.19         | 3.19       | 0.0            |
| oltp_read_write       | 11.45        | 11.45      | 0.0            |
| oltp_update_index     | 3.3          | 3.3        | 0.0            |
| oltp_write_only       | 6.32         | 6.32       | 0.0            |
| types_delete_insert   | 6.91         | 6.91       | 0.0            |

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>46acdca</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>46acdca</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
| tpcc-scale-factor-1 | 44.17            | 44.98          | 1.83           |


| tpcc-scale-factor-1 | dolt             | 4a69899a3aeec3cee42867ee63a7d4fa0b4c1863 | 53.43    | dolt           | 27f1f3e0e1db99e5017b70cd7a8a028824ba3506 | 53.22  | -0.39          |

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>c513e73</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>c513e73</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a28db5a</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a28db5a</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>8d7d0ea</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>8d7d0ea</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>d3b9652</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>d3b9652</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>76b1a73</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>76b1a73</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>30d36b2</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>30d36b2</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/libraries/doltcore/schema/collation_comparator.go
```diff
@@ -93,12 +93,25 @@ func (c CollationTupleComparator) Validated(types []val.Type) val.TupleComparato
 		if types[i].Enc == val.StringEnc {
 			panic("string type encoding is missing its collation")
 		}
-		newCollations[i] = sql.Collation_Unspecified
+		if isCollatedStringEnc(types[i].Enc) {
+			newCollations[i] = sql.Collation_Default
```
nicktobey, line 103: I don't understand the change here. We panic if there's a StringEnc without a collation but not if there's a StringAdaptiveEnc without a collation?

go/store/prolly/tree/blob_builder_test.go
```diff
@@ -401,6 +405,294 @@ func mustNewBlob(ctx context.Context, ns NodeStore, len, chunkSize int) hash.Has
...
+				val.TupleDescriptorArgs{ValueStore: cjs},
+				val.Type{Enc: val.JsonAdaptiveEnc},
+			)
+
+			cjs.bytesReads = 0
+			cjs.jsonCmps = 0
+			got := td.Comparator().CompareValues(ctx, 0, lVal, rVal, val.Type{Enc: val.JsonAdaptiveEnc})
+			require.Equal(t, c.want, got)
+			require.Equal(t, 1, cjs.jsonCmps, "JsonAdaptiveEnc must dispatch through CompareJsonAdaptiveValues")
+			require.Equal(t, 0, cjs.bytesReads, "JSON compare must not fall back to the byte-streaming path")
```
nicktobey: This confuses me. If the values are stored out of band, then they need to be read from the chunk store to be compared. So why is bytesReads 0 here?

go/store/prolly/tree/blob_builder.go
```diff
@@ -242,6 +242,52 @@ func (b *BlobBuilder) write(ctx context.Context, keys, vals [][]byte, subtrees [
 	return h, nil
 }
 
+// blobChunkReader streams the leaf chunks of a blob tree in order via val.ValueChunkReader.
+// It walks the tree lazily — only the nodes on the current path from the root to the next
+// leaf chunk are loaded, so callers can stop early without materializing the whole blob.
+type blobChunkReader struct {
```
nicktobey: An improvement on this concept would be to use a differ that walks both trees at once.

If the first level 0 chunk is the same in both blobs, then the first level 1 chunk in each blob will store the same hash, and we can skip loading the first level 0 chunk entirely. This lets us seek directly to the first chunk of the blob that's actually different instead of reading every prior chunk.

go/store/val/tuple_descriptor.go
```diff
@@ -72,6 +72,9 @@ type TupleTypeHandler interface {
 type TupleDescriptorArgs struct {
 	Comparator TupleComparator
 	Handlers   []TupleTypeHandler
+	// ValueStore, if non-nil, is attached to the TupleDesc's Comparator via |WithValueStore|.
+	// It is required when comparing tuples with adaptive-encoded fields.
+	ValueStore ValueStore
```
nicktobey, line 77: Does this field need to exist if the Comparator also holds a reference to the ValueStore? It also feels like in practice, this should never be nil, right?

go/store/val/adaptive_value.go
```diff
@@ -224,6 +268,19 @@ func (v AdaptiveValue) convertToJsonStorage(ctx context.Context, vs ValueStore)
 	return NewJsonStorageOutOfBand(addr, vs, int64(length)), nil
 }
 
+// OutOfBandAdaptiveValueAddr returns the content address embedded in an out-of-band
+// AdaptiveValue. It returns an error if v is NULL or inline.
+func OutOfBandAdaptiveValueAddr(v AdaptiveValue) (hash.Hash, error) {
```
nicktobey: Should this be a method?

go/store/prolly/tree/blob_builder_test.go
zachmu: Test could be badly stated, I'll review it.

go/libraries/doltcore/schema/collation_comparator.go
zachmu: Agree this is suspect.

go/store/prolly/tree/blob_builder.go
zachmu: Good suggestion, will do.

go/store/val/tuple_descriptor.go
zachmu: I don't think we can guarantee it is never nil, too many code paths that don't have a value store at hand.

Does seem substantially duplicated by the comparator value though.

go/store/val/adaptive_value.go
zachmu: Good call.

go/store/val/tuple_descriptor.go
zachmu: So we use the ValueStore on the Args to set the one on the Comparator during construction. I'll add a note about this.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b704c77</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b704c77</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b3f7ade</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b3f7ade</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>b79ff79</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>b79ff79</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>1095112</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>1095112</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tuple_range_iter_test.go
```diff
@@ -149,23 +160,32 @@ func testIterPrefixRange(t *testing.T, om testMap, tuples [][2]val.Tuple) {
 		stop, err := getKeyPrefix(tuples[z][0], prefixDesc)
 		require.NoError(t, err)
 
+		openR, err := openRange(ctx, start, stop, prefixDesc)
```
nicktobey: This should use the same `mustRange` helper function approach used further down in the file.

go/store/prolly/tuple_range_iter_test.go
```diff
@@ -227,7 +249,15 @@ func getExpectedRangeSize(rng Range, tuples [][2]val.Tuple) (sz int) {
 	ctx := context.Background()
 	for i := range tuples {
 		k := tuples[i][0]
-		if rng.aboveStart(ctx, k) && rng.belowStop(ctx, k) {
+		above, err := rng.aboveStart(ctx, k)
+		if err != nil {
+			panic(err)
```
nicktobey: This shouldn't panic, it should either return the error, or accept `t` and call `require.NoError(t, err)`

go/store/val/codec.go
```diff
@@ -877,31 +886,27 @@ func compareCollatedChunkReaders(ctx context.Context, l, r ValueChunkReader, col
...
-// or when one stream is exhausted while the other still has data.
-func compareChunkReaders(ctx context.Context, l, r ValueChunkReader) (int, error) {
+// compareChunkDiffer compares two values using the output of a ChunkDiffer built on them, using the diffs it generates.
+func compareChunkDiffer(ctx context.Context, d ChunkDiffer) (int, error) {
 	var lBuf, rBuf []byte
 	lDone, rDone := false, false
 	for {
-		if len(lBuf) == 0 && !lDone {
-			chunk, err := l.NextChunk(ctx)
+		for (len(lBuf) == 0 && !lDone) || (len(rBuf) == 0 && !rDone) {
```
nicktobey: Why would we expect to get an empty chunk before we finish iterating?

go/store/val/codec.go
nicktobey: Given that the differ is designed to return the first chunk where the two values diff, is there any reason to suspect that this comparison will ever be 0?

The only thing I can think of collations such as case insensitivity... but in that case we should be using a collation aware comparator, not bytes.Compare.

So for text blobs with a collation, I think it makes sense to loop here but we need a collation sensitive comparator. For other types of adaptive encoded values we should never end up in a loop here.

zachmu: Yeah this entire method iterates and shouldn't, kind of brain-dead. Fixed now.

go/store/prolly/tree/blob_chunk_differ.go
```diff
@@ -0,0 +1,183 @@
...
+		rh := d.r.stack[0].node.HashOf()
+		if lh == rh {
+			d.l.stack = nil
+			d.r.stack = nil
+		}
+	}
+	return d, nil
+}
+
+func loadBlobDiffSide(ctx context.Context, ns NodeStore, side *blobDiffSide, v val.AdaptiveValue) error {
```
nicktobey: Should this be a method on blobDiffSide?

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>bbe2623</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>bbe2623</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/json_indexed_document.go
```diff
@@ -134,6 +136,73 @@ func getBytesFromIndexedJsonMap(ctx context.Context, m StaticJsonMap) (bytes []b
 	return bytes, err
 }
 
+// openJsonAdaptiveValue resolves an adaptive JSON value to a sql.JSONWrapper. Inline values
+// are wrapped as LazyJSONDocuments over their in-memory payload; out-of-band values become
+// IndexedJsonDocuments (or, when the underlying tree predates the JSON-indexed format, fall
+// back to a LazyJSONDocument via JSONDoc.ToIndexedJSONDocument). NULL adaptive values return
+// (nil, nil) so the caller can apply NULL ordering.
+func openJsonAdaptiveValue(ctx context.Context, ns NodeStore, v val.AdaptiveValue) (sql.JSONWrapper, error) {
```
nicktobey: There's some code duplication here and in tree.GetField. We should probably have a canonical function for this operation.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>d5cc646</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>d5cc646</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/merge.go
```diff
@@ -179,14 +179,22 @@ func SendPatches[K ~[]byte, O Ordering[K]](
 		leftLevel, _ := l.getLevel()
 		rightLevel, _ := r.getLevel()
 		if leftLevel > 0 && rightLevel > 0 {
-			if compareWithNilAsMin(ctx, order, K(left.EndKey), K(right.KeyBelowStart)) <= 0 {
+			cmpLR, cmpErr := compareWithNilAsMin(ctx, order, K(left.EndKey), K(right.KeyBelowStart))
+			if cmpErr != nil {
+				return cmpErr
+			}
+			cmpRL, cmpErr := compareWithNilAsMin(ctx, order, K(right.EndKey), K(left.KeyBelowStart))
```
nicktobey: Nit: the added variable names are vague, and we don't need to compute both before doing any comparisons.

I suggest cmpLR -> compareLeftEndWithRightStart, and cmpRL -> compareRightEndWithLeftStart

nicktobey: Or heck, if the result is used immediately after the call to `compareWithNilAsMin` instead of getting them both up front, they could just be `cmp`

go/store/prolly/tree/merge.go
```diff
@@ -271,7 +294,15 @@ func SendPatches[K ~[]byte, O Ordering[K]](
 		}
 
 		if leftLevel > 0 {
-			if compareWithNilAsMin(ctx, order, K(right.EndKey), K(left.KeyBelowStart)) <= 0 {
+			cmpRL, cmpErr := compareWithNilAsMin(ctx, order, K(right.EndKey), K(left.KeyBelowStart))
```
nicktobey: Same as above, no need to compute these both in advance and give them different vague names.

go/store/prolly/tuple_range_iter_test.go
```diff
@@ -48,26 +48,35 @@ func testIterRange(t *testing.T, om testMap, tuples [][2]val.Tuple) {
...
+		require.NoError(t, err)
+		closedR, err := closedRange(ctx, start, stop, desc)
+		require.NoError(t, err)
+
 		tests := []rangeIterTest{
 			// two-sided ranges
 			{
 				name:      "OpenRange",
-				testRange: openRange(ctx, start, stop, desc),
+				testRange: openR,
```
nicktobey: Probably better to wrap these in a `mustRange` helper function.

go/store/prolly/tuple_range_iter_test.go
```diff
@@ -149,23 +160,33 @@ func testIterPrefixRange(t *testing.T, om testMap, tuples [][2]val.Tuple) {
...
+		openStartR := mustRange(openStartRange(ctx, start, stop, prefixDesc))
+		openStopR := mustRange(openStopRange(ctx, start, stop, prefixDesc))
+		closedR := mustRange(closedRange(ctx, start, stop, prefixDesc))
+
 		tests := []prefixRangeTest{
 			// two-sided ranges
 			{
 				name:      "OpenRange",
-				testRange: openRange(ctx, start, stop, prefixDesc),
+				testRange: openR,
```
nicktobey: Same as above, prefer a `mustRange` helper function.

go/store/val/value_store.go
```diff
@@ -37,6 +37,30 @@ type ValueStore interface {
...
+type ChunkDiffer interface {
+	// Next returns the bytes of the next chunks that differ between left and right. io.EOF signals
+	// the end of the diff.
+	Next(ctx context.Context) (left, right []byte, err error)
+}
+
+// ChunkDiffValueStore is an optional interface implemented by ValueStores that can compare two
+// out-of-band AdaptiveValues by walking their underlying chunked storage in parallel.
+// adaptive encodings (BytesAdaptiveEnc, StringAdaptiveEnc, GeomAdaptiveEnc).
+type ChunkDiffValueStore interface {
```
nicktobey: This whole thing is weird.

I can see why this is a separate interface instead of just being a method on ValueStore: in theory, a ValueStore implementation might not use nodes or chunking, nodes and chunking are a facet of the NodeStore, which implements ValueStore.

But in practice, NodeStore is effectively the only implementation, and the others just wrap it. Currently, every ValueStore implementation also implements ChunkDiffValueStore, so most of the code in `getChunkDiffer` is unreachable. And all the implementations just defer to calling `newBlobChunkDiffer`. 

Ultimately, I think this approach is leaking implementation details into the `val` package, because this package isn't supposed to know or care about chunking. And the caller hear doesn't need to know or care about chunking: what the caller actually wants is the ability to ask the ValueStore to compare two adaptive values, and then allow the ValueStore implementation (in this case, NodeStore) to leverage implementation details to do that comparison efficiently.

So given that, I think the more correct approach is to add a method to ValueStore for comparing two adaptive values, with a parameter that describes encoding/collation, then having NodeStore implement that method. I don't think we need to have a public ChunkDiffer interface at all.

go/store/prolly/tree/blob_chunk_differ.go
```diff
@@ -0,0 +1,225 @@
...
+import (
+	"context"
+	"io"
+
+	"github.com/dolthub/dolt/go/store/val"
+)
+
+// blobChunkDiffer is an implementation of val.ChunkDiffer for chunked blob values.
+//
+// This implementation may return identical chunks as diffs after the first differing chunk is found. It
```
nicktobey: The entire comment doesn't really make sense. I think I see what it's getting at: if we're comparing text with a collation, the split may have occurred in the middle of a rune (I think?), and we need to load the immediate next chunk to complete the runes and actually do the comparison. But this has nothing to do with "the byte offsets of those identical chunks may be different", which isn't relevant at all.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>07c7c11</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>07c7c11</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@coffeegoddd __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>26b08f5</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>26b08f5</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>86de1a1</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>86de1a1</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

go/store/prolly/tree/node_store.go
```diff
@@ -247,6 +247,130 @@ func (ns *nodeStore) WriteBytes(ctx context.Context, b []byte) (hash.Hash, error
...
+		case len(lBuf) == 0:
+			return -1, nil
+		case len(rBuf) == 0:
+			return 1, nil
+		}
+		lr, lread := utf8.DecodeRune(lBuf)
+		rr, rread := utf8.DecodeRune(rBuf)
+		lErr := lr == utf8.RuneError && lread == 1
+		rErr := rr == utf8.RuneError && rread == 1
+		if lErr || rErr {
```
nicktobey, line 354: What's going on here? If one size has a rune decoding error and the other doesn't, the one that has a valid ending is considered greater? A comment making this clear might help.

go/store/prolly/tree/node_store.go
```diff
@@ -247,6 +247,130 @@ func (ns *nodeStore) WriteBytes(ctx context.Context, b []byte) (hash.Hash, error
...
+
+// compareCollatedChunkDiffer is the streaming version of CompareCollatedStrings, getting just the diff between two
+// values when comparing them.
+func compareCollatedChunkDiffer(ctx context.Context, d chunkDiffer, collation sql.CollationID) (int, error) {
+	getRuneWeight := collation.Sorter()
+	var lBuf, rBuf []byte
+	lDone, rDone := false, false
+
+	pullFrom := func() error {
+		for (!lDone && !utf8.FullRune(lBuf)) || (!rDone && !utf8.FullRune(rBuf)) {
```
nicktobey, line 316: Does utf8.FullRune scan the entire buffer? This could be O(N^2) if there are N consecutive chunks that end in the middle of runes, but I'm not sure of a good way around that and it might not matter in practice.

go/store/prolly/tree/node_store.go
```diff
@@ -247,6 +247,130 @@ func (ns *nodeStore) WriteBytes(ctx context.Context, b []byte) (hash.Hash, error
...
+			lChunk, rChunk, err := d.Next(ctx)
```
nicktobey, line 317: I think there's a bug that can happen here when a chunk boundary occurs in the middle of a rune.

Imagine both sides have three chunks, where every chunk boundary occurs in the middle of a rune:
- Chunk A contains different bytes but are equal under collation
- Chunk B is identical
- Chunk C contains another difference

If we diff these, the differ will first return the contents of Chunk A. We will identify that we don't have a full rune and call the differ again, which will now return the contents of Chunk C. This function will then concatenate chunks A and C in the buffer, skipping over Chunk A. This will result in a codepoint appearing on the boundary that is not actually in either string, and could influence the result of the comparison..

nicktobey: Never mind, I looked at the differ implementation and realized that all calls after the first return subsequent chunks. This can't actually happen. However, this is a bit unexpected from looking at the call sites. It might be better to have separate NextDiff and NextChunk methods, making it clear that NextChunk doesn't skip ahead.

Having two separate methods also helps us avoid a potential slow path, where we're comparing two text values with collation, and the first chunk contains a byte diff (but are equal under collation), and then the next byte diff doesn't occur for many many chunks. We should be able to skip ahead in that case.

Finally, it might be worth considering a case where we're comparing two blobs that have different chunking strategies and thus may contain different chunk boundaries for the same contents. This isn't supposed to be possible for TEXT or BLOB values, but could happen as a result of fixing bugs. And it may be possible with JSON columns, since a JSON value might be copied into a an adaptive JSON column from a JSON address column containing values that used the old fixed-length chunker instead of the JSON chunker.

In order to correctly handle these cases, we should make sure we have correct behavior if one chunk in a diff is a superstring of the other and there are additional chunks to read.

coffeegoddd:
@zachmu __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>27f1f3e</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>27f1f3e</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

PR #11012, opened 2026-05-07
Warn users about lone dolt_checkout() in `dolt sql -q`

### LINKED ISSUE #6876: Warn users about running `dolt sql -q "call dolt_checkout(...);"`

kevglynn:
## Summary

- When `dolt sql -q "call dolt_checkout('branch')"` is run as a single statement, print a yellow warning to stderr explaining that the branch change only persists for the current session
- Multi-statement queries (e.g. `call dolt_checkout('branch'); select * from t`) do NOT trigger the warning, since combining `dolt_checkout()` with other statements in a single invocation is valid usage
- Suggests using `dolt checkout <branch>` instead for persistent branch changes

## Changes

- `go/cmd/dolt/commands/sql.go`: Added `warnIfLoneDoltCheckout()` and `isLoneDoltCheckoutCall()` — called from `queryMode` before executing the query
- `go/cmd/dolt/commands/sql_test.go`: Added `TestIsLoneDoltCheckoutCall` with 11 test cases (single call, uppercase, mixed case, `-b` flag, different procedures, multi-statement, empty, bad syntax)
- `integration-tests/bats/sql-checkout.bats`: Added 2 BATS integration tests — one confirming the warning appears for a lone checkout, one confirming it does NOT appear with multiple statements

## Test plan

- [x] `go test -run TestIsLoneDoltCheckoutCall ./cmd/dolt/commands/` — all 11 cases pass
- [x] Manual E2E verification: warning appears for lone `dolt_checkout()`, does not appear for other procedures or multi-statement input
- [ ] CI: BATS `sql-checkout` suite

Closes #6876

### DIFF
--- a/go/cmd/dolt/commands/debug.go
+++ b/go/cmd/dolt/commands/debug.go
@@ -428,5 +428,5 @@ func execDebugMode(ctx *sql.Context, qryist cli.Queryist, queryFile *os.File, co
 	}()
 	input := bufio.NewReader(transform.NewReader(queryFile, textunicode.BOMOverride(transform.Nop)))
 
-	return execBatchMode(ctx, qryist, input, continueOnErr, format, false)
+	return execBatchMode(ctx, qryist, input, continueOnErr, format, false, nil)
 }
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -292,7 +292,7 @@ func (cmd SqlCmd) Exec(ctx context.Context, commandStr string, args []string, dE
 			}
 		} else {
 			input = transform.NewReader(input, textunicode.BOMOverride(transform.Nop))
-			err := execBatchMode(queryist.Context, queryist.Queryist, input, continueOnError, format, binaryAsHex)
+			err := execBatchMode(queryist.Context, queryist.Queryist, input, continueOnError, format, binaryAsHex, nil)
 			if err != nil {
 				return sqlHandleVErrAndExitCode(queryist.Queryist, errhand.VerboseErrorFromError(err), usage)
 			}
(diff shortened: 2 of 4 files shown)


zachmu:
Hi @kevglynn, thanks for the contribution. Sorry for the delay in response!

I like this idea, but the implementation requires parsing the query string an additional time. We already parse all queries in given to `sql -q`, so this warning logic should take advantage of that fact.

Please make that change and mention me when you have, and I'll take another look for merging.

kevglynn:
@zachmu Thanks for the review! Good catch — I've refactored the implementation to eliminate the redundant parse.

**What changed:**

- Removed `isLoneDoltCheckoutCall()` and `warnIfLoneDoltCheckout()` which called `sqlparser.Parse()` separately
- Added a `batchExecInfo` struct that `execBatchMode` populates as it processes each already-parsed statement
- `execBatchMode` now accepts an optional `*batchExecInfo` parameter — when non-nil, it tracks statement count and whether any statement is a `CALL dolt_checkout(...)` using `isDoltCheckoutCall()` on the `sqlparser.Statement` that `ParseWithOptions` already produced
- After the loop, if `stmtCount != 1`, `isLoneDoltCheckout` is cleared — so multi-statement input never triggers the warning
- `queryMode` (the `-q` entry point) passes a `&batchExecInfo` and checks the result after execution
- Other callers (`Exec` batch path, `debug.go`) pass `nil` — zero behavior change for them

No extra parse. Same observable behavior. Unit tests and BATS tests still pass.

kevglynn:
@zachmu Rebased onto latest main — the refactored implementation (using `batchExecInfo` to piggyback on the existing parse in `execBatchMode`, no redundant parse) is ready for another look whenever you have a chance. Thanks!

zachmu:
Thank you for your contribution, @kevglynn! I carried your commit forward into https://github.com/dolthub/dolt/pull/11791, simplified the batch metadata to a returned struct with statement counting kept internal, updated the warning wording, and replaced the Go test with stronger Bats coverage. The tests now verify the exact warning on stderr and exercise multi-statement input through -q. All 25 SQL checkout tests and 7 batch tests pass. Closing this PR in favor of the replacement; your contribution is preserved in its commit history.

PR #11019, opened 2026-05-07
.github/actions/setup-go-toolchain: Make a common toolchain setup action which go test and bats test workflows can depend on.

reltuk:
This is responsible for installing Golang toolchain and getting ICU4C in the environment so it can be used.


coffeegoddd (approved):
LGTM

PR #11020, opened 2026-05-07
.github: Factor out build-dolt and setup-dolt-config actions. Use them from workflows that want dolt binaries, like bats and sql-server-integration tests.

reltuk:


coffeegoddd (approved):
LGTM!

PR #11023, opened 2026-05-08
[no-release-notes] Update future version string in version check tests

nicktobey:
The future has become the present.

To properly test the "version out of date" warning, we must invent a version number from even further in the future.

I look forward to breaking this test again in the future.


reltuk (approved):
LGTM!

PR #11030, opened 2026-05-11
Gate DELETE on dolt_conflicts_<t> with branch_control Write permission

tbantle22:
## Summary

- `ProllyConflictsTable.Deleter` returned its row deleter without any `branch_control` check. A user with only `Read` permission on a branch could `DELETE` conflict rows directly via `dolt_conflicts_<t>`. The artifact-map editor it constructs bypasses the source-table writer entirely, so the existing `WritableDoltTable.Deleter` `Write` check never fired transitively either.
- Mirror the pattern from `WritableDoltTable.Deleter` (`tables.go:973`): check `Permissions_Write` on entry to `ProllyConflictsTable.Deleter` and return a `sqlutil.NewStaticErrorEditor` on denial.
- Plumb `dsess.SqlDatabase` through `NewConflictsTable` and `newProllyConflictsTable` so the check has the right db handle. Only one production caller of `NewConflictsTable` exists (`database.go:578`) and no test callers, so the signature change is safe.

`UPDATE` on `dolt_conflicts_<t>` already inherits the source-table `Write` check via the eager `srcUpdater` construction in `newProllyConflictOurTableUpdater` — no change needed there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11032, opened 2026-05-11
Let merge-permission users write through dolt_conflicts_<t>

tbantle22:
## Summary

Stacks on top of #11030. Adds a branch-control carve-out so a user with only `Permissions_Merge` on a branch can resolve a merge that produced data conflicts by writing through `dolt_conflicts_<t>` and via `DOLT_CONFLICTS_RESOLVE`. Lets a PR reviewer on the SQL workbench finish a conflicting merge without being granted full write access on the target branch.

Admission rule applied at every entry point — **Write, OR (Merge AND MergeActive AND data conflicts exist)**:

- `ProllyConflictsTable.Updater` (`dtables/conflicts_tables_prolly.go`) — new gate via `checkConflictsTableWriteAccess`.
- `ProllyConflictsTable.Deleter` (same file) — replaces the bare `Write` gate from #11030 with the same helper.
- `DoDoltConflictsResolve` (`dprocedures/dolt_conflicts_resolve.go`) — new `dsess.CheckAccessOrMergeActive` helper; "data conflicts exist" scoped to the working root.

The `Updater` path's source-table `srcUpdater` is constructed eagerly inside `newProllyConflictOurTableUpdater`, so a merge-permission caller admitted at the outer gate would otherwise be rejected by the inner `WritableDoltTable.Updater` Write check. To bridge that delegation chain:

- New `dsess.ConflictsBypassMarker` keyed on `(db, branch, table)`. `prollyConflictOurTableUpdater` sets it on a derived context just for the source-table writer factory call.
- `WritableDoltTable.Updater` honors the marker as a Write-check bypass **only when** it matches the target table. A marker for `t1` cannot leak into a write on `t2`.

## Test plan

- [x] Extended `"Merge permission allows merge with conflicts"` (`branch_control_test.go`) with the full positive workflow: merge-only user runs `UPDATE dolt_conflicts_test` (write goes through to source table), `CALL DOLT_CONFLICTS_RESOLVE`, `CALL DOLT_COMMIT`. Source-table direct writes remain rejected.
- [x] New `"Merge permission allows DELETE from dolt_conflicts_<t> to resolve conflicts"` — exercises the DELETE path end-to-end.
- [x] New `"Merge permission does not allow conflicts writes outside an active merge"` — without an active merge, even a merge-permission user is rejected by `DOLT_CONFLICTS_RESOLVE` and by direct source-table writes.
- [x] `go test ./libraries/doltcore/sqle/enginetest/ -run "TestBranchControl"` passes.
- [x] `go vet ./libraries/doltcore/sqle/...` clean.

## Known limitation

When a merge-only user runs `UPDATE`/`DELETE` on `dolt_conflicts_<t>` with **zero matching rows** (e.g., empty conflicts table outside an active merge), the engine never invokes the writer's per-row methods, so the static error from the factory's permission check never surfaces. This is the same property that affects every `WritableDoltTable` write check — consistent with existing Dolt behavior. The negative test asserts via `DOLT_CONFLICTS_RESOLVE` and a direct source-table write instead, both of which run the gate regardless of matching-row count.

### DIFF
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_conflicts_resolve.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_conflicts_resolve.go
@@ -22,7 +22,6 @@ import (
 	"github.com/dolthub/go-mysql-server/sql"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
-	"github.com/dolthub/dolt/go/libraries/doltcore/branch_control"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb"
 	"github.com/dolthub/dolt/go/libraries/doltcore/doltdb/durable"
 	"github.com/dolthub/dolt/go/libraries/doltcore/env/actions"
@@ -375,7 +374,7 @@ func ResolveDataConflicts(ctx *sql.Context, dSess *dsess.DoltSession, root doltd
 }
 
 func DoDoltConflictsResolve(ctx *sql.Context, args []string) (int, error) {
-	if err := branch_control.CheckAccess(ctx, branch_control.Permissions_Write); err != nil {
+	if err := dsess.CheckAccessOrMergeActive(ctx); err != nil {
 		return 1, err
 	}
 	dbName := ctx.GetCurrentDatabase()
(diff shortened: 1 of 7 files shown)


reltuk (approved):
LGTM!

PR #11038, opened 2026-05-12
remove adapative encoding matrixes

zachmu:
Now that adaptive encoding is on by default, we don't need to double our CI testing efforts anymore.


reltuk (approved):
LGTM!

PR #11045, opened 2026-05-14
fix(sql): warn when dolt_checkout used as single -q statement

### LINKED ISSUE #6876: Warn users about running `dolt sql -q "call dolt_checkout(...);"`

superShen0916:
## Summary
Fixes #6876 - When `dolt sql -q "call dolt_checkout('branch');"` is run as a single statement, the branch change does not persist because each `-q` invocation uses its own session. This PR adds a warning to help users understand this behavior.

## Changes
- `go/cmd/dolt/commands/sql.go`: Added warning in `queryMode()` when query contains `dolt_checkout`

### DIFF
--- a/go/cmd/dolt/commands/sql.go
+++ b/go/cmd/dolt/commands/sql.go
@@ -23,6 +23,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -73,6 +74,8 @@ If a server is running for the database in question, then the query will go thro
 }
 
 var ErrMultipleDoltCfgDirs = errors.NewKind("multiple .doltcfg directories detected: '%s' and '%s'; pass one of the directories using option --doltcfg-dir")
+
+var doltCheckoutWarned sync.Once
 
 const (
 	QueryFlag             = "query"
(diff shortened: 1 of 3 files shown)


go/cmd/dolt/commands/sql.go
```diff
@@ -411,6 +411,10 @@ func queryMode(
 		return sqlHandleVErrAndExitCode(qryist, errhand.VerboseErrorFromError(err), usage)
 	}
 
+	if strings.Contains(strings.ToLower(query), "dolt_checkout") {
```
macneale4, line 417: It would be best to see this warning only once. Can you set a one time flag to ensure that's the case.

macneale4 (commented):
Thanks for the contribution! Please address the feedback and add a bats tests in the `sql.bats` file to verify that the warning shows once per invocation. Thanks again!

go/cmd/dolt/commands/sql.go
superShen0916: Added a Bats regression in `integration-tests/bats/sql-checkout.bats` that runs two `DOLT_CHECKOUT` statements in one `dolt sql -q` invocation and asserts the warning is emitted exactly once. The targeted Bats case passes on the updated branch.

superShen0916:
Closing this in favor of #11012, which addresses #6876 with parsed-statement-aware detection and avoids warning for valid multi-statement queries. I re-tested this branch on current main, and its old multi-`-q` regression no longer matches the current CLI parser. Thanks for the review.

PR #11057, opened 2026-05-15
Branch_control regression tests for more user-table write shapes

tbantle22:
Adds branch_control regression tests for the user-table write shapes that TestBranchControlBlocks wasn't already pinning; all were already correctly gated, so this is pure regression coverage with no behavior change.


reltuk (approved):
LGTM

Issue #11068, opened 2026-05-19
conjoin ENOSPC: panic + leftover partial nbs_table_* files cause unrecoverable disk-fill loop under supervised dolt-server


tim-white87:
**Summary**

When the dolt sql-server's chunk-store `conjoin` operation hits `ENOSPC` on `copy_file_range` while writing a new NBS table file (often into a configured backup remote), two things happen that together turn one disk-pressure event into an unrecoverable disk-fill loop:

1. The dolt server **panics** on the failed flush instead of returning the error to the caller. The panic propagates up through `store/nbs.(*journalWriter).flush → commitRootHashUnlocked → commitRootHash`.
2. The **partial NBS table file is left behind** in the backup remote (or wherever conjoin was writing). The next conjoin attempt creates a fresh partial file rather than reusing or cleaning up the previous one.

If the dolt server is running under any kind of supervisor that auto-restarts it on crash (which is a common deployment pattern — Gas City does this, but the same pattern exists for any service-managed dolt deployment), each restart triggers another conjoin attempt against the same low-disk filesystem, which fails the same way and leaves another orphan. The orphan rate observed in two independent reports has been **~2.8 GB/h to ~140 GB/h** until the filesystem hits 100%.

**Panic signature**

```
time="2026-05-19T01:42:14Z" level=warning msg="conjoin of database failed with error" error="write /...path.../.beads/backup/nbs_table_2015467144: copy_file_range: no space left on device" generation=new pkg=store.noms
panic: goroutine 70482693 [running]:
  runtime/debug.Stack()
  github.com/dolthub/dolt/go/libraries/utils/errors.Fatalf(...)
  github.com/dolthub/dolt/go/store/nbs.(*journalWriter).flush(...)
    go/store/nbs/journal_writer.go:544
  github.com/dolthub/dolt/go/store/nbs.(*journalWriter).commitRootHashUnlocked(...)
    go/store/nbs/journal_writer.go:462
  github.com/dolthub/dolt/go/store/nbs.(*journalWriter).commitRootHash(...)
    go/store/nbs/journal_writer.go:444
```

**Reproduction**

1. Configure a dolt server with a `file://` backup remote on the same filesystem as the live dolt data dir.
2. Fill the filesystem until free space drops below approximately `2 × largest_table_file` (conjoin needs that headroom to write its merged output).
3. Either trigger a conjoin manually or let a long-running write workload run; the server's internal auto-backup-sync will eventually call conjoin on the backup remote.
[...]

PR #11069, opened 2026-05-19
errors: surface ENOSPC instead of crashing under FatalBehaviorCrash

tim-white87:
## Summary

Refs #11068.

Disk-exhaustion (`syscall.ENOSPC`) returned from a write inside an NBS fatal path currently triggers `FatalBehaviorCrash`, which restarts the dolt sql-server. Under any supervisor that auto-restarts the server (systemd, container orchestrator, Gas City, etc.), the restart triggers another write against the still-full filesystem. That second attempt fails identically, can leave another partial `nbs_table_*` file behind in the backup remote, and the cycle repeats every 30s under common health-check cadences. Two independent reports in #11068 observed this consume **~60 GB of free disk overnight on one host and ~7 GB in 3 minutes on another**.

ENOSPC is operator-recoverable by freeing disk space — it is not a data-corruption or invariant-violation condition that requires the process to die. Returning the error to the caller is safe (every existing `Fatalf` callsite already returns or assigns the result, so they handle the error correctly) and prevents the supervisor-restart amplification cycle.

## Change

A single special-case in `Fatalf`: when behavior is `FatalBehaviorCrash` and any of `args` is an error chain containing `syscall.ENOSPC`, return the error rather than crashing. Other fatal conditions continue to crash as before. `FatalBehaviorError` mode is unchanged.

```go
func Fatalf(behavior FatalBehavior, msg string, args ...any) error {
    if behavior == FatalBehaviorCrash {
        if anyArgIsEnospc(args) {
            return fmt.Errorf("fatal error: "+msg, args...)
        }
        // ... existing crash path unchanged ...
    }
    ...
}
```

`anyArgIsEnospc` uses `errors.Is` against `syscall.ENOSPC`, so wrapped errors (the common case — an OS write returns ENOSPC and the caller wraps it before passing) are detected.

## Existing callsites that funnel through this

```
go/store/nbs/journal.go:424         UpdateGCGen
go/store/nbs/store.go:1673          closing table persister
go/store/nbs/file_manifest.go:466   renaming manifest
go/store/nbs/file_manifest.go:470   fsync manifest dir
go/store/nbs/journal_writer.go:471  syncing journal
go/store/nbs/journal_writer.go:544  writing to journal file   <- panic site in #11068
```

All six callsites either return or assign the `Fatalf` result, so each handles a returned error correctly.

## Test plan

- [x] `gofmt -l` clean
- [x] `go vet ./libraries/utils/errors/` clean
- [x] `go build ./libraries/utils/errors/ ./store/nbs/` succeeds
- [x] New unit tests cover the special-case branches:
  - Direct `syscall.ENOSPC` arg under `FatalBehaviorCrash` returns wrapped error
  - Wrapped `ENOSPC` (`fmt.Errorf("%w", syscall.ENOSPC)`) also returns
  - `FatalBehaviorError` mode unchanged
  - `anyArgIsEnospc` helper covered with empty / no-error / unrelated / direct / wrapped / multi-arg cases
- [x] Journal-related tests in `./store/nbs/` pass (`go test -run "Journal" ./store/nbs/`)

The crash-path-for-non-ENOSPC case isn't easily testable in-process without a subprocess fork (since it actually panics the test binary). The existing crash branch is preserved untouched; coverage there would need a separate test infrastructure pattern. Happy to add it if maintainers can point to an existing example.

## Not addressed by this PR

Per #11068, there is a related concern: when `conjoin` fails mid-write with ENOSPC, a partial `nbs_table_*` file is left behind in the target. This PR breaks the panic-restart amplification cycle, which means each ENOSPC produces at most one partial file rather than dozens, but it does not clean up that one file. The cleanup belongs in the conjoin write path (separate code, separate review surface) and should be addressed in a follow-up.

Similarly, the dolt sql-server's internal auto-backup-sync mechanism that triggers conjoin independent of any external scheduler is not changed here — that's an architectural question for the maintainers.

## Risk

Low. The change adds one branch to a single function, gated on an error-chain match for `syscall.ENOSPC`. The non-ENOSPC crash path is byte-for-byte unchanged. The `FatalBehaviorError` path is unchanged. All existing callsites of `Fatalf` already handle the returned-error case (verified by inspection — they return or assign).

### DIFF
--- a/go/libraries/utils/errors/panic.go
+++ b/go/libraries/utils/errors/panic.go
@@ -15,8 +15,10 @@
 package errors
 
 import (
+	stderrors "errors"
 	"fmt"
 	"runtime/debug"
+	"syscall"
 	"time"
 )
 
@@ -34,8 +36,20 @@ const (
 // be entering an unsafe state due to the encountered error. If |behavior| is
 // FatalBehaviorCrash, this function will never return. Otherwise, an error value is
 // returned, built with fmt.Errorf on |msg| and |args|.
+//
+// As a special case, even when |behavior| is FatalBehaviorCrash, if any of |args|
+// is an error whose chain contains syscall.ENOSPC, Fatalf returns the error
+// instead of crashing the process. Disk-exhaustion is operator-recoverable by
+// freeing space; crashing under ENOSPC and being auto-restarted by a supervisor
+// triggers another write attempt against the still-full filesystem, which
+// fails identically and can leave additional partial files behind. Surfacing
+// the error to the caller lets the failure be reported cleanly without the
+// recovery cycle amplifying it. See https://github.com/dolthub/dolt/issues/11068.
 func Fatalf(behavior FatalBehavior, msg string, args ...any) error {
 	if behavior == FatalBehaviorCrash {
+		if anyArgIsEnospc(args) {
+			return fmt.Errorf("fatal error: "+msg, args...)
+		}
 		stack := string(debug.Stack())
 		args := append([]any{stack}, args...)
 		go func() {
@@ -47,4 +61,16 @@ func Fatalf(behavior FatalBehavior, msg string, args ...any) error {
 	} else {
 		return fmt.Errorf("fatal error: "+msg, args...)
 	}
+}
+
+// anyArgIsEnospc reports whether any of |args| is an error whose unwrap chain
+// contains syscall.ENOSPC. Used by Fatalf to distinguish disk-exhaustion from
+// other fatal conditions.
+func anyArgIsEnospc(args []any) bool {
+	for _, a := range args {
+		if e, ok := a.(error); ok && stderrors.Is(e, syscall.ENOSPC) {
+			return true
+		}
+	}
+	return false
 }
--- a/go/libraries/utils/errors/panic_test.go
+++ b/go/libraries/utils/errors/panic_test.go
@@ -1,0 +1,88 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


reltuk:
No, this is not a correct approach. Dolt is correct in crashing when it receives errors from storage. There is no way to safely restore durability invariants without a crash once storage has violated its contract to store what you gave it.

tim-white87:
Thanks for the quick call on this — you're right, and I'd underweighted the durability-invariant argument when I wrote the PR.

To make sure I'm tracking it correctly: ENOSPC isn't "the disk is full and the next write will fail." It's "this write failed and dolt cannot determine from the error alone which previously-acknowledged writes are actually durable on disk vs. still in user-space buffers or page cache that won't be flushed." Continuing to operate would let dolt accept further commits that reference chunks dolt thinks are durable but aren't — silent corruption surfacing later. The panic forces a full restart and a journal rescan, which is the only way to recover the truth about what's actually on disk.

That's a much stronger correctness argument than "ENOSPC is recoverable, just bubble it up" — apologies for missing it.

The operational cascade we observed in [gastownhall/gascity#2158](https://github.com/gastownhall/gascity/issues/2158) (the supervisor's 30s health check auto-restarting dolt after every panic, triggering another write attempt against the still-full filesystem, leaking another partial `nbs_table_*` file per cycle) is the orchestration layer's bug, not dolt's. I've fixed that in [gastownhall/gascity#2347](https://github.com/gastownhall/gascity/pull/2347): the supervisor's recover step now inspects the dolt log tail for ENOSPC signatures and refuses to restart when it sees them. That's the correct layer for the policy.

Closing the loop on the issue I filed at [#11068](https://github.com/dolthub/dolt/issues/11068) — I'll add a correction comment there pointing back here so anyone landing on it gets the right framing. There's still a separate concern about partial `nbs_table_*` files left behind by `copy_file_range` aborting mid-write (an `unlink` on ENOSPC before returning), which is independent of the panic-vs-return question. Happy to file that as a focused follow-up if it'd be useful, but I don't want to spam your bug tracker.

Thanks again — saved me from shipping a regression into a production database engine.

PR #11135, opened 2026-06-01
fix: make findCommonCommit deterministic for criss-cross merges

### LINKED ISSUE #11042: dolt_merge_base is non-deterministic on multi-LCA criss-cross merges
## Summary

`dolt_merge_base(C, D)` returns different results on repeated invocations of the same scenario when `C` and `D` have multiple maximal common ancestors (the classic criss-cross merge case). Across 10 runs of the setup below, the function returned commit `a1` 6 times and commit `b1` 4 times — both are valid maximal LCAs, but `dolt_merge_base` arguably shouldn't be unstable.

Tested against `dolt version 1.88.0` (latest at time of writing).

## Repro

Initialize a fresh repo and run:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v INT);
INSERT INTO t VALUES (1, 10);
CALL dolt_add('-A');
CALL dolt_commit('-m', 'base', '--date', '2020-01-01T00:00:00');
CALL dolt_branch('A');
CALL dolt_branch('B');
CALL dolt_checkout('A');
INSERT INTO t VALUES (2, 20);
CALL dolt_add('-A');
CALL dolt_commit('-m', 'a1', '--date', '2030-01-01T00:00:00');
CALL dolt_checkout('B');
INSERT INTO t VALUES (3, 30);
CALL dolt_add('-A');
CALL dolt_commit('-m', 'b1', '--date', '2020-06-01T00:00:00');
CALL dolt_branch('C', 'A');
CALL dolt_branch('D', 'B');
CALL dolt_checkout('C');
CALL dolt_merge('B');
CALL dolt_checkout('D');
CALL dolt_merge('A');
```

Then:

```sql
SELECT message FROM dolt_log WHERE commit_hash = dolt_merge_base('C', 'D');
```

Repeat the SELECT 10 times in fresh `dolt sql -c` invocations (or in a loop in one session). Observed (in our 10-run sample):

- `a1` — 6 times
- `b1` — 4 times

## Topology

```
       base (2020-01-01)
       /  \
      A    B
      |    |
     a1    b1
 (2030-01)  (2020-06)
      |\\  //|
      | \\/  |
      |   X  |
      |  /\\ |
      | /  \\|
      C     D
   merge(B) merge(A)
```

Maximal common ancestors of `C` and `D` are `{a1, b1}` — neither is an ancestor of the other, and `base` is an ancestor of both. A correct `merge_base` implementation should pick one of `{a1, b1}` deterministically (e.g., timestamp-ordered, hash-ordered, or topological).

## Expected

`dolt_merge_base(C, D)` returns the same value on every call for a given pair of commits.

## Observed

Result varies across runs — appears to depend on traversal order which is not seeded deterministically.

## Discovery context

Surfaced while writing parity oracle tests in [doltlite#847](https://github.com/dolthub/doltlite/pull/847) (doltlite implements `dolt_merge_base` with a deterministic timestamp-then-hash tiebreaker and was failing the Dolt-parity assertion because Dolt's answer wasn't stable). The doltlite oracle now uses an "any maximal LCA is acceptable" helper for these criss-cross cases as a workaround, but the underlying determinism gap in Dolt remains.

## Suggested fix direction

After finding the set of maximal common ancestors, apply a deterministic tiebreaker. Options:

- Commit timestamp (max first), then commit-hash lex order as the second tier — what doltlite uses.
- Pure commit-hash lex order — simpler, but loses the "prefer most recent" property.
- Topological order over the commit graph — most "git-like" but more expensive.

toller892:
## Problem

`dolt_merge_base(C, D)` returns different results on repeated invocations when `C` and `D` have multiple maximal common ancestors (criss-cross merge topology). As reported in #11042, across 10 runs the function returned different commits — both valid maximal LCAs, but the result was unstable.

## Root Cause

`findCommonCommit()` in `go/store/datas/commit.go` converts one set of commits to a Go map and iterates over the other to find matches:

```go
for s, c := range as {
    if _, present := bs[s]; present {
        return c, true
    }
}
```

Go map iteration order is non-deterministic. When multiple common ancestors exist at the same height, the first match found varies between runs.

## Fix

Collect all common ancestors, sort by hash address, and return the first one deterministically:

```go
var common []*Commit
for s, c := range as {
    if _, present := bs[s]; present {
        common = append(common, c)
    }
}
sort.Slice(common, func(i, j int) bool {
    return common[i].Addr().Less(common[j].Addr())
})
return common[0], true
```

## Testing

- All existing `TestFindCommonAncestor` tests pass
- Added `TestFindCommonAncestorCrissCrossDeterministic` that creates a criss-cross merge topology and verifies the result is identical across 20 runs
- `go vet ./store/datas/` passes

Fixes #11042

### DIFF
--- a/go/store/datas/commit.go
+++ b/go/store/datas/commit.go
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"time"
 
 	flatbuffers "github.com/dolthub/flatbuffers/v23/go"
@@ -536,12 +537,21 @@ func findCommonCommit(a, b []*Commit) (*Commit, bool) {
 	}
 
 	as, bs := toAddrMap(a), toAddrMap(b)
+	var common []*Commit
 	for s, c := range as {
 		if _, present := bs[s]; present {
-			return c, true
+			common = append(common, c)
 		}
 	}
-	return nil, false
+	if len(common) == 0 {
+		return nil, false
+	}
+	// Sort by address to ensure deterministic selection when multiple
+	// common ancestors exist at the same height (e.g., criss-cross merges).
+	sort.Slice(common, func(i, j int) bool {
+		return common[i].Addr().Less(common[j].Addr())
+	})
+	return common[0], true
 }
 
 func firstError(l, r error) error {
--- a/go/store/datas/commit_test.go
+++ b/go/store/datas/commit_test.go
@@ -435,6 +435,62 @@ func TestFindCommonAncestor(t *testing.T) {
 	})
 }
 
+func TestFindCommonAncestorCrissCrossDeterministic(t *testing.T) {
+	// Criss-cross merge with two maximal common ancestors at the same height.
+	// Verifies that findCommonCommit returns deterministically.
+	//
+	//       base
+	//       /  \
+	//      A    B
+	//      |    |
+	//     a1    b1
+	//      |\  /|
+	//      | \/ |
+	//      | /\ |
+	//      |/  \|
+	//      C    D
+	//
+	// Maximal common ancestors of C and D: {a1, b1}
+	// findCommonCommit must pick the same one every time.
+
+	storage := &chunks.TestStorage{}
+	db := NewDatabase(storage.NewViewWithDefaultFormat()).(*database)
+	defer db.Close()
+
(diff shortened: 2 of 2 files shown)


zachmu (approved):

Issue #11196, opened 2026-06-11
GC kill-connections safepoint leaks connection slots: `dss_concurrent_connections` counts sessions whose client socket is already gone (pegs at `max_connections`)


julianknutsen:
# GC kill-connections safepoint leaks connection slots: `dss_concurrent_connections` counts sessions whose client socket is already gone (pegs at `max_connections`)

## Summary

A `dolt sql-server` GC cycle (auto-GC, or `CALL DOLT_GC()`) can leave
`dss_concurrent_connections` counting sessions whose **client TCP socket is
already gone**. Under sustained concurrent write load the gauge climbed to
exactly `max_connections` (4000) while the OS showed only ~68 live sockets to
the server port, and the server then rejected every new client:

```
max waiting connections reached. Client rejected. Increase server max_connections and back_log
```

The server could not be stopped cleanly — it hung inside `sqle/auto_gc` until
`SIGKILL`, with:

```
sqle/auto_gc: Attempt to auto GC database <db> failed with error: context canceled
```

Root cause (source analysis below, confirmed in the v2.1.6 source **and at
HEAD**): the per-connection accounting (`dss_concurrent_connections`, and the
listener's `max_connections` semaphore slot) is released **only when the
connection's handler goroutine returns**. A writer that is parked inside the
store waiting for GC (`nbs.waitForGC` → `gcCond.Wait()`) does not observe its
peer's TCP disconnect, because that wait only wakes on GC-complete /
cycle-advance / `ctx.Done()` — never on socket EOF. So the slot stays counted
for the whole remaining GC duration; and when the default kill-connections
safepoint's 10s backoff expires and GC **aborts**, those sessions are never
reaped at all.

We have a **fully-stock-client** reproduction (no custom MySQL-protocol code —
just `/usr/bin/mysql` + `SIGKILL`) that pins the prod signature: the gauge stays
at N while the N killed clients have **zero ESTABLISHED sockets**.

## Affected versions

- Reproduced on **`dolt 2.1.6`** (`go/cmd/dolt/doltversion/version.go:18` →
  `Version = "2.1.6"`).
- **Still present at HEAD** (`82e98ffc70`, checked 2026-06-11). The three
  files central to the mechanism are byte-identical between the `v2.1.6` tag and
  HEAD:
  - `go/libraries/doltcore/sqle/dprocedures/dolt_gc.go` — **identical**
    (`git diff v2.1.6..HEAD` empty). `var UseSessionAwareSafepointController =
[...]

PR #11199, opened 2026-06-11
Bound the session-aware GC pre-finalize safepoint wait (DOLT_GC_SAFEPOINT_MAX_WAIT)

julianknutsen:
## What

Adds an optional `DOLT_GC_SAFEPOINT_MAX_WAIT` (a Go duration, e.g. `5m`). When set and non-zero, the **session-aware** safepoint controller bounds how long `EstablishPreFinalizeSafepoint` waits for sessions to quiesce. On expiry the GC cycle aborts and is retried on the next trigger, instead of blocking indefinitely behind a session that never ends its command. Unset/zero preserves today's unbounded behavior exactly.

## Why — #11196

On a `sql-server` with auto-GC enabled (which forces the session-aware controller in `engine/sqlengine.go`), `EstablishPreFinalizeSafepoint` is `GCSafepointWaiter.Wait(ctx)` with no timeout: it waits for every session that was mid-command when the waiter was created to reach `SessionCommandEnd`. A session whose client peer has gone away but whose handler goroutine is wedged mid-command (e.g. blocked writing a result to a non-reading peer under a long `write_timeout`) never quiesces, so the auto-GC cycle stalls indefinitely. While stalled it holds `gcInProgress`, blocks every subsequent GC, and the per-connection accounting (released only when the handler goroutine returns) ratchets `dss_concurrent_connections` toward `max_connections` until new clients are rejected. That is the failure mode in #11196.

We reproduced the stall deterministically against `v2.1.6` with stock clients: with auto-GC on, a single `SELECT SLEEP(N)` session holding an outstanding command stalls `CALL DOLT_GC()`'s pre-finalize safepoint for the full N seconds (vs a ~6s baseline), and a mid-stall `dolt_thread_dump()` shows the GC goroutine parked at `GCSafepointWaiter.Wait` ← `sessionAwareSafepointController.EstablishPreFinalizeSafepoint`.

## Why it's safe

The pre-finalize safepoint runs **strictly before** `finalize()`/the chunk swap in `ValueStore.gc()`, so returning an error there aborts the cycle having collected nothing — the deferred `collector.EndGC` still runs and releases every parked session, exactly as on any other GC error path. The change is a no-op when the env var is unset (the default), so existing deployments are unaffected.

## Testing

`go/libraries/doltcore/sqle/dprocedures/dolt_gc_test.go` (white-box, modeled on `gcctx/gc_safepoint_controller_test.go`):

- `AbortsAfterMaxWait` — a never-quiescing session aborts with `context.DeadlineExceeded` at the bound rather than blocking forever.
- `UnboundedWhenUnset` — with no max wait, the safepoint still blocks until the session quiesces (original behavior preserved).
- `SucceedsWhenSessionsQuiesce` — a configured max wait does not interfere with a safepoint that establishes normally.

Passes under `-race`; `gofmt`/`go vet` clean.

## Scope / intent

This is offered as a **minimal reference mitigation** for the stall half of #11196 — **not** as the prescribed fix. We'll gladly defer to whatever shape you prefer (a config-file knob instead of/in addition to the env var, a different default, an escalation rather than a plain abort, etc.).

It deliberately does **not** address the deeper connection-accounting behavior also described in #11196 — that connection slots and `dss_concurrent_connections` are released only when the handler goroutine returns, and that the OK-result (DML) query path starts no dead-peer poller. That part lives in `go-mysql-server`/`vitess` and is yours to fix properly.

We are running this patch as a fork in our production environment to stop the recurring outages until an upstream fix lands.

### DIFF
--- a/go/libraries/doltcore/dconfig/envvars.go
+++ b/go/libraries/doltcore/dconfig/envvars.go
@@ -54,6 +54,14 @@ const (
 	// Will go away after session_aware is made default-and-only.
 	EnvGCSafepointControllerChoice = "DOLT_GC_SAFEPOINT_CONTROLLER_CHOICE"
 
+	// Optional Go duration (e.g. "5m"). When set and non-zero, bounds how long
+	// the session-aware GC safepoint will wait for all sessions to quiesce
+	// before the pre-finalize safepoint. On expiry the GC cycle is aborted
+	// (and retried on the next trigger) rather than blocking indefinitely
+	// behind a session that never ends its command. Unset/zero preserves the
+	// original unbounded wait.
+	EnvGCSafepointMaxWait = "DOLT_GC_SAFEPOINT_MAX_WAIT"
+
 	// Used for tests. If set, Dolt will error if it would rebuild a table's row data.
 	EnvAssertNoTableRewrite         = "DOLT_TEST_ASSERT_NO_TABLE_REWRITE"
 	EnvAssertNoInMemoryArchiveIndex = "DOLT_TEST_ASSERT_NO_IN_MEMORY_ARCHIVE_INDEX"
--- a/go/libraries/doltcore/sqle/dprocedures/dolt_gc.go
+++ b/go/libraries/doltcore/sqle/dprocedures/dolt_gc.go
@@ -51,10 +51,22 @@ func init() {
 			panic("Invalid value for " + dconfig.EnvGCSafepointControllerChoice + ". must be session_aware or kill_connections")
 		}
 	}
+	if raw := os.Getenv(dconfig.EnvGCSafepointMaxWait); raw != "" {
+		d, err := time.ParseDuration(raw)
+		if err != nil {
+			panic("Invalid value for " + dconfig.EnvGCSafepointMaxWait + ": " + err.Error())
+		}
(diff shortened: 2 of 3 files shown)


julianknutsen:
Closing/withdrawing this.

After capturing production goroutine dumps, the connection-pool exhaustion in #11196 turned out **not** to be the GC safepoint — @reltuk was right to doubt that the safepoint controller had anything to do with the leaked connections.

The actual root cause is a read-lock leak in `lestrrat-go/strftime`'s `Lookup` (pinned v1.0.4), triggered by `DATE_FORMAT`: it does `defer RLock()` instead of `RUnlock()`, so with no writer the leaked read count overflows the int32 `readerCount` and `RLock` deadlocks every `DATE_FORMAT` under load. Fixes:

- gms: dolthub/go-mysql-server#3586 (cache compiled formatters)
- dolt: #11201 (bump strftime to the fixed v1.2.0)

This bounded-pre-finalize-wait change isn't the fix for #11196, and you're already making `SessionAwareSafepointController` the default and adding staleness monitoring to DML/DDL, so I'm withdrawing it rather than leave a misleading PR open. Happy to revisit the bounded-wait idea separately if it's ever useful.

Issue #10866, opened 2026-04-13
Dolt shell should read input per character/token instead of per line


angelamayxie:
A lot of Dolt shell issues (such as #10860, #10861, #10862, possibly #10865) could be fixed if input was parsed per token (or per character, like in `sql_statement_scanner`) instead of per line.

codeaucafe:
Is that fine if I try to fix these 3 (maybe 4) issues as you suggested?

angelamayxie:
@codeaucafe Sure, the fix will be in the dolthub/ishell repo

codeaucafe:
Thank you @angelamayxie. You beat me to my next question haha.

angelamayxie:
@codeaucafe It would be nice to be able to reuse the code from [`sql_statement_scanner.go`](https://github.com/dolthub/dolt/blob/main/go/cmd/dolt/commands/sql_statement_scanner.go) but we might have to move things around to avoid a cyclical import

codeaucafe:
update: still working on this. Should have more time to get it reviewable next week. have made some changes, but haven't full tested everything

codeaucafe:
@angelamayxie I believe the two PRs are ready for review now.

ishell: dolthub/ishell#8
dolt: dolthub/dolt#11202

Issue #10862, opened 2026-04-13
Dolt shell does not ignore delimiters inside comments


angelamayxie:
Similar to #10861 

Dolt
```
tmp/main*> select /* ;
syntax error at position 11 near '/* '
tmp/main*> select -- ;
syntax error at position 11 near '-- '
```

MySQL
```
mysql> select /* ;
   /*> */ null;
+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)

mysql> select -- ;
    -> null;
+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)
```

PR #11202, opened 2026-06-12
dolthub/dolt#10866: fix(sql-shell): parse input per character instead of per line

### LINKED ISSUE #10860: Dolt shell has issues parsing multiple statements in the same line
dolt
```
tmp/main*> select null; select -- comment
        -> null;
syntax error at position 20 near 'select'
tmp/main*> select null; select null;
syntax error at position 20 near 'select'
```

MySQL
```
mysql> select null; select -- comment
+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)

    -> null;
+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)

mysql> select null; select null;
+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)

+------+
| NULL |
+------+
| NULL |
+------+
1 row in set (0.00 sec)
```

### LINKED ISSUE #10861: Dolt shell does not ignore delimiters inside quotes
Dolt
```
tmp/main*> select ';
syntax error at position 9 near 'select'
```

MySQL
```
mysql> select ';
    '> ';
+----+
| ;
 |
+----+
| ;
 |
+----+
1 row in set (0.00 sec)
```

### LINKED ISSUE #10865: Dolt shell expects more input after empty line whereas MySQL shell returns nothing
MySQL
```
mysql> 
mysql> 
```

Dolt
```
tmp/main*> 
        ->  
```

Behavior is the same when the line contains only whitespaces (MySQL also expects more input)

codeaucafe:
## Summary

The interactive `dolt sql` shell accumulated input one physical line at a time and decided "is this statement done?" by checking whether the trimmed line ended with `;`. That check has no awareness of SQL quotes, comments, or the active `DELIMITER`, causing four bugs reported in #10860–#10862 and #10865. This PR replaces the blind suffix check with dolt's existing per-character `StreamScanner` (already used by batch mode) and replaces the single-statement `Parse` call in the shell callback with a scanner-driven split-and-execute loop.

The shell-side completeness check is wired in via a new opt-in `IsComplete` callback added to the `dolthub/ishell` fork (PR: dolthub/ishell#8). `go/go.mod` contains a `replace` directive pointing at to PR's head commit: `061915e`. 

## Issues fixed

- [#10860](https://github.com/dolthub/dolt/issues/10860) — `select null; select null;` on one line gave `syntax error at position 20 near 'select'`. The accumulated buffer was fed to `sqlparser.Parse` once; that parser accepts only a single statement.
- [#10861](https://github.com/dolthub/dolt/issues/10861) — `select ';` was treated as complete because the line ends with `;`, even though that `;` is inside an open string literal.
- [#10862](https://github.com/dolthub/dolt/issues/10862) — `select /* ;` and `select -- ;` were treated as complete for the same reason; the `;` is inside an open block or line comment.
- [#10865](https://github.com/dolthub/dolt/issues/10865) — pressing Enter on an empty line dropped the shell into the `->` continuation prompt instead of re-issuing the primary `>` prompt.

### `go/cmd/dolt/commands/sql_statement_scanner.go`

- Add `NewStreamScannerWithDelimiter(r io.Reader, delimiter string)` constructor. The shell processes one command at a time, so the delimiter set by a prior `DELIMITER` command must be passed in rather than discovered in-band (as batch mode does when scanning a whole script). `NewStreamScanner` delegates to this and is unchanged for existing callers.
- Add `endedAtEOF bool` field on `StreamScanner`, reset at the top of each `Scan()` call and set when the input runs out before the active delimiter is matched.
- Add three accessors exposing parse state to callers: `InsideQuote()`, `InsideBlockComment()`, `EndedAtEOF()`.
- Add `IsShellInputComplete(text, delimiter string) bool`. Runs the scanner over the accumulated buffer and returns `false` when the input ends inside an open quote or block comment, or with a trailing unterminated statement. Blank/whitespace-only input returns `true` so an empty Enter does not trigger continuation mode (fixes #10865).

### `go/cmd/dolt/commands/sql.go`

- Set `IsComplete: IsShellInputComplete` in `UninterpretedConfig`. The ishell library calls this after each physical line with the cumulative buffer and the current delimiter; returning `true` stops accumulation. This is the single change that makes completeness detection SQL-aware (fixes #10861, #10862, #10865).
- Replace the single `sqlparser.Parse(query)` call in the `Uninterpreted` callback with a `NewStreamScannerWithDelimiter` split-and-execute loop. Each statement in the buffer is parsed and executed individually so multiple `;`-separated statements on one line each run and print their own result set (fixes #10860). Per-statement errors are printed and execution continues with the next statement, matching MySQL CLI semantics.

### Unit tests (`go/cmd/dolt/commands/sql_statement_scanner_test.go`)

- `TestIsShellInputComplete` — 22 cases covering each issue branch (multi-statement, open quote, open block/line comment, empty input) plus `DELIMITER` cases.
- `TestEndedAtEOFFlag` — 3 sub-tests verifying the `endedAtEOF` flag is set on EOF-terminated input and cleared on delimiter-terminated input.

### Integration tests (`integration-tests/bats/sql-shell.bats`)

Four new bats tests, one per issue, each driving a real interactive `dolt sql` session:

- `sql-shell: multiple statements on one line (#10860)`
- `sql-shell: unclosed quote continues input (#10861)`
- `sql-shell: unclosed comment continues input (#10862)`
- `sql-shell: empty line stays at primary prompt (#10865)`

## Merge Strategy
- merge https://github.com/dolthub/doltgresql/pull/2846
  - this will allow DoltgreSQL Integration tests in Dolt PR to pass
- After Dolt tests pass, merge ishell PR: https://github.com/dolthub/ishell/pull/8
- Merge this Dolt PR
- clean up replace statements in Doltgres and Dolt go.mod's

Fixes: #10860, #10861, #10862, #10865, #10866


go/go.sum
```diff
@@ -187,8 +187,8 @@ github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDk
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/cncf/xds/go v0.0.0-20251210132809-ee656c7534f5 h1:6xNmx7iTtyBRev0+D/Tv1FZd4SCg8axKApyNyRsAt/w=
 github.com/cncf/xds/go v0.0.0-20251210132809-ee656c7534f5/go.mod h1:KdCmV+x/BuvyMxRnYBlmVaq4OLiKW6iRQfvC62cvdkI=
-github.com/cockroachdb/apd/v3 v3.2.3 h1:4Zx+I3R35bFXMnltzmjP79i2cravE4jTRL6ps9Aux80=
-github.com/cockroachdb/apd/v3 v3.2.3/go.mod h1:klXJcjp+FffLTHlhIG69tezTDvdP065naDsHzKhYSqc=
+github.com/codeaucafe/ishell v0.0.0-20260607220657-061915e86568 h1:zAWHQAgb62ygwjQsDscI9btOt/lWEa5NbgMjKL2y1qM=
```
angelamayxie: this should be dolthub/ishell

angelamayxie (commented):
please take a look at the failing tests

go/go.mod
```diff
@@ -208,4 +208,6 @@ require (
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 )
 
+replace github.com/dolthub/ishell => github.com/codeaucafe/ishell v0.0.0-20260607220657-061915e86568
```
angelamayxie: remove this line. run `go get github/com/dolthub/ishell@[commit]` to update `go.mod` and `go mod tidy` to clean up `go.sum`

codeaucafe: I don't believe this works for forks/outside contributors like myself. I had to do this same thing when making changes in Doltgres that a Dolt Pr I made needed for it to work/be fixed. That's why I did it this way. I'll find those PRs later tonight to show u what I mean.

codeaucafe: forgot to add: but I'll also try that just to be sure again

codeaucafe:
@angelamayxie thank you for reviewing. the following CI always fail for my outside contributor PRs:
- Benchmark Mini Sysbench...
- Benchmark SQL Correctness...
- Check for correctness_approved label...
- Check for performance_approved label...
- Check formatting, Commuters....

`Tests Bats Unix Remote/Bats tests` is failing at the `Install Maven` step. For some reason this CI gate has been failing at this step for me for my last 3-4 contributions this year. I'm not entirely sure and never got a reason when I asked and looked into it. 

`Test Integration with DoltgreSQL/ test-integration...` is failing due to it requiring my  change to ishell to pass. I had this same issue with before with a change I made to dolt that required me to make use the replace directive pointing in DoltgreSQL in order for this DoltgreSQL integration test suite to pass. This is why I state previously that this replace directive is required for forked contributions like the ones I have to make as an outside contributor. @nicktobey helped me on that change. These are the two PRs for that for reference/example:
- https://github.com/dolthub/doltgresql/pull/2196
- https://github.com/dolthub/dolt/pull/10227

I made a PR the same way in Doltgres so we can get this integration test passing: https://github.com/dolthub/doltgresql/pull/2846

go/go.mod
angelamayxie: Okay, I think you're right this doesn't work because it's a forked repo, instead of a regular branch in the original repo.

go/cmd/dolt/commands/sql_statement_scanner.go
```diff
@@ -376,3 +390,47 @@ func (s *StreamScanner) seekDelimiter() (error, bool) {
...
+
+// IsShellInputComplete reports whether |text| forms one or more complete
+// statements terminated by |delimiter|. It is false when input ends inside an
+// open quote or block comment, or with a trailing unterminated statement;
+// whitespace-only input is complete (issue #10865).
+func IsShellInputComplete(text string, delimiter string) bool {
+	if strings.TrimSpace(text) == "" {
+		return true
+	}
+	sc := NewStreamScannerWithDelimiter(strings.NewReader(text), delimiter)
```
angelamayxie: Looking at how this is integrated with the ishell, it seems like for each line of the input, the entire input so far is scanned again. Correct me if I'm misunderstanding this, but that seems like if an input has n lines, the scanner will need to scan through (n^2)/2 lines.

codeaucafe: @angelamayxie you are correct. `IsShellInputComplete` is stateless, so ishell calls it after each line with the whole buffer and it rescans from the start every time, leading to n²/2 time (inefficiency). (Same root issue as your comment on the ishell PR on line 529)

I'll work on refactoring this. The idea is to make the completeness check incremental; carry scanner's quote/comment/delimiter state across lines and only consume each new line's bytes instead of re-scanning the whole buffer, which I believe makes the accumulation O(n). I think this requires factoring the per-character state machine out of seekDelimiter so the streaming check and the statement splitter can share it.

I plan to scope it to removing the quadratic behavior; the separate "scanned twice" point (completeness pass + execution pass) is a constant factor, so I was going hold off on refactoring that for now (or handle it in a follow up PR). How does that sound to you?

angelamayxie: I think we can avoid "scanning twice" in the ishell by reading the input per-character, similar to the sql scanner in Dolt. The sql scanner logic in Dolt can probably be refactored into the ishell to avoid a circular import.

codeaucafe: ok I'll work on that. thank you for help! 

just a small FYI that I may not get to it until Sunday or early next week because of the holiday weekend.

codeaucafe: @angelamayxie just a quick check before I start. To do the single pass, I'll move the SQL scanner (StreamScanner) into ishell so ishell reads the input per-character and emits complete statements like we discussed, and the dolt shell just executes them (dropping IsShellInputComplete and the second split scan). That matches what you suggested and avoids the circular import.

One thing I want to confirm too: StreamScanner isn't only used by the shell. execBatchMode (dolt sql < file.sql), filter-branch, and debug also use it. If I move it into ishell, those non-shell dolt paths would end up importing their SQL scanner from ishell. That is fine right?

angelamayxie: @codeaucafe I think that should be fine. Ideally the StreamScanner should just be its own repo that both dolt and ishell import from, but having it as a package in the ishell repo is probably the easiest right now.

codeaucafe: Got it. I'll work on the refactor this week/weekend. After it's merged I can work on splitting it out to its own repo in another pr/change (or at least make an issue for tracking)

angelamayxie: I think it's better to make an issue for tracking. We'd want the new repo to still be part of the dolthub organization and I'm not sure if someone outside the organization is able to do that.

codeaucafe: that is right. sorry, I was only thinking about the actual work not the repo setup 🤦‍♂️ (after this PR/change is merged). Happy to help write the code if y'all want help (after repo setup). I'll make the ticket after I wrap this change up.

codeaucafe:
FYI still working on PR refactor. Made the changes, but reviewing and fixing some issues I found.

codeaucafe:
@angelamayxie I believe this is ready for another review. Most of the new changes were done in ishell like we discussed:
                                                                                  
ishell (the meat) — PR #8                                                                                                                                              
                                                                                                                                                                         
- Relocated SQL StreamScanner into ishell so input can be scanned where the shell reads it — avoids a circular import.                  
- Rewrote uninterpreted read path for single streaming pass: added uninterpretedReader (adapts line-at-a-time readline into an io.Reader) and                   readUninterpretedCommand, replacing the old per-line readMultiLinesFunc predicate. Resolved the O(n²) re-scan and the double-scan issues. I wrote a couple unit tests and benchmark tests to verify this, but I wasn't sure if it was overkill to include it or not. If you think its worth I'll make that commit and push it up
- Added Context.Statements to give pre-scanned statements to consumers.                                                                                             
- Fixed Ctrl-D at continuation prompt to discard the partial statement.
- Cleanups: removed a no-op Delimiter() line, reordered a deferred-error check, trimmed stored statement text.
- (Uncommitted local: the scaling test + benchmark.)                                                                                                                   
                                                                                                                                                                         
dolt — PR #11202                                                                                                                                                       
                                                                                                                                                                         
- Deleted the local scanner, made shell handler consume c.Statements instead of re-scanning; repointed batch/debug/filter-branch to ishell.NewStreamScanner.       
- Kept the \edit path scanning its own editor output.                                                                                                                  
- Bumped go.mod/go.sum to the pushed ishell fork pseudo-version.                                                                                                       
- Tests: Ctrl-D discard test, large-input correctness test, tightened the quote-continuation assertion, plus the earlier per-issue tests (#10860/#10861/#10862/#10865).
                                                                                                                                                                         
doltgresql — PR #2846                                                                                                                                                  
                                                                                                                                                                         
- go.mod replace bump to the same ishell commit (950b93b)

codeaucafe:
FYI looked at the Test Bats Unix / Bats tests... CI tests failures and its 3 BATS tests failing but it doesn't appear related to my PR's changes

Issue #11210, opened 2026-06-15
3 Dolt JSON Bugs for large JSON payloads / parallel execution of JSON payloads.


tianhuil:
The following bugs are reproduced on CI tests in a [Github repo here](https://github.com/tianhuil/dolt-json-bug/blob/main/README.md) and [a 2nd Github repo](https://github.com/tianhuil/dolt-repro).

It affects at least Dolt version 2.1.7.

See @zachmu and @nicktobey 

## Bug 1: JSON column silently stores NULL for large values

When inserting JSON values > ~4004 bytes containing certain control characters into
a `JSON` column, Dolt silently stores `NULL` instead of the value. No error is
raised. The same value stored in a `LONGTEXT` column round-trips correctly
byte-for-byte.

### Trigger conditions

The bug requires **both** conditions:

1. **JSON input text > ~4004 bytes**
2. **Value contains control characters that `JSON.stringify` serializes as `\uXXXX` (6-byte escape sequences)**

These control characters trigger the bug (no short escape → `\uXXXX`):
- U+0000–U+0007 (NUL through BEL)
- U+000B (VT, vertical tab)
- U+000E–U+001F

These do **not** trigger (short escape or literal):
- U+0008 (BS) → `\b` (2 bytes)
- U+0009 (TAB) → `\t` (2 bytes)
- U+000A (LF) → `\n` (2 bytes)
- U+000C (FF) → `\f` (2 bytes)
- U+000D (CR) → `\r` (2 bytes)
- U+007F (DEL) → literal (1 byte)
- Normal ASCII / UTF-8 → literal (1 byte each)

### Minimum reproduction

```json
{"d":"\u000b\u000b\u000b...666 times...\u000b"}
```

Total JSON input = 4004 bytes. A single `\u000b` in a 50KB string also triggers.
But 100KB of plain `x` with no escapes works fine.

### Hypothesis

Dolt's JSON serializer allocates an output buffer proportional to the input
string length. Control characters needing `\uXXXX` escaping expand from 1 byte
(internal) to 6 bytes (JSON output). The serializer writes past the end of the
allocated buffer, corrupting internal state and causing a silent NULL. Characters
with shorter escapes or no expansion fit within the original buffer.

---

## Bug 2: JSON_VALID returns 0 for LONGTEXT values > 2040 bytes

`JSON_VALID()` returns `0` for valid JSON stored in `LONGTEXT` columns when the
[...]

PR #11211, opened 2026-06-15
test: reproduce JSON bugs from issue #11210 as failing tests

tianhuil:
Reproduces all three bugs from GitHub issue #11210 as failing unit tests:

**Bug 1** — Large JSON values with control characters (U+0000–U+001F) are silently stored as NULL or become invalid when the serialized representation exceeds ~4004 bytes. Tests verify that JSON_TYPE, JSON_LENGTH, and JSON_UNQUOTE(JSON_EXTRACT) all fail.

**Bug 2** — JSON_VALID() returns 0 for valid JSON strings stored in LONGTEXT columns without a primary key when the value exceeds ~2040 bytes.

**Bug 3** — Concurrent JSON reads return corrupted data due to a data race in unescapeHTMLCodepoints() (blob_builder.go), which modifies the backing byte slice in place without synchronization.

All three tests pass on the expected buggy behavior (they fail on the current codebase, demonstrating the bugs).

### DIFF
--- a/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
+++ b/go/libraries/doltcore/sqle/enginetest/dolt_engine_test.go
@@ -1082,6 +1082,11 @@ func TestLargeJsonObjects(t *testing.T) {
 	RunLargeJsonObjectsTest(t, harness)
 }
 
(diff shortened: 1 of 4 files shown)


timsehn:
This was covered in our own PR.

PR #11248, opened 2026-06-24
Add a global credential helper for remotesapi

### LINKED ISSUE #11244: Support refreshable custom gRPC metadata for remotes
We run a Dolt `remotesapi` endpoint behind an identity-aware proxy. Both layers require authentication:

- The proxy expects a short-lived bearer token in `Proxy-Authorization`.
- Dolt expects its normal credentials in `Authorization`.

Native gRPC over HTTPS works here. The problem is that the Dolt client has no supported way to attach the second header while preserving its own authentication.

Our current workaround is a local HTTP/2 proxy that obtains and refreshes the identity token, injects `Proxy-Authorization`, and forwards the gRPC stream unchanged. It works, but adds another process and more operational complexity.

Would the maintainers be open to a credential-helper mechanism for additional gRPC metadata? For example, Dolt could invoke a configured command that returns headers and an expiration time, then refresh them when necessary. Dolt would continue managing its own `Authorization` header.

Important properties would be:

- Scoped to a specific remote
- Works for unary and streaming RPCs
- Supports expiring credentials
- Does not persist or log returned values
- Cannot accidentally override Dolt's own authentication

A static `--header` option would help some cases, but would not be sufficient for short-lived tokens.

I'm happy to draft a PR once we agree on the configuration and credential-provider shape. Is `PerRPCCredentials` the right extension point, or is there an existing mechanism I've missed?

I don't want to just throw up some AI slop without checking first :)

iamthebot:
## Why

Some `remotesapi` deployments sit behind an identity-aware proxy or gateway which needs its own short-lived credential. In that setup there are two separate auth layers:

- the gateway expects something like `Proxy-Authorization: Bearer ...`
- Dolt still expects its normal `Authorization` metadata

The gRPC transport itself already works. The missing piece is a supported way to add the gateway header without replacing Dolt's own credentials.

Our current workaround is a local HTTP/2 proxy which obtains the gateway token, forwards the full gRPC stream, and injects the additional header. It works, but it is a lot of transport machinery for what is ultimately a credential lookup.

This PR adds a global credential helper for `remotesapi` interactions so Dolt can make that request directly.

## How it works

The helper is configured globally:

```bash
dolt config --global --add remotesapi.credential_helper /path/to/helper
```

Dolt uses the Bazel credential helper protocol. It invokes `<helper> get` and sends only the canonical origin on stdin:

```json
{"uri":"https://example.com:443"}
```

The helper returns additive headers and an optional expiration:

```json
{
  "headers": {
    "Proxy-Authorization": ["Bearer TOKEN"]
  },
  "expires": "2026-06-24T20:00:00Z"
}
```

Those headers are added to unary and streaming gRPC calls while Dolt's existing per-RPC credentials remain in place. They are also added to HTTP requests which match the same origin. They are deliberately not forwarded to signed object-store URLs returned by the remote.

Responses are cached until shortly before expiration. A response without an expiration is used once.

## Why global

The executable is read directly from global config, not from the repository config or per-remote parameters. This keeps repository contents and `dolt_remote("add", ...)` from selecting a local binary to execute.

The helper still sees the destination origin and is responsible for returning credentials only for destinations it trusts. This allows one global helper to multiplex safely, for example by using a strict host allowlist or minting audience-bound credentials.

Helper headers are additive. Dolt rejects attempts to replace its `Authorization` header or set transport-controlled headers.

## Validation

- A real gRPC request carries both Dolt's Basic `Authorization` and the helper's `Proxy-Authorization`
- Unary and streaming RPCs are covered
- Expiration, refresh, expired credentials, header validation, and concurrent access are covered
- Same-origin HTTP requests receive helper headers
- Signed URLs on another origin do not receive helper headers
- Repository-local helper configuration is ignored
- The Bazel-compatible subprocess request and response are exercised with an executable test helper
- `go test -race ./libraries/doltcore/credentialhelper`
- `go test ./libraries/doltcore/env`
- `go test ./cmd/dolt/commands -run '^TestConfig'`
- `go test ./libraries/doltcore/dbfactory ./libraries/utils/config`
- `go vet ./libraries/doltcore/credentialhelper`
- `go run ./utils/copyrightshdrs`
- `./utils/repofmt/check_fmt.sh`
- Full `dolt` CLI build

I also ran the full `cmd/dolt/commands` package. Its only failure locally was the signed-commit test because `gpg` is not installed.

Closes #11244.

## AI disclosure

I used OpenAI Codex extensively (using an undisclosed model) to investigate the existing credential path, implement the change, and write tests and documentation. I reviewed the resulting design and code, ran the validation above, and take responsibility for the contribution.

### DIFF
--- a/go/cmd/dolt/commands/config.go
+++ b/go/cmd/dolt/commands/config.go
@@ -67,7 +67,19 @@ Valid configuration variables:
 
 	- remotes.default_port - sets default port for authenticating with doltremoteapi.
 
+	- remotesapi.credential_helper - sets a global credential helper for remotesapi requests. This option is ignored in repository local config.
+
 	- push.autoSetupRemote - if set to "true" assume --set-upstream on default push when no upstream tracking exists for the current branch.
+
+Credential helpers use the Bazel credential helper protocol. Dolt invokes the configured executable with the {{.EmphasisLeft}}get{{.EmphasisRight}} argument and sends the canonical remotesapi origin on stdin:
+
+	{"uri":"https://example.com:443"}
+
+The helper returns additive HTTP headers and, optionally, an RFC 3339 expiration time:
+
+	{"headers":{"Proxy-Authorization":["Bearer TOKEN"]},"expires":"2026-06-24T20:00:00Z"}
+
+Dolt caches headers until shortly before their expiration. Responses without an expiration are used once. The helper is responsible for returning credentials only for origins it trusts. It cannot replace Dolt's own Authorization header or transport-controlled headers.
 `,
 
 	Synopsis: []string{
--- a/go/libraries/doltcore/credentialhelper/credential_helper.go
+++ b/go/libraries/doltcore/credentialhelper/credential_helper.go
@@ -1,0 +1,353 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package credentialhelper
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	"golang.org/x/net/http/httpguts"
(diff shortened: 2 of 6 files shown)


timsehn:
Some of these are async. They look failed until a remote job finishes.

Some of them look like GitHub flakes with the install process. @reltuk is working another customer sissue but he'll probably get to this today.

timsehn:
Actually, if I remember correctly, the async jobs can't publish to outside contributor PRs. A quick inspection seems to say this passed our test suite.

reltuk (approved):
LGTM!

reltuk:
@iamthebot This PR looked great. Thanks for the due diligence and the contribution. I've merged it to main and it will be available in the next Dolt release. We will probably make further passes on some integration tests and stuff, but this is a great start.

bpf120:
@iamthebot, thank you for alerting us to this issue and getting it closed. :)  We'd love to learn more about your Dolt use case sometime. Feel free to [email me](mailto:brianf@dolthub.com) or swing by our [Discord](https://discord.gg/gqr7K4VNKe) if you are able to share.

PR #11274, opened 2026-07-08
Fix unreachable dolt-login hint on PermissionDenied push errors

### LINKED ISSUE #783: Prompt user to run "dolt login" on PermissionDenied error on "dolt push"
This at least eliminates the case where they have not yet authenticated. We should also indicate that if they have already authenticated some more digging might be needed.

kevglynn:
## Summary

- The hint text in `handlePushError` ("have you logged into DoltHub using 'dolt login'?") has been unreachable since `PushToRemoteBranch` started wrapping `ErrUnknownPushErr` — the sentinel comparison used `switch`/`==` on a wrapped error, and the gRPC status was stringified so `status.FromError` couldn't see `codes.PermissionDenied`
- Also fixes the `RpcError` type assertion in the same function, which suffered the same wrapping issue (the `FullDetails()` code path was similarly dead)

## Changes

- `go/libraries/doltcore/env/actions/remotes.go`: Changed `fmt.Errorf("%w; %s", ErrUnknownPushErr, err.Error())` to `fmt.Errorf("%w: %w", ErrUnknownPushErr, err)` — preserves both the sentinel and the original error (including `RpcError` with its captured gRPC status) in the chain
- `go/cmd/dolt/commands/push.go`: Replaced `switch err` / `err.(*RpcError)` with `errors.Is` / `errors.As`; extracted `isPermissionDeniedErr` helper that checks the `RpcError` status via `errors.As`, falling back to error text matching
- `go/cmd/dolt/commands/push_test.go`: 6 test cases for `isPermissionDeniedErr` covering wrapped `RpcError` with PermissionDenied, Internal, and Unauthenticated codes; plain text matching; non-matching errors; nil

## Test plan

- [x] `go test -run TestIsPermissionDeniedErr ./cmd/dolt/commands/ -v` — all 6 cases pass
- [x] `go test ./cmd/dolt/commands/` — full package passes
- [x] `go test ./libraries/doltcore/env/actions/` — passes (error format change)
- [x] `go test ./libraries/doltcore/remotestorage/` — passes
- [x] `go vet ./cmd/dolt/commands/` — clean
- [ ] CI

Closes #783

Made with [Cursor](https://cursor.com)

### DIFF
--- a/go/cmd/dolt/commands/push.go
+++ b/go/cmd/dolt/commands/push.go
@@ -26,7 +26,6 @@ import (
 	"github.com/gocraft/dbr/v2"
 	"github.com/gocraft/dbr/v2/dialect"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/dolthub/dolt/go/cmd/dolt/cli"
 	"github.com/dolthub/dolt/go/cmd/dolt/errhand"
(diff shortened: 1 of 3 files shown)


elianddb:
Hi @kevglynn, using `Contains` to detect the specific errors can break easily if this error changes in the future. Also, can you reduce the description above to a gist paragraph and then a list of bullet points with the specific changes to the code. It'll help the review process and fit into our release notes better.

elianddb:
Closed no activity.

PR #11281, opened 2026-07-09
go: sqle: Fix concurrent-clone deadlock between two remotesapi sql-servers

### LINKED ISSUE #11280: Deadlock when two sql-servers `CALL DOLT_CLONE` from each other's `--remotesapi-port` concurrently
## What happened

Two `dolt sql-server` instances, each serving a `--remotesapi-port`, deadlock permanently when they each `CALL DOLT_CLONE` from the other's remotesapi at the same time. The databases are trivial (a single row), so a clone should take milliseconds. Instead both `CALL DOLT_CLONE` statements hang forever.

The hang is server-side and not client-recoverable: killing a clone's MySQL client does not cancel the server-side clone or free the other server. Only restarting a server recovers it.

Notes:
- Concurrent `DOLT_PULL` does **not** deadlock — only the initial `DOLT_CLONE`.
- Sequential clones (A clones from B, completes, then B clones from A) are fine.

## Version

Reproduced on `dolt v2.1.10` and on current `main` (`d1da417d27`, 2026-07-08).

## Minimal reproduction

```bash
#!/usr/bin/env bash
set -uo pipefail
DOLT="${DOLT:-dolt}"
W="$(mktemp -d)"; cd "$W"; mkdir a b
"$DOLT" sql-server --data-dir a -H 127.0.0.1 -P 3401 --remotesapi-port 3501 --loglevel=warning >a.log 2>&1 </dev/null & APID=$!
"$DOLT" sql-server --data-dir b -H 127.0.0.1 -P 3402 --remotesapi-port 3502 --loglevel=warning >b.log 2>&1 </dev/null & BPID=$!
sql(){ "$DOLT" --host 127.0.0.1 --port "$1" --user root --password '' --no-tls sql -q "$2"; }
for i in $(seq 1 40); do (: </dev/tcp/127.0.0.1/3401) 2>/dev/null && (: </dev/tcp/127.0.0.1/3402) 2>/dev/null && break; sleep 0.5; done; sleep 1
sql 3401 "CREATE DATABASE da; USE da; CREATE TABLE t(i int primary key); INSERT INTO t VALUES(1); CALL DOLT_COMMIT('-Am','seed');" >/dev/null 2>&1
sql 3402 "CREATE DATABASE db; USE db; CREATE TABLE t(i int primary key); INSERT INTO t VALUES(2); CALL DOLT_COMMIT('-Am','seed');" >/dev/null 2>&1
sql 3401 "CALL DOLT_CLONE('http://127.0.0.1:3502/db','db')" & C1=$!
sql 3402 "CALL DOLT_CLONE('http://127.0.0.1:3501/da','da')" & C2=$!
sleep 8
if kill -0 $C1 2>/dev/null && kill -0 $C2 2>/dev/null; then echo "DEADLOCK: both clones still running after 8s"; else echo "no deadlock this run (retry; timing-dependent)"; fi
kill -9 $C1 $C2 $APID $BPID 2>/dev/null; cd /; rm -rf "$W"
```

It is timing-dependent (both clones must be fetching at once); run it a few times if a given run prints "no deadlock".

## Root cause

`DoltDatabaseProvider.CloneDatabaseFromRemote` (`go/libraries/doltcore/sqle/database_provider.go`) takes the provider's `RWMutex` **write lock** and holds it (`defer p.mu.Unlock()`) across the entire clone, including the network fetch of all chunks from the remote. The code already carries a TODO noting this:

```go
// TODO: This holds the database provider lock across the entire duration of
// the clone, which is much too long to hold this lock.
```

Meanwhile, when the server serves a `--remotesapi-port`, its gRPC `ChunkStoreService` handlers resolve the requested database through `remotesrvStore.Get` → `DoltDatabaseProvider.Database` → `SessionDatabase`, which needs a **read lock** on the same provider mutex.

So when A and B clone from each other simultaneously:

- A's clone holds A's write lock and blocks fetching from B's remotesapi.
- B's remotesapi handler (answering A) blocks trying to take B's read lock, because B's clone holds B's write lock.
- B's clone holds B's write lock and blocks fetching from A's remotesapi.
- A's remotesapi handler (answering B) blocks trying to take A's read lock, because A's clone holds A's write lock.

A dependency cycle → permanent deadlock. `DOLT_PULL` is immune because it acts on an already-registered database and does not hold the provider write lock across its fetch. Because Go's `sync.RWMutex` is write-preferring, the waiting `RLock` can never proceed while the writer holds the lock.

### Goroutine evidence (`kill -QUIT` on both wedged servers)

Server A — the clone holding the write lock, parked in the fetch:

```
goroutine [sync.WaitGroup.Wait]:
  store/datas/pull.WithDiscardingStatsCh(...)
  sqle.(*DoltDatabaseProvider).cloneDatabaseFromRemote(...)
  sqle.(*DoltDatabaseProvider).CloneDatabaseFromRemote(...)   // holds p.mu.Lock()
  sqle/dprocedures.doltClone(...)
```

Server A — its own remotesapi handler starved on the read lock:

```
goroutine [sync.RWMutex.RLock]:
  sqle.(*DoltDatabaseProvider).SessionDatabase(...)           // wants p.mu.RLock()
  sqle.(*DoltDatabaseProvider).Database(...)
  sqle.remotesrvStore.Get(...)
  remotesrv.(*RemoteChunkStore).getOrCreateStore(...)
  remotesrv.(*RemoteChunkStore).GetRepoMetadata(...)
```

Server B shows the mirror image (its clone blocked in the client-side `GetRepoMetadata`, its own remotesapi handler parked on `SessionDatabase`'s `RLock`). A third victim on server A — the stats worker (`AllDatabases` → `rlockAwaitingEmptyDeletingDatabases`) — is also parked on `RLock`, starved by the same held write lock.

### The cycle

```
A.clone (holds A.mu write) ── waits on ──▶ B.remotesapi.GetRepoMetadata
                                                   │ needs B.mu read
                                                   ▼
                                          B.clone (holds B.mu write)
                                                   │ waits on
                                                   ▼
                                          A.remotesapi.GetRepoMetadata
                                                   │ needs A.mu read
                                                   ▼
                                          blocked on A.mu write (held by A.clone)  ── cycle closed
```

## Suggested fix

Do not hold the provider write lock during the clone's network fetch. Take the lock only briefly to reserve the new database name (with the existing existence checks), release it for the fetch, then re-acquire it only to register the finished database. A small "reserved names" set keeps concurrent create/clone/undrop of the same name mutually exclusive without blocking unrelated readers (including the server's own remotesapi handler).

I have a fix implementing this and will open a PR referencing this issue. It includes a bats regression test that starts two remotesapi servers cloning from each other concurrently and asserts no deadlock.

mccraigmccraig:
`DoltDatabaseProvider.CloneDatabaseFromRemote` held the provider `RWMutex` write lock across the entire clone, including the network fetch of all chunks from the remote. When a server serves a `--remotesapi-port`, that fetch hits a peer's remotesapi whose handler needs a **read** lock on this same provider (via `SessionDatabase`). Two sql-servers that `CALL DOLT_CLONE` from each other's remotesapi at the same time therefore deadlock: each holds its own provider write lock while blocked on the peer, and neither peer can take the read lock needed to answer. `DOLT_PULL` is immune because it operates on an already-registered database and does not hold the write lock across its fetch. This is the lock the existing `// TODO: This holds the database provider lock across the entire duration of the clone ...` was warning about.

The fix narrows the lock scope so the network fetch does not run under the provider write lock:

- Reserve the database name under a brief lock acquisition (a new `creatingDatabases` set), release the lock for the fetch, then re-acquire it only to register the finished database (`registerNewDatabase` still requires the lock).
- The reservation keeps concurrent `CREATE DATABASE` / `dolt_clone` / `undrop` of the same name mutually exclusive, so those races behave as before even though the lock is dropped for the fetch.
- A shared `databaseNameUnavailableLocked` helper performs the availability checks for all creation paths and preserves the interrupted-create in-progress marker behaviour (`ErrIncompleteDatabaseDir` / `dbfactory.MarkDatabaseInProgress`). The in-memory reservation and the on-disk marker are complementary: the marker guards crash recovery across processes, while `creatingDatabases` serialises concurrent in-process operations while the lock is dropped.
- Names in `creatingDatabases` and `deletingDatabases` are normalised with `formatDbMapKeyName` (Dolt database names are case-insensitive), matching how `p.databases` is keyed. Unlike `deletingDatabases`, `creatingDatabases` deliberately does **not** gate database enumeration: an in-progress clone is simply invisible until it registers, so enumeration / remotesapi reads never block on it.

Tests:

- `integration-tests/bats/sql-server-remotesrv.bats`: a regression test that starts two `--remotesapi-port` servers cloning from each other concurrently. It hangs (clones time out) on the buggy build and passes on the fixed build; it also fails if a clone errors for any non-deadlock reason.
- `go/libraries/doltcore/sqle/database_provider_test.go`: `TestCreatingDatabaseReservation` covering the reservation semantics — conflict, case-insensitivity across create/clone/undrop, release, and that a reservation does not block `AllDatabases`.

Verified with `go test -race ./libraries/doltcore/sqle/` and the new bats test; `gofmt` / `goimports` clean.

Fixes #11280

### DIFF
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -79,6 +79,29 @@ type DoltDatabaseProvider struct {
 	// when accessed.
 	deletingDatabases map[string]struct{}
 
+	// Databases named in creatingDatabases are currently being created by a
+	// clone that is fetching its data from a remote. The provider lock is
+	// deliberately NOT held for the duration of that network fetch (see
+	// CloneDatabaseFromRemote), so this set is used to reserve the name and
+	// prevent concurrent CREATE DATABASE / dolt_clone / undrop operations from
+	// racing on the same name while the clone is in progress.
+	//
+	// Names are stored normalized via formatDbMapKeyName, matching p.databases
+	// and deletingDatabases (Dolt database names are case-insensitive).
+	//
+	// Unlike deletingDatabases, this set deliberately does NOT gate database
+	// enumeration (see rlockAwaitingEmptyDeletingDatabases). A clone builds its
+	// database directory in place and non-atomically, so a partially-populated
+	// directory exists on disk for the duration of the fetch. That would be a
+	// hazard only if something enumerated databases by scanning the data
+	// directory at runtime, and nothing does: in a running server all
+	// enumeration goes through the in-memory p.databases map (AllDatabases /
+	// DoltDatabases), the remotesapi resolves databases by exact name, and the
+	// only data-directory scan (MultiEnvForDirectory) runs at startup, before
+	// any clone can be in flight. An in-progress clone is therefore simply
+	// invisible until it finishes and registers under the write lock.
+	creatingDatabases map[string]struct{}
+
 	txLocks keymutex.Keymutex
 
 	defaultBranch     string
@@ -208,6 +231,7 @@ func NewDoltDatabaseProviderWithDatabases(defaultBranch string, fs filesys.Files
 		dbLocations:            dbLocations,
 		databases:              dbs,
 		deletingDatabases:      make(map[string]struct{}),
+		creatingDatabases:      make(map[string]struct{}),
 		functions:              funcs,
 		tableFunctions:         tableFuncs,
 		externalProcedures:     externalProcedures,
@@ -697,17 +721,8 @@ func (p *DoltDatabaseProvider) CreateCollatedDatabase(ctx *sql.Context, name str
 		}
 	}()
 
-	if _, ok := p.deletingDatabases[name]; ok {
-		return sql.ErrDatabaseExists.New(name)
-	}
-	exists, isDir := p.fs.Exists(name)
-	if exists && isDir {
-		if subFs, ferr := p.fs.WithWorkingDir(name); ferr == nil && env.IsIncompleteDatabaseDir(subFs) {
-			return NewErrIncompleteDatabaseDir(name)
-		}
-		return sql.ErrDatabaseExists.New(name)
-	} else if exists {
-		return fmt.Errorf("Cannot create DB, file exists at %s", name)
+	if err = p.databaseNameUnavailableLocked(name, true /* checkDisk */); err != nil {
+		return err
 	}
 
 	err = p.fs.MkDirs(name)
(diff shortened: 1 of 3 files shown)


reltuk:
This is awesome. Thank you for the bug report and thank you for the contribution. I will make a secondary pass on comments and maybe where the testing lands, but in general this approach seems totally reasonable and I had no blocking feedback about the implementation itself.

reltuk:
If you're curious, follow up PR is here: #11281. It just slightly reworks some comments and function names and moves the test around.

mccraigmccraig:
excellent news! I'm using dolt in a local first single master, many follows mesh pattern, and this was quite an issue... now I can get back on official releases 😁

PR #11291, opened 2026-07-12
Handle `\c` in the interactive SQL shell

### LINKED ISSUE #10867: Dolt shell doesn't handle `\c`
Dolt
```
tmp/main*> hello \c
        ->  
```

MySQL
```
mysql> hello \c
mysql> 
```

vishnujayvel:
## What

Fixes #10867 — the interactive `dolt sql` shell doesn't handle the MySQL client's `\c` escape (cancel/clear the statement currently being entered). A line ending in `\c` was folded into the multi-line buffer as ordinary text, so the next line concatenated onto it and produced a syntax error near the stray backslash instead of discarding the input.

## What changed

- `go/cmd/dolt/commands/sql.go` — register `\c` alongside the existing `\g`/`\G` in `SpecialTerminators`, and in the shell callback, when the accumulated query ends with `\c`, discard the buffered input (no execute, no history entry) and reset the prompt via `postCommandUpdate`. This mirrors the `mysql` CLI's behavior.
- `integration-tests/bats/sql-shell.bats` + a new `integration-tests/bats/sql-shell-clear-statement.expect` — interactive PTY coverage: a partial statement followed by `\c` is cancelled, and a subsequent statement runs standalone.

## Verification

- Built with the repo's icu4c CGO flags; `go build ./cmd/dolt/...` and `go vet` clean.
- Live PTY repro: on an unpatched build the bug reproduces (syntax error near the stray `\`); with the patch, `\c` discards the buffer and the next statement executes on its own. Negative control on the unpatched binary confirmed the repro.

## Note on the in-flight shell rework (#11202)

I see #11202 is reworking the shell's input handling to a per-character `StreamScanner` (fixing #10860/#10861/#10862/#10865/#10866). That PR doesn't cover `\c` (#10867), so this change is complementary — but it does touch the same `Uninterpreted` callback, so the two will likely conflict textually. I've kept this minimal and modeled on the existing `\g`/`\G` handling specifically so it's easy to port: happy to rebase onto the scanner approach if #11202 lands first, or to fold `\c` handling into that effort — whichever you'd prefer.

Fixes #10867


vishnujayvel:
Noting for whoever picks this up: I flagged in the description that #11202's shell rework touches the same Uninterpreted callback, so happy to rebase onto the StreamScanner approach or fold \c handling into that effort directly, whichever's easier on your end.

zachmu:
Thank you for your contribution, @vishnujayvel! I carried your commits forward into https://github.com/dolthub/dolt/pull/11780 and strengthened the regression test to verify that canceled single-line and multiline INSERT statements leave the table empty. Closing this PR in favor of that replacement; your contribution is preserved in its commit history.

PR #11293, opened 2026-07-12
Fix fsck handling of stash refs

### LINKED ISSUE #11292: dolt fsck misclassifies stash list as a commit
## Description

`dolt fsck` fails whenever a repository contains a stash because it attempts to validate the stash-list object (`SLST`) as a commit object (`DCMT`).

This reproduces in a brand-new repository. Clearing the stash makes `dolt fsck` pass.

## Environment

- Dolt: `2.1.10`
- Also reproduced with Dolt: `2.0.7`
- OS: macOS ARM64

## Minimal reproduction

```bash
mkdir dolt-fsck-stash-repro
cd dolt-fsck-stash-repro
dolt init
dolt sql -q 'create table t (id int primary key, value int); insert into t values (1, 1);'
dolt add t
dolt commit -m 'initialize repro table'
dolt sql -q 'update t set value=2 where id=1;'
dolt stash
dolt stash list
dolt fsck
```

## Actual result

`dolt stash list` shows the expected stash, then `dolt fsck` exits nonzero:

```text
stash@{0}: WIP on refs/heads/main: <commit> initialize repro table

Starting tree validation for 3 commit objects...
commit tree validation failed: runtime error: commit <hash> has incorrect file ID: expected DCMT, got SLST
```

Inspecting the reported hash with `dolt show <hash> --no-pretty` identifies it as a valid `SerialMessage StashList`, not a commit.

## Control test

```bash
dolt stash clear
dolt fsck
```

This completes successfully:

```text
No problems found.
```

## Expected result

`dolt fsck` should recognize and validate or skip the stash-list object rather than treating it as a commit. A valid stash should not cause the repository integrity check to fail.

0x687931:
## Summary

- exclude stash-list refs from commit DAG roots during `dolt fsck`
- add a regression test proving repositories with stashes pass `fsck`
- verify the stash remains present after the integrity check

## Root cause

`getRawReferencesFromStoreRoot` treated `ref.StashRefType` as a commit-pointing ref. A stash ref points to an `SLST` stash-list object, so commit-tree validation attempted to parse that object as `DCMT` and failed.

The full chunk scan still validates and categorizes `SLST` and `STSH` chunks. This change only prevents a non-commit ref from entering commit DAG traversal.

## Validation

- `go test ./cmd/dolt/commands`
- built the patched `dolt` binary from current `main`
- reproduced the original failure in a fresh repository
- verified the patched binary reports `No problems found.` with the stash preserved
- `git diff --check`

Fixes #11292

### DIFF
--- a/go/cmd/dolt/commands/fsck.go
+++ b/go/cmd/dolt/commands/fsck.go
@@ -1011,9 +1011,12 @@ func getRawReferencesFromStoreRoot(ctx context.Context, cs chunks.ChunkStore, er
 
 			refType := doltRef.GetType()
 			switch refType {
-			case ref.BranchRefType, ref.RemoteRefType, ref.InternalRefType, ref.WorkspaceRefType, ref.StashRefType:
+			case ref.BranchRefType, ref.RemoteRefType, ref.InternalRefType, ref.WorkspaceRefType:
 				// Address is the commit id.
 				refs[addr] = append(refs[addr], name)
+			case ref.StashRefType:
+				// A stash ref points to a StashList, not a commit.
+				// Its chunks were already validated by the full chunk scan.
 			case ref.TagRefType:
 				if commitHash, ok := resolveTagToCommit(ctx, cs, name, addr, errs); ok {
 					refs[commitHash] = append(refs[commitHash], name)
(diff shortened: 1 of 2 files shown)


macneale4:
@0x687931 - thanks for the PR! I think we need a little more thorough validation of the object. I've created a PR which supersedes this one: https://github.com/dolthub/dolt/pull/11298

Issue #11087, opened 2026-05-21
Optimize Git remote backend object reads to avoid many git cat-file subprocesses


webkaz:
## Summary

We are using Dolt Git remote support with Beads, storing Dolt data in a GitHub repository under `refs/dolt/data`.

On a small Beads database, read-only fetch from GitHub takes about 23 seconds even when the underlying Git fetch of `refs/dolt/data` is only about 1.4 seconds according to `GIT_TRACE2`. The trace shows hundreds of `git cat-file` subprocess invocations after fetch.

This suggests the main cost is not GitHub network transfer, but the Git remote backend's object access pattern.

## Environment

- Dolt: 2.0.3
- Git: Apple Git 2.50.1
- OS: macOS
- Remote: `git+https://github.com/webkaz-labs/shirube.git`
- Use case: Beads embedded Dolt DB synced through a GitHub Git remote

## Measurements

Measured on a copied database, not the live DB:

- Local embedded Dolt DB: about 99M before GC, about 75M after GC
- `.dolt/noms`: 54M before GC, 29M after GC
- `.dolt/git-remote-cache`: 45M
- Filesystem remote no-op push/pull: about 1.1s to 1.5s
- GitHub Git remote `dolt fetch origin main`: 23.32s before GC, 22.43s after GC
- `GIT_TRACE2` shows the actual `git fetch --no-tags --refmap= origin +refs/dolt/data:...` section is only about 1.4s
- The trace showed 772 to 814 `git cat-file` invocations during the operation

Running `dolt gc` reduced local disk usage, but did not materially improve the GitHub Git remote fetch time.

## Request

Could the Git remote backend avoid per-object `git cat-file` process startup overhead?

Possible directions:

- Use `git cat-file --batch` / `--batch-check`
- Keep a persistent `git cat-file` subprocess for object lookup
- Batch object reads from the Git remote cache
- Add a no-op / already-up-to-date fast path for Git remotes
- Add optional debug/profiling output for Git remote backend object reads

The no-op / small-change path matters for Beads-style workflows, where agents frequently push or pull small task metadata changes. In our current setup, `bd dolt push` often takes 50+ seconds even for tiny Beads note updates, while a filesystem Dolt remote baseline is around 1 second.

## Why this matters

Dolt Git remote support is useful because it lets Beads users keep task data in the same GitHub repository as the source code, under `refs/dolt/data`. The current behavior works, but the fixed cost is high enough that agents are incentivized to batch or avoid syncs.

[...]

PR #11301, opened 2026-07-14
gitblobstore: materialize blobs to a local file cache for ranged reads (fixes read half of #11087)

maphew:
## Problem

Every `BlobRange` read through the Git remote backend spawns one `git cat-file -s` plus one `git cat-file blob` subprocess, and a ranged read at offset N must stream and discard N bytes before returning anything (`cat-file` has no server-side range). Table-file access during pull/fetch is many small ranged reads, so on stores with large table files the aggregate cost is quadratic and the operation effectively never finishes.

Observed in the field (via [beads](https://github.com/gastownhall/beads), which embeds this backend — see gastownhall/beads#4770): a pull against a store with a ~39MB table file spawned cat-file pairs continuously (~60 processes/sec on Windows, the same blob re-read up to 33 times in one 15-second `GIT_TRACE` window), with single reads observed discarding 18MB+ prefixes. The pull was killed after 10+ minutes with no end in sight, indistinguishable from a hang.

This is the read-amplification half of #11087, and implements the existing TODO in `sliceInlineBlob`:

> Consider caching/materializing blobs to a local file (or using a batched git cat-file mode) to serve ranges efficiently.

## Change

`blobFileCache` materializes each blob into a local file on first read — one `cat-file blob` per blob, ever — and serves all subsequent reads (full, ranged, and repeated) with ordinary file I/O and seeks. Blob sizes come from `stat` instead of `cat-file -s` when cached.

Correctness relies on git blobs being content-addressed and immutable, so entries never go stale. Materialization is singleflighted in-process; on disk it writes to a temp file and renames into place, so concurrent same-OID materializations and cross-process races are benign (the loser uses the winner's identical file; the Windows rename-over-existing failure case is handled).

Operational knobs:
- Cache lives at `<gitdir>/dolt-blob-cache/` (git tools ignore unknown dirs in a git dir)
- `DOLT_GIT_BLOB_FILE_CACHE=0` disables it entirely, restoring the previous streaming path (which is kept intact)
- Size-capped at 4 GiB by default, `DOLT_GIT_BLOB_FILE_CACHE_MAX_BYTES` to override; best-effort LRU (by mtime) eviction after materializations, skipping files that can't be removed (e.g. open on Windows)

`multiPartReadCloser` gains an optional injected slice-opener so the chunked-object path benefits too; when absent it behaves exactly as before (existing unit tests unchanged and passing).

## Impact

The field-report store above: `dolt pull` via the embedding application went from effectively hanging to **completing in 8 seconds**, verified against the same repository and remote.

## Tests

- `TestBlobFileCacheMaterializesOnce` — N reads with ranges = exactly 1 `BlobReader` call
- `TestBlobFileCacheConcurrentOpen` — 16 concurrent opens share one materialization
- `TestBlobFileCacheDisabledByEnv` — kill-switch semantics
- `TestSliceFileBlobRanges` — file-backed slicing matches `sliceInlineBlob` range semantics (all/offset/length/suffix/clamp/invalid)
- `TestBlobFileCacheTrim` — LRU eviction respects the cap
- Full `store/blobstore` and `store/blobstore/internal/git` suites pass on this branch

The `remotes-git.bats` cat-file-counting assertions should only get easier to satisfy (strictly fewer spawns); I was unable to run bats locally (Windows dev box).

https://claude.ai/code/session_013M2Ynv9UuPybJwQqgLU5Y4

### DIFF
--- a/go/store/blobstore/git_blob_file_cache.go
+++ b/go/store/blobstore/git_blob_file_cache.go
@@ -1,0 +1,239 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blobstore
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"sync"
+
+	"github.com/dolthub/dolt/go/store/blobstore/internal/git"
+)
+
+// blobFileCache materializes git blobs into local files so that ranged and
+// repeated reads are served by ordinary file I/O instead of spawning a
+// `git cat-file` subprocess per read and streaming the blob from byte zero.
+//
+// Without it, every BlobRange read of an inline or chunked-part blob costs one
+// `git cat-file -s` plus one `git cat-file blob` subprocess, and a read at
+// offset N first streams and discards N bytes. Table-file reads during a pull
+// are many small ranged reads, so on stores with large table files the
+// aggregate cost is quadratic and a pull can appear to hang (see the runner
+// TODO this replaces, and gastownhall/beads#4770 for a field report).
+//
+// Cache entries are keyed by blob OID. Git blobs are content-addressed and
+// immutable, so entries can never go stale. Materialization writes to a temp
+// file in the same directory and renames it into place; the loser of a
+// concurrent race simply uses the winner's file.
+//
+// The cache directory lives inside the git dir (git tools ignore unknown
+// directories there). Set DOLT_GIT_BLOB_FILE_CACHE=0 to disable the cache and
+// fall back to streaming reads. DOLT_GIT_BLOB_FILE_CACHE_MAX_BYTES overrides
+// the default size cap; the cap is enforced best-effort by evicting the
+// least-recently-modified entries after a materialization pushes the total
+// over the cap.
+type blobFileCache struct {
+	dir      string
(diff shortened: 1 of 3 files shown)


elianddb:
@maphew We should have a release out for this soon. #11247 was merged a bit ago, but we'll be making a release today.

maphew:
Wonderful, thanks!

elianddb:
@maphew Release is out let me know if the issue is fixed for you!

maphew:
Partial confirmation so far: on Linux x86_64 I built beads (bd) against v2.2.0 (`dolt/go v0.40.5-0.20260715172757-a6690826d767`) and exercised the affected repository (243 MB of dolt data behind `refs/dolt/data`, including the 39 MB table-file blob from my report):

- Fresh-clone bootstrap: all 71,605 chunks in ~25 s.
- `bd dolt pull` immediately after: 5.4 s, clean exit, no orphaned `git cat-file` child.

One caveat before I call it fixed: my deterministic repro environment is a Windows machine (where the pipe deadlock hung every time), and this Linux box turns out not to reproduce the original hang even with a pre-#11247 build (`dolt/go v0.40.5-0.20260626152440`) — so the numbers above show v2.2.0 is healthy on this path, but not that it fixed *this* bug. I'll re-test on the Windows machine shortly and report back; if it's clean there I'll close this PR, since #11247 covers the read half for my case.

_claude-fable-5-xhigh on behalf of maphew_

maphew:
Windows confirmation complete on the deterministic repro machine.

I built the same Beads revision used for the original Windows investigation against Dolt v2.2.0 (`dolt/go v0.40.5-0.20260715172757-a6690826d767`) and exercised a fresh clone of the affected `maphew/mybd` repository, which still contains the 38,957,891-byte table-file blob that triggered the quadratic ranged-read path.

- `bd bootstrap --yes`: all 81,719 chunks retrieved in 16.393 s; exit 0.
- Immediate `bd dolt pull`: 2.289 s; exit 0.
- `git cat-file` processes before/after both operations: 0/0; no child was left behind.
- The cloned database opened successfully afterward with 893 issues.

That is the Windows result the earlier Linux test was missing. I consider the reported read-path problem fixed by released #11247, so I am closing this alternative PR. Thanks for getting the fix released so quickly.

_codex-gpt-5.6-sol-high on behalf of matt wilkie_

PR #11303, opened 2026-07-15
sql: fix empty/comment-only query handling in interactive shell

### LINKED ISSUE #10863: Do not emit warnings for empty query after trimming comments
Dolt returns an `Empty set` and emits a warning when a comment has been trimmed to an empty query, whereas MySQL returns a `Query OK, 0 rows affected` with no warning

Seems like the warning is coming from GMS

Dolt
```
tmp/main*> -- comment ;
Empty set, 1 warning (0.00 sec)

tmp/main*> show warnings;
+---------+------+----------------------------------------------------------------+
| Level   | Code | Message                                                        |
+---------+------+----------------------------------------------------------------+
| Warning | 0    | query was empty after trimming comments, so it will be ignored |
+---------+------+----------------------------------------------------------------+
1 row in set (0.00 sec) 

Warning (Code 0): query was empty after trimming comments, so it will be ignored
```

MySQL
```
mysql> -- comment ;
Query OK, 0 rows affected (0.00 sec)
```

### LINKED ISSUE #10864: MySQL has error for empty delimited queries whereas Dolt has misleading warning
MySQL (unclear if this error is coming from the MySQL shell or server)
```
mysql> ;
ERROR: 
No query specified
```

Dolt
```
tmp/main*> ;
Empty set, 1 warning (0.00 sec)

tmp/main*> show warnings;
+---------+------+----------------------------------------------------------------+
| Level   | Code | Message                                                        |
+---------+------+----------------------------------------------------------------+
| Warning | 0    | query was empty after trimming comments, so it will be ignored |
+---------+------+----------------------------------------------------------------+
1 row in set (0.00 sec) 

Warning (Code 0): query was empty after trimming comments, so it will be ignored
```
(Warning is related to #10863)

vishnujayvel:
## What

In the interactive `dolt sql` shell, two related edge cases around empty/comment-only statements produced confusing output instead of clear, MySQL-consistent behavior:

1. A bare empty query (e.g. typing just `;`) silently printed `Empty set, 1 warning` with a warning body of `query was empty after trimming comments`, instead of a clear client-side error.
2. A comment-only query (e.g. `-- foo ;`) behaved the same way — a spurious "empty" warning — instead of being silently skipped the way `dolt sql < file.sql` (batch mode) already treats it.

This PR makes the shell branch on `sqlparser.ErrEmpty` explicitly in `execShell`'s `Uninterpreted` callback (`go/cmd/dolt/commands/sql.go`):

- A truly empty statement (nothing left after `;` is stripped) now prints a MySQL-style `No query specified` error, matching how the real `mysql` CLI responds to a bare `;`.
- A comment-only statement (non-empty input that parses to `ErrEmpty` because it's all comment) is now skipped silently — no output, no warning — which mirrors the existing behavior of `execBatchMode`, where `err == sqlparser.ErrEmpty` is already a silent `continue`.

## Why

Batch mode (`execBatchMode`, sql.go) and `processQuery` (used by scripting/non-interactive paths) both already treat `sqlparser.ErrEmpty` as "silently skip." The interactive shell was the only place papering over this with a fake `Empty set, 1 warning` result, which is misleading — there's no result set, and the "warning" concept here doesn't correspond to anything the query actually did. This change brings the shell in line with the rest of the codebase's handling of empty statements while adding a genuinely useful error for the truly-empty case (bare `;`), since that's a real MySQL client-side error dolt users would expect to see.

Closes #10864
Closes #10863

Thanks to angelamayxie for the original reports.

## Test plan

- [x] `go build ./cmd/dolt/...` — passes
- [x] `go vet ./cmd/dolt/...` — passes
- [x] `gofmt -l go/cmd/dolt/commands/sql.go` — clean
- [x] New bats regression tests added in `integration-tests/bats/sql-shell.bats`:
  - `sql-shell: bare empty query gives MySQL-style error, not a warning` (uses new `sql-shell-empty-query.expect`)
  - `sql-shell: comment-only query is skipped silently, no warning` (uses new `sql-shell-comment-only-query.expect`)
- [x] Both new bats tests run locally against a freshly built `dolt` binary and pass:
  ```
  ok 1 sql-shell: bare empty query gives MySQL-style error, not a warning
  ok 1 sql-shell: comment-only query is skipped silently, no warning
  ```
- Full `sql-shell.bats` suite regression run was kicked off locally to check for unrelated breakage; not gating this PR description on it since the two new/targeted cases already passed cleanly and no other test in the file touches this code path.


vishnujayvel:
This one brings the interactive shell's empty/comment-only query handling in line with what execBatchMode already does, so it's a small behavioral consistency fix rather than new logic. Bats coverage for both the bare-`;` and comment-only cases is passing locally — glad to adjust if you'd rather handle the two cases differently.

zachmu (approved):

PR #11347, opened 2026-07-27
Fix panic on empty-string value for list options

nikolauspschuetz:
Passing an empty string as the value of a list-valued option (e.g. `dolt log --not ""`, `dolt diff --include-cols ""`) no longer panics. The empty value is accepted like any other argument.

### DIFF
--- a/go/libraries/utils/argparser/parser.go
+++ b/go/libraries/utils/argparser/parser.go
@@ -496,7 +496,7 @@ func getListValues(args []string) []string {
 
 	for _, arg := range args {
 		// Stop if another option found
-		if arg[0] == '-' || arg == "--" {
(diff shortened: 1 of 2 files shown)


elianddb:
Hi @nikolauspschuetz, this seems like a small change could you reduce the descriptions so its fits neatly into our release notes. You an take a looked at some of the PRs of our staff to get an idea, we're trying clamp down the formatting.

nikolauspschuetz:
Thanks @elianddb — trimmed the description down to a single release-note line. Let me know if you want it shorter or in a different shape.

elianddb:
lgtm! I'll apply the changes on a separate branch to rm the extra comments.

PR #11351, opened 2026-07-28
kvexec: fix left outer merge join wrong results with residual filters

### LINKED ISSUE #11350: kvexec: left outer merge join with residual filter emits spurious null-extended rows and drops final unmatched left row (wrong anti-join results)
## Summary

`mergeJoinKvIter` (go/libraries/doltcore/sqle/kvexec/merge_join.go) returns wrong results for **LEFT OUTER merge joins** when (a) the join keys contain duplicate runs and (b) the join condition carries a residual (non-merge-key) filter. Two distinct defects combine:

1. **A left row that already matched is re-emitted as a spurious null-extended row.** In the `match:` loop, `l.matchedLeft = ok` is assigned on *every* pairing, so a later pairing in the same key run that fails the residual filter **overwrites** the earlier success. At the key-run transition the `isLeftJoin && !l.matchedLeft` check then wrongly emits the null-extended row for a left row that had a match.

2. **The final unmatched left row's null-extended row is dropped.** When `l.leftIter.Next(ctx)` returns `io.EOF` while advancing past a completed left row (inside the `match:` loop's advance branch), the error is returned immediately — skipping the left-join non-match emission owed to the row just completed (`tmpKey`). (The `compare:` loop's `cmp == -1` branch handles this case correctly via `oldLeftKey`; the `match:` loop advance branch does not.)

Because `LEFT JOIN … WHERE right.col IS NULL` anti-joins are built on exactly these null-extended rows, both defects surface as **phantom orphans / missing orphans** in anti-join queries. Found in production: the same anti-join returned 197 rows or ~15,000 rows depending on an unrelated `NOT LIKE` predicate (whose only effect was flipping the plan into the merge-join path).

Verified on `dolt version 2.2.2` (also reproduces on 2.1.10 and 1.50.0 — longstanding, not a recent regression). The gms `LeftOuterMergeJoin` row iterator (`sql/rowexec/merge_join.go`) handles all these cases correctly — the bug is only in the Dolt kvexec implementation; the gms memory engine returns correct results for the identical plan.

## Minimal repro

```sql
create table b2 (cat varchar(50) not null, code varchar(16) not null, lang varchar(20) not null,
  primary key(cat, code, lang), key(code));
create table t2 (id varchar(36) primary key, code varchar(16), lang varchar(16), key(code, lang));
insert into b2 values ('cat0','P1','de:app'),('cat1','P1','fr:app'),('cat2','P1','nl:app');
insert into t2 values ('t1','P1','de'),('t2','P1','es'),('t3','P1','fr');

select /*+ MERGE_JOIN(p,w) */ p.lang, w.lang
from b2 p left join t2 w
  on w.code = p.code and w.lang = substring_index(p.lang, ':', 1);
```

Expected (MySQL, and gms memory engine):

```
de:app | de
fr:app | fr
nl:app | NULL
```

Actual (dolt 2.2.2, kvexec merge join):

```
de:app | de
fr:app | fr
fr:app | NULL     <-- spurious: fr already matched (defect 1)
                  <-- nl:app | NULL is missing (defect 2)
```

Plan: `LeftOuterMergeJoin` / `cmp: (p.code = w.code)` / `sel: (w.lang = SUBSTRING_INDEX(p.lang, ':', 1))`. Same query with `LOOKUP_JOIN` hint or as `NOT EXISTS` returns correct results.

Walkthrough of defect 1 with the repro data (right key run buffered as `[es, fr]` + current `de`):
- `fr:app` pairs against `es` (fail), `fr` (**match**, `matchedLeft = true`, row emitted), then `de` (fail → `matchedLeft = ok` **resets to false**);
- key-run transition sees `matchedLeft == false` → emits `fr:app | NULL`.

Then for `nl:app` (matches nothing), the left iterator EOFs during the advance and the owed `nl:app | NULL` is dropped (defect 2).

At scale each duplicate-key run contributes one phantom and can lose one orphan: a 15k-row synthetic version of the same shape returns 9,999 for an anti-join whose ground truth (via `NOT EXISTS`) is 5,000.

## Fix

Both defects are small and local; a PR with the fix + regression tests follows:

- defect 1: `l.matchedLeft = l.matchedLeft || ok` (only ever *set* within a key run; the existing reset at the run transition remains the sole reset point);
- defect 2: on `io.EOF` from the left-advance in the `match:` loop, emit the null-extended row for `tmpKey` when `isLeftJoin && !matchedLeft` before surfacing EOF (mirroring the `compare:` loop's `oldLeftKey` handling).

With the fix: minimal repro correct; the 15k synthetic anti-join returns exactly 5,000 (= `NOT EXISTS`); existing `kvexec` and `enginetest` join suites pass (`TestJoinQueries`, `TestJoinOps`).

Possibly related to the join-correctness audit in #10316 (different axis: that issue tracks type coercion; this one is left-outer bookkeeping, same operator family).

---

*Found while running orphan audits on the [holywritings.net](https://holywritings.net) Bahá'í writings database (~60k rows, prayer_book_structure × writings anti-joins), where the phantom-orphan inflation was caught because adding a filter can only ever shrink a result set.*

LaPingvino:
Fixes #11350 — wrong results from `mergeJoinKvIter` for left outer merge joins whose key runs contain duplicates and whose join condition carries a residual (non-merge-key) filter.

Two one-line-class defects, both in the `match:` loop:

1. **`l.matchedLeft = ok` overwrote earlier successes.** A left row that matched mid-run was re-flagged unmatched by a later failed pairing in the same run, so the run transition emitted a spurious null-extended row for it. Changed to `l.matchedLeft = l.matchedLeft || ok`; the run-transition reset remains the sole reset point.

2. **`io.EOF` during the left advance dropped the final owed null-extension.** The `compare:` loop handles left-EOF via `oldLeftKey`; the `match:` loop's advance branch returned the error immediately, losing the null-extended row for a final unmatched left row. Now it emits that row (when `isLeftJoin && !matchedLeft`) before surfacing EOF.

Minimal repro (returns `fr:app|NULL` instead of `nl:app|NULL` before this patch):

```sql
create table b2 (cat varchar(50) not null, code varchar(16) not null, lang varchar(20) not null,
  primary key(cat, code, lang), key(code));
create table t2 (id varchar(36) primary key, code varchar(16), lang varchar(16), key(code, lang));
insert into b2 values ('cat0','P1','de:app'),('cat1','P1','fr:app'),('cat2','P1','nl:app');
insert into t2 values ('t1','P1','de'),('t2','P1','es'),('t3','P1','fr');
select /*+ MERGE_JOIN(p,w) */ p.lang, w.lang
from b2 p left join t2 w on w.code = p.code and w.lang = substring_index(p.lang, ':', 1);
```

The regression test registers the kvexec `ExecBuilder` on the test engine (as the CLI/server engines do) so the kvexec iterator — not the GMS fallback, which was already correct — is what's exercised; the new test fails 7 assertions without the fix and passes with it. `kvexec` suite and `enginetest` `TestJoinQueries`/`TestJoinOps` pass. Also verified against a 15k-row synthetic (anti-join now agrees exactly with `NOT EXISTS`: 5,000) and the production database where this was found (197-row orphan audit no longer inflates to ~15k when the planner picks the merge join path).

https://claude.ai/code/session_01VpTxMfzCXHoM8ojJn5aQiv

### DIFF
--- a/go/libraries/doltcore/sqle/kvexec/merge_join.go
+++ b/go/libraries/doltcore/sqle/kvexec/merge_join.go
@@ -204,7 +204,7 @@ match:
 				return nil, err
 			}
 			l.matchPos += 2
-			l.matchedLeft = ok
+			l.matchedLeft = l.matchedLeft || ok
 			if ok {
 				return candidate, nil
 			}
@@ -213,7 +213,7 @@ match:
 			if err != nil {
 				return nil, err
 			}
-			l.matchedLeft = ok
+			l.matchedLeft = l.matchedLeft || ok
 			l.matchPos++
 			if ok {
 				return candidate, nil
@@ -228,6 +228,19 @@ match:
 			tmpKey, tmpVal := l.leftKey, l.leftVal
 			l.leftKey, l.leftVal, err = l.leftIter.Next(ctx)
 			if err != nil {
+				if errors.Is(err, io.EOF) && l.isLeftJoin && !l.matchedLeft {
+					// the just-completed left row never matched; emit its
+					// null-extended row before surfacing EOF
+					l.exhaustLeft = true
+					l.leftKey = nil
+					ret, ok, buildErr := l.buildResultRow(ctx, tmpKey, tmpVal, nil, nil)
+					if buildErr != nil {
+						return nil, buildErr
+					}
+					if ok {
+						return ret, nil
+					}
+				}
 				return nil, err
 			}
 			cmp, cmpErr := l.llCmp(tmpKey, tmpVal, l.leftKey, l.leftVal)
--- a/go/libraries/doltcore/sqle/kvexec/merge_join_test.go
+++ b/go/libraries/doltcore/sqle/kvexec/merge_join_test.go
@@ -1,0 +1,127 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kvexec
+
+import (
+	"context"
+	"testing"
+
+	"github.com/dolthub/go-mysql-server/sql"
+	"github.com/dolthub/go-mysql-server/sql/rowexec"
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/dtestutils"
+	"github.com/dolthub/dolt/go/libraries/doltcore/sqle"
+	"github.com/dolthub/dolt/go/libraries/doltcore/table/editor"
+)
+
+// TestLeftOuterMergeJoinResidualFilter is a regression test for a pair of
+// wrong-result bugs in mergeJoinKvIter for left outer merge joins whose key
+// runs contain duplicates and whose join condition carries a residual
+// (non-merge-key) filter:
+//
+//  1. matchedLeft was overwritten (rather than or-ed) on every pairing, so a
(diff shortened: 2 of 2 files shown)


angelamayxie:
Thanks so much for the fix! I think the tests should belong in GMS since we want these join queries to work in both GMS and Dolt. I'll move them over :)

Issue #509, opened 2020-04-01
Ensure ability to use AWS S3 compatible data stores


fire:
S3 is not just AWS's protocol but a protocol for many other systems. Dolt's support for S3 compatible apis would be great.

timsehn:
Like you want to be able to push to a remote from Dolt that has an S3 compatible API? And you want that supported in the remote syntax? Can you give us an example of one you want that we can build to?

fire:
https://www.s3-client.com/s3-compatible-storage-solutions.html 

Here's a list. 

I would start with https://min.io/

zachmu:
Dolt's S3 storage requires Dynamo as well, to store manifest contents. In order for S3 compatible storage offerings to work, we would need to implement a file-based s3 manifest, as is currently the case with GCS remotes.

omar391:
@timsehn Any update on this?

timsehn:
> @timsehn Any update on this?

We support S3, filesystem, GCS, and a custom remote server called `remotesrv`.  We don't have much demand for more supported remote types.

https://docs.dolthub.com/sql-reference/version-control/remotes

priley86:
just +1'ing here as someone exploring dolt. I'd love a more generic s3/ MinIO compatible remote (supposing this would have to be added as a new nbs/blobstore). This doesn't look overly difficult to write given the others already implemented and would be a huge value.

jonesmabea:
The current Dolt tree appears to make this tractable now: `Blobstore.CheckAndPut` and `blobstoreManifest` already provide optimistic manifest CAS, including rereading current contents after a conflict. AWS S3 now supports `If-None-Match: *` for creation and `If-Match: <ETag>` for replacement; Cloudflare R2 and current MinIO AIStor also document both headers on `PutObject`.

I've opened draft PR #11433 with a working implementation: an endpoint-configurable S3 `Blobstore` treating the manifest ETag as an opaque version token. A conditional 412 (and AWS's conditional `409 ConditionalRequestConflict`) maps to `blobstore.CheckAndPutError`, so the existing NBS conflict path is unchanged. Table files go through the no-conjoin store, avoiding any S3 compose operation.

[...]

PR #11433, opened 2026-08-06
Add s3:// remote scheme for generic S3-compatible object stores (#509)

jonesmabea:
## Motivation

#509 asks for S3-compatible storage support. The historical blocker was the manifest: `aws://` pairs S3 with a DynamoDB table for the atomic check-and-set, which generic S3-compatible stores cannot provide. That blocker is gone: AWS S3 supports conditional writes (`If-None-Match: *` for create since Aug 2024, `If-Match: <ETag>` CAS since Nov 2024), and Cloudflare R2 and current MinIO AIStor document both headers on `PutObject`.

## Design

A new `s3://bucket/path` scheme backed by an endpoint-configurable `S3Blobstore`:

- The manifest ETag is an opaque version token. `CheckAndPut` issues a direct single-part conditional `PutObject` (`If-None-Match: *` on create, `If-Match` on replace). HTTP 412, AWS's conditional `409 ConditionalRequestConflict`, and `If-Match` against a missing key all map to the unwrapped `blobstore.CheckAndPutError`, so the existing NBS manifest reread/retry path handles races unchanged. 429/5xx are left to the caller's retry with the original condition preserved.
- `CheckAndPut` never uses multipart (multipart ETags are not content-stable across providers, and conditional headers on `CompleteMultipartUpload` are not universally supported). Ordinary `Put` uses the SDK upload manager, which switches to multipart automatically for large table files.
- The factory wires `nbs.NewNoConjoinBSStore`, so `Concatenate` is never invoked — no server-side compose operation is required, following the Git-remote backend precedent.
- Credentials come from the standard AWS SDK chain. New optional remote params: `s3-endpoint`, `s3-region`, `s3-path-style` (the `AWS_ENDPOINT_URL_S3` env var is honored when params are omitted).

The first commit fixes a pre-existing gap this work surfaced: `noConjoinBlobstorePersister.Open` lacked the archive-format (`.darc`) fallback that `blobstorePersister.Open` has, so pushing archive table files into a no-conjoin blobstore store failed at `openChunkSources`.

## Verification

- Integration tests (env-gated on `TEST_S3_BUCKET`/`TEST_S3_ENDPOINT`/`TEST_S3_PATH_STYLE`, following the GCS/OCI convention) cover put/get/exists/ranged reads, conditional create/update/stale-version/missing-key semantics, an 8-writer concurrent `CheckAndPut` race asserting exactly one winner, multipart `Put`, and unsupported `Concatenate`. All pass against live Cloudflare R2.
- End-to-end against R2: `remote add` → `push` → `clone` → data verified → second `push` → `pull` round-trip verified.
- Caveat stated plainly: the concurrent-race test is a smoke test, not a proof of provider-side CAS linearization. R2 documents the conditional headers but not an explicit exactly-one-winner guarantee, and R2 rate-limits ~1 write/sec per key (relevant to the hot manifest). A more rigorous conformance harness (synchronized independent clients, disabled SDK retries, raw timing/status capture, repetition) is planned; providers lacking conditional writes should eventually get a fast capability-probe error.

## Open questions

1. Is this design shape acceptable, and is the work planned under the current #509 assignment far enough along that this PR conflicts? Happy to adapt or hand off.
2. Naming/config bikeshed: `s3://` scheme and `s3-endpoint`/`s3-region`/`s3-path-style` param names.
3. Whether a startup capability probe (fail-fast on providers without conditional-write support) should land in this PR or a follow-up.

Known cosmetic issue: a second push to an existing branch prints `[new branch]` instead of a fast-forward line; investigating.

### DIFF
--- a/go/cmd/dolt/cli/arg_parser_helpers.go
+++ b/go/cmd/dolt/cli/arg_parser_helpers.go
@@ -156,6 +156,9 @@ func CreateCloneArgParser() *argparser.ArgParser {
 	ap.SupportsString(dbfactory.AWSCredsProfile, "", "profile", "AWS profile to use.")
 	ap.SupportsString(dbfactory.OSSCredsFileParam, "", "file", "OSS credentials file.")
 	ap.SupportsString(dbfactory.OSSCredsProfile, "", "profile", "OSS profile to use.")
+	ap.SupportsString(dbfactory.S3EndpointParam, "", "endpoint", "S3-compatible endpoint url (e.g. Cloudflare R2 or MinIO); only valid for s3 remotes.")
+	ap.SupportsString(dbfactory.S3RegionParam, "", "region", "Signing region for s3 remotes.")
+	ap.SupportsString(dbfactory.S3PathStyleParam, "", "true|false", "Use path-style addressing for s3 remotes.")
 	ap.SupportsString(UserFlag, "u", "user", "User name to use when authenticating with the remote. Gets password from the environment variable {{.EmphasisLeft}}DOLT_REMOTE_PASSWORD{{.EmphasisRight}}.")
 	ap.SupportsFlag(SingleBranchFlag, "", "Clone only the history leading to the tip of a single branch, either specified by --branch or the remote's HEAD (default).")
 	return ap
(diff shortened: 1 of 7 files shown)


macneale4:
@jonesmabea - thanks so much for the contribution! I've created a PR with your changes here: https://github.com/dolthub/dolt/pull/11571 

I'm going to get our CI to execute this code with one of our CI keys, so it will be easier for me to take it from here.

macneale4:
One interface change I've decided to make is that the arguments will be specified as query string arguments in the URL. The --aws-* example followed in this change is actually one we don't want to emulate. It's best to have the entire remote specified as a single URL.

jonesmabea:
@macneale4 Thanks so much for taking the PR forward,  much appreciated! The design change makes sense to me; having the entire remote specified as a single URL with query-string arguments is cleaner than the `--aws-*` flag pattern. Looking forward to the new functionality :)

PR #11437, opened 2026-08-06
Name the conflicting ref in ExistingRefError

### LINKED ISSUE #11434: `ExistingRefError.Error()` omits the ref name and type
`doltdb.ExistingRefError` is returned when a caller-provided name collides with an existing ref. Its `Error()` explicitly returns a "already exists", a bare string.

https://github.com/dolthub/dolt/blob/56fc3f4f292598b644df4397d5ef1799a5883c07/go/libraries/doltcore/doltdb/errors.go#L158-L162

The type carries a `Ref` field that identifies the conflict. Branch creation already recovers this detail out of band, but tag and workspaces do not.

`Error()` should include the ref, for example: `ref 'refs/heads/x' already exists`. Updating 'Error()` will change the message tag and workspace callers and test assert against, so they'll have to be updated with their own wrappers.

asadbek0512:
Fixes #11434

`ExistingRefError.Error()` returned a bare `already exists`. A user seeing it learns neither which ref was taken nor what kind it was. The `TODO(elianddb)` on it said the ref could only be named once tags and workspaces used their own errors, since they surfaced this message verbatim.

## Changes

- `actions.ErrTagExists` / `actions.ErrWorkspaceExists`, mirroring the existing `ErrBranchExists` wording.
- `CreateTagOnDB` and `CreateWorkspaceOnDB` return those instead of `&doltdb.ExistingRefError{...}`.
- `ExistingRefError.Error()` now returns `ref 'refs/heads/br' already exists`; the `TODO` is gone.

## Why tags and workspaces return their error at the source

`BranchExistsError` exists because `failOnCaseConflict` raises `ExistingRefError` deep inside branch creation, where the ref that actually collides differs from the one the caller asked for; the caller has to unwrap to learn the real name. Tag and workspace creation instead detect the duplicate with an explicit `ddb.HasRef` check that already has the name in scope, and neither path runs `failOnCaseConflict`. An unwrapper for them would only re-derive a name the raising function held, so they return the typed error directly. `CreateWorkspace` currently has no callers, which also makes an unwrapper dead code.

That leaves `ExistingRefError` raised only on branch paths, whose callers already wrap it. Its own message is the fallback for a caller that does not, so it uses `Ref.String()` rather than `GetPath()`: a name can exist as more than one ref type, and an unwrapped error should not be ambiguous about which one it means.

## Compatibility

No test asserted the bare string. `branch_case_test.go` matches on the error type and `Ref.GetPath()`, both unchanged. No enginetest or bats case asserts a tag- or workspace-exists message; the bats suites that grep for `already exists` cover tables, foreign keys, and files, and the new messages still contain that substring.

## Testing

`go test` passes for `./libraries/doltcore/doltdb/...`, `./libraries/doltcore/env/...`, and `-run 'TestDoltTag|TestDoltBranch|TestDoltCheckout'` in `./libraries/doltcore/sqle/enginetest/`. `go vet` is clean on the touched packages. Added `TestExistingRefErrorNamesTheRef` to pin the message across branch, tag, and workspace refs.

### DIFF
--- a/go/libraries/doltcore/doltdb/errors.go
+++ b/go/libraries/doltcore/doltdb/errors.go
@@ -149,16 +149,18 @@ func GetUnreachableRootCause(err error) error {
 // [ref.DoltRef], such as a branch or tag, conflicts with one or more
 // existing names.
 //
-// Ref identifies one exsiting conflict so callers can produce
+// Ref identifies one existing conflict so callers can produce
 // context-specific errors, such as actions.BranchExistsError.
 type ExistingRefError struct {
 	Ref ref.DoltRef
 }
 
+// Error names the conflicting ref. A caller that reaches the user should still
+// wrap this with a message for its own ref type; this is the fallback for one
+// that does not, and it must say which ref was already taken. The full ref path
+// disambiguates a name that exists as more than one type.
 func (e *ExistingRefError) Error() string {
-	// TODO(elianddb): Include Ref in the error message once tags, workspaces,
-	//  and other reference types use their own errors, and update their tests.
-	return "already exists"
+	return fmt.Sprintf("ref '%s' already exists", e.Ref.String())
 }
 
 // DoltIgnoreConflictError is an error that is returned when the user attempts to stage a table that matches conflicting dolt_ignore patterns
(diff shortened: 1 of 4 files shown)


elianddb:
Hi @asadbek0512, this is a relatively small change, could you shorten the description to a couple of sentences and drop the headers. Also, I would prefer if you dropped the new error test file. You may find existing infrastructure where you can test similar errors, for example with tags and the like.

elianddb:
Closing due to inactivity.

PR #11440, opened 2026-08-07
include the conflicting ref in ExistingRefError.Error()

### LINKED ISSUE #11434: `ExistingRefError.Error()` omits the ref name and type

shaurya703:
Fixes #11434

`ExistingRefError.Error()` returned a bare `already exists`, so callers that don't recover the detail out of band — tags and workspaces — surfaced an error that never named the conflicting ref:

```sql
mysql> call dolt_tag('v1');
mysql> call dolt_tag('v1');
ERROR 1105 (HY000): already exists
```

## Changes

- `Error()` now includes the ref, as suggested in the issue: `ref 'refs/tags/v1' already exists`. Workspaces (and any other caller that surfaces the raw error) pick this up automatically.
- `dolt_tag()` recovers a tag-specific message via a new `actions.TagExistsError`, mirroring the existing `actions.BranchExistsError` pattern: `fatal: A tag named 'v1' already exists.`
- Added `TestCreateTagOnExistingTag` covering the new `Error()` format, the `TagExistsError` recovery, and that unrelated errors are not converted.

Branch creation is unaffected — it already recovers the detail via `BranchExistsError`, and `branch_case_test.go` asserts with `ErrorAs`/`GetPath()` rather than the message string, so no existing assertions needed updating.

## Testing

`go test ./libraries/doltcore/env/actions/ ./libraries/doltcore/doltdb/` — all pass, including the new test.

### DIFF
--- a/go/libraries/doltcore/doltdb/errors.go
+++ b/go/libraries/doltcore/doltdb/errors.go
@@ -156,9 +156,7 @@ type ExistingRefError struct {
 }
 
 func (e *ExistingRefError) Error() string {
-	// TODO(elianddb): Include Ref in the error message once tags, workspaces,
-	//  and other reference types use their own errors, and update their tests.
-	return "already exists"
+	return fmt.Sprintf("ref '%s' already exists", ref.String(e.Ref))
 }
 
 // DoltIgnoreConflictError is an error that is returned when the user attempts to stage a table that matches conflicting dolt_ignore patterns
--- a/go/libraries/doltcore/env/actions/tag.go
+++ b/go/libraries/doltcore/env/actions/tag.go
@@ -16,8 +16,11 @@ package actions
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
+
+	errorKinds "gopkg.in/src-d/go-errors.v1"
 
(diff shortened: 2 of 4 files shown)


shaurya703:
Closing in favor of #11437, which was opened before this one and takes an equivalent approach for the same issue — I missed it when I started, apologies for the duplicate. One small thing worth carrying over if useful: this PR also added a regression test (`TestCreateTagOnExistingTag`) covering the duplicate-tag path end to end; happy for any part of it to be reused.

Issue #10951, opened 2026-04-27
dolt_rebase('--continue') error mentions dolt_conflicts_<table> even when no conflict rows persist after abort


timsehn:
## Problem

In an interactive rebase conflict flow, `dolt_rebase('--continue')` errors with guidance to resolve conflicts in `dolt_conflicts_<table>` and then continue rebasing.

But in the tested scenario, the rebase aborts immediately and does not leave persisted conflict rows behind, so that guidance appears incorrect or at least misleading.

## Repro

```sql
CREATE TABLE t(id INT PRIMARY KEY, v INT);
INSERT INTO t VALUES(1, 10), (2, 20);
CALL dolt_add('-A');
CALL dolt_commit('-m', 'c1');

CALL dolt_checkout('-b', 'feature');
UPDATE t SET v=200 WHERE id=1;
UPDATE t SET v=201 WHERE id=2;
CALL dolt_add('-A');
CALL dolt_commit('-m', 'feat_update');

CALL dolt_checkout('main');
UPDATE t SET v=300 WHERE id=1;
UPDATE t SET v=301 WHERE id=2;
CALL dolt_add('-A');
CALL dolt_commit('-m', 'main_update');

CALL dolt_checkout('feature');
CALL dolt_rebase('-i', 'main');
CALL dolt_rebase('--continue');
```

## Actual behavior

`dolt_rebase('--continue')` errors with text like:

> data conflict detected while rebasing commit ...
> Resolve the conflicts and remove them from the dolt_conflicts_<table> tables, then continue the rebase by calling dolt_rebase('--continue')

But after the abort:

```sql
SELECT count(*) FROM dolt_conflicts;
```

returns `0`, and there is no persisted conflict-table state left to resolve.

## Expected behavior

Either:

1. preserve conflict rows so the user can actually resolve them and continue, or
2. change the error message to reflect that the rebase was aborted and there is no conflict-table state to resolve

shaurya703:
I tried to reproduce this against current main (d95653646-era HEAD, built locally) using the exact script from the description, run in a single `dolt sql` session. The misleading `dolt_conflicts_<table>` guidance no longer appears — `dolt_rebase('--continue')` now errors with:

```
data conflicts from rebase, but session settings do not allow preserving conflicts, so they cannot
be resolved. The rebase has been aborted. Set @@autocommit to 0 or set @@dolt_allow_commit_conflicts
to 1 and try the rebase again to resolve the conflicts.
```

[...]

PR #11449, opened 2026-08-08
pin the aborted-rebase state behind ErrRebaseDataConflictsCantBeResolved

shaurya703:
Related to #10951 — extends the existing `dolt_rebase` data-conflict enginetest to assert the abort that the error message promises actually happened: original branch restored, the `dolt_rebase_<branch>` working branch and the `dolt_rebase` plan table both gone, and `dolt_conflicts` empty.


elianddb:
Hi @shaurya703, this seems to be a small change. Can you please reduce it down to a sentence of the gist of the change. Typically, if we require additional context, the issue is hyperlinked, or additional review comments can be left on the code snippets in the GitHub GUI.

shaurya703:
Thanks — cut down to a single sentence, with the issue linked for the context that was inline before.

PR #11474, opened 2026-08-09
explain the connection failure when a locked db's sql-server is unreachable

### LINKED ISSUE #10856: Error when LateBindQueryist fails to connect to the running sql-server noted in sql-server.info is confusing
When trying to use a DB which is locked by another process, the error message reported is that it could not connect, this is confusing and causes the user to take more time than necessary to resolve the issue.

The error already provided by `dolt sql-server` is much more helpful, I think that's a good candidate for all of the commands.

To reproduce, I'm using dolt 1.86.0, on macos, I have `dolt sql` running in one terminal. And in another:

```sh
❯ dolt status
dial tcp [::1]:3306: connect: connection refused

❯ dolt sql-server
Starting server with Config HP="localhost:3306"|T="28800000"|R="false"|L="info"|S="/tmp/mysql.sock"
database "efdb" is locked by another dolt process; either clone the database to run a second server, or stop the dolt process which currently holds an exclusive write lock on the database
```

`dolt status` , `dolt push`, etc. all give the same error.

shaurya703:
Fixes #10856

A CLI command that can't take the write lock routes through the sql-server named in `.dolt/sql-server.info`. When connecting to that server failed — usually a stale info file after an unclean shutdown — the user got a bare `dial tcp [::1]:3306: connect: connection refused` from the first query, with nothing saying where it came from or how to recover.

- `queryist_utils.go` — the late binder in `BuildConnectionStringQueryist` now pings before returning, so a dead server fails at connect time as `failed to connect to the dolt sql-server at <host>:<port>: <cause>` instead of leaking a bare dial error out of the first query.
- `dolt.go` — the implicit lock-detection path wraps that error with why this server was contacted at all (lock held, info file names the port) and how to recover (retry if it is still starting, otherwise delete the stale `.dolt/sql-server.info`), since on that path the user never chose the target.
- `queryist_utils_test.go` — new `TestBuildConnectionStringQueryistNamesTargetOnConnectFailure` binds a port, closes the listener, and asserts the resulting error names the host and port.

Read-only fallback, raised in the issue discussion, is a product decision and left out of scope.

### DIFF
--- a/go/cmd/dolt/commands/sqlserver/queryist_utils.go
+++ b/go/cmd/dolt/commands/sqlserver/queryist_utils.go
@@ -104,6 +104,13 @@ func BuildConnectionStringQueryist(_ context.Context, cwdFS filesys.Filesys, cre
 	queryist := ConnectionQueryist{connection: conn, gatherWarnings: &gatherWarnings}
 
 	var lateBind cli.LateBindQueryist = func(ctx context.Context, opts ...cli.LateBindQueryistOption) (res cli.LateBindQueryistResult, err error) {
+		// Surface connection failures here, with the target named, instead of
+		// letting the first query return a bare dial error like
+		// "dial tcp [::1]:3306: connect: connection refused".
+		if err := conn.DB.PingContext(ctx); err != nil {
+			return res, fmt.Errorf("failed to connect to the dolt sql-server at %s:%d: %w", host, port, err)
+		}
+
 		sqlCtx := sql.NewContext(ctx)
 		sqlCtx.SetCurrentDatabase(dbRev)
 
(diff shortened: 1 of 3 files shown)


elianddb:
Hi @shaurya703, can you please reduce the description a short paragraph of what you've fixed? Also, then a couple of bullet points on the specific code changes, and why they're there. These descriptions go into release notes, so it's best to keep them compressed and it also helps us complete reviews quicker.

shaurya703:
Thanks — trimmed to a short paragraph on the problem plus a bullet per file on what changed and why. Dropped the reproduction transcript and verification notes; happy to put any of it in review comments on the relevant lines if useful.

PR #11476, opened 2026-08-09
sqlfmt: omit generated columns from diff -r sql

### LINKED ISSUE #11445: `dolt diff -r sql` emits statements that assign generated columns
`dolt diff -r sql` writes generated columns into the statements it produces, and the database rejects any statement that assigns one, so the output cannot be replayed. Given:

```sql
CREATE TABLE t (id INT PRIMARY KEY, a INT, g INT GENERATED ALWAYS AS (a+1) STORED)
```

## Got
```sh
$ dolt diff -r sql
UPDATE `t` SET `a`=99,`g`=100 WHERE `id`=1;
INSERT INTO `t` (`id`,`a`,`g`) VALUES (2,20,21);
```

## Want
```sh
$ dolt diff -r sql
UPDATE `t` SET `a`=99 WHERE `id`=1;
INSERT INTO `t` (`id`,`a`) VALUES (2,20);
```

`SqlRowAsUpdateStmt` never filters generated columns, unlike `InsertStatementPrefix` and `SqlRowAsTupleString`. The schema handed to the diff writer by `RowIter` has an empty `Generated` field on every column.

https://github.com/dolthub/dolt/blob/8e5f4ba650ee6eb07eae6290bffabadace432451/go/cmd/dolt/commands/diff.go#L1143-L1154

Reproduced on dolthub/dolt@78b1425c6b.

vishnujayvel:
`dolt diff -r sql` no longer assigns generated columns in the `INSERT` and `UPDATE` statements it emits, so its output can be replayed. Previously the engine rejected those assignments; where the diff is emitted as `DELETE`+`INSERT` — keyless tables, and tables whose primary key is itself generated — replay deleted the old row and then failed to reinsert it, silently losing the row. `dolt_patch()`'s `statement` output is fixed by the same change.
- `schemaFromCreateTableStmt` now populates `Generated`, `Virtual`, and `OnUpdate` on `schema.Column`, so `IsGenerated()` is correct for schemas built from `SHOW CREATE TABLE`.
- `SqlRowAsUpdateStmt` skips generated columns, and `GenerateDataDiffStatement` drops them from `colsToUpdate`.
- STORED and VIRTUAL generated columns are both covered, on primary-keyed and keyless tables.

Fix dolthub/dolt#11445
Close dolthub/dolt#11699

### DIFF
--- a/go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
+++ b/go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
@@ -287,8 +287,8 @@ func SqlRowAsDeleteStmt(ctx *sql.Context, r sql.Row, tableName string, tableSch 
 // The row to change is keyed by the primary key columns of
 // |tableSch|, using their values from |r|.
 //
-// TODO(elianddb): Schema isn't recording column's Generated marker
-// correctly, so Column.IsGenerated doesn't filter.
+// Generated columns are skipped, matching InsertStatementPrefix and
+// SqlRowAsTupleString.
 func SqlRowAsUpdateStmt(ctx *sql.Context, r sql.Row, tableName string, tableSch schema.Schema, colsToUpdate *set.StrSet) (string, error) {
 	var b strings.Builder
 	b.WriteString("UPDATE ")
@@ -300,7 +300,7 @@ func SqlRowAsUpdateStmt(ctx *sql.Context, r sql.Row, tableName string, tableSch 
 	i := 0
 	seenOne := false
 	err := tableSch.GetAllCols().Iter(func(_ uint64, col schema.Column) (stop bool, err error) {
-		if colsToUpdate.Contains(col.Name) {
+		if colsToUpdate.Contains(col.Name) && !col.IsGenerated() {
 			if seenOne {
 				b.WriteRune(',')
 			}
--- a/go/libraries/doltcore/sqle/sqlfmt/row_fmt_test.go
+++ b/go/libraries/doltcore/sqle/sqlfmt/row_fmt_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/dolthub/dolt/go/libraries/doltcore/schema/typeinfo"
 	_ "github.com/dolthub/dolt/go/libraries/doltcore/sqle"
 	"github.com/dolthub/dolt/go/libraries/doltcore/sqle/sqlfmt"
+	"github.com/dolthub/dolt/go/libraries/utils/set"
 	"github.com/dolthub/dolt/go/store/types"
 )
 
@@ -81,6 +82,17 @@ func TestInsertStatementPrefixSkipsGeneratedCols(t *testing.T) {
 
 	require.NoError(t, err)
 	assert.Equal(t, "INSERT INTO `table_name` (`id`,`a`) VALUES ", prefix)
+}
+
+func TestSqlRowAsUpdateStmtSkipsGeneratedCols(t *testing.T) {
+	// See https://github.com/dolthub/dolt/issues/11445
+	sch := newGeneratedColSchema()
+	colsToUpdate := set.NewStrSet([]string{"a", "c"})
+
+	stmt, err := sqlfmt.SqlRowAsUpdateStmt(sql.NewEmptyContext(), sql.Row{int64(1), "x", "x!"}, "table_name", sch, colsToUpdate)
+
+	require.NoError(t, err)
+	assert.Equal(t, "UPDATE `table_name` SET `a`='x' WHERE `id`=1;", stmt)
 }
 
 func TestSqlRowAsTupleString(t *testing.T) {
--- a/go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
+++ b/go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
@@ -43,7 +43,11 @@ func GenerateDataDiffStatement(ctx *sql.Context, tableName string, sch schema.Sc
 		updatedCols := set.NewEmptyStrSet()
 		for i, diffType := range colDiffTypes {
 			if diffType != diff.None {
-				updatedCols.Add(sch.GetAllCols().GetByIndex(i).Name)
+				col := sch.GetAllCols().GetByIndex(i)
+				if col.IsGenerated() {
+					continue
+				}
+				updatedCols.Add(col.Name)
 			}
 		}
 		if updatedCols.Size() == 0 {
--- a/go/cmd/dolt/commands/diff.go
+++ b/go/cmd/dolt/commands/diff.go
@@ -1146,9 +1146,9 @@ func schemaFromCreateTableStmt(createTableStmt string) (schema.Schema, error) {
 			IsPartOfPK:    primaryCols[col.Name.Lowered()],
 			TypeInfo:      typeInfo,
 			Default:       defBuf.String(),
-			Generated:     "",    // TODO
-			OnUpdate:      "",    // TODO
-			Virtual:       false, // TODO
+			Generated:     genBuf.String(),
+			OnUpdate:      onUpBuf.String(),
+			Virtual:       col.Type.GeneratedExpr != nil && !bool(col.Type.Stored),
 			AutoIncrement: col.Type.Autoincrement == true,
 			Comment:       comment,
 		}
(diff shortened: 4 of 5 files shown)


elianddb:
@vishnujayvel Could you shorten the description to be similar to our teams' release notes? They come up as release notes and this change is relatively small.

vishnujayvel:
Done — shortened the description to match the team's release-note style.

Thanks again for the repro in #11445 — it named both failure points, which is why this landed as a small diff.

I've parked the long-form root cause and test plan below so it stays on the PR for review without ending up in the release notes.

<details>
<summary>Original description (root cause + test plan)</summary>


`dolt diff -r sql` was writing generated columns into the INSERT and
UPDATE statements it emits. The engine rejects any assignment into a generated
column, so the SQL patch could not be replayed. Where the diff is emitted as
DELETE+INSERT — keyless tables, and tables whose primary key is itself a
generated column — replaying a patch deleted the old row and then failed the
INSERT, silently losing the row.

The same emission path backs the `dolt_patch()` table function, so its
`statement` output is fixed by the same change.

Fixes #11445

Thanks to @elianddb (Elian) for the clear repro and for tracing both failure
points in #11445.

### Got

```sql
UPDATE `t` SET `a`=99,`g`=100 WHERE `id`=1;
INSERT INTO `t` (`id`,`a`,`g`) VALUES (2,20,21);
```

### Want / after this change

```sql
UPDATE `t` SET `a`=99 WHERE `id`=1;
INSERT INTO `t` (`id`,`a`) VALUES (2,20);
```

## Root cause (two parts)

1. **Schema construction on the CLI SQL-diff path.**
   `schemaFromCreateTableStmt` (`go/cmd/dolt/commands/diff.go`) already
   parsed `col.Type.GeneratedExpr` into a buffer, but wrote
   `Generated: ""` and `Virtual: false` into `schema.Column`. That made
   `Column.IsGenerated()` always false for schemas built from
   `SHOW CREATE TABLE`, so the existing INSERT filters in
   `InsertStatementPrefix` / `SqlRowAsTupleString` never fired on this path.

2. **UPDATE emission never filtered generated columns.**
   `SqlRowAsUpdateStmt` (`go/libraries/doltcore/sqle/sqlfmt/row_fmt.go` —
   not `diff.go`) only gated on `colsToUpdate`, unlike the INSERT helpers
   which already skip `col.IsGenerated()`. Even with an honest schema,
   UPDATE would still emit generated columns when their values changed
   alongside base columns.

## Fix

- Populate `Generated` from the already-formatted expression buffer and
  set `Virtual` from `GeneratedExpr != nil && !Stored` (and fill
  `OnUpdate` from its parallel buffer while touching the same struct).
- Skip `col.IsGenerated()` columns in `SqlRowAsUpdateStmt`, and drop them
  from `colsToUpdate` in `GenerateDataDiffStatement` so the UPDATE column
  set never includes them.

## Test plan

- Unit: `TestSqlRowAsUpdateStmtSkipsGeneratedCols` in
  `go/libraries/doltcore/sqle/sqlfmt/row_fmt_test.go`, mirroring
  `TestInsertStatementPrefixSkipsGeneratedCols`.
  - **Red on base:** expected
    `UPDATE ... SET a='x' WHERE id=1` vs actual
    `UPDATE ... SET a='x',c='x!' WHERE id=1`.
  - **Green at fix:** pass.
- Full package: `go test ./libraries/doltcore/sqle/sqlfmt/ -count=1` → ok.
- Integration: bats case
  `sql-diff: stored generated columns omitted from INSERT and UPDATE` in
  `integration-tests/bats/sql-diff.bats` — INSERT+UPDATE, assert `g` absent,
  replay after `dolt reset --hard`, assert final rows. **Passed** with the
  fixed binary on PATH, and fails against an unpatched binary. The full
  `sql-diff.bats` file was not run to completion locally (slow host); its
  first three cases (INSERT/UPDATE/DELETE reconcile) pass with the fixed
  binary, and CI covers the rest.
- Manual end-to-end repro (same as #11445 Got/Want): fixed binary omits
  `g` from both statements; replay succeeds; rows `(1,99,100)`, `(2,20,21)`.
  Also verified by direct comparison against an unpatched build: VIRTUAL
  column tables, keyless tables, generated-column primary keys, and
  `dolt_patch()` output are corrected; DELETE statements and non-generated
  tables are byte-identical.

## Notes

- Coverage: STORED and VIRTUAL generated columns are both fixed (VIRTUAL
  previously leaked into INSERT column lists as `NULL` values and broke
  replay the same way), on primary-keyed and keyless tables — including
  when the generated column is itself the primary key. Filters use
  `IsGenerated()` (any non-empty `Generated` string), consistent with the
  existing INSERT path.
- Scope of the newly populated fields: nothing on any diff render path
  reads `Generated`/`OnUpdate`/`Virtual` except the generated-column
  filters this PR relies on. ALTER-DDL statements come from
  `tds.AlterStmts` and the raw `SHOW CREATE TABLE` text, not from this
  parsed schema, so schema-diff output is unchanged (verified empirically
  at base and at this branch: ALTER/tabular/JSON outputs byte-identical).


</details>

go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
```diff
@@ -287,8 +287,8 @@ func SqlRowAsDeleteStmt(ctx *sql.Context, r sql.Row, tableName string, tableSch
 // The row to change is keyed by the primary key columns of
 // |tableSch|, using their values from |r|.
 //
-// TODO(elianddb): Schema isn't recording column's Generated marker
-// correctly, so Column.IsGenerated doesn't filter.
+// Generated columns are skipped, matching InsertStatementPrefix and
```
elianddb, line 290: No need to mention the other shared helpers, only the the Generated Columns behavior. Would also move to second paragraph.

go/libraries/doltcore/sqle/sqlfmt/schema_fmt.go
```diff
@@ -43,7 +43,11 @@ func GenerateDataDiffStatement(ctx *sql.Context, tableName string, sch schema.Sc
 		updatedCols := set.NewEmptyStrSet()
 		for i, diffType := range colDiffTypes {
 			if diffType != diff.None {
-				updatedCols.Add(sch.GetAllCols().GetByIndex(i).Name)
+				col := sch.GetAllCols().GetByIndex(i)
+				if col.IsGenerated() {
+					continue
+				}
+				updatedCols.Add(col.Name)
```
elianddb, line 50: Remove no need to duplicate filtering logic here.

go/libraries/doltcore/sqle/sqlfmt/row_fmt.go
```diff
@@ -300,7 +300,7 @@ func SqlRowAsUpdateStmt(ctx *sql.Context, r sql.Row, tableName string, tableSch
 	i := 0
 	seenOne := false
 	err := tableSch.GetAllCols().Iter(func(_ uint64, col schema.Column) (stop bool, err error) {
-		if colsToUpdate.Contains(col.Name) {
+		if colsToUpdate.Contains(col.Name) && !col.IsGenerated() {
 			if seenOne {
```
elianddb, line 304: There's an empty list edge case here `!seenOne` add an assertion to existing test.

go/libraries/doltcore/sqle/sqlfmt/row_fmt_test.go
```diff
@@ -83,6 +84,17 @@ func TestInsertStatementPrefixSkipsGeneratedCols(t *testing.T) {
...
 
+func TestSqlRowAsUpdateStmtSkipsGeneratedCols(t *testing.T) {
+	// See https://github.com/dolthub/dolt/issues/11445
+	sch := newGeneratedColSchema()
+	colsToUpdate := set.NewStrSet([]string{"a", "c"})
+
+	stmt, err := sqlfmt.SqlRowAsUpdateStmt(sql.NewEmptyContext(), sql.Row{int64(1), "x", "x!"}, "table_name", sch, colsToUpdate)
+
+	require.NoError(t, err)
+	assert.Equal(t, "UPDATE `table_name` SET `a`='x' WHERE `id`=1;", stmt)
```
elianddb, line 95: assertion for empty string

elianddb (changes requested):
A couple of requested changes to rm extra code, and uncovered edge case.

zachmu:
Thank you for your contribution, @vishnujayvel! I carried your commit forward into https://github.com/dolthub/dolt/pull/11782 and addressed Elian’s feedback: simplified the function documentation, removed duplicate filtering, and added handling and a regression assertion for updates containing only generated columns. The formatter tests and SQL diff integration suite pass. Closing this PR in favor of the replacement; your contribution is preserved in its commit history.

PR #11527, opened 2026-08-12
sqle: reclaim incomplete/abandoned database directories on create/clone instead of erroring

### LINKED ISSUE #11533: sql-server: an incomplete leftover directory from an interrupted CREATE DATABASE / DOLT_CLONE wedges every retry and cannot be cleared in-process
## Summary

When a `CREATE DATABASE` or `CALL DOLT_CLONE(...)` is interrupted (client crash, cancelled context, transport flap), it can leave a directory on disk that database discovery ignores but that still occupies the name. A later `CREATE`/`CLONE` of the same name then fails and **cannot recover in-process** — the only remedy is to delete the directory out of band, which races a live `dolt sql-server`.

Reproduced on **dolt 2.2.2** and present on current `main`.

## The three leftover shapes

An interrupted create/clone can leave a directory that discovery skips (logging `skipping in-progress database directory` / `skipping incomplete database directory`):

1. a directory carrying the in-progress marker `.dolt_safe_to_ignore`;
2. a directory with `.dolt` storage but no `.dolt/repo_state.json`;
3. a **bare** directory with no `.dolt` storage at all — a clone cancelled before any `.dolt` was written.

## Repro (standalone, no server required)

```bash
mkdir demo && cd demo && dolt init

# Simulate an interrupted create/clone leftover occupying the name "foo".
# Pick any one of the three shapes:
mkdir foo && touch foo/.dolt_safe_to_ignore   # shape 1: in-progress marker
# mkdir -p foo/.dolt/noms                      # shape 2: partial .dolt, no repo_state.json
# mkdir foo                                     # shape 3: bare dir, no .dolt

dolt sql -q "CREATE DATABASE foo;"
```

### Actual behavior

```
# shapes 1 and 2:
error on line 1 for query CREATE DATABASE foo: cannot create database foo: incomplete database directory from an interrupted create already exists; remove the directory and try again

# shape 3 (bare dir):
error on line 1 for query CREATE DATABASE foo: can't create database foo; database exists
```

`CALL DOLT_CLONE('<remote>', 'foo')` fails identically (it never reaches the remote). Meanwhile:

- `SHOW DATABASES` does **not** list `foo` (discovery ignores the leftover), and
- there is **no SQL** that clears it: `DROP DATABASE foo` and `CALL dolt_undrop('foo')` don't see it either.

The directory must be removed from the filesystem. On a running `dolt sql-server` that out-of-band `rm` races the server, and until then the name is permanently wedged.

### Expected behavior

A retried `CREATE`/`CLONE` of a name occupied **only** by an incomplete, non-servable leftover should succeed (reclaim the name) with no out-of-band filesystem surgery. A **complete, servable** database directory must of course still conflict with `database exists`.

## Impact

Any automated recovery that retries a clone after a failure — e.g. a drop → purge → reclone loop, or orchestration that reclones a replica after a transport error — gets permanently stuck: the target name is neither usable (discovery skips it) nor recreatable (`CREATE`/`CLONE` errors), and clearing it means stopping or racing the live server.

## Proposed fix

PR #11527 reclaims such a leftover at name-reservation time (under the provider lock): an incomplete, non-servable directory is moved into the `.dolt_dropped_databases` holding area (recoverable via `dolt_undrop`, purged by `dolt_purge_dropped_databases`) and the name reported available; a complete database still conflicts. It also extends `IsIncompleteDatabaseDir` to recognise the bare no-`.dolt` case (shape 3).

mccraigmccraig:
Fixes #11533

## Problem

An interrupted `CREATE DATABASE` or `CALL DOLT_CLONE(...)` can leave a directory on disk that discovery ignores:

- one carrying the in-progress marker (`.dolt_safe_to_ignore`),
- one with `.dolt` storage but no repo-state file, or
- a bare directory created before any `.dolt` storage was written (a clone cancelled very early).

A subsequent `CREATE`/`CLONE` of the same name then fails and cannot recover in-process: the leftover has to be removed out of band, which races a live server. On released dolt the marker / partial-`.dolt` cases fail with `incomplete database directory from an interrupted create already exists; remove the directory and try again`, and the bare no-`.dolt` case fails with `database exists` (Error 1007).

## Fix

At name-reservation time (under `p.mu`), when a directory occupies the name but is **not** a complete, servable database, move it into the `.dolt_dropped_databases` holding area (recoverable via `dolt_undrop`, purged by `dolt_purge_dropped_databases`) and report the name available. A complete database still conflicts. Because the quarantine runs under the provider lock it is atomic with the reservation, so it is safe against a concurrent same-name create/clone (exactly one wins).

`IsIncompleteDatabaseDir` now also reports a directory with **no `.dolt` storage at all** as non-servable, covering an early-cancelled clone interrupted before `.dolt` was written. Transient/ambiguous cases still bias to `ErrDatabaseExists`, so a healthy database is never quarantined.

## Repro

```sql
-- An interrupted create/clone left a directory squatting on the name "foo". Any of:
--   mkdir foo && touch foo/.dolt_safe_to_ignore   -- in-progress marker
--   mkdir -p foo/.dolt/noms                        -- partial .dolt, no repo_state.json
--   mkdir foo                                      -- bare, no .dolt

-- Before this change (released dolt) the retry is wedged, with no in-process recovery:
CREATE DATABASE foo;
--   ERROR: incomplete database directory from an interrupted create already exists; remove the directory and try again
--   (bare shape: can't create database foo; database exists)

-- After this change the retry reclaims the name:
CREATE DATABASE foo;   -- OK: foo is a fresh, usable database
CALL dolt_undrop();    -- lists foo: the reclaimed leftover was quarantined, not deleted
```

`CALL DOLT_CLONE('<remote>', 'foo')` behaves the same: it quarantines the leftover and proceeds to clone from the remote.

## Tests

- **Go unit tests** — `go test -race ./go/libraries/doltcore/sqle/`: new `DoltDatabaseProvider` tests cover all three leftover shapes (in-progress marker / partial `.dolt` without repo-state / no `.dolt`) across `CREATE`, `CLONE`, and the shared reserve gate; assert a complete database still returns `ErrDatabaseExists`; and run 8 concurrent same-name reservations that serialize to exactly one winner.
- **BATS integration tests** — `integration-tests/bats/`: `sql-create-database.bats` and `clone-drop.bats` cover reclaim over all three leftover shapes for `CREATE DATABASE` and `CALL DOLT_CLONE` (name reclaimed, resulting database usable, leftover quarantined and recoverable via `dolt_undrop`). The existing `clone-drop: in-progress marker lifecycle on clone` test — which asserted the removed error — is updated to the new reclaim behavior.

### DIFF
--- a/go/libraries/doltcore/env/environment.go
+++ b/go/libraries/doltcore/env/environment.go
@@ -304,15 +304,20 @@ func (dEnv *DoltEnv) Valid() bool {
 	return dEnv != nil && dEnv.CfgLoadErr == nil && dEnv.RSLoadErr == nil && dEnv.DBLoadError == nil && dEnv.HasDoltDir() && dEnv.HasDoltDataDir()
 }
 
-// IsIncompleteDatabaseDir reports whether the database directory rooted at |fs| holds a database whose creation
-// never finished, either because it carries the in-progress marker or because it has Dolt storage without the
-// repo state file that every complete database has. Discovery ignores such a directory, so it cannot be served.
+// IsIncompleteDatabaseDir reports whether the directory rooted at |fs| is NOT a complete, servable Dolt
+// database, so discovery ignores it and a create/clone may reclaim its name instead of colliding with it.
+// This holds when the directory carries the in-progress marker, has Dolt storage without the repo state
+// file that every complete database has, or holds no Dolt storage at all — a bare or non-database leftover,
+// e.g. an early-cancelled clone interrupted before .dolt was written. A complete database (Dolt storage
+// plus its repo state file, marker cleared) returns false.
 func IsIncompleteDatabaseDir(fs filesys.Filesys) bool {
 	if dbfactory.IsDatabaseInProgress(fs) {
 		return true
 	}
 	if exists, isDir := fs.Exists(dbfactory.DoltDir); !exists || !isDir {
-		return false
+		// No Dolt storage at all: not a database, only a bare or non-database directory squatting on the
+		// name. Discovery would never serve it, so it is safe to reclaim.
+		return true
 	}
 	exists, _ := fs.Exists(getRepoStateFile())
 	return !exists
--- a/go/libraries/doltcore/sqle/database_provider.go
+++ b/go/libraries/doltcore/sqle/database_provider.go
@@ -675,12 +675,6 @@ func commitTransaction(ctx *sql.Context, dSess *dsess.DoltSession, rsc *doltdb.R
 	return nil
 }
 
-var ErrIncompleteDatabaseDir = errors.New("incomplete database directory from an interrupted create already exists; remove the directory and try again")
-
-func NewErrIncompleteDatabaseDir(db string) error {
-	return fmt.Errorf("cannot create database %s: %w", db, ErrIncompleteDatabaseDir)
-}
-
 func (p *DoltDatabaseProvider) CreateCollatedDatabase(ctx *sql.Context, name string, collation sql.CollationID) (err error) {
 	// We have to validate the name before attempting to create a directory. If a directory contains a delimiter, when
 	// registerNewDatabase errors out a directory with the exact name will be leftover due to a process lock. This then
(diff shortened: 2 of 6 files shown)


coffeegoddd:
@mccraigmccraig thanks for the contribution! I've been discussing this with @elianddb and I think we are discussing some concurrency edge cases to handle and have some concerns about moving an incomplete database into the dropped databases dir. We are gonna follow up with a fix here once we nail down what the correct behavior should be here.

mccraigmccraig:
ofc... I don't really mind what happens to the partial dbs, I just didn't want to delete them!

elianddb:
Hi @mccraigmccraig, we're a little apprehensive when it comes to removing these incomplete directories. When it comes to concurrent writes, it's hard to tell whether a user is writing the directory or it's actually an abandoned directory. For now, we will be punting this issue until we can come up with a more comprehensive solution that addresses our concerns.

PR #11590, opened 2026-08-23
Distinguish updating a branch from creating one in push output

### LINKED ISSUE #11352: dolt push prints '[new branch]' for every successful push of an existing branch
## Summary

Every successful `dolt push` of an existing branch prints ` * [new branch]   main -> main` — the message git reserves for branch *creation*. Updates of long-existing remote branches are reported as new-branch creations on every push.

## Repro (dolt 2.2.2, file remote, no network needed)

```
$ dolt init; dolt sql -q "create table t (i int primary key)"; dolt add t; dolt commit -m c1
$ dolt remote add origin file:///tmp/remote
$ dolt push origin main
 * [new branch]          main -> main        # correct: first push
$ dolt sql -q "insert into t values (1)"; dolt add t; dolt commit -m c2
$ dolt push origin main
 * [new branch]          main -> main        # wrong: branch already exists remotely
```

`dolt branch -a` shows `remotes/origin/main` tracked locally before the second push; git in the same situation prints `<old>..<new>  main -> main`.

## Root cause

`go/libraries/doltcore/env/actions/remotes.go`, `DoPush`: the success message is hardcoded — any successful push that isn't a delete appends the `[new branch]` line. The adjacent TODO already acknowledges the missing plumbing ("we don't have sufficient information here to know what actually happened in the push"), and `buildReturnMsg` carries a second TODO ("updated branches info is missing"). So the gap is known — but the placeholder text is actively misleading rather than merely sparse: it asserts creation on every routine update. (It cost us a re-verification of a production push to DoltHub today: seeing `[new branch]` for a months-old `main` reads as "did I just push to the wrong remote/branch?".)

## Suggested fix (two tiers)

1. **Minimal, no plumbing:** before pushing, `PushToRemoteBranch` (or `DoPush` via the local remote-tracking ref `refs/remotes/<remote>/<branch>`, which is already maintained — `SetHeadToCommit(ctx, remoteRef, commit)` updates it after each push) can distinguish "destination branch existed" from "didn't", and print `[updated] main -> main` vs `[new branch] main -> main`. That fixes the misleading half without the full git-style output.
2. **Full:** plumb a result struct out of `push()`/`PushToRemoteBranch` carrying old/new head hashes, enabling the git-style `74476cf..080b073  main -> main` line the TODO asks for; the old head is cheaply available from the same remote-tracking ref (or from `destDB.ResolveBranchRoots` which the fast-forward path already calls).

rjc123:
Fixes #11352.

`dolt push` printed ` * [new branch]          main -> main` for every successful push of an existing branch, so routine updates were reported as branch creations. `[new branch]` is the message git reserves for creation, and seeing it for a long-lived branch reads as "did I just push to the wrong remote?" — the true outcome was only knowable by separately diffing local and remote heads.

## The change

The success message in `DoPush` was hardcoded: any successful non-delete push appended the `[new branch]` line. This samples whether the destination ref exists **before** the push — the push itself creates it — and prints `[updated]` when it did.

```
push 1  →  * [new branch]          main -> main     (branch is new)
push 2  →  * [updated]             main -> main
push 3  →  * [updated]             main -> main
```

Three things worth calling out for review:

1. **A failed existence check is swallowed, not returned.** It affects only the wording of a status line, and a push that would otherwise succeed must not fail because we could not describe it. On error the previous wording is used.
2. **This is the minimal fix (tier 1 in the issue).** Printing git-style commit ranges (`74476cf38..080b073e7`) still needs results plumbed out of `Push()`. I updated the existing TODO rather than removing it, so the remaining gap stays documented.
3. **`HasRef` is a single `GetDataset`**, which is negligible next to what `push()` already does against the destination (`ResolveBranchRoots`, `FastForwardWithWorkspaceCheck`, `SetHeadAndWorkingSetToCommit`).

## Tests

One existing bats assertion encoded the old behaviour. In `remotes-push-pull.bats`, "push --all with remote specified" pushes `branch1` twice and the second push asserted `[new branch]`; it now asserts `[updated]`. Every other `[new branch]` assertion in the suite is a genuine first push and is unchanged — including `sql-server-remotesrv.bats`, where `main:new_branch` really is a creation.

Added a regression test covering creation → update → update, and asserting `[new branch]` does *not* appear on the second push.

## What I verified, and what I did not

Verified against a binary built from this branch: creation prints `[new branch]`, subsequent pushes print `[updated]`, and the branch-deletion (`[deleted]`) and up-to-date paths are unchanged. I also replayed the modified test's exact scenario end to end and confirmed `* [updated]  branch1 -> branch1`.

`go test ./libraries/doltcore/env/...` passes; `gofmt` and `go vet` are clean.

**I was not able to run the bats suite locally** — `bats` isn't installed on this machine. The bats changes are reasoned from the surrounding test bodies and confirmed by reproducing the scenarios by hand against the real binary, but they have not been executed. Worth a careful look in CI.

### DIFF
--- a/go/libraries/doltcore/env/actions/remotes.go
+++ b/go/libraries/doltcore/env/actions/remotes.go
@@ -110,14 +110,31 @@ func Push(ctx context.Context, tempTableDir string, mode ref.UpdateMode, destRef
 func DoPush[C doltdb.Context](ctx C, pushMeta *env.PushOptions[C], statsCh chan pull.Stats) (returnMsg string, err error) {
 	var successPush, setUpstreamPush, failedPush []string
 	for _, targets := range pushMeta.Targets {
+		// Whether the destination branch already exists has to be sampled before the push, because the
+		// push itself creates the ref. It is used only to choose between the "[new branch]" and
+		// "[updated]" status lines below.
+		//
+		// A failure here is deliberately swallowed rather than returned: it affects the wording of a
+		// status message only, and a push that would otherwise succeed must not be failed because we
+		// could not describe it. On error we fall back to the previous behaviour.
+		destExisted := false
+		if targets.SrcRef != ref.EmptyBranchRef {
+			if exists, hasRefErr := pushMeta.DestDb.HasRef(ctx, targets.DestRef); hasRefErr == nil {
+				destExisted = exists
+			}
+		}
+
 		err = push(ctx, pushMeta.Rsr, pushMeta.TmpDir, pushMeta.SrcDb, pushMeta.DestDb, pushMeta.Remote, targets, statsCh)
 		if err == nil {
-			// TODO: we don't have sufficient information here to know what actually happened in the push. Supporting
-			// git behavior of printing the commit ids updated (e.g. 74476cf38..080b073e7  branch1 -> branch1) isn't
-			// currently possible. We need to plumb through results in the return from the Push(). Having just an error
-			// response is not sufficient, as there are many "success" cases that are not errors.
+			// TODO: we still don't have sufficient information here to print the commit ids updated the
+			// way git does (e.g. 74476cf38..080b073e7  branch1 -> branch1). That needs results plumbed
+			// through the return from Push(); an error response alone is not sufficient, as there are
+			// many "success" cases that are not errors. Branch creation vs update is distinguished above,
+			// so the status line is no longer actively misleading, but it remains coarser than git's.
 			if targets.SrcRef == ref.EmptyBranchRef {
 				successPush = append(successPush, fmt.Sprintf(" - [deleted]             %s", targets.DestRef.GetPath()))
+			} else if destExisted {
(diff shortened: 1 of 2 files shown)


elianddb:
Hi @rjc123, I'll be putting a pin on this contribution, we already have a branch in development for this.

rjc123:
Understood, thanks for saying so — I'll leave this as is and won't push anything further to it. Happy for it to be closed in favour of your branch.

Is there a rough timeframe for when yours is likely to land? Not chasing — it just helps to know whether that's the next release or a few cycles out, given #11352 currently misreports every routine push.

Two things that may save you a few minutes, whichever fix lands: #11352 still reproduces on 2.3.1, so it has survived two minor releases since it was filed; and `remotes-push-pull.bats` has one assertion that encodes the current behaviour ("push --all with remote specified" pushes branch1 twice and expects `[new branch]` on the second), so it will need updating. Every other `[new branch]` assertion in that suite is a genuine first push.

PR #11592, opened 2026-08-23
Say that a failed conjoin was deferred, not that something went fatal

### LINKED ISSUE #11591: Conjoin failure logs git's fatal auth text on a push that succeeded
## Summary

When background conjoin fails during a push, dolt logs the underlying error verbatim at `warning` level. For a `git+ssh` remote that error is git's own **fatal** text — `fatal: Could not read from remote repository`, plus a `run 'ssh-add <key>'` hint — and it is printed immediately before `Push complete.`

The push has succeeded. Conjoin is asynchronous, best-effort compaction whose failure does not affect it. But the output contains what reads as an authentication failure alongside a success, with nothing saying the two are unrelated.

## Observed

```
level=info    msg="beginning conjoin of database" pkg=store.noms upstream_len=<n>
level=warning msg="conjoin of database failed with error" pkg=store.noms
              error="Connection closed by <host> port <port>\r\nfatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
hint: dolt does not support interactive credential prompts
hint: configure git credentials (credential helper, token) for HTTPS remotes
hint: run `ssh-add <key>` to pre-load your key for SSH remotes
hint: ensure non-interactive auth is configured for GCM"
Push complete.
```

The push genuinely succeeded — local and remote were confirmed at the same hash immediately afterwards, via `dolt_branches` vs `dolt_remote_branches`.

## Why this is a problem rather than cosmetic

Every phrase in that error says "your credentials are broken": `fatal`, `Could not read from remote repository`, `make sure you have the correct access rights`, `run 'ssh-add <key>'`. None of it is true here. The transport authenticated fine; the connection was closed part-way through compaction, which on a store large enough for conjoin to fire can run long enough for that to happen.

The consequences:

- **An operator cannot tell it from a real failure** without separately diffing local and remote heads after every push.
- **An automated caller cannot either.** A script or agent reading the tail of that output has no way to classify it. It will either raise false alarms on healthy pushes, or — worse — learn to ignore `fatal:` and `Could not read from remote repository`, which is exactly the text that matters when auth *is* broken.
- **It sends you to the wrong place.** The suggested remedy is `ssh-add`, which has nothing to do with the cause. In our case it was initially diagnosed as a network-level block before anyone noticed the next push simply worked.

## Mechanism

`startConjoinIfRequired` fires when `conjoinRequired(nbs.tables)` is true, logs `beginning conjoin of database`, then deliberately runs the operation on `context.Background()` so it outlives the caller:

https://github.com/dolthub/dolt/blob/0d6b06015867dd0cabd2eb51ff222691aae992dc/go/store/nbs/store.go#L354-L378

On failure `finalizeConjoin` attaches the raw error and logs it at warn:

https://github.com/dolthub/dolt/blob/0d6b06015867dd0cabd2eb51ff222691aae992dc/go/store/nbs/store.go#L385-L397

So the severity is already correct internally — it *is* a warning, and the code treats it as non-fatal by returning cleanly and retrying on a later push. The problem is that the rendered message contradicts that: it carries another tool's fatal vocabulary and says nothing about being deferred, retryable, or independent of the push result.

## Suggested fix

Any one of these would resolve it; the first is the smallest:

1. **Make the message self-describing.** Say that compaction was deferred, will retry on a later push, and does not affect whether the push succeeded. The wording is what misleads, not the fact of logging it.
2. **Don't surface the embedded git fatal text at this level.** Summarise the cause (connection closed during compaction) and keep the raw output behind a debug/verbose flag.
3. **A structured or quiet output mode** that callers can parse deterministically, so success and failure are machine-distinguishable rather than inferred from prose.

## Environment and a caveat worth stating

dolt 2.3.1, pushing to a `git+ssh` remote, on a store large enough that conjoin runs during a push.

**I have not reproduced this with the `dolt` CLI directly.** It was observed through an application embedding dolt's storage layer. Pushing a much smaller store over the same kind of remote is clean, which is consistent with conjoin simply not firing at that size. The code path above is dolt's own, and I would expect the CLI to behave identically wherever `conjoinRequired` returns true and the transport can drop mid-operation — but I can't claim to have seen it, so it's worth confirming before acting on the wording.

Happy to put up a PR for option 1 if that's the direction you'd want.

rjc123:
Reword the conjoin failure log line to say that compaction was deferred and will be retried, and add a `deferred` field so log consumers can classify it without parsing prose.
Fixes #11591
Close #11624
Blocked by #11624

### DIFF
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -392,15 +392,6 @@ func (nbs *NomsBlockStore) finalizeConjoin(ctx context.Context, err error) {
 	}()
 
 	if err != nil {
-		// Conjoin is best-effort compaction running on a goroutine that outlives the caller, and at
-		// this point nothing has been changed — the store is untouched and conjoin is retried on a
-		// later write. The attached error is the raw underlying one, though, and for a git+ssh remote
-		// that is git's own text: "fatal: Could not read from remote repository", an access-rights
-		// sentence and an `ssh-add` hint. Emitted next to the push it coincided with, that reads as an
-		// authentication failure on an operation which in fact succeeded.
-		//
-		// So say what happened rather than only that something failed. The severity was already right;
-		// the wording was borrowed from a tool whose failure means something else entirely.
 		nbs.logger.WithError(err).WithField("deferred", true).
 			Warn("background compaction (conjoin) failed and was deferred; it will be retried on a later write, and the operation that triggered it was unaffected")
 		return
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -392,7 +392,17 @@ func (nbs *NomsBlockStore) finalizeConjoin(ctx context.Context, err error) {
 	}()
 
 	if err != nil {
-		nbs.logger.WithError(err).Warn("conjoin of database failed with error")
+		// Conjoin is best-effort compaction running on a goroutine that outlives the caller, and at
+		// this point nothing has been changed — the store is untouched and conjoin is retried on a
+		// later write. The attached error is the raw underlying one, though, and for a git+ssh remote
+		// that is git's own text: "fatal: Could not read from remote repository", an access-rights
+		// sentence and an `ssh-add` hint. Emitted next to the push it coincided with, that reads as an
+		// authentication failure on an operation which in fact succeeded.
+		//
+		// So say what happened rather than only that something failed. The severity was already right;
+		// the wording was borrowed from a tool whose failure means something else entirely.
+		nbs.logger.WithError(err).WithField("deferred", true).
+			Warn("background compaction (conjoin) failed and was deferred; it will be retried on a later write, and the operation that triggered it was unaffected")
 		return
 	}
 
--- a/go/store/nbs/store.go
+++ b/go/store/nbs/store.go
@@ -392,8 +392,7 @@ func (nbs *NomsBlockStore) finalizeConjoin(ctx context.Context, err error) {
 	}()
 
 	if err != nil {
-		nbs.logger.WithError(err).WithField("deferred", true).
-			Warn("background compaction (conjoin) failed and was deferred; it will be retried on a later write, and the operation that triggered it was unaffected")
+		nbs.logger.WithError(err).WithField("deferred", true).Warn("background conjoin (compaction) deferred")
 		return
 	}
 


elianddb:
Hi @rjc123, thank for addressing this ambiguity. It's a small change, so let's keep your description to one sentence max. You can keep the context of your change in the comments.

I think the verbosity of the inline comment is a bit much too. This change is self-explanatory, so please remove it.

Your addition of the `deferred` label is good!

go/store/nbs/store.go
```diff
@@ -392,7 +392,17 @@ func (nbs *NomsBlockStore) finalizeConjoin(ctx context.Context, err error) {
 	}()
 
 	if err != nil {
-		nbs.logger.WithError(err).Warn("conjoin of database failed with error")
+		// Conjoin is best-effort compaction running on a goroutine that outlives the caller, and at
```
elianddb: Drop this comment. Sure, it provides context, but the change is self-explanatory.

go/store/nbs/store.go
```diff
@@ -392,7 +392,17 @@ func (nbs *NomsBlockStore) finalizeConjoin(ctx context.Context, err error) {
...
+		// this point nothing has been changed — the store is untouched and conjoin is retried on a
+		// later write. The attached error is the raw underlying one, though, and for a git+ssh remote
+		// that is git's own text: "fatal: Could not read from remote repository", an access-rights
+		// sentence and an `ssh-add` hint. Emitted next to the push it coincided with, that reads as an
+		// authentication failure on an operation which in fact succeeded.
+		//
+		// So say what happened rather than only that something failed. The severity was already right;
+		// the wording was borrowed from a tool whose failure means something else entirely.
+		nbs.logger.WithError(err).WithField("deferred", true).
```
elianddb: Logrus encourages [structured logging through fields for quick searching](https://github.com/sirupsen/logrus#fields). Keep the message string short so it can act as an easy event identifier. I'd recommend:
```
Warn("background conjoin (compaction) deferred")
```

elianddb (changes requested):

rjc123:
Thanks — all three done. Inline comment removed, description cut to one sentence, and the `deferred` field kept. Sent as a follow-up commit rather than a force-push so the review history stays readable.

Context for the change, moved out of the code as requested:

The raw error attached at this call site is often git's own text when the remote is `git+ssh` — `fatal: Could not read from remote repository`, an access-rights sentence and an `ssh-add` hint. Printed next to the operation it coincided with, it reads as an authentication failure on a push that in fact succeeded. Nothing has been changed at that point: the store is untouched and conjoin is retried on a later write, so the warning severity was already correct — only the wording was borrowed from a tool whose failure means something else.

Scoped to this one call site deliberately. The three other `Warn` calls in `finalizeConjoin` share the wording but not the consequences — one fires after the manifest has already advanced, and one after conjoin has succeeded — so describing them the same way would be wrong.

elianddb:
Hi @rjc123, could you also address the other review comment that's unresolved.

go/store/nbs/store.go
rjc123: Done in b0c3d696a — adopted your wording verbatim. The `deferred` field carries the classification, so keeping the message short as an event identifier loses nothing.

rjc123:
An observation from outside the project, offered only because I went looking to check
whether the integration failure was mine.

It doesn't appear to be, and the evidence is slightly more useful than "probably a flake":
commit `b0c3d696a` ran `ci-binlog-tests` twice with opposite results — success on the fork
branch at 08:26 UTC, failure on `elian/11591` at 19:28 UTC. Same SHA, same tree; the only
change in it is a log message in `store/nbs`.

Widening it out, `ci-binlog-tests` failed 8 of 87 completed runs over the last 11 days
(~9%), spread across 8 different branches rather than concentrated in one. In the failures
whose logs are still readable the failing test isn't consistent —
`TestBinlogPrimary_InitializeGTIDPurged` here, `TestBinlogPrimary_Reset` and a broader
`TestBinlog` elsewhere — which suggests the lane rather than a single test.

I couldn't find an open issue tracking it, but I may simply have missed it. Happy to file
one with the run links if that would be useful, or to leave it entirely with you — you have
far better context on whether this is known.

Nothing needed on this PR as far as I can see. Flagging it only because a ~9% failure rate
on a pre-merge gate is the kind of thing that gets absorbed as background noise from the
inside, and it was visible to me precisely because I had no priors about it.

Relevant to https://github.com/dolthub/dolt/pull/11624

elianddb (approved):

elianddb:
> An observation from outside the project, offered only because I went looking to check whether the integration failure was mine.
> 
> It doesn't appear to be, and the evidence is slightly more useful than "probably a flake": commit `b0c3d696a` ran `ci-binlog-tests` twice with opposite results — success on the fork branch at 08:26 UTC, failure on `elian/11591` at 19:28 UTC. Same SHA, same tree; the only change in it is a log message in `store/nbs`.
> 
> Widening it out, `ci-binlog-tests` failed 8 of 87 completed runs over the last 11 days (~9%), spread across 8 different branches rather than concentrated in one. In the failures whose logs are still readable the failing test isn't consistent — `TestBinlogPrimary_InitializeGTIDPurged` here, `TestBinlogPrimary_Reset` and a broader `TestBinlog` elsewhere — which suggests the lane rather than a single test.
> 
> I couldn't find an open issue tracking it, but I may simply have missed it. Happy to file one with the run links if that would be useful, or to leave it entirely with you — you have far better context on whether this is known.
> 
> Nothing needed on this PR as far as I can see. Flagging it only because a ~9% failure rate on a pre-merge gate is the kind of thing that gets absorbed as background noise from the inside, and it was visible to me precisely because I had no priors about it.
> 
> Relevant to #11624

This is usually under @fulghum domain, I think we'll wait and see a little longer.

PR #11596, opened 2026-08-24
Fix macOS-only bats CI failures

fulghum:
Fixes the long-standing macOS-only bats CI failures (verified: 38 → 0 failures across two full nightly-workflow runs).

- Skip docker-entrypoint.bats/tzdata.bats for MacOS (no docker available on PATH in MacOS runners)
- Fix BSD awk/grep incompatibilities in sql-diff.bash and GNU-only `sed -i` in sql-pull.bats
- Disable the detached `dolt send-metrics` process during bats runs (raced with test teardown cleanup)
- Default `DOLT_PAGER=cat` (less races with expect scripts typing ahead, dropping keystrokes)

### DIFF
--- a/integration-tests/bats/diff-system.expect
+++ b/integration-tests/bats/diff-system.expect
@@ -2,6 +2,9 @@
 
 set timeout 5
 set env(NO_COLOR) 1
+# less reads keypresses from /dev/tty directly, racing with characters this script types
+# ahead into the pty for the next command and dropping them.
+set env(DOLT_PAGER) "cat"
 
 source  "$env(BATS_CWD)/helper/common_expect_functions.tcl"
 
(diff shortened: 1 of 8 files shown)


fulghum:
Confirmed fix with a full macOS bats CI run: https://github.com/dolthub/dolt/actions/runs/32756894838 — 3670/3670 tests passed.

fulghum:
Confirmed with the scoped pager fix: https://github.com/dolthub/dolt/actions/runs/32767179402 — macOS bats 3670/3670 passed, and all Ubuntu CI checks (including `Bats tests (ubuntu-22.04)`) are green.

macneale4 (approved):
LGTM

PR #11598, opened 2026-08-25
feat: add orcarouter provider to dolt assist

kuswardhanietidims-svg:
The `dolt assist` command is Dolt's chat assistant for running dolt commands and queries in plain language. This PR adds a named [OrcaRouter](https://www.orcarouter.ai) provider via `--provider orcarouter`, mirroring the existing OpenAI wiring: it reads `ORCAROUTER_API_KEY`, targets `https://api.orcarouter.ai/v1/chat/completions`, and defaults to the `orcarouter/auto` model. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents — like OpenRouter it exposes a provider/model namespace across many models, but also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint, so Dolt users can use that stack without treating it as an anonymous base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes. OpenAI stays the default; no behavior change. I also fixed a pre-existing bug in `getJsonPrompt` that emitted a leading comma (invalid JSON) on the first request.

Verification: `go build` + `go vet` pass, `gofmt` clean, live-tested `--provider orcarouter` against OrcaRouter (200). `go test ./cmd/dolt/commands/` passes; the only failure is pre-existing `TestSignAndVerifyCommit` (needs `gpg`, not installed here).

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/go/cmd/dolt/commands/assist.go
+++ b/go/cmd/dolt/commands/assist.go
@@ -40,14 +40,23 @@ import (
 	"github.com/dolthub/dolt/go/libraries/utils/argparser"
 )
 
+const (
+	providerOpenAi     = "openai"
+	providerOrcaRouter = "orcarouter"
+	openAiApiBase      = "https://api.openai.com/v1"
+	orcaRouterApiBase  = "https://api.orcarouter.ai/v1"
+	openAiDefaultModel = "gpt-3.5-turbo"
+	orcaRouterModel    = "orcarouter/auto"
+)
+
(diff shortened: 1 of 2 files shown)


zachmu (approved):
LGTM, thanks for the contribution!

PR #11640, opened 2026-08-31
[no-release-notes] Fix macOS vector index BATS data generation

fulghum:
BSD `seq -s,` emits a trailing separator, which makes the out-of-line vector GC test generate invalid JSON on macOS. Build the three test vectors with `paste -sd, -` so the generated arrays have separators only between values.

Validated the focused regression and the complete `vector-index.bats` file on macOS.


fulghum:
macOS BATS CI passed: [Test Bats MacOS run 33412157190, attempt 2](https://github.com/dolthub/dolt/actions/runs/33412157190/attempts/2).

elianddb (approved):
lgtm!

Issue #10525, opened 2026-02-18
Add Some sort of Git Visibility to Git Remotes


timsehn:
The current implementation of Git remotes creates a "hidden" ref on the remote and pushes Dolt files to that ref. You can only see it in Git by cloning the Git repository and running `git ls-remote`. 

I think some users are going to want to see that there is a Dolt remote in their Git remote. Moreover, I think some users are going to want to know and act on Dolt metadata from their Git repository.

I'm not sure if the following options are the default or are configurable. But here are some user goals for this.

1. I'm a Git remote user and I want to know if my Git repository has a remote without cloning it and running `ls-remote`. I want some indicator in the GitHub or GitLab UI. Potential solution: Make a tracking branch (maybe called `dolt-<remotename>-remote`) in the Git repository for the Dolt database. I believe this would allow users to see that there is a Dolt remote in their repository.  
2. I'm a Git remote user and I want to see what branches are available in the Dolt remote. Posible solution. Manifest a file on push called `dolt-branches.(txt|json|yaml)`. This is dependent on fixing user goal 1.
3. I'm a Git remote user and I want to know what commits each branch is on so I can tie Git commits to Dolt commits for say CI. Again, we can manifest a file on push.
4. I want to know the log of the Dolt remote so I can act on changes in my Git code. Again, we can manifest a file on push.

As you can see there aqre a ton of possibilities here.

deg:
Just to add a strawman into this, here's a quick script that I (well, Claude) wrote that met my personal single-user needs.
Better than nothing and might already help other folks.
https://github.com/deg/bin/blob/master/dolt-remote-check

Revised: Now at https://github.com/deg/beads-utils/blob/main/bd-dolt-check

Free for any use.

PR #11650, opened 2026-09-01
fix: push git-remote info branch outside refs/heads/ (Fixes #11648)

### LINKED ISSUE #11648: git-remote: __dolt_remote_info__ under refs/heads/ makes concurrent `git fetch` fail with "incorrect old value provided"
## Summary

Dolt's git-remote backend keeps its data outside `refs/heads/` — `refs/dolt/data`, plus UUID-owned `refs/dolt/remotes/…` and `refs/dolt/blobstore/…` in the local cache repo. The **info branch does not follow that rule**. `pushInfoBranch` in `go/store/blobstore/git_blobstore.go` maintains it locally at `refs/dolt/info/<name>` but force-pushes it to the remote as a real branch:

```go
infoRef := "refs/heads/" + gbs.infoBranch
localInfoRef := "refs/dolt/info/" + gbs.infoBranch
...
_ = gbs.api.ForcePushRef(ctx, gbs.remoteName, localInfoRef, infoRef)
```

Because it lands under `refs/heads/`, the default fetch refspec `+refs/heads/*:refs/remotes/origin/*` sweeps it into **every bare `git fetch`** by every consumer of that repository — including people who have no idea Dolt is involved.

That is normally just clutter. It stops being clutter because the ref is force-pushed on **every** data write, which makes it the highest-churn ref in the repository. Git updates each remote-tracking ref with a compare-and-swap whose expected-old-value is read when the fetch begins, so two fetches overlapping in the same clone race it and the loser gets:

```
error: fetching ref refs/remotes/origin/__dolt_remote_info__ failed: incorrect old value provided
```

and `git fetch` **exits non-zero**.

## Why I think this is a real bug and not just untidiness

`RemoteTrackingRef` in `go/store/blobstore/git_refs.go` carries this comment:

> Each instance gets its own tracking ref to avoid concurrent git-fetch races when multiple blobstore instances share the same cache repo.

So concurrent-git-fetch races on a high-churn ref are a known hazard that Dolt already engineered around — with UUID-owned refs — for its own cache repo. The info branch exports that same hazard into the **user's** repository, where the user has no equivalent mitigation available and no indication of where the failure is coming from.

## Reproduction

A clone whose remote has a `main` and an `__dolt_remote_info__`. `main` advances once per round (an ordinary commit landing) while `__dolt_remote_info__` churns continuously (what every data push does); 8 concurrent `git fetch origin` per round, 15 rounds:

| | fetches | exited non-zero | attributed to `__dolt_remote_info__` |
|---|---|---|---|
| stock refspec | 120 | 98 | 98 |
| `git config --add remote.origin.fetch "^refs/heads/__dolt_remote_info__"` | 120 | 0 | 0 |

The ratio matters: with every ref churning equally, ordinary branches lose the race too. Modelling the real ratio — one branch update against continuous info-branch churn — puts essentially 100% of the failures on the info branch.

Over an SSH remote the race is much rarer (network latency staggers the fetches enough to serialise the ref write), which is why this presents as maddeningly intermittent rather than reproducible.

## Impact

The fetch does not abort — refs it already wrote stay written, so `origin/main` still updates. The damage is the **non-zero exit**: `git fetch && …` pipelines take an error path, CI marks a step failed, and automated agents read it as "remote unreachable" and fall back to stale local state. All from a ref that has nothing to do with the work in front of them, in a repository where nobody has knowingly touched anything Dolt-related.

## Workaround, and why it is a poor default

Per clone, per remote:

```bash
git config --add remote.origin.fetch "^refs/heads/__dolt_remote_info__"
```

This works (git ≥ 2.29 for negative refspecs). But it is **local git config**: not carried by `.gitattributes`, not carried by `git clone`, not carried to collaborators. Every clone that ever fetches the repository has to be fixed by hand or by a bespoke setup script, and any clone that misses it keeps the bug silently. We hit this across four clones on three operating systems and had to write an installer for it.

I realise `DOLT_REMOTE_INFO_BRANCH` can disable the branch outright — the constant's doc comment says so explicitly. Two notes on that: it is an all-or-nothing opt-out of a feature that is otherwise useful (#10525), and the documented "set to an empty string" mechanism is not reachable from a Windows shell, where assigning an empty value deletes the variable rather than setting it empty, so `os.LookupEnv` returns `false` and you get the default back. A whitespace value appears to work via the `strings.TrimSpace` in `ResolveInfoBranch`, but I have not tested that path.

## Requested change

1. **Push the info branch outside `refs/heads/`** — the natural target is `refs/dolt/info/<name>`, which is *already* the local ref name in `pushInfoBranch`, so the remote layout would simply become symmetric with the local one and with `refs/dolt/data`. This keeps the visibility goal of #10525: the ref still shows in `git ls-remote` and in forge ref browsers; it just stops being fetched by default and stops colliding.
2. Or, less good but much smaller: **let `DOLT_REMOTE_INFO_BRANCH` accept a full ref path** when the value begins with `refs/`. Today it only renames the branch *within* `refs/heads/`, so an operator cannot move it out of the way themselves.

## Versions

`bd` (beads) 1.2.2 embedding Dolt; git 2.53.0.windows.1; remote is a self-hosted Gitea over SSH. Also reproduced against a plain local bare repo, so it is not Gitea-specific. Source references are `dolthub/dolt@main` as of 2026-08-31.

## Related

- #10525 — the design rationale for making the info branch visible. This report is meant as "keep the visibility, change the namespace", not as an argument against that feature.
- gastownhall/beads#5266 — documents the `refs/heads/` placement and independently audits the info branch as write-only and safe to delete, but does not mention the fetch race.

chiliec:
## Summary

`pushInfoBranch` (`go/store/blobstore/git_blobstore.go`) force-pushes the info branch to `refs/heads/<name>` on **every data write**. Because it lands under `refs/heads/`, the default fetch refspec `+refs/heads/*:refs/remotes/origin/*` sweeps it into **every** bare `git fetch` by every consumer of the repository — including people who have no idea Dolt is involved.

Since it is force-pushed on every write, it is the highest-churn ref in the repo. Git updates each remote-tracking ref with a compare-and-swap whose expected-old-value is read when the fetch begins, so two fetches overlapping in the same clone race it and the loser gets:

```
error: fetching ref refs/remotes/origin/__dolt_remote_info__ failed: incorrect old value provided
```

and `git fetch` **exits non-zero** — for a ref that has nothing to do with the consumer's work. This is the failure mode reported in #11648.

## Fix

Push the info branch to `refs/dolt/info/<name>` instead of `refs/heads/<name>`. That is already the local ref name in `pushInfoBranch`, so the remote layout simply becomes symmetric with the local one and with `refs/dolt/data`. It keeps the visibility goal of #10525 — the ref still shows in `git ls-remote` and forge ref browsers — but it is no longer fetched by default and no longer collides with concurrent fetches.

The info branch is write-only on Dolt's side (only `pushInfoBranch` touches it; nothing reads it back from `refs/heads/`), so moving the push target is safe.

Fixes #11648.

## Testing

Added `go/store/blobstore/git_blobstore_info_branch_test.go` (real git, mirrors the existing `requireGitOnPath` / `newRemoteAndLocalRepos` harness):

- `TestGitBlobstore_InfoBranch_PushedOutsideRefsHeads` — after a data write + manifest flush, the info ref exists on the remote under `refs/dolt/info/__dolt_remote_info__` and **not** under `refs/heads/`.
- `TestGitBlobstore_InfoBranch_NotFetchedByDefault` — a fresh clone doing a default `git fetch origin` succeeds and does **not** pull the info branch into `refs/remotes/origin/`.

Both fail before the change and pass after (verified by reverting the one-line fix):

```
# before the fix
--- FAIL: TestGitBlobstore_InfoBranch_PushedOutsideRefsHeads
    info ref refs/dolt/info/__dolt_remote_info__ should exist on the remote; ... 'not a valid ref'
--- FAIL: TestGitBlobstore_InfoBranch_NotFetchedByDefault
    info branch should not be fetched into refs/remotes/origin/ by default

# after the fix
ok  github.com/dolthub/dolt/go/store/blobstore
```

Full package: `go test ./store/blobstore/` — `ok` (12.3s). `gofmt` and `go vet ./store/blobstore/` clean. Verified locally with Go 1.26.

---

First-time contributor here. The issue also suggested an alternative (letting `DOLT_REMOTE_INFO_BRANCH` accept a full `refs/...` path); I went with moving the default because it fixes every existing deployment without operator action, but happy to adjust if you'd prefer the opt-in shape or a migration note for remotes that already carry the old `refs/heads/` branch.

### DIFF
--- a/go/store/blobstore/git_blobstore.go
+++ b/go/store/blobstore/git_blobstore.go
@@ -829,8 +829,16 @@ func (gbs *GitBlobstore) pushInfoBranch(ctx context.Context, headOID git.OID) {
 		return
 	}
 
-	infoRef := "refs/heads/" + gbs.infoBranch
-	localInfoRef := "refs/dolt/info/" + gbs.infoBranch
+	// Push the info branch outside refs/heads/ so it is not swept into every
+	// consumer's default `git fetch` (+refs/heads/*:refs/remotes/origin/*).
+	// Keeping it under refs/dolt/info/ mirrors the local ref name and the
+	// refs/dolt/data layout: it stays visible in `git ls-remote` and forge ref
+	// browsers, but no longer collides with concurrent fetches. Landing this
+	// high-churn, force-pushed ref under refs/heads/ made overlapping fetches in
+	// the same clone race git's compare-and-swap and fail with
+	// "incorrect old value provided", aborting `git fetch` with a non-zero exit.
+	infoRef := "refs/dolt/info/" + gbs.infoBranch
+	localInfoRef := infoRef
 
 	content := formatDoltRemoteInfo(gbs.remoteRef, headOID, time.Now())
 
--- a/go/store/blobstore/git_blobstore_info_branch_test.go
+++ b/go/store/blobstore/git_blobstore_info_branch_test.go
@@ -1,0 +1,112 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blobstore
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	git "github.com/dolthub/dolt/go/store/blobstore/internal/git"
+	"github.com/dolthub/dolt/go/store/testutils/gitrepo"
+)
+
+// The info branch is force-pushed on every data write. It must NOT land under
+// refs/heads/, or the default fetch refspec (+refs/heads/*:refs/remotes/origin/*)
+// sweeps this high-churn ref into every consumer's `git fetch`, where concurrent
+// fetches race git's compare-and-swap and fail with "incorrect old value
+// provided". It must land under refs/dolt/info/ instead (still visible in
+// `git ls-remote`, but not fetched by default). See dolthub/dolt#11648.
+func TestGitBlobstore_InfoBranch_PushedOutsideRefsHeads(t *testing.T) {
+	requireGitOnPath(t)
+
+	ctx := context.Background()
+	remoteRepo, localRepo, _ := newRemoteAndLocalRepos(t, ctx)
+	_, err := remoteRepo.SetRefToTree(ctx, DoltDataRef, nil, "seed empty")
+	require.NoError(t, err)
+
+	bs, err := NewGitBlobstoreWithOptions(localRepo.GitDir, DoltDataRef, GitBlobstoreOptions{
+		RemoteName: "origin",
+		Identity:   testIdentity(),
+		InfoBranch: DefaultInfoBranch,
+	})
+	require.NoError(t, err)
+
+	// A data write + manifest flush triggers the best-effort info-branch push.
+	_, err = PutBytes(ctx, bs, "k", []byte("v\n"))
+	require.NoError(t, err)
+	_, err = bs.CheckAndPutManifest(ctx, "", []byte("manifest\n"))
+	require.NoError(t, err)
+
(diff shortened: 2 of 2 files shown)


coffeegoddd:
@chiliec thanks for the contribution, unfortunately this pr isn't the correct solution. see this comment: https://github.com/dolthub/dolt/issues/11648#issuecomment-5516276809

the purpose of the info branch being on refs/heads/ is so that it will render in the git remote's UI, and moving it to refs/dolt/info, breaks this.

PR #11666, opened 2026-09-02
[no-release-notes] Close superseded dependency bumps regardless of CI

fulghum:
Simplify stale dependency-bump cleanup: once a newer bump PR is opened, close older bump PRs regardless of their CI result. The `keep-alive` label remains the explicit way to preserve an older bump.


zachmu (approved):
LGTM

PR #11736, opened 2026-09-05
[no-release-notes] ci: optimize bats unix test and fix expect race

elianddb:


zachmu (approved):
LGTM assuming this actually makes things faster

PR #11738, opened 2026-09-07
fix: clear upstreams when removing a remote

### LINKED ISSUE #11529: `dolt remote remove` leaves an upstream config behind
A branch can record an upstream, the branch that's merged into it by default, or that it's rebased onto. In Dolt, it's stored under `.dolt/repo_state.json` as two fields: the remote and a ref naming the branch on that remote. `dolt remote remove` clears the remote name but keeps the ref, making it ambiguous with a local branch upstream config.

```sh
$ dolt push -u origin b1:other
$ dolt remote remove origin
$ cat .dolt/repo_state.json
  "branches": {
    "b1": {
      "head": "refs/heads/other",
      "remote": ""
    }
  }

$ dolt branch --set-upstream-to other b1
branch 'b1' set up to track 'other'
$ cat .dolt/repo_state.json
  "branches": {
    "b1": {
      "head": "refs/heads/other",
      "remote": ""
    }
  }
```

When the upstream lives in the local repository, Git indicates it with a '.' character: [the relative path setting](https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchnamemerge).

```sh
$ git config --get-regexp ^branch\.b1\.
branch.b1.remote origin
branch.b1.merge refs/heads/other
$ git remote remove origin
$ git config --get-regexp ^branch\.b1\.
(no output)

$ git branch --set-upstream-to=other2 b1
$ git config --get-regexp ^branch\.b1\.
branch.b1.remote .
branch.b1.merge refs/heads/other2
```

I've left a TODO in `UpstreamRef` since this doesn't allow us to discern if we want the ref.
https://github.com/dolthub/dolt/blob/92de41bbaef077eb6db53bce2b5d5f8ac001891f/go/libraries/doltcore/env/remotes.go#L811-L816

CBOSSX:
`dolt remote remove` now clears branch upstream configurations that reference the deleted remote, matching Git and preventing stale merge refs in `repo_state.json`. The regression preserves branches tracking other remotes; the `env/...` and CLI command suites, focused race test, vet, formatting, diff check, and CLI build pass.
Fixes #11529.
Close #11760

### DIFF
--- a/go/libraries/doltcore/env/repo_state.go
+++ b/go/libraries/doltcore/env/repo_state.go
@@ -232,12 +232,17 @@ func (rs *RepoState) AddRemote(r Remote) {
 
 func (rs *RepoState) RemoveRemote(r Remote) {
 	rs.Remotes.Delete(r.Name)
-	if rs.Branches == nil {
+	ClearTrackingBranches(rs.Branches, r.Name)
+}
+
+// ClearTrackingBranches removes tracking configurations for the named remote.
+func ClearTrackingBranches(branches *concurrentmap.Map[string, BranchConfig], remoteName string) {
+	if branches == nil {
 		return
 	}
-	for branchName, branch := range rs.Branches.Snapshot() {
-		if branch.Remote == r.Name {
-			rs.Branches.Delete(branchName)
+	for branchName, branch := range branches.Snapshot() {
+		if branch.Remote == remoteName {
+			branches.Delete(branchName)
 		}
 	}
 }
--- a/go/libraries/doltcore/env/repo_state.go
+++ b/go/libraries/doltcore/env/repo_state.go
@@ -232,6 +232,11 @@ func (rs *RepoState) AddRemote(r Remote) {
 
 func (rs *RepoState) RemoveRemote(r Remote) {
 	rs.Remotes.Delete(r.Name)
+	for branchName, branch := range rs.Branches.Snapshot() {
+		if branch.Remote == r.Name {
+			rs.Branches.Delete(branchName)
+		}
+	}
 }
 
 func (rs *RepoState) AddBackup(r Remote) {
--- a/go/libraries/doltcore/env/repo_state.go
+++ b/go/libraries/doltcore/env/repo_state.go
@@ -232,6 +232,9 @@ func (rs *RepoState) AddRemote(r Remote) {
 
 func (rs *RepoState) RemoveRemote(r Remote) {
 	rs.Remotes.Delete(r.Name)
+	if rs.Branches == nil {
+		return
+	}
 	for branchName, branch := range rs.Branches.Snapshot() {
 		if branch.Remote == r.Name {
 			rs.Branches.Delete(branchName)
--- a/go/libraries/doltcore/env/repo_state_test.go
+++ b/go/libraries/doltcore/env/repo_state_test.go
@@ -1,0 +1,46 @@
+// Copyright 2026 Dolthub, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package env
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
+	"github.com/dolthub/dolt/go/libraries/utils/concurrentmap"
+)
+
+func TestRemoveRemoteClearsTrackingBranches(t *testing.T) {
+	remotes := concurrentmap.New[string, Remote]()
+	remotes.Set("origin", Remote{Name: "origin"})
+	branches := concurrentmap.New[string, BranchConfig]()
+	branches.Set("tracked", BranchConfig{
+		Merge:  ref.MarshalableRef{Ref: ref.NewBranchRef("main")},
+		Remote: "origin",
+	})
+	branches.Set("other", BranchConfig{
+		Merge:  ref.MarshalableRef{Ref: ref.NewBranchRef("main")},
+		Remote: "upstream",
+	})
+
+	repoState := RepoState{Remotes: remotes, Branches: branches}
+	repoState.RemoveRemote(Remote{Name: "origin"})
+
+	_, found := repoState.Branches.Get("tracked")
+	require.False(t, found)
+	_, found = repoState.Branches.Get("other")
+	require.True(t, found)
+}
--- a/go/libraries/doltcore/env/repo_state_test.go
+++ b/go/libraries/doltcore/env/repo_state_test.go
@@ -44,3 +44,15 @@ func TestRemoveRemoteClearsTrackingBranches(t *testing.T) {
 	_, found = repoState.Branches.Get("other")
 	require.True(t, found)
 }
+
+func TestRemoveRemoteWithUninitializedBranches(t *testing.T) {
+	remotes := concurrentmap.New[string, Remote]()
+	remotes.Set("origin", Remote{Name: "origin"})
+	repoState := RepoState{Remotes: remotes}
+
+	require.NotPanics(t, func() {
+		repoState.RemoveRemote(Remote{Name: "origin"})
+	})
+	_, found := repoState.Remotes.Get("origin")
+	require.False(t, found)
+}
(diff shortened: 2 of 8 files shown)


CBOSSX:
I inspected the failing checks. Six are fork-permission / required-label workflow failures. The macOS sql-server integration job failed because `/tmp/mysql.sock` already existed (`listen unix /tmp/mysql.sock: bind: file exists`), after which the test could not connect to the server. This is outside the two changed `RepoState` files, and the same workflow has passed on adjacent PRs. I do not have repository permission to rerun the job (GitHub returns 403). Could a maintainer rerun the failed macOS job?

go/libraries/doltcore/env/repo_state.go
```diff
@@ -232,6 +232,11 @@ func (rs *RepoState) AddRemote(r Remote) {
 
 func (rs *RepoState) RemoveRemote(r Remote) {
 	rs.Remotes.Delete(r.Name)
+	for branchName, branch := range rs.Branches.Snapshot() {
```
elianddb: There's potential for `rs.Branches == nil` causing a panic.

elianddb: This manual branch iteration sets `branch.Remote = ""` before calling `RemoveRemote`:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/go/libraries/doltcore/sqle/dprocedures/dolt_remote.go#L175-L185

  When `RemoveRemote` runs, `branch.Remote == r.Name` is already false:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/go/libraries/doltcore/env/repo_state.go#L236-L238

This causes the cleanup in `RepoState.RemoveRemote` to be completely bypassed in production.

The root cause of callers bypassing domain rules is that `RepoState.Remotes` and `RepoState.Branches` are exposed as public mutable maps:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/go/libraries/doltcore/env/repo_state.go#L72-L74

Callers across the codebase can directly invoke `.Delete()` or `.Set()` on these fields, bypassing the invariant cleanup added in this PR.

elianddb: `SessionStateAdapter.RemoveRemote` directly calls `repoState.Remotes.Delete(name)`:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/go/libraries/doltcore/sqle/dsess/session_state_adapter.go#L200

  This bypasses the tracking cleanup logic in `repoState.RemoveRemote(remote)`. It should delegate to `repoState.RemoveRemote(remote)` and synchronize `s.branches` in session state.

go/libraries/doltcore/env/repo_state_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/dolthub/dolt/go/libraries/doltcore/ref"
+	"github.com/dolthub/dolt/go/libraries/utils/concurrentmap"
+)
+
+func TestRemoveRemoteClearsTrackingBranches(t *testing.T) {
```
elianddb, line 26: Add a unit test verifying that calling `RemoveRemote` when branches are uninitialized does not panic.

elianddb, line 26: Add an assertion verifying that after removing a remote via CLI/SQL, `dolt status` does not report the branch as up to date with itself:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/integration-tests/bats/remote-cmd.bats#L121-L125

go/libraries/doltcore/env/repo_state.go
elianddb: Remove or update this TODO comment:
  https://github.com/dolthub/dolt/blob/cb23b4be4c1167f2ef9dbcb2eb81db106805a04d/go/libraries/doltcore/env/remotes.go#L811-L812

Deleting the branch config entry upon remote removal directly addresses this item.

elianddb (changes requested):
I have a couple of comment on implementation and testing. We want to test our interfaces for the end user (SQL/CLI) to make sure changes make it downstream.

go/libraries/doltcore/env/repo_state.go
CBOSSX: Added `TestRemoveRemoteWithUninitializedBranches` and a nil guard in `RepoState.RemoveRemote`; the test failed with the reported panic before the guard and now passes.

CBOSSX: Removed the procedure-level branch rewrite. CLI and SQL removal now delegate the invariant to `RepoState.RemoveRemote`, so the tracked branch config is deleted instead of pre-cleared and bypassed.

CBOSSX: `SessionStateAdapter.RemoveRemote` now delegates persisted cleanup to `repoState.RemoveRemote(remote)` and removes the same tracking entries from `s.branches` after the repo state save succeeds.

go/libraries/doltcore/env/repo_state_test.go
CBOSSX: Added the requested uninitialized-branches regression. It reproduced the nil-pointer panic before the fix and passes with the guard.

CBOSSX: Expanded the CLI BATS assertion and added a SQL removal case. Both failed before with `Your branch is up to date with main` and now pass; the full 9-test `remote-cmd.bats` suite also passes.

go/libraries/doltcore/env/repo_state.go
CBOSSX: Removed the TODO now that remote removal deletes the matching branch config entries across the CLI and SQL paths.

go/libraries/doltcore/env/repo_state.go
```diff
@@ -232,6 +232,14 @@ func (rs *RepoState) AddRemote(r Remote) {
 
 func (rs *RepoState) RemoveRemote(r Remote) {
 	rs.Remotes.Delete(r.Name)
+	if rs.Branches == nil {
```
elianddb: Move into `ClearTrackingBranches` helper in this file so you can share with the session state adapter.

go/libraries/doltcore/sqle/dsess/session_state_adapter.go
```diff
@@ -197,8 +196,18 @@ func (s SessionStateAdapter) RemoveRemote(_ context.Context, name string) error
...
 		return env.ErrRemoteNotFound
 	}
-	repoState.Remotes.Delete(name)
-	return repoState.Save(fs)
+	repoState.RemoveRemote(remote)
+	if err := repoState.Save(fs); err != nil {
+		return err
+	}
+
+	s.remotes.Delete(remote.Name)
```
elianddb, line 204: nil check missing for branches, however use the helper recommendation from above.

integration-tests/bats/remote-cmd.bats
```diff
@@ -122,6 +122,36 @@ teardown() {
...
+
+    run dolt status
+    [ "$status" -eq 0 ]
+    [[ "$output" =~ "Your branch is up to date with 'origin/main'" ]] || false
+
+    dolt sql -q "CALL DOLT_REMOTE('remove','origin')"
+
+    run dolt status
+    [ "$status" -eq 0 ]
+    [[ ! "$output" =~ "origin" ]] || false
```
elianddb, line 151: line `152` is redundant since `origin` is included in `153`

elianddb (changes requested):
Looks good! I have a couple of small changes.

go/libraries/doltcore/env/repo_state.go
CBOSSX: Addressed in 778848d44: both RepoState.RemoveRemote and SessionStateAdapter.RemoveRemote now call the shared ClearTrackingBranches helper. Affected Go suites and remote/pull/sql-pull BATS pass.

go/libraries/doltcore/sqle/dsess/session_state_adapter.go
CBOSSX: Addressed in 778848d44: the session adapter now uses the shared nil-safe helper. Added TestClearTrackingBranchesWithUninitializedBranches and reran affected Go suites.

integration-tests/bats/remote-cmd.bats
CBOSSX: Addressed in 778848d44: removed the redundant origin/main assertion in both CLI and SQL tests, retaining the broader origin assertion and the self-tracking main assertion. The 9 remote-cmd tests pass.

CBOSSX:
Addressed the three review comments in 778848d44. Both persisted and session tracking cleanup now share the nil-safe ClearTrackingBranches helper. The custom-remote pull tests also now fetch and explicitly track test-remote/main after removing origin, since deleting origin correctly deletes its tracking entry.

Validation: affected env, dsess, and command Go suites; CLI build; vet and repository formatting; and all 103 remote-cmd/pull/sql-pull BATS cases (102 passed, one mode-specific skip). The two custom-remote cases failed before this fixture correction and pass now.

The previous CI logs also contain Lambda unexpected-EOF responses and local remotesrv connection-refused failures. The benchmark dispatch/checkout and label-comment failures require repository workflow permissions. Please re-review the updated head and enable the repository-owned checks as needed.

PR #11740, opened 2026-09-07
nbs: bound conjoin's table-file open concurrency

### LINKED ISSUE #11739: conjoin opens unbounded concurrent connections — exceeds sshd MaxStartups on git+ssh remotes, so conjoin never completes
### Summary

`conjoinTables` in `go/store/nbs/conjoiner.go` starts one goroutine per conjoinee with no
concurrency limit. For a persister whose `Open` is a network operation this fans out to an
unbounded number of simultaneous connections. Against a **Git remote over SSH** each `Open`
costs its own SSH connection, so a single conjoin exceeds the server's limit on concurrent
*unauthenticated* connections (OpenSSH `MaxStartups`, default `10:30:100`). The server drops
the excess pre-auth, conjoin fails, and the table-file count never comes down.

The push itself succeeds — only the compaction fails — so the store silently degrades.

### Environment

- Transport: Dolt Git remote, `git+ssh://` to a self-hosted bare repo
- Server: OpenSSH with stock `MaxStartups 10:30:100`, `persourcemaxstartups none` (confirmed via `sshd -T`)
- Observed through an application embedding the Dolt Go module (`github.com/dolthub/dolt/go v0.40.5-0.20260715172757-a6690826d767`)
- The unbounded fan-out is unchanged on `main` at time of writing

### Observed behaviour

Every push logs, at warning level:

```
level=info    msg="beginning conjoin of database" pkg=store.noms upstream_len=2259
level=warning msg="conjoin of database failed with error"
              error="kex_exchange_identification: read: Connection reset by peer
                     Connection reset by <redacted> port <ssh-port>
                     fatal: Could not read from remote repository."
Push complete.
```

`kex_exchange_identification` places the failure in **key exchange** — before authentication
and before any payload — which is exactly what `MaxStartups` random-drop produces. Credentials
are not involved: sequential connections succeed 12/12.

### Measurement

Concurrent TCP connections to the remote's SSH port, sampled locally at ~5ms resolution
(5,310 samples) across one conjoin-triggering push:

| Metric | Value |
|---|---|
| Peak concurrent connections | **18** |
| Samples at/above `MaxStartups` threshold of 10 | **72** of 5,310 |
| Total connections opened during the push | **~247** |
| Same metric for a no-op push (no conjoin) | **2** |

Burst shape, consecutive ~5ms samples: `3 → 6 → 9 → 12 → 15 → 15 → 14 → 13 → 8 → 3`. Roughly
150ms wide, and repeated — the threshold is crossed many times within one push, peaking at
nearly double it.

A graduated concurrency probe against the same server reproduces the server-side behaviour
independently: 5 concurrent connections → 0 failures; 15 → 1 failure; 30 → 10 failures.

### Root cause

`go/store/nbs/conjoiner.go`, `conjoinTables`:

```go
eg, ectx := errgroup.WithContext(ctx)
toConjoin := make(chunkSources, len(conjoinees))

for idx := range conjoinees {
    i, spec := idx, conjoinees[idx]
    eg.Go(func() (err error) {
        toConjoin[i], err = p.Open(ectx, spec.name, spec.chunkCount, stats)
        return
    })
}
```

One goroutine per conjoinee; `SetLimit` is not called anywhere in the file. The fan-out is
bounded only by the size of the conjoinee set, which is itself a function of how far behind
compaction has fallen.

### Why this surfaces now

Unbounded fan-out is harmless for the HTTP remotesapi, where connection pooling and HTTP/2
multiplexing absorb it. Git remotes as Dolt remotes shipped in February 2026, and over
`git+ssh` each `Open` becomes a distinct SSH connection — so the same loop now presents as a
connection storm. The feature's design write-up does not discuss transfer concurrency.

### Impact — it is self-reinforcing

NBS triggers a conjoin once active table files exceed its limit (256). Each failed conjoin
leaves the count higher, so the next attempt has more conjoinees and a wider fan-out, and is
more likely to fail. The store cannot recover unaided. On our store the count has gone
1,962 → 2,113 → 2,259 over three days — roughly 9× the conjoin threshold — and pulls now take
**10+ minutes**, lengthening steadily.

### Proposed fix

Bound the fan-out with `eg.SetLimit(...)`. A default of 8 sits below the stock `MaxStartups`
of 10 and is ample for local persisters, where opening a table file is dominated by the index
read. PR to follow.

Happy to make the limit configurable instead, or to scope it to remote persisters only, if
either is preferred.

rjc123:
Fixes #11739.

`conjoinTables` starts one goroutine per conjoinee with no limit. Where the persister's `Open`
is a network operation this fans out to an unbounded number of simultaneous connections; over a
Git remote reached by SSH each one is a separate SSH connection, so a single conjoin can exceed
the server's `MaxStartups` (default `10:30:100`) and have its connections dropped during key
exchange. The conjoin then fails while the push reports success, so the table-file count grows
without bound and never compacts.

Measured against a self-hosted Git remote over SSH, sampling concurrent connections locally at
~5ms: **peak 18 concurrent, ~247 connections opened in a single conjoin push**, versus 2 for a
push that triggers no conjoin. 72 of 5,310 samples sat at or above the limit of 10.

This adds `eg.SetLimit(conjoinOpenConcurrency)` with a default of 8 — below the stock
`MaxStartups` so the default configuration works, and ample for local persisters where opening
a table file is dominated by the index read.

Happy to make the value configurable, or to apply the limit only for remote persisters, if
either is preferred.

**Testing:** `go vet ./store/nbs/` clean; `go test ./store/nbs/ -count=1` passes (30.9s).

**Diff:**

```go
+// conjoinOpenConcurrency bounds how many table files conjoinTables opens at
+// once. Opening a table file is a network round trip for a remote persister,
+// and for a Git remote reached over SSH it costs a separate SSH connection.
+// An unbounded fan-out here therefore scales with the number of conjoinees and
+// can exceed the server's limit on concurrent unauthenticated connections
+// (sshd's MaxStartups defaults to 10), which drops the excess before
+// authentication and fails the conjoin.
+//
+// Local persisters are not meaningfully constrained by this: opening table
+// files is dominated by the index read, and eight in flight already saturates
+// a local disk.
+const conjoinOpenConcurrency = 8
+
 func conjoinTables(...) {
 	eg, ectx := errgroup.WithContext(ctx)
+	eg.SetLimit(conjoinOpenConcurrency)
 	toConjoin := make(chunkSources, len(conjoinees))
```

### DIFF
--- a/go/store/nbs/conjoiner.go
+++ b/go/store/nbs/conjoiner.go
@@ -284,8 +284,22 @@ func conjoin(ctx context.Context, behavior dherrors.FatalBehavior, s conjoinStra
 	return mc, op.conjoinedSrc, cf, nil
 }
 
+// conjoinOpenConcurrency bounds how many table files conjoinTables opens at
+// once. Opening a table file is a network round trip for a remote persister,
+// and for a Git remote reached over SSH it costs a separate SSH connection.
+// An unbounded fan-out here therefore scales with the number of conjoinees and
+// can exceed the server's limit on concurrent unauthenticated connections
+// (sshd's MaxStartups defaults to 10), which drops the excess before
+// authentication and fails the conjoin.
+//
+// Local persisters are not meaningfully constrained by this: opening table
+// files is dominated by the index read, and eight in flight already saturates
+// a local disk.
+const conjoinOpenConcurrency = 8
+
 func conjoinTables(ctx context.Context, behavior dherrors.FatalBehavior, conjoinees []tableSpec, p tablePersister, stats *Stats) (conjoined tableSpec, src chunkSource, cleanup cleanupFunc, err error) {
 	eg, ectx := errgroup.WithContext(ctx)
+	eg.SetLimit(conjoinOpenConcurrency)
 	toConjoin := make(chunkSources, len(conjoinees))
 
 	for idx := range conjoinees {


reltuk:
Quick feed back here – I think if we accepted this, we would need to do it only for something like Git remotes – I guess key exchange there is expensive enough and the transport isn't multiplexing.

With most remotes, we want to fan this out quite aggressively. But this also raises a question in my mind, because when we go to conjoin, we typically already have a table set open and it will always have the vast majority of the targeted table files already open and cloneable within it. Maybe the right approach here is to thread the lifecycle of the already-open table files through so that we don't have to open so many table files?

rjc123:
Thanks — and agreed on both counts. Your second point is the better fix, and I think the numbers support it.

**On scoping.** Agreed, and I'd go further: I don't think this is a general Dolt concurrency question at all. It's a bug in the Git-remote path specifically, where the assumption that `Open` is cheap and multiplexed stops holding. Restricting any limit to Git remotes seems right, and leaving aggressive fan-out alone everywhere else.

**On reusing the already-open table set.** The measurements suggest conjoin is not currently reusing it. Sampling concurrent connections at ~5ms against the same store:

| | server limit `10:30:100` (stock) | server limit `100:30:200` |
|---|---|---|
| Peak concurrent connections | 18 | **445** |
| Samples at/above 200 | — | 141 |

The peak of 18 under the stock limit was just the server suppressing it. Given headroom, one conjoin opens 445+ concurrently — and that is still a floor, since `100:30:200` drops connections above 200. For a store with ~2,260 table files, opening hundreds fresh is consistent with your hypothesis that the already-open set isn't being threaded through.

If that lifecycle were plumbed in, most of those opens would disappear and the concurrency question becomes moot for every transport rather than being papered over for one. That's clearly the better outcome.

**Where I can help.** Happy to either:

- narrow this PR to apply the limit only for Git remotes, as an interim while the lifecycle work is scoped; or
- close it in favour of the table-set-reuse approach.

Your call entirely — I don't know the persister lifecycle well enough to attempt the second unaided, but I'm glad to try if you can point me at where the open table set is reachable from `conjoinTables`.

For context on where this is coming from: I'm a downstream consumer rather than a Dolt developer. [beads](https://github.com/gastownhall/beads) embeds the Dolt Go module and syncs its issue database over a `git+ssh://` remote, which is how this surfaced — conjoin has been failing on every push for months, so the table-file count has grown to ~9x the conjoin threshold and pulls now take 10+ minutes. Tracked downstream at gastownhall/beads#6359. I'm happy with whatever shape of fix you think is right.

reltuk:
Do you still have a repro available? I would love to hear the results of a test using a version of Dolt built to include

https://github.com/dolthub/dolt/pull/11742

and

https://github.com/dolthub/dolt/pull/11745

rjc123:
Not the original one — that store compacted itself out of existence. Once I raised the server's
`MaxStartups`, a conjoin finally completed, the backlog collapsed from ~2,264 table files to 32,
and conjoin no longer fires at all. So I built a synthetic one instead. Results below; the short
version is that #11742 halves conjoin's Git work, and that I could **not** reproduce the
connection storm locally, for a reason I think is worth knowing.

**Setup.** Both binaries built from the same tree at `031742f11a`: `main` unmodified, versus
`main` + #11742 + #11745 (both merged cleanly). A scratch database pushed 128 times to a
`git+file://` remote to reach 256 table files, snapshotted, and restored between runs so each
binary saw a byte-identical store. The measured push takes it to 257 and triggers conjoin.

**Measurement.** Since the Git blobstore shells out to `git` (`exec.LookPath` in
`internal/git/runner.go`), every table-file open is a separate `git` invocation — one SSH
connection each on a `git+ssh://` remote. So I put a shim named `git` first on `PATH` and logged
start/end timestamps for every invocation. That counts the fan-out exactly rather than sampling
it, which is how I measured it the first time.

| | `main` | +#11742/#11745 |
|---|---|---|
| Total `git` subprocesses | 1,657 | **1,137** (−31%) |
| of which `cat-file` | 1,562 | 1,048 |
| `git fetch` | 9 | 6 |
| Push wall time | 8.3–8.6s | 6.5–7.7s |
| Outcome | 257 → 2 table files | 257 → 2 table files |

Splitting that on your own `beginning conjoin` / `conjoin completed` log timestamps attributes it
cleanly:

| `git` subprocesses started | `main` | patched |
|---|---|---|
| **during** the conjoin window | 1,110 | **570** (−49%) |
| outside it | 547 | 567 (unchanged) |

Four runs of each; the counts were identical every time.

**What I could not reproduce, and I don't want to overstate this.** The connection storm itself
never appeared: peak concurrent `git fetch` was **1** on both binaries, even with 150ms of
latency injected into every fetch to imitate an SSH round trip. So this run says nothing
quantitative about connections — only about opens.

I think the reason is `syncForRead`'s dedup window, `defaultSyncForReadTTL = 1s`. My push runs
8.4s and issues 9 fetches — one per expiry, near enough exactly. The conjoin that originally
failed ran for **ten minutes plus**, so it crossed that window ~600 times, each crossing
releasing whatever readers were waiting. That matches your #11745 note about the window never
renewing against a quiet remote, and it means the storm needs a long, slow conjoin to show up at
all. An 8-second local one can't.

**One other limit worth flagging:** I couldn't test whether the effect scales, because a working
conjoin won't let a backlog accumulate — it fires at 257, compacts to 2, and climbs again in a
sawtooth. A deep backlog only exists when conjoin is *failing*, which is how the real store
reached 2,264. So 257 files is the largest honest local data point, and it is the mildest end of
the range.

I also deliberately haven't quoted peak concurrency: it varied between 74 and 103 for the *same*
binary on the *same* store, so any apparent difference there is scheduling noise, not signal.

Happy to go further if it's useful — I can point this at a real SSH remote, or force conjoin to
fail while building so there's a genuinely deep backlog to test against. And #11740 still looks
redundant next to these two; glad to close it whenever you'd like.

reltuk:
Thank you for the follow up here! Both #11742 and #11745 landed in main.

I agree there are probably still lots of improvements we could make here, but I'm not sure hard limiting the fan out of the Opens at this layer is the right approach. I'm going to go ahead and close this PR.

I think there might still be work to do in this space. If you have interactions which seem too chatty or which are slower than they should be, please don't hesitate to follow up in issues and/or separate PRs. I'm happy to keep investigating and improving things here.

Thanks for your contribution and thanks for your interest in the project!

PR #11873, opened 2026-09-16
go/store/nbs: Fix an issue in how ValueStore and NodeStore chunk caches interact with unlocked reads during GC.

reltuk:
During an unlocked read, the `keeper` will be `nil` when no GC was going on when the read started. `BeginGC` would happily install a new `keeper` while such reads were outstanding. Then `PurgeCaches` would clear the existing ValueStore and NodeStore caches. But chunks coming back from those unlocked reads which started before the `keeper` was installed could then get inserted into the now-cleared caches.

Serialize the `keeper` installation after all outstanding unlocked reads, so that every chunk returned from the NomsBlockStore from that point forward is correctly registered as a GC dependency. In addition, add some machinery so that the caches do not cache things which may have been read before the purge.


coffeegoddd:
@reltuk __DOLT__
| covering_index_scan  | 2.3          | 2.3        | 0.0            |
| groupby_scan         | 65.65        | 64.47      | -1.8           |
| index_join           | 1.93         | 1.93       | 0.0            |
| index_join_scan      | 1.32         | 1.32       | 0.0            |
| index_scan           | 204.11       | 204.11     | 0.0            |
| oltp_point_select    | 0.25         | 0.26       | 4.0            |
| oltp_read_only       | 5.0          | 5.0        | 0.0            |
| select_random_points | 0.51         | 0.51       | 0.0            |
| select_random_ranges | 0.64         | 0.65       | 1.56           |
| table_scan           | 204.11       | 200.47     | -1.78          |
| types_table_scan     | 475.79       | 467.3      | -1.78          |


| oltp_delete_insert    | 6.09         | 6.09       | 0.0            |
| oltp_insert           | 3.07         | 3.02       | -1.63          |
| oltp_read_write       | 11.24        | 11.24      | 0.0            |
| oltp_update_index     | 3.3          | 3.3        | 0.0            |
| oltp_update_non_index | 3.02         | 2.97       | -1.66          |
| oltp_write_only       | 6.21         | 6.21       | 0.0            |

coffeegoddd:
@reltuk __DOLT__
<table>
</TR>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>result</TH>
<TH>total</TH>
</TR>
<TR><TD>a23fea6</TD>
<TD>ok</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
<TR><TH>version</TH>
<TH>total_tests</TH>
</TR>
<TR><TD>a23fea6</TD>
<TD>5937471</TD>
</TR>
</table>
<table>
</TR>
<TR><TD>100.0</TD>
</TR>
</table>

coffeegoddd:
@reltuk __DOLT__
| tpcc-scale-factor-1 | 46.63            | 47.47          | 1.8            |


| tpcc-scale-factor-1 | dolt             | 6333e4ba10d0a5ab24fefeecd9a84dab78d91969 | 52.4     | dolt           | a23fea69e5d2a2e13d5d8ea2dfa6aa2ef48c2194 | 51.99  | -0.78          |

go/store/util/sizecache/size_cache.go
```diff
@@ -45,6 +45,14 @@ type SizeCache struct {
 	totalSize uint64
 	maxSize   uint64
 	mu        sync.Mutex
+
+	// Incremented by every Purge. Read out of Get and rechecked by Add,
+	// both under mu, so that a value which was fetched from the backing
+	// store before a Purge cannot be cached after it.
+	//
+	// The read which fetched it took no GC read dependency on it, so
```
nicktobey, line 53: The fact that this comment mentions GC in a component that is lower-level than GC feels like a code smell. It's unclear looking at SizeCache why the output of Get would be passed back to Add, why Add needs to take a generation, why we need to discard attempted inserts from a prior generation, etc. It means that there's now a possibility that a line like `c.Add(hashFromString("data-1"), 200, "data-1", c.Generation())` might experience a purge between the call to c.Generation() and c.Add()`

go/store/nbs/store.go
```diff
@@ -123,12 +123,20 @@ type NomsBlockStore struct {
...
+	// unlocked read is in progress.
+	outstandingReads int
+
+	// Set while BeginGC is waiting for |outstandingReads| to drain so
+	// that it can install its keeper. New reads block rather than start
+	// while this is set. This prevents starvation on installing a new
+	// keeper when reads are constantly arriving.
+	gcInstallPending bool
+
+	mu sync.RWMutex // protects the current nbs state
```
nicktobey: Is this comment adding anything?

go/store/util/sizecache/size_cache.go
reltuk: An alternative is to make it something like

```go
Get(key string) (interface{}, Key, bool)
Put(key Key, interface{})
```

And hide the generation inside the Key.

It's definitely subtle that the Put is by-necessity tied to a previous Get that returned negative-presence and that that negative-presence's validity can change out from under the client while they are materializing the result to be stored at the key.

reltuk: But, to clarify, the intent was to make them miss the Put and it's totally fine if they do. If there was a concurrent Purge, it's the same as if the entire Add() was sequenced before the Purge, right?

reltuk:
Gonna break this up into some independent pieces...

